PR #3, opened 2019-05-01
Implement more TIR data structures.

vext01:
This is a companion commit to:
https://github.com/softdevteam/ykrustc/pull/18

### DIFF
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -10,12 +10,17 @@
 //! Types for the Yorick intermediate language.
 
 use serde::{Deserialize, Serialize};
-use std::fmt::{self, Display};
+use std::{
+    fmt::{self, Display},
+    marker::PhantomData,
+};
 
 pub type CrateHash = u64;
 pub type DefIndex = u32;
 pub type BasicBlockIndex = u32;
 pub type LocalIndex = u32;
+pub type VariantIndex = u32;
+pub type PromotedIndex = u32;
 
 /// A mirror of the compiler's notion of a "definition ID".
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
@@ -93,7 +98,8 @@ impl Display for BasicBlock {
 pub enum Statement {
     Nop,
     Assign(Place, Rvalue),
-    Unimplemented, // FIXME
+    SetDiscriminant(Place, VariantIndex),
+    Unimplemented,
 }
 
 impl Display for Statement {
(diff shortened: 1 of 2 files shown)


ykpack/src/types.rs
```diff
@@ -93,7 +96,16 @@ impl Display for BasicBlock {
 pub enum Statement {
     Nop,
     Assign(Place, Rvalue),
-    Unimplemented, // FIXME
+    SetDiscriminant{place: Place, variant_index: VariantIndex},
```
ltratt: To fit in with the other members of the `enum`, should this be `SetDiscriminant(Place, VariantIndex)`?

ykpack/src/types.rs
```diff
@@ -102,19 +114,112 @@ impl Display for Statement {
...
-pub enum Rvalue {
+pub struct Projection<B, V> {
+    pub base: B,
+    pub elem: ProjectionElem<V>,
+}
+
+/// Describes a projection operation upon a projection base.
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
+pub enum ProjectionElem<V> {
+    Unimplemented(PhantomData<V>), // FIXME
```
ltratt: Why fixme?

ykpack/src/types.rs
vext01: Yeah, I've been pondering on this myself. In MIR it's a random mix.

Ideally I'd use only tuple-like variants, but I know I'd be upset that I'd have to break my rule if we had a variant with lots of the same type.

E.g.:
```
enum Taps {
    ...
    Mixer(usize, usize, usize, usize),
}
```

Here a struct-like would be preferred, so that we can name the fields.

Thoughts?

ykpack/src/types.rs
vext01: There's supposed to be another structure in there, but I don't want to implement it just yet.

The missing struct is parameterised by V.

ykpack/src/types.rs
ltratt: I agree that when you get repetitions of the same type you want to use names, but that's not the case here. [I also sometimes use names when the type doesn't give much away about what the "thing" is.] Up to you.

vext01: OK

vext01:

ltratt:
Please squash.

vext01:
My bad, the tests were broken, and I also detected rogue tabs.

Still OK to squash?

ltratt:
The new commits look fine. Please squash.

vext01:
Squashed. Left the tab fixes separate, because they applied to different lines to what was touched before.

ltratt:
You can try using `git absorb` to assign the tab fixes to the right commits. It can't always work, but it's pretty good.

vext01:
Yeah, it's on my to do list to look at.

ltratt:
You can try:

```
$ cargo install git-absorb
$ git reset --soft "HEAD^"
$ git absorb
```

then manually check what it's done before doing:

```
$ git rebase --autosquash origin/master
```

I think it's worth trying before we merge this. If it's a disaster, we can merge as-is.

vext01:
I'll give it a go, but I expect it to be a no-op, since I already squashed.

ltratt:
If those tab changes happened in this PR, it is li-kely do the right thing. [Sometimes it doesn't, but not often.]

vext01:
As expected, it's a no-op. Let's r+?

vext01:
Wait, it does something I dont understand. Can we merge and I'll look at this another day.

ltratt:
Ah, I might have missed this:

>Left the tab fixes separate, because they applied to different lines to what was touched before.

I think you're saying "the tabs fix is for stuff that wasn't changed in this PR"? If so, yes, git can't work for this case.

vext01:
The tab stuff was in this PR, but it was already in a separate commit how I desired it :P

ltratt:
What I meant was: I think you're changing tabs to spaces in code that wasn't touched by the first commit. [Indeed, I've just checked and this is the case.]

bors r+

vext01:
Can I squash that?

ltratt:
Please squash.

vext01:
splat

PR #5, opened 2019-05-08
Implement yet more TIR.

vext01:
This change implements more TIR structures in ykpack.

I'm going to point out a couple of areas for discussion.

### DIFF
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -10,10 +10,7 @@
 //! Types for the Yorick intermediate language.
 
 use serde::{Deserialize, Serialize};
-use std::{
-    fmt::{self, Display},
-    marker::PhantomData,
-};
+use std::fmt::{self, Display};
 
 pub type CrateHash = u64;
 pub type DefIndex = u32;
@@ -21,6 +18,7 @@ pub type BasicBlockIndex = u32;
 pub type LocalIndex = u32;
 pub type VariantIndex = u32;
 pub type PromotedIndex = u32;
+pub type FieldIndex = u32;
 
 /// A mirror of the compiler's notion of a "definition ID".
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
@@ -112,7 +110,7 @@ impl Display for Statement {
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
 pub enum Place {
     Base(PlaceBase),
-    Projection(PlaceProjection),
+    Projection(Projection),
 }
 
 /// The "base" of a place projection.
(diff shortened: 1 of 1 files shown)


ykpack/src/types.rs
```diff
@@ -133,14 +131,65 @@ pub struct PlaceProjection {
...
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
+pub enum Constant {
+    Scalar(Scalar),
+    Unimplemented,
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
+pub enum Scalar {
+    Bits {
```
vext01: We've already spoken about this. We are going to use an enum for the different numeric types.

Notice how signedness isn't mentioned. I assume in MIR that'd be in the `Ty`, which we don't model.

So How about we have two numeric constant variants, signed and unsigned?

ltratt: I'm not too worried about how we do it at the moment, TBH: we're bound to learn from experience, so do something that seems sensible and we can iterate as we learn more.

ykpack/src/types.rs
```diff
@@ -133,14 +131,65 @@ pub struct PlaceProjection {
 /// Describes a projection operation upon a projection base.
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
 pub enum ProjectionElem<V> {
```
vext01: The parameterisation here irritates me. Especially since `ProjectionElem` is only ever instantiated with `V` as either `()` or `Local`. Toying with flattening it out...

vext01:
The last commit implements the changes to constant integers.

So whereas we would have had:
```
Constant(Scalar(Bits { size: 4, hi_word: 0, lo_word: 100 }))
```

Now we have:
```
Constant(SignedInt(I32(100)))
```

ykpack/src/types.rs
vext01: Projection elements are sub-views on aggregate structures.

So if we have `mystruct.f`, then `mystruct` is the projection "Base" and `f` is the "Element".

Similar notions for slices, indexing and derefs.

I'm going to have a play and see if we can come up with something a little more concise that what MIR does.

vext01: [Taking a quick detour to get my upstream Rust change merged, because it will affect the TIR lowering]

ltratt: OK.

vext01:
The last commit gets rid of parameterised projections.

I think we can review this now.

A companion PR for the compiler coming soon...

ykpack/src/types.rs
```diff
@@ -155,38 +155,62 @@ pub enum Operand {
...
 }
 
-impl Scalar {
-    pub fn bits_from_u128(size: u8, val: u128) -> Self {
-        Scalar::Bits {
-            size,
-            hi_word: (val >> 64) as u64,
-            lo_word: val as u64,
+impl UnsignedInt {
+    pub fn from_u128(val: u128) -> Self {
```
ltratt: Are we saying that every `UnsignedInt` will be created from a `u128` or ...?

vext01: No. Just there's a helper for the `u128` (and `i128`) variant since it needs to be split into two for serialisation. Other types are just directly created using the regular constructor.

Why do they need to be split?:
https://github.com/3Hren/msgpack-rust/issues/169

ltratt: OK.

ltratt:
Are we ready to squash? If so, go ahead.

vext01:
splat

vext01:
My fault. I'm going to try to remember to do that in the future.

Squash?

ltratt:
Please squash.

vext01:
splat.

PR #7, opened 2019-05-14
First attempt at TIRv2.

vext01:
Hi,

Here are our first baby steps with "TIR v2".

To demonstrate how it might work, consider the Rust code:
```
struct Point {
    x: u32,
    y: u32,
}
...
let mut p = Point {x: 25, y: 50};
p.y += 100;
```

I can imagine this lowering to (supposing this is a textual representation of the in-memory structs):
```
$1 = Alloca(64);
$2 = AddressOf(LocalIndex);
MemStore($2, 0, 25u32, 4);
MemStore($2, 4, 50u32, 4);
$3 = MemLoad($2, 4, 4);
$4 = AddU32($3, 100u32);
MemStore($2, 4, $4, 4);
```

There are lots of design decisions to take into account here. Please comment on the following...

## Should we have Explicit Variable Types?

Initially I'd set out to have each TIR local variable typed, as PyPy does, e.g.:
```
%i0 = ...
%p0 = ...
```

However, I went off the idea because it's kind of irritating to carry around the type information with the variable. I'd considered thing like integer tagging and `(type, index)` pairs, but neither option feels right somehow. It's much easier if a variable is just a number.

The proposed TIR essentially contains *no* explicit type information. What do you think? Does the JIT runtime need the types of things to hand?

## Correctness

The proposed TIR is flexible, but the encoding allows for nonsensical programs. For example:
```
$0 = Alloca(1);
$1 = AddressOf($0);
MemStore($1, 0, 100u64, 2); // ptr, offset, val, num_bytes
```

This TIR would be wrong in two ways:
 * We try to write a 64-bit constant to a one-byte variable.
 * We passed a different byte-size (2) than both the constant we are storing and the size of the allocation.

Is that OK, or would be want a TIR which, by construction, can never be incorrect? Is it even possible? I suspect not...

## AddressOf

`AddressOf` seems necessary to make a pointer from a value (and I suppose me might need a deref later). What do you think?

## Typed Operations

One idea I toyed with was the idea of "typed operations", where you can always know the type of the thing on the left-hand side of an assignment from the operation on the right-hand side.

This would mean, for example, that instead of having a single load from memory operation:
```
MemLoad{ptr: Operand, offset: Operand, num_bytes: Operand},
```

We'd perhaps have many:
```
MemLoad_U8{ptr: Operand, offset: Operand},
MemLoad_U16{ptr: Operand, offset: Operand},
MemLoad_U32{ptr: Operand, offset: Operand},
...
```

This would make the set of operations much larger, but makes knowing the types of things at runtime a bit easier. Notice also that the `num_bytes` parameter disappears.

Assigning one variable from another becomes odd though, as there's no operation on the right-hand side that could imply the types. Perhaps we'd have to do something like:
```
$1 = IdU8($2);
```

(i.e. the ID function)

Perhaps one upside of typed operations is that we could encode user struct and enum types more easily, for example:

```
MemLoadADT{ptr: Operand, type: ADTIndex}
```

(ADT = abstract data type -- Rust's name for structs and enums)

Here we've added a `type` operand, which identifies the type of the struct we are loading. There could be a table elsewhere which keeps track of the sizes of each ADT, thus removing the need for a `num_bytes` operand for user structs.

Struct/enum allocas could similarly accept a `ADTIndex` instead of a byte size.

For comparison purposes, the Rust code in the beginning of the PR might lower to the following typed-operations TIR:

```
# Suppose `struct Point` is ADTIndex 33
$1 = AllocaADT(33);
$2 = AddressOfADT(LocalIndex, 33);
MemStoreU32($2, 0, 25u32);
MemStoreU32($2, 4, 50u32);
$3 = MemLoadU32($2, 4);
$4 = AddU32($3, 100u32);
MemStoreU32($2, 4, $4);
```

Here the stores feel clumsy, as there's no lvalue to associate type with.

## Summary

Hopefully it's evident that I'm really not sure what the right thing to do is. I'm keen to hear your ideas.

Thanks -- End of essay ;)

### DIFF
--- a/ykpack/src/lib.rs
+++ b/ykpack/src/lib.rs
@@ -40,7 +40,10 @@ pub use types::*;
 
 #[cfg(test)]
 mod tests {
-    use super::{BasicBlock, Decoder, DefId, Encoder, Pack, Statement, Terminator, Tir};
+    use super::{
+        BasicBlock, Constant, Decoder, DefId, Encoder, Local, Operand, Pack, Rvalue, Statement,
+        Terminator, Tir, UnsignedInt,
+    };
     use fallible_iterator::{self, FallibleIterator};
     use std::io::{Cursor, Seek, SeekFrom};
 
@@ -133,9 +136,40 @@ mod tests {
 
     #[test]
     fn test_text_dump() {
-        let stmts_t1_b0 = vec![Statement::Nop];
+        let stmts_t1_b0 = vec![
+            Statement::Assign(
+                Local::new(0, 0),
+                Rvalue::Operand(Operand::Local(Local::new(1, 0))),
+            ),
+            Statement::Assign(Local::new(2, 0), Rvalue::GetField(Local::new(3, 0), 4)),
+            Statement::Assign(
+                Local::new(4, 0),
+                Rvalue::Operand(Operand::Constant(Constant::UnsignedInt(UnsignedInt::U8(
+                    10,
+                )))),
+            ),
+            Statement::Nop,
+        ];
         let term_t1_b0 = Terminator::Abort;
-        let stmts_t1_b1 = vec![Statement::Unimplemented];
+        let stmts_t1_b1 = vec![
+            Statement::Assign(Local::new(5, 0), Rvalue::Load(Local::new(6, 0))),
+            Statement::Store(Local::new(5, 0), Operand::Local(Local::new(4, 0))),
+            Statement::Assign(
+                Local::new(7, 0),
+                Rvalue::Add(
+                    Operand::Local(Local::new(8, 0)),
+                    Operand::Local(Local::new(9, 0)),
+                ),
+            ),
+            Statement::Assign(
+                Local::new(7, 0),
+                Rvalue::Sub(
+                    Operand::Local(Local::new(9, 0)),
+                    Operand::Local(Local::new(10, 0)),
+                ),
+            ),
+            Statement::Assign(Local::new(11, 0), Rvalue::Alloca(0)),
+        ];
         let term_t1_b1 = Terminator::Goto { target_bb: 50 };
 
         let blocks_t1 = vec![
@@ -148,7 +182,10 @@ mod tests {
             Pack::Tir(Tir::new(
                 DefId::new(3, 4),
                 String::from("item2"),
-                Vec::new(),
+                vec![BasicBlock::new(
+                    vec![Statement::Unimplemented],
+                    Terminator::Unreachable,
+                )],
             )),
         ];
 
@@ -159,19 +196,27 @@ mod tests {
         let got_lines = got.split("\n");
 
         let expect = "[Begin TIR for item1]\n\
-            DefId(1, 2):
-            bb0:
-                Nop
-                term: Abort
-
-            bb1:
-                Unimplemented
-                term: Goto { target_bb: 50 }
-
-        [End TIR for item1]
-        [Begin TIR for item2]
-            DefId(3, 4):
-        [End TIR for item2]\n";
+    DefId(1, 2):
+    bb0:
+        $0: t0 = $1: t0
+        $2: t0 = get_field($3: t0, 4)
+        $4: t0 = U8(10)
+        nop
+        Abort
+    bb1:
+        $5: t0 = load($6: t0)
+        store($5: t0, $4: t0)
+        $7: t0 = int_add($8: t0, $9: t0)
+        $7: t0 = int_sub($9: t0, $10: t0)
+        $11: t0 = alloca(0)
+        Goto { target_bb: 50 }
+[End TIR for item1]
+[Begin TIR for item2]
+    DefId(3, 4):
+    bb0:
+        unimplemented
+        Unreachable
+[End TIR for item2]\n";
 
         let expect_lines = expect.split("\n");
 
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -16,6 +16,34 @@ pub type CrateHash = u64;
 pub type DefIndex = u32;
 pub type BasicBlockIndex = u32;
 pub type LocalIndex = u32;
+pub type TyIndex = u32;
+pub type FieldIndex = u32;
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Copy)]
+pub struct Local {
+    idx: LocalIndex,
+    ty: TyIndex,
+}
+
+impl Local {
+    pub fn new(idx: LocalIndex, ty: TyIndex) -> Self {
+        Self { idx, ty }
+    }
+
+    pub fn idx(&self) -> LocalIndex {
+        self.idx
+    }
+
+    pub fn ty(&self) -> TyIndex {
+        self.ty
+    }
+}
+
+impl Display for Local {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "${}: t{}", self.idx, self.ty)
+    }
+}
 
 /// A mirror of the compiler's notion of a "definition ID".
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
@@ -62,9 +90,12 @@ impl Display for Tir {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         writeln!(f, "[Begin TIR for {}]", self.item_path_str)?;
         writeln!(f, "    {}:", self.def_id)?;
+        let mut block_strs = Vec::new();
         for (i, b) in self.blocks.iter().enumerate() {
-            write!(f, "    bb{}:\n{}", i, b)?;
+            block_strs.push(format!("    bb{}:\n{}", i, b));
         }
+        println!("{:?}", block_strs);
+        writeln!(f, "{}", block_strs.join("\n"))?;
         writeln!(f, "[End TIR for {}]", self.item_path_str)?;
         Ok(())
     }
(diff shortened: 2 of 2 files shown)


ltratt:
I think there are two questions we can ask that will help us here: why does LLVM bitcode have a type system? why do RPython traces have a type system? The answer to both questions will probably be enlightening for us too.

vext01:
> Why does LLVM bitcode have a type system?

The answer is "it depends who you ask".

The least convincing is the one from the [LLVM docs](http://llvm.org/docs/LangRef.html#type-system):

> The LLVM type system is one of the most important features of the intermediate
representation. Being typed enables a number of optimizations to be performed
on the intermediate representation directly, without having to do extra
analyses on the side before the transformation. A strong type system makes it
easier to read the generated code and enables novel analyses and
transformations that are not feasible to perform on normal three address code
representations.

But we find better answers if we dig in early 2000s papers.

In [this paper](http://llvm.org/pubs/2003-10-01-LLVA.pdf), the authors first argree with the above quote:
> standard language-independent optimizations use only some subset of these
types (if any),including optimizations that require array dependence analysis,
pointer analysis (even field-sensitive algorithms [16]),and call graph
construction.

But then goes on to say:
> The most important purpose of the type system, however,is to enable typed
memory access. LLVA achieves this via type-safe  pointer  arithmetic  using
the getelementptr instruction. This enables pointer arithmetic to be
expressed directly in LLVA without exposing implementation details, such as
pointer size or endianness. To do this, offsets are specified in terms of
abstract type properties (field number for a structure and element index for an
array).

Which suggests that types are more for portability. This is then confirmed a second time with:
> "As noted in Section 2, a key design goal for a V-ISA isto maintain object code
portability across a family of processor implementations"

In [Lattner's thesis](http://llvm.org/pubs/2002-12-LattnerMSThesis.pdf), he says:
> LLVM is a strictly typed representation,where every SSA value and memory
location has an associated type, and all operations obey strict type
rules. This type information enables a broad class of high-level transformations
on low-level code. In addition, type mismatches can be used to detect errors in
optimizations by the LLVM consistency checker.

and

> "For example, classes in C++ with inheritance and virtual methods can be
represented using structures for the data values and a typed function table
with indirect function calls for inheritance. This permits many high-level
language-independent optimizations (e.g., virtual function resolution) to be
performed on the LLVM code"

and

> "Type information enables high-level information to be easily extracted from
the low-level code, enabling novel transformations at link time. To do this,
however, it must be possible to allow type-safe access to fields of data in
memory. For this reason, a critical instruction in LLVM (for maintaining
type-safety) is the getelementptr instruction."

Other semi-related items from these sources which interested me:

> "Registers can only hold scalar values, viz., boolean, integer, floating point, and pointer"

i.e. large aggregates can never live inside a register. Makes sense really, as you typically can't fit a struct into a machine register.

> A keypoint is that LLVM instructions are polymorphic: a single instruction
(like add) can operate on several different types of operands. This greatly
reduces the number of distinct opcodes.

So perhaps we want type-parameterised operations in our TIR?

Any comments? In the meantime I'll look into why RPython traces are typed.

vext01:
By the way, I think LLVA was the name of LLVM IR at the time the papers were written.

vext01:
As for PyPy. I've had a chat with Armin as to why there are different register types in traces (the `i`, `f`, `p`) stuff.

 * It's useful to be able to distinguish floats since they will need to be stored in non-general purpose registers.
 * Floats are always 64-bit, whereas integers depend on the arch, and that can matter.
 * It's useful to know what's a pointer, for GC root purposes.

It appears that RPython does have a notion of either pointer types, or typed operations, e.g.:
```
i2 = getfield_gc_i(p0, descr=<FieldS pypy.interpreter.pyframe.PyFrame.inst_last_instr 40>)
```

Here we are getting the field `inst_last_isntr` from offset `40` from the pointer `p0` which is a `PyFrame`.

I suppose this is the only way to do it with a language like Python, where instances of the same class could have their fields laid out differently (hence maps etc.)?

jacob-hughes:
I realise from this PR that my underlying assumptions of how the hardware tracer is intended to work aren't that clear. I don't think I can give you any good answers to this PR unless they're clarified. So please correct or confirm below:

**The primary goal**
---
It should be possible to find the corresponding MIR (stored somewhere in the binary) from a TIR trace (and therefore emit efficient assembly for hot loops from Rust's LLVM codegen pipeline).

***or***

It should be possible to _construct_ MIR from the TIR trace, from which we can then generate efficient LLVM IR to swap in.

***or***

Something else entirely and I've completely misunderstood ;)

---

There is also a secondary goal, which is that there might be interesting properties in MIR (and its type system) which allow us to make clever, context specific optimisations, and we'd like to preserve them for this reason.

vext01:
Hi Jake.

I believe that we are implementing the first of the three options you've outlined. We only really care about making executable code, but within the ABI constraints of Rust.

ltratt:
> It should be possible to find the corresponding MIR (stored somewhere in the binary) from a TIR trace (and therefore emit efficient assembly for hot loops from Rust's LLVM codegen pipeline).

Not quite. We need to find the corresponding "intermediate TIR" (i.e. what we save out at compile-time) from a PT stream, then we convert that into "full"/"real" TIR. [I'm not sure what good names are here. But there's definitely a difference between the TIR stored in the executable and the TIR that we convert to machine code. The latter is SSA etc. etc.]

ltratt:
> I suppose this is the only way to do it with a language like Python, where instances of the same class could have their fields laid out differently (hence maps etc.)?

No, there's nothing Python-specific here: by compile time things like ` pypy.interpreter.pyframe.PyFrame` are basically C structs.

Taking everything on board, I think I stand by my idea from a day or two back that what we're looking for is a type system that's basically Java pre-generics.

vext01:
> everything on board, I think I stand by my idea from a day or two back that what we're looking for is a type system that's basically Java pre-generics.

OK, so my next questions are:

 * Are user-types "first class" in TIR. In other words, if a user defines a struct `MyAmazingStruct`, does this get a new type identifier, or are all user structs one type?

 *  Where do we store the types? I suppose each variable will need a type identifier of some kind, and if we are to allow for user types, then we have an unbounded number of types which me must represent.

At the moment, a TIR variable is simply a u32. We could make them a u64 (or a pair of u32s, whatever) and use the top 32-bits as a type identifier. This would allow us to represent 2^32 possible types.

ltratt:
> Are user-types "first class" in TIR?

Yes, otherwise you don't have a type system: you just have `void *`! Like I said: think Java and you'll be in the right sort of design space. [Not entirely coicidentally, RPython's type system is basically Java with simple generics.] Note this does mean we need compound types (for structs and so on).

>Where do we store the types? I suppose each variable will need a type identifier of some kind, and if we are to allow for user types, then we have an unbounded number of types which me must represent.

Where we store things is, IMHO, a complete red herring at this stage. `u32`, `u64`, whatever... it doesn't matter very much yet.

vext01:
> Are user-types "first class" in TIR?
> Yes

OK, works for me.

> Where we store things is, IMHO, a complete red herring at this stage.

I don't think it's a *total* red herring. I'd like to try and get it right so that we don't have to unpick/refactor things later.

Anyway, I'll start with a numeric type tag. A u64 for each variable (32-bit ID, 32-bit type-ID). This will be efficient on 64-bit systems at least.

vext01:
(While I'm making changes to my code, I wonder if we are re-inventing LLVM here. Why is it we are not tracing at the LLVM-level again?)

vext01:
I've just pushed a change to TIR in light of what we've learned.

In short:
 * Each variable now carries a type index.
 * All operations are polymorphic, but strongly typed.
 * Get fields from aggregates by index, not byte offset.
 * Remove `AddressOf`. This was inspired by LLVM IR, which too has no address-of operator.

If we were to go ahead with this design, we'd need a second type of pack which is a type description. We'd have array of these, and each `TyIndex` is a index into the array.

Our `GetField` operator is effectively a limited version of LLVM's `getelementptr`. Ours doesn't deal with array indexing. It seems this is a great source of confusion in LLVM IR. We'd probably need another operation for array indexing.

Thoughts at this point?

The question in my previous comment still stands.

vext01:
I'm hoping for some input on this before we continue. Any comments?

ltratt:
If you think we might be able to use LLVM IR for this, why not give it a go? Maybe we can! None of us knows the right answer here, so don't assume that I have a secret that I'm not sharing :)

vext01:
The only downside that I we can think of (well, it was Jake's idea) is that we'd probably lose trait information(?).

Admittedly I've not looked at how traits lower to LLVM, so maybe that's something I should look at before making a decision.

vext01:
I've been weighing up the pros/cons of an approach that uses LLVM IR as the tracing language:

Pros:
 * Already in SSA.
 * More faithful (fewer steps of abstraction to) the executable code (big plus).
 * We can use libllvm to parse LLVM IR in the JIT runtime.

Cons:
 * Tracer runtime would have to use raw pointers from libllvm (unless we wrote our own LLVM IR parser). Using a custom IR we simply use Serde.
 * We are become more reliant on specific LLVM versions.
 * I'd have to re-write the software tracing MIR pass to be an LLVM pass (superficial qualm, but still).
 * Loss of trait information (see below).

And there were a couple of other items that I was curious about:

## Loss of Trait Information

I looked at the LLVM IR for code using trait objects. Traits have no real manifestation beyond a (untyped) pointer and a vtable.

This compiler explorer link explains nicely:
https://godbolt.org/z/SsNZlh

Here we see the pointer and its vtable being loaded:
```
  %2 = extractvalue { {}*, [3 x i64]* } %1, 0, !dbg !38
  %3 = extractvalue { {}*, [3 x i64]* } %1, 1, !dbg !38
```

Notice also how the fat pointer is actually three 64-bit integers in size (not two as I might expect). I'm not sure what the last pointer is used for... Even if you add `-O` fat pointers are still three pointers long.

I've asked in discord, because I'm really curious to know what the last pointer is used for.

Anyway, Jake is right. Traits will not map to distinct types in LLVM.

## LLVM already has some kind of JIT. Why does no one use it?

TLDR; It's more like a C compiler than can be used at runtime, and it's quite slow.

https://stackoverflow.com/questions/6833068/why-is-llvm-considered-unsuitable-for-implementing-a-jit

Currently I'm still leaning towards sticking with our own IR, but does anyone have any comments?

ltratt:
Traits are very different from normal types, because there's no struct `S` matching a trait `T`. So I wouldn't be surprised if you could ignore them, at least for quite a while.

vext01:
Hrm. I think I'm going to cautiously continue with our own IR...

vext01:
Assuming everyone thinks it's OK to continue with TIR, here's the next set of changes.

I  have a branch where the compiler builds against this, although I've only implemented (some) assignments at the moment.

In short, the changes since last time:
 * Locals are now a 64-bit struct consisting of an ID and a type tag.
 * I decided not to type operations themselves, since the operands are now all typed.
 * I restricted the types of the arguments of some operations a bit.
 * Implemented `Display` for everything, in a more human readable form (I may change the formatting of constants and terminators. See below.).

Here's an (arbitrary) example of what textual TIR will look like:
```
    DefId(1, 2):
    bb0:
        $0: t0 = $1: t0
        $2: t0 = get_field($3: t0, 4)
        $4: t0 = U8(10)
        nop
        Abort
    bb1:
        $5: t0 = load($6: t0)
        store($5: t0, $4: t0)
        $7: t0 = int_add($8: t0, $9: t0)
        $7: t0 = int_sub($9: t0, $10: t0)
        $11: t0 = alloca(0)
        Goto { target_bb: 50 }
```

 * Variables are prefixed with `$`.
 * Variables are typed, the type index is after the colon. In the future we can print something more meaningful than the index. Don't read into all of the types being `t0`. Assigning the correct type ID will come later.

As mentioned before, I'd like to change how terminators and constants format. E.g.:
 * `U8(10)` would be better as `10u8`
 * `Goto {target_bb: 50}` would be better as `goto(bb50)`.

Possibly do those as the next PR(s).

vext01:
The above referenced PR are the corresponding compiler changes.

ltratt:
I'm fine with this.

jacob-hughes:
Looks good. So `DefId(1, 2):` here means that this is a function declaration representation in TIR. Which means that you could do a query for that same `DefId` to retrieve the MIR representation?

vext01:
That's right.

vext01:
Everyone OK for some squashing?

ltratt:
Fine with me.

jacob-hughes:
Yes

vext01:
Splat.

PR #10, opened 2019-05-30
Implement enough stuctures to lower a comparison with a constant int.

vext01:
TLDR: Implement enough stuff to lower an instruction like:
```
$3: t0 = ge($2: t0, I32(2000))
```
In doing so, I've noticed a few oddities with our intial design which
I've also addressed:

 - It seems odd to have an Operand as a Rvalue. It makes more sense to
   lift Operand's constituent bits directly into the Rvalue enum. In
   doing so we can reduce the number of "this can't happen"s later. E.g.
   we know something can only accept a constant, but we'd have to use an
   Operand and consider the Local case too.

   Operand still remains for operations which can accept both a constant
   *or* a local.

 - Instead of having binary integer operations as top-level rvalues, group them
   under a BinaryOp enum. This will allow us to more easily treat binary
   ops uniformly based on structure.

   Nullary and Unary ops will follow at some point.

 - Similarly for constant integers. Group them into a ConstantInt enum.

 - TIR operations are polymorphic, so the Add operation should Display
   as "add", not "int_add".

Additionally, I've made the Display for an unimplemented statement
lowering include the stringified MIR of the statement, so I can easily
see what lowering need to be implemented next.

Companion compiler PR will come soon.

### DIFF
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -165,7 +165,8 @@ pub enum Statement {
     /// Store into the memory.
     Store(Local, Operand),
     /// Any unimplemented lowering maps to this variant.
-    Unimplemented,
+    /// The string inside is the stringified MIR statement.
+    Unimplemented(String),
 }
 
 impl Display for Statement {
@@ -174,7 +175,7 @@ impl Display for Statement {
             Statement::Nop => write!(f, "nop"),
             Statement::Assign(l, r) => write!(f, "{} = {}", l, r),
             Statement::Store(ptr, val) => write!(f, "store({}, {})", ptr, val),
-            _ => write!(f, "unimplemented"),
+            Statement::Unimplemented(mir_stmt) => write!(f, "unimplemented_stmt: {}", mir_stmt),
         }
     }
 }
@@ -183,14 +184,15 @@ impl Display for Statement {
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
 pub enum Rvalue {
     /// Another local variable.
-    Operand(Operand),
+    Local(Local),
+    /// A constant value.
+    Constant(Constant),
     /// Get a pointer to a field.
     GetField(Local, FieldIndex),
     /// Load a value of specified type from a pointer.
     Load(Local),
-    /// Binary Ops.
-    Add(Operand, Operand),
-    Sub(Operand, Operand),
+    /// Nullary, Unary and Binary Ops.
+    BinaryOp(BinOp, Operand, Operand),
     /// Allocate space for the specified type on the stack and return a pointer to it.
     Alloca(TyIndex),
 }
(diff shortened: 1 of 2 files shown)


ykpack/src/types.rs
```diff
@@ -225,21 +227,85 @@ impl Display for Operand {
...
 }
 
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
+pub enum ConstantInt {
+    UnsignedInt(UnsignedInt),
+    SignedInt(SignedInt),
+}
+
+impl ConstantInt {
+    // When constructing constant integers, truncation is deliberate.
```
ltratt: What does this mean?

ltratt: You might be able to do something like:

```rust
impl From<uwhatever> for ConstantInt {
    fn from(x: uwhatever) -> Self { ... }
}
```

and then avoid having to call differently named functions.

vext01: > What does this mean?

It's justification for the excessive use of `as`.

> You might be able to do something like:

Can I special-case u128? It breaks the mould sadly.

vext01: > Can I special-case u128? It breaks the mould sadly.

I don't think it's going to work. Each different type lowers to a different ykpack variant, so each type needs different functionality anyway.

ltratt: >Each different type lowers to a different ykpack variant, so each type needs different functionality anyway.

OK.

Can we make the comment a bit less gnomic?

vext01: Let me try some macros.

vext01:
The last commit is the least offensive solution I could come up with.

Any good?

ltratt:
Do we want to test all the integer conversions instead of just 1? And do we want to test the `128` case in particular?

vext01:
> Do we want to test all the integer conversions instead of just 1?

I think *all* of them might be excessive, but I think you are right that testing a 128 case might be good.

ltratt:
OK.

vext01:
Last push tests a few more cases, including -ve values where incorrect truncation could yield a +ve value.

ykpack/src/types.rs
```diff
@@ -524,9 +524,32 @@ mod tests {
...
+
+    #[test]
+    fn const_u64_from_bits() {
+        let v = std::u64::MAX;
+        let cst = ConstantInt::u64_from_bits(v as u128);
+        assert_eq!(cst, ConstantInt::UnsignedInt(UnsignedInt::U64(v)));
+    }
+
+    #[test]
+    fn const_i32_from_bits() {
```
ltratt: Let's order these low to high, so swap the 32 and 64 bit tests around?

vext01:
> Let's order these low to high, so swap the 32 and 64 bit tests around?

Fixed

ltratt:
Please squash.

vext01:
splat.

vext01:
Sorry about that. Shall I squash?

ltratt:
Please squash.

vext01:
splat.

PR #13, opened 2019-06-14
Tiri

vext01:
> A preliminary design for a TIR trace interpreter.
>
> This is designed only to show that the traces we are collecting are somewhat sane, and as such I've put zero effort into trying to make it fast.
>
> In the long run we will have a TIR trace *compiler*.

What do you think of the design? A few notes:

 * The stack works out quite nicely. Since all temporary computations get stored into TIR locals, and we know how many of those there are ahead of time, there's no need to have a growable/shrinkable frame, nor is there a need to figure out how big each frame needs to be.

 * Each frame has a program counter, which can double up as return address storage.

 * The interpreter struct itself is immutable, and all mutable state is in the scope of the `run()` function. This (I hope) should make it easier to dodge the borrow checker.

Thanks.

### DIFF
--- a/tiri/src/main.rs
+++ b/tiri/src/main.rs
@@ -1,0 +1,110 @@
+// Copyright 2019 King's College London.
+// Created by the Software Development Team <http://soft-dev.org/>.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Tiri -- TIR interpreter.
+//!
+//! No effort has been made to make this fast.
+
+use elf;
+use fallible_iterator::FallibleIterator;
+use std::collections::HashMap;
+use std::convert::TryFrom;
+use std::io::Cursor;
+use std::path::{Path, PathBuf};
+use ykpack::{BasicBlockIndex, Body, Decoder, DefId, Pack, Statement, StatementIndex, Terminator};
+
+/// The interpreter maintains a stack of these to keep track of calls.
+struct Frame<'p> {
+    /// A reference to the TIR body corresponding with this `Frame`.
+    body: &'p Body,
+    /// The current execution position in the above TIR body.
+    pc: (BasicBlockIndex, StatementIndex),
+}
+
+impl<'p> Frame<'p> {
+    /// Create a new frame and position its program counter at the beginning of the first block.
+    fn new(body: &'p Body) -> Self {
+        Self { body, pc: (0, 0) }
+    }
+}
+
+/// The interpreter itself.
+/// The struct itself holds only immutable program information.
+struct Interp {
+    /// Maps a `DefId` to the corresponding TIR body.
+    tir_map: HashMap<DefId, Body>,
+    /// The `DefId` of the `main()` function, which serves as the entry point.
+    main_defid: DefId,
+}
+
+impl Interp {
+    /// Create a new interpreter, using the TIR found in the `.yk_tir` section of the binary `bin`.
+    fn new(bin: &Path) -> Self {
+        let ef = elf::File::open_path(&PathBuf::from(bin)).unwrap();
+        let sec = ef.get_section(".yk_tir").unwrap();
+        let mut curs = Cursor::new(&sec.data);
+        let mut dec = Decoder::from(&mut curs);
+
+        let mut tir_map = HashMap::new();
+        let mut main_defid = None;
+        while let Some(pack) = dec.next().unwrap() {
+            let Pack::Body(body) = pack;
+
+            if body.def_path_str == "main" {
+                main_defid = Some(body.def_id.clone());
+            }
+            tir_map.insert(body.def_id.clone(), body);
+        }
+
+        Self {
+            tir_map: tir_map,
+            main_defid: main_defid.expect("Couldn't find main()"),
+        }
+    }
+
+    /// Start interpreting TIR from the `main()` function.
+    fn run(&self) {
+        let mut stack = Vec::new();
+        stack.push(Frame::new(&self.tir_map[&self.main_defid]));
+
+        // The main interpreter loop.
+        loop {
+            let cur_frame = stack.last().unwrap();
+            let cur_block = &cur_frame.body.blocks[usize::try_from(cur_frame.pc.0).unwrap()];
+            let block_len = cur_block.stmts.len();
+            let pc_stmt_usize = usize::try_from(cur_frame.pc.1).unwrap();
+
+            if pc_stmt_usize < block_len {
+                let stmt = &cur_block.stmts[pc_stmt_usize];
+                self.interp_stmt(stmt);
+            } else if pc_stmt_usize == block_len {
+                // We take statement index one past the end to mean the block terminator.
+                let term = &cur_block.term;
+                self.interp_term(term);
+            } else {
+                unreachable!();
+            }
+        }
+    }
+
+    /// Interpret the specified statement.
+    fn interp_stmt(&self, _stmt: &Statement) {
+        unimplemented!();
+    }
+
+    /// Interpret the specified terminator.
+    fn interp_term(&self, _term: &Terminator) {
+        unimplemented!();
+    }
+}
+
+fn main() {
+    let bin = std::env::args().skip(1).next().unwrap();
+    Interp::new(&PathBuf::from(bin)).run();
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+target
+Cargo.lock
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,5 +2,6 @@
 members=[
     "ykpack",
     "ykrt",
-    "yktrace"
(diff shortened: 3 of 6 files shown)


tiri/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+use elf;
+use fallible_iterator::FallibleIterator;
+use std::collections::HashMap;
+use std::convert::TryFrom;
+use std::io::Cursor;
+use std::path::{Path, PathBuf};
+use ykpack::{BasicBlockIndex, Body, Decoder, DefId, Pack, Statement, StatementIndex, Terminator};
+
+/// The interpreter maintains a stack of these to keep track of calls.
+struct Frame<'p> {
```
ltratt: Why `'p` and not the more normal `'a` (or something else that makes sense in context)?

tiri/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+    /// Maps a `DefId` to the corresponding TIR body.
+    tir_map: HashMap<DefId, Body>,
+    /// The `DefId` of the `main()` function, which serves as the entry point.
+    main_defid: DefId,
+}
+
+impl Interp {
+    /// Create a new interpreter, using the TIR found in the `.yk_tir` section of the binary `bin`.
+    fn new(bin: &Path) -> Self {
+        let ef = elf::File::open_path(&PathBuf::from(bin)).unwrap();
```
ltratt: Is `ef` needed as in intermediary? `let sec = elf::File::open_path(...).unwrap().get_section(".yk_tir").unwrap()` might be more succinct.

tiri/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        Self {
+            tir_map: tir_map,
+            main_defid: main_defid.expect("Couldn't find main()"),
+        }
+    }
+
+    /// Start interpreting TIR from the `main()` function.
+    fn run(&self) {
+        // Set up the stack to start executing at the entry point.
```
ltratt: This comment seems a bit pointless.

tiri/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+        // The main interpreter loop.
+        loop {
+            let cur_frame = stack.last().unwrap();
+            let cur_block = &cur_frame.body.blocks[usize::try_from(cur_frame.pc.0).unwrap()];
+            let block_len = cur_block.stmts.len();
+            let pc_stmt_usize = usize::try_from(cur_frame.pc.1).unwrap();
+
+            if pc_stmt_usize < block_len {
+                let stmt = &cur_block.stmts[pc_stmt_usize];
+                debug!(
```
ltratt: I'm really, really reluctant to merge any code which spews to stdout, even if it's only in debug mode. In my experience, this code sticks around for decades, even when it no longer serves any purpose.

tiri/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+            } else {
+                unreachable!();
+            }
+
+            break; // FIXME -- Just to prevent infinite loop until statement actions are implemented.
+        }
+    }
+
+    /// Interpret the specified statement.
+    fn interp_stmt(&self, _stmt: &Statement) {}
```
ltratt: Should have `unimplemented!()` in the body?

tiri/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    /// Interpret the specified terminator.
+    fn interp_term(&self, _term: &Terminator) {}
```
ltratt: Ditto.

tiri/src/main.rs
vext01: `p` for pack, since that's where the stuff inside is borrowed from. But I don't feel strongly.

tiri/src/main.rs
vext01: Sadly, if we do:
```
let sec = elf::File::open_path(&PathBuf::from(bin)).unwrap().get_section(".yk_tir").unwrap();
```

We get:
```
error[E0716]: temporary value dropped while borrowed
  --> tiri/src/main.rs:52:19
   |
52 |         let sec = elf::File::open_path(&PathBuf::from(bin)).unwrap().get_section(".yk_tir").unwrap();
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                - temporary value is freed at the end of this statement
   |                   |
   |                   creates a temporary which is freed while still in use
53 |         let mut curs = Cursor::new(&sec.data);
   |                                    --------- borrow later used here
```

tiri/src/main.rs
vext01: Yeah, perhaps. Let's kill it.

tiri/src/main.rs
ltratt: Ah, OK.

tiri/src/main.rs
ltratt: OK.

tiri/src/main.rs
vext01: I have to disagree with you on this one. There are several reasons I find `debug!` useful, but to cut to the chase and address your main objection:

> this code sticks around for decades, even when it no longer serves any purpose.

These calls will always serve a purpose: they will print a trace of TIR statements leading up to a crash I'm debugging. That's the kind of info that you can't get from a back-trace. When would this not be useful? Especially given that (as you have already noted) they only get compiled in under debug mode, but even then you only see them when a environment variable is set.

I agree that excessive `debug!`s is silly, but strategically placed debug statements save me time overall and there is a precedent to use them in Rust code. The ones in the rust compiler have been especially useful: in the compiler inserting a debug print triggers a lengthy compilation cycle.

Unless you feel very strongly, I'd like to ask politely if we I leave these in :)

tiri/src/main.rs
vext01: We could, but I wouldn't worry too much, as these are the next items to be populated.

tiri/src/main.rs
vext01: I was kind of hoping the non-linear lifetimes work would render this kind of error a thing of the past.

Perhaps that work never hit Rust?

ltratt: NLL is largely addressing a different issue, whereby the borrow checker rejects programs that appear to borrow a variable mutably and immutably. Here you've got a different problem, which is a temporary being freed while still being borrowed. I agree that this often feels quite annoying but I'm not sure if there's a sensible resolution or not.

tiri/src/main.rs
ltratt: I do feel *very* strongly about this (sorry). This is how chatty programs start life, and once the precedent is set, it's difficult to stop it happening again and again. If you need them for debugging, that's fine, but that's probably something for your local benefit that isn't useful for anyone else. In my experience, if I think I need to leave debugging statements in, it invariably means that I have an incomplete PR.

> The ones in the rust compiler have been especially useful: in the compiler inserting a debug print triggers a lengthy compilation cycle.

I assume they have some benefit other than making compilation slower? ;)

tiri/src/main.rs
ltratt: I'm a big fan of `unimplemented`, because it's also a clear guide to other people (e.g. reviewers) that things are as yet unfilled.

tiri/src/main.rs
vext01: Seems to be some confusion, but I think I know what's going on.

> This is how chatty programs start life, and once the precedent is set, it's difficult to stop it happening again and again. 

I think you have mixed-up `debug!` with `dbg!`? The former *doesn't* lead to chatty programs because you have to opt-in to seeing the messages by setting an environment variable (`RUST_LOG`). You can think of it a bit like `PYPY_LOG`.

You certainly wouldn't want `dbg!` all over the place!

> I assume they have some benefit other than making compilation slower?

No, `debug!` and friends aim to *save* compilation time. Any change to the source tree will trigger recompilation, so if coarse debugging statements are ever-present, often all you need to do to diagnose an issue is set the `RUST_LOG` appropriately.

For example if you wanted info-level information about what the `rustc_driver` crate is doing, you can set `RUST_LOG=rustc_driver=info`. Note that you can filter by crate.

If you still aren't convinced, my last stand (military term that) would be to point out that these macros are really standard. Grepping the compiler:

```
$ grep -R 'debug!' src | wc -l
    2406
```

ltratt: Custer/Barrett's last stand? ;) Here's the problem I have with this: of those 2406, how many are still relevant? My guess is very few. So turning on logging almost certainly overwhelms us with stuff that's irrelevant. That's the problem with putting this stuff in the code base: it never gets GCd.

vext01: Well the same is true of documentation and comments in my opinion. Just as you have to make sure your comments and docs are up to date, you have to make sure you `debug!`s are too.

I also suspect that as code churns you are forced to review your `debug!`s, since they will surely be the cause of compilation errors during development.

I can see you can't be swayed, so let's move on. Perhaps this will be something we will never agree upon.

vext01:
That should address your comments.

tiri/src/main.rs
ltratt: Yeah, sorry that this is one where it's impossible to find a compromise :/ Thanks for being a good sport about it!

ltratt:
Please squash.

vext01:
Splat. Kept the addition of the gitignore separate.

PR #14, opened 2019-06-18
Mir trace to tir frag boilerplate

vext01:
I'm raising this PR quite early so that we can discuss the design.

The way I envisage the TIR interpreter working is as follows:

 * Language interpreter written in our framework eventually decides to trace.
 * Tracing completes, yielding a MIR trace: a trace of MIR block locations.
 * We take the TIR from the ELF file and specialise it the the trace, yielding a *TIR Fragment*.
 * Next time we encounter the hot path, we interpret the TIR fragment.

A couple of items for discussion:

## The name "Tir Fragment"

Why not "TIR Trace"? As discussed the other day on MM, the TIR equivalent of a MIR trace isn't necessarily straight line code, so calling it a trace would be misleading.

I also considered "Tirlet".

## Trait fun.

The following trait irritates me:

```
/// Represents a thread which is currently tracing.                                 
pub trait ThreadTracer {                                                            
    /// Stops tracing on the current thread, returning the MIR trace on success. `None` is returned    
    /// on error or if the trace was invalidated. **The consumers should not call this**.    
    fn stop_tracing_impl(self: Box<Self>) -> Option<Box<dyn MirTrace>>;             
    /// Stops tracing on the current thread, returning a specialised TIR fragment on success.    
    /// Returns `None` if the trace was invalidated. **Do not override this**.      
    fn stop_tracing(self: Box<Self>) -> Option<TirFrag> {                           
        self.stop_tracing_impl().map(|mir_trace| TirSpecialiser::new(&*mir_trace).specialise())    
    }                                                                               
}
```

Notice the "Consumers should not call this" and "Do not override this" comments? I couldn't get the design I wanted using the Rust type system:

 * `stop_tracing` is the same for all backends.
 * `stop_tracing_impl` is specific to a backend, but shouldn't be part of the public API.

To get the desired API, what we could do is wrap `ThreadTracer` in another struct. Thoughts?

### DIFF
--- a/yktrace/Cargo.toml
+++ b/yktrace/Cargo.toml
@@ -6,3 +6,7 @@ edition = "2018"
 
 [dependencies]
 libc = "0.2"
+elf = "0.0"
+fallible-iterator = "0.2"
+ykpack = { path = "../ykpack" }
+lazy_static = "1.3"
--- a/yktrace/src/lib.rs
+++ b/yktrace/src/lib.rs
@@ -72,15 +72,24 @@ pub enum TracingKind {
 }
 
 /// Represents a thread which is currently tracing.
-pub trait ThreadTracer {
-    /// Stops tracing on the current thread, returning the MIR trace on success. `None` is returned
-    /// on error or if the trace was invalidated. **The consumers should not call this**.
-    fn stop_tracing_impl(self: Box<Self>) -> Option<Box<dyn MirTrace>>;
+pub struct ThreadTracer {
+    /// The tracing implementation.
+    t_impl: Box<dyn ThreadTracerImpl>,
+}
+
+impl ThreadTracer {
     /// Stops tracing on the current thread, returning a specialised TIR fragment on success.
-    /// Returns `None` if the trace was invalidated. **Do not override this**.
-    fn stop_tracing(self: Box<Self>) -> Option<TirFrag> {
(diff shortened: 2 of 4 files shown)


vext01:
eac0a61 shows what I mean by "wrap ThreadTracer in another struct.".

Effectively, we rename the trait to `ThreadTracerImpl` and embed an instance of this in an outer `ThreadTracer`. Then we can move the method which shouldn't be overridden up into the outer struct.

ltratt:
Our terminology here doesn't quite add up. For example:

>Tracing completes, yielding a MIR trace: a trace of MIR block locations.
>We take the TIR from the ELF file and specialise it the the trace, yielding a TIR Fragment.

MIR block locations have become TIR in the ELF? And that becomes a TIR fragment? I don't know what we should call things, but we have: MIR that rustc produced; whatever format we decide to encode in the binary (which will probably be very MIR-like, but won't be MIR), let's call it XYZ; then we do PT and we get XYZ; we convert that into a trace. Terminology is always a tedious thing to get right, but I think we've shown before that, in this case, if we don't get it right, it's normally because we don't understand what we want! So it's probably worth getting this right(ish) now.

ltratt:
I think I'm missing some high-level design here: what's a TIR specialiser for example?

vext01:
This has been superseded.

PR #15, opened 2019-06-24
TIR -> SIR.

vext01:
In light of our face-to-face meeting, rename TIR to SIR (serialised IR). A compiler change will follow.

We also discussed a second IR which we could call TIR to express traces, but in hindsight, I don't think we need to give it such a grand name.

In our system, a trace is simply a list of block locations and we've decided to assume that blocks have a one-to-one mapping between MIR, SIR and (hopefully) binary code. So our "TIR" would probably simply be a Vec. Thoughts?

[That discussion point isn't strictly necessary for this PR, but I'd like to check before continuing]

### DIFF
--- a/ykpack/src/lib.rs
+++ b/ykpack/src/lib.rs
@@ -69,7 +69,7 @@ mod tests {
             BasicBlock::new(stmts1_b1, dummy_term.clone()),
             BasicBlock::new(stmts1_b2, dummy_term.clone()),
         ];
-        let tir1 = Pack::Body(Body::new(DefId::new(1, 2), String::from("item1"), blocks1));
+        let sir1 = Pack::Body(Body::new(DefId::new(1, 2), String::from("item1"), blocks1));
 
         let stmts2_b1 = vec![Statement::Nop; 7];
         let stmts2_b2 = vec![Statement::Nop; 200];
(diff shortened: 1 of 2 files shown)


vext01:
By the way, I've not updated any names in tiri yet, (which itself needs a rename). The interpreter will need the modified compiler and ykpack, so we have to do this in stages.

vext01:
This PR must be merged first.

vext01:
I was hoping to resolve the above discussion point first :P

ltratt:
OK, I don't really understand what the discussion point is TBH!

vext01:
The other day we agreed we'd have a SIR *and* a TIR, but it seems to be that be don't really need a TIR: A vector of block locations should suffice.

ltratt:
Ah, I'm with you.

I disagree, because the TIR will presumably have done things like SSA renaming in unrolled loops and so on. In other words: TIR will follow the *structure* of TIR, but its innards will be subtly different.

I'm not sure that's very relevant to this PR though?

vext01:
> I'm not sure that's very relevant to this PR though?

You are right, it's not really, but as I said:
> [That discussion point isn't strictly necessary for this PR, but I'd like to check before continuing]

>  because the TIR will presumably have done things like SSA renaming in unrolled loops and so on

Unrolling is a good counter-example. I'm with you now. In light of that, TIR will be *similar* to MIR, but without any blocks or terminators.

If you are happy, let's continue to bors this PR.

PR #17, opened 2019-06-26
Kill unwind edges/terminators, add stuff needed for guards.

vext01:
This PR should be merged before the companion PR.

Companion PR: https://github.com/softdevteam/ykrustc/pull/32

This change does three things:

## Remove all unwind/cleanup edges and terminators.

As mentioned in a face-to-face, we are going to use the abort panic strategy for our system. There's a handful of things in terminators which we therefore no longer care about:

 * Terminators soley for unwinding: `Abort` and `Resume`.
 * Edges out of terminators which only apply if there's the possibility of unwinding (`cleanup` and `unwind` edges).

## Add stuff that we will need for guards.

There were some bits missing from some terminators which we need to generate guards.

Example:

An `Assert` terminator has a condition, which we will need to guard. Add this field.

## Add missing info from Drop terminators.

The location-related fields were missing from `Drop` and `DropAndReplace`.

### DIFF
--- a/ykpack/src/lib.rs
+++ b/ykpack/src/lib.rs
@@ -61,7 +61,7 @@ mod tests {
 
     // Makes some sample stuff to round trip test.
     fn get_sample_packs() -> Vec<Pack> {
-        let dummy_term = Terminator::Abort;
+        let dummy_term = Terminator::Goto(10);
 
         let stmts1_b1 = vec![Statement::Nop; 16];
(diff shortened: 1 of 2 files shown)


vext01:
I've just realised I forgot to fmt this. Go ahead with the review and I'll fmt at the end. Mi culpa.

ltratt:
Please format and squash at the same time.

vext01:
Fmt+splat.

PR #19, opened 2019-07-03
Add TIR guards.

vext01:
For now we implement only enough guards to cover the behaviours of our SIR terminators.

I've had to make a few changes to our terminators in light of some subtle MIR semantics which I initially missed.

I don't like the name `OtherInteger` as a guard variant name. Can you think of a better one? It should capture that we are guarding that the value of a local is not a member of a list of integer values.

Companion PR: https://github.com/softdevteam/ykrustc/pull/33

### DIFF
--- a/yktrace/src/tir.rs
+++ b/yktrace/src/tir.rs
@@ -16,7 +16,7 @@ use elf;
 use fallible_iterator::FallibleIterator;
 use std::{collections::HashMap, convert::TryFrom, env, io::Cursor};
 #[cfg(debug_assertions)]
-use ykpack::{BasicBlockIndex, Terminator};
+use ykpack::{BasicBlockIndex, Local, SerU128, Terminator};
 use ykpack::{Body, Decoder, DefId, Pack, Statement};
 
 // The SIR Map lets us look up a SIR body from the SIR DefId.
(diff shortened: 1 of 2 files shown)


vext01:
I should explain, that this is a necessary step for trace interpretation in TIRI. I'm hoping to get this working outside of ykrt for now (limiting scope and all that ;) )

yktrace/src/tir.rs
```diff
@@ -87,7 +87,39 @@ impl TirTrace {
...
+                | Terminator::Call { .. }
+                | Terminator::Unimplemented(_) => None,
+                Terminator::Unreachable => panic!("Traced unreachable code"),
+                Terminator::SwitchInt {
+                    local,
+                    ref values,
+                    ref target_bbs,
+                    otherwise_bb
+                } => {
+                    // Peek at the next block in the trace to see which outgoing edge was taken and
```
ltratt: Do we know that there's always another block in the trace?

vext01: I'm working on the assumption that there is. I don't think we can stop tracing on a `SwitchInt`. Either way, it'll crash if there isn't one, so at least it's not a silent failure.

What would you prefer?

ltratt: If that's a valid assumption, let's document it. If we're unsure, I suggest an `assert` which will at least make clear to someone that we're not entirely sure if this code is correct or not.

vext01: I agree. Just testing the change now.

vext01:
Try that last push.

ltratt:
Please squash.

vext01:
splat

vext01:
```
---- tir::tests::nonempty_tir_trace stdout ----
thread 'tir::tests::nonempty_tir_trace' panicked at 'called `Result::unwrap()` on an `Err` value: Syntax("invalid length 3, expected struct variant Terminator::SwitchInt with 4 elements")', src/libcore/result.rs:999:5
```

The ykrustc tarball we use for testing used an older ykpack.

We have a circular dependency. We can't update ykrustc without updating yk...

Any ideas on what to do? The only think i can think of is to update ykrustc overriding the yk dependency to the feature branch to break the cycle.

ltratt:
I didn't realise that there's a circular dependency. Should we break it once and for all? Or is that impossible?

vext01:
Let me try to explain better. I want you to understand because I'm not sure what to do.

The problem is subtle. Ideally we want to test ykrustc and yk as one unit. That's not easy due to repo practicalities. Our workaround is to use the ykrustc tarball for the last succeeded CI build. In this case the last successdful CI build used an older ykpack with an incompatible SIR format. So boom.

ltratt:
Can't we just delete the last CI build and this PR will therefore build everything from scratch?

vext01:
Sadly not. Then it will crash because there is no tarball ;)

This snippet may help to explain:
https://github.com/softdevteam/yk/blob/cdc99012998c469fb665f334c051b87b6917d92a/.buildbot.sh#L5-L7

There are a couple of options I can think of:

 * We make the compiler temporarily depend on a feature branch of yk to break the cycle, then do the yk PR, and then finally adjust back the dependency in the compiler repo. This is the safest option, but involves 3 steps.

 * We manually make a tarball of the new compiler and copy it in place on the CI server (it's only the tests that make a cycle). This is easier in terms of PRs, but is racy, as another unrelated PR may overwrite our tarball at any time.

ltratt:
> Then it will crash because there is no tarball ;)

Ah. My original intention was that it should say "if there's no tarball, rebuild everything" which would, I believe, cure this problem entirely. One reason I suggested that is precisely for cases where we get some weirdness that can be solved by rebuilding everything from scratch. Shall we do that? [Assuming that it will actually fix things! Maybe even rebuilding from source isn't enough, at which point we really will have to think again!]

vext01:
> Ah. My original intention was that it should say "if there's no tarball, rebuild everything" which would, I believe, cure this problem entirely.

It would only work if you had some way of communicating to the yk build that it should use the as-yet unmerged ykrustc branch. Remember, we need a compiler that was built against the very yk branch that we are PRing at any given time.

ltratt:
> It would only work if you had some way of communicating to the yk build that it should use the as-yet unmerged ykrustc branch.

Good point! Well, the easiest way to solve this in this case is to manually replace the tarball with one that's "good" and I suggest we do that. In the long run, we can perhaps think of something more clever.

vext01:
Ack. So that's option 2 in this comment:
https://github.com/softdevteam/yk/pull/19#issuecomment-508480082

I might as well do the stage 2 test manually while i'm here then ;)

ltratt:
OK.

vext01:
bors try

vext01:
You can r+ this one now.

PR #20, opened 2019-07-10
Boilerplate refactorings in preparation for trace interpretation.

vext01:
Before tiri was a general SIR interpreter, which is not correct. We want
a TIR trace interpreter. As a result:

 * There is no stack.
 * We are not interested in SIR block information at this stage.
 * Interpreter state is an index into the trace and a variable store.

This will likely not be the final design, but it's a good starting point
to iterate upon.

[Note that `main.rs` moved to `lib.rs` as we won't need a `main()`. The changes were large enough for git to not recognise this as a moved file]


yktrace/src/tir.rs
```diff
@@ -126,6 +127,10 @@ impl TirTrace {
         Ok(Self { ops })
     }
 
+    pub fn op(&self, idx: usize) -> &TirOp {
+        &self.ops.get(idx).expect("bogus trace index")
```
ltratt: I think we should make this `get_unchecked` (and document that) on the basis that you should never pass an index in that you haven't checked elsewhere is safe. [We can add a `debug_assert` that checks the index is OK, but then we only pay the cost in debug mode.] Thoughts?

vext01: Yeah. Works for me.

As long as it crashes in debug mode with some clue of where to look, I'm happy :)

vext01:
Try this.

ltratt:
Please squash.

vext01:
Splat.

PR #21, opened 2019-07-12
An error type for trace invalidation.

vext01:
The meat of this PR is in the first commit (see the commit message for more) and then there's a boring style change.

### DIFF
--- a/tiri/src/lib.rs
+++ b/tiri/src/lib.rs
@@ -11,6 +11,9 @@
 //!
 //! No effort has been made to make this fast.
 
+#![feature(test)]
+extern crate test;
+
 use yktrace::tir::{Guard, Statement, TirOp, TirTrace};
 
 /// Storage space for one local variable.
@@ -68,5 +71,35 @@ impl<'t> Interp<'t> {
     /// Interpret the specified terminator.
     fn interp_guard(&self, _guard: &Guard) {
         unimplemented!();
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::Interp;
+    use test::black_box;
+    use yktrace::{start_tracing, TracingKind};
+
+    // Some work to trace.
+    #[inline(never)]
+    fn work(x: usize, y: usize) -> usize {
+        let mut res = 0;
+        while res < y {
+            res += x;
+        }
+        res
+    }
+
+    #[test]
+    #[ignore] // FIXME -- This will fail until we investigate why SIR is missing.
+    fn interp_simple_trace() {
+        let tracer = start_tracing(Some(TracingKind::SoftwareTracing));
+        let res = work(black_box(3), black_box(13));
+        let tir_trace = tracer.stop_tracing().unwrap();
+        assert_eq!(res, 15);
+        assert!(tir_trace.len() > 0);
+
+        let interp = Interp::new(&tir_trace);
+        interp.run();
     }
 }
--- a/yktrace/src/errors.rs
+++ b/yktrace/src/errors.rs
@@ -1,0 +1,29 @@
+// Copyright 2019 King's College London.
+// Created by the Software Development Team <http://soft-dev.org/>.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use std::fmt::{self, Display, Formatter};
+use ykpack::DefId;
+
+#[derive(Debug)]
+/// Reasons that a trace can be invalidated.
+pub enum InvalidTraceError {
+    /// There is no SIR for the DefId of a location in the trace.
+    NoSir(DefId),
+    /// Something went wrong in the compiler's tracing code
+    InternalError
+}
+
+impl Display for InvalidTraceError {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            InvalidTraceError::NoSir(def_id) => write!(f, "No SIR for location: {}", def_id),
+            InvalidTraceError::InternalError => write!(f, "Internal tracing error")
+        }
+    }
+}
--- a/yktrace/src/lib.rs
+++ b/yktrace/src/lib.rs
@@ -13,17 +13,22 @@
 extern crate test;
 
 use core::yk_swt::SirLoc;
-use std::iter::{IntoIterator, Iterator};
+use std::{
+    fmt::Debug,
+    iter::{IntoIterator, Iterator}
+};
 #[macro_use]
 extern crate lazy_static;
 
+mod errors;
 mod swt;
 pub mod tir;
 
+use errors::InvalidTraceError;
 use tir::TirTrace;
 
 /// Generic representation of a trace of SIR block locations.
-pub trait SirTrace {
+pub trait SirTrace: Debug {
     /// Returns the length of the trace, measured in SIR locations.
     fn len(&self) -> usize;
     /// Returns the SIR location at index `idx`.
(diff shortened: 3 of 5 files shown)


tiri/src/lib.rs
```diff
@@ -70,3 +73,33 @@ impl<'t> Interp<'t> {
...
+            res += x;
+        }
+        res
+    }
+
+    #[test]
+    #[ignore] // FIXME -- This will fail until we investigate why SIR is missing.
+    fn interp_simple_trace() {
+        let tracer = start_tracing(Some(TracingKind::SoftwareTracing));
+        let res = black_box(work(black_box(3), black_box(13)));
```
ltratt: Surely there's no point in doing `black_box(3)` -- an integer can't realistically be optimised!

yktrace/src/errors.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use std::fmt::{self, Display, Formatter};
+use ykpack::DefId;
+
+#[derive(Debug)]
+/// Reasons that a trace can be invalidated.
+pub enum InvalidTraceError {
+    /// There is no SIR for the DefId of a location in the trace.
+    NoSir(DefId),
+    /// Something went wrong in low-level compiler support.
```
ltratt: This sentence doesn't parse.

ltratt:
One thing that *might* be worth thinking about is whether you want to `Box` your error. Basically if you don't expect errors to be returned very often, and if the size of the `enum` is bigger than `usize` (which it may well be already), then you'll reduce the size of the function's return type by `Box`ing errors. It's not easy to know when this is/isn't a good idea, unfortunately...

tiri/src/lib.rs
vext01: It's to prevent constant propagation into the function body. Otherwise the body may be replaced by a value.

yktrace/src/errors.rs
vext01: I'm not sure what you mean. Seems OK here.

Does "Something went wrong in the compiler's tracing code" work better?

vext01:
A `DefId` is already bigger than a box, so it does make sense to use a box. We might aswell implement `Error` if we are going that route too. Then `?` will work.

tiri/src/lib.rs
ltratt: Doesn't the outer `black_box` prevent that happening?

yktrace/src/errors.rs
ltratt: Yep!

ltratt:
> We might aswell implement Error if we are going that route too.

Works for me.

tiri/src/lib.rs
vext01: Not necessarily. The addition could still be removed. Ultimately it's the arguments to the addition operator that we must prevent the compiler from knowing ahead of time.

The outer black box is actually probably not necessary.

vext01: (It's probably sufficient to only guard only one argument with a black box, too, but having two won't hurt)

vext01:
> then you'll reduce the size of the function's return type by Boxing errors.

Ah interesting.

Remember what Jake showed us LLVM does. If a return value is larger than a register, it'll put it on the caller's stack and pass a reference into the callee. I'd have to check if we can make that happen from Rust, but that'd be more efficient than heap allocating.

Thoughts before I investigate?

tiri/src/lib.rs
ltratt: We should understand what `black_box` does before we end up using it, otherwise we've no real idea if we're over or under using it. My guess -- but it is a guess! -- is that we've overused it here. But I might be wrong, and perhaps we need it in more places! We should understand it better IMHO.

ltratt:
> Thoughts before I investigate?

I think it's worth investigating! I'd forgotten about that LLVM discovery.

vext01:
Successfully reproduced!

image

Notice how the return value is a formal parameter?

ltratt:
I'd like to pretend I can make that sense of that... but it's beyond me!

vext01:
The highlighted line of the right-hand side shows the LLVM IR for the call on line 15 of the Rust program on the left.

You can see that the signature of `wiggle()` has been transformed from:
```
pub fn wiggle() -> Result<(), BiggidyError>;
```

To:
```
pub fn wiggle(err: &Result<(), BiggidyError>);
```

So the storage for the return value is allocated on the stack *before* the call, and a reference is passed into the callee. The callee then populates the pre-allocated space with the (conceptual) return value. Very cool.

ltratt:
Aha, I see!

vext01:
So now the question is, do we prefer to leave the return value unboxed (thus using stack allocation) or box it and use heap allocation?

My gut says the former is more efficient because stack allocation amounts to only subtracting from the stack pointer, whereas heap allocation requires locating a spot large enough on the heap. Using the heap also contributes to memory fragmentation and I'd also guess has worse locality.

Would you agree?

ltratt:
Why don't you try on your example? My guess is that boxing it still transforms things into a parameter, at which point the boxing is largely pointless (unless the struct involved is really big -- at some point the waste of stack space must be a factor, but I doubt we're at that point).

tiri/src/lib.rs
vext01: `black_box()` just inserts a chunk of inline assember code which is essentially the ID function:
https://doc.rust-lang.org/1.1.0/src/test/lib.rs.html#1054-1059

It stops optimisations crossing the black box.

My understanding, based on this, is that If you guard an argument to our work function with a BB, then operations that use the argument can't be replaced with a value.

If we guard the whole call, but not the args, I'd expect the call to be uninlinable, since the compiler loses knowledge about the return value, but not the computation in the leadup.

I can investigate further, or we can drop the outer black box.

ltratt: The docs say:

> A function that is opaque to the optimizer, to allow benchmarks to pretend to use outputs to assist in avoiding dead-code elimination.

which feels to me to be somewhat different than what you've said?

vext01: Yeah, it'd prevent DCE, but I'm very sure it'd block other optimisations. Let me try and build a proof.

vext01:
> My guess is that boxing it still transforms things into a parameter

The opposite, because the Box fits in a register.


image

vext01:
The opposite, because the Box fits in a register:

image

The call is:
```
%3 = tail call noalias align 8 dereferenceable_or_null(24) i64* @_ZN7example6wiggle17h8f5262ea6c257490E()
```

Notice the `i64*` return type? That's our Box, and there are no formal params.

ltratt:
That's probably because you've got `()` as the "success" part of the `Result`: put something in there which can't be squeezed in a pointer's spare bits, and I imagine it'll become bigger than a word.

tiri/src/lib.rs
vext01: Here's a case where we clearly see the whole call to the function `myadd` being replaced with a constant folded value `999`:

image

And here's the same with the arguments black boxed:
image

Notice here that the compiler is unable to jump to the conclusion `999`, as it doesn't know what the inline assembler code does. It is forced to perform the addition.

[As an aside, notice that the function is still inlined regardless of the black box]

Makes sense?

vext01:
Yes, I'd expect the compiler to use the stack if the Ok case is larger than a register. Indeed:

```
call void @_ZN7example6wiggle17h4367e95236baf0f2E(%"core::result::Result<u128, alloc::boxed::Box<InvalidTraceError>>"* noalias nocapture nonnull sret dereferenceable(24) %_12)
```

tiri/src/lib.rs
ltratt: What happens if you put the `black_box` around the outer function in the second case?

vext01: I suspect the call will be replaced with `999` again.

[black_boxes around entire function calls are useful for something else though: keeping code from re-ordering]

One moment...

vext01: As expected, the compiler performs constant propagation:
image

ltratt: That's very interesting! `black_box` does not work as I expected, which is probably my fault.

OK, so going back to this PR ;) I now think we should drop the outer `black_box` call since it doesn't do anything useful in this context?

vext01: Yep, I agree.

vext01:
And what do we learn about Boxing the error type? I suppose we shouldn't as it seems LLVM does a good job of optimising it if it gets big. Agree?

tiri/src/lib.rs
vext01: You've also highlighted that the Rust `black_box` docs could be improved. I might do a PR against upstream.

ltratt: Cool!

ltratt:
Agreed.

vext01:
The last two small changes address the outstanding comments.

ltratt:
Please squash.

vext01:
splat

PR #26, opened 2019-09-30
Implement trace trimming.

vext01:
See the commit message for a detailed description.

A companion PR follows. This PR ought to go in first, but we have a CI cycle. We need to think about a better way to deal with these kinds of breaking changes, as I expect many more in the future.

Companion PR: https://github.com/softdevteam/ykrustc/pull/55

### DIFF
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -108,6 +108,12 @@ impl Display for DefId {
     }
 }
 
+/// Bits in the `flags` bitfield in `Body`.
+pub mod bodyflags {
+    pub const TRACE_HEAD: u8 = 1;
+    pub const TRACE_TAIL: u8 = 1 << 1;
+}
+
 /// A tracing IR pack.
 /// Each Body maps to exactly one MIR Body.
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
@@ -115,14 +121,16 @@ pub struct Body {
     pub def_id: DefId,
     pub def_path_str: String,
     pub blocks: Vec<BasicBlock>,
+    pub flags: u8,
 }
 
 impl Body {
-    pub fn new(def_id: DefId, def_path_str: String, blocks: Vec<BasicBlock>) -> Self {
+    pub fn new(def_id: DefId, def_path_str: String, blocks: Vec<BasicBlock>, flags: u8) -> Self {
         Self {
             def_id,
             def_path_str,
             blocks,
+            flags,
         }
     }
 }
--- a/yktrace/src/debug.rs
+++ b/yktrace/src/debug.rs
@@ -9,11 +9,11 @@
 
 //! Debugging utilities.
 
-use crate::SirTrace;
+use crate::{tir::SIR, SirLoc, SirTrace};
 use elf;
 use fallible_iterator::FallibleIterator;
-use std::{collections::HashMap, env, io::Cursor};
-pub use ykpack::Statement;
+use std::{collections::HashMap, env, io::Cursor, iter::IntoIterator};
+pub use ykpack::{bodyflags, Statement};
 use ykpack::{Decoder, DefId, Pack};
 
 // The SIR Debug Map lets us map a DefId to a definition path.
@@ -48,21 +48,38 @@ pub fn def_path(def_id: &DefId) -> Option<&str> {
 }
 
 /// Prints a SIR trace to stdout for debugging purposes.
-pub fn print_sir_trace(trace: &dyn SirTrace) {
+pub fn print_sir_trace(trace: &dyn SirTrace, trimmed: bool) {
+    let locs: Vec<&SirLoc> = match trimmed {
+        false => (0..(trace.raw_len())).map(|i| trace.raw_loc(i)).collect(),
+        true => trace.into_iter().collect()
+    };
+
     println!("---[ BEGIN SIR TRACE DUMP ]---");
-    for loc in trace {
+    for loc in locs {
         let def_id = DefId::from_sir_loc(&loc);
         let def_path_s = match def_path(&def_id) {
             Some(s) => s,
             None => "<unknown>"
         };
+
         println!(
-            "[{}] crate={}, index={}, bb={}",
+            "[{}] crate={}, index={}, bb={}, flags=[",
             def_path_s,
             loc.crate_hash(),
             loc.def_idx(),
-            loc.bb_idx()
+            loc.bb_idx(),
         );
+
+        let body = SIR.bodies.get(&def_id);
+        if let Some(body) = body {
+            if body.flags & bodyflags::TRACE_HEAD != 0 {
+                print!("HEAD ");
+            }
+            if body.flags & bodyflags::TRACE_TAIL != 0 {
+                print!("TAIL ");
+            }
+        }
+        println!("]");
     }
     println!("---[ END SIR TRACE DUMP ]---");
 }
--- a/yktrace/src/lib.rs
+++ b/yktrace/src/lib.rs
@@ -13,10 +13,7 @@
 extern crate test;
 
 use core::yk_swt::SirLoc;
-use std::{
-    fmt::Debug,
-    iter::{IntoIterator, Iterator}
-};
+use std::{fmt::Debug, iter::Iterator};
 #[macro_use]
 extern crate lazy_static;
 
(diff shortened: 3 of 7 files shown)


yktrace/src/lib.rs
```diff
@@ -13,10 +13,7 @@
 extern crate test;
 
 use core::yk_swt::SirLoc;
-use std::{
```
vext01: This hunk is due to a rustfmt change. I wasn't sure if it should be in a separate commit or not.

ykpack/src/types.rs
```diff
@@ -108,21 +108,29 @@ impl Display for DefId {
     }
 }
 
+/// Bits in the `flags` bitfield in `Body`.
+pub mod bodyflags {
+    pub const TRACE_HEAD: u8 = 1;
+    pub const TRACE_TAIL: u8 = 1 << 1;
```
ptersilie: `pub const TRACE_TAIL: u8 = 2;`
or
`pub const TRACE_TAIL: u8 = 0x10;`
?

ptersilie: I would probably write it like this:
```
pub const TRACE_HEAD: u8 = 0x01;
pub const TRACE_TAIL: u8 = 0x10;
```
So it's immediately clear these are flags, and doesn't require any additional decoding by someone reading the code.

yktrace/src/lib.rs
```diff
@@ -26,42 +23,70 @@ mod swt;
...
-        if self.next_idx < self.trace.len() {
-            let ret = self.trace.loc(self.next_idx);
-            self.next_idx += 1;
-            Some(ret)
+        if self.next_idx < self.trace.raw_len() {
+            let def_id = DefId::from_sir_loc(self.trace.raw_loc(self.next_idx));
+            if SIR.markers.trace_tails.contains(&def_id) {
+                // Stop when we find the start of the code that stops the tracer, thus trimming the
+                // end of the trace. By setting the next index to one above the last one in the
+                // trace, we we ensure the iterator will return `None` forever more.
```
ptersilie: `we we` typo

ykpack/src/types.rs
vext01: I hate to break it to you, but `0x10` is not 2 :P

But more seriously, the `<<` notation is more conventional, because it means you don't have to calculate the power of 2 manually. Admittedly that's not a big issue for a `u8` bitfield, but still.

yktrace/src/lib.rs
vext01: Good catch. Thanks.

yktrace/src/tir.rs
```diff
@@ -19,50 +19,97 @@ use std::{collections::HashMap, convert::TryFrom, env, io::Cursor};
...
 pub struct TirTrace {
     ops: Vec<TirOp>
 }
 
 impl TirTrace {
-    /// Create a TirTrace from a SirTrace.
-    /// Returns Err if a DefId is encountered for which no SIR is available.
-    pub(crate) fn new<'s>(trace: &'s dyn SirTrace) -> Result<Self, InvalidTraceError> {
+    /// Create a TirTrace from a SirTrace, trimming remnants of the code which starts/stops the
+    /// tracer. Returns a TIR trace and the bounds the SIR trace was trimmed to, or Err if a DefId
```
ptersilie: `and the(n) bounds the SIR trace` typo

ykpack/src/types.rs
ptersilie: > I hate to break it to you, but 0x10 is not 2 :P
Haha, sorry, I meant `0b02`. :D

> But more seriously, the << notation is more conventional
If this is the convention then that's fair enough. I personally just find it easier to read `0b10` instead of `1 << 1`.

vext01: `0b` is fine for a `u8`, but now consider a u32:
`0b00000000100000000000000000000000` vs. `1<<23`.

Or a `u64`...

ltratt: Can a trace be both `TRACE_HEAD` and `TRACE_TAIL`? If not, then there's no need to worry about bits in quite this way: just a numeric differentiation would be enough, right?

yktrace/src/debug.rs
```diff
@@ -48,20 +48,38 @@ pub fn def_path(def_id: &DefId) -> Option<&str> {
...
-    for loc in trace {
+    for loc in locs {
         let def_id = DefId::from_sir_loc(&loc);
         let def_path_s = match def_path(&def_id) {
             Some(s) => s,
             None => "<unknown>"
         };
+
+        let body = SIR.bodies.get(&def_id);
+        let mut flags = Vec::new();
```
ltratt: This intermediate `Vec` feels like a real song and dance for nothing. Why not move the `if` after the `println` and just `println("HEAD")` (or `TAIL`) as appropriate?

yktrace/src/lib.rs
```diff
@@ -207,12 +233,43 @@ mod test_helpers {
...
+    fn trim_trace() {
+        let tracer = start_tracing(Some(TracingKind::SoftwareTracing));
+        work(black_box(100));
+        let sir_trace = tracer.t_impl.stop_tracing().unwrap();
+
+        let contains_tracer_start_stop = |locs: Vec<&SirLoc>| {
+            let mut found_start_code = false;
+            let mut found_stop_code = false;
+            for loc in locs {
+                let body = match SIR.bodies.get(&DefId::from_sir_loc(&loc)) {
```
ltratt: This can just be an `unwrap` I think?

yktrace/src/tir.rs
```diff
@@ -19,50 +19,97 @@ use std::{collections::HashMap, convert::TryFrom, env, io::Cursor};
...
+pub struct Sir {
+    /// Lets us map a DefId from a trace to a SIR body.
+    pub bodies: HashMap<DefId, Body>,
+    // Interesting locations that we need quick access to.
+    pub markers: SirMarkers
+}
+
+/// Contains the DefIds of interesting locations required for trace manipulation.
+pub struct SirMarkers {
+    /// The function which starts tracing and whose suffix gets trimmed off the top of traces.
```
ltratt: It's not the function's suffix that gets trimmed off, I assume, but the whole function that gets trimmed off the trace?

yktrace/src/tir.rs
```diff
@@ -19,50 +19,97 @@ use std::{collections::HashMap, convert::TryFrom, env, io::Cursor};
...
+    /// There is only one such function: `yktrace::start_tracing()`.
+    ///
+    /// The compiler is made aware of this location by the `#[trace_head]` annotation.
+    pub trace_head: DefId,
+    /// Functions which stop tracing and whose prefix gets trimmed off the bottom of traces.
```
ltratt: Similarly, I assume the entirety of this function is removed from the end of the trace?

yktrace/src/tir.rs
```diff
@@ -19,50 +19,97 @@ use std::{collections::HashMap, convert::TryFrom, env, io::Cursor};
...
-    pub(crate) fn new<'s>(trace: &'s dyn SirTrace) -> Result<Self, InvalidTraceError> {
+    /// Create a TirTrace from a SirTrace, trimming remnants of the code which starts/stops the
+    /// tracer. Returns a TIR trace and the bounds the SIR trace was trimmed to, or Err if a DefId
+    /// is encountered for which no SIR is available.
+    pub fn new<'s>(trace: &'s dyn SirTrace) -> Result<Self, InvalidTraceError> {
         let mut ops = Vec::new();
-        let num_locs = trace.len();
+        let mut itr = trace.into_iter().peekable();
+        loop {
+            let loc = itr.next();
```
ltratt: Isn't this just a `for` loop in disguise?

ykpack/src/types.rs
vext01: In theory it could happen, but I doubt it. I've used a flags bitfield to allow for additional flags in the future.

yktrace/src/debug.rs
vext01: Because there may be more than one flag per trace location.

yktrace/src/lib.rs
vext01: Ah yes.

yktrace/src/debug.rs
ltratt: OK... but then still just do it as a series of `print`s after the one that's already there maybe?

yktrace/src/tir.rs
vext01: Yes, but the `peek()` later causes borrow check issues if you use a for loop.

yktrace/src/tir.rs
ltratt: Aha, I see what you mean. OK, maybe tweak a bit to make this clearer?

yktrace/src/tir.rs
ltratt: Ah, OK. So maybe:

```rust
loop {
    if let Some(loc) = itr.next() {
        ...
    } else {
        break;
    }
}
```

although that may also break the borrow checker... I'm just trying to avoid the conceptual and run-time overhead of `X.is_none(); ... ; X.unwrap()`.

yktrace/src/debug.rs
vext01: Yeah, I suppose we could do. I didn't put a great deal of effort into optimising this, as it's debug code that won't be called in production.

yktrace/src/tir.rs
vext01: Sure.

yktrace/src/debug.rs
ltratt: Understood. At the moment the intermediate list is just a bit of a distraction, so I think simplifying is a good idea.

yktrace/src/tir.rs
vext01: Your solution does work:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=77ecf18f4300a16e725651434c42825d

Whereas this is what I'd hoped to write:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ccf28d6a3d8635b051d65e554600b442

ltratt: Ah, the borrow checker :)

bjorn3: Have you tried `while let Some(loc) = iter.next() { ... }`?

vext01: Indeed. Just got the same suggestion from someone on #rust IRC:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d6ee07b93af9c802f93c9eea076b3ebb

ltratt: Good idea!

vext01: Or even better, check this out:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4f03faceeef0f67cf30e7446f3b9e00b

Credit: M-ou-se on IRC :)

ltratt: Does that work OK for the last element in the list?

vext01: Ah, you are right. It's only gone around the loop twice. Shame.
```
[src/main.rs:5] i[0] = 1
[src/main.rs:6] i[1] = 2
[src/main.rs:5] i[0] = 2
[src/main.rs:6] i[1] = 3
```

vext01:
I think that addresses the comments.

ltratt:
Please squash.

vext01:
splat, but we need to look at the other PR first.

vext01:
OK, so the passing CI from ykrustc will have made new tarballs using the new ykpack and we can r+ this now.

PR #29, opened 2019-10-04
Sir changes

vext01:
These changes are the first phase needed to have a proof of concept TIR interpreter (in the long run we will *compile* traces, and these changes help there too).

A companion PR will follow and I also expect this to make a [CI cycle](https://github.com/softdevteam/ykrustc/wiki/Continuous-Integration-Cycles).

Companion PR: https://github.com/softdevteam/ykrustc/pull/60

Previously, the on-disk SIR had included calls to to trace recorder if software tracing was in effect. This meant that I had to unpick the SIR structure at runtime and besides, we don't actually care to trace the trace recorder. I realised that I could just encode SIR in such a way that the calls to the trace recorder don't appear.

Second, when a trace follows a function call, the callee has its own local variables but with indices which would clash with the caller's. In order to interpret or compile the trace we'd have to do *something* so that the callee doesn't hose the caller's variables. How we do this, using a stack or variable renaming, doesn't matter for now, but either way, we need to know how many arguments and how many local variables each function uses.

### DIFF
--- a/ykpack/src/lib.rs
+++ b/ykpack/src/lib.rs
@@ -71,12 +71,14 @@ mod tests {
             BasicBlock::new(stmts1_b1, dummy_term.clone()),
             BasicBlock::new(stmts1_b2, dummy_term.clone()),
         ];
-        let sir1 = Pack::Body(Body::new(
-            DefId::new(1, 2),
-            String::from("item1"),
-            blocks1,
-            0,
-        ));
+        let sir1 = Pack::Body(Body {
+            def_id: DefId::new(1, 2),
+            def_path_str: String::from("item1"),
+            blocks: blocks1,
+            num_args: 3,
+            num_locals: 4,
+            flags: 0,
+        });
 
         let stmts2_b1 = vec![Statement::Nop; 7];
         let stmts2_b2 = vec![Statement::Nop; 200];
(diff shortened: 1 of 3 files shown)


vext01:
bors try

verifying that this makes a CI cycle.

vext01:
```
89  |         let sir2 = Pack::Body(Body::new(
    |                                     ^^^ function or associated item not found in `types::Body`
```

That's not the failure I was looking for. Let me fix that.

vext01:
bors try

vext01:
```
thread 'debug::tests::test_load_debug_map' panicked at 'called `Result::unwrap()` on an `Err` value: Syntax("invalid length 4, expected struct Body with 6 elements")', src/libcore/result.rs:1165:5
```

There we go. This is a CI cycle. See the wiki post linked in the description for how to fix this.

ptersilie:
Okay, I've already reviewed all PRs. I suggest you do squashing on all of them if it needs to be done. And then you add the depdency commit to `ykrustc` and I merge it, right?

vext01:
Squashed. Will now override the ykpack dep in ykrustc to `b858c5c`.

vext01:
This PR is phase 2 in the CI cycle. It's ready for bors.

PR #30, opened 2019-10-25
SIR changes in preparation for trace interpretation.

vext01:
In some places, the structure of SIR needlessly differed from MIR. It
becomes quite confusing to keep track of the two similar
representations, so we should only really diverge if we have good reason
to do so.

The representation of `Rvalue` is much more similar now, and we
introduce the notion of a `Place` into SIR, as this is an intuitive
abstraction for a storage location.

This is going to need to be a 3-phase commit. ~~Compiler PR coming soon.~~

Compiler change:
https://github.com/softdevteam/ykrustc/pull/63

Review this one first.

### DIFF
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -55,7 +55,7 @@ macro_rules! new_ser128 {
 new_ser128!(SerU128, u128);
 new_ser128!(SerI128, i128);
 
-#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Copy)]
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Copy, Hash)]
 pub struct Local {
     idx: LocalIndex,
     ty: TyIndex,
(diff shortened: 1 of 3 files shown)


ykpack/src/types.rs
```diff
@@ -81,6 +81,65 @@ impl Display for Local {
...
+            base: PlaceBase::Local(l),
+            projections: Vec::new(),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Hash)]
+pub enum PlaceBase {
+    Local(Local),
+    Static, // FIXME not implemented
```
ptersilie: Should we fix this?

vext01: I think we should implement that in a future PR.

There are lots of language features that we don't yet support which we can incrementally implement as needed.

It's like your gameboy emulator. Implementing all opcodes at once would kill you.

ptersilie: Point taken!

ptersilie:
Right, I think this is ready for the 3-phase commit now. However, we probably want to address the comment in the other PR first.

vext01:
Agreed.

ptersilie:
`bors r+`

vext01:
Bors doesn't like verbose formatting ;)

ptersilie:
Hah, no idea why I put that in code.

bors r+

vext01:
This is a rustfmt failure. Odd, because I thought I had already done a fmt. Investigating.

vext01:
Fixed. OK to squash?

ptersilie:
Yes, please squash.

vext01:
Squashed.

PR #31, opened 2019-10-29
Interpreting our first trace.

vext01:
This gets us to the point where we can interpret a simple trace.

See commit messages for more.

Note that the interpreter can only do basic stuff with integer types, and will currently treat signed types incorrectly.

### DIFF
--- a/tiri/src/lib.rs
+++ b/tiri/src/lib.rs
@@ -9,30 +9,131 @@
 
 //! Tiri -- TIR trace interpreter.
 //!
+//! This is a proof of concept interpreter for TIR traces. It is fairly simple since TIR traces are
+//! straight line code for one path through a SIR flow-graph and (most) conventional control flow
+//! has been replaced by guards (calls can still appear in traces).
+//!
+//! The interpreter takes as input a TIR trace and repeatedly evaluates TIR statements at the index
+//! indicated by the program counter (initially zero). Each statement will update the program
+//! counter accordingly.
+//!
+//! FIXME: talk about memory once it is implemented.
+//!
+//! FIXME: talk about calls once they are implemented.
+//!
+//! Interpretation continues until either a guard fails, or the end of the trace is reached. When
+//! interpretation is complete, the interpreter state is returned, allowing us to inspect the
+//! values of local variables and whether the trace prematurely aborted.
+//!
 //! No effort has been made to make this fast.
 
+#![feature(exclusive_range_pattern)]
 #![feature(test)]
 extern crate test;
 
-use yktrace::tir::{Guard, Statement, TirOp, TirTrace};
+use std::collections::HashMap;
+use yktrace::tir::{
+    BinOp, Constant, ConstantInt, Guard, Local, Operand, Place, PlaceBase, PlaceProjection, Rvalue,
+    SignedInt, Statement, TirOp, TirTrace, UnsignedInt,
+};
 
-/// Storage space for one local variable.
-/// FIXME: Not yet populated.
-struct Local {}
+/// A `Value` represents the contents of a local variable.
+#[derive(Debug, Clone, Eq, PartialEq)]
+enum Value {
+    /// A small (u128-sized or smaller) value.
+    Scalar(Scalar),
+    /// A pair of the above, used for checked operations.
+    ScalarPair(Scalar, Scalar),
+    // FIXME -- Implement memory accesses.
+}
+
+impl Value {
+    /// A helper to create a raw scalar value of the given value and size.
+    #[cfg(test)]
+    fn raw_scalar(data: u128, size: u8) -> Self {
+        Self::Scalar(Scalar::Raw { data, size })
+    }
+
+    /// A helper to cast a value to a Boolean, if appropriate.
+    fn as_bool(&self) -> bool {
+        match self {
+            // Casting a `ScalarPair` to a `bool` performs a scalar cast on the second element.
+            // This is used to inspect the overflow status of a checked binary operation.
+            Self::Scalar(s) | Self::ScalarPair(_, s) => s.as_bool(),
+        }
+    }
+
+    /// A helper to get the u128 value from inside a raw immediate.
+    fn as_u128(&self) -> u128 {
+        match self {
+            Self::Scalar(Scalar::Raw { data, .. }) => *data,
+            _ => panic!("Invalid cast from Value to u128"),
+        }
+    }
+}
+
+impl From<bool> for Value {
+    fn from(b: bool) -> Self {
+        Self::Scalar(Scalar::from(b))
+    }
+}
+
+/// A single u128-sized (or smaller) value.
+#[derive(Debug, Clone, Eq, PartialEq)]
+enum Scalar {
+    Raw { data: u128, size: u8 },
+    _Ptr, // FIXME Not implemented.
+}
+
+impl From<bool> for Scalar {
+    fn from(b: bool) -> Self {
+        if b {
+            Scalar::Raw { data: 1, size: 1 }
+        } else {
+            Scalar::Raw { data: 0, size: 1 }
+        }
+    }
+}
+
+impl Scalar {
+    fn as_bool(&self) -> bool {
+        let data = match self {
+            Scalar::Raw { data, .. } => data,
+            Scalar::_Ptr => panic!("Invalid cast from Scalar to bool"),
+        };
+        *data != 0
+    }
+}
 
 /// Mutable interpreter state.
-struct InterpState {
+pub struct InterpState {
     /// The next position in the trace to interpret.
-    trace_pos: usize,
+    pc: usize,
     /// Local variable store.
-    _locals: Vec<Local>,
+    locals: HashMap<u32, Value>,
+    /// Set true upon aborting the execution of a trace.
+    abort: bool,
 }
 
 impl InterpState {
-    fn new() -> Self {
+    pub fn new() -> Self {
         Self {
-            trace_pos: 0,
-            _locals: Vec::new(),
+            pc: 0,
+            locals: HashMap::new(),
+            abort: false,
+        }
+    }
+}
+
+impl InterpState {
+    fn store_local(&mut self, local: &Local, val: Value) {
+        self.locals.insert(local.idx(), val);
+    }
+
+    fn local(&self, l: &Local) -> Value {
+        match self.locals.get(&l.idx()) {
+            Some(v) => v.clone(),
+            None => panic!("uninitialised read from ${}", l.idx()),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


tiri/src/lib.rs
```diff
@@ -11,28 +11,131 @@
...
-            _locals: Vec::new(),
+            pc: 0,
+            locals: HashMap::new(),
+            abort: false,
+        }
+    }
+}
+
+impl InterpState {
+    fn store_local(&mut self, local: &Local, val: Value) {
```
bjorn3: Maybe implement Copy for Local and pass it by val instead. That will be faster when this is not inlined.

tiri/src/lib.rs
```diff
@@ -11,28 +11,135 @@
...
+        self.locals.insert(local.idx(), val);
+    }
+
+    fn local(&self, l: &Local) -> Value {
```
bjorn3: Same here

tiri/src/lib.rs
```diff
@@ -11,28 +11,135 @@
...
 
 /// Mutable interpreter state.
-struct InterpState {
+pub struct InterpState {
     /// The next position in the trace to interpret.
-    trace_pos: usize,
+    pc: usize,
     /// Local variable store.
-    _locals: Vec<Local>,
+    locals: HashMap<u32, Value>,
```
bjorn3: MaybEme use HashMap<Local, Value>, or even rustc_index::IndexMap<Local, Option<Value>> is most locals are expected to be initialized.

tiri/src/lib.rs
vext01: I believe you, but why would it be faster?

tiri/src/lib.rs
vext01: Interesting. What are they initialised to before their first store?

tiri/src/lib.rs
bjorn3: Because otherwise calling `.idx()` requires accessing the memory behind a pointer, while using `Local` without `&` it is possible (and likely) for the compiler to put the `Local` in a register.

vext01: OK.

(we are likely to make `Local` a plain integer in the future anyway, so we will in fact copy soon)

tiri/src/lib.rs
bjorn3:
```diff
-rustc_index::IndexMap<Local, Option>
+rustc_index::IndexVec<Local, Option<Value>>`
```

> What are they initialised to before their first store?

You will need to add elements to it yourself, but using [`IndexVec::from_elem_n`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_index/vec/struct.IndexVec.html#method.from_elem_n) you can easily create an `IndexVec` with multiple identical elements, like `None`.

tiri/src/lib.rs
bjorn3: I thought `Local` was already just a newtype around an integer.

ltratt:
I don't say this very often, but we need to add a few comments to this code: at the moment I can't really work out what's going on. In particular, we need to be clear what the major structs/enums are e.g. `Value`, `Intermediate`, `Scalar` and so on. I also think the commit messages need to be clearer about what we're doing: I get that this is, in some sense, an AST interpreter, but I don't know where to start looking to get some sense of the control flow.

vext01:
I'll add comments.

vext01:
> In particular, we need to be clear what the major structs/enums are e.g. Value, Intermediate, Scalar and so on.

The last push adds comments to these types.

> I get that this is, in some sense, an AST interpreter

It's not! I've added some blurb in the module-level comment to try and clarify.

Any good?

ltratt:
I don't have any experience with Miri, and we're not likely to follow Miri for very long (I suspect), so it's not ideal to direct people to Miri. I think this repo needs to stand on its own two feet. Can we try and make it self contained?

vext01:
Is this any better? I've killed all references to Miri and added an overview.

tiri/src/lib.rs
```diff
@@ -13,9 +13,21 @@
...
+//! Program state is mutated through assignments to local variables, each of which is an instance
+//! of `Value`. Memory locations are mutated via the `Value::Indirect` variant.
+//!
+//! When a guard is encountered, the guard condition must evaluate true for interpretation to
+//! continue; otherwise interpretation is aborted.
+//!
+//! FIXME: talk about calls once they are implemented.
+//!
+//! Interpretation continues until either a guard fails, or the end of the trace is reached. When
+//! interpretation is complete, the state of the interpreter state is returned, allowing us to
```
vext01: Just noticed another typo here. Sorry, fixing.

vext01:
Ready for re-review.

tiri/src/lib.rs
```diff
@@ -13,9 +13,21 @@
...
-//! evaluate compile-time constant code). Where a Tiri struct has a direct Miri equivalent, a
-//! comment will point this out. You can find the Miri structs in src/librustc_mir/interpret/.
+//! The interpreter takes as input a TIR trace and repeatedly evaluates TIR statements at the index
+//! indiciated by the program counter (initially zero). Each statement will update the program
+//! counter accordingly.
+//!
+//! Program state is mutated through assignments to local variables, each of which is an instance
+//! of `Value`. Memory locations are mutated via the `Value::Indirect` variant.
+//!
+//! When a guard is encountered, the guard condition must evaluate true for interpretation to
```
ltratt: I think this para is redundant given that the para below says the same (minus the bit about aborting).

tiri/src/lib.rs
```diff
@@ -34,10 +46,11 @@ use yktrace::tir::{
 const DUMP_CONTEXT: usize = 4;
 
 /// Each local variable is represented by one `Value` struct.
-/// This corresponds with the Miri's `Operand` type.
 #[derive(Debug, Clone, Eq, PartialEq)]
 enum Value {
+    /// A stand-alone immediate value.
     Immediate(Immediate),
+    /// Memory.
```
ltratt: Heap memory? Stack memory? Both?

ltratt:
OK, this is starting to make more sense. I think we need to define our terminology a tiny bit more carefully, but we're not far off!

tiri/src/lib.rs
vext01: OK

tiri/src/lib.rs
vext01: I believe both. Let's roll with both for now.

tiri/src/lib.rs
```diff
@@ -34,10 +46,11 @@ use yktrace::tir::{
 const DUMP_CONTEXT: usize = 4;
 
 /// Each local variable is represented by one `Value` struct.
-/// This corresponds with the Miri's `Operand` type.
 #[derive(Debug, Clone, Eq, PartialEq)]
 enum Value {
+    /// A stand-alone immediate value.
```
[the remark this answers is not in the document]
vext01: Either variant of this struct are storage for local variables. Immediates are the subset not referring to a memory address.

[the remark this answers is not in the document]
ltratt: I must admit, that doesn't make much sense to me. Remember I'm dumb :)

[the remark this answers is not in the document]
vext01: I borrowed this from Miri as it seemed a sensible way to do things. Stuff that fits in a register is an "immediate" value whereas other stuff which has to be in-memory is a different variant.

So stuff word-size integers (or smaller) are immediate values, whereas a large struct is not immediate, but "indirect".

I think that's a useful distinction to make, but I can be swayed if you disagree. This is new to me too.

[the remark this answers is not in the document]
vext01: I lifted the names `Immediate` and `Indirect` from Miri, but I also know them as ( similar) addressing modes in assembler programming. I suggest we keep them.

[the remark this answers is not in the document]
vext01: Here's the Intel manual's blurb on immediates:
image

Their notion is little different, as they have a separate addressing mode for registers, which are a bit like local variables.

We probably don't want to obsess about CPU semantics too much though ;)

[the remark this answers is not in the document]
ltratt: I'm OK with the names, so long as we clearly document what they actually mean. I would never have guessed e.g. that `Immediate` means "register sized thing", so that's the key bit of information etc. etc.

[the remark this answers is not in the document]
vext01: I've just realised that Miri bends the rules a little, in that immediates are typically larger than register size.

An Immeidate is either a Scalar or a ScalarPair, and each Scalar is by a u128 value with a size....

I'll find some wording that makes sense...

tiri/src/lib.rs
```diff
@@ -45,12 +42,13 @@ use yktrace::tir::{
 #[allow(dead_code)]
 const DUMP_CONTEXT: usize = 4;
 
-/// Each local variable is represented by one `Value` struct.
+/// The contents of each local variable is represented by a `Value`.
```
ltratt: How about just "A `Value` represents the contents of a local variable."

tiri/src/lib.rs
```diff
@@ -45,12 +42,13 @@ use yktrace::tir::{
...
 #[derive(Debug, Clone, Eq, PartialEq)]
 enum Value {
-    /// A stand-alone immediate value.
+    /// A value that can be stored immediately into a local variable without touching memory.
```
ltratt: I think this might need to be something like "A `Value` which is small enough that it can potentially be stored in 1 or more registers."? I'm not sure if the "potentially" thing is true or not though!

tiri/src/lib.rs
```diff
@@ -45,12 +42,13 @@ use yktrace::tir::{
...
     Immediate(Immediate),
-    /// Memory.
+    /// A reference to a potentially larger (and potentially composite) value in memory (either on
```
ltratt: "A `Value` which is too big for us to want to store in register(s)." Note that I'm assuming that there are some values which are, say, 6 machine words big and which we *could* store in registers, but won't, because it's not a very good use of them.

tiri/src/lib.rs
ltratt: I'm trying to understand what the difference between these two things is, so I'm trying to find a description you're unhappy enough with that you'll fix to something better :)

There is no machine I know of that has 4GiB of registers, so that's why I wondered if the "fits in one or more registers" things might be the right route. I'm not saying it is, but hopefully you can see what I'm getting at!

tiri/src/lib.rs
```diff
@@ -84,8 +82,7 @@ impl From<bool> for Value {
     }
 }
 
-/// An immediate value: a value immediately available, and which doesn't require any kind of lookup
-/// or memory access.
+/// A value which can be stored immediately inside a local variable without touching memory.
```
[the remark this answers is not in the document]
ltratt: >Why don't we try rolling immediate into Value and seeing how far we get? 

Works for me.

> It's documented against the struct, (but not the variant as) "A scalar (non-compound) value.".

If I'm being cruel, defining "X" by using the name "X" doesn't help me. I really can't derive any meaningful information from that comment at the moment. Remember (repeat after me)... Laurie is dumb and needs things spelled out!

tiri/src/lib.rs
```diff
@@ -45,67 +45,47 @@ const DUMP_CONTEXT: usize = 4;
...
     fn raw_scalar(data: u128, size: u8) -> Self {
-        Value::Immediate(Immediate::Scalar(Scalar::Raw { data, size }))
+        Self::Scalar(Scalar::Raw { data, size })
     }
 
     /// A helper to cast a value to a Boolean, if appropriate.
     fn as_bool(&self) -> bool {
         match self {
-            Value::Immediate(i) => i.as_bool(),
+            // In the case of a `ScalarPair` we assume we are casting to bool to inspect the result
```
ltratt: "Assume" is a bit vague. Is this always the case or ...?

ltratt:
This definitely feels clearer to me. What about you?

tiri/src/lib.rs
vext01: My current thinking is that we shouldn't generate SIR where you want to cast the first element of a pair to a bool. Let me fiddle the comment.

vext01:
I think I can live with it.

tiri/src/lib.rs
ltratt: OK.

ltratt:
OK, cool. Can I suggest you do a squash, I'll do a light second round of reviewing, then hopefully we can get to the point of merging soon.

tiri/src/lib.rs
```diff
@@ -9,30 +9,134 @@
...
 //!
+//! This is a proof of concept interpreter for TIR traces. It is fairly simple since TIR traces are
+//! straight line code for one path through a SIR flow-graph and (most) conventional control flow
+//! has been replaced by guards (calls can still appear in traces).
+//!
+//! The interpreter takes as input a TIR trace and repeatedly evaluates TIR statements at the index
+//! indicated by the program counter (initially zero). Each statement will update the program
+//! counter accordingly.
+//!
+//! Program state is mutated through assignments to local variables, each of which is an instance
```
ltratt: This comment no longer seems to make sense: it suggest variables map to `Value`, but memory locations (what are they? heap? stack?) map to `Value::Memory`.

tiri/src/lib.rs
ltratt: Ping.

tiri/src/lib.rs
```diff
@@ -50,35 +154,186 @@ impl<'t> Interp<'t> {
...
+    fn eval_rvalue(&self, state: &InterpState, rval: &Rvalue) -> Value {
+        match rval {
+            Rvalue::Use(o) => self.eval_operand(state, o),
+            Rvalue::BinaryOp(oper, o1, o2) => self.eval_binop(state, oper, o1, o2, false),
+            Rvalue::CheckedBinaryOp(oper, o1, o2) => self.eval_binop(state, oper, o1, o2, true),
+            _ => panic!("unimplemented rvalue eval"),
+        }
+    }
+
+    fn eval_place(&self, state: &InterpState, place: &Place) -> Value {
```
ltratt: What does the `place` in `eval_place` mean?

tiri/src/lib.rs
vext01: I've deliberately been vague about stack and heap, as I'm not yet sure how I'm going to implement them.

In the meantime, would it help to just drop this sentence of the comment?

ltratt: As a good rule of thumb, if we're going to be vague, let's be clear about it :) In other words, when something's vague, and it's not clear why, I start to worry. I can't help it!

I think I'd ditch this comment and also delete `Value::_Memory`, which we don't use (I think), and which is possibly a bit confusing.

tiri/src/lib.rs
vext01: > (Although it's defined in Miri, the same type is directly re-used in MIR).

Hold up, that's nonsense! I meant to link to this one:
https://github.com/softdevteam/ykrustc/blob/e103291e3553ec58444caa056765e5c3312caae4/src/librustc/mir/mod.rs#L1841

(There are two `Place`s in upstream rust, one for MIR, once for Miri).

tiri/src/lib.rs
vext01: > I think I'd ditch this comment and also delete Value::_Memory, which we don't use (I think), and which is possibly a bit confusing.

That's fine. It will probably come back later, when we start dealing with the intepreter's notion of stack/heap. Is that OK?

yktrace/src/tir.rs
```diff
@@ -177,9 +187,19 @@ pub enum Guard {
     Integer(Place, SerU128),
     /// The local must not be a member of the specified collection of integers.
     /// This is necessary due to the "otherwise" semantics of the SwitchInt terminator in MIR.
-    OtherInteger(Place, Vec<SerU128>),
+    OtherInteger(Place, Vec<u128>),
```
ltratt: Is it just me, or does `Guard` always have a first argument of `Place`? At which point that suggests that `Guard` should be a `struct` along the lines of:

```rust
struct Guard {
    lhs: Place
    rhs: GuardKind
}

enum GuardKind {
    Dunno(Vec<u128>)
    Bool(bool)
}
```

tiri/src/lib.rs
ltratt: Works for me!

yktrace/src/tir.rs
vext01: We could do. I'm indifferent. Yours makes sense, but mine is marginally easier to de-structure. You choose :)

ltratt: We're going to add more variants to `GuardKind` I assume? In which case the split I suggested is definitely more idiomatic. If not, then it's maybe less important.

vext01:
I think that fixes all of your comments.

ltratt:
Please squash.

vext01:
Splat.

(It made sense to separate the guard change out)

vext01:
Bors task is queued. Lukas has builds running.

PR #43, opened 2020-02-19
Disable software tracing and default to hardware tracing.

vext01:


ptersilie:
Hah, that change was smaller than I thought it would be.

bors r+

PR #45, opened 2020-02-19
Add return statement to ykpack

ptersilie:


ykpack/src/types.rs
```diff
@@ -170,6 +170,8 @@ pub enum Statement {
     Nop,
     /// An assignment.
     Assign(Place, Rvalue),
+    /// A return instruction
+    Return,
```
vext01: So later will will have an `Option` inside so we can return values?

ptersilie: Sure. I just wanted to get this PR out quickly so you are not blocked anymore. We can change this later.

PR #46, opened 2020-02-26
Add load and store statements.

vext01:
These are changes required by:
https://github.com/softdevteam/ykrustc/pull/85

My next yk PR will remove a load of stuff from ykpack. There's a lot of stuff there that no longer makes sense.

### DIFF
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -169,9 +169,10 @@ pub enum Statement {
     /// Do nothing.
     Nop,
     /// An assignment.
-    Assign(Place, Rvalue),
+    Assign(Local, Rvalue),
     /// A return instruction
     Return,
+    Store,
     /// Any unimplemented lowering maps to this variant.
     /// The string inside is the stringified MIR statement.
     Unimplemented(String),
@@ -183,6 +184,7 @@ impl Display for Statement {
             Statement::Nop => write!(f, "nop"),
             Statement::Assign(l, r) => write!(f, "{} = {}", l, r),
             Statement::Return => write!(f, "return"),
+            Statement::Store => write!(f, "store"),
             Statement::Unimplemented(mir_stmt) => write!(f, "unimplemented_stmt: {}", mir_stmt),
         }
     }
@@ -195,6 +197,7 @@ pub enum Rvalue {
     BinaryOp(BinOp, Operand, Operand),
     CheckedBinaryOp(BinOp, Operand, Operand),
     Unimplemented,
+    Load(LocalIndex), // FIXME arg may not be local.
 }
 
 impl Display for Rvalue {
@@ -206,6 +209,7 @@ impl Display for Rvalue {
                 write!(f, "checked_{}({}, {})", op, oper1, oper2)
             }
             Self::Unimplemented => write!(f, "unimplemented rvalue"),
+            Self::Load(l) => write!(f, "load({})", l),
         }
     }
 }
--- a/tiri/src/lib.rs
+++ b/tiri/src/lib.rs
@@ -192,13 +192,10 @@ impl<'t> Interp<'t> {
     /// Interpret the specified statement.
     fn interp_stmt(&self, state: &mut InterpState, stmt: &Statement) {
         match stmt {
-            Statement::Assign(plc, rval) => match (&plc.base, &plc.projections.len()) {
-                (PlaceBase::Local(l), 0) => {
-                    state.store_local(*l, self.eval_rvalue(state, rval).clone());
-                    state.pc += 1;
-                }
-                _ => unimplemented!("unhandled assignment"),
-            },
+            Statement::Assign(l, rval) => {
+                state.store_local(*l, self.eval_rvalue(state, rval).clone());
+                state.pc += 1;
+            }
             _ => panic!("unhandled statement: {}", stmt),
         }
     }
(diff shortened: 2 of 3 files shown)


ykpack/src/types.rs
```diff
@@ -195,6 +197,7 @@ pub enum Rvalue {
     BinaryOp(BinOp, Operand, Operand),
     CheckedBinaryOp(BinOp, Operand, Operand),
     Unimplemented,
+    Load(LocalIndex), // FIXME arg may not be local.
```
ltratt: Perhaps we should, at least temporarily, call this `LoadLocal`?

vext01: This will be fixed very soon, so I wouldn't worry about it for now.

ltratt: The immediate next PR or ...?

vext01: It'll be either the next or the one after.

I don't want to start fleshing out instructions implementations until I know that I can deal with other operand types. Spent some time the other day looking at how constants are born and I have a plan.

ltratt: Then I would still prefer to call this `LoadLocal` and then we no longer have the `FIXME` and the wait for a couple of PRs. As you can tell, I am allergic to `FIXME`s :p

vext01: We can call it `LoadLocal` if you wan't, but I'd still want a `FIXME` saying "this shouldn't be `LoadLocal`, it should be `Load` and accept other operand types too" :P

I suspect we are worrying too much too early. I don't even know if this approach is going to work yet.

vext01: Having thought some more: here's what we should do for the *next* PR.

Keep this statement as `Load`, but change the `Local` inside to `Operand` and have `Operand` be an enum with (for now) one `Local` variant. This implies there are more to come.

It'll be quite a large change because we already have a `Operand` enum which contains a load of old stuff and tests rely on them. I'd planned to sort this later, but we can bring that work forward to keep everyone happy :)

Sound good?

ltratt: We don't even really need to make a decision on this yet: if we say `LoadLocal` now, in a future PR, we might add `LoadGlobal` or rename this to `Load` and use an internal `enum` or whatever (FWIW, it's probably easier to have separate variants for LoadLocal/LoadGlobal rather than an internal VM, but this is partly a matter of taste). The beauty of calling it `LoadLocal` is that it doesn't need a `FIXME`: the name is impossible to misread.

vext01: Wait, there seems to be confusion over the role of the term "Local" here.

LLVM's `load` instruction loads something from memory. Where to load from is expressed by a pointer operand:
https://llvm.org/docs/LangRef.html#load-instruction

My `Load(Local)` variant allows a load from memory where the pointer operand is stored in a local variable. i.e.
```
%p = <some-pointer>
%r = load(%p)
```

Adding new instructions for different operand types doesn't seem like the right thing to do. Consider what would happen to a `Add` and `Sub`:
```
enum Statement {
    ...
    AddConstConst(Const, Const)
    AddLocalLocal(Local, Local)
    AddConstGlobal(Const, Global)
    AddLocalGlobal(Local, Global)
    AddGlobalGlobal(Global, Global)
    AddLocalConst(Local, Const)
    SubConstConst(Const, Const)
    SubLocalLocal(Local, Local)
    SubConstGlobal(Const, Global)
    SubLocalGlobal(Local, Global)
    SubGlobalGlobal(Global, Global)
    SubLocalConst(Local, Const)
}
```

It'd better be expressed as:
```
enum Statement {
    ...
    Add(Operand, Operand)
    Sub(Operand, Operand)
}

enum Operand {
    Local,
    Global,
    Const
}
```

Makes sense? Sorry, it's quite hard to explain.

ltratt: Dumb question: do we have to follow LLVM slavishly here or is this our own IR where we can structure things as we see fit?

vext01: We have a little bit of wiggle room, but not a lot. Sadly our IR is going to have to be very similar to LLVM's due to the way the codegen works.

ltratt: Yes, I understand that we're currently somewhat constrained, but does it need to be exactly the same in this case?

vext01: It may help to look at the point where the [generic backend code calls the llvm backend for a load](https://github.com/softdevteam/ykrustc/blob/043bf15b2c4fd0b21c02f6fe3d82bb8f663faf75/src/librustc_codegen_llvm/builder.rs#L438-L444):

```
    fn load(&mut self, ptr: &'ll Value, align: Align) -> &'ll Value {
        unsafe {
            let load = llvm::LLVMBuildLoad(self.llbuilder, ptr, UNNAMED);
            llvm::LLVMSetAlignment(load, align.bytes() as c_uint);
            load
        }
    }
```

`codegen_ssa` calls this passing a C++ pointer (`ptr`) which represents where to load from and returns another C++ pointer representing the result of the load.

In order for the Sir generator to know what `ptr` is, we must have cached its creation during an earlier call to the LLVM API.

That's the granularity we are working at unfortunately. We have to decide how each individual call to the LLVM API should mutate the SIR.

You are wondering why we are working so close to LLVM?

ltratt: > You are wondering why we are working so close to LLVM?

Yes, partly.

PR #47, opened 2020-03-06
Update dependency for object crate

ptersilie:
This updates the object crate to version 0.17, since otherwise we can't compile it with hardware tracing enabled.

We could merge this as is, or take the opportunity to update some other crates as well. Thoughts?


vext01:
> We could merge this as is, or take the opportunity to update some other crates as well. Thoughts?

I thought it would be good to get it in now and get us moving :)

PR #48, opened 2020-03-12
Allow passing strings to unimplemented Rvalues

ptersilie:
This allows us to add the type of the unimplemented Rvalue so that it
shows up in emitted sir code.

### DIFF
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -194,7 +194,7 @@ pub enum Rvalue {
     Use(Operand),
     BinaryOp(BinOp, Operand, Operand),
     CheckedBinaryOp(BinOp, Operand, Operand),
-    Unimplemented,
+    Unimplemented(String),
 }
 
 impl Display for Rvalue {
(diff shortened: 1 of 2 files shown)


vext01:
Fine. If we merge this, does it break the compiler? Do we need a 3-phase with a compiler change?

ptersilie:
I don't think the compiler currently has any code that uses `Rvalue::Unimplemented`, but I have a PR lined up that adds those, once this is merged.

vext01:
> Couldn't find the end of the code that starts the tracer.

I guess we don't expect that to work right now anyway? The mapper surely needs work first.

ptersilie:
I guess, though I'm not sure why this test isn't commented out.

vext01:
Please add a commit to skip it.

ptersilie:
Fixed.

vext01:
Ah. I think you annotated the test helper, but not the tests. There are two tests we should mark.

ptersilie:
Let's try this.

bors try

vext01:
please squash

ptersilie:
Squashed.

PR #49, opened 2020-03-13
Improve textual notation of Call targets.

vext01:
The textual notation for a Call target is kind of weird.

If there's a known symbol, it looks like this:
```
call operand=sym_name=_ZN50_$LT$T$u20$as$u20$core..convert..Into$LT$U$GT$$GT$4into17hcac79ab93d213ca6E, ret_bb=bb23
```

I think `operand=sym_name=` looks weird.

This change kills `sym_name=` and in case of an unknown target uses `<unknown>` (instead of `unknown`).

### DIFF
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -354,8 +354,8 @@ pub enum CallOperand {
 impl Display for CallOperand {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            CallOperand::Fn(sym_name) => write!(f, "sym_name={}", sym_name),
-            CallOperand::Unknown => write!(f, "unknown"),
+            CallOperand::Fn(sym_name) => write!(f, "{}", sym_name),
+            CallOperand::Unknown => write!(f, "<unknown>"),
         }
     }
 }


vext01:
> thread 'main' panicked at 'Cannot run dynamic test fn out-of-process', src/libtest/lib.rs:497:22

That's new...

bjorn3:
`-Zpanic-abort-tests` was recently enabled in `ykrustc`.

vext01:
> -Zpanic-abort-tests was recently enabled in ykrustc.

Yeah, I wondered if it might be related to that.

Next week I'll be investigating. At this point I'm not sure what a "dynamic test" is :P

bjorn3:
Dynamic tests are for example used by [`compiletest_rs`](https://crates.io/crates/compiletest_rs). Basically a dynamic test can be a closure, while a static test must be a function pointer. I can't find where you are using a dynamic test in `ykrt` though.

vext01:
I can confirm that the failure is caused by the abort strategy.

The failing tests are the benchmark tests marked `#[bench]`. Comment those and the failures go away...

vext01:
And here's the problem: you cannot do benchmark tests with the abort strategy
https://github.com/rust-lang/cargo/issues/3166

It looks like normal tests and benchmark tests get mixed into the same binary, so we can't build just benchmark tests with the unwind strategy and the others with the abort strategy. And besides, we probably do want to allow benchmarks of our tracer.

So we might have to honour the unwind strategy after all.

@ptersilie @ltratt any thoughts on this?

vext01:
I've reverted the switch to the abort strategy for now.

bors try

vext01:
@ptersilie this is ready for merge.

vext01:
bors ping

PR #50, opened 2020-03-17
Better call terminators

vext01:
This is the companion PR to:
https://github.com/softdevteam/ykrustc/pull/93

See commit messages for details.

### DIFF
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -56,17 +56,17 @@ impl Display for Local {
 
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Hash)]
 pub struct Place {
-    pub base: PlaceBase,
-    pub projections: Vec<PlaceProjection>,
+    pub local: Local,
+    pub projection: Vec<PlaceElem>,
 }
 
 impl Display for Place {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if self.projections.is_empty() {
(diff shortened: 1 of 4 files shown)


ykpack/src/types.rs
```diff
@@ -384,7 +388,9 @@ pub enum Terminator {
     },
     Call {
         operand: CallOperand,
-        ret_bb: Option<BasicBlockIndex>,
+        args: Vec<Operand>,
+        /// If the call converges, the return value and basic block to continue at.
```
ptersilie: Sentence doesn't parse.

vext01: Makes sense to me, but we can try a re-word.

How about:
```
`Some` if the call is convergent, `None` otherwise. When `Some` contains the return value and the block to return to.
```

ptersilie: This also doesn't parse:
When `Some` contains the return value and the block to return to...what?

ptersilie: I think it's the ending sentences on prepositions that confuses me.

ptersilie: Okay I think I understand it now. It makes more sense to me if we just turn the sentence around: `The return value and basic block to continue at, if the call converges.`

vext01: Fixed

ptersilie:
Please squash.

vext01:
squashed. companion PR need squash and cycle breaker.

PR #51, opened 2020-03-19
Start work on assembler backend

ptersilie:
**This is work in progress and not ready for merging yet**

This is a very basic outline of our assembler to be used as a replacement for the assembler library I am currently using in my [ykcompile](https://github.com/ptersilie/yk/tree/ykcompile) branch.

Before I dive deeper into this and make decisions I will later regret, I think it would be good to start some discussion on this and think together where we would like this to go and how would like this to work.

One immediately obvious flaw of this API is that any encoder needs to return a `Vec`, which is slow. A solution I see for this is to have the encoder write the machine code directly to memory, which is also not ideal from an API perspective. So I'm open to suggestions here.

Ready...fight!

### DIFF
--- a/assembler/src/lib.rs
+++ b/assembler/src/lib.rs
@@ -1,3 +1,5 @@
+mod backend;
+
 use libc;
 use std::mem;
 use std::process::Command;
@@ -18,11 +20,9 @@ impl Assembler {
             let mut page: *mut libc::c_void = mem::MaybeUninit::uninit().as_mut_ptr();
             libc::posix_memalign(&mut page, PAGE_SIZE, PAGE_SIZE);
             libc::mprotect(page, PAGE_SIZE, libc::PROT_WRITE);
-
-            // XXX correct use of assume_init()?
-
             // Transmute to raw u8 pointer
             let memmap: *mut u8 = mem::transmute(page);
+            // TODO correct use of MaybeUninit::assume_init()
             memmap
         };
 
@@ -33,17 +33,12 @@ impl Assembler {
     }
 
     pub fn add_instruction(&mut self, instr: &str) {
-        let output = Command::new("/usr/bin/rasm2")
-            .arg("-B")
-            .arg(instr)
-            .output()
-            .expect("failed to execute process");
-        let instr = output.stdout.as_ptr();
+        let mc = backend::encode(backend::EncoderBackend::RasmBackend, instr);
         unsafe {
             let off = self.memmap.offset(self.instrptr as isize);
-            libc::memcpy(off as *mut _, instr as *const _, output.stdout.len());
+            libc::memcpy(off as *mut _, mc.as_ptr() as *const _, mc.len());
         }
-        self.instrptr += output.stdout.len();
+        self.instrptr += mc.len();
     }
 
     fn make_executable(&self) {
--- a/assembler/src/lib.rs
+++ b/assembler/src/lib.rs
@@ -1,0 +1,79 @@
+use libc;
+use std::mem;
+use std::process::Command;
+
+// getpagesize() (sycall)
+const PAGE_SIZE: usize = 4096;
+
+struct Assembler {
+    memmap: *mut u8,
+    instrptr: usize
+}
+
+impl Assembler {
+
+    pub fn new() -> Self {
+        // Create a single page of memory to store our instructions in and make it writable
+        let memmap = unsafe {
+            let mut page: *mut libc::c_void = mem::MaybeUninit::uninit().as_mut_ptr();
+            libc::posix_memalign(&mut page, PAGE_SIZE, PAGE_SIZE);
+            libc::mprotect(page, PAGE_SIZE, libc::PROT_WRITE);
+
+            // XXX correct use of assume_init()?
+
+            // Transmute to raw u8 pointer
+            let memmap: *mut u8 = mem::transmute(page);
+            memmap
+        };
+
+        Assembler {
+            memmap,
+            instrptr: 0
+        }
+    }
+
+    pub fn add_instruction(&mut self, instr: &str) {
+        let output = Command::new("/usr/bin/rasm2")
+            .arg("-B")
+            .arg(instr)
+            .output()
+            .expect("failed to execute process");
+        let instr = output.stdout.as_ptr();
+        unsafe {
+            let off = self.memmap.offset(self.instrptr as isize);
+            libc::memcpy(off as *mut _, instr as *const _, output.stdout.len());
+        }
+        self.instrptr += output.stdout.len();
+    }
+
+    fn make_executable(&self) {
+        unsafe {
+            libc::mprotect(self.memmap as *mut _, PAGE_SIZE, libc::PROT_EXEC);
+        }
+    }
+
+    fn function_pointer(&self) -> fn() -> i64 {
+        unsafe {
+            mem::transmute(self.memmap)
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::Assembler;
+
+    #[test]
+    pub(crate) fn test_basic() {
+        let mut assemb = Assembler::new();
+        assemb.add_instruction("mov rdx, 3");
+        assemb.add_instruction("mov rcx, 4");
+        assemb.add_instruction("add rcx, rdx");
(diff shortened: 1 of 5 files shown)


assembler/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::Assembler;
+
+    #[test]
+    pub(crate) fn test_basic() {
+        let mut assemb = Assembler::new();
```
ltratt: I think we can separate out the "building instructions" phase from "something that's executable" phase? So we might end up with something like:

```
let mut asm = AsmBuilder::new();
asm.push("mov rdx, 3");
asm.push("mov rcx, 4");
...
let f = asm.assemble();
assert_eq!(f(), 7);
```

Where the type structure might look something like:

```
struct AsmBuilder { ... }

impl AsmBuilder {
    fn push(&mut self, instr: &str) { ... }
    fn assemble(self) -> Assembled { ... }
}

struct AssembledFunction { ... }

impl AssembledFunction {
    fn as_ptr(&self) -> ...
}
```

Note that `assemble` takes `self` (not `&self`): when you call "assemble" you would consume the `AsmBuilder`. There are some advantages to this:

* We can choose whether we convert strings to instructions on each `push` or at `assemble` (the latter stage allows us to do things with labels etc.).
* Once `assemble` is called, people can't add instructions to a page of memory marked as executable. This will make us more easily compatible with W^X.

I can't think of any immediate disadvantages. We could even have a way of consuming an `AssembledFunction` back to an `AsmBuilder` for when we want to patch code. But that's a little way off!

vext01:
> One immediately obvious flaw of this API is that any encoder needs to return a Vec, which is slow. A solution I see for this is to have the encoder write the machine code directly to memory, which is also not ideal from an API perspective

The memory you'd write directly to... you'd probably have to heap allocate that anyway (assembled code may get large), so I actually don't see much problem with using a `Vec`.

ltratt:
I do think we might want to take a leaf out of [dynasm's book](https://luajit.org/dynasm_examples.html) and have some sort of macro that allows people to build up chunks of asm code gradually. But that may be something we can transparently add on top once we've got the basic API right?

assembler/src/lib.rs
```diff
@@ -0,0 +1,74 @@
+mod backend;
+
+use libc;
+use std::mem;
+use std::process::Command;
+
+// getpagesize() (sycall)
```
vext01: obviously, replace the constant with the syscall :)

ltratt:
Rather than an encoder *returning* a `Vec`, why don't we pass it a `&mut Vec` to which it can append? That way we don't continually malloc new blocks.

vext01:
> Rather than an encoder returning a Vec, why don't we pass it a &mut Vec to which it can append? That way we don't continually malloc new blocks.

Sounds good to me.

assembler/src/lib.rs
vext01: I'm in favour of assembling as we go rather than caching strings. The binary encodings are going to be smaller, thus less memory to store.

Consuming self sounds good to me.

ltratt: > I'm in favour of assembling as we go rather than caching strings.

The API I've suggested doesn't expose to users which we do.

ptersilie: I like this. I assume the `AssembledFunction` would hold on to the memory map as well, so that it doesn't go out of scope. Is there anything else that `AssemblerFunction` could do apart from giving us a pointer?

vext01: > The API I've suggested doesn't expose to users which we do.

Yes

ltratt: > I assume the AssembledFunction would hold on to the memory map as well, so that it doesn't go out of scope

Yes, we can add a `Drop` to `AssembledFunction` that unmaps and frees the memory automatically.

> Is there anything else that AssemblerFunction could do apart from giving us a pointer?

I don't know what else it might need to do, but I haven't thought deeply about use cases here.

ptersilie:
> I do think we might want to take a leaf out of [dynasm's book](https://luajit.org/dynasm_examples.html) and have some sort of macro that allows people to build up chunks of asm code gradually. But that may be something we can transparently add on top once we've got the basic API right?

But aren't macros resolved at compile time? That was the whole reason why the `dynasmrt` create isn't really that useful to us.

assembler/src/lib.rs
vext01: > Is there anything else that AssemblerFunction could do apart from giving us a pointer?

Eventually it'll have to return a signature of the code it generated, but I think we can assume a fixed sig for now. We can build up to that later.

ltratt:
> But aren't macros resolved at compile time?

Yes, but resolved to *what*? We can choose! If we wanted, it could just translate them to `push` (as string).

assembler/src/lib.rs
ltratt: > Eventually it'll have to return a signature of the code it generated

I can't think of anyway of encoding this statically. I think ultimately we give users some sort of pointer back and they have to cast it (with `unsafe`) to the correct Rust/FFI function type.

ptersilie:
> > But aren't macros resolved at compile time?
> 
> Yes, but resolved to _what_? We can choose! If we wanted, it could just translate them to `push` (as string).

But it means we won't be able to chose registers dynamically at runtime. But I guess for that we still have the default API.

ltratt:
> But it means we won't be able to chose registers dynamically at runtime.

The macro system is Turing complete: you can do anything you want with it. [Though it is quite awkward to use, but that's a separate matter.] Conceptually you can do things like:

```
let x: RegNumber = ...;
asm!(
   mov $x, 0
)
```

and so on.

ptersilie:
> ```
> let x: RegNumber = ...;
> asm!(
>    mov $x, 0
> )
> ```

What if `RegNumber` depends on some other input, e.g. a trace which can't be resolved at compile time?

ltratt:
The code I've pasted is 100% dynamic: you could say `let x = std::env::args()`.

vext01:
I don't see any need for a macro to build instructions.

Is there too much stringly-typed programming going on here?

E.g.
```
const RAX = "rax";
const RBX = "rbx";

asm.push(format!("mov {}, {}", RAX, RBX));
```

To encode that using your own backend (not rasm2), you are going to need an asm parser. We don't need one.

I'd envisaged something like:
```
enum Operand {
    Reg(Reg),
    Indirect(Reg),
    Immediate(Imm),
}

enum Reg {
    RAX,
    RBX,
    ....
}

asm.mov(Operand::Reg(Reg::RAX), Operand::Reg(Reg::RBX));
```

And the backend does whatever it needs to do. In the case of rasm2 it builds a string. In the case of our own backend, you build binary code.

Thoughts?

ltratt:
@vext01 I think this is a big issue and is where we need to understand what other libraries do and probably what other VMs use internally (hence why I suggested dynasm as one point of comparison). There are clear advantages and disadvantages to both approaches (the API-only approach is borderline unreadable, but is probably fast). Maybe we should even support both (API and stringly). Although I think it's difficult (possibly impossible) in Rust, we might even be able to convert stringly-things to API-things at compile-time.

bjorn3:
> > > But aren't macros resolved at compile time?
> > 
> > 
> > Yes, but resolved to _what_? We can choose! If we wanted, it could just translate them to `push` (as string).
> 
> But it means we won't be able to chose registers dynamically at runtime. But I guess for that we still have the default API.

See https://github.com/bytecodealliance/wasmtime/blob/1c55eb1a8bba9e59df604a0df2b6552dd54e76a7/crates/lightbeam/src/backend.rs#L754-L755 for an example how to chose the register at runtime.

ptersilie:
> The code I've pasted is 100% dynamic: you could say `let x = std::env::args()`.

Then why aren't we using this to assemble our trace?

vext01:
> See ... for an example how to chose the register at runtime.

Thanks @bjorn3. That save us a *lot* of time.

If at any point (after COVID-19 has gone away) you find yourself in London, please come and visit us! We'd love to meet you and we surely owe you a coffee and/or a beer by now :)

(We are based at King's College London. My contact details can be found in my GitHub profile)

bjorn3:
Thanks for the invitation! You're welcome!

PR #52, opened 2020-03-20
Relax the dependency on rmp-serde.

vext01:
This is required for an upstream sync.

### DIFF
--- a/ykpack/Cargo.toml
+++ b/ykpack/Cargo.toml
@@ -7,4 +7,4 @@ edition = "2018"
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
 fallible-iterator = "0.2"
-rmp-serde = "=0.14.0"
+rmp-serde = "0.14"


ykpack/Cargo.toml
```diff
@@ -7,4 +7,4 @@ edition = "2018"
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
 fallible-iterator = "0.2"
-rmp-serde = "=0.14.0"
+rmp-serde = "0.14.0"
```
ptersilie: Why not make this "0.14" so we get the non-breaking bug fixes?

vext01: In the eyes of `cargo` both are the same, but I take your point: it's more obvious if we drop the `.0`.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

vext01:
Fixed.

ykpack/Cargo.toml
ptersilie: Oh you're right. But then this means that doing `0.14` will update us to `0.15` and so on, so it actually may not be what we want.

ptersilie: Nevermind, they are equivalent. I was confused. Move along.

vext01:
Can I squash?

ptersilie:
Please do.

vext01:
squashed.

PR #54, opened 2020-03-23
Initiate work on trace compiler

ptersilie:
This PR updates `HWTMapper` to address changes in the new SIR and starts work on the trace compiler. Requires a 3-way merge.

### DIFF
--- a/ykcompile/Cargo.toml
+++ b/ykcompile/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "ykcompile"
+version = "0.1.0"
+authors = ["Lukas Diekmann <lukas.diekmann@gmail.com>"]
+edition = "2018"
+license = "Apache-2.0 OR MIT"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+yktrace = { path = "../yktrace" }
+ykpack = { path = "../ykpack" }
+assembler = "0.10.1"
+dynasmrt = "0.6"
+dynasm = "0.6"
+
+[profile.dev]
+panic = "abort"
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -1,0 +1,178 @@
+#![feature(proc_macro_hygiene)]
+#![feature(test)]
+
+#[macro_use]
+extern crate dynasm;
+extern crate dynasmrt;
+extern crate test;
+
+use std::collections::HashMap;
+use std::mem;
+
+use yktrace::tir::{
+    Constant, ConstantInt, Operand, Rvalue, Statement, TirOp, TirTrace, UnsignedInt,
+};
+
+use dynasmrt::DynasmApi;
+
+/// A compiled SIRTrace.
+pub struct CompiledTrace {
+    /// A compiled trace.
+    mc: dynasmrt::ExecutableBuffer,
+}
+
+impl CompiledTrace {
+    pub fn execute(&self) -> u64 {
+        // For now a compiled trace always returns whatever has been left in register RAX. We also
+        // assume for now that this will be a `u64`.
+        let func: fn() -> u64 = unsafe { mem::transmute(self.mc.ptr(dynasmrt::AssemblyOffset(0))) };
+        func()
+    }
+}
+
+/// The `TraceCompiler` takes a `SIRTrace` and compiles it to machine code. Returns a `CompiledTrace`.
+pub struct TraceCompiler {
+    asm: dynasmrt::x64::Assembler,
+    /// Contains the list of currently available registers.
+    available_regs: Vec<u8>,
+    /// Maps locals to their assigned registers.
+    assigned_regs: HashMap<u32, u8>,
+}
+
+impl TraceCompiler {
+    fn local_to_reg(&mut self, l: u32) -> u8 {
+        // This is a really dumb register allocator, which runs out of available registers after 7
+        // locals. We can do better than this by using StorageLive/StorageDead from the MIR to free
+        // up registers again, and allocate additional locals on the stack. Though, ultimately we
+        // probably want to implement a proper register allocator, e.g. linear scan.
+        if l == 0 {
+            0
+        } else {
+            let reg = self
+                .available_regs
+                .pop()
+                .expect("Can't allocate more than 7 locals yet!");
+            *self.assigned_regs.entry(l).or_insert(reg)
+        }
+    }
+
+    /// Move constant `c` of type `usize` into local `a`.
+    pub fn mov_local_usize(&mut self, local: u32, cnst: usize) {
+        let reg = self.local_to_reg(local);
+        dynasm!(self.asm
+            ; mov Rq(reg), cnst as i32
+        );
+    }
+
+    /// Move constant `c` of type `u8` into local `a`.
+    pub fn mov_local_u8(&mut self, local: u32, cnst: u8) {
+        let reg = self.local_to_reg(local);
+        dynasm!(self.asm
+            ; mov Rq(reg), cnst as i32
+        );
+    }
+
+    /// Move local `var2` into local `var1`.
+    fn mov_local_local(&mut self, l1: u32, l2: u32) {
+        let lreg = self.local_to_reg(l1);
+        let rreg = self.local_to_reg(l2);
+        dynasm!(self.asm
+            ; mov Rq(lreg), Rq(rreg)
+        );
+    }
+
+    fn nop(&mut self) {
+        dynasm!(self.asm
+            ; nop
+        );
+    }
+
+    fn c_mov_int(&mut self, local: u32, constant: &ConstantInt) {
+        let reg = self.local_to_reg(local);
+        let val = match constant {
+            ConstantInt::UnsignedInt(UnsignedInt::U8(i)) => *i as i64,
+            ConstantInt::UnsignedInt(UnsignedInt::Usize(i)) => *i as i64,
+            e => todo!("SignedInt, etc: {}", e),
+        };
+        dynasm!(self.asm
+            ; mov Rq(reg), QWORD val
+        );
+    }
+
+    fn c_mov_bool(&mut self, local: u32, b: bool) {
+        let reg = self.local_to_reg(local);
+        dynasm!(self.asm
+            ; mov Rq(reg), QWORD b as i64
+        );
+    }
+
+    fn statement(&mut self, stmt: &Statement) {
+        match stmt {
+            Statement::Assign(l, r) => {
+                let local = l.local.0;
+                match r {
+                    Rvalue::Use(Operand::Place(p)) => self.mov_local_local(local, p.local.0),
+                    Rvalue::Use(Operand::Constant(c)) => match c {
+                        Constant::Int(ci) => self.c_mov_int(local, ci),
+                        Constant::Bool(b) => self.c_mov_bool(local, *b),
+                        c => todo!("Not implemented: {}", c),
+                    },
+                    unimpl => todo!("Not implemented: {:?}", unimpl),
+                };
+            }
+            Statement::Return => {}
+            Statement::Nop => {}
+            Statement::Unimplemented(mir_stmt) => todo!("Can't compile: {}", mir_stmt),
+        }
+    }
+
+    fn finish(mut self) -> dynasmrt::ExecutableBuffer {
+        dynasm!(self.asm
+            ; ret
+        );
+        self.asm.finalize().unwrap()
+    }
+
+    pub fn compile(tt: TirTrace) -> CompiledTrace {
+        // Set available registers to R11-R8, RDX, RCX
+        let regs = vec![11, 10, 9, 8, 2, 1];
+        let assembler = dynasmrt::x64::Assembler::new().unwrap();
+        let mut tc = TraceCompiler {
+            asm: assembler,
+            available_regs: regs,
+            assigned_regs: HashMap::new(),
+        };
+        for i in 0..tt.len() {
+            let t = tt.op(i);
+            match t {
+                TirOp::Statement(st) => tc.statement(st),
+                TirOp::Guard(_) => tc.nop(), // FIXME Implement guards.
+            }
+        }
+        CompiledTrace { mc: tc.finish() }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::TraceCompiler;
+    use yktrace::tir::TirTrace;
+    use yktrace::{start_tracing, TracingKind};
+
+    #[inline(never)]
+    fn simple() -> u8 {
+        let x = 13;
+        x
+    }
+
+    #[test]
+    pub(crate) fn test_simple() {
+        let th = start_tracing(Some(TracingKind::HardwareTracing));
+        simple();
+        let sir_trace = th.stop_tracing().unwrap();
+        let tir_trace = TirTrace::new(&*sir_trace).unwrap();
+        let ct = TraceCompiler::compile(tir_trace);
+        assert_eq!(ct.execute(), 13);
+    }
+}
--- a/yktrace/src/hwt/mapper.rs
+++ b/yktrace/src/hwt/mapper.rs
@@ -7,7 +7,7 @@ use std::{borrow, collections::HashMap, env, fs};
 
 pub struct HWTMapper {
     phdr_offset: u64,
-    labels: Option<HashMap<u64, String>>
+    labels: Option<HashMap<u64, (String, u32)>>
 }
 
 impl HWTMapper {
@@ -33,18 +33,15 @@ impl HWTMapper {
         for b in trace.iter_blocks() {
             match b {
                 Ok(block) => {
-                    let addr = block.start_vaddr() - self.phdr_offset;
-                    match labels.get(&addr) {
-                        Some(l) => {
-                            // FIXME Do the splitting at the time we load from DWARF.
-                            let data: Vec<&str> = l.split(':').collect();
-                            debug_assert!(data.len() == 3);
-                            let sym = String::from(data[1]);
-                            let bb_idx = data[2].parse::<u32>().unwrap();
-                            annotrace.push(SirLoc::new(sym, bb_idx))
+                    let start_addr = block.start_vaddr() - self.phdr_offset;
+                    let end_addr = start_addr + block.len();
+                    for (addr, (sym, bb_idx)) in labels {
+                        if *addr >= start_addr && *addr < end_addr {
+                            // found matching label
+                            annotrace.push(SirLoc::new(sym.to_string(), *bb_idx));
+                            break;
                         }
-                        None => {}
-                    };
+                    }
                 }
                 Err(_) => {}
             }
@@ -60,7 +57,7 @@ fn get_phdr_offset() -> u64 {
 }
 
 /// Extracts YK debug labels and their addresses from the executable.
-fn extract_labels() -> Result<HashMap<u64, String>, gimli::Error> {
+fn extract_labels() -> Result<HashMap<u64, (String, u32)>, gimli::Error> {
     // Load executable
     let pathb = env::current_exe().unwrap();
     let file = fs::File::open(&pathb.as_path()).unwrap();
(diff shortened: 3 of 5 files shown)


ykcompile/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "ykcompile"
+version = "0.1.0"
+authors = ["Lukas Diekmann <lukas.diekmann@gmail.com>"]
+edition = "2018"
```
vext01: Please add a license line.

ptersilie: Done.

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+
+use yktrace::tir::{TirTrace, TirOp, Statement, Rvalue, Operand, Constant, ConstantInt,
+UnsignedInt};
+
+use dynasmrt::DynasmApi;
+
+/// A compiled SIRTrace.
+pub struct CompiledTrace {
+    /// A compiled trace.
+    mc: dynasmrt::ExecutableBuffer
```
vext01: Is `mc` machine code?

ptersilie: Yes.

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+        // For now a compiled trace always returns whatever has been left in register RAX. We also
+        // assume for now that this will be a u64.
+        let func: fn() -> u64 = unsafe {
+            mem::transmute(self.mc.ptr(dynasmrt::AssemblyOffset(0)))
+        };
+        func()
+    }
+}
+
+/// The TraceCompiler takes a SIRTrace and compiles it to machine code. Returns a CompiledTrace.
```
vext01: Should we backtick types in comments to be consistent with our other code-bases?

ptersilie: Sure.

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+pub struct TraceCompiler {
+    asm: dynasmrt::x64::Assembler,
+    available_regs: Vec<u8>,
+    assigned_regs: HashMap<u32, u8>
+}
+
+impl TraceCompiler {
+
```
vext01: Have we done a `cargo +nightly fmt`?

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+        let func: fn() -> u64 = unsafe {
+            mem::transmute(self.mc.ptr(dynasmrt::AssemblyOffset(0)))
+        };
+        func()
+    }
+}
+
+/// The TraceCompiler takes a SIRTrace and compiles it to machine code. Returns a CompiledTrace.
+pub struct TraceCompiler {
+    asm: dynasmrt::x64::Assembler,
```
vext01: Should we build our API so that adding other architectures later will be easier?

For example, should we parametrise the type of the `asm` field?

ykcompile/src/lib.rs
ptersilie: Nope. Will do.

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+    available_regs: Vec<u8>,
+    assigned_regs: HashMap<u32, u8>
```
vext01: Is that local index to register number?

We should add comments.

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+            mem::transmute(self.mc.ptr(dynasmrt::AssemblyOffset(0)))
+        };
+        func()
+    }
+}
+
+/// The TraceCompiler takes a SIRTrace and compiles it to machine code. Returns a CompiledTrace.
+pub struct TraceCompiler {
+    asm: dynasmrt::x64::Assembler,
+    available_regs: Vec<u8>,
```
vext01: We could use a bit-field for this to save space?

ykcompile/src/lib.rs
ptersilie: We can do it now or we can do it later. I don't think we'll care about other architectures for a while.

vext01: Yeah. I'm just wondering if we can avoid the tedium of having someone go over a (possibly large) existing code-base later, abstracting all of the x86-specifics with an interface.

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+
+impl TraceCompiler {
+
+    fn local_to_reg(&mut self, l: u32) -> u8 {
+        // This is a really dumb register allocator, which runs out of available registers after 7
+        // locals. We can do better than this by using StorageLive/StorageDead from the MIR to free
+        // up registers again, and allocate additional locals on the stack. Though, ultimately we
+        // probably want to implement a proper register allocator, e.g. graph colouring or linear
+        // scan.
+        match self.assigned_regs.get(&l) {
```
vext01: I wonder if we can make this more concise using `Entry`?

https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html

vext01: Even better, we can just use a `Vec` where the index is the local variable number. That removes the hashmap altogether.

But we can treat that as an optimisation later tbh.

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+                        .pop()
+                        .expect("Can't allocate more than 7 locals yet!");
+                    self.assigned_regs.insert(l, reg);
+                    reg
+                }
+            }
+        }
+    }
+
+    pub fn local_usize(&mut self, a: u32, c: usize) {
```
vext01: Can we add comments so that I know what `c` is here?

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+        let mut tc = TraceCompiler {
+            asm: assembler,
+            available_regs: regs,
+            assigned_regs: HashMap::new()
+        };
+        for i in 0..tt.len()-1 {
+            let t = tt.op(i);
+            match t {
+                TirOp::Statement(st) => tc.statement(st),
+                TirOp::Guard(_) => { println!("Not implemented: Guard") }
```
vext01: Don't let Laurie see that `println!`!

vext01:
I'm going to pause my review for a moment (until those method comments appear).

ykcompile/src/lib.rs
ptersilie: We could. But then we also need a mapper that maps the bits to the correct registers, since currently we are using registers 1, 2, 8, 9, 10, 11.

But all of this is going to change anyways once we have a proper register allocator, so I don't think it's worth optimising things already.

vext01: Very good point. Let's keep it simple.

ykcompile/src/lib.rs
ptersilie: I'll try.

ykcompile/src/lib.rs
ptersilie: We used to just print them so we could still compile stuff without implementing guards. Though I'm not sure if guards still appear in our traces.

ykcompile/src/lib.rs
ptersilie: Done.

ykcompile/src/lib.rs
ptersilie: Done.

ykcompile/src/lib.rs
ptersilie: > Even better, we can just use a `Vec` where the index is the local variable number. That removes the hashmap altogether.
> 
> But we can treat that as an optimisation later tbh.

Not really, since our locals are all over the place i.e. not consecutive. But since all of this is going to get replaced by a proper register allocator some day, there's no point spending time optimising things now.

ykcompile/src/lib.rs
bjorn3: https://github.com/bnjbvr/regalloc.rs is a crate that is currently being made to be a new register allocator for Cranelift. It has both linear-scan and a backtracking support.

ykcompile/src/lib.rs
vext01: ok

ykcompile/src/lib.rs
vext01: We could emit an instruction which aborts execution, or a nop?

The former stops the trace, the latter allows it to continue.

ykcompile/src/lib.rs
vext01: That's definitely worth looking into then! Thanks

ptersilie: Thanks, this certainly will be useful in the future. One thing I am wondering though is, since we only need to allocate linear traces (so there's no real need for liveness analysis) what advantages would a proper register allocater give us?

For now our rough plan was to use `StorageLive/StorageDead` information from the MIR to allocate/deallocate registers.

ykcompile/src/lib.rs
ptersilie: I don't think we want to abort since it's gonna be a while before we implement guards.

ykcompile/src/lib.rs
vext01: OK, that helps.

So I'm now wondering if we can come up with some better names for things.

`a` is a local index, right? So how about `l` or `local`?

The function name itself isn't super-intuitive, but I'm not sure what we'd call it?

`const_into_local`?

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+                .expect("Can't allocate more than 7 locals yet!");
+            *self.assigned_regs.entry(l).or_insert(reg)
+        }
+    }
+
+    /// Move constant `c` of type `usize` into local `a`.
+    pub fn local_usize(&mut self, a: u32, c: usize) {
+        let reg = self.local_to_reg(a);
+        dynasm!(self.asm
+            ; mov Rq(reg), c as i32
```
vext01: Is the cast here safe? `c` is a `u64`.

Should we be using `try_from`?

ykcompile/src/lib.rs
bjorn3: A proper register allocator tries to optimize when to spill what register in case of register pressure.

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    fn finish(mut self) -> dynasmrt::ExecutableBuffer {
+        dynasm!(self.asm
+            ; ret
+        );
+        self.asm.finalize().unwrap()
+    }
+
+    pub fn compile(tt: TirTrace) -> CompiledTrace {
+        let regs = vec![11, 10, 9, 8, 2, 1];
```
vext01: Shalll we comment the mnemonic names of these registers? Is that r11-r8, rbx, rdx?

vext01:
This is looking like a good start! Just minor niggles really.

ykcompile/src/lib.rs
vext01: OK, so let's add a NOP for now. Then at least we can (potentially) see an unimplemented instruction in the code.

ykcompile/src/lib.rs
ptersilie: Fixed in a548b5a

ykcompile/src/lib.rs
ptersilie: Like this? b515975

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+        let mut tc = TraceCompiler {
+            asm: assembler,
+            available_regs: regs,
+            assigned_regs: HashMap::new(),
+        };
+        for i in 0..tt.len() {
+            let t = tt.op(i);
+            match t {
+                TirOp::Statement(st) => tc.statement(st),
+                TirOp::Guard(_) => tc.nop(),
```
bjorn3:
```suggestion
                TirOp::Guard(_) => tc.nop(), // FIXME implement an actual guard.
```

ptersilie: Thanks!

ykcompile/src/lib.rs
vext01: Something like that, yes.

Currently we crash on unimplemented statements. Is that what we want?

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+        match stmt {
+            Statement::Assign(l, r) => {
+                let local = l.local.0;
+                match r {
+                    Rvalue::Use(Operand::Place(p)) => {
+                        self.local_local(local, p.local.0)
+                    },
+                    Rvalue::Use(Operand::Constant(c)) => {
+                        match c {
+                            Constant::Int(ci) => {
```
ltratt: I wonder if you might prefer to call a `c_int` function here, and inside that you would inline `local_u8` and `local_usize`.

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+                                    e => panic!("SignedInt, etc: {}", e)
+                                }
+                            }
+                            Constant::Bool(b) => {
+                                self.local_u8(local, *b as u8)
+                            }
+                            c => panic!("Not implemented: {}", c)
+                        }
+                    }
+                    unimpl => panic!("Not implemented: {:?}", unimpl)
```
ltratt: You might prefer to say `unimplemented!(unimpl)` or `todo!(unimpl)` in cases like this (the Rust docs make a nit-picking separation between `unimplemented` and `todo` that seems a bit pointless to me).

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+                };
+            },
+            Statement::Return => {},
+            Statement::Nop => {},
+            Statement::Unimplemented(mir_stmt) => println!("Can't compile: {}", mir_stmt)
```
ltratt: This should surely cause a `panic`, not try and carry on after a `println`?

ykcompile/src/lib.rs
ltratt: Edd knows me well. But now I've seen it, it must go!

yktrace/src/hwt/mapper.rs
```diff
@@ -33,18 +33,15 @@ impl HWTMapper {
...
-                            // FIXME Do the splitting at the time we load from DWARF.
-                            let data: Vec<&str> = l.split(':').collect();
-                            debug_assert!(data.len() == 3);
-                            let sym = String::from(data[1]);
-                            let bb_idx = data[2].parse::<u32>().unwrap();
-                            annotrace.push(SirLoc::new(sym, bb_idx))
+                    let start_addr = block.start_vaddr() - self.phdr_offset;
+                    let end_addr = start_addr + block.len();
+                    for (addr, (sym, bb_idx)) in labels {
+                        if addr >= &start_addr && addr < &end_addr {
```
ltratt: Is `end_addr` not inclusive? It's also generally more idiomatic to write:

```rust
if *addr >= start_addr && *addr < end_addr
```

though they're functionally equivalent.

yktrace/src/hwt/mapper.rs
```diff
@@ -138,3 +135,11 @@ fn extract_labels() -> Result<HashMap<u64, String>, gimli::Error> {
     }
     Ok(labels)
 }
+
+fn split_symbol(s: &str) -> (String, u32) {
+    let data: Vec<&str> = s.split(':').collect();
+    debug_assert!(data.len() == 3);
+    let sym = String::from(data[1]);
```
ltratt: `data[1].to_owned()` is more idiomatic.

yktrace/src/hwt/mapper.rs
bjorn3: `end_addr` is computed as `start_add + len`, so it is indeed exclusive.

ykcompile/src/lib.rs
ptersilie: For some reason dynasm tells me it expects an `i32` here, though we should be able to pass it an `i64`. I'm looking into why that is.

ykcompile/src/lib.rs
ptersilie: It's gone already. As for unimplemented statements, I think for now we should crash. We can change it later if that means we'll be stuck. I will try and keep the tests small enough that they'll allow us to implement one statement at a time.

yktrace/src/hwt/mapper.rs
ltratt: I stand corrected!

ykcompile/src/lib.rs
ptersilie: Done.

ykcompile/src/lib.rs
ltratt: Works for me!

[For some reason when I looked at the code I didn't see all these extra comments. Ah, GitHub updates :/]

yktrace/src/hwt/mapper.rs
vext01: How about `(start_addr..end_addr).contains(&addr)`?

ykcompile/src/lib.rs
ptersilie: Fixed.

ykcompile/src/lib.rs
ptersilie: Fixed.

ykcompile/src/lib.rs
ptersilie: This was already fixed.

yktrace/src/hwt/mapper.rs
ltratt: Isn't that going to be rather inefficient?

ptersilie: Agreed. That's why I used your suggestion.

vext01: I investigated.

It's a little less efficient because `Range::contains()` uses a trait (`RangeBounds`) under the hood.

But `Range` doesn't actually allocate the range (ala Python 2 `range`). It just stores the lower and upper bounds.

ykcompile/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+        );
+    }
+
+    fn nop(&mut self) {
+        dynasm!(self.asm
+            ; nop
+        );
+    }
+
+    fn c_mov_int(&mut self, local: u32, constant: &ConstantInt) {
```
vext01: I'm still not certain if these function names are descriptive enough.

Here we have `mov_int` to mean move a constant integer to a register.

Suppose we need to move a constant into a memory location. What would we call that?

Do we need to put the operand types in the function name? E.g.

 * `c_mov_const_to_reg`
 * `c_mov_const_to_mem`

maybe we should have a method `store_const_to_local` which calls one of the above based upon where the local currently lives (register or memory).

Thoughts?

ltratt: I wouldn't worry *too* much about this now: as we expand functionality, a better naming scheme might make itself apparent.

vext01: OK. I'm happy to continue taking into account that these function names are not fixed.

ykcompile/src/lib.rs
ptersilie: Fixed in 78cbcdb

yktrace/src/hwt/mapper.rs
ptersilie: Fixed in 18eba3a

ykcompile/src/lib.rs
ptersilie: I agree with both of you. I think be probably should make the function names more descriptive. But I also want to get this PR merged now and think we don't need to worry about this just yet.

ykcompile/src/lib.rs
ptersilie: Since this PR has already been going for a while I suggest we get it merged and leave this for a later PR.

vext01:
Please squash and then add a cycle breaker commit.

vext01:
This is ready for bors now?

ptersilie:
Is the other PR merged? I think bors has finished by Github is still waiting for something: https://github.com/softdevteam/ykrustc/pull/96

vext01:
It's showing as merged here.

ptersilie:
Ah, it wasn't before. I just noticed the two commits have been swapped around accidentally. Let me fix that and then it's ready for merging.

ptersilie:
Nevermind, I just got confused because Github displays the commits in a different order than `git log`. All is well. Let the merge commence!

ptersilie:
bors try

ptersilie:
It works. We can squash and merge tomorrow.

vext01:
Please squash.

ptersilie:
Squashed!

PR #55, opened 2020-03-27
Crash dumper

vext01:
These PR adds a crash dumper, and fixes a couple of other small bits. See individual commits.

Here's an example crash dump:
```
The trace compiler crashed!

Reason: Ran out of registers.

Executable code buffer:
  movabs rcx, 0xd
  mov rax, rcx
  movabs r8, 0
  mov r9, rcx

Register allocation (local -> reg):
  51 ->   8 (r8)
   7 ->   9 (r9)
   1 ->   1 (rcx)

thread 'tests::test_simple' panicked at 'stopped due to trace compilation error', ykcompile/src/lib.rs:222:9
```

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -7,13 +7,54 @@ extern crate dynasmrt;
 extern crate test;
 
 use std::collections::HashMap;
+use std::fmt::{self, Display, Formatter};
 use std::mem;
+use std::process::Command;
 
 use yktrace::tir::{
-    Constant, ConstantInt, Operand, Rvalue, Statement, TirOp, TirTrace, UnsignedInt,
+    Constant, ConstantInt, Guard, Operand, Rvalue, Statement, TirOp, TirTrace, UnsignedInt,
 };
 
 use dynasmrt::DynasmApi;
+
+#[derive(Debug)]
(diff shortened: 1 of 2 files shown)


ykcompile/src/lib.rs
```diff
@@ -137,19 +229,31 @@ impl TraceCompiler {
...
-            let t = tt.op(i);
-            match t {
+            let res = match tt.op(i) {
                 TirOp::Statement(st) => tc.statement(st),
-                TirOp::Guard(_) => tc.nop(), // FIXME Implement guards.
+                TirOp::Guard(g) => tc.guard(g),
+            };
+
+            if let Err(e) = res {
+                err = Some(e);
```
ptersilie: Why don't we call `tc.crash_dump()` here so we don't need to assign the `err` variable.

ptersilie:
Just one comment. Rest looks good.

ykcompile/src/lib.rs
vext01: Fixed.

vext01:
Ready for re-review.

ptersilie:
Looks good. Please squash.

vext01:
splat.

PR #56, opened 2020-03-27
Fix a bug in the register allocator and a few other bits.

vext01:
See individual commits.

This will clash with the other PR, so I'll probably have to resolve some conflicts later.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -89,10 +89,15 @@ impl TraceCompiler {
         if l == 0 {
             Ok(0)
         } else {
-            if let Some(reg) = self.available_regs.pop() {
-                Ok(*self.assigned_regs.entry(l).or_insert(reg))
+            if self.assigned_regs.contains_key(&l) {
+                Ok(self.assigned_regs[&l])
             } else {
-                Err(CompileError::OutOfRegisters)
+                if let Some(reg) = self.available_regs.pop() {
+                    self.assigned_regs.insert(l, reg);
+                    Ok(reg)
+                } else {
+                    Err(CompileError::OutOfRegisters)
+                }
             }
         }
     }
@@ -252,8 +257,7 @@ impl TraceCompiler {
 
 #[cfg(test)]
 mod tests {
-
-    use super::TraceCompiler;
+    use super::{HashMap, TraceCompiler};
     use yktrace::tir::TirTrace;
     use yktrace::{start_tracing, TracingKind};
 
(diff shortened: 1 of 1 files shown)


ykcompile/src/lib.rs
```diff
@@ -36,44 +36,52 @@ pub struct TraceCompiler {
...
+            if self.assigned_regs.contains_key(&l) {
+                self.assigned_regs[&l]
+            } else {
+                let reg = self.available_regs.pop().expect(
+                    "Can't allocate more than 7 locals
+                    yet!",
+                );
+                self.assigned_regs.insert(l, reg);
+                reg
+            }
```
bjorn3:
```suggestion
            let available_regs = &mut self.available_regs;
            *self.assigned_regs.entry(l).or_insert_with(|| {
                available_regs.pop().expect("Can't allocate more than 7 locals yet!");
            })
```

vext01: I tried something similar to this and it whined about mutable borrows inside the `or_insert_with` closure. Let's give it a shot.

vext01: Worked with a minor tweak :)

I think the trick is to do `let available_regs = &mut self.available_regs`.

I had repeated `self.available_regs` inside the closure, which upset the borrow checker.

Thanks!

bjorn3: > I think the trick is to do let available_regs = &mut self.available_regs.

See my code :) A closure currently captures a whole variable (`self`), not just a part of it (`self.available_regs`). There is an accepted RFC to only capture the used part, but the implementation is currently on hold.

bjorn3: https://github.com/rust-lang/rfcs/pull/2229

ykcompile/src/lib.rs
```diff
@@ -45,7 +45,10 @@ impl TraceCompiler {
         // locals. We can do better than this by using StorageLive/StorageDead from the MIR to free
         // up registers again, and allocate additional locals on the stack. Though, ultimately we
         // probably want to implement a proper register allocator, e.g. linear scan.
+
         if l == 0 {
+            // In SIR, `Local` zero is the (implicit) return value, so it makes sense to allocate it
+            // in return register of the underlying X86_64 calling convention.
```
ptersilie: `in the return register` (missing `the`)

ptersilie:
One typo, rest looks good.

vext01:
Fixed typo, but we this will need re-sync after #55

ptersilie:
Looks good, squash and rebase once #55 is done.

vext01:
That was quite involved to rebase!

We can't use the entry API after all, as there's no way to get the `OutOfRegisters` error out of the `or_insert_with` closure.

PR #59, opened 2020-04-02
Compile calls

ptersilie:
This PR adds function calls to `ykcompile`, allowing us to compile function calls, including arguments and return variables.

In order to not overwrite in-use registers during a function call, we currently store all registers on the stack before entering a called block. This is really dumb but gets us going until we add a better register allocator soon.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -12,7 +12,8 @@ use std::mem;
 use std::process::Command;
 
 use yktrace::tir::{
-    Constant, ConstantInt, Guard, Local, Operand, Rvalue, Statement, TirOp, TirTrace, UnsignedInt,
+    CallOperand, Constant, ConstantInt, Guard, Local, Operand, Place, Rvalue, Statement, TirOp,
+    TirTrace, UnsignedInt,
 };
 
 use dynasmrt::DynasmApi;
@@ -78,6 +79,8 @@ pub struct TraceCompiler {
     available_regs: Vec<u8>,
     /// Maps locals to their assigned registers.
     assigned_regs: HashMap<Local, u8>,
+    /// Stores the destination locals to which we copy RAX to after returning from a call.
+    returns: Vec<Option<Place>>,
 }
 
 impl TraceCompiler {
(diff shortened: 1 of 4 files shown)


ykcompile/src/lib.rs
```diff
@@ -78,6 +79,8 @@ pub struct TraceCompiler {
     available_regs: Vec<u8>,
     /// Maps locals to their assigned registers.
     assigned_regs: HashMap<Local, u8>,
+    /// Stores return destinations from calls
```
vext01, line 82: full stop.

ykcompile/src/lib.rs
```diff
@@ -160,6 +180,49 @@ impl TraceCompiler {
...
+        // FIXME find a better way to get rid of the unneeded call to stop_tracing
+        match op {
+            ykpack::CallOperand::Fn(s) => if s.contains("stop_tracing") { return Ok(()) },
+            ykpack::CallOperand::Unknown => {}
+        };
+        // Save all registers to the stack
+        self.store_registers();
+        // Move call arguments into registers.
+        for (i, op) in args.iter().enumerate() {
+            let reg = Local((i + 1) as u32);
```
vext01, line 198: It seems like we are bypassing the register allocator here. Shouldn't we use `local_to_reg()`?

ptersilie: We are not, since `mov_local_local`, etc., use the register allocator, but I agree that the variable `reg` is wrongly named here, and should instead be called `local`.

ykcompile/src/lib.rs
```diff
@@ -160,6 +180,49 @@ impl TraceCompiler {
         Ok(())
     }
 
+    fn c_call(
+        &mut self,
+        op: &CallOperand,
+        args: &Vec<Operand>,
+        dest: &Option<Place>,
+    ) -> Result<(), CompileError> {
+        // FIXME find a better way to get rid of the unneeded call to stop_tracing
```
vext01, line 189: Full-stop, and can we expand the comment with a little more context?

ykcompile/src/lib.rs
```diff
@@ -160,6 +180,49 @@ impl TraceCompiler {
...
+            match op {
+                Operand::Place(p) => self.mov_local_local(reg, p.local)?,
+                Operand::Constant(c) => match c {
+                    Constant::Int(ci) => self.c_mov_int(reg, ci)?,
+                    Constant::Bool(b) => self.c_mov_bool(reg, *b)?,
+                    c => todo!("Not implemented: {}", c),
+                },
+            }
+        }
+        // Remember the return destination
```
vext01, line 208: Check for full-stops on all comments :)

ykcompile/src/lib.rs
vext01: I think I get why this is needed (so we know where to copy rax to), but it'd help to expand the comment.

ykcompile/src/lib.rs
vext01: ah! I see.

Yes, please rename.

vext01:
That's all of my comments.

ptersilie:
Think I've addressed everything.

vext01:
One typo and we are done.

ptersilie:
Fixed.

vext01:
Please squash.

ptersilie:
Squashed. This PR modifies `ykpack` but I hope we can get away without a 3-way merge since we haven't changed `ykrustc`. Let's try it.

vext01:
rust fmt :)

ptersilie:
Fixed.

vext01:
plz squash

ptersilie:
Fixed.

vext01:
```
---- tests::test_simple stdout ----
thread 'tests::test_simple' panicked at 'called `Result::unwrap()` on an `Err` value: Syntax("invalid type: string \"discriminant(_6) = 0\", expected tuple variant Statement::Call")', yktrace/src/tir.rs:36:40
---- tests::test_function_call stdout ----
thread 'tests::test_function_call' panicked at 'Once instance has previously been poisoned', src/libstd/sync/once.rs:395:21
```

ptersilie:
Indeed, even though we haven't touch `ykrustc`, `ykpack` is built into the binary which we need to run this PR, so a 3-way merge is required again.

ptersilie:
Alright, I have done a manual build of ykrustc with the changed dependencies on our ci server. Building this should work now. :crossed_fingers: 

bors try

ptersilie:
That worked! Ready for merging!

ptersilie:
Damn, branch out of date.

ptersilie:
Rebased. Let's go!

PR #61, opened 2020-04-07
Use storage dead statement to free up registers.

ptersilie:


ykcompile/src/lib.rs
```diff
@@ -108,6 +108,12 @@ impl TraceCompiler {
         }
     }
 
+    fn free_register(&mut self, local: &Local) {
+        if let Some(reg) = self.assigned_regs.get(local) {
+            self.available_regs.push(*reg);
```
vext01, line 113: What about the hashmap storing register assignments? Does the entry there need to be freed too?

I guess it depends if a local index can be reused. I don't think they can...

ykcompile/src/lib.rs
```diff
@@ -108,6 +108,12 @@ impl TraceCompiler {
         }
     }
 
+    fn free_register(&mut self, local: &Local) {
+        if let Some(reg) = self.assigned_regs.get(local) {
```
vext01, line 112: Is this guard necessary?

vext01:
Couple of comments.

ykcompile/src/lib.rs
ptersilie: Yes, I noticed that only after I assigned you. Fix is already pushed.

vext01: I'm suggesting that it may not be strictly necessary. Read again.

ptersilie: They can be reused, e.g. by entering a function call (all arguments are assigned to $1..$N). So we should definitely clean up the `assigned_regs` too.

Though now that I think about it, entering a function call messes up `assigned_regs` already, i.e. we would have to reset and reload `assigned_regs` when entering/leaving a call, which we currently don't do. Well, investigating proper register allocation is next on my todo list anyways, so we should fix this properly.

vext01: Agreed. The problem you point out above will go away with a proper allocator.

ykcompile/src/lib.rs
```diff
@@ -344,6 +344,7 @@ impl TraceCompiler {
         }
 
         tc.ret();
+        tc.crash_dump(CompileError::OutOfRegisters);
```
vext01, line 347: did you mean to commit this?

ptersilie: No, sorry.

ptersilie: Fixed.

vext01:
Please squash.

ykcompile/src/lib.rs
```diff
@@ -252,6 +258,8 @@ impl TraceCompiler {
             }
             Statement::Return => self.c_return()?,
             Statement::Call(op, args, dest) => self.c_call(op, args, dest)?,
+            Statement::StorageLive(_) => {},
+            Statement::StorageDead(l) => self.free_register(l),
```
bjorn3, line 262: What about

```rust
_1: u8;

bb0: {
    StorageLive(_1);
    _1 = 42u8;
    Goto bb2;
}

bb1: {
    StorageDead(_1);
    return;
}

bb2: {
    Use(_1);
    Goto ebb1;
}

Comment (vext01) [reply ykcompile/src/lib.rs:262]: As I understand, the input to the trace compiler is a trace, so there are no blocks to speak of. Everything should have been inlined.

The above would be:
```
    StorageLive(_1);
    _1 = 42u8;
    Use(_1);
    Goto ebb1;
    StorageDead(_1);
    return;
```

Is that right @ptersilie ?

bjorn3: /me forgot that this is a trace compiler and not a normal compiler.

vext01: No worries! You got me thinking for a moment there!

ptersilie:
Squashed. Though it requires a 3-way merge so we better merge your PR first.

ptersilie:
@vext01 The ykrustc build is running. Can I rebase this while we wait?

vext01:
yep!

ptersilie:
ykrustc has finished building so we should be able to bors this now. Let's give it a try.

bors try

ptersilie:
Formatting again. Why do I never learn. :sob: 

bors try

ptersilie:
@vext01 I need to squash and then this is ready for merging.

vext01:
please squash.

ptersilie:
Squashed.

PR #63, opened 2020-04-14
Calls to things we don't have SIR for.

vext01:
This PR starts dealing with calls to stuff we don't have SIR for.

See commits for more detail.

Will need a rebuilt ykrustc.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -174,7 +174,7 @@ impl TraceCompiler {
         let val = match constant {
             ConstantInt::UnsignedInt(UnsignedInt::U8(i)) => *i as i64,
             ConstantInt::UnsignedInt(UnsignedInt::Usize(i)) => *i as i64,
-            e => todo!("SignedInt, etc: {}", e),
+            e => return Err(CompileError::Unimplemented(format!("{}", e))),
         };
         dynasm!(self.asm
             ; mov Rq(reg), QWORD val
@@ -218,7 +218,7 @@ impl TraceCompiler {
                 Operand::Constant(c) => match c {
                     Constant::Int(ci) => self.c_mov_int(argidx, ci)?,
                     Constant::Bool(b) => self.c_mov_bool(argidx, *b)?,
-                    c => todo!("Not implemented: {}", c),
+                    c => return Err(CompileError::Unimplemented(format!("{}", c))),
                 },
             }
         }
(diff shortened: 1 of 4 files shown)


ykcompile/src/lib.rs
```diff
@@ -468,4 +473,31 @@ mod tests {
...
+    // A trace which contains a call to something which we don't have SIR for should emit a TIR
+    // call operation.
+    #[test]
+    #[ignore]
+    pub fn call_symbol() {
+        let th = start_tracing(Some(TracingKind::HardwareTracing));
+        let g = core::intrinsics::wrapping_add(10u64, 40u64);
+        let sir_trace = th.stop_tracing().unwrap();
+
+        println!("{}", g);
```
ptersilie, line 487: Some left over prints.

ykpack/src/types.rs
```diff
@@ -175,6 +175,9 @@ pub enum Statement {
     /// Information about which locals are currently live/dead.
     StorageLive(Local),
     StorageDead(Local),
+    /// A (non-inlined) call from a TIR trace to a binary symbol using the system ABI. This does
+    /// not appear in SIR.
```
ptersilie, line 179: `in the SIR`.

yktrace/Cargo.toml
```diff
@@ -11,7 +11,7 @@ elf = "0.0"
 fallible-iterator = "0.2"
 ykpack = { path = "../ykpack" }
 lazy_static = "1.3"
-hwtracer = { git = "https://github.com/softdevteam/hwtracer" }
+hwtracer = { path = "../../hwtracer" }
```
ptersilie, line 14: This shouldn't be here, right?

ykcompile/src/lib.rs
```diff
@@ -477,7 +477,6 @@ mod tests {
     // A trace which contains a call to something which we don't have SIR for should emit a TIR
     // call operation.
     #[test]
-    #[ignore]
```
ptersilie: Does this need to be squashed into the previous commit?

ptersilie:
Just some high-level comments.

ykcompile/src/lib.rs
vext01: That's on purpose. The test belonged in the previous commit, but didn't work because of the mapper bug fixed in this commit.

ykpack/src/types.rs
vext01: I  think `in SIR` is more correct. We are saying that this kind of statement cannot ever appear in any instance of SIR.

yktrace/Cargo.toml
vext01: Oops. That's not needed I don't think. Will check.

vext01:
OK to squash?

ptersilie:
What about the missing "the"? I guess without it is also ok.

vext01:
> What about the missing "the"? I guess without it is also ok.

I must admit, I prefer without.

ptersilie:
Fine with me, but then we should also change these other locations accordingly ;-) :
```
yktrace/src/tir.rs
2://! into a Tracing IR (TIR) Trace using the SIR found in the `.yk_sir` section of the currently                                                                                   
98:/// A TIR trace is conceptually a straight-line path through the SIR with guarded speculation.                                                                                                                                                                
120:            // statements in the SIR) so that we have the freedom to mutate them later.  
```

vext01:
I'm not sure there's a hard and fast rule.

I think there's a distinction to be made:
 * "SIR" -- the language of SIR.
 * "the SIR" -- an instance of, or instances (plural) of SIR.

Changing some of those in your grep output would make it sound a bit weird IMHO.

ptersilie:
Fair enough. Squash away!

vext01:
squashed, will now do a manual build of ykrustc.

ptersilie:
Okay, let me know when to bors this.

vext01:
bors try

vext01:
I need to fmt all the individual commits. Mind if I force push?

ptersilie:
Go ahead!

vext01:
bors try

vext01:
ready.

PR #64, opened 2020-04-22
Bring TIR traces into SSA form.

ptersilie:
This PR turns TIR into SSA form, which means we don't have to store all
registers onto the stack before entering an inline function call
anymore.

We use alpha renaming to bring the TIR into SSA form during translation
from SIR to TIR. The idea is quite simple: whenever we see a
`StorageLive(local)` statement we assign the contained `Local` a new
variable name (`var_count`) which starts at `1` and increases with each
new variable. We store a mapping `oldname => newname` for each variable
in `rename_map`, which allows us to rename references later in the
trace.

This becomes a bit tricky when it comes to inline function calls, which
assume a clean slate, i.e. arguments are named `1..N`. Renaming those
arguments would overwrite any existing mappings, e.g. assigning `5` to
argument `1`, would overwrite the mapping of a variable `1` which exists
outside of the inline function call and make it impossible to rename
references that appear after the function call to their correct new
name. Each mapping is thus a `Vector` to which we push additional
mappings. For the above example the mapping would be `1 => [2, 5]`. We
remove mappings again when they run out of scope, i.e. either when the
inlined function call ends (`Statement::Leave`), or we encounter
`Statement::StorageDead`.

The next thing we need to watch out for is that the `TraceCompier`
assigns arguments to the correct `Local`s (which may have been renamed).
Luckily, it seems that each inline function call assigns a new variable
(to store the return value) just before entering the call. We thus can
simply assign arguments to `returnvalue+1..args.len()` instead of the
previous `1...args.len()`. However, it may turn out that not every
inlined function call assigns a new variable (though I couldn't find one
that doesn't even if the function's return type is empty). In that case
we may need to forward the "argument offset" within `Statement::Enter`.

I've left comments similar to the above within the code, though they may need updating if they don't explain things properly.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -119,22 +119,6 @@ impl TraceCompiler {
         }
     }
 
-    fn store_registers(&mut self) {
-        for reg in &[15, 14, 13, 12, 11, 10, 9, 8, 2, 1] {
-            dynasm!(self.asm
-                ; push Rq(reg)
-            );
-        }
-    }
-
-    fn restore_registers(&mut self) {
-        for reg in &[1, 2, 8, 9, 10, 11, 12, 13, 14, 15] {
-            dynasm!(self.asm
-                ; pop Rq(reg)
-            );
-        }
-    }
-
     /// Move constant `c` of type `usize` into local `a`.
     pub fn mov_local_usize(&mut self, local: Local, cnst: usize) -> Result<(), CompileError> {
         let reg = self.local_to_reg(local)?;
@@ -208,16 +192,15 @@ impl TraceCompiler {
             }
             ykpack::CallOperand::Unknown => {}
         };
-        // Save all registers to the stack.
-        self.store_registers();
         // Move call arguments into registers.
-        for (i, op) in args.iter().enumerate() {
-            let argidx = Local((i + 1) as u32);
+        let vars_off = dest.as_ref().unwrap().local;
+        for (op, i) in args.iter().zip(1..) {
+            let arg_idx = Local(i + vars_off.0);
             match op {
-                Operand::Place(p) => self.mov_local_local(argidx, p.local)?,
+                Operand::Place(p) => self.mov_local_local(arg_idx, p.local)?,
                 Operand::Constant(c) => match c {
-                    Constant::Int(ci) => self.c_mov_int(argidx, ci)?,
-                    Constant::Bool(b) => self.c_mov_bool(argidx, *b)?,
+                    Constant::Int(ci) => self.c_mov_int(arg_idx, ci)?,
+                    Constant::Bool(b) => self.c_mov_bool(arg_idx, *b)?,
                     c => return Err(CompileError::Unimplemented(format!("{}", c))),
                 },
             }
@@ -228,7 +211,6 @@ impl TraceCompiler {
     }
 
     fn c_leave(&mut self) -> Result<(), CompileError> {
-        self.restore_registers();
         let dest = self.leaves.pop();
         if let Some(d) = dest {
             if let Some(d) = d {
--- a/yktrace/src/tir.rs
+++ b/yktrace/src/tir.rs
@@ -8,7 +8,7 @@ use elf;
 use fallible_iterator::FallibleIterator;
 use std::{
     collections::HashMap,
-    convert::TryFrom,
+    convert::{From, TryFrom},
     env,
     fmt::{self, Display},
     io::Cursor
@@ -108,6 +108,9 @@ impl TirTrace {
     pub fn new<'s>(trace: &'s dyn SirTrace) -> Result<Self, InvalidTraceError> {
         let mut ops = Vec::new();
         let mut itr = trace.into_iter().peekable();
+        let mut rename_map: HashMap<Local, Vec<Local>> = HashMap::new();
+        let mut var_len = 1;
+        let mut cur_call_args: Vec<u32> = Vec::new();
         while let Some(loc) = itr.next() {
             let body = match SIR.bodies.get(&loc.symbol_name) {
                 Some(b) => b,
@@ -119,13 +122,51 @@ impl TirTrace {
             // When adding statements to the trace, we clone them (rather than referencing the
             // statements in the SIR) so that we have the freedom to mutate them later.
             let user_bb_idx_usize = usize::try_from(loc.bb_idx).unwrap();
-            ops.extend(
-                body.blocks[user_bb_idx_usize]
-                    .stmts
-                    .iter()
-                    .cloned()
-                    .map(TirOp::Statement)
-            );
+            // When converting the SIR trace into a TIR trace we turn it into SSA form by alpha
+            // renaming all `Local`s and storing a mapping from old to new name in `rename_map`.
+            // This is a bit tricky when it comes to inlined function calls, which assume arguments
+            // start at `Local(1)`, which may have already been used. We thus need to store
+            // multiple levels in `rename_map` and have an additional variable `cur_call_args`,
+            // which provides information about which `Local`s we need to remove again from
+            // `rename_map` once we leave the inlined function call again (since we won't get
+            // `StorageDead` statements for arguments).
+            for stmt in body.blocks[user_bb_idx_usize].stmts.iter() {
+                let op = match stmt {
+                    Statement::StorageLive(local) => {
+                        let newlocal = Local(var_len);
+                        match rename_map.get_mut(local) {
+                            Some(v) => v.push(newlocal),
+                            None => {
+                                rename_map.insert(local.clone(), vec![newlocal]);
+                            }
+                        };
+                        var_len += 1;
+                        Statement::StorageLive(newlocal)
+                    }
+                    Statement::StorageDead(local) => {
+                        if let Some(v) = rename_map.get_mut(local) {
+                            let l = v.pop();
+                            Statement::StorageDead(l.unwrap())
+                        } else {
+                            stmt.clone()
+                        }
+                    }
+                    Statement::Assign(place, rvalue) => {
+                        let newplace = TirTrace::rename_place(&rename_map, &place);
+                        let newrvalue = TirTrace::rename_rvalue(&rename_map, &rvalue);
+                        Statement::Assign(newplace, newrvalue)
+                    }
+                    Statement::Call(_, _, _) => {
+                        // Statement::Call is a terminator and thus should never appear here.
+                        unreachable!();
+                    }
+                    Statement::Nop => stmt.clone(),
+                    Statement::Enter(_, _, _) => unreachable!(),
+                    Statement::Leave => stmt.clone(),
+                    Statement::Unimplemented(_) => stmt.clone()
+                };
+                ops.push(TirOp::Statement(op));
+            }
 
             match &body.blocks[user_bb_idx_usize].term {
                 Terminator::Call {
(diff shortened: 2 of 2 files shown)


yktrace/src/tir.rs
```diff
@@ -119,13 +123,50 @@ impl TirTrace {
...
-                    .map(TirOp::Statement)
-            );
+            // When converting the SIR trace into a TIR trace we turn it into SSA form by alpha
+            // renaming all `Local`s and storing a mapping from old to new name in `rename_map`.
+            // This is a bit tricky when it comes to inlined function calls, which assume arguments
+            // start at `Local(1)`, which may have already been used. We thus need to store
+            // multiple levels in `rename_map` and have an additional variable `cur_call_args`,
+            // which provides information about which `Local`s we need to remove again from
+            // `rename_map` once we leave the inlined function call again (since we won't get
+            // `SotrageDead` statements for arguments).
```
bjorn3, line 133:
```suggestion
            // `StorageDead` statements for arguments).
```

bjorn3:
I think this won't handle the following correctly:

```
_1: (u32, u32);
_1.0 = 1;
_1.1 = 2;
```

If I skimmed it correctly, this will result in:

```
_1: (u32, u32);
_1.0 = 1;
_2: (u32, u32);
_2.1 = 2;
```

Notice how `_2.0` is uninitialized instead of `1`.

ptersilie:
Good spot. When renaming a `Place` I forgot to reapply the `projection`, which means
```
let x = (1, 2);
x.0 = 3;
x.1 = 4;
```
would result in
```
$1 = U8(1)
$1 = U8(2)
$1 = U8(3)
$1 = U8(4)
```
After fixing this (see 9308ff1) we now get:
```
  ($1).(unimplemented projection: "Field(field[0], u8)") = U8(1)                                                                                                                     
  ($1).(unimplemented projection: "Field(field[1], u8)") = U8(2)                                                                                                                     
  ($1).(unimplemented projection: "Field(field[0], u8)") = U8(3)                                                                                                                     
  ($1).(unimplemented projection: "Field(field[1], u8)") = U8(4)
```

ltratt:
I'm fine with this for now:

> This becomes a bit tricky when it comes to inline function calls, which
assume a clean slate, i.e. arguments are named 1..N. Renaming those
arguments would overwrite any existing mappings, e.g. assigning 5 to
argument 1, would overwrite the mapping of a variable 1 which exists
outside of the inline function call and make it impossible to rename
references that appear after the function call to their correct new
name. Each mapping is thus a Vector to which we push additional
mappings. For the above example the mapping would be 1 => [2, 5].

but I don't think this is a viable long-term solution as the malloc of a `Vec` is probably going to be quite a punishing cost. I also think we should be able to find a proxy that removes this need. In essence, when we recurse into an inlined function can we give it an `offset: usize` variable that says "alpha rename variable `x` to `offset + x`"? Then we can return from that function a `usize` that says how many variables have been used and update accordingly. I *think* that gives the same effect but I might be wrong!

[Also, starting the count from 1 seems a bit odd. Should we start from 0? It's not a big deal though!]

ykcompile/src/lib.rs
```diff
@@ -208,11 +192,10 @@ impl TraceCompiler {
             }
             ykpack::CallOperand::Unknown => {}
         };
-        // Save all registers to the stack.
-        self.store_registers();
         // Move call arguments into registers.
+        let local_offset = dest.as_ref().unwrap().local;
```
ltratt, line 196: Sometimes we use "offset", sometimes "idx" as a suffix. Should we be consistent?

ykcompile/src/lib.rs
```diff
@@ -208,11 +192,10 @@ impl TraceCompiler {
...
         for (i, op) in args.iter().enumerate() {
-            let argidx = Local((i + 1) as u32);
+            let argidx = Local(i as u32 + local_offset.0 + 1);
```
ltratt, line 198: In general `as` scares me, as it can do some surprising things. It might not be possible, but perhaps there's a way of shuffling the code that means we can consistently use one integer type?

yktrace/src/tir.rs
```diff
@@ -108,6 +109,9 @@ impl TirTrace {
     pub fn new<'s>(trace: &'s dyn SirTrace) -> Result<Self, InvalidTraceError> {
         let mut ops = Vec::new();
         let mut itr = trace.into_iter().peekable();
+        let mut rename_map: HashMap<Local, Vec<Local>> = HashMap::new();
+        let mut var_count = 1;
```
ltratt, line 113: And now "count" as a suffix :)

yktrace/src/tir.rs
```diff
@@ -119,13 +123,50 @@ impl TirTrace {
...
+            // start at `Local(1)`, which may have already been used. We thus need to store
+            // multiple levels in `rename_map` and have an additional variable `cur_call_args`,
+            // which provides information about which `Local`s we need to remove again from
+            // `rename_map` once we leave the inlined function call again (since we won't get
+            // `StorageDead` statements for arguments).
+            for stmt in body.blocks[user_bb_idx_usize].stmts.iter() {
+                let op = match stmt {
+                    Statement::StorageLive(local) => {
+                        let newlocal = Local(var_count);
+                        if let Some(v) = rename_map.get_mut(local) {
```
ltratt, line 138: It's a minor matter but this might be shorter as:

```rust
match rename_map.get_mut(local) {
    Some(v) => v.push(newlocal),
    None => rename_map.insert(local.clone(), vec![newlocal])
}
```

yktrace/src/tir.rs
```diff
@@ -138,11 +179,29 @@ impl TirTrace {
...
+                                let newlocal = Local(var_count);
+                                if let Some(v) = rename_map.get_mut(&oldlocal) {
+                                    v.push(newlocal);
+                                } else {
+                                    rename_map.insert(oldlocal, vec![newlocal]);
+                                }
+                                var_count += 1;
+                            }
+                            cur_call_args.push(newargs.len() as u32);
+                            // FIXME It seems that calls always have a destination despite it being
```
ltratt, line 200: Should we add an `assert` along the lines of `assert!(call.destination.is_some())` to make sure this assumption can't be violated?

yktrace/src/tir.rs
```diff
@@ -216,6 +286,67 @@ impl TirTrace {
...
+                        println!("warning: mapping for {} already empty", &place.local);
+                        place.clone()
+                    }
+                }
+            }
+            None => {
+                if &place.local != &Local(0) {
+                    // Local(0) is used for function returns and is thus never
+                    // renamed, so we don't need to print a warning for it.
+                    println!("warning: could not find mapping for {}", &place.local);
```
ltratt, line 343: I am generally opposed to stdout (or stderr) stuff like this: either this is a condition we should crash over (assert/panic/whatever) to warn us to fix it; or it's not a problem and execution can continue. `print`s are a horrible halfway house :)

ptersilie:
> but I don't think this is a viable long-term solution as the malloc of a `Vec` is probably going to be quite a punishing cost. I also think we should be able to find a proxy that removes this need. In essence, when we recurse into an inlined function can we give it an `offset: usize` variable that says "alpha rename variable `x` to `offset + x`"? Then we can return from that function a `usize` that says how many variables have been used and update accordingly. I _think_ that gives the same effect but I might be wrong!

I added the arguments to the map so that I can also rename references to the arguments. But you are right, that I may not need this, though I'm not sure where to store the offset especially with nested functions. I will have to think about this for a bit.

> [Also, starting the count from 1 seems a bit odd. Should we start from 0? It's not a big deal though!]

That's because `$0` is reserved for returns. The SIR also starts at `$1` so I was just doing the same.

yktrace/src/tir.rs
bjorn3: Maybe use the `log` crate? You can then use `log::warn!("could not find mapping for {}", &place.local);`. To show them you need to call `env_logger::setup();` from the `env_logger` crate **and** set the `RUST_LOG` env var to one of `warn`, `info`, `debug` or `trace`.

yktrace/src/tir.rs
ptersilie: Good idea.

ltratt:
> I added the arguments to the map so that I can also rename references to the arguments. But you are right, that I may not need this, though I'm not sure where to store the offset especially with nested functions. I will have to think about this for a bit.

I don't think you should fix that for this PR, but for the next PR.

> That's because $0 is reserved for returns. The SIR also starts at $1 so I was just doing the same.

Oh -- makes sense!

yktrace/src/tir.rs
ltratt: In this case, it seems to me that if we can't find a mapping, we've somehow got a serious internal consistency error which we need to fix? That doesn't feel like a warning to me, but I'm prepared to be convinced otherwise.

ptersilie: The problem is that if we crash our tests won't work anymore and we can't continue working on the compiler. This is something we need to fix in the future (i.e. clean up the trace from the remnants of the start/stop tracing calls), so I thought a warning will make sure we don't forget. I'm also ok with putting a `FIXME` here, or used `log` as @bjorn3 suggested.

ltratt: How many tests fail? Should we mark them as temporarily failing?

ptersilie: All of them. Basically our current TIR traces have stuff in them we don't want/need, but getting rid of this is something I will need to look into with @vext01 at some point the future.

ltratt: OK, fair enough. Let's do `eprintln` at least then.

ykcompile/src/lib.rs
ptersilie: The problem is that `Local` takes `u32`s but enumerate returns `usize`. I think I may be able to use `zip` to enumerate while using `u32` as an iterator.

ptersilie: Fixed in 5dc9bb8

ykcompile/src/lib.rs
ptersilie: The things is these all describe different things:
- `Local offset` is offset for the arguments.
- `var_count` tells us how many variables we've assigned so far so we know what the next variables name is.

I feel like calling them all `idx` isn't very accurate either.

yktrace/src/tir.rs
ptersilie: Fixed in a62443d

yktrace/src/tir.rs
ptersilie: Fixed in 24eabc1

ykcompile/src/lib.rs
ltratt: Fair enough. I do feel like we can probably be more consistent than we currently are though. For example, `vars_len` might work better (and it's shorter than `count`) and `args_off` might work. I'm not saying those are the right names, but just lobbing a few ideas into the pot.

yktrace/src/tir.rs
ptersilie: Fixed in 4ed196a

ykcompile/src/lib.rs
ptersilie: Fixed in dae83bd (I didn't come up with any better names though).

ptersilie:
I think I have addressed all comments. Ready for re-review.

ptersilie:
Yes, it is. I guess you could call them either, I'm happy either way.

ltratt:
Let's do `args_len` then, as it's less easily confused IMHO.

ptersilie:
Actually, I was wrong. `arg_off` just describes the offset which we need to add to each argument so it has nothing to do with the number of arguments passed into the function.

ltratt:
So should it be `vars_off` or similar?

ptersilie:
Yes, that might describe it better.

ptersilie:
Fixed in 8654053.

ltratt:
Please squash.

ptersilie:
Squashed. Now I just need to manually build `rustc`. I'll ping you when that's done.

ptersilie:
bors try

ptersilie:
bors try

ptersilie:
bors try

ptersilie:
bors try

ptersilie:
Finally! This needs another squash and then this is ready for merging.

ltratt:
Please squash.

ptersilie:
Squashed!

PR #65, opened 2020-05-13
Avoid using hashmaps for SSA conversion

ptersilie:
This PR updates the current SSA renaming approach to use a single vector instead of hashmaps of vectors. Before doing this we first remove any remnants of `start_tracing` and `stop_tracing` from the headers (772c560). We then move the existing SSA renaming code into its own struct (5ab2645) to make our life a bit easier. Finally, we update the SSA algorithm to use a vector instead of a hashamp (see commit message for more details).

This PR depends on https://github.com/softdevteam/ykrustc/pull/103.

### DIFF
--- a/yktrace/src/tir.rs
+++ b/yktrace/src/tir.rs
@@ -285,6 +285,39 @@ impl TirTrace {
             }
         }
 
+        // Remove the remnants of the `start_tracing` (1 instruction) and `stop_tracing` (5
+        // instructions) calls at the beginning and end of the trace.  To be sure we don't remove
+        // something important, let's put some asserts in to check these instructions are always
+        // roughly the same.
+        match ops.pop() {
+            Some(TirOp::Statement(Statement::Enter(CallOperand::Fn(s),_,_))) => {
+                debug_assert!(s.contains("stop_tracing"))
+            }
+            e => panic!("Expected call to `stop_tracing` here, instead got {:?}.", e)
+        }
+        match ops.pop() {
+            Some(TirOp::Statement(Statement::Assign(_, _))) => {
+            }
+            e => panic!("Expected `Assign` here, instead got {:?}.", e)
+        }
+        match ops.pop() {
+            Some(TirOp::Statement(Statement::Assign(_, Rvalue::Use(Operand::Constant(Constant::Bool(false)))))) => {
+            }
+            e => panic!("Expected `Assign(_, false)` here, instead got {:?}.", e)
+        }
+        for _ in 0..3 {
+            match ops.pop() {
+                Some(TirOp::Statement(Statement::StorageLive(_))) => {
+                }
+                e => panic!("Expected `StorageLive` here, instead got {:?}.", e)
+            }
+        }
+        match ops.remove(0) {
+            TirOp::Statement(Statement::StorageDead(_)) => {
+            }
+            e => panic!("Expected `StorageDead` here, instead got {:?}.", e)
+        }
+
         Ok(Self { ops })
     }
 
@@ -315,10 +348,13 @@ impl TirTrace {
     }
 
     fn rename_place(rename_map: &HashMap<Local, Vec<Local>>, place: &Place) -> Place {
-        // In the future there should always be a mapping for any local in the trace.
-        // Unfortunately, since we are still getting remnants of the trace header and tail in our
-        // trace, this is not always the case. So for now print warnings and don't attempt to
-        // rename those locals.
+        // Local(0) is always used for returning values from calls, so they don't need to be
+        // renamed.
+        if &place.local == &Local(0) {
+            return place.clone();
+        }
+        // While the TIR trace still contains remnants of the trace header and tail, they will be
+        // removed later, so we can savely ignore them here.
         match rename_map.get(&place.local) {
             Some(v) => match v.last() {
                 Some(l) => {
@@ -327,16 +363,10 @@ impl TirTrace {
                     p
                 }
                 None => {
-                    eprintln!("warning: mapping for {} already empty", &place.local);
                     place.clone()
                 }
             },
             None => {
-                if &place.local != &Local(0) {
-                    // Local(0) is used for function returns and is thus never
-                    // renamed, so we don't need to print a warning for it.
-                    eprintln!("warning: could not find mapping for {}", &place.local);
-                }
                 place.clone()
             }
         }
--- a/yktrace/src/tir.rs
+++ b/yktrace/src/tir.rs
@@ -8,7 +8,7 @@ use elf;
 use fallible_iterator::FallibleIterator;
 use std::{
     collections::HashMap,
-    convert::{From, TryFrom},
+    convert::TryFrom,
     env,
     fmt::{self, Display},
     io::Cursor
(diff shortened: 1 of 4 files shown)


yktrace/src/tir.rs
```diff
@@ -285,74 +258,143 @@ impl TirTrace {
             }
         }
 
+        // Remove the remnants of the `start_tracing` (1 instruction) and `stop_tracing` (5
+        // instructions) calls at the beginning and end of the trace.  To be sure we don't remove
+        // something important, let's put some asserts in to check these instructions are always
+        // roughly the same.
+        match ops.pop() {
+            Some(TirOp::Statement(Statement::Enter(CallOperand::Fn(s),_,_,_))) => {
+                assert!(s.contains("stop_tracing"))
```
vext01, line 267: Did you think about if these should be ,`debug_assert!`?

yktrace/src/tir.rs
```diff
@@ -285,74 +258,143 @@ impl TirTrace {
...
+    }
+
+    /// Return the length of the trace measure in operations.
+    pub fn len(&self) -> usize {
+        self.ops.len()
+    }
+}
+
+struct SSA {
+    /// Largest original variable name at the time of entering a call.
```
vext01, line 311: Could we make this a little more precise? It's a stack of variable indices of some kind I think?

yktrace/src/tir.rs
```diff
@@ -285,74 +258,143 @@ impl TirTrace {
...
+        unsafe { &self.ops.get_unchecked(idx) }
+    }
+
+    /// Return the length of the trace measure in operations.
+    pub fn len(&self) -> usize {
+        self.ops.len()
+    }
+}
+
+struct SSA {
```
vext01, line 310: Does this need to be revised in light of our IR not quite being SSA? This only does renamings, but doesn't make new inflections of variable upon mutations I think?

yktrace/src/tir.rs
```diff
@@ -285,74 +258,143 @@ impl TirTrace {
...
+    }
+
+    fn offset(&self) -> u32 {
+        self.offset
+    }
+
+    fn enter(&mut self, num_locals: usize) {
+        // When entering an inlined function call we push the current offset onto the stack, so we
+        // can return to this offset after leaving the function. Afterwards we increase the offset
+        // by the number of variables in the outer context.
```
vext01, line 332: The order of the operations is reversed in the comment. We increase the offset first.

yktrace/src/tir.rs
```diff
@@ -285,74 +258,143 @@ impl TirTrace {
...
+        if let Some(v) = self.stack.last() {
+            self.offset = *v;
+        } else {
+            panic!("Unbalanced enter/leave statements!")
+        }
+    }
+
+    fn rename_args(&mut self, args: &Vec<Operand>) -> Vec<Operand> {
+        let mut newargs = Vec::new();
+        for (i, op) in args.iter().enumerate() {
```
vext01, line 350: We might be able to make it slightly more concise with `.map()`? (but I wouldn't worry too much).

yktrace/src/tir.rs
```diff
@@ -285,74 +258,143 @@ impl TirTrace {
...
-                if &place.local != &Local(0) {
-                    // Local(0) is used for function returns and is thus never
-                    // renamed, so we don't need to print a warning for it.
-                    eprintln!("warning: could not find mapping for {}", &place.local);
-                }
-                place.clone()
-            }
+    fn rename_place(&self, place: &Place) -> Place {
+        if &place.local == &Local(0) {
+            // Local(0) is used for returning values from calls, so it doesn't need to be renamed.
```
vext01, line 385: This feels fishy.

Suppose we have:
```
fn f() {
  $0 = 5;
  $1 = g(); // $1 unused
}

fn g() {
  $0 = 6;
}
```

Wouldn't `g`'s $0 obliterate `f`'s?

vext01:
This is looking good. Most comments are minor, but we should think carefully about $0.

yktrace/src/tir.rs
ptersilie: These can probably be `debug_assert"s, yes.

yktrace/src/tir.rs
ptersilie: I'm open to suggestions. I don't have any.

yktrace/src/tir.rs
ptersilie: This used to do more stuff, but has been simplified since then, so I agree this can be a map now.

yktrace/src/tir.rs
ptersilie: To the best of my knowledge, `$0` is only ever used to return variables from a function so it's always the last instruction of a block. A function `f` as above would thus not be possible.

bjorn3: 1. For MIR after deaggregation there can be multiple statements each assigning to a different part of `_0`.
2. Also for MIR there is https://github.com/rust-lang/rust/pull/72205 that will hopefully be merged soon(ish).

ptersilie: >     1. For MIR after deaggregation there can be multiple statements each assigning to a different part of `_0`.

This should be fine as long as the `_0` is the return variable and not used for anything else.

>     2. Also for MIR there is [rust-lang/rust#72205](https://github.com/rust-lang/rust/pull/72205) that will hopefully be merged soon(ish).

Does this mean that as of now `_0` can be used as a normal variable, but this PR is going to fix that?

yktrace/src/tir.rs
vext01: `VarRenamer`?

yktrace/src/tir.rs
vext01: >[$0 is] always the last instruction of a block.

I'm not sure if that's true or not.

bjorn3: > Does this mean that as of now _0 can be used as a normal variable, but this PR is going to fix that?

Rather the opposite. Before the PR the return value was copied to `_0` just before returning. After the PR the return value may be written to `_0` and then be read from within the same function at a later point.

```rust
bb0: {
    _1 = const 0u8;
    _2 = Add(_1, const 0u8);
    SwitchInt(_2) -> [0: bb1, otherwise: bb2];
}
bb1: {
    _0 = _1;
    return;
}
bb2: {
    goto -> bb2;
}
```

might turn into

```rust
bb0: {
    _0 = const 0u8;
    _2 = Add(_0, const 0u8);
    SwitchInt(_2) -> [0: bb1, otherwise: bb2];
}
bb1: {
    return;
}
bb2: {
    goto -> bb2;
}
```

if I understand it correctly.

ptersilie: Ah, thanks for the clarification. That's good to know as this will mess with inlined functions. But I think I'll do that in a separate PR and in the meantime add an `assert` to check that `_0` is not overwritten by an inlined function.

yktrace/src/tir.rs
ptersilie: Fixed in 51ea6c0

yktrace/src/tir.rs
ptersilie: Fixed in 364952d

yktrace/src/tir.rs
ptersilie: Fixed in 48ca04d

yktrace/src/tir.rs
ptersilie: Fixed in b7e6880

yktrace/src/tir.rs
ptersilie: Fixed in 0f6e990

vext01:
Changes so far look good.

ptersilie:
Okay to squash? I will fix the `$0` problem in a separate PR.

vext01:
please squash

ptersilie:
Squashed and rebased.

ptersilie:
I now have to manually build https://github.com/softdevteam/ykrustc/pull/103 on bors, then we can merge this PR and afterward we can merge https://github.com/softdevteam/ykrustc/pull/103.

vext01:
OK, ping when ready.

vext01:
@ptersilie so this is ready to merge I think?

ptersilie:
There's another change you need to approve in https://github.com/softdevteam/ykrustc/pull/103 before I can build rustc on bors so we can merge this.

ptersilie:
This should work now.

bors try

ptersilie:
Let's try again.

bors try

ptersilie:
Ok to squash?

vext01:
OK

ptersilie:
Squashed.

PR #66, opened 2020-05-22
Codegen calls.

vext01:
This makes a start on code-genning calls. 

See individual commit messages.

The second commit is unrelated and should stay separate.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -20,6 +20,8 @@ use yktrace::tir::{
 
 use dynasmrt::DynasmApi;
 
+const X86_86_NUM_REGS: u8 = 16;
+
 #[derive(Debug, Hash, Eq, PartialEq)]
 pub enum CompileError {
     /// We ran out of registers.
@@ -28,6 +30,8 @@ pub enum CompileError {
     /// Compiling this statement is not yet implemented.
     /// The string inside is a hint as to what kind of statement needs to be implemented.
     Unimplemented(String),
+    /// The binary symbol could not be found.
+    UnknownSymbol(String),
 }
 
 impl Display for CompileError {
@@ -35,6 +39,7 @@ impl Display for CompileError {
         match self {
             Self::OutOfRegisters => write!(f, "Ran out of registers"),
             Self::Unimplemented(s) => write!(f, "Unimplemented compilation: {}", s),
+            Self::UnknownSymbol(s) => write!(f, "Unknown symbol: {}", s),
         }
     }
 }
@@ -73,7 +78,14 @@ impl CompiledTrace {
         // For now a compiled trace always returns whatever has been left in register RAX. We also
         // assume for now that this will be a `u64`.
         let func: fn() -> u64 = unsafe { mem::transmute(self.mc.ptr(dynasmrt::AssemblyOffset(0))) };
-        func()
+        self.exec_trace(func)
+    }
+
+    /// Jump to JITted code. This is a separate unmangled function to make it easy to set a
+    /// debugger breakpoint right before entering the trace.
+    #[no_mangle]
+    fn exec_trace(&self, t_fn: fn() -> u64) -> u64 {
+        t_fn()
     }
 }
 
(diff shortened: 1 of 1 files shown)


ykcompile/src/lib.rs
```diff
@@ -73,7 +78,14 @@ impl CompiledTrace {
...
         // assume for now that this will be a `u64`.
         let func: fn() -> u64 = unsafe { mem::transmute(self.mc.ptr(dynasmrt::AssemblyOffset(0))) };
-        func()
+        self.exec_trace(func)
+    }
+
+    /// Jump to JITted code. This is a separate unmangled function to make it easy to set a
+    /// debugger breakpoint right before entering the trace.
+    #[no_mangle]
+    fn exec_trace(&self, t_fn: fn() -> u64) -> u64 {
```
bjorn3, line 87: 1. You can use `rbreak exec_trace` (gdb) or `break set -r exec_trace` (lldb) to set a breakpoint on any symbol containing `exec_trace`.
2. LLVM can inline this function and I expect it to do so in release mode.

vext01: > You can use rbreak exec_trace (gdb) or break set -r exec_trace (lldb) to set a breakpoint on any symbol containing exec_trace.

Didn't know that! Thanks!

> LLVM can inline this function and I expect it to do so in release mode.

Yeah. I realised that and initially I had a `#[inline(never]` on this function, but realised that it wasn't necessary for us right now: we explicitly turn off the optimiser for now because we don't want LLVM to re-order blocks.

bjorn3: Would it be possible to enable optimizations for yorick itself, but keep it disabled for user code? That would speed up the tracing and jitting while keeping LLVM from reordering blocks in the user code.

ykcompile/src/lib.rs
```diff
@@ -222,6 +234,63 @@ impl TraceCompiler {
...
+                "unknown call target".to_owned(),
+            ));
+        };
+
+        // FIXME implement support for arguments.
+        if !args.is_empty() {
+            return Err(CompileError::Unimplemented("call with args".to_owned()));
+        }
+
+        // Generally speaking, we don't know what the callee will clobber, so we save our registers
```
ptersilie, line 258: s/clobber/overwrite?

ltratt: "Clobber" is the normal term here (for better or worse), in the sense that it implies an element of chaos to what the callee might do.

ykcompile/src/lib.rs
```diff
@@ -222,6 +234,63 @@ impl TraceCompiler {
...
+        for reg in 1..X86_86_NUM_REGS {
+            dbg!(reg);
+            dynasm!(self.asm
+                ; push Rq(reg)
+            );
+        }
+
+        let sym_addr = TraceCompiler::find_symbol(sym)? as i64;
+        dynasm!(self.asm
+            ; xor rax, rax
```
ptersilie, line 270: Since the `call` will overwrite `rax` do we really need to clear it beforehand?

ykcompile/src/lib.rs
```diff
@@ -222,6 +234,63 @@ impl TraceCompiler {
...
+        // FIXME implement support for arguments.
+        if !args.is_empty() {
+            return Err(CompileError::Unimplemented("call with args".to_owned()));
+        }
+
+        // Generally speaking, we don't know what the callee will clobber, so we save our registers
+        // before doing the call. For now we assume that the caller is OK with RAX being clobbered
+        // though.
+        for reg in 1..X86_86_NUM_REGS {
+            dbg!(reg);
```
ptersilie, line 262: I think we should remove these `dbg!`'s before merging. There's another one below.

ykcompile/src/lib.rs
ptersilie: Ah, I did not know that. Not the word I would have used.

ykcompile/src/lib.rs
vext01: It might be, yeah. Might be able to filter based on crate, but we may have to move the tests to another crate so that they don't get optimised (they emulate user code).

vext01: I've made an issue for this:
https://github.com/softdevteam/ykrustc/issues/105

bjorn3: It seems that optimizations are disabled by default for tests: https://doc.rust-lang.org/cargo/reference/profiles.html#test

ykcompile/src/lib.rs
vext01: It's a GNU assembler term, yeah:

image


https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html

ykcompile/src/lib.rs
vext01: In the Sys-V calling convention, `rax` (well, part of it) specifies the number of vector register args for a varag call.

Snippet from sysv ABI spec:
image

https://uclibc.org/docs/psABI-x86_64.pdf

We can't handle arguments (let alone varags) yet, but I thought it'd be wise to zero the "hidden argument" to avoid surprises later.

I remember a time during my PhD when I spent half a day figuring out why I couldn't call `printf` from asm. It was because `rax` contained left-over junk.

Perhaps I should add a comment?

ykcompile/src/lib.rs
vext01: Oops. That was unintentionally left-in.

vext01:
Give this a go.

ykcompile/src/lib.rs
ptersilie: Ah, that makes sense then. I guess a short comment can't hurt though.

ykcompile/src/lib.rs
ltratt: It's much more widely used than just GNU's assembler!

ptersilie:
Changes look good. Please squash.

ykcompile/src/lib.rs
vext01: Fair enough! I don't use many other assemblers these days!

vext01:
splat.

PR #67, opened 2020-05-28
Replace return variable with its destination.

ptersilie:
This change replaces the default return variable `$0` with the variable
in the outer context where the return value will end up after leaving
the function. This saves us an instruction when we compile the trace.
More importantly however, this guards us against a future optimisation
in rustc that allows SIR to assign to $0 multiple times and at the
beginning of a block, which could lead to another function overwriting
its value (see https://github.com/rust-lang/rust/pull/72205).

While we are here, also fix a bug in the variable renaming code.
Currently, functions that are called consecutively (i.e. they are not                                                                                                            
nested) use the same offset to rename their variables (the offset of                                                                                                             
their outer context), which leads to them sharing the same variable                                                                                                              
names. By using an accumulator which is continuously increased to                                                                                                                
calculate the offset, we make sure consecutive functions have increasing                                                                                                         
variable names, even when after leaving a function the offset is                                                                                                                 
temporarily reset to that of the outer context.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -96,8 +96,9 @@ pub struct TraceCompiler {
     available_regs: Vec<u8>,
     /// Maps locals to their assigned registers.
     assigned_regs: HashMap<Local, u8>,
-    /// Stores the destination locals to which we copy RAX to after leaving an inlined call.
-    leaves: Vec<Option<Place>>,
+    /// Stores the destination local of the outermost function and moves its content into RAX at
+    /// the end of the trace.
+    rtn_var: Option<Place>,
 }
 
 impl TraceCompiler {
@@ -127,6 +128,15 @@ impl TraceCompiler {
 
     fn free_register(&mut self, local: &Local) {
         if let Some(reg) = self.assigned_regs.remove(local) {
+            if local == &self.rtn_var.as_ref().unwrap().local {
+                // We currently assume that we only trace a single function which leaves its return
+                // value in RAX. Since we now inline a function's return variable this won't happen
+                // automatically anymore. To keep things working, we thus copy the return value of
+                // the most outer function into RAX at the end of the trace.
+                dynasm!(self.asm
+                    ; mov rax, Rq(reg)
+                );
+            }
             self.available_regs.push(reg);
         }
     }
@@ -217,19 +227,9 @@ impl TraceCompiler {
                 },
             }
         }
-        // Remember the return destination.
-        self.leaves.push(dest.as_ref().cloned());
-        Ok(())
-    }
-
-    fn c_leave(&mut self) -> Result<(), CompileError> {
-        let dest = self.leaves.pop();
-        if let Some(d) = dest {
-            if let Some(d) = d {
-                // When we see a leave statement move whatever's left in RAX into the destination
-                // local.
-                self.mov_local_local(d.local, Local(0))?;
-            }
+        if self.rtn_var.is_none() {
+            // Remember the return variable of the most outer function.
+            self.rtn_var = dest.as_ref().cloned();
         }
         Ok(())
     }
@@ -313,7 +313,7 @@ impl TraceCompiler {
                 };
             }
             Statement::Enter(op, args, dest, off) => self.c_enter(op, args, dest, *off)?,
-            Statement::Leave => self.c_leave()?,
+            Statement::Leave => {}
             Statement::StorageLive(_) => {}
             Statement::StorageDead(l) => self.free_register(l),
             Statement::Call(target, args, dest) => self.c_call(target, args, dest)?,
@@ -388,7 +388,7 @@ impl TraceCompiler {
             // Use all the 64-bit registers we can (R15-R8, RDX, RCX).
             available_regs: vec![15, 14, 13, 12, 11, 10, 9, 8, 2, 1],
             assigned_regs: HashMap::new(),
-            leaves: Vec::new(),
+            rtn_var: None,
         };
 
         for i in 0..tt.len() {
@@ -452,7 +452,7 @@ mod tests {
             asm: dynasmrt::x64::Assembler::new().unwrap(),
             available_regs: vec![15, 14, 13, 12, 11, 10, 9, 8, 2, 1],
             assigned_regs: HashMap::new(),
-            leaves: Vec::new(),
+            rtn_var: None,
         };
 
         for _ in 0..32 {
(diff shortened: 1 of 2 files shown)


ptersilie:
Last commit needs squashing.

ykcompile/src/lib.rs
```diff
@@ -125,8 +129,26 @@ impl TraceCompiler {
         }
     }
 
+    /// Returns the currently assigned register for a given `Local`. Similar to `local_to_reg` but
+    /// read-only, i.e. this function doesn't assign `Local`'s to registers.
+    fn get_reg(&self, local: &Local) -> Result<u8, CompileError> {
```
ltratt: AFAICS this function is dead, so we should (later) squash https://github.com/softdevteam/yk/pull/67/commits/d3394ddde1bfc23dffdbb5bd541fedd56cc36071 into this commit?

ltratt: I see you already mentioned this.

ptersilie: Yes, I noticed shortly after pushing, but didn't know if force pushing is okay if I haven't assigned anyone yet.

ltratt: You did the right thing.

ykcompile/src/lib.rs
```diff
@@ -27,6 +27,8 @@ pub enum CompileError {
     /// We ran out of registers.
     /// In the long-run, when we have a proper register allocator, this won't be needed.
     OutOfRegisters,
+    /// We tried to retrieve the register for a local which doesn't have a register assigned to it.
+    UnknownLocal,
```
vext01, line 31: Shall we put the Local inside and report it when we crash?

ykcompile/src/lib.rs
```diff
@@ -96,8 +99,9 @@ pub struct TraceCompiler {
     available_regs: Vec<u8>,
     /// Maps locals to their assigned registers.
     assigned_regs: HashMap<Local, u8>,
-    /// Stores the destination locals to which we copy RAX to after leaving an inlined call.
-    leaves: Vec<Option<Place>>,
+    /// Stores the destination local of the outer most function and moves its content into RAX at
```
vext01, line 102: s/outer most/outermost/

ykcompile/src/lib.rs
```diff
@@ -125,8 +129,26 @@ impl TraceCompiler {
...
+        match self.assigned_regs.get(local) {
+            Some(u) => Ok(*u),
+            None => Err(CompileError::UnknownLocal),
+        }
+    }
+
     fn free_register(&mut self, local: &Local) {
         if let Some(reg) = self.assigned_regs.remove(local) {
+            if local == &self.rtn_var.as_ref().unwrap().local {
+                // We currently assume that we only trace a single function which leaves its return
```
vext01, line 135: Gross hack, but OK to get us going ;)

yktrace/src/tir.rs
```diff
@@ -380,8 +384,17 @@ impl VarRenamer {
 
     fn rename_place(&self, place: &Place) -> Place {
         if &place.local == &Local(0) {
-            // Local(0) is used for returning values from calls, so it doesn't need to be renamed.
-            place.clone()
+            // Replace the default return variable $0 with the variable in the outer context where
+            // the return value will end up after leaving the function. This saves us an
+            // instruction when we compile the trace. More importantly however, this guards us
```
vext01, line 409: I'd kill the "More importantly however, this guards us against a future optimisation..." bit and just leave it for the commit message. The future will soon catch up with us and the comment will go stale.

yktrace/src/tir.rs
```diff
@@ -117,6 +117,10 @@ impl TirTrace {
                 }
             };
 
+            // Initialise VarRenamer's accumulator (and thus also set the first offset) to the
+            // traces most outer number of locals.
+            rnm.init_acc(body.num_locals);
```
vext01, line 122: would this be better as an argument to `VarRename::new()`? It's only done once?

yktrace/src/tir.rs
```diff
@@ -314,6 +318,11 @@ struct VarRenamer {
     stack: Vec<u32>,
     /// Current offset used to rename variables.
     offset: u32,
+    /// Accumulator keeping track of total amount of variables used. Needed to use different
```
vext01, line 321: I think this would read better as "Accumulator keeping track of the total *number* of variables used"

I might be wrong, but "amount" feels like it implies an uncountable quantity?

yktrace/src/tir.rs
```diff
@@ -330,12 +340,22 @@ impl VarRenamer {
...
+        }
+    }
+
     fn enter(&mut self, num_locals: usize, dest: Place) {
-        // When entering an inlined function call update the current offset by adding the number of
-        // assigned variables in the outer context. Also add this offset to the stack, so we can
-        // restore it once we leave the inlined function call again.
-        self.offset += num_locals as u32;
+        // When entering an inlined function call set the offset to the current accumulator. Then
+        // increase the accumulator with the number of locals in the current function. Also add the
```
vext01, line 351: Minor wording tweak:
"increase the accumulator with the number of locals" -> "increment the accumulator by the number of locals"

vext01:
Just a handful of small comments from me.

yktrace/src/tir.rs
ptersilie: The problem is when we create `VarRename` we don't have access to `body` yet (at least not without jumping through some hoops).

ykcompile/src/lib.rs
ptersilie: Indeed. It'll be a while before we won't need this workaround anymore.

yktrace/src/tir.rs
vext01: I see. Never mind then.

ykcompile/src/lib.rs
vext01: Ah! In fact I need this for my non-inline call work, which -- like for inlined calls -- puts the return value in the destination register, which may not be rax.

ykcompile/src/lib.rs
ptersilie: I just realised this isn't even used anymore, so I removed it in 
61e680e.

ykcompile/src/lib.rs
ptersilie: Fixed in bff4293.

yktrace/src/tir.rs
ptersilie: Fixed in 35abcb1

yktrace/src/tir.rs
ptersilie: Fixed in 61cadb7

yktrace/src/tir.rs
ptersilie: Fixed in 61cadb7

ptersilie:
Think I have addressed all comments.

vext01:
So `UnknownLocal` was in fact unused?

ptersilie:
It was used in `get_reg` which I removed since it wasn't needed anymore. I just missed that the error was only needed for that.

vext01:
OK. Please squash.

ptersilie:
Squashed.

PR #68, opened 2020-06-01
Yk codegen calls with args

vext01:
This adds the ability to use arguments to calls. See commit messages for more.

The second commit is optional. It's not needed, but it might save us time/confusion in the future.

A compiler change is required. PR coming.


ykcompile/src/lib.rs
```diff
@@ -250,41 +258,99 @@ impl TraceCompiler {
...
 
-        // Generally speaking, we don't know what the callee will clobber, so we save our registers
-        // before doing the call. For now we assume that the caller is OK with RAX being clobbered
-        // though.
-        for reg in 1..X86_86_NUM_REGS {
+        // Figure out where the return value (if there is one) is going.
+        let dest_reg: Option<u8> = if let Some(d) = dest {
+            Some(self.local_to_reg(d.local)?)
+        } else {
+            None
```
ptersilie, line 282: We currently assume that there is always a return variable which is guarded by an assert (https://github.com/softdevteam/yk/blob/master/yktrace/src/tir.rs#L177). So technically there's no need for this else branch here.

ptersilie: Ah wait. That assert is only for inlined function calls. Though we could make the same assumption for native calls, since I haven't yet managed to produce code that generates calls that don't have a destination. Given `dest` is an `Option` there must be a way, but until we know for sure, we can just assume there isn't, and hopefully this will blow up some day on some example.

ykcompile/src/lib.rs
```diff
@@ -250,41 +258,99 @@ impl TraceCompiler {
...
+        // will store the return value. It's safe to assume the caller expects this to be
+        // clobbered.
+        //
+        // FIXME: Note that we don't save rax. Although this is a callee save register, the way the
+        // tests currently work is they check the last value returned at the end of the trace. This
+        // value is assumed to remain in rax. If we were to restore rax, we'd break that. Note that
+        // the register allocator never gives out rax for this precise reason.
+        //
+        // rdi, rsi, rdx, rcx, r8, r9, r10, r11.
+        let save_regs = [7, 6, 2, 1, 8, 9, 10, 11]
```
ptersilie, line 295: Maybe [remove_item](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.remove_item) is a bit shorter here and easier to read, though it is only in nightly, so I'll leave that decision up to you.

ykcompile/src/lib.rs
vext01: So `else { panic() }`?

ykcompile/src/lib.rs
vext01: Well, our ykrustc fork is technically nightly, so sounds good to me.

bjorn3: Technically the channel you use is called `dev` :)

https://github.com/softdevteam/ykrustc/blob/1e8ac85b61ba402d6bff972b48933f94f0e3f893/config.toml.example#L349-L352

```rust
# The "channel" for the Rust build to produce. The stable/beta channels only
# allow using stable features, whereas the nightly and dev channels allow using
# nightly features
#channel = "dev"
```

ykcompile/src/lib.rs
```diff
@@ -250,41 +258,99 @@ impl TraceCompiler {
...
+            .map(|r| *r)
+            .collect::<Vec<u8>>();
+        for reg in &save_regs {
             dynasm!(self.asm
                 ; push Rq(reg)
             );
         }
 
+        let stack_index = |reg: u8| -> i32 {
+            i32::try_from(save_regs.iter().rev().position(|&r| r == reg).unwrap()).unwrap()
```
ptersilie, line 307: I think this could benefit from a comment as it's meaning is difficult to unravel. For example, why is it beneficial here to reverse the iterator? Do we know that `reg` can always be found towards the end of the vec?

ykcompile/src/lib.rs
```diff
@@ -250,41 +258,99 @@ impl TraceCompiler {
...
+        };
+
+        // Sys-V ABI dictates the first 6 arguments are passed in: rdi, rsi, rdx, rcx, r8, r9.
+        let mut arg_regs = vec![9, 8, 1, 2, 6, 7]; // reversed so they pop() in the right order.
+        for arg in args {
+            let arg_reg = if let Some(r) = arg_regs.pop() {
+                r
+            } else {
+                return Err(CompileError::Unimplemented(
+                    "out of argument registers".to_owned(),
```
ptersilie, line 317: Did we not already make sure this can't happen in line 261?

ykcompile/src/lib.rs
```diff
@@ -250,41 +258,99 @@ impl TraceCompiler {
...
             dynasm!(self.asm
-                ; pop Rq(reg)
+                ; mov Rq(dest_reg), rax
             );
         }
 
-        // Put return value in place.
-        if let Some(dest) = dest {
-            let dest_reg = self.local_to_reg(dest.local)?;
+        // For tests we need this same hack as `c_enter()` uses for now.
```
ptersilie, line 356: Maybe better like "To avoid breaking tests we need to the same hack as in `c_enter.`?

ykcompile/src/lib.rs
```diff
@@ -125,9 +128,15 @@ impl TraceCompiler {
         }
     }
 
-    fn free_register(&mut self, local: &Local) {
+    fn free_register(&mut self, local: &Local) -> Result<(), CompileError> {
         if let Some(reg) = self.assigned_regs.remove(local) {
-            if local == &self.rtn_var.as_ref().unwrap().local {
+            if local
```
ptersilie, line 133: Is this how `rustfmt` formats this? Can we not at least move the `== &self` into the line above, e.g.
```
            if local == &self
                       .rtn_var
                       .as_ref()
                       .ok_or_else(|| CompileError::NoReturnLocal)?
                       .local
```

ykcompile/src/lib.rs
ptersilie: This will all be rewritten once we implement spilling, so it's not super important. But you can just add an `assert!(newdest.is_some());` to `tir.rs` before converting the `Call`, and then we can get rid of the else branch entirely (or put `unreachable` in it).

ptersilie: Actually now that I look at the `Call` conversion: how come we don't rename its return destination? https://github.com/softdevteam/yk/blob/master/yktrace/src/tir.rs#L196

ptersilie:
Some comments.

ptersilie:
By the way, did you try to see if Dynasmrs has an enum with the registers so we don't need to hard code numbers? Maybe we can use this: https://github.com/CensoredUsername/dynasm-rs/blob/5012c7fa79a62df60c0e7d5db427f1f9e866fa2a/plugin/src/arch/x64/ast.rs

ykcompile/src/lib.rs
vext01: It's due to the way the registers will be laid out on a stack. The first element in `save_regs` will be at the highest stack offset, because it will be pushed first.

Will add a comment.

ykcompile/src/lib.rs
vext01: We did. I suggest we kill the one on 261.

ykcompile/src/lib.rs
vext01: We do, I think. I'm not sure that's relevant.

`dest` is an option, so we have do something with the possibility of it being `None`, even if we know that's impossible. I'm happy with `unreachable!`. Or we could get rid of `Option` (later)?

ykcompile/src/lib.rs
ptersilie: Isn't it better to bail earlier so we avoid extra work before erroring?

vext01: We can do. We'd need a `panic()` or `unreachable()` here then.

vext01:
I found that too, but I assumed we can't get at them because they are in a compiler plugin.

ykcompile/src/lib.rs
ptersilie: Removing the `Option` might be a bit premature as I don't yet know if there are cases where it can't be `None`. We can do `unreachable!` or you just unwrap it (which will also crash if the "impossible" happens). That's up to you.

ptersilie: > We do, I think. I'm not sure that's relevant.

I don't see it. Notice how we rename the destination for the `Enter` case (https://github.com/softdevteam/yk/blob/master/yktrace/src/tir.rs#L170) but not for the `Call` case.

ptersilie:
Yes, it didn't look easily accessible. Well, let's leave it for now as it is. It's not super important.

ykcompile/src/lib.rs
vext01: This is sadly how rustfmt formats this :P

ykcompile/src/lib.rs
vext01: It needs to happen for both.

Do we just need to lift that rename out of the `if let...`?

ykcompile/src/lib.rs
ptersilie: Works for me.

ykcompile/src/lib.rs
ptersilie: If you want to finish this PR first, we can do this in a separate PR. We just need to copy this

vext01:
Ready.

ykcompile/src/lib.rs
ptersilie: Did you want to do this too or leave as is?

ykcompile/src/lib.rs
```diff
@@ -353,7 +351,7 @@ impl TraceCompiler {
             );
         }
 
-        // For tests we need this same hack as `c_enter()` uses for now.
+        // To avoid breaking tests we need this same hack as `c_enter()` uses for now.
```
ptersilie, line 354: At the risk of sounding nitpicky: "the same hack"?

ykcompile/src/lib.rs
vext01: It's kind of annoying, because `dest_reg` is `Option<u8>` not `u8`. I'd have to wrap every element of the list in `Some`.

ptersilie:
Looks good. Please squash.

vext01:
splat.

ptersilie:
Okay, guess it's time to manually built softdevteam/ykrustc#106 on builtbot so we can merge this.

vext01:
I think this is ready.

bors try.

vext01:
OK to squash?

ltratt:
Please squash.

vext01:
splat. ready for merge.

PR #70, opened 2020-06-08
Correctly rename the return value destination for external calls.

vext01:
Last week, @ptersilie correctly [pointed out](https://github.com/softdevteam/yk/pull/68#discussion_r433890619) that the return value of a call was only being renamed if the call was to be inlined.

This fixes that.

When moving code, I also moved a `FIXME` comment, which I must admit, I'm no longer able to decipher (I think Lukas may know what it means). I've opted to keep it in, but I almost removed it.

I wonder if the never type `!` would cause a `None` destination. In which case the comment is moot. Will look into it. Marking this a draft in the meantime.

### DIFF
--- a/yktrace/src/tir.rs
+++ b/yktrace/src/tir.rs
@@ -160,30 +160,32 @@ impl TirTrace {
                     args,
                     destination: dest
                 } => {
+                    // Rename the return value.
+                    //
+                    // FIXME It seems that calls always have a destination despite the field being
+                    // `Option`. If this is not always the case, we may want add the `Local` offset
+                    // (`var_len`) to this statement so we can assign the arguments to the correct
+                    // `Local`s during trace compilation.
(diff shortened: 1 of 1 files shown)


ltratt:
I don't think we need to worry about the `None` destination: the `unwrap` will catch it should we ever encounter one. I suggest we merge this PR and move on if you're OK with that.

vext01:
Bear with me a moment. I think I know a case to cause the unwrap to panic.

ltratt:
OK.

vext01:
Hrm. I was perhaps wrong.

I thought that tracing a function like:
```
fn f() -> !;
```

Would make a `None` destination, but of course we never find out, because we will never return from `f` when we trace it.

So I think the `dest` field should probably not be `Option`. I'd want to discuss with Lukas, but for now we can merge this I think.

PR #73, opened 2020-06-10
Make the textual TIR representation more consistent.

vext01:


vext01:
Try this:

```
  live($3)
  enter(sym(_ZN9ykcompile5tests7fnested17hee940008725af9d0E), [], $3, 69)
  live($70)
  enter(sym(_ZN9ykcompile5tests8fnested217h1011e2edaa6f9971E), [20u8], $70, 71)
  live($73)
  $73 = $72
  enter(sym(_ZN9ykcompile5tests8fnested317h4d600e9c273f3501E), [$73, 10u8], $70, 74)
  live($77)
  $77 = $75
  $70 = $77
  dead($77)
  leave
  dead($73)
  leave
  $3 = $70
  dead($70)
  leave
  dead($3)
```

vext01:
(I also shortened `StorageLive` and `StorageDead` to `live` and `dead`).

ltratt:
Is the `sym(...)` wrapper useful -- could we have anything that isn't a `sym`?

vext01:
There's an enum underlying that with an additional`Unknown` variant which I have seen appear in the past. Presumably when we figure out when that happens, `Unknown` will be replaced by something else that has another stringification.

Thoughts?

ltratt:
Given how often we're going to encounter this, maybe it's worth getting some sense of what this other thing might be? Or, to be honest, I'd just get rid of the `sym` wrapper and hope that when we do encounter the unknown thing we find a nice way of differentiating it from symbol names. If we don't, at worst we have to change the existing tests a bit.

vext01:
https://github.com/softdevteam/yk/blob/0c83ad251268fc102067366adb98492d48f0382e/ykpack/src/types.rs#L403

vext01:
> I'd just get rid of the sym wrapper and hope that when we do encounter the unknown thing we find a nice way of differentiating it from symbol names.

Alright, lets kill the `sym(...)` wrapping.

vext01:
Done

ltratt:
Please squash.

vext01:
Squashed and updated PR descr.

vext01:
Oops. Ran the compiler tests, but not the yk ones!

vext01:
Do we see any value in a test which tests the format of a textual TIR dump?

Arguably that's not much use, and anyway, the up and coming TIR trace tester will do a better job of it?

I'd like to remove the test if you agree?

ltratt:
> Do we see any value in a test which tests the format of a textual TIR dump?

Yes, I think it's a good idea. We might one day drop it when we have other ways of testing the same thing, but at the moment we don't have those things.

vext01:
Fixed.

ltratt:
Please squash.

vext01:
splat.

PR #74, opened 2020-06-23
Implement register spilling.

ptersilie:
This PR implements a new register allocator which means we won't have to crash anymore when we run out of registers. Note, that while our target algorithm, Bottom-Up Allocator [1], compiles traces from the bottom up, we will continue to compile things from the top, due to current restrictions in Dynasm.

This PR ended up a bit bigger than expected, but all changes should be split up nicely in separate commits that all make sense on their own.

e2196fb prepares the code for a new allocation algorithm by renaming a few variables.
c28a081 adds an `enum` which will hold the locations (e.g. stack/register) of locals.
4beb035 implements some basic register spilling.
570d791 adds some more spilling operations.
71947c5 makes sure that we only allocate as much space on the stack as is needed to hold the spilled locals.
ba58a55 implements spilling operations related to external calls.
b992e31 implements moving/spilling `u64` immediates to the stack, which isn't allowed by x86.

[1] https://dl.acm.org/doi/pdf/10.1145/3132190.3132209

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -202,14 +202,25 @@ impl TraceCompiler {
                     ; mov Rq(lreg), Rq(rreg)
                 );
             }
-            (Location::Register(_reg), Location::Stack(_off)) => {
-                todo!();
+            (Location::Register(reg), Location::Stack(off)) => {
+                dynasm!(self.asm
+                    ; mov Rq(reg), [rsp + off]
+                );
             }
-            (Location::Stack(_off), Location::Register(_reg)) => {
-                todo!();
+            (Location::Stack(off), Location::Register(reg)) => {
+                dynasm!(self.asm
+                    ; mov [rsp + off], Rq(reg)
+                );
             }
-            (Location::Stack(_off1), Location::Stack(_off2)) => {
-                todo!();
+            (Location::Stack(off1), Location::Stack(off2)) => {
+                // Since RAX is currently not available to the register allocator, we can use it
+                // here to simplify moving values from the stack back onto the stack (which x86
+                // does not support). Otherwise, we would have to free up a register via spilling,
+                // making this operation more complicated and costly.
+                dynasm!(self.asm
+                    ; mov rax, [rsp + off2]
+                    ; mov [rsp + off1], rax
+                );
             }
             _ => unreachable!(),
         }
@@ -264,8 +275,11 @@ impl TraceCompiler {
                     ; mov Rq(reg), QWORD b as i64
                 );
             }
-            Location::Stack(_offset) => {
-                todo!();
+            Location::Stack(offset) => {
+                let val = b as i32;
+                dynasm!(self.asm
+                    ; mov QWORD [rsp+offset], val
+                );
             }
             Location::NotLive => unreachable!(),
         }
@@ -400,7 +414,8 @@ impl TraceCompiler {
                     // Load argument back from the stack.
                     let idx = match self.local_to_location(*local)? {
                         Location::Register(reg) => stack_index(reg),
-                        _ => unreachable!(), // Is this true?
+                        Location::Stack(_off) => todo!(),
+                        Location::NotLive => unreachable!(),
                     };
                     dynasm!(self.asm
                         ; mov Rq(arg_reg), [rsp + idx * 8]
@@ -559,8 +574,25 @@ impl TraceCompiler {
         self.asm.finalize().unwrap()
     }
 
+    #[cfg(test)]
+    fn test_compile(tt: TirTrace) -> (CompiledTrace, i32) {
+        // Changing the registers available to the register allocator affects the number of spills,
+        // and thus also some tests. To make sure we notice when this happens we also check the
+        // number of spills in those tests. We thus need a slightly different version of the
+        // `compile` function that provides this information to the test.
+        let tc = TraceCompiler::_compile(tt);
+        let spills = tc.cur_stack_offset;
+        let ct = CompiledTrace { mc: tc.finish() };
+        (ct, spills)
+    }
+
     /// Compile a TIR trace, returning executable code.
     pub fn compile(tt: TirTrace) -> CompiledTrace {
+        let tc = TraceCompiler::_compile(tt);
+        CompiledTrace { mc: tc.finish() }
+    }
+
+    fn _compile(tt: TirTrace) -> TraceCompiler {
         let assembler = dynasmrt::x64::Assembler::new().unwrap();
 
         let mut tc = TraceCompiler {
@@ -591,7 +623,7 @@ impl TraceCompiler {
         }
 
         tc.ret();
-        CompiledTrace { mc: tc.finish() }
+        tc
     }
 
     /// Returns a pointer to the static symbol `sym`, or an error if it cannot be found.
@@ -844,6 +876,7 @@ mod tests {
         let _e = 5;
         let _f = 6;
         let h = 7;
+        let _g = true;
         h
     }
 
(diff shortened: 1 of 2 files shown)


ykcompile/src/lib.rs
```diff
@@ -49,25 +49,28 @@ impl Display for CompileError {
...
-        8 => "r8",
-        9 => "r9",
-        10 => "r10",
-        11 => "r11",
-        12 => "r12",
-        13 => "r13",
-        14 => "r14",
-        15 => "r15",
-        _ => unimplemented!(),
+fn reg_num_to_name(loc: &Location) -> &'static str {
```
vext01, line 52: Shouldn't this really be named differently now?

bjorn3:
> Bottom-Up Allocator [1]

I think I now what you are referring to, but you seem to have missed the link.

ykcompile/src/lib.rs
ptersilie: Oh, has this been merged already? I'll probably to that in a separate PR though.

ptersilie:
> > Bottom-Up Allocator [1]
> 
> I think I now what you are referring to, but you seem to have missed the link.

Whoops. Fixed. Thanks.

ykcompile/src/lib.rs
vext01: I mean, the name of the function `reg_num_to_name` is confusing, because it no longer takes a register number. How about `local_to_reg_name`?

vext01:
> current restrictions in Dynasm.

By the way, the dynasm guy mentioned in IRC earlier that backwards assembly turned out to be much harder than he anticipated, and has thus put that work on hold.

ykcompile/src/lib.rs
```diff
@@ -94,77 +97,133 @@ impl CompiledTrace {
...
         if l == Local(0) {
             // In SIR, `Local` zero is the (implicit) return value, so it makes sense to allocate
             // it to the return register of the underlying X86_64 calling convention.
-            Ok(0)
+            Ok(Location::Register(0))
         } else {
-            if self.assigned_regs.contains_key(&l) {
-                Ok(self.assigned_regs[&l])
+            if let Some(location) = self.variable_location_map.get(&l) {
+                // We already have a location for this local
```
vext01, line 132: Missing full-stop.

ykcompile/src/lib.rs
```diff
@@ -94,77 +97,133 @@ impl CompiledTrace {
...
+                .local;
+        match self.variable_location_map.get(local) {
+            Some(Location::Register(reg)) => {
+                // If this local is currently stored in a register, free it.
+                self.register_content_map.insert(*reg, None);
+                if is_rtn_var {
+                    // We currently assume that we only trace a single function which leaves its return
+                    // value in RAX. Since we now inline a function's return variable this won't happen
+                    // automatically anymore. To keep things working, we thus copy the return value of
+                    // the most outer function into RAX at the end of the trace.
```
vext01, line 175: outer-most

ykcompile/src/lib.rs
```diff
@@ -94,77 +97,133 @@ impl CompiledTrace {
...
+                dynasm!(self.asm
+                    ; mov Rq(reg), [rsp + off]
+                );
+            }
+            (Location::Stack(off), Location::Register(reg)) => {
+                dynasm!(self.asm
+                    ; mov [rsp + off], Rq(reg)
+                );
+            }
+            (Location::Stack(off1), Location::Stack(off2)) => {
```
vext01, line 215: So if we didn't have RAX reserved, we'd kick another variable to the stack, but just temporarily?

ykcompile/src/lib.rs
```diff
@@ -181,20 +240,50 @@ impl TraceCompiler {
...
+                    ; mov Rq(reg), QWORD c_val
+                );
+            }
+            Location::Stack(offset) => {
+                if c_val <= u32::MAX.into() {
+                    let val = c_val as u32 as i32;
+                    dynasm!(self.asm
+                        ; mov QWORD [rsp+offset], val
+                    );
+                } else {
```
vext01, line 257: Let's comment this up so that we remember what it does a few months down the line?

ykcompile/src/lib.rs
```diff
@@ -458,18 +588,39 @@ impl TraceCompiler {
         self.asm.finalize().unwrap()
     }
 
+    #[cfg(test)]
```
vext01, line 591: Can we add a comment saying why we did this. I know why, but let's help those who read our code later.

ykcompile/src/lib.rs
```diff
@@ -717,4 +877,132 @@ mod tests {
...
+
+    #[test]
+    fn test_spilling_simple() {
+        let th = start_tracing(Some(TracingKind::HardwareTracing));
+        many_locals();
+        let sir_trace = th.stop_tracing().unwrap();
+        let tir_trace = TirTrace::new(&*sir_trace).unwrap();
+        let (ct, spills) = TraceCompiler::test_compile(tir_trace);
+        assert_eq!(ct.execute(), 7);
+        // Changing the registers available to the register allocator affects the amount of spills,
```
vext01, line 901: This is a nitpick, but I *think* `s/amount/number/` as `amount` implies an uncountable unit, like water. Might be wrong. Just my gut feeling.

vext01:
OK, I think this looks good. Just a few comments.

ykcompile/src/lib.rs
```diff
@@ -94,77 +97,133 @@ impl CompiledTrace {
     }
 }
 
+#[derive(Debug, Clone, PartialEq)]
+enum Location {
+    Register(u8),
+    Stack(i32),
```
vext01, line 103: For `Stack`, where is the offset relative to?

ykcompile/src/lib.rs
ptersilie: Yes, that would be the plan.

ykcompile/src/lib.rs
ptersilie: It's relative to `rsp`. At the beginning of the trace we do `sub rsp`, so to get to a stack location we need to do `rsp + offset`.

vext01: OK, there's one detail missing in my mental model: how do we make sure the offsets remain correct when more stuff gets spilled?

It looks like (and I may be wrong) that it'd work like this. Suppose:

 - registers are full
 - allocate for local var `a`. Is given `Location::stack(0)`.
 - allocate for local var `b`. Is given `Location::stack(8)`.

But that doesn't look right...

ptersilie: If you mean making sure that the same variable always get allocated to the same offset, then that isn't implemented yet. This PR was already getting quite big, so I was going to do that in another one.

vext01: That wasn't what I meant, but we resolved this comment in the team meeting I think.

ykcompile/src/lib.rs
ptersilie: This is being changed in the last commit. But I can add a comment to that commit to explain what it does.

ykcompile/src/lib.rs
ptersilie: There's a comment at https://github.com/softdevteam/yk/pull/74/commits/570d79148dabc4a0434b9eae164ccccaa4a1352f#diff-162e1dfdca14767450c768ad342d54f2R887 which explains this, but I can moved it up here if you think that's a better place.

vext01: I think the additional comment would aid readability.

ykcompile/src/lib.rs
ptersilie: Fixed in 20600b7

ykcompile/src/lib.rs
ptersilie: Fixed in 20600b7

ykcompile/src/lib.rs
ptersilie: Fixed in 8524b22

ykcompile/src/lib.rs
ptersilie: Fixed in a60544c

ykcompile/src/lib.rs
ptersilie: Fixed in 11b078a

ykcompile/src/lib.rs
ptersilie: Fixed in 11b078a

ptersilie:
I think that addresses all of your comments.

vext01:
Please squash.

ptersilie:
Squashed (the autosquash feature is quite useful, though the fact you can't edit the commit message isn't great for reviewing).

PR #82, opened 2020-07-09
Fix tests in the wake of a ykrustc upstream sync.

vext01:
The Rust compiler now emits one fewer local sometimes, and this broke our spill tests.

This should go in after the upstream sync PR.


vext01:
This can go in now.

vext01:
bors try

PR #84, opened 2020-07-13
Load dwarf labels once

vext01:
See commit messages.

### DIFF
--- a/yktrace/src/hwt/mapper.rs
+++ b/yktrace/src/hwt/mapper.rs
@@ -3,36 +3,34 @@ use phdrs::objects;
 
 use crate::SirLoc;
 use hwtracer::Trace;
+use lazy_static::lazy_static;
 use std::{borrow, env, fs};
 
-pub struct HWTMapper {
-    phdr_offset: u64,
+lazy_static! {
     /// Maps a label address to its symbol name and block index.
     ///
     /// We use a vector here since we never actually look up entries by address; we only iterate
     /// over the labels checking if each address is within the range of a block.
-    labels: Option<Vec<(u64, (String, u32))>>
+    ///
+    /// The labels are the same for each trace, and they are immutable, so it makes sense for this
+    /// to be a lazy static, loaded only once and shared.
+    ///
+    /// FIXME if we want to support dlopen(), we will have to rethink this.
+    static ref LABELS: Vec<(u64, (String, u32))> = extract_labels().unwrap();
+}
+
+pub struct HWTMapper {
+    phdr_offset: u64
 }
 
 impl HWTMapper {
     pub fn new() -> HWTMapper {
         let phdr_offset = get_phdr_offset();
-        let labels = match extract_labels() {
-            Ok(l) => Some(l),
-            Err(_) => None
-        };
-        HWTMapper {
-            phdr_offset,
-            labels
-        }
+        HWTMapper { phdr_offset }
     }
 
     /// Maps each entry of a hardware trace to the appropriate SirLoc.
     pub fn map(&self, trace: Box<dyn Trace>) -> Option<Vec<SirLoc>> {
-        if !self.labels.is_some() {
-            return None;
-        }
-        let labels = self.labels.as_ref().unwrap();
         let mut annotrace = Vec::new();
         for b in trace.iter_blocks() {
             match b {
@@ -49,7 +47,7 @@ impl HWTMapper {
                     // b) `labels` is sorted, so the blocks will be appended to the trace in the
                     // correct order.
                     let mut locs = Vec::new();
-                    for (addr, (sym, bb_idx)) in labels {
+                    for (addr, (sym, bb_idx)) in &*LABELS {
                         if *addr >= start_addr && *addr <= end_addr {
                             // Found matching label.
                             locs.push((addr, SirLoc::new(sym.to_string(), *bb_idx)));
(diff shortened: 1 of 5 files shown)


yktrace/src/hwt/mapper.rs
```diff
@@ -2,73 +2,66 @@ use object::Object;
...
+lazy_static! {
     /// Maps a label address to its symbol name and block index.
     ///
     /// We use a vector here since we never actually look up entries by address; we only iterate
     /// over the labels checking if each address is within the range of a block.
-    labels: Option<Vec<(u64, (String, u32))>>
+    ///
+    /// The labels are the same for each trace, and they are immutable, so it makes sense for this
+    /// to be a lazy static, loaded only once and shared.
+   static ref LABELS: Vec<(u64, (String, u32))> = extract_labels().unwrap();
```
bjorn3, line 17:
```suggestion
    static ref LABELS: Vec<(u64, (String, u32))> = extract_labels().unwrap();
```

yktrace/src/hwt/mapper.rs
```diff
@@ -2,73 +2,66 @@ use object::Object;
...
-    phdr_offset: u64,
+lazy_static! {
     /// Maps a label address to its symbol name and block index.
     ///
     /// We use a vector here since we never actually look up entries by address; we only iterate
     /// over the labels checking if each address is within the range of a block.
-    labels: Option<Vec<(u64, (String, u32))>>
+    ///
+    /// The labels are the same for each trace, and they are immutable, so it makes sense for this
+    /// to be a lazy static, loaded only once and shared.
```
bjorn3, line 16: What about `dlopen`? Or do you plan to simply not support it?

yktrace/src/hwt/mapper.rs
vext01: Funny that rustfmt didn't spot this...

yktrace/src/hwt/mapper.rs
vext01: Not for now, no. We will have to eventually come back to it though.

yktrace/src/hwt/mapper.rs
bjorn3: Rustfmt pretty much bails on macros. It doesn't know what it is looking at. It just sees a bunch of tokens.

yktrace/src/hwt/mapper.rs
bjorn3: Maybe add a fixme?

vext01: Fixed in 1c5e833

yktrace/src/hwt/mapper.rs
```diff
@@ -2,73 +2,68 @@ use object::Object;
...
+    ///
+    /// FIXME if we want to support dlopen(), we will have to rethink this.
+    static ref LABELS: Vec<(u64, (String, u32))> = extract_labels().unwrap();
```
ltratt, line 19: Hmm, I must admit that I'm not convinced by this. The labels don't really feel like a global variable (as the dlopen possibility suggests). Is there somewhere better we can attach them to e.g. `HWTMapper`?

vext01: I don't think we should over-think this.

No part of Yorick supports dlopen right now, not just the mapper.

ltratt: I agree with not over-thinking, but global variables are very dangerous: if you ever want to make them non-local, you often find you're missing the plumbing necessary to pass them where needed, and that problem invariably gets worse over time. Put another way, I think we should only use `lazy_static` in exceptional cases.

vext01: OK.

I don't want to think about this now. So I think it's best to close the PR.

vext01: Note that SIR is loaded in a similar manner, using a global immutable lazy_static.

The SIR (like the labels) is assumed immutable, as we rely on static linkage to collect these kinds of things at compile time.

ltratt: > I don't want to think about this now. So I think it's best to close the PR.

Just to be clear: I'm not saying we shouldn't merge this PR, because, as far as I can tell, it doesn't introduce *new* global variables, right? But I think we should be more cautious in the future about introducing new global variables.

vext01: It adds a global variable as a mechanism to share the labels between all instances of a mapper. Before each new mapper re-loaded the labels from disk.

Since we don't have a "god struct" where we can store such things, I don't know where else to store this stufff other than in a lazy static.

ltratt: Ok, so I misread things. Before this PR labels weren't stored in a global variable, but this PR turns them into a global variable. Dumb question: how did we manage before this PR?

vext01: > Before each new mapper re-loaded the labels from disk.

Into a local variable.

vext01:
We discussed this offline, and decided that we should measure the overhead of loading DWARF.

For the test binary, it takes 2 or 3 seconds to load the DWARF labels. Since this cost is per-trace, it would be beneficial to merge this, although in the mid/long term, we should find a more suitable place for this stuff to live (i.e. not a lazy_static).

If you are still OK with this @ltratt, I can squash the branch.

ltratt:
Please squash.

ptersilie:
What's going on with this? Seems bors hasn't merged this properly (or Github wasn't informed properly)?

ltratt:
It's been merged (I just checked the main commits at https://github.com/softdevteam/yk/commits/master). I guess this was the period when GitHub was going wonky. Closing.

PR #85, opened 2020-07-17
Use trace inputs to get variables into and out of a trace.

ptersilie:
With the added support for trace inputs in https://github.com/softdevteam/ykrustc/pull/108 we can now get variables into and out of a trace.

We start preparing for this change by refactoring the code to take `Place` instead of `Local` in many places, so we can deal with projections correctly (69fa323). We then compile trace inputs and update all tests to use them instead of the return variable hack (d6e02a1). 229a1da adds a new test, which also tests getting variables into a trace. Finally, ccc8878 removes the now obsolete return variable hack.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -3,6 +3,7 @@
 #![feature(proc_macro_hygiene)]
 #![feature(test)]
 #![feature(core_intrinsics)]
+#![feature(yk_swt)]
 
 #[macro_use]
 extern crate dynasmrt;
@@ -20,8 +21,8 @@ use std::process::Command;
 
 use stack_builder::StackBuilder;
 use yktrace::tir::{
-    CallOperand, Constant, ConstantInt, Guard, Local, Operand, Place, Rvalue, Statement, TirOp,
-    TirTrace,
+    CallOperand, Constant, ConstantInt, Guard, Local, Operand, Place, Projection, Rvalue,
+    Statement, TirOp, TirTrace,
 };
 
 use dynasmrt::{DynasmApi, DynasmLabelApi};
@@ -78,25 +79,27 @@ fn local_to_reg_name(loc: &Location) -> &'static str {
 }
 
 /// A compiled `SIRTrace`.
-pub struct CompiledTrace {
+pub struct CompiledTrace<TT> {
     /// A compiled trace.
     mc: dynasmrt::ExecutableBuffer,
+    _pd: PhantomData<TT>,
 }
 
-impl CompiledTrace {
+impl<TT> CompiledTrace<TT> {
     /// Execute the trace by calling (not jumping to) the first instruction's address.
-    pub fn execute(&self) -> u64 {
+    pub fn execute(&self, args: TT) -> TT {
         // For now a compiled trace always returns whatever has been left in register RAX. We also
         // assume for now that this will be a `u64`.
-        let func: fn() -> u64 = unsafe { mem::transmute(self.mc.ptr(dynasmrt::AssemblyOffset(0))) };
-        self.exec_trace(func)
+        let func: fn(TT) -> TT =
+            unsafe { mem::transmute(self.mc.ptr(dynasmrt::AssemblyOffset(0))) };
+        self.exec_trace(func, args)
     }
 
     /// Actually call the code. This is a separate unmangled function to make it easy to set a
     /// debugger breakpoint right before entering the trace.
     #[no_mangle]
-    fn exec_trace(&self, t_fn: fn() -> u64) -> u64 {
-        t_fn()
+    fn exec_trace(&self, t_fn: fn(TT) -> TT, args: TT) -> TT {
+        t_fn(args)
     }
 }
 
@@ -104,11 +107,14 @@ impl CompiledTrace {
 enum Location {
     Register(u8),
     Stack(i32),
+    Arg(i32),
     NotLive,
 }
 
+use std::marker::PhantomData;
+
 /// The `TraceCompiler` takes a `SIRTrace` and compiles it to machine code. Returns a `CompiledTrace`.
-pub struct TraceCompiler {
+pub struct TraceCompiler<TT> {
     /// The dynasm assembler which will do all of the heavy lifting of the assembly.
     asm: dynasmrt::x64::Assembler,
     /// Stores the content of each register.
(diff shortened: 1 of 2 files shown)


ykcompile/src/lib.rs
```diff
@@ -118,16 +123,29 @@ pub struct TraceCompiler {
...
 
-impl TraceCompiler {
+impl<TT> TraceCompiler<TT> {
+
     fn place_to_location(&mut self, p: &Place) -> Result<Location, CompileError> {
         if !p.projection.is_empty() {
-            return Err(CompileError::Unimplemented(format!("{}", p)));
+            if Some(p.local) == self.trace_inputs_local {
+                match &p.projection[0] {
+                    // This is a hack since we have not implement proper projection lowering yet.
```
vext01, line 135: Perhaps make that a FIXME?

ykcompile/src/lib.rs
```diff
@@ -194,6 +212,7 @@ impl TraceCompiler {
                     );
                 }
             }
+            Some(Location::Arg(_)) => {},
```
vext01: Can this ever happen? Should it be `unreachable!`?

yktrace/src/lib.rs
```diff
@@ -159,10 +159,11 @@ trait ThreadTracerImpl {
 /// behaviour.
 #[trace_head]
 pub fn start_tracing(kind: Option<TracingKind>) -> ThreadTracer {
-    match kind {
+    let th = match kind {
         Some(TracingKind::SoftwareTracing) => unimplemented!("software tracing is broken"),
         None | Some(TracingKind::HardwareTracing) => hwt::start_tracing()
-    }
+    };
+    th
```
vext01, line 166: Is there a reason for this?

yktrace/src/tir.rs
```diff
@@ -98,7 +99,8 @@ pub struct SirMarkers {
 /// A TIR trace is conceptually a straight-line path through the SIR with guarded speculation.
 #[derive(Debug)]
 pub struct TirTrace {
-    ops: Vec<TirOp>
+    ops: Vec<TirOp>,
+    trace_inputs_local: Option<Local>
```
vext01, line 103: Did we want to kill `Option`?

ykcompile/src/lib.rs
```diff
@@ -96,7 +96,6 @@ impl<TT> CompiledTrace<TT> {
 
     /// Actually call the code. This is a separate unmangled function to make it easy to set a
     /// debugger breakpoint right before entering the trace.
-    #[no_mangle]
```
vext01, line 97: Kill the comment above too.

ykcompile/src/lib.rs
```diff
@@ -1126,4 +1125,17 @@ mod tests {
...
+    fn test_trace_inputs() {
+        let mut inputs = trace_inputs((1, 2, 3));
+        let th = start_tracing(Some(TracingKind::HardwareTracing));
+        inputs.0 = unsafe { add6(inputs.0, inputs.1, inputs.2, 4, 5, 6) };
+        let sir_trace = th.stop_tracing().unwrap();
+        let tir_trace = TirTrace::new(&*sir_trace).unwrap();
+        let mut args = (1,2,3);
+        TraceCompiler::<&(u64, u64, u64)>::compile(tir_trace).execute(&mut args);
+        assert_eq!(inputs.0, 21);
+        assert_eq!(inputs.0, args.0);
```
vext01, line 1103: Let's add another trace execution to the end of this test, with arguments different from those at trace time?

vext01:
This looks good!

It's satisfying to get rid of the return variable hack!

ykcompile/src/lib.rs
ptersilie: Now that it's mangled, do we even still need this function?

ykcompile/src/lib.rs
ptersilie: Whoops, that's an old comment. I just needs to go.

ykcompile/src/lib.rs
ptersilie: To be honest, I don't know. Since only trace inputs return `Arg` I need to check if they always outlive the trace so we won't see a `StorageDead` for them. This might be the case if the `inputs` variable is not used after tracing has finished.

yktrace/src/lib.rs
ptersilie: Looks like a debugging left-over. Will remove.

yktrace/src/tir.rs
ptersilie: I'll double check this.

ykcompile/src/lib.rs
ptersilie: A quick test showed that even if the `inputs` variable is not used after the trace it doesn't seem to die during the trace, though I wouldn't take that as a guarantee that this can never happen. If it does happen I'm not quite sure what the right solution is.

yktrace/src/tir.rs
ptersilie: I think in practise we should always have trace inputs (esp. since users will only use the merge point function, which does all of this behind the scenes). Though if we remove this option we should produce a proper error somewhere if trace inputs aren't set. So we should probably do this in a separate PR.

ykcompile/src/lib.rs
vext01: Shall we make it something which crashes and we can address it later if it comes up?

ykcompile/src/lib.rs
vext01: Can you try the rust-gdb thing and see if you can get it to break on the mangled name?

If that works, we should leave this as a func.

ykcompile/src/lib.rs
ptersilie: Fine with me. I'll just add `unreachable` to this and the other two cases.

yktrace/src/tir.rs
vext01: Agreed on all counts.

ptersilie:
That should address all comments, I think.

vext01:
Please squash

ptersilie:
Squashed! But just noticed I'll need to rebase as well. OK?

ltratt:
Squashing includes the right to rebase, so go ahead!

ptersilie:
Rebased!

PR #86, opened 2020-07-21
Implement binary operation `checked_add`.

ptersilie:


ltratt:
I'm not very clear: what types of integers does this work on? Is it just `u64` or `usize` or similar?

ykcompile/src/lib.rs
```diff
@@ -503,6 +507,109 @@ impl<TT> TraceCompiler<TT> {
         Ok(())
     }
 
+    fn c_checked_binop(
+        &mut self,
+        dest: &Place,
+        binop: &BinOp,
+        op1: &Operand,
+        op2: &Operand,
+    ) -> Result<(), CompileError> {
```
bjorn3, line 516: This doesn't seem to write a tuple anywhere. `CheckedBinop` returns a tuple of `(is_overflow, result_value)`.

ykcompile/src/lib.rs
```diff
@@ -503,6 +507,109 @@ impl<TT> TraceCompiler<TT> {
...
+        // Move `op1` into `dest`.
+        match op1 {
+            Operand::Place(p) => self.mov_place_place(dest, &p)?,
+            Operand::Constant(Constant::Int(ci)) => self.mov_place_constint(dest, &ci)?,
+            Operand::Constant(Constant::Bool(b)) => self.mov_place_bool(dest, *b)?,
```
bjorn3, line 521: There are no `CheckedBinops` that take a bool as operand. Only ints.

ykcompile/src/lib.rs
```diff
@@ -503,6 +507,109 @@ impl<TT> TraceCompiler<TT> {
...
+                }
+            }
+            Operand::Constant(Constant::Int(ci)) => match binop {
+                BinOp::Add => self.checked_add_const(lloc, ci),
+                _ => todo!(),
+            },
+            Operand::Constant(Constant::Bool(_b)) => todo!(),
+            Operand::Constant(c) => return Err(CompileError::Unimplemented(format!("{}", c))),
+        };
+        // TODO check for overflow with `jo <label>`
```
bjorn3, line 541: This is a very important check. For cg_clif many things in libcore broke until I performed overflow checks. Especially in the float parse and display implementation.

ykcompile/src/lib.rs
```diff
@@ -1115,4 +1225,58 @@ mod tests {
         ct.execute(&mut args2);
         assert_eq!(args2.0, 39);
     }
+
+    #[inline(never)]
+    fn checked_add(a: u8) -> u8 {
```
bjorn3, line 1230: This is only a `checked_add` when debug assertions are enabled. The integer types have a `checked_add` method that is guaranteed to be a checked add. It uses the `add_with_overflow` intrinsic internally.

ykcompile/src/lib.rs
ptersilie: Since this just compiles parts of our trace, which is then run later, we don't need to return anything here. In other words: w can't know at this point whether we will overflow or not, until we actually run the compiled trace.

ltratt: I don't think that's what Bjorn is saying (but I might be wrong): I think he's saying that if we don't return the tuple, there's no way for us to check whether `is_overflow` is true or not.

vext01: It'd be a `JC` (jump if carry flag set) instruction in the trace, wouldn't it?

ltratt: > It'd be a JC (jump if carry flag set) instruction in the trace, wouldn't it?

Jump to what? How does that code know if it's the overflow/not-overflow case? One way or the other we have to choose an encoding and, presumably, the current Rust codegen defaults to writing a tuple to heap (and, I would guess furthermore, probably expects LLVM to optimise that away).

So in a sense the answer to your question could be "yes" but we just need a consistent encoding and we have to make sure it corresponds with what other parts of the rustc codegen are expecting.

ptersilie: Actually, I think I understand what you mean now. In MIR, `checked_add`, writes the result to a tuple:
```
$8 = checked_add($7, 2u8)
$1 = ($8).0 // 0 holds the result, 1 the overflow flag
```
We are currently only writing the result to a normal `Local`. We definitely need to proper projections first before we can do that, which we currently don't have.

ltratt: Should we wait for a projections PR before going ahead with this PR then?

vext01: > Jump to what? 

To a `ud2` instruction or similar.

But the missing detail here is that the check is MIR (and thus SIR) level, so in fact @bjorn3  is right. This should return a tuple eventually (once we have support for types that are no u64).

I'm currently wrestling with `fm`, otherwise the types branch is close...

ykcompile/src/lib.rs
ptersilie: Fair enough! I can change this to `unreachable!()`.

ykcompile/src/lib.rs
ptersilie: Tracing currently only works with debug enabled, but this is useful to keep in mind, if/when we managed to get tracing to work without debug mode.

ykcompile/src/lib.rs
ptersilie: > This should return a tuple eventually (once we have support for types that are no u64).

For avoidance of confusion: this should compile code that writes the result and overflow flag to a tuple (so the function this comment is annotated will continue to not return anything).

ptersilie: After some offline discussion with @vext01, we've reached the following conclusion:

We can't trace a program that overflows, since the interpreter would just crash during tracing. This means that the trace assumes things never overflow, so if they do we'll need guards which exit the trace gracefully. The guards would have to check the tuple returned by `checked_add`, so we will eventually need to write the overflow flag to that tuple. However, since we don't have blackholing we currently don't have guards either. So for the time being we will have to just crash the trace instead, which we can do with a jump to an `ud2` instructions as explained above by @vext01.

This means, even if do projections first and write the overflow flag, we won't use it anyway for the time being. Though I can be convinced to do projections first if @ltratt deems it necessary before we proceed with this PR.

ltratt: Am I right in thinking that we'd have to rewrite all this after projections is done anyway?

ptersilie: Not really. There's only one place here which uses projections: https://github.com/softdevteam/yk/pull/86/files#diff-162e1dfdca14767450c768ad342d54f2R143. But even with projections that line shouldn't look all that different.

What we can do however, once we have projections, is write the overflow flag to the `checked_add` return tuple, but we won't be using it until we have blackholing.

ltratt: OK, I'm happy to go with what you and Edd think.

ykcompile/src/lib.rs
bjorn3: There is a differemce between debuginfo and debug assertions. The former is -Cdebuginfo the later -Cdebug-assertions. They can independently be changed.

ykcompile/src/lib.rs
ptersilie: Fixed in b264e6c

ykcompile/src/lib.rs
ptersilie: Fixed in d434138

ykcompile/src/lib.rs
ptersilie: See d434138

ykcompile/src/lib.rs
ptersilie: Fixed in 9b3d3f5. I've renamed the functions to make clear that these tests can also be used to test unchcked `add` (e.g. by passing `overflow-checks=off`). However, doing so currently results in an `UnimplementedError`, as we haven't implemented unchecked `add` yet.

ykcompile/src/lib.rs
```diff
@@ -607,6 +721,8 @@ impl<TT> TraceCompiler<TT> {
         // Jump to the label that reserves stack space for spilled locals.
         dynasm!(self.asm
             ; jmp ->reserve
+            ; ->crash:
```
vext01, line 724: Did you (manually) check if the trace crashes ok?

vext01:
Just one more question from me

ykcompile/src/lib.rs
ptersilie: Yes, I changed one of the tests to overflows in the trace and it segfaulted with `Invalid operation` (or something similar).

vext01:
Please squash.

ptersilie:
Squashed!

PR #91, opened 2020-07-27
Compile references and dereferences.

ptersilie:
I've left a few `todo`'s in the code to try and keep this PR small. We won't be needing them right now so we can implement them at a later stage.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -106,6 +106,7 @@ enum Location {
     Register(u8),
     Stack(i32),
     Arg(i32),
+    Deref(Box<Location>),
     NotLive,
 }
 
@@ -132,6 +133,7 @@ impl<TT> TraceCompiler<TT> {
             if Some(p.local) == self.trace_inputs_local {
                 match &p.projection[0] {
                     Projection::Field(idx) => Ok(Location::Arg((idx * 8) as i32)),
+                    Projection::Deref => unreachable!(),
                     Projection::Unimplemented(s) => {
                         Err(CompileError::Unimplemented(format!("{}", s)))
                     }
@@ -140,6 +142,9 @@ impl<TT> TraceCompiler<TT> {
                 // TODO deal with remaining projections
                 match &p.projection[0] {
                     Projection::Field(0) => self.local_to_location(p.local),
+                    Projection::Deref => self
+                        .local_to_location(p.local)
+                        .map(|l| Location::Deref(Box::new(l))),
                     _ => Err(CompileError::Unimplemented(format!("{}", p))),
                 }
             }
@@ -192,6 +197,7 @@ impl<TT> TraceCompiler<TT> {
             }
             Some(Location::Stack(_)) => {}
             Some(Location::Arg(_)) => unreachable!(),
+            Some(Location::Deref(_)) => unreachable!(),
             Some(Location::NotLive) => unreachable!(),
             None => unreachable!(),
         }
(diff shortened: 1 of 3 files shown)


ykcompile/src/lib.rs
```diff
@@ -251,11 +257,77 @@ impl<TT> TraceCompiler<TT> {
                     ; mov [rdi + aoff], rax
                 );
             }
+            (Location::Register(reg), Location::Deref(boxed)) => match *boxed {
+                Location::Stack(off) => {
+                    dynasm!(self.asm
+                        ; mov Rq(reg), [rbp - off]
```
vext01, line 263: Do I see it right that we dereference two things here?

vext01: Once to get the pointer off the stack, again to dereference the pointer?

ykcompile/src/lib.rs
```diff
@@ -251,11 +257,77 @@ impl<TT> TraceCompiler<TT> {
...
         }
         Ok(())
     }
 
+    fn mov_place_ref(&mut self, p1: &Place, p2: &Place) -> Result<(), CompileError> {
+        let lloc = self.place_to_location(p1)?;
+        // We can only reference Locals living on the stack. So move it there if it doesn't.
+        let rloc = match self.place_to_location(p2)? {
+            Location::Register(reg) => {
+                let offset = self.stack_builder.alloc(8, 8) as i32;
```
vext01, line 300: (8, 8) here is the size and align of a pointer? I wonder if it should be a constant?

ykcompile/src/lib.rs
ptersilie: Yes and no. We have a reference/pointer on the stack (in `[rbp-off]`). So we first move that pointer into a register (first instruction), and then dereference the pointer to get its value (second instruction).

ykcompile/src/lib.rs
```diff
@@ -1331,4 +1427,90 @@ mod tests {
...
+
+    #[test]
+    fn test_ref_deref() {
+        let mut inputs = trace_inputs((0,));
+        let th = start_tracing(Some(TracingKind::HardwareTracing));
+        inputs.0 = ref_deref();
+        let sir_trace = th.stop_tracing().unwrap();
+        let tir_trace = TirTrace::new(&*sir_trace).unwrap();
+        let ct = TraceCompiler::<&(u64,)>::compile(tir_trace);
+        let mut args = (0,);
```
vext01, line 1447: Let's get into the habit of trying different args in each test case. I think it'll reduce "works by luck" test passes.

ykcompile/src/lib.rs
ptersilie: Dunno, haven't really been thinking about the type stuff yet, so just copy-pasted from above. We can fix that when we do the type changes?

ykcompile/src/lib.rs
```diff
@@ -106,6 +106,7 @@ enum Location {
     Register(u8),
     Stack(i32),
     Arg(i32),
+    Deref(Box<Location>),
```
vext01, line 109: Perhaps not for this PR, but IIRC we agreed that all of the projections would fall under one "Projection" variant?

I'm also still wondering if `Stack(offset)` should be `Mem(reg, offset)` to handle things like heap allocs or offsets from temporary registers, but again, possibly for another PR.

ykcompile/src/lib.rs
vext01: ok

ykcompile/src/lib.rs
ptersilie: It depends what you do with the args. If you overwrite it it won't matter, but if the output depends on it, we can't just randomly choose an input value.

ykcompile/src/lib.rs
vext01: ok

ykcompile/src/lib.rs
vext01: Ah, I see. The "input" here is actually only an output.

ptersilie: Yes.

ykcompile/src/lib.rs
ptersilie: I agree. One thing about `Deref`s though is that I think it needs to be recursive, which is why I just made it part of `Location`.

I always assumed that some of these things will change later on. It's hard to say what this stuff needs to look like until we have actual examples (e.g. heap alloc) to refer to. And we can't really solve everything in one go. So we'll have to come back to it in another PR.

vext01: Yep, that's fine.

Let's be on the lookout for variants in Location that can be generalised later.

ptersilie:
If there are no comments for me to address, then I think this is ready for merging.

PR #92, opened 2020-08-06
Assembly helpers.

vext01:
@ptersilie, here's a draft of the assembler helpers work.

 - `asm_mem_reg!` emits an assembler instruction of your choosing (`mov`, `add`, ...) using the right size qualifier. Eventually we can ask SIR for the size of a type (but not just yet).
 - If this looks good I'll add a `asm_reg_mem!` too.
 - It's not clear if it's necessary for a `asm_reg_reg!`. Wouldn't `mov eax, ebx` do the same as `mov rax, rbx`?

Tests pass.

Raising as a draft to check I've not overlooked anything. Does this look usable?

Thanks to @CensoredUsername for helping with macro details. Please feel free to comment :)

### DIFF
--- a/ykcompile/src/asm_helpers.rs
+++ b/ykcompile/src/asm_helpers.rs
@@ -1,0 +1,30 @@
+//! Helpers for generating assembler code with dynasm.
+
+/// Emits a 'mem <- reg'  assembler instruction using the desired size qualifier.
+macro_rules! asm_mem_reg {
+    ($dasm: expr, $size: expr, $op: expr, $mem: expr, $reg: expr) => {
+        match $size {
+            1 => {
+                dynasm!($dasm
+                    ; $op BYTE $mem, Rb($reg)
+                );
+            }
+            2 => {
+                dynasm!($dasm
+                    ; $op WORD $mem, Rw($reg)
+                );
+            },
+            4 => {
+                dynasm!($dasm
+                    ; $op DWORD $mem, Rd($reg)
+                );
+            },
+            8 => {
+                dynasm!($dasm
+                    ; $op QWORD $mem, Rq($reg)
+                );
+            }
+            _ => panic!("Invalid size operand: {}", $size),
+        }
+    }
+}
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -9,6 +9,8 @@
 extern crate dynasmrt;
 extern crate test;
 
+#[macro_use]
+mod asm_helpers;
 mod stack_builder;
 
 use dynasmrt::{x64::Rq::*, Register};
@@ -25,6 +27,10 @@ use yktrace::tir::{
 };
 
 use dynasmrt::{DynasmApi, DynasmLabelApi};
+
+// FIXME: this is a lie. I'm using this only to identify areas that need proper size lookups from
+// SIR types.
+const SIZE_ALL: i32 = 8;
 
 #[derive(Debug, Hash, Eq, PartialEq)]
 pub enum CompileError {
@@ -221,9 +227,7 @@ impl<TT> TraceCompiler<TT> {
                 );
             }
             (Location::Stack(off), Location::Register(reg)) => {
-                dynasm!(self.asm
-                    ; mov [rbp - off], Rq(reg)
-                );
+                asm_mem_reg!(&mut self.asm, SIZE_ALL, mov, [rbp - off], reg);
             }
             (Location::Stack(off1), Location::Stack(off2)) => {
                 // Since RAX is currently not available to the register allocator, we can use it
@@ -232,8 +236,8 @@ impl<TT> TraceCompiler<TT> {
                 // making this operation more complicated and costly.
                 dynasm!(self.asm
                     ; mov rax, [rbp - off2]
-                    ; mov [rbp - off1], rax
                 );
+                asm_mem_reg!(&mut self.asm, SIZE_ALL, mov, [rbp - off1], RAX.code());
             }
             (Location::Register(reg), Location::Arg(off)) => {
                 dynasm!(self.asm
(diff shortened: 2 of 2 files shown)


ptersilie:
I don't think `asm_reg_reg!` is necessary either, as moving registers doesn't require specifying the size. Though I'm unsure if we will do something like `mov rax, eax` in the future.

vext01:
In dynasm you can specify the register sizes, kind of, with `Rq`, `Rd`, ... but I think it will generate the same code.

I'm going to verify that now to be sure.

bjorn3:
> Though I'm unsure if we will do something like `mov rax, eax` in the future.

That would be `movzx` or `movsx`, depending on if you want zero or sign extending.

ykcompile/src/lib.rs
```diff
@@ -232,8 +236,8 @@ impl<TT> TraceCompiler<TT> {
                 // making this operation more complicated and costly.
                 dynasm!(self.asm
                     ; mov rax, [rbp - off2]
-                    ; mov [rbp - off1], rax
                 );
+                asm_mem_reg!(&mut self.asm, SIZE_ALL, mov, [rbp - off1], RAX.code());
```
ptersilie, line 240: Maybe we should just pass `RAX` here. Thought for that we need to refactor a bit of code, so maybe we do a refactoring PR at some later point.

vext01: It may or may not be possible. The macro is expecting something which can go inside a `Rq()`.

vext01:
```rust
use dynasmrt::{dynasm, DynasmApi, x64::Rq::{RAX, RBX}, Register};
use std::fs::File;
use std::io::Write;

fn main() {
    let mut ops = dynasmrt::x64::Assembler::new().unwrap();

    dynasm!(ops
        // Using "direct registers". Note that adding QWORD or DWORD here makes an error.
        ; mov rax, rbx
        ; mov eax, ebx
        //; mov QWORD rax, rbx                          error: size hint with direct register.
        //; mov QWORD eax, ebx                          error: same error ^

        // Using "dynamic registers".
        ; mov Rq(RAX.code()), Rq(RBX.code())
        ; mov Rd(RAX.code()), Rd(RBX.code())
        //; mov QWORD Rq(RAX.code()), Rq(RAX.code())    error: size hint with direct register.
        //; mov DWORD Rd(RAX.code()), Rd(RBX.code())    error: same error ^
    );

    let buf = ops.finalize().unwrap();
    let mut f = File::create("/tmp/code").unwrap();
    f.write_all(&buf).unwrap();
}
```

Disassembled code according to r2:
```
            0x00000000      4889d8         mov rax, rbx                                                   
            0x00000003      89d8           mov eax, ebx                                                   
            0x00000005      4889d8         mov rax, rbx                                                   
            0x00000008      4089d8         mov eax, ebx                                                   
```

That wasn't what I was expecting!

I also think that when I come to implement `asm_reg_mem!` I'm going to hit some of these same errors i pointed out above. I'd better try that now.

ykcompile/src/asm_helpers.rs
```diff
@@ -0,0 +1,30 @@
+//! Helpers for generating assembler code with dynasm.
+
+/// Emits a 'mem <- reg'  assembler instruction using the desired size qualifier.
+macro_rules! asm_mem_reg {
```
ptersilie, line 4: We could merge `asm_mem_reg` and `asm_reg_mem` into one, by parsing the expressions. Though I am not sure if it's worth it.

ptersilie: Also, will we be needing a `asm_mem_mem` at some point?

ptersilie:
I've added some thoughts. Otherwise looks sensible.

ykcompile/src/lib.rs
ptersilie: Can't we just say `$reg.code()` inside the macro?

ykcompile/src/asm_helpers.rs
vext01: Is that even encodable in x86_64?

ptersilie:
> ```
>             0x00000000      4889d8         mov rax, rbx                                                   
>             0x00000003      89d8           mov eax, ebx                                                   
>             0x00000005      4889d8         mov rax, rbx                                                   
>             0x00000008      4089d8         mov eax, ebx                                                   
> ```
> 
> That wasn't what I was expecting!

I assume you were expecting the two `mov eax, ebx` to be the same.

ykcompile/src/asm_helpers.rs
ptersilie: You are right, it's not. And it's probably best to leave the spilling outside of the helper.

vext01:
> I assume you were expecting the two mov eax, ebx to be the same.

Yes, but it only has a prefix byte. They may be equivalent encodings.

CensoredUsername:
>That wasn't what I was expecting!

And yet it's what x64 requires ;). There's good reasons for them. 

A first rule to remember is that dynasm produces templates of a fixed length, so if you use dynamic registers the chosen encoding must allow for _all_ dynamic registers to be interpolated into the template properly.

With that in mind, look at the different variants:

            0x00000000      4889d8         mov rax, rbx        ; Moves 64 bits from register rbx to rax. REX prefix required for 64-bit operand size
            0x00000003      89d8           mov eax, ebx        ; Moves 32 bits from register rbx to rax. Zeros the upper 32 bits of rax.
            0x00000005      4889d8         mov rax, rbx        ; Moves 64 bits. REX prefix required for 64-bit operand size, as well as possible use of R8-R15 due to dynamic registers used.
            0x00000008      4089d8         mov eax, ebx       ; Moves 32 bits from register rbx to rax. Zeros the upper 32 bits of rax. REX prefix used for possible use of R8-R15 due to dynamic registers used.

vext01:
I see! Thanks for the explanation. Must admit, my raw encoding knowledge is rusty :)

ykcompile/src/asm_helpers.rs
vext01: >  probably best to leave the spilling outside of the helper.

Why's that?

vext01:
> I don't think asm_reg_reg! is necessary either, as moving registers doesn't require specifying the size. Though I'm unsure if we will do something like mov rax, eax in the future.

I think we do actually need `asm_reg_reg!` for operations like `add` where the status flags will be set differently depending upon the size of the operands.  I think that's correct, anyway.

ykcompile/src/asm_helpers.rs
ptersilie: More for aesthetic reasons as we'd have to forward the register allocator to the helper function.

ykcompile/src/lib.rs
vext01: I'll investigate.

vext01: If we put `.code()` inside the macro, then we cannot use it with `Location::Register()` which stores the raw register number already.

vext01:
Closing this for now. May try again (with more limited scope) later.

PR #93, opened 2020-08-07
Combine `Location::Arg` and `Location::Stack`.

vext01:
This means we have fewer cases to consider.

While we are here, let's have the stack builder directly hand out `Location::Mem` instances, rather than raw offsets.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -101,13 +101,26 @@ impl<TT> CompiledTrace<TT> {
     }
 }
 
+/// Represents a memory location using a register and an offset.
+#[derive(Debug, Clone, PartialEq)]
+struct RegAndOffset {
+    reg: u8,
+    offs: i32,
+}
+
 #[derive(Debug, Clone, PartialEq)]
 enum Location {
     Register(u8),
-    Stack(i32),
-    Arg(i32),
+    Mem(RegAndOffset),
     Deref(Box<Location>),
     NotLive,
+}
+
+impl Location {
+    /// Creates a new memory location from a register and an offset.
+    fn new_mem(reg: u8, offs: i32) -> Self {
+        Self::Mem(RegAndOffset { reg, offs })
+    }
 }
 
 use std::marker::PhantomData;
@@ -132,7 +145,7 @@ impl<TT> TraceCompiler<TT> {
         if !p.projection.is_empty() {
             if Some(p.local) == self.trace_inputs_local {
                 match &p.projection[0] {
-                    Projection::Field(idx) => Ok(Location::Arg((idx * 8) as i32)),
+                    Projection::Field(idx) => Ok(Location::new_mem(RDI.code(), (idx * 8) as i32)),
                     Projection::Deref => unreachable!(),
                     Projection::Unimplemented(s) => {
                         Err(CompileError::Unimplemented(format!("{}", s)))
@@ -178,8 +191,10 @@ impl<TT> TraceCompiler<TT> {
                 } else {
                     // All registers are occupied, so we need to spill the local to the stack. For
                     // now we assume that all spilled locals are 8 bytes big.
-                    let loc = Location::Stack(self.stack_builder.alloc(8, 8) as i32);
-                    loc
+                    Location::new_mem(
+                        RBP.code(),
+                        -i32::try_from(self.stack_builder.alloc(8, 8)).unwrap(),
+                    )
                 };
                 let ret = loc.clone();
                 self.variable_location_map.insert(l, loc);
@@ -195,8 +210,7 @@ impl<TT> TraceCompiler<TT> {
                 // If this local is currently stored in a register, free it.
                 self.register_content_map.insert(*reg, None);
             }
-            Some(Location::Stack(_)) => {}
-            Some(Location::Arg(_)) => unreachable!(),
+            Some(Location::Mem { .. }) => {}
             Some(Location::Deref(_)) => unreachable!(),
             Some(Location::NotLive) => unreachable!(),
             None => unreachable!(),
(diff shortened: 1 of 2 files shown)


ykcompile/src/lib.rs
```diff
@@ -101,15 +101,28 @@ impl<TT> CompiledTrace<TT> {
...
+    reg: u8,
+    offs: i32,
+}
+
 #[derive(Debug, Clone, PartialEq)]
 enum Location {
     Register(u8),
-    Stack(i32),
-    Arg(i32),
+    Mem(RegAndOffset),
```
ptersilie, line 114: Can we not merge the `struct` into the `enum` here:
```
enum Location {
    Register(u8),
    Mem { reg: u8, offs: i32 },
```

This also means we can get rid of the `new_mem` function below.

ykcompile/src/lib.rs
```diff
@@ -178,8 +191,10 @@ impl<TT> TraceCompiler<TT> {
                 } else {
                     // All registers are occupied, so we need to spill the local to the stack. For
                     // now we assume that all spilled locals are 8 bytes big.
-                    let loc = Location::Stack(self.stack_builder.alloc(8, 8) as i32);
-                    loc
+                    Location::new_mem(
+                        RBP.code(),
+                        -i32::try_from(self.stack_builder.alloc(8, 8)).unwrap(),
```
ptersilie: Maybe writing this as `-1 * i32::...` makes it more difficult to overlook the `-` here.

ptersilie: We should also change `stack_builder` to return an i32, but that's for another PR.

ptersilie: That being said, it does return `u32`. Shouldn't it be safe to cast to `i32` using `as` in this case?

ykcompile/src/lib.rs
vext01: Can do. I actually started out with something like this, but went for (too much?) abstraction.

ykcompile/src/lib.rs
vext01: The stack builder now returns a Location directly. See the second commit.

ykcompile/src/lib.rs
```diff
@@ -362,22 +354,16 @@ impl<TT> TraceCompiler<TT> {
                     let v1 = c_val as u32 as i32;
                     let v2 = (c_val >> 32) as u32 as i32;
                     dynasm!(self.asm
-                        ; mov DWORD [rbp-offset], v1
-                        ; mov DWORD [rbp-offset+4], v2
+                        ; mov DWORD [Rq(ro.reg) + ro.offs], v1
+                        ; mov DWORD [Rq(ro.reg) + ro.offs + 4], v2
```
ptersilie, line 364: I think we need to be careful here. This is still correct for stack locations. But I don't think it is for `Arg`s, though I'll have to think a bit about what this should be instead.

ykcompile/src/stack_builder.rs
```diff
@@ -12,12 +15,12 @@ pub struct StackBuilder {
 ///
 /// The top of the stack, before any calls to `alloc()` is assumed to be appropriately aligned.
 impl StackBuilder {
-    /// Allocate an object of given size and alignment on the stack, returning an offset which
-    /// is intended to be subtracted from the base pointer.
-    pub fn alloc(&mut self, size: u64, align: u64) -> u32 {
+    /// Allocate an object of given size and alignment on the stack, returning a `Location::Mem`
+    /// describing the poisition of the allocation. The stack is assumed to grow down.
```
ptersilie, line 19: s/poisition/position

ykcompile/src/lib.rs
ptersilie: Less code is rarely a bad idea. :-)

ykcompile/src/lib.rs
ptersilie: Should we still make that change there? At least now it's only one place to change. ;-)

ykcompile/src/lib.rs
vext01: Actually, having started this change, I think I prefer the additional struct, as it makes constructing and de-structuring less pleasant. Can we keep it as-is?

ptersilie: Not sure what you mean. It should be just replacing `Location::new_mem(` with `Location::Mem {`. And in the matches `Location::Mem(ro1)` with `Location::Mem(reg, off)`. To me it feels like the code would becleaner without the struct.

ykcompile/src/lib.rs
vext01: Are you sure it's incorrect? I've just scribbled stack diagrams on paper and I think it remains correct.

E.g. writing two words at rbp-8 and rdi+0:

```
Stack                              Arg

+--------+  rbp                   +---------+  rdi + 8
|write2  |                        |write2   |
|        |                        |         |
+--------+  rbp - 4               +---------+  rdi + 4
|write1  |                        |write1   |
|        |                        |         |
+--------+  rbp - 8               +---------+  rdi
```

In both cases, adding 4 seems correct. Right?

ykcompile/src/lib.rs
vext01: I think it feels a little awkward as you have to keep trace of the two fields inside.

e.g.:
```diff
-            (Location::Register(reg), Location::Mem(ro)) => {
+            (Location::Register(reg), Location::Mem{reg: reg2, offs}) => {
                 dynasm!(self.asm
-                    ; mov Rq(reg), [Rq(ro.reg) + ro.offs]
+                    ; mov Rq(reg), [Rq(reg2) + offs]
                 );
```

I much prefer having the two related items encapsulated in `ro`.

ykcompile/src/lib.rs
vext01: Do you mean, why don't we use `as` in the stackbuilder when creating a location? Instead of `try_from`?

```rust
Location::new_mem(RBP.code(), -i32::try_from(self.stack_top).unwrap())
```

Using `as` wouldn't be safe, as an i32 can't express all of the numbers `stack_top` can (it's a `u64`).

Generally, I try to avoid `as` unless really necessary. Using `try_from` costs nothing when there can be no error, and it protects us, should the type change behind our backs.

ykcompile/src/lib.rs
ptersilie: I think I prefer this version. You could simplify it a bit more:
```
```diff
-            (Location::Register(regl), Location::Mem(ro)) => {
+            (Location::Register(regl), Location::Mem{reg, offs}) => {
                 dynasm!(self.asm
-                    ; mov Rq(regl), [Rq(ro.reg) + ro.offs]
+                    ; mov Rq(regl), [Rq(reg) + offs]
                 );
```
```

But if it really pains you that much then we can go with yours.

ykcompile/src/lib.rs
ptersilie: I meant adding the `-1 * i32::...`, otherwise the `-` can be easily overlooked (I barely noticed it was there the first time I saw this code).

ykcompile/src/lib.rs
ptersilie: You're right!

ykcompile/src/lib.rs
vext01: Fixed.

ykcompile/src/lib.rs
vext01: It doesn't pain me, per se. I just prefer it :P

vext01:
I think that's everything?

ptersilie:
I see you chose to ignore my `Mem` suggestion. ;-) Fair enough. Let's go with your version. Please squash.

vext01:
splat.

PR #94, opened 2020-08-11
Use sized operations throughout the trace compiler.

vext01:
In doing so:
 - Fix missing local decls in TIR compiler.
 - Implement more type serialisations.
 - Ignore some now problematic tests.
 - Since Tuples and Structs are similar, make a Fields struct.
 - Update spilling checks -- we now spill different sizes.

We only cover cases required by the test suite for now. In the future
I'd like to try and devise a solution to prevent the exponential
explosion of cases we are already beginning to see.

~~A companion PR will be up soon.~~
https://github.com/softdevteam/ykrustc/pull/117

Sorry this is such a big change!

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -19,6 +19,8 @@ use std::convert::TryFrom;
 use std::fmt::{self, Display, Formatter};
 use std::mem;
 use std::process::Command;
+use ykpack::{SignedIntTy, Ty, TypeId, UnsignedIntTy};
+use yktrace::sir::SIR;
 use yktrace::tir::{
     BinOp, CallOperand, Constant, ConstantInt, Guard, Local, Operand, Place, Projection, Rvalue,
     Statement, TirOp, TirTrace,
@@ -133,6 +135,7 @@ impl Location {
 }
 
 use std::marker::PhantomData;
+use ykpack::LocalDecl;
 
 /// The `TraceCompiler` takes a `SIRTrace` and compiles it to machine code. Returns a `CompiledTrace`.
 pub struct TraceCompiler<TT> {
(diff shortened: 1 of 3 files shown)


ykcompile/src/lib.rs
```diff
@@ -144,12 +147,42 @@ pub struct TraceCompiler<TT> {
...
     stack_builder: StackBuilder,
     _pd: PhantomData<TT>,
 }
 
 impl<TT> TraceCompiler<TT> {
+    fn can_live_in_register(tyid: &TypeId) -> bool {
+        // FIXME: optimisation: small structs and tuples etc. could actually live in a register.
+        let ty = SIR.ty(tyid);
+        match ty {
+            Ty::UnsignedInt(ui) => match ui {
```
ptersilie, line 162: Since these are all true and we won't see any other types here, should we just squash this down to a single `true` statement?

vext01: Good point.

ykcompile/src/lib.rs
```diff
@@ -225,65 +261,158 @@ impl<TT> TraceCompiler<TT> {
         Ok(())
     }
 
+    /// Get the type of a place.
+    fn place_ty(&self, p: &Place) -> &Ty {
+        let base_ty = SIR.ty(&self.local_decls[&p.local].ty);
+        match p {
```
ptersilie, line 267: I found this unnecessarily hard to read. Can we turn this into a normal if block, e.g.
```
if let Place { local, projection } = p {
    if projection.is_empty() {
        base_ty
    } else {
        ...
    }
}
```

vext01: Sure

ykcompile/src/lib.rs
```diff
@@ -225,65 +261,158 @@ impl<TT> TraceCompiler<TT> {
...
+                // arbitrarily long chains of projections.
+                assert_eq!(projection.len(), 1);
+                match projection[0] {
+                    Projection::Field(idx) => match base_ty {
+                        Ty::Struct(sty) => SIR.ty(&sty.fields.tys[usize::try_from(idx).unwrap()]),
+                        Ty::Tuple(tty) => {
+                            let sidx = usize::try_from(idx).unwrap();
+                            let fty = &tty.fields.tys[sidx];
+                            let sty = SIR.ty(fty);
+                            println!("{} {:?} {}", sidx, fty, sty);
```
ptersilie, line 286: Forgotten print? If we remove it we can also squash this into a one liner as in the `Struct` case.

ykcompile/src/lib.rs
```diff
@@ -144,12 +147,42 @@ pub struct TraceCompiler<TT> {
...
+        // FIXME: optimisation: small structs and tuples etc. could actually live in a register.
+        let ty = SIR.ty(tyid);
+        match ty {
+            Ty::UnsignedInt(ui) => match ui {
+                UnsignedIntTy::U8 => true,
+                UnsignedIntTy::U16 => true,
+                UnsignedIntTy::U32 => true,
+                UnsignedIntTy::U64 => true,
+                UnsignedIntTy::Usize => true,
+                UnsignedIntTy::U128 => true,
```
vext01, line 168: There's actually a bug here!

ykcompile/src/lib.rs
vext01: Yes, that's left over debugging.

ykcompile/src/lib.rs
ptersilie: Indeed.

ptersilie: Should still do something like
```
match ty {
    U128 => false,
    _ => true
}
```

vext01: Yep, that's what I've got.

ykcompile/src/lib.rs
```diff
@@ -596,25 +758,30 @@ impl<TT> TraceCompiler<TT> {
...
-            Operand::Constant(Constant::Int(ci)) => self.mov_place_constint(dest, &ci)?,
+            Operand::Place(p) => self.mov_place_place(&val_dest, &p)?,
+            Operand::Constant(Constant::Int(ci)) => self.mov_place_constint(&val_dest, &ci)?,
             Operand::Constant(Constant::Bool(_b)) => unreachable!(),
             Operand::Constant(c) => return Err(CompileError::Unimplemented(format!("{}", c))),
         };
-        // Add together `dest` and `op2`.
-        let lloc = self.place_to_location(dest)?;
+        // Add together `val_dest` and `op2`.
+        let lloc = self.place_to_location(&val_dest)?;
```
ptersilie, line 749: I think this means that we currently only allocate the first field of the tuple (and never the tuple itself). Which at the moment is fine, since those are equivalent. Though I'm not sure if this could cause problems in the future.

I think ideally at some point the tuple should be allocated somewhere, and `val_dest` should just be the location of the tuple + the offset of the field. Which makes me think that we might need a `Location::Mem(mem, offset)` in the future.

vext01: Good spot! I'd like to fix this now. It will surely bite use later!

ykcompile/src/lib.rs
```diff
@@ -1417,6 +1634,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore] // FIXME: Need to type our projections.
```
ptersilie, line 1613: Is just the test broken or is ref/deref in general broken now?

ykcompile/src/lib.rs
vext01: But hold on. 

```
let lloc = self.place_to_location(&val_dest)?;
```

^ still results in a call to `local_to_location` which should allocate the whole tuple.

Do you agree?

ykcompile/src/lib.rs
vext01: When you have a deref on the left hand side of the assignment, we currently don't have a way to know its type. The same will be true for other projection types.

This can be fixed, but we don't need it right now, I don't think.

ykcompile/src/lib.rs
ptersilie: Yes, I wasn't entirely right. `local_to_location` ignores the projections so it will allocate the `Local` of the tuple. What we need to do later though is, if the projection is bigger than `0`, we need to update the offset. So let's say the tuple is allocated in `Location::Stack(offset)`, then we need to return `Location::Stack(offset + fieldidx * size)`.

What happens if the tuple was squeezed into a register though? Can we use `al, ax, eax` to access the fields, or do we need to move the tuple to the stack first?

ykcompile/src/lib.rs
ptersilie: Okay, we can fix this in a separate PR.

ykcompile/src/lib.rs
vext01: > What we need to do later though is, if the projection is bigger than 0, we need to update the offset.

place_to_location almost handles that already with:
```
Projection::Field(idx) => Ok(Location::new_mem(RDI.code(), (idx * 8) as i32))
```

`idx * 8` is wrong. I'll fix that. I also think this should panic if there is more than one projection in the chain.

> What happens if the tuple was squeezed into a register though? 

That can't happen for now. `can_live_in_register()` returns false for tuples.

ptersilie: > place_to_location almost handles that

It's similar but not the same. We won't have `RDI` here, but `local_to_location` will return a `Location` to which we need to add the offset of the field.

vext01: I think that's the fix I'm writing. Let me finish it and we can see.

vext01: Give that last push a shot.

ykcompile/src/lib.rs
```diff
@@ -401,6 +442,7 @@ impl<TT> TraceCompiler<TT> {
         // We can only reference Locals living on the stack. So move it there if it doesn't.
         let rloc = match self.place_to_location(p2)? {
             Location::Register(reg) => {
+                #[cfg(target_pointer_width = "64")]
```
vext01, line 445: I was meant to kill this. The whole file is for x86_64.

vext01:
OK, I think we are done here?

ykcompile/src/lib.rs
```diff
@@ -175,18 +175,59 @@ impl<TT> TraceCompiler<TT> {
...
+                    unreachable!(); // Trace inputs are always a tuple.
                 }
             } else {
-                // TODO deal with remaining projections
+                // FIXME deal with remaining projections and chains of projections.
+                let base_ty = self.place_ty(&Place::from(p.local)).clone();
                 match &p.projection[0] {
-                    Projection::Field(0) => self.local_to_location(p.local),
+                    Projection::Field(idx) => match base_ty {
+                        Ty::Struct(sty) => {
```
ptersilie, line 203: The code for `Struct` and `Tuple` looks identical to me. Is there a reason we don't combine the two, e.g. `Struct | Tuple => {`?

ptersilie:
Just one more comment.

ykcompile/src/lib.rs
vext01: I was thinking about this too.

I think there may be scenarios later where we might like the distinction, but I was thinking of making the innards common (in another PR).

ptersilie: Should we combine them for now and copy them when/if we actually have to?

vext01: Ah, you mean this piece of code specifically.

we can't because `tty` and `sty` are different types.

The change I was proposing would make `tty` and `sty` the same type, but it'd be a larger change.

ptersilie: Oh they are different types! That wasn't clear from the code. In that case this is for a future PR.

ptersilie:
Okay, please squash.

vext01:
Squashed. Will now prime CI on the companion.

vext01:
ci primed.

PR #98, opened 2020-08-27
A more robust way of ignoring statements that use thread tracers.

vext01:
I've written a pretty detailed commit message for this as there are some ropey details. Please check it makes sense, as I have a feeling we may be re-visiting this later.

I'm going to submit some follow up PRs after this. I think some other parts of code could now be tidied-up.

### DIFF
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -283,6 +283,16 @@ pub struct Place {
     pub projection: Vec<Projection>,
 }
 
+impl Place {
+    fn push_maybe_defined_locals(&self, locals: &mut Vec<Local>) {
+        locals.push(self.local);
+    }
+
+    fn push_used_locals(&self, locals: &mut Vec<Local>) {
+        locals.push(self.local);
+    }
+}
+
 impl Display for Place {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         if self.projection.is_empty() {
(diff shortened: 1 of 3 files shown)


ykpack/src/types.rs
```diff
@@ -439,6 +449,83 @@ pub enum Statement {
...
+                    a.push_used_locals(&mut ret);
+                }
+            }
+            Statement::Unimplemented(_) => (),
+        }
+
+        ret
+    }
+
+    /// Returns a vector of locals that this SIR statement removes the existing definitions of.
```
ptersilie, line 507: Can we reword this? I can't quite understand the comment. Does this say: "Return the locals that die during the trace?"

ykpack/src/types.rs
```diff
@@ -439,6 +449,83 @@ pub enum Statement {
...
+            | Statement::Assign(..)
+            | Statement::Enter(..)
+            | Statement::Leave
+            | Statement::Call(..)
+            | Statement::Unimplemented(_) => Vec::new(),
+            Statement::StorageDead(lcl) => vec![*lcl],
+        }
+    }
+
+    /// Returns a vector of locals either used of defined by this statement.
```
ptersilie, line 521: s/of/or

ykpack/src/types.rs
```diff
@@ -439,6 +449,83 @@ pub enum Statement {
...
+                }
+            }
+            Statement::Unimplemented(_) => (),
+        }
+
+        ret
+    }
+
+    /// Returns a vector of locals that this SIR statement removes the existing definitions of.
+    /// The locals are considered dead at the start of the following statement.
```
ptersilie, line 508: s/following/next? Otherwise one could interpret this as "the statement following this comment".

yktrace/src/tir.rs
```diff
@@ -554,4 +616,18 @@ mod tests {
...
         assert!(tir_trace.len() > 0);
     }
+
+    #[test]
+    #[should_panic]
+    fn use_undefined_var() {
+        let outside_var = 100;
+
+        #[cfg(tracermode = "sw")]
+        let tracer = start_tracing(Some(TracingKind::SoftwareTracing));
```
ptersilie, line 626: Why are we invoking the software tracer here?

ptersilie (commented):
A few typos/comments. There's also some typos in the commit message:
s/then we can check/so we can check
s/StoragLive/StorageLive

ykpack/src/types.rs
```diff
@@ -439,6 +449,83 @@ pub enum Statement {
     Unimplemented(String),
 }
 
+impl Statement {
+    /// Returns a vector of locals that this SIR statement *may* define.
+    /// Whether or not the local is actually defined depends upon if this is the first write into
```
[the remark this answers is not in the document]
vext01: agreed.

ykpack/src/types.rs
vext01: Actually we can kill the function. We don't insert StorageDead until after we call this, so it is never actually used.

yktrace/src/tir.rs
vext01: It would only call the software tracer if we pass `-C tracer=sw` otherwise the line below is used instead.

It's an idiom I copied from a nearby test.

ptersilie: I thought it was unnecessary since software tracing is currently dead anyway. But we can leave it in if you like.

vext01:
That should address the comments.

ptersilie:
Please squash.

vext01:
splat.

ptersilie:
Think you forgot the typos in the commit message.

vext01:
oops. fixed.

PR #101, opened 2020-08-27
[3/3] Move trace input to yktrace

bjorn3:
Depends on https://github.com/softdevteam/ykrustc/pull/121

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -1115,14 +1115,13 @@ impl<TT> TraceCompiler<TT> {
 mod tests {
     use super::{CompileError, HashMap, Local, Location, TraceCompiler};
     use crate::stack_builder::StackBuilder;
-    use core::yk::trace_inputs;
     use dynasmrt::{x64::Rq::*, Register};
     use fm::FMBuilder;
     use libc::{abs, c_void, getuid};
     use regex::Regex;
     use std::marker::PhantomData;
     use yktrace::tir::TirTrace;
-    use yktrace::{start_tracing, TracingKind};
+    use yktrace::{start_tracing, trace_inputs, TracingKind};
 
     extern "C" {
         fn add6(a: u64, b: u64, c: u64, d: u64, e: u64, f: u64) -> u64;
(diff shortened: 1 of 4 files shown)


vext01:
This looks good to me.

Please squash.

bjorn3:
Should I keep the second and third commit separate?

vext01:
It's up to you.

(We used to try and ensure every commit compiled so that we could bisect, but in reality it's very hard due to the dependency on very specific versions of ykrustc)

bjorn3:
All commits should work independent of each other and except for ab7e4eb independent of the ykrustc changes.

vext01:
Well, if you can make it so that every commit works, then that's great, but don't obsess about it if it's too annoying :)

bjorn3:
I try to keep the history clean for myself. I often have changes applied locally that shouldn't be committed (like making it compile without ykrustc in a hackish in this case), so creating independent commits for independent changes is not much more effort.

vext01:
So no squashing required, or?

bjorn3:
I squashed everything that makes sense for this PR. It will likely need a cycle breaker though.

vext01:
Yes.

In your ykrustc PR, edit the top level `Cargo.toml` and override the ykpack dependency to this repo and branch, then do a `bors try`.

Our CI caches the last successful build of ykrustc and uses it for subsequent builds of this repo, so we can then merge this.

Then once that's done, we revert the dep patching in `ykrustc` and update the lock file, before merging that.

 It's not pretty, but we've not found a better way...

bjorn3:
Wait, does the ykrustc test suite run the tests of this repo? If not then the cycle breaker should not be necessary.

vext01:
Currently the ykrustc CI doesn't run the yk tests.

vext01:
> This branch is out-of-date with the base branch

Can you rebase and check it doesn't break anything?

bjorn3:
Rebased and tested.

vext01:
You probably didn't see that because it's in hardware tracer specific code.

bjorn3:
I accidentally edited the hwtracer checkout in ~/.cargo/git/checkouts`. I dropped 9cd766d, so it should now work.

PR #104, opened 2020-08-28
Refactor projection resolution.

ptersilie:
This PR refactors the code that deals with projections and prepares us for being able to deal with chains of projections soon (we'll need types of projections for that though).

A nice side-effect of these changes is that `Deref` has moved out of `Location` which reduces the branches we have to deal with when compiling operations. However, we pay for this with extra instructions in the compiled trace:  since projections are resolved independently to the operation they are used in, we often store the result of the projection in a temporary when it could have been stored directly in the target location. However, we can fix this by passing the target location to `resolve_projection` in the future. For now, I'm happy to take the performance hit if it allows us to reason about this stuff more easily.

I'd like to squash these commits into one before merging. I have only left them intact as I think it makes this PR easier to review.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -102,7 +102,6 @@ pub struct RegAndOffset {
 pub enum Location {
     Register(u8),
     Mem(RegAndOffset),
-    Deref(Box<Location>),
     NotLive,
 }
 
@@ -122,6 +121,12 @@ impl Location {
     }
 }
 
+enum RegAlloc {
+    Local(Local),
+    Temp,
+    Free,
+}
+
 use std::marker::PhantomData;
 use ykpack::LocalDecl;
 
@@ -130,7 +135,7 @@ pub struct TraceCompiler<TT> {
     /// The dynasm assembler which will do all of the heavy lifting of the assembly.
     asm: dynasmrt::x64::Assembler,
     /// Stores the content of each register.
-    register_content_map: HashMap<u8, Option<Local>>,
+    register_content_map: HashMap<u8, RegAlloc>,
     /// Maps trace locals to their location (register, stack).
     variable_location_map: HashMap<Local, Location>,
     /// Local referencing the input arguments to the trace.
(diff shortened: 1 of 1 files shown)


ykcompile/src/lib.rs
```diff
@@ -263,22 +263,88 @@ impl<TT> TraceCompiler<TT> {
...
             Some(Location::NotLive) => unreachable!(),
             None => unreachable!("freeing unallocated register"),
         }
         self.variable_location_map.insert(*local, Location::NotLive);
         Ok(())
     }
 
+    /// Returns whether the register content map contains any temporaries. This is used as a sanity
+    /// check at the end of a trace to make sure we haven't forgotten to free temporaries at the
+    /// end of an instruction.
```
ptersilie, line 337: This should be `operation` not `instruction`.

vext01:
I'm using this to test a buildbot change.

bors try

vext01:
bors try

vext01:
bors try-

ykcompile/src/lib.rs
```diff
@@ -122,6 +121,12 @@ impl Location {
     }
 }
 
+enum RegisterVal {
```
vext01, line 124: Total bikeshed, but I think the name of this could be better. `Val` suggests it expresses a value, but it actually expresses a register allocation decision.

How about `RegAlloc`?

ykcompile/src/lib.rs
```diff
@@ -236,14 +257,12 @@ impl<TT> TraceCompiler<TT> {
...
                 // Find a free register to store this local.
-                let loc = if let Some(reg) = self.register_content_map.iter().find_map(|(k, v)| {
-                    if v == &None {
-                        Some(*k)
-                    } else {
-                        None
-                    }
-                }) {
-                    self.register_content_map.insert(reg, Some(l));
+                let loc = if let Some(reg) =
```
vext01: I wonder if we should factor this out into a method now? It'd make it easier to decouple the allocator later?

ykcompile/src/lib.rs
```diff
@@ -383,49 +460,45 @@ impl<TT> TraceCompiler<TT> {
...
+        // Free temporary if one was created.
+        self.free_temporary(rloc);
         Ok(())
     }
 
     fn mov_place_ref(&mut self, p1: &Place, p2: &Place) -> Result<(), CompileError> {
         let lloc = self.place_to_location(p1)?;
+
+        // Deal with the special case `&*`, i.e. referencing a `Deref` on a reference just returns
+        // the reference.
```
vext01, line 464: Is that still true in light of custom `Deref` impls? Do we need a fixme?

ykcompile/src/lib.rs
```diff
@@ -437,13 +510,13 @@ impl<TT> TraceCompiler<TT> {
                 // This Local lives now on the stack...
                 self.variable_location_map.insert(p2.local, loc.clone());
                 // ...so we can free its old register.
-                self.register_content_map.insert(reg, None);
+                self.register_content_map.insert(reg, RegisterVal::Free);
```
vext01, line 502: Can this not go through the freeing method I saw before?

ykcompile/src/lib.rs
```diff
@@ -455,18 +528,9 @@ impl<TT> TraceCompiler<TT> {
...
-                Location::Register(reg2) => {
-                    dynasm!(self.asm
-                        ; mov Rq(reg1), Rq(reg2)
-                    );
-                }
-                _ => todo!(),
-            },
             (_, _) => todo!(),
         };
+        self.free_temporary(rloc);
```
vext01, line 522: I wonder if the localities of these can be improved? It's not immediately obvious to a new reader where the corresponding temp is allocated.

Did you experiment with something that has a `Drop` implementation to automatically free?

ykcompile/src/lib.rs
bjorn3: Custom `Deref` impls are desugared when creating MIR.

vext01: Do they become calls?

ykcompile/src/lib.rs
vext01: I see this was done in a later commit, so ignore this.

ykcompile/src/lib.rs
```diff
@@ -282,15 +276,26 @@ impl<TT> TraceCompiler<TT> {
...
 
+    /// Returns a free register or `None` if all registers are occupied.
+    fn get_free_register(&self) -> Option<u8> {
+        self.register_content_map.iter().find_map(|(k, v)| match v {
+            RegisterVal::Free => Some(*k),
+            _ => None,
+        })
+    }
+
+    /// Spill a local to the stack and return its location.
```
vext01, line 274: Should this also update the mapping from locals to locations?

ykcompile/src/lib.rs
ltratt: I tend to agree that `RegisterVal` is confusing (I misread it at first glance), not that I fully understand all of this anyway!

ykcompile/src/lib.rs
```diff
@@ -169,51 +169,56 @@ impl<TT> TraceCompiler<TT> {
...
-                todo!("chains of projections");
-            }
+            self.resolve_projection(p)
+        } else {
+            self.local_to_location(p.local)
+        }
+    }
 
-            // FIXME deal with remaining projections and chains of projections.
+    /// Takes a Place, resolves all projections, and returns a Location containting the result.
```
vext01, line 178: Backtick `Location`?

ykcompile/src/lib.rs
```diff
@@ -169,51 +169,56 @@ impl<TT> TraceCompiler<TT> {
...
+    fn resolve_projection(&mut self, p: &Place) -> Result<Location, CompileError> {
+        let mut curloc = self.local_to_location(p.local)?;
+        if p.projection.len() > 1 {
+            todo!("Deal with remaining projections");
+        }
+        for proj in &p.projection {
```
vext01, line 184: Can we deal with chains yet? Is this loop just in preparation for when we can?

ykcompile/src/lib.rs
```diff
@@ -169,51 +169,56 @@ impl<TT> TraceCompiler<TT> {
...
-                                _ => unreachable!("{:?}", loc),
+                                _ => unreachable!("{:?}", curloc),
                             }
                         }
                     }
                     _ => todo!("{:?}", base_ty),
                 },
                 Projection::Deref => {
-                    let loc = self.local_to_location(p.local)?;
+                    // FIXME Dereferencing a reference to a copyable struct/tuple copies it to the
```
vext01, line 208: I take it we don't yet have a way to know if something is `Copy`?

ykcompile/src/lib.rs
```diff
@@ -243,7 +233,11 @@ impl<TT> TraceCompiler<TT> {
     /// Given a local, returns the register allocation for it, or, if there is no allocation yet,
     /// performs one.
     fn local_to_location(&mut self, l: Local) -> Result<Location, CompileError> {
-        if let Some(location) = self.variable_location_map.get(&l) {
+        if Some(l) == self.trace_inputs_local {
+            // If the Local references trace_inputs return its location on the stack, which is
```
vext01, line 237: Backtick `trace_inputs()` and add parentheses?

ykcompile/src/lib.rs
bjorn3: Yes

vext01:
This looks really good. I'm particularly happy to see `Deref` removed from `Location`.

Some comments and questions.

ykcompile/src/lib.rs
bjorn3:
```rust
fn main() {
    &*Vec::<()>::new();
}
```

```rust
// WARNING: This output format is intended for human consumers only
// and is subject to change without notice. Knock yourself out.
fn main() -> () {
    let mut _0: ();
    let _1: &[()];
    let mut _2: &std::vec::Vec<()>;
    let _3: std::vec::Vec<()>;

    bb0: {
        StorageLive(_1);
        StorageLive(_2);
        StorageLive(_3);
        _3 = const std::vec::Vec::<()>::new() -> bb2;
    }

    bb1 (cleanup): {
        resume;
    }

    bb2: {
        _2 = &_3;
        _1 = const <std::vec::Vec<()> as std::ops::Deref>::deref(move _2) -> [return: bb3, unwind: bb4];
    }

    bb3: {
        StorageDead(_2);
        drop(_3) -> bb5;
    }

    bb4 (cleanup): {
        drop(_3) -> bb1;
    }

    bb5: {
        StorageDead(_3);
        StorageDead(_1);
        _0 = const ();
        return;
    }
}
```

ptersilie: > Is that still true in light of custom `Deref` impls? Do we need a fixme?

Note that I check that the place `p2` is a reference. Dereferencing a reference apparently does not trigger custom `Deref` operations.

> Custom Deref impls are desugared when creating MIR.

That's a very useful information that hopefully makes our lives easier.

ykcompile/src/lib.rs
ptersilie: No problem. I have no attachment to that name, `RegAlloc` works for me.

ykcompile/src/lib.rs
ptersilie: Not quite. `free_register()` also sets the `variable_location_map` to `NotLive`, which we don't want since the local now lives on the stack.

ykcompile/src/lib.rs
ptersilie: I have not. I didn't want to overcomplicate this PR. Will we have access to the `register_content_map` if we add a `Drop` implementation to `Location`?

ykcompile/src/lib.rs
ptersilie: I guess it could. I left it this way due to the way it's used in https://github.com/softdevteam/yk/pull/104/files/6b48c0933870c5148cfd3a8d3c2416687cb50ef6..43ad10f19a2ac4eeb8419734c38affc40af7d86e#diff-162e1dfdca14767450c768ad342d54f2R265. So this would need some more refactoring.

vext01: For now let's add to the comment saying that it doesn't actually update the mapping?

ykcompile/src/lib.rs
ptersilie: > This PR refactors the code that deals with projections and prepares us for being able to deal with chains of projections soon (we'll need types of projections for that though).

I initially planned for this PR to deal with chains of projections, but then decided to keep this a refactoring PR that doesn't add new functionality.

ykcompile/src/lib.rs
ptersilie: I don't think so, but I haven't looked into it much, yet. This is for a future PR.

ykcompile/src/lib.rs
bjorn3: `Drop` with side-effects would be very confusing I think.

ptersilie: While it's not great that it isn't clear where temporaries are allocated, I agree with @bjorn3 that side-effects on `Drop` would be worse.

vext01: So `free_temproary` only sometimes frees a temp, right? (i've lost the definition).

vext01: Yes, it seems passing a location which is not a temp reg is also allowed.

Should this be the case (should it panic?)? If so, should we rename the function to be more indicative of the behaviour? `maybe_free_temp()`?

ptersilie: Yes. You give it a `Local` and it's been assigned to a temporary that temporary is freed. Otherwise it does nothing. It's not great, so I'll happily take any suggestions on improving this.

ykcompile/src/lib.rs
vext01: ok

ykcompile/src/lib.rs
ptersilie: We can't `panic` since we will pass it a non temporary a lot. `maybe_free_temp` sounds like it only sometimes frees temporaries. How about something like`free_if_temp`?

vext01: I think the semantics are fine, just the name could be clearer.

vext01: `free_if_temp` or `maybe_free_temp` work for me.

Another option might be to implement a `free_temps()` at the statement level and call it after the codegen of each statement. It means we'd only need to call it in one place, but we'd have less control over freeing and would probably leave temps around longer than necessary.

ykcompile/src/lib.rs
ptersilie: Happy for me to fix this after squashing things down, since this will introduce conflicts in each commit.

ykcompile/src/lib.rs
ptersilie: We don't have access to `rloc` at the statement level, so we'd have to iterate over all allocated locals after each statement to free the temps.

ptersilie:
Addressed all comments, except the `RegisterVal` one, since that's easier to do after squashing.

ykcompile/src/lib.rs
```diff
@@ -234,7 +234,7 @@ impl<TT> TraceCompiler<TT> {
     /// performs one.
     fn local_to_location(&mut self, l: Local) -> Result<Location, CompileError> {
         if Some(l) == self.trace_inputs_local {
-            // If the Local references trace_inputs return its location on the stack, which is
+            // If the Local references `trace_inputs` return its location on the stack, which is
```
vext01, line 237: and `Local` :P

vext01:
One comment, squash when fixed.

ykcompile/src/lib.rs
ptersilie: Done.

ptersilie:
Squashed and renamed `RegisterVal`. So requires one more squash.

vext01:
LGTM. Please squash.

ptersilie:
Squashed!

ptersilie:
Hold on, it needs rebasing.

ptersilie:
Rebased.

PR #105, opened 2020-08-29
Add viewer for the SIR in executables

bjorn3:
This can be useful while debugging.

I have also added a way to generate and show TIR for a trace specified in a format like:

```
_ZN21mini_core_hello_world20yk_swt_start_tracing17hd35172730ed8f9e0E 2
_ZN21mini_core_hello_world5start17hc857f9701ee10e6bE 2
_ZN21mini_core_hello_world5start17hc857f9701ee10e6bE 4
_ZN21mini_core_hello_world5start17hc857f9701ee10e6bE 5
_ZN21mini_core_hello_world4main17h4614255fce60abb6E 1
_ZN21mini_core_hello_world8take_f3217h7c2802eb13dcfab7E 1
_ZN21mini_core_hello_world4main17h4614255fce60abb6E 2
_ZN21mini_core_hello_world5start17hc857f9701ee10e6bE 6
_ZN63_$LT$$LP$$RP$$u20$as$u20$mini_core_hello_world..Termination$GT$6report17hcfb2e8675d965aaeE 1
_ZN21mini_core_hello_world5start17hc857f9701ee10e6bE 7
_ZN21mini_core_hello_world19yk_swt_stop_tracing17h5af4cd5c43ee1fdeE 1
```

I have not been able to test that functionality yet, as cg_clif currently doesn't generate enough sir for it to work without errors.

PR #108, opened 2020-08-29
Fix the yktrace half of the software tracer

bjorn3:
Includes #105

### DIFF
--- a/yktrace/src/swt.rs
+++ b/yktrace/src/swt.rs
@@ -2,10 +2,9 @@
 
 use super::{SirTrace, ThreadTracer, ThreadTracerImpl};
 use crate::{errors::InvalidTraceError, sir::SIR, SirLoc};
-use ykpack::Local;
-use core::yk::{swt, SirLoc as CoreSirLoc};
 use libc;
 use std::convert::TryFrom;
+use ykpack::Local;
 
 /// A trace collected via software tracing.
 /// Since the trace is a heap-allocated C buffer, we represent it as a pointer and a length.
@@ -18,13 +17,19 @@ impl SWTSirTrace {
     /// Create a SWTSirTrace from a raw buffer and (element) length.
     ///
     /// `buf` must have been allocated by malloc(3); this function will free(2) it.
-    fn from_buf(buf: *const CoreSirLoc, len: usize) -> Self {
+    fn from_buf(buf: *const SwtLoc, len: usize) -> Self {
         // When we make a SWTSirTrace, we convert all of the locations from core::SirLoc up to
         // crate::SirLoc and store them in self so that we can hand out references via raw_loc().
         let locs = (0..len)
             .map(|idx| {
                 let idx = isize::try_from(idx).unwrap();
-                SirLoc::from(unsafe { &*buf.offset(idx) })
+                let swt_loc = unsafe { &*buf.offset(idx) };
+                let symbol_name = unsafe { std::ffi::CStr::from_ptr(swt_loc.symbol_name) };
+                SirLoc {
+                    symbol_name: symbol_name.to_str().unwrap().to_string(),
+                    bb_idx: swt_loc.bb_idx,
+                    addr: None
+                }
             })
             .collect();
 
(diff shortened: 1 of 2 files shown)


yktrace/src/swt.rs
```diff
@@ -55,21 +60,139 @@ struct SWTThreadTracer;
...
+        std::intrinsics::abort();
+    }
+
+    TRACE_BUF_CAP = TL_TRACE_INIT_CAP;
+    TRACING = true;
+}
+
+/// Record a location into the trace buffer if tracing is enabled on the current thread.
+#[do_not_trace]
+#[no_mangle]
```
vext01, line 130: Remind me why this is `no_mangle`

yktrace/src/swt.rs
```diff
@@ -55,21 +60,139 @@ struct SWTThreadTracer;
...
+        if TRACE_BUF as usize == 0 {
+            TRACING = false;
+            return;
+        }
+
+        TRACE_BUF_CAP = new_cap;
+    }
+
+    *((TRACE_BUF as usize + TRACE_BUF_LEN * std::intrinsics::size_of::<SwtLoc>()) as *mut SwtLoc) = SwtLoc { symbol_name, bb_idx };
+    TRACE_BUF_LEN = TRACE_BUF_LEN + 1;
```
vext01, line 194: Use +=?

yktrace/src/swt.rs
```diff
@@ -55,21 +60,139 @@ struct SWTThreadTracer;
...
+#[thread_local]
+static mut TRACE_BUF_CAP: usize = 0;
+
+/// true = we are tracing, false = we are not tracing or an error occurred.
+#[thread_local]
+static mut TRACING: bool = false;
+
+// =======================
+// The following code may not call any non `#[do_not_trace]` function to prevent an infinite
+// recursion of `__yk_swt_rec_loc` calling a function calling `__yk_swt_rec_loc`.
```
vext01, line 111: Shall we mention that we manage our own trace buffer rather than using a `Vec`, otherwise the whole of `Vec` would need to be `no_trace`?

yktrace/src/swt.rs
bjorn3: So that cg_clif can find this function by calling `__yk_swt_rec_loc`.

yktrace/src/swt.rs
bjorn3: Yes, I initially didn't use `+=` as I tested it first with cg_clif's `mini_core_hello_world` test, which doesn't have access to `+=` because `libcore` is not imported.

yktrace/src/swt.rs
```diff
@@ -121,36 +122,68 @@ unsafe fn yk_swt_start_tracing_impl() {
...
 
     TRACE_BUF_CAP = TL_TRACE_INIT_CAP;
-    TRACING = true;
+    __YK_SWT_ACTIVE = true;
 }
 
 /// Record a location into the trace buffer if tracing is enabled on the current thread.
 #[do_not_trace]
 #[no_mangle]
+#[cfg(not(all(target_arch = "x86_64", target_os = "linux")))]
```
vext01, line 131: Let's add comments with a few words about the "fast path" and "slow path" functions.

yktrace/src/swt.rs
```diff
@@ -121,36 +122,68 @@ unsafe fn yk_swt_start_tracing_impl() {
...
+}
+
+#[cfg(all(target_arch = "x86_64", target_os = "linux"))]
+global_asm!(r#"
+.globl __yk_swt_rec_loc
+.type __yk_swt_rec_loc,@function
+.section .text.__yk_swt_rec_loc,"ax",@progbits
+__yk_swt_rec_loc:
+    push %rdi
+    data16 lea __YK_SWT_ACTIVE@tlsgd(%rip),%rdi
```
vext01, line 147: What's the purpose of `data16` here?

vext01: Is the `@tlsgd` suffix guaranteed to be fixed?

yktrace/src/swt.rs
```diff
@@ -121,36 +122,68 @@ unsafe fn yk_swt_start_tracing_impl() {
...
     }
 
+    __yk_swt_rec_loc_impl(symbol_name, bb_idx);
+}
+
+#[cfg(all(target_arch = "x86_64", target_os = "linux"))]
+global_asm!(r#"
+.globl __yk_swt_rec_loc
+.type __yk_swt_rec_loc,@function
+.section .text.__yk_swt_rec_loc,"ax",@progbits
```
vext01, line 144: "ax" is segment flags here? executable and (what's `a`)?

yktrace/src/swt.rs
vext01: Ah, the `data16` is like an `.asciz` in gnu asm?

vext01: i.e. inlining data.

yktrace/src/swt.rs
```diff
@@ -121,36 +122,68 @@ unsafe fn yk_swt_start_tracing_impl() {
...
+__yk_swt_rec_loc:
+    push %rdi
+    data16 lea __YK_SWT_ACTIVE@tlsgd(%rip),%rdi
+.byte 0x66 # data16
+.byte 0x66 # data16
+.byte 0x48 # rex
+    callq __tls_get_addr-4
```
vext01, line 151: I think you are getting a handle on the thread local from global code, right? Please add comments.

yktrace/src/swt.rs
```diff
@@ -121,36 +122,68 @@ unsafe fn yk_swt_start_tracing_impl() {
...
+    rex    movzbl (%rax),%eax
```
vext01, line 152: Is the rex prefix needed?

yktrace/src/swt.rs
bjorn3: The block of assembly must result in an exact combination of bytes for the linker to be able to rewrite the expensive `__tls_get_addr` call to a sequence using `%fs:0` and a fixed offset. If you fail to use this exact block of assembly you get a linker error.

yktrace/src/swt.rs
bjorn3: See https://github.com/softdevteam/yk/pull/108#discussion_r481059219

yktrace/src/swt.rs
vext01: Sorry for being a dummy, but can you add a comment explaining about why `tls_get_addr` is expensive?

yktrace/src/swt.rs
bjorn3: Probably not. I just copied the sequence from cranelift, which has the tendency to emit redundant rex prefixes.

yktrace/src/swt.rs
bjorn3: According to https://ftp.gnu.org/old-gnu/Manuals/gas-2.9.1/html_chapter/as_7.html#SEC119 it means "allocatable".

vext01:
I appreciate that this won't be fun, but in this repo we've used intel syntax for our asm.

Can we use `.intel_syntax noprefix`?

Otherwise looks good (if a little scary :P )

bjorn3:
Sure, I can switch to the intel syntax. The at&t one is a bit confusing for me too. I just wanted to stick with the default.

yktrace/src/swt.rs
vext01: OK.

yktrace/src/swt.rs
bjorn3: `__tls_get_addr` walks a linked list containing the tls data for every object linked into the current process.

yktrace/src/swt.rs
vext01: Not much on the internet as to what that really means, but if it works, OK.

yktrace/src/swt.rs
vext01: I wonder if we should try without? The simpler the better.

vext01:
> The at&t one is a bit confusing for me too.

For us all then :) Since I've always used the intel manual as my tome, I find at&t confusing.

yktrace/src/swt.rs
vext01: Gotcha. That's a detail that should appear in a comment :)

yktrace/src/swt.rs
bjorn3: Allocatable means that it will be mapped into memory when loading.

yktrace/src/swt.rs
bjorn3: It works without.

yktrace/src/swt.rs
```diff
@@ -90,6 +90,9 @@ const TL_TRACE_INIT_CAP: usize = 1024;
 const TL_TRACE_REALLOC_CAP: usize = 1024;
 
 /// The trace buffer.
+///
+/// `calloc` and `free` are directly used instead of `Vec` to avoid calling into traced function
```
vext01, line 94: into traced function**s**

yktrace/src/swt.rs
```diff
@@ -90,6 +90,9 @@ const TL_TRACE_INIT_CAP: usize = 1024;
...
+/// from `__yk_swt_rec_loc`. Otherwise there would be an infinite recursion.
```
vext01, line 95: Kill "an".

yktrace/src/swt.rs
```diff
@@ -126,6 +129,10 @@ unsafe fn yk_swt_start_tracing_impl() {
 }
 
 /// Record a location into the trace buffer if tracing is enabled on the current thread.
+///
+/// This function is separate from `__yk_swt_rec_loc_impl` to prevent the fast path of tracing
```
vext01, line 133: I think we could simplify this to "This function calls a fast or a slow implementation depending on the platform".

yktrace/src/swt.rs
```diff
@@ -137,19 +144,28 @@ unsafe extern "C" fn __yk_swt_rec_loc(symbol_name: *const i8, bb_idx: u32) {
     __yk_swt_rec_loc_impl(symbol_name, bb_idx);
 }
 
+// Cranelift produces suboptimal code for the fast path. When running on x86_64 and using ELF TLS
```
vext01, line 147: This implementation *is* the fast path, isn't it?

In hindsight, I think something like: "An optimised trace recorder which avoids the cost of `__tls_get_addr`."

bjorn3: This is not about avoiding the cost of `__tls_get_addr`. The linker already does that when possible. This is about avoiding the cost of unnecessary spilling of registers when tracing is disabled.

yktrace/src/swt.rs
```diff
@@ -55,21 +60,190 @@ struct SWTThreadTracer;
...
+
+    movzx eax, BYTE PTR [rax]
+    test eax, eax
+    jne __yk_swt_rec_loc_active
+    pop rdi
+    ret
+__yk_swt_rec_loc_active:
+    pop rdi
+    call __yk_swt_rec_loc_impl
+    ret
```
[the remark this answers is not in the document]
vext01: Perhaps another PR?

yktrace/src/swt.rs
vext01: So would something like this be accurate?

> Cranelift produces suboptimal code for the Rust version of `__yk_swt_rec_loc` (redundant spills). When running...

yktrace/src/swt.rs
bjorn3: That is not quite the reason. Does the new comment make it clearer?

vext01:
Great! Please squash.

bjorn3:
Splat

vext01:
Just formatting.

bjorn3:
Fixed and squashed

bjorn3:
bors r-

```
relocation R_X86_64_PC32 against symbol `__tls_get_addr@@GLIBC_2.3' can not be used when making a PIE object; recompile with -fPIC
```

yktrace/src/swt.rs
```diff
@@ -167,7 +167,7 @@ __yk_swt_rec_loc:
 .byte 0x66 # data16
 .byte 0x66 # data16
 .byte 0x48 # rex
-    call __tls_get_addr-4
+    call __tls_get_addr@plt-4
```
vext01, line 170: Relocation is pure magic from my POV :P

vext01:
Please squash.

bjorn3:
Splat

PR #110, opened 2020-09-01
Try to remove unused definitions.

vext01:
This is not perfect. It works well for simple locals, but consider `x.0 = 0; x.1 = 1` and then x is not used after. The first operation will be seen to define `x`, the second will be seen as a use of `x`, and thus neither of these statements will be removed.

Still, it's better than nothing.

### DIFF
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -520,6 +520,14 @@ impl Statement {
         }
         ret
     }
+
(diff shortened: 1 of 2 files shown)


ptersilie:
Looks good. Just needs a rebased.

vext01:
Rebased.

PR #120, opened 2020-09-18
Select tracer kind

### LINKED ISSUE #119: Tracer choice should be automatic from -Ctracermode
I was reviewing some yk code and realised that it's kind of odd that we offer the illusion of a runtime choice of tracer. We don't because you have to compile with the right `-C tracer=`.

I wonder if we should remove the option to `start_tracing` as it's kind of misleading.

vext01:
Working towards a merge point design, this PR allows us to specify a tracing kind via the builder.

While we are here, fix #119.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -10,6 +10,7 @@ use std::{
     },
     thread::{self, JoinHandle},
 };
+use yktrace::TracingKind;
 
 pub type HotThreshold = u32;
 const DEFAULT_HOT_THRESHOLD: HotThreshold = 50;
@@ -49,6 +50,8 @@ impl Location {
 /// Configure a meta-tracer. Note that a process can only have one meta-tracer active at one point.
 pub struct MTBuilder {
     hot_threshold: HotThreshold,
+    /// The kind of tracer to use.
+    tracing_kind: TracingKind,
 }
 
 impl MTBuilder {
@@ -56,18 +59,25 @@ impl MTBuilder {
     pub fn new() -> Self {
         Self {
             hot_threshold: DEFAULT_HOT_THRESHOLD,
+            tracing_kind: TracingKind::default(),
         }
     }
 
     /// Consume the `MTBuilder` and create a meta-tracer, returning the
     /// [`MTThread`](struct.MTThread.html) representing the current thread.
     pub fn init(self) -> MTThread {
-        MTInner::init(self.hot_threshold)
+        MTInner::init(self.hot_threshold, self.tracing_kind)
     }
 
     /// Change this meta-tracer builder's `hot_threshold` value.
     pub fn hot_threshold(mut self, hot_threshold: HotThreshold) -> Self {
         self.hot_threshold = hot_threshold;
+        self
+    }
+
+    /// Select the kind of tracing to use.
+    pub fn tracing_kind(mut self, tracing_kind: TracingKind) -> Self {
+        self.tracing_kind = tracing_kind;
         self
     }
 }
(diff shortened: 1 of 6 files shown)


ykrt/src/mt.rs
```diff
@@ -213,12 +230,19 @@ impl MTThread {
 struct MTThreadInner {
     mt: MT,
     hot_threshold: HotThreshold,
+    #[allow(dead_code)] // not used just yet.
```
ltratt, line 233: I think the comment is a bit redundant?

ltratt:
I'm not sure I totally understand: we appear to be giving the user the illusion they can switch between software/hardware tracing, but I think at the moment we compile with one or the other, so they can't actually select between the two in practise?

yktrace/src/lib.rs
```diff
@@ -26,6 +29,16 @@ pub enum TracingKind {
     HardwareTracing
 }
 
+impl Default for TracingKind {
+    /// Returns the default tracing kind.
+    fn default() -> Self {
+        #[cfg(tracermode = "hw")]
+        return TracingKind::HardwareTracing;
+        #[cfg(tracermode = "sw")]
+        return TracingKind::SoftwareTracing;
```
bjorn3, line 38: Maybe use `compile_error!` with a nice message when `-Ctracer` is not used.

vext01:
> I'm not sure I totally understand: we appear to be giving the user the illusion they can switch between software/hardware tracing, but I think at the moment we compile with one or the other, so they can't actually select between the two in practise?

Indeed. That's what #119 was all about.

I thought we had settled on a plan to keep the existing API, which would allow a runtime selection at a later date, but for now crash if the user chooses a tracer they didn't compile for.

Maybe I misunderstood?

ltratt:
I guess my possible-problem with this PR is that, AIUI, it *always* forces the user to specify what type of tracer they want, which I don't think is desirable. I think we should choose for them, but we can allow them an additional way to override that choice in the (rather unlikely case) that they want something specific.

vext01:
At the `MTBuilder` level, we choose a tracer for the user unless they specify:
https://github.com/softdevteam/yk/pull/120/files#diff-0c76abf7b121c7956d6451de0e4b1f6cR62

ltratt:
Mea culpa! I'm with you now.

ltratt:
OK, so I'm OK with this PR, but I think we need to address @bjorn3's comments before I bors it.

yktrace/src/lib.rs
vext01: Using `compile_error!` seems to break doc tests and I don't seem to be able to fix it:

```
$ RUSTFLAGS="-C tracer=hw" cargo test
...
   Doc-tests yktrace
error: Please compile with `-C tracer=T`, where T is one of 'hw' or 'sw'
  --> /home/vext01/research/yorick/yk/yktrace/src/lib.rs:70:5
   |
70 |     compile_error!("Please compile with `-C tracer=T`, where T is one of 'hw' or 'sw'");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Despite:

```
#[cfg(all(doctest, not(any(tracermode = "hw", tracermode = "sw"))))]                           
compile_error!("Please compile with `-C tracer=T`, where T is one of 'hw' or 'sw'");           
```

Any ideas?

vext01: Gah, I spotted it the moment I posted it.

vext01:
@ltratt try this.

ltratt:
Please squash.

vext01:
Splat.

PR #121, opened 2020-09-21
Control point

vext01:
See the first commit message for the overall gist.

Please ignore the second commit. I will remove this one the libipt issue is closed.

A good starting point is probably [this test](https://github.com/vext01/yk_pv/blob/b06025211859a32febffb213049b81afb7cd1569/ykrt/src/mt.rs#L373-L409).

This is a draft because I have questions:
 - Should `prog` be a trace input? We have to get that inside `interp_step` somehow (without using the closure environment).
 - Laurie expressed distaste of the term "trace inputs", so shall we try to decide on a better name now?
 - What do we think of the overall design?

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -1,8 +1,11 @@
+use std::cmp::{Eq, PartialEq};
 #[cfg(test)]
 use std::time::Duration;
 use std::{
     io,
+    ops::Deref,
     panic::{catch_unwind, resume_unwind, UnwindSafe},
+    ptr,
     rc::Rc,
     sync::{
         atomic::{AtomicBool, AtomicU32, AtomicUsize, Ordering},
@@ -10,7 +13,7 @@ use std::{
     },
     thread::{self, JoinHandle},
 };
-use yktrace::TracingKind;
+use yktrace::{start_tracing, ThreadTracer, TracingKind};
 
 pub type HotThreshold = u32;
 const DEFAULT_HOT_THRESHOLD: HotThreshold = 50;
@@ -25,20 +28,50 @@ const PHASE_COMPILED: u32 = 0b00 << 30;
 const PHASE_TRACING: u32 = 0b01 << 30;
 const PHASE_COUNTING: u32 = 0b10 << 30; // The value specifies the current hot count.
 
-/// A `Location` uniquely identifies a control point position in the end-user's program (and is
-/// used by the `MT` to store data about that location). In other words, every position
-/// that can be a control point also needs to have one `Location` value associated with it, and
-/// that same `Location` value must always be used to identify that control point.
+/// A `Location` is a handle on a unique identifier for control point position in the end-user's
+/// program (and is used by the `MT` to store data about that location). In other words, every
+/// position that can be a control point also needs to have one `Location` value associated with
+/// it. Note however that instances of `Location` may be freely cloned and shared across threads.
 ///
-/// As this may suggest, program positions that can't be control points don't need an associated
-/// `Location`. For interpreters that can't (or don't want) to be as selective, a simple (if
-/// moderately wasteful) mechanism is for every bytecode or AST node to have its own `Location`
-/// (even for bytecodes or nodes that can't be control points).
+/// Program positions that can't be control points don't need an associated `Location`. For
+/// interpreters that can't (or don't want) to be as selective, a simple (if moderately wasteful)
+/// mechanism is for every bytecode or AST node to have its own `Location` (even for bytecodes or
+/// nodes that can't be control points).
+#[derive(Debug, Clone)]
 pub struct Location {
-    pack: AtomicU32,
+    inner: Arc<LocationInner>,
 }
 
 impl Location {
+    pub fn new() -> Self {
+        Self {
+            inner: Arc::new(LocationInner::new()),
+        }
+    }
+}
+
+impl PartialEq for Location {
+    fn eq(&self, other: &Location) -> bool {
+        ptr::eq(Arc::as_ptr(&self.inner), Arc::as_ptr(&other.inner))
+    }
+}
+
+impl Eq for Location {}
+
+impl Deref for Location {
+    type Target = LocationInner;
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+#[derive(Debug)]
+pub struct LocationInner {
+    pack: AtomicU32,
+}
+
+impl LocationInner {
     /// Create a fresh Location suitable for passing to `MT::control_point`.
     pub fn new() -> Self {
         Self {
@@ -190,14 +223,27 @@ impl MTThread {
     }
 
     /// Attempt to execute a compiled trace for location `loc`.
-    pub fn control_point(&self, loc: &Location) {
+    pub fn control_point<'l, S, I>(&mut self, loc: Option<&'l Location>, step_fn: S, inputs: I) -> I
+    where
+        S: Fn(I) -> I,
+    {
+        // If a loop can start at this position then update the location and potentially start/stop
+        // this thread's tracer.
+        if let Some(loc) = loc {
+            self._control_point(loc);
+        }
+
+        step_fn(inputs)
+    }
+
+    pub fn _control_point(&mut self, loc: &Location) {
         // Since we don't hold an explicit lock, updating a Location is tricky: we might read a
         // Location, work out what we'd like to update it to, and try updating it, only to find
         // that another thread interrupted us part way through. We therefore use compare_and_swap
         // to update values, allowing us to detect if we were interrupted. If we were interrupted,
         // we simply retry the whole operation.
         loop {
-            let pack = &loc.pack;
+            let pack = &loc.inner.pack;
             // We need Acquire ordering, as PHASE_COMPILED will need to read information written to
             // external data as a result of the PHASE_TRACING -> PHASE_COMPILED transition.
             let lp = pack.load(Ordering::Acquire);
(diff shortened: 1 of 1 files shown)


ykrt/src/mt.rs
```diff
@@ -305,20 +363,58 @@ mod tests {
...
+        // The program is silly. Do nothing twice, then start again.
+        enum ByteCode {
+            Nop,
+            Restart,
+        }
+        let prog = vec![ByteCode::Nop, ByteCode::Nop, ByteCode::Restart];
+
+        // Suppose the bytecode compiler for this imaginary language knows that the first bytecode
+        // is the only place a loop can start.
+        let locs = vec![Some(Location::new()), None, None];
```
ptersilie, line 386: Does this mean the locations vector is always the size of the input program bytecode?

vext01: It doesn't have to be. The user can store them however they want. It could have been a hashmap, or any other sparse data structure.

(we probably don't want a hashmap, it was just an example)

vext01: @ltratt What kind of data structure would you envisage for a production VM actually?

You'd need some kind of sparse mapping, but you said you want to avoid hashing. What would you use?

ltratt:
> Accept and `Option<&Location>` to allow calling the control point for  program positions which cannot be the start of a loop. For such positions there will be no `Location` so the user should pass `None`.

This seems an unnecessary complexity: surely the user just shouldn't call the function in such cases?

> Add a `*const Location` to each `MTThreadInner` to indicate which location the thread is tracing from. This is necessary because only the thread that started tracing a location should stop tracing (and update the pack) for that location (thread 2 shouldn't stop tracing where thread 1 started -- doing so doesn't make sense and would cause a crash).

I might be missing something, but I don't think this stops the race, because two threads could have the same `*const Location` pointer? However, fortunately, I think that you *might* be worrying over nothing, because the atomic nature of a `Location` means that thread 2 should notice that the `Location` is already `PHASE_TRACING` and not start tracing.

ykrt/src/mt.rs
ltratt: Let's remember that there's no need to have a "locations vector": it's just one way (the simplest, probably; the least efficient, almost certainly) to encode a notion of locations. Ultimately the VM just has to give the meta-tracer a `Location` for every loop (or loop-like thing like a function).

A high-performance VM would probably do things like have a "spare" word after each loop header instruction to store `Location`s in.

vext01:
> This seems an unnecessary complexity: surely the user just shouldn't call the function in such cases?

If you were to not call the control point then the bytecode's handler would not be executed. It's the last thing the control point does.

> I might be missing something, but I don't think this stops the race, because two threads could have the same *const Location pointer?

OK we have a bug there for sure. You might be right about the fix. I'll get back to you.

ltratt:
> If you were to not call the control point then the bytecode's handler would not be executed. It's the last thing the control point does.

Ah, I see. I think that's OK.

ykrt/src/mt.rs
vext01: > A high-performance VM would probably do things like have a "spare" word after each loop header instruction to store Locations in.

Clever.

vext01:
> Ah, I see. I think that's OK.

The other option is to *not* have the bytecode handler passed in to the control point like that, but by doing so we give the user a little more (unnecessary?) freedom with the bazooka they are holding. By having the call inside the control point, we can guarantee that it gets called at the appropriate time.

ltratt:
> The other option is to not have the bytecode handler passed in to the control point like that, but by doing so we give the user a little more (unnecessary?) freedom with the bazooka they are holding. By having the call inside the control point, we can guarantee that it gets called at the appropriate time.

If we weren't to pass it in, how would everything work? I think it's possible that we don't have any choice but to use the current scheme!

vext01:
> If we weren't to pass it in, how would everything work?

The user would have to call the step function immediately after calling the control point. I prefer the API we have where there is no ability for the user to do the wrong thing (either not call the handler, or do it an the wrong time etc.).

ltratt:
>  The user would have to call the step function immediately after calling the control point.

I don't think that's enough: if a trace is present for a location they shouldn't call it.

But, thinking about it, we did once have this design:

```rust
loop {
  if control_point(&location) {
    continue;
  }
}
```

where returning `true` from `control_point` means "a trace was executed so carry on". But I think this is impractical because the trace would have to interfere with the function-containing-the-loop's stack frame to update locals etc.

The advantage the closure-per-iteration approach has is that we're in full control of locals.

vext01:
Yes, you'd need a way to ensure the user only calls the handler if there was no code for that location.

We can make that decision automatically, which is what [this comment](https://github.com/vext01/yk_pv/blob/06bfd89bf1f82dea6c4fe106033d127899e49cca/ykrt/src/mt.rs#L271) is alluding to.

So I think we are on the same page :)

ltratt:
Agreed!

vext01:
> I think that you might be worrying over nothing, because the atomic nature of a Location means that thread 2 should notice that the Location is already PHASE_TRACING and not start tracing.

I think I have a counter example.

Suppose there are two loop headers L1 and L2, and there are two threads T1 and T2. Perhaps the code looks like this:
```
L1:
for ... {
    ...
}

L2:
for ... {
    ...
}
```

T2 has just started tracing L2 and T1 is tracing the final iteration of L1.

T1 falls out of L1 (still tracing) and reaches L2.

The location for L2 indicates `TRACING` (as T1 is tracing it) and therefore T2 stops its tracer **and flips L1's pack to `COMPILED`**. `T1` is now confused, as another thread fiddled with the location it was tracing.

Only T1 should have been allowed to transition T2 from `TRACING` to `COMPILED`.

Is that right, or have I made a mistake?

ltratt:
Each thread has to know which `Location` it started tracing on, otherwise it can't know when it's finished properly.

vext01:
> Each thread has to know which Location it started tracing on, otherwise it can't know when it's finished properly.

Precisely.

I thought you were [objecting to the addition of the addition of `Location` to the `MTThread`](https://github.com/softdevteam/yk/pull/121#issuecomment-696226449)?

ltratt:
> I thought you were objecting to the addition of the addition of Location to the MTThread?

That was because things were muddled up with the multiple-thread stuff. I don't think it's an issue now. *However*, I should note that I think it's UB to turn an `&Location` to a `const *Location` and later compare a `&Location` to that `*const Location`: in theory the `&Location` could have moved. We could force `Location`s to be `Pin`ed, but honestly I don't fully understand the consequences of that. Alternatively we could simply store the `&Location` directly. That will make the lifetimes a bit more complex, but it might be the best trade-off.

vext01:
> I don't think it's an issue now. However, I should note that I think it's UB to turn an &Location to a const *Location and later compare a &Location to that *const Location: in theory the &Location could have moved

Early on we were talking about the pointer address of a location being its ident, and I was working under that assumption. But you are right, we should re-think this.

> We could force Locations to be Pined, but honestly I don't fully understand the consequences of that. Alternatively we could simply store the &Location directly. That will make the lifetimes a bit more complex, but it might be the best trade-off.

I failed to shoehorn that into Rust. Once you've finished scattering your new lifetime all over the place, the Rust compiler can't prove that the locations live long enough for the bytecode handler closure.

ltratt:
> Early on we were talking about the pointer address of a location being its ident

References and pointers in Rust aren't quite the same: the original design was for reference equality. Pointer equality has slightly different rules. Unfortunately a) this stuff is very hard to understand full stop b) it's not properly documented at the moment :(

vext01:
Fair enough.

So where do we stand on this PR? Shall we merge it as a basis for iteration, and you can try some stuff out?

ltratt:
I don't think we can merge the `*const` bit: we need to find a design that's not quite so clearly UB. [Or, alternatively, prove that this design isn't UB.] I'm fine with coming up with an inefficient temporary solution, if that helps.

vext01:
Ah wait, we need the libipt change in first. I'll check the status.

vext01:
> I don't think we can merge the *const bit: we need to find a design that's not quite so clearly UB.

I *don't think* (but you will probably prove me wrong) it's UB. But whether it's UB or not is moot as the code is certainly incorrect if we assume that locations can move: If a `Location` moves, then a thread which is tracing there will not recognise it as the place that should stop the tracer.

> I'm fine with coming up with an inefficient temporary solution, if that helps.

Thinking hat on. Or maybe I should stop wimping out and try to get it working with a proper reference.

About the libipt stuff, we need to wait a little longer:
https://github.com/intel/libipt/issues/73

I do have an (incorrect) hack for hwtracer, which happens to allow the tests to pass, if we are desperate to move on.

ltratt:
> I don't think (but you will probably prove me wrong) it's UB.

I think `Pin` shows that it doesn't work https://doc.rust-lang.org/std/pin/struct.Pin.html because the `&Location` can be moved and then its address will not correspond to the stored `*const Location`.

> About the libipt stuff, we need to wait a little longer:

I'm happy to leave that to your discretion (i.e. if you want to merge the hack or not).

vext01:
Maybe this is a step in the right direction?

I've pushed two commits.

The first shows what happens if you blindly replace `*const Location` with `&Location`. The result? Lifetime errors as mentioned before. Rust is unable to prove that references to `Location`s live long enough for the bytecode closure. In fact, it's fine, as we join the threads before they die, but Rust doesn't know that.

We can't even use a [scoped spawn](https://docs.rs/crossbeam-utils/0.7.2/crossbeam_utils/thread/struct.Scope.html) here, as the locality of our spawns may not be inside the same scope.

So the second commit is an attempt to fix that. Using the same idiom as you (@ltratt) used for `MT` and `MTInner`, make `Location` a clonable handle on a `LocationInner`. Then we can store a cloned `Location, and never any references, thus solving the lifetimes problem.

Comparison of `Locations` is done by comparing the pointers you get out of `Arc::as_ptr()` (I've done that in a `PartialEq` implementation).

The docs for `Location` need to be better, but my brain is fried. This is about the fourth thing I've tried today 😵 💫 💥 

Does this at least look like we are going in the right direction? FWIW tests pass.

ykrt/src/mt.rs
```diff
@@ -25,20 +28,50 @@ const PHASE_COMPILED: u32 = 0b00 << 30;
...
-/// `Location`. For interpreters that can't (or don't want) to be as selective, a simple (if
-/// moderately wasteful) mechanism is for every bytecode or AST node to have its own `Location`
-/// (even for bytecodes or nodes that can't be control points).
+#[derive(Debug, Clone)]
 pub struct Location {
-    pack: AtomicU32,
+    inner: Arc<LocationInner>,
 }
 
+/// A `Location` is a handle on a unique identifier for control point position in the end-user's
```
vext01, line 36: Hah. This docstring is even in the wrong place. What a plonker. Sorry.

ykrt/src/mt.rs
```diff
@@ -346,4 +471,20 @@ mod tests {
...
+    #[test]
+    fn clone_loc_ptr_ident() {
+        let l = Location::new();
+        for _ in 1..1000 {
+            let c1 = l.clone();
+            let c2 = l.clone();
+            let c3 = l.clone();
+            let c4 = l.clone();
+
+            assert!(ptr::eq(Arc::as_ptr(&l.inner), Arc::as_ptr(&c1.inner)));
```
vext01, line 484: This can be simplified.

ltratt:
As soon as you `Clone` a `Location`, you've broken the relationship between a `Location` and its address. In other words, this pretty much destroys the idea of a `Location` I'm afraid! Put another way, a `&` is really a pointer (plus some other rules); if you clone the underlying thing, you've got two pointers to two different objects, so they won't compare equal for pointer equality.

vext01:
It's the address of the data inside the `Arc` that now forms the identity of a location. [All clones share the same identity](https://github.com/softdevteam/yk/pull/121/files/e07140d2df63c442ba8e646682f80279794cb79f#diff-0c76abf7b121c7956d6451de0e4b1f6cR475-R489)

ltratt:
Oh, I see. That's incredibly inefficient, but I did say that inefficient is OK! I think we should merge this, then I should take a look to see what we can do to regain the efficiency we need. Sound like a plan?

vext01:
It's only inefficient if you clone locations a lot, which we shouldn't.

But yeah, we can merge and you can try!

ltratt:
`Arc`s aren't cheap to create. Anyway, let's not worry about that for now. Please squash.

vext01:
Ah, we can't merge without the libipt fix or hack.

vext01:
Squashed, and can be merged after the hack goes in (https://github.com/softdevteam/hwtracer/pull/90).

PR #123, opened 2020-09-23
Implement basic array indexing.

ptersilie:
This implementation makes use of some of assumptions that are commented
by FIXMEs. Solving everything in one go would have made this PR
unneccessarily large and complicated. In particular two assumptions we
will have to deal with in the near future:

1) For now a `Deref` on values that don't fit inside a register is
always followed by another projection like `Index` or `Field`. This
means that dereferencing a copyable array/struct won't currently copy it
to the stack.

2) An `Index` projection is only applied to `&Array`. Because of 1) and
bbecause we haven't implemented array construction yet, this is a valid
assumption. Implementing indexing on `Array` means we probably have to
restructure that part of the code and use more temporaries.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -161,6 +161,30 @@ impl<TT> TraceCompiler<TT> {
                 SignedIntTy::I128 => false,
                 _ => true,
             },
+            Ty::Array(_) => false,
+            Ty::Ref(_) | Ty::Bool => true,
+            Ty::Struct(..) | Ty::Tuple(..) => false,
+            Ty::Unimplemented(..) => todo!("{}", ty),
+        }
+    }
+
+    /// Determine if the type needs to be copied when it is being dereferenced.
+    fn is_copyable(tyid: &TypeId) -> bool {
+        // FIXME For now let's do the dumbest thing and disallow copying of arrays/tuples/structs.
+        // This means that dereferencing an &Array just returns the reference, which is incorrect
+        // if the element types are copyable too.  Once we've implemented copying of
+        // arrays/tuples/structs we will update this function.
+        let ty = SIR.ty(tyid);
+        match ty {
+            Ty::UnsignedInt(ui) => match ui {
+                UnsignedIntTy::U128 => false,
+                _ => true,
+            },
+            Ty::SignedInt(si) => match si {
+                SignedIntTy::I128 => false,
+                _ => true,
+            },
+            Ty::Array(_) => false,
             Ty::Ref(_) | Ty::Bool => true,
             Ty::Struct(..) | Ty::Tuple(..) => false,
             Ty::Unimplemented(..) => todo!("{}", ty),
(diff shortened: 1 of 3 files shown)


ykcompile/src/lib.rs
```diff
@@ -208,23 +209,87 @@ impl<TT> TraceCompiler<TT> {
...
-                    // stack. So this may also return a memory location.
+                    // FIXME We currently assume Deref is only called on Refs.
+
+                    // Are we dereferencing a reference, if so, what's its type.
+                    let tyid = match ty {
+                        Ty::Ref(rty) => rty.clone(),
+                        _ => todo!(),
+                    };
+
+                    if Self::can_live_in_register(&tyid) {
```
vext01, line 220: I don't think it's true that because a type *could* live in a register, that it has to. For example, if we are out of registers, then we would spill.

Should this check be looking at the `Location` of the instance, rather than whether its type could fit in a register?

ykcompile/src/lib.rs
```diff
@@ -208,23 +209,87 @@ impl<TT> TraceCompiler<TT> {
...
+                                ; imul Rq(temp), [Rq(ro.reg) + ro.offs], elemsize as i32
+                            );
+                        }
+                        _ => todo!(),
+                    }
+                    // Add together the index and the array address and retrieve its value.
+                    match &curloc {
                         Location::Mem(ro) => {
                             dynasm!(self.asm
-                                ; mov Rq(temp), [Rq(ro.reg) + ro.offs]
```
vext01, line 217: So looks like it wasn't practical to use `lea`?

ykcompile/src/lib.rs
ptersilie: This check is only to differentiate things like `u8`s from arrays, structs, etc.

ykcompile/src/lib.rs
ptersilie: I'm more than happy if you have suggestions on how to do this stuff in less instructions. But I don't think `lea` is gonna help us here.

vext01:
We discussed this offline and have decided to use a sequence of incremental PRs to implement arrays.

I believe @ptersilie will push a couple of small changes to this one before we merge it.

ptersilie:
Just one more commit to be more clear about how we currently deal with copyable data types.

ptersilie:
Also needs a rebase.

vext01:
Please squash and rebase.

ptersilie:
Squashed and rebased.

vext01:
```
let th = start_tracing(Some(TracingKind::HardwareTracing));
```

Get rid of `Some` or pass just `TracingKind::default()`.

ptersilie:
Whoops. Didn't run tests after the rebase. Fixed.

ptersilie:
Also need to prime the ci first.

ptersilie:
Sorry, need to squash first.

ptersilie:
Okay, now go!

PR #124, opened 2020-09-25
Refactor implementation of deref projections.

ptersilie:
Make sure we don't silently return a reference to a value, when we should have copied it.
We will now hit a `todo!()` when trying to dereference an array without also indexing it.


vext01:
Can't argue with this.

bors r+

PR #128, opened 2020-10-01
Left-hand sides

vext01:
See the first commit message for details.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -44,7 +44,7 @@ impl Display for CompileError {
 /// Converts a register number into it's string name.
 fn local_to_reg_name(loc: &Location) -> &'static str {
     match loc {
-        Location::Register(r) => match r {
+        Location::Register(r, _) => match r {
             0 => "rax",
             1 => "rcx",
             2 => "rdx",
@@ -97,7 +97,8 @@ pub struct RegAndOffset {
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum Location {
-    Register(u8),
+    /// Register number, and if it contains a pointer.
+    Register(u8, bool),
     Mem(RegAndOffset),
     NotLive,
 }
@@ -188,9 +189,9 @@ impl<TT> TraceCompiler<TT> {
         }
     }
 
-    fn place_to_location(&mut self, p: &Place) -> (Location, Ty) {
+    fn place_to_location(&mut self, p: &Place, store: bool) -> (Location, Ty) {
         if !p.projection.is_empty() {
-            self.resolve_projection(p)
+            self.resolve_projection(p, store)
         } else {
             let ty = self.place_ty(&Place::from(p.local)).clone();
             (self.local_to_location(p.local), ty)
@@ -198,7 +199,7 @@ impl<TT> TraceCompiler<TT> {
     }
 
     /// Takes a `Place`, resolves all projections, and returns a `Location` containing the result.
-    fn resolve_projection(&mut self, p: &Place) -> (Location, Ty) {
+    fn resolve_projection(&mut self, p: &Place, store: bool) -> (Location, Ty) {
         let mut curloc = self.local_to_location(p.local);
         let mut ty = self.place_ty(&Place::from(p.local)).clone();
         let mut iter = p.projection.iter().peekable();
@@ -235,19 +236,22 @@ impl<TT> TraceCompiler<TT> {
                                 Location::Mem(ro) => {
                                     dynasm!(self.asm
                                         ; lea Rq(temp), [Rq(ro.reg) + ro.offs + i32::try_from(offs).unwrap()]
-                                        ; mov Rq(temp), [Rq(temp)]
                                     );
                                 }
-                                Location::Register(reg) => {
+                                Location::Register(reg, _) => {
                                     dynasm!(self.asm
                                         ; lea Rq(temp), [Rq(reg) + i32::try_from(offs).unwrap()]
-                                        ; mov Rq(temp), [Rq(temp)]
                                     );
                                 }
                                 _ => unreachable!(),
                             }
+                            if !store {
+                                dynasm!(self.asm
+                                    ; mov Rq(temp), [Rq(temp)]
+                                );
+                            }
                             self.free_if_temp(curloc);
-                            curloc = Location::Register(temp);
+                            curloc = Location::Register(temp, store);
                         }
                         Ty::Tuple(_tty) => todo!(),
                         _ => unreachable!(),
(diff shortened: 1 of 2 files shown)


ykcompile/src/lib.rs
```diff
@@ -1910,4 +1910,30 @@ mod tests {
         ct.execute(&mut args);
         assert_eq!(args.0, 100);
     }
+
+    /// Test codegen of copying something which doesn't fit in a register.
+    #[ignore] // Places, once resolved, are currently assumed  to git in a register.
```
ptersilie, line 1915: Assumed to what?

vext01: lol, `to fit`. my bad.

ptersilie: Duh, that somehow went right past me.

ykcompile/src/lib.rs
```diff
@@ -97,7 +97,8 @@ pub struct RegAndOffset {
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum Location {
-    Register(u8),
+    /// Register number, and if it contains a pointer.
+    Register(u8, bool),
```
jacob-hughes, line 101: I think the codegen logic would be a little easier to follow if instead this was an enum, e.g.:
```
enum RegTy {
    Pointer,
    Immediate
}
```

ykcompile/src/lib.rs
```diff
@@ -473,160 +485,99 @@ impl<TT> TraceCompiler<TT> {
...
+                match rloc {
+                    Location::Register(reg2, _) => {
                         dynasm!(self.asm
-                            ; mov Rq(reg1), Rq(reg2)
+                            ; mov Rq(reg), Rq(reg2)
                         );
                     }
                     _ => todo!(),
                 }
+                self.free_if_temp(rloc); // ????
```
jacob-hughes, line 551: This comment doesn't fill me with confidence lol

ykcompile/src/lib.rs
```diff
@@ -935,128 +778,65 @@ impl<TT> TraceCompiler<TT> {
...
+                        Constant::Bool(b) => self.c_bool(temp, *b),
                         c => todo!("{}", c),
                     },
                     Rvalue::CheckedBinaryOp(binop, op1, op2) => {
-                        self.c_checked_binop(l, binop, op1, op2)
+                        self.c_checked_binop(temp, binop, op1, op2);
+                        // FIXME deal with overflow
+                        let mut l = l.clone();
+                        l.projection.push(Projection::Field(0));
+                        // FIXME dedup code
```
vext01, line 828: We should fix this now.

ykcompile/src/lib.rs
```diff
@@ -789,7 +666,6 @@ impl<TT> TraceCompiler<TT> {
         let save_regs = [RDI, RSI, RDX, RCX, R8, R9, R10, R11]
```
vext01, line 666: The comment above this is inaccurate. Shall we fix it now?

ykcompile/src/lib.rs
```diff
@@ -473,160 +485,99 @@ impl<TT> TraceCompiler<TT> {
...
+        self.free_if_temp(loc);
     }
 
-    fn mov_place_ref(&mut self, p1: &Place, p2: &Place) {
-        let (lloc, _) = self.place_to_location(p1);
-
+    /// Codegen a reference `p` and store the result in `reg`.
+    fn c_ref(&mut self, reg: u8, p: &Place) {
         // Deal with the special case `&*`, i.e. referencing a `Deref` on a reference just returns
         // the reference.
```
jacob-hughes, line 528: I find this a bit hard to read so I have a dumb question. Will this work with multiple deref operators: `&**`, e.g.:
```
let s = &String::new("hello");
let str_ = &**s;
```
I _think_ `str_` should be of type `&str` here, because we first deref the `&String` to get `String`, and then `&*` will convert it to `&str` because it will call `String::deref` (where the deref method here does implicit conversion).

ykcompile/src/lib.rs
ptersilie: Yes, I think we should fix it.

ykcompile/src/lib.rs
ptersilie: Okay, I think it's not a super trivial fix though.

jacob-hughes (commented):
There's quite a lot of code here and it's a part of rustc which I'm not too familiar with so unfortunately this review is a little shallow. Looks mostly good, but a few comments.

ykcompile/src/lib.rs
ptersilie: Whoops. This was a marker because we wanted to double check if this statement is actually needed. Will remove.

ykcompile/src/lib.rs
ptersilie: Apart from the fact that we can't deal with custom deref implementations anyway yet, there is still a bug here. Your example shouldn't hit the special case, which is limited to `&*` on `Ref`s. So it should instead hit the bottom case which would just resolve the projections in the correct order. However, because our conditions are a bit wonky, your example would actually hit the special case which is wrong and needs to be fixed. But that belongs in a separate PR.

ykcompile/src/lib.rs
ptersilie: You mean in place of the `bool`. I have no preference here. @vext01 Thoughts?

ykcompile/src/lib.rs
vext01: I've changed this line in my branch, so perhaps let's leave it for this PR.

ykcompile/src/lib.rs
```diff
@@ -1069,6 +849,75 @@ impl<TT> TraceCompiler<TT> {
...
+    /// Store a value from the register `reg` into place `p`.
+    fn store(&mut self, p: &Place, reg: u8) {
+        let (lloc, ty) = self.place_to_location(p, true);
+        match &lloc {
+            Location::Register(reg1, is_ptr) => {
+                if *is_ptr {
+                    // If the lhs is a projection that results in a memory address (e.g.
+                    // `(*$1).0`), then the value in `reg1` is a pointer to store into.
+                    match ty.size() {
+                        0 => (), // ZST.
```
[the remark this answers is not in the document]
vext01: IIRC, MIR semantics allow assignment to a ZST. For example when calling a function that returns unit. The way the codegen currently works, I think we would still allocate a ZST and if we are out of registers, that might be a memory location.

Is that right @ptersilie ?

ykcompile/src/lib.rs
vext01: At least converting it from a tuple variant to a struct variant, so fields could have names would help. But that'd muddy this PR quite a bit, so save for later?

ykcompile/src/lib.rs
vext01: hah!

ykcompile/src/lib.rs
jacob-hughes: Personally, I think an enum for this purpose is a lot more descriptive and harder to misuse, but it's not a hill I'll die on.

ykcompile/src/lib.rs
jacob-hughes: Ok I see. Perhaps add a comment that summarises this in a FIXME?

bjorn3: > Apart from the fact that we can't deal with custom deref implementations anyway yet

Those are lowered into calls when building MIR.

ptersilie: Right, I've already forgotten about that. :-)

ykcompile/src/lib.rs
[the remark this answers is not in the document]
ptersilie: I agree. Changing this seems to be a bit more involved so this is for a later PR.

ykcompile/src/lib.rs
vext01: I agree, but since I don't think it'll be a small change, could we vacate the hill until a later date? :P

vext01:

I removed the ignored test, as it will appear in a future PR.

ykcompile/src/lib.rs
jacob-hughes: Works for me

ykcompile/src/lib.rs
jacob-hughes: I think this still needs fixing?

ykcompile/src/lib.rs
jacob-hughes: Can we FIXME here before merging?

jacob-hughes:
Ok, please squash

PR #129, opened 2020-10-01
Large places

vext01:
The first commit message explains.

There's some renaming mixed in with the first commit. I tried to separate it out, but it's more trouble than it's worth. Poor planning on my part. Sorry about that.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -15,6 +15,7 @@ use libc::{c_void, dlsym, RTLD_DEFAULT};
 use stack_builder::StackBuilder;
 use std::collections::HashMap;
 use std::convert::TryFrom;
+use std::ffi::CString;
 use std::fmt::{self, Display, Formatter};
 use std::mem;
 use std::process::Command;
@@ -1206,8 +1207,6 @@ impl<TT> TraceCompiler<TT> {
 
     /// Returns a pointer to the static symbol `sym`, or an error if it cannot be found.
     fn find_symbol(sym: &str) -> Result<*mut c_void, CompileError> {
-        use std::ffi::CString;
-
         let sym_arg = CString::new(sym).unwrap();
         let addr = unsafe { dlsym(RTLD_DEFAULT, sym_arg.into_raw()) };
 
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -27,6 +27,8 @@ use yktrace::tir::{
 
 use dynasmrt::{DynasmApi, DynasmLabelApi};
 
+const CALLER_SAVE_REGS: [dynasmrt::x64::Rq; 8] = [RDI, RSI, RDX, RCX, R8, R9, R10, R11];
+
 #[derive(Debug, Hash, Eq, PartialEq)]
 pub enum CompileError {
     /// The binary symbol could not be found.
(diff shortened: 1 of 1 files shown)


ykcompile/src/lib.rs
```diff
@@ -550,11 +550,12 @@ impl<TT> TraceCompiler<TT> {
                     }
                     _ => todo!(),
                 }
-                return;
+                return Location::Register(reg, false);
```
jacob-hughes, line 554: Is the `return` keyword here necessary?

ykcompile/src/lib.rs
```diff
@@ -850,73 +871,179 @@ impl<TT> TraceCompiler<TT> {
...
+                            dynasm!(self.asm
+                                ; mov Rq(temp), QWORD [Rq(src_ro.reg) + src_ro.offs]
+                                ; mov QWORD [Rq(dest_ro.reg) + dest_ro.offs], Rq(temp)
+                            );
+                        }
+                        _ => unreachable!(),
+                    }
+                    self.free_if_temp(Location::Register(temp, false));
+                } else {
+                    self.c_memcpy(dest_ro, src_ro, ty.size());
```
jacob-hughes, line 972: Can these locations overlap?

ykcompile/src/lib.rs
```diff
@@ -480,50 +494,35 @@ impl<TT> TraceCompiler<TT> {
...
+        let sym = Self::find_symbol("memcpy").unwrap();
+        self.caller_save();
+        dynasm!(self.asm
+            ; push rax
+            ; xor rax, rax
+            ; lea rdi, [Rq(dest.reg) + dest.offs]
+            ; lea rsi, [Rq(src.reg) + src.offs]
+            ; mov rdx, size as i32
+            ; mov r11, QWORD sym as i64
+            ; call r11
```
jacob-hughes, line 509: Why is it necessary to lookup the `memcpy` fn pointer, put it in a register and then call it. Can't you just call memcpy directly here?

vext01: I've been unable to achieve that using `dynasm` so far. Looking at the docs, it seems you can't.

image


Notice there is no `imm64` variant. [This SO post](
https://stackoverflow.com/questions/19552158/call-an-absolute-pointer-in-x86-machine-code) confirms it too.

I think you can use `rel32` [if you know where memcpy is relative to the instruction pointer](https://stackoverflow.com/questions/38961192/how-to-execute-a-call-instruction-with-a-64-bit-absolute-address)

ykcompile/src/lib.rs
vext01: Yes, it's an intentional early return. During development I removed it accidentally and hilarity ensued.

ykcompile/src/lib.rs
vext01: And thus should we use `memmove()`. Good question.

I don't think they can overlap. @ptersilie?

vext01: Can you guys think of a rust assignment that would copy into memory which overlaps with the source operand? Does the borrow checker disallow that perhaps.

bjorn3: The MIR semantics regarding aliasing are not clear right now.

ptersilie: I honestly don't know. But gut feeling says no, but who knows?

ykcompile/src/lib.rs
jacob-hughes: Huh, strange. That stack overflow post was interesting.

ykcompile/src/lib.rs
jacob-hughes: Ah yes, I didn't see that it was part of the if block.

ykcompile/src/lib.rs
vext01: It sounds like if the buffers overlap, we invoke undefined behaviour. So to be completely safe, I suppose we should use `memmove` and take the performance penalty.

jacob-hughes: Yeah I agree that's the sensible approach.

jacob-hughes:
LGTM. Does this need squashing?

vext01:
If that last push is OK, I'll squash.

jacob-hughes:
Yeah, please squash

vext01:
splat.

PR #130, opened 2020-10-02
Replace the is_ptr flag with a new Location variant.

vext01:
This is much clearer and it doesn't cause case explosion, as the flag already exploded store().

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -47,7 +47,7 @@ impl Display for CompileError {
 /// Converts a register number into it's string name.
 fn local_to_reg_name(loc: &Location) -> &'static str {
     match loc {
-        Location::Register(r, _) => match r {
+        Location::Register(r) => match r {
             0 => "rax",
             1 => "rcx",
             2 => "rdx",
@@ -100,9 +100,13 @@ pub struct RegAndOffset {
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum Location {
-    /// Register number, and if it contains a pointer.
-    Register(u8, bool),
+    /// A value in a register.
+    Register(u8),
+    /// A statically known memory location relative to a register.
     Mem(RegAndOffset),
+    /// A runtime memory location stored in a register.
+    Addr(u8),
+    /// A non-live location. Used by the register allocator.
     NotLive,
 }
 
@@ -122,9 +126,9 @@ impl Location {
     }
 
     /// If `self` is a `Register` then unwrap it, otherwise panic.
-    fn unwrap_reg(&self) -> (u8, bool) {
-        if let Location::Register(reg, is_ptr) = self {
-            (*reg, *is_ptr)
+    fn unwrap_reg(&self) -> u8 {
+        if let Location::Register(reg) = self {
+            *reg
         } else {
             panic!("tried to unwrap a Register location when it wasn't a Register");
         }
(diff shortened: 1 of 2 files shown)


ykcompile/src/lib.rs
```diff
@@ -266,7 +278,7 @@ impl<TT> TraceCompiler<TT> {
                                 );
                             }
                             self.free_if_temp(curloc);
-                            curloc = Location::Register(temp, store);
+                            curloc = Self::reg_or_addr(temp, store);
```
ptersilie, line 281: Since this is always preceded by `if !store`, should we remove `reg_or_addr` and do something like:
```
if !store {
    dynasm!(self.asm
     ; mov Rq(temp), [Rq(temp)]
    );
    curloc = Location::Register(temp);
 } else {
  curloc = Location::Address(temp);
}
```
I don't know if that makes things clearer.

vext01:
Try this.

By the way, should `ty` be updated each time `curloc` is? That's currently not the case.

ykcompile/src/lib.rs
```diff
@@ -323,14 +318,17 @@ impl<TT> TraceCompiler<TT> {
                             }
                             _ => unreachable!(),
                         };
-                        if !store {
+                        let nextloc = if store {
```
ptersilie, line 321: I think we can just do:
```
self.free_if_temp(curloc);
curloc = if store {
    Location::Add(temp);
} else {
    ...
    Location::Register(temp);
}
```

ptersilie:
I think we should update `ty` each time and I thought we did.

ykcompile/src/lib.rs
vext01: But you have to free `curloc`.

vext01: Doing it this way we don't duplicate the freeing.

vext01:
> I think we should update ty each time and I thought we did.

OK, I can fix this in this PR, but I want to get the previous stuff signed off first.

ptersilie:
Do you have an example where we don't update `ty`. I couldn't see any.

ptersilie:
Also, there's another comment for you.

vext01:
We don't update `ty` in the [Index case](https://github.com/softdevteam/yk/blob/cedda5224f094bad684123da39cc0ee106cba3b8/ykcompile/src/lib.rs#L324)

ykcompile/src/lib.rs
ptersilie: I don't follow. I am freeing `curloc` and then I'm overwriting it with a new location.

ptersilie:
You're right. That needs to be fixed.

vext01:
> You're right. That needs to be fixed.

I think that last commit is right, but I was unable to write a test to cover it. We don't support enough of arrays yet.

ptersilie:
Yes, hopefully that will change soon. Looks good. Please squash.

vext01:
splat.

ptersilie:
Whoot? ` cannot install while Rust is installed`

vext01:
```
warning: it looks like you have an existing installation of Rust at...
```

That's new! Investigating.

vext01:
bors try

vext01:
bors try

vext01:
OK to squash?

ptersilie:
Please do.

vext01:
splat.

PR #135, opened 2020-10-10
New trimming scheme for yk

vext01:
Here's the new trimming scheme for yk.

The first commit contains changes to the codegen to fix problems that were exposed by the trimming changes. One previously passing test remains marked `#[ignore] // FIXME`, but as agreed offline, since we are thinking of changing the IR quite fundamentally, we won't get hung up on it.

The second commit removes all the stuff that we no longer need.

Companion PR ~~coming~~ https://github.com/softdevteam/ykrustc/pull/136.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -356,7 +356,7 @@ impl<TT> TraceCompiler<TT> {
                                 ; add Rq(temp), [Rq(ro.reg) + ro.offs]
                             );
                         }
-                        Location::Register(reg) => todo!(),
+                        Location::Register(_) => todo!(),
                         Location::Addr(reg) => {
                             dynasm!(self.asm
                                 ; add Rq(temp), Rq(reg)
@@ -423,9 +423,7 @@ impl<TT> TraceCompiler<TT> {
     /// performs one.
     fn local_to_location(&mut self, l: Local) -> Location {
         if l == INTERP_STEP_ARG {
-            // If the local references `trace_inputs` return its location on the stack, which is
-            // stored in the first argument of the executed trace.
-            Location::Register(RDI.code())
+            Location::Register(RDI.code()) // i.e. the first arg register.
         } else if let Some(location) = self.variable_location_map.get(&l) {
             // We already have a location for this local.
             location.clone()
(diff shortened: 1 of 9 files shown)


bjorn3:
How would usage of `#[interp_step]` work when using an interpreter loop? How would something like the following work?

```rust
let mut state = State::default();
let tracer = start_tracing();
loop {
    step_fn(&mut state);
    if state.finished {
        break;
    }
}
tracer.stop_tracing();
```

The `if` is no longer traced.

vext01:
> The if is no longer traced.

[This test](https://github.com/softdevteam/yk/blob/b8296b72a8234140e3653dca3b05488ff00b52ae/ykrt/src/mt.rs#L422-L429) shows how we envisage it to work.

But you're right that it needs to return true/false to say if the program is finished or not!

ykcompile/src/lib.rs
```diff
@@ -346,16 +302,35 @@ impl<TT> TraceCompiler<TT> {
                             Location::Register(temp)
                         };
                         ty = SIR.ty(&tyid).clone();
+                    } else {
+                        // Dereferencing when the pointee is uncopyable the location to an address.
```
ptersilie, line 306: Dereferencing a pointer, where the pointee is uncopyable, turns the location to an address?

ykcompile/src/lib.rs
```diff
@@ -346,16 +302,35 @@ impl<TT> TraceCompiler<TT> {
...
+                        self.free_if_temp(curloc);
+                        curloc = Location::Addr(temp);
                     }
                 }
                 Projection::Index(local) => {
                     // Get the type of the array elements.
                     let elem_ty = match ty {
-                        Ty::Array(_) => {
+                        Ty::Array(ety) => {
                             // FIXME Since we can't compile array construction yet, we can assume
```
ptersilie, line 330: Think this FIXME can go now.

ptersilie:
Only two minor comments.

vext01:
Fixes pushed.

ptersilie:
Please squash.

vext01:
Those can stay separate.

Shall I prime?

ptersilie:
I meant the `Fix comments` commit which you squashed as I was typing the "Please squash." message. Yes, please prime.

PR #137, opened 2020-10-20
Implement casting.

ptersilie:


ykcompile/src/lib.rs
```diff
@@ -214,7 +214,7 @@ impl<TT> TraceCompiler<TT> {
             // An array is copyable if its elements are.
             Ty::Array(ety) => Self::is_copyable(ety),
             Ty::Slice(ety) => Self::is_copyable(ety),
-            Ty::Ref(_) | Ty::Bool => true,
+            Ty::Ref(_) | Ty::Bool | Ty::Char => true,
```
vext01, line 217: Are these unicode chars? If so they may not fit in a register, but I think we can address that later.

vext01:
Other than that one comment LGTM.

ykcompile/src/lib.rs
bjorn3: They are unicode chars. Those are represented as 32bit integers, of which only 24bit are used. The rest must be all zeros.

ykcompile/src/lib.rs
```diff
@@ -633,6 +633,86 @@ impl<TT> TraceCompiler<TT> {
...
+                            }
+                            2 => match ty.size() {
+                                1 => {
+                                    dynasm!(self.asm
+                                        ; movzx Rw(temp), Rb(reg)
+                                    );
+                                }
+                                2 | 4 | 8 => {
+                                    dynasm!(self.asm
+                                        ; mov Rw(temp), Rw(reg)
```
bjorn3, line 657: You should prefer 32bit operations over 8 and 16bit operations to prevent partial register stalls.

ykcompile/src/lib.rs
vext01: So not UTF-8?

I was expecting chars to be variable width utf-8 encoded.

ykcompile/src/lib.rs
```diff
@@ -633,6 +633,86 @@ impl<TT> TraceCompiler<TT> {
...
+                                        ; mov Rd(temp), Rd(reg)
+                                    );
+                                }
+                                _ => todo!("{}", ty.size()),
+                            },
+                            8 => {
+                                match ty.size() {
+                                    1 => {
+                                        dynasm!(self.asm
+                                            ; movzx Rq(temp), Rb(reg)
```
bjorn3, line 684: What about signed integers?

ykcompile/src/lib.rs
```diff
@@ -633,6 +633,86 @@ impl<TT> TraceCompiler<TT> {
...
+                                    2 => {
+                                        dynasm!(self.asm
+                                            ; movzx Rq(temp), Rw(reg)
+                                        );
+                                    }
+                                    4 | 8 => {
+                                        // mov reg32, reg32 in x64 automatically zero-extends to
+                                        // reg64.
+                                        dynasm!(self.asm
+                                            ; mov Rq(temp), Rq(reg)
```
bjorn3, line 696: For 32bit integers you need to use `Rd` not `Rq`.

ykcompile/src/lib.rs
vext01: Then the value should be sign extended. Good catch.

ykcompile/src/lib.rs
ptersilie: Good point.

ykcompile/src/lib.rs
ptersilie: How would I then make sure that the operand is cast down?

ykcompile/src/lib.rs
ptersilie: Fixed in c89306c

ykcompile/src/lib.rs
bjorn3: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c87c93dc6af84ebf7cbaaa1d1a04e8f9

```rust
fn main() {
    assert_eq!(std::mem::size_of::<char>(), 4);
}
```

ptersilie: Ah, yes, `chars` are 4 bytes, which is addressed in `test_cast`.

ykcompile/src/lib.rs
bjorn3: Got confused. I though this was casting to a bigger size. If it is to a smaller size, you can make it a nop and just say that the upper bits of a register are undefined. This is the model Cranelift chose.

ptersilie: Why a nop? I still need to move the value though, don't I, e.g. `let a = b as u8` where `b` is `u16` should result in `mov Rb(a), Rb(b)`.

ykcompile/src/lib.rs
ptersilie: I've split `c_cast` into multiple functions and added a `todo!` for now. Commit coming soon once I figured out what's going on with ykrustc.

ykcompile/src/lib.rs
ptersilie: I think I get it now. The cast does nothing, and then the `store` function handles the move.

vext01:
Is this ready for squashing?

ykcompile/src/lib.rs
ptersilie: "Fixed" in 19fe606

ykcompile/src/lib.rs
ptersilie: Actually, we can't currently do a `nop` here, since the `store` function doesn't care about the size and always moves quadwords. We probably want to change the `store` function. But since everything is going to change with the new IR, we keep it as is for now.

ptersilie:
Can I squash?

vext01:
Yup

ptersilie:
Squashed and ready for merging.

PR #140, opened 2020-11-04
New IR.

vext01:
This is a new design for SIR (and therefore TIR). The new design
resolves as many projections at SIR generation time so that they don't
have to be done at runtime. By doing so, we lessen the burden on the
trace code generator and the register allocator.

Companion to: https://github.com/softdevteam/ykrustc/pull/140

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -21,10 +21,9 @@ use std::ffi::CString;
 use std::fmt::{self, Display, Formatter};
 use std::mem;
 use std::process::Command;
-use ykpack::{SignedIntTy, Ty, TypeId, UnsignedIntTy};
+use ykpack::{IPlace, OffT, SignedIntTy, Ty, TypeId, UnsignedIntTy};
 use yktrace::tir::{
-    BinOp, CallOperand, Constant, ConstantInt, Guard, GuardKind, Local, Operand, Place, Projection,
-    Rvalue, Statement, TirOp, TirTrace,
+    BinOp, CallOperand, Constant, Guard, GuardKind, Local, Statement, TirOp, TirTrace,
 };
 use yktrace::{sir::SIR, INTERP_STEP_ARG};
 
@@ -35,14 +34,13 @@ lazy_static! {
     static ref CALLER_SAVE_REGS: [u8; 8] = [RDI.code(), RSI.code(), RDX.code(), RCX.code(),
                                             R8.code(), R9.code(), R10.code(), R11.code()];
 
-    // Register partitioning. These arrays must not overlap.
+    // The register partitioning. These arrays must not overlap.
     // FIXME add callee save registers to the pool. Trace code will need to save/restore them.
-    static ref TEMP_REGS: [u8; 3] = [R9.code(), R10.code(), R11.code()];
-    static ref LOCAL_REGS: [u8; 3] = [R8.code(), RDX.code(), RCX.code()];
-}
+    static ref TEMP_REG: u8 = R11.code();
+    static ref REG_POOL: [u8; 5] = [R10.code(), R9.code(), R8.code(), RDX.code(), RCX.code()];
 
-fn is_temp_reg(reg: u8) -> bool {
-    TEMP_REGS.contains(&reg)
+    static ref TEMP_LOC: Location = Location::Register(*TEMP_REG);
+    static ref PTR_SIZE: u64 = u64::try_from(mem::size_of::<usize>()).unwrap();
 }
 
 #[derive(Debug, Hash, Eq, PartialEq)]
@@ -111,7 +109,16 @@ impl<TT> CompiledTrace<TT> {
 #[derive(Debug, Clone, PartialEq)]
 pub struct RegAndOffset {
     reg: u8,
-    offs: i32,
+    off: OffT,
+}
+
+/// Describes the location of the pointer in Location::Indirect.
+#[derive(Debug, Clone, PartialEq)]
+pub enum IndirectLoc {
+    /// There's a pointer in this register.
+    Register(u8),
+    /// There's a pointer in memory somewhere.
+    Mem(RegAndOffset),
 }
 
 #[derive(Debug, Clone, PartialEq)]
@@ -120,18 +127,21 @@ pub enum Location {
     Register(u8),
     /// A statically known memory location relative to a register.
     Mem(RegAndOffset),
-    /// A runtime memory location stored in a register.
-    Addr(u8),
+    /// A location that contains a pointer to some underlying storage.
+    Indirect { ptr: IndirectLoc, off: OffT },
+    /// A statically known constant.
+    Const { val: Constant, ty: TypeId },
     /// A non-live location. Used by the register allocator.
     NotLive,
 }
 
 impl Location {
     /// Creates a new memory location from a register and an offset.
-    fn new_mem(reg: u8, offs: i32) -> Self {
-        Self::Mem(RegAndOffset { reg, offs })
+    fn new_mem(reg: u8, off: OffT) -> Self {
+        Self::Mem(RegAndOffset { reg, off })
     }
 
+    #[cfg(test)]
     /// If `self` is a `Mem` then unwrap it, otherwise panic.
     fn unwrap_mem(&self) -> &RegAndOffset {
         if let Location::Mem(ro) = self {
@@ -141,17 +151,64 @@ impl Location {
         }
     }
 
-    /// If `self` is a `Register` then unwrap it, otherwise panic.
-    fn unwrap_reg(&self) -> u8 {
-        if let Location::Register(reg) = self {
-            *reg
+    /// If `self` is a `Mem` then return a mutable reference to its innards, otherwise panic.
+    fn unwrap_mem_mut(&mut self) -> &mut RegAndOffset {
+        if let Location::Mem(ro) = self {
+            ro
         } else {
-            panic!("tried to unwrap a Register location when it wasn't a Register");
+            panic!("tried to unwrap a Mem location when it wasn't a Mem");
         }
+    }
+
+    /// Returns which register (if any) is used in addressing this location.
+    fn uses_reg(&self) -> Option<u8> {
+        match self {
+            Location::Register(reg) => Some(*reg),
+            Location::Mem(RegAndOffset { reg, .. }) => Some(*reg),
+            Location::Indirect {
+                ptr: IndirectLoc::Register(reg),
+                ..
+            }
+            | Location::Indirect {
+                ptr: IndirectLoc::Mem(RegAndOffset { reg, .. }),
+                ..
+            } => Some(*reg),
+            Location::Const { .. } => None,
+            Location::NotLive => unreachable!(),
+        }
+    }
+
+    /// Apply an offset to the location, returning a new one.
+    fn offset(self, off: OffT) -> Self {
+        if off == 0 {
+            return self;
+        }
+        match self {
+            Location::Mem(ro) => Location::Mem(RegAndOffset {
+                reg: ro.reg,
+                off: ro.off + off,
+            }),
+            Location::Indirect { ptr, off: ind_off } => Location::Indirect {
+                ptr,
+                off: ind_off + off,
+            },
+            Location::Register(..) | Location::Const { .. } => todo!("offsetting a constant"),
+            Location::NotLive => unreachable!(),
+        }
+    }
+
+    /// Converts a direct place to an indirect place for use as a pointer.
+    fn to_indirect(&self) -> Self {
+        let ptr = match self {
+            Location::Register(r) => IndirectLoc::Register(*r),
+            Location::Mem(ro) => IndirectLoc::Mem(ro.clone()),
+            _ => unreachable!(),
+        };
+        Location::Indirect { ptr, off: 0 }
     }
 }
 
-/// Allocation of one of the LOCAL_REGS. Temporary registers are tracked separately.
+/// Allocation of one of the REG_POOL. Temporary registers are tracked separately.
 enum RegAlloc {
     Local(Local),
     Free,
@@ -168,20 +225,24 @@ pub struct TraceCompiler<TT> {
     register_content_map: HashMap<u8, RegAlloc>,
     /// Maps trace locals to their location (register, stack).
     variable_location_map: HashMap<Local, Location>,
-    /// Available temproary registers.
-    temp_regs: Vec<u8>,
     /// Local decls of the tir trace.
     local_decls: HashMap<Local, LocalDecl>,
     /// Stack builder for allocating objects on the stack.
     stack_builder: StackBuilder,
+    /// Stores the memory addresses of local functions.
     addr_map: HashMap<String, u64>,
     _pd: PhantomData<TT>,
 }
 
 impl<TT> TraceCompiler<TT> {
-    fn can_live_in_register(tyid: &TypeId) -> bool {
+    fn can_live_in_register(decl: &LocalDecl) -> bool {
+        if decl.referenced {
+            // We must allocate it on the stack so that we can reference it.
+            return false;
+        }
+
         // FIXME: optimisation: small structs and tuples etc. could actually live in a register.
-        let ty = SIR.ty(tyid);
+        let ty = SIR.ty(&decl.ty);
         match ty {
             Ty::UnsignedInt(ui) => match ui {
                 UnsignedIntTy::U128 => false,
@@ -191,7 +252,7 @@ impl<TT> TraceCompiler<TT> {
                 SignedIntTy::I128 => false,
                 _ => true,
             },
-            Ty::Array(_) => false,
+            Ty::Array { .. } => false,
             Ty::Slice(_) => false,
             Ty::Ref(_) | Ty::Bool | Ty::Char => true,
             Ty::Struct(..) | Ty::Tuple(..) => false,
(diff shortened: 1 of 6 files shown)


vext01:
I've just realised that I should have marked the commit as co-authored with Lukas. I'll apply that at squash time.

ykpack/src/types.rs
```diff
@@ -58,7 +64,9 @@ impl Display for Ty {
             Ty::UnsignedInt(ui) => write!(f, "{}", ui),
             Ty::Struct(sty) => write!(f, "{}", sty),
             Ty::Tuple(tty) => write!(f, "{}", tty),
-            Ty::Array(aty) => write!(f, "&{:?}", aty),
+            Ty::Array { elem_ty, len, .. } => {
+                write!(f, "[({}, {}); {}]", elem_ty.0, elem_ty.1, len)
+            }
             Ty::Slice(sty) => write!(f, "&{:?}", sty),
```
bjorn3, line 70: `&[{:?}]`?

ykpack/src/types.rs
```diff
@@ -516,46 +561,57 @@ impl Statement {
...
+                Self::maybe_push_local(&mut ret, dest.local());
+                Self::maybe_push_local(&mut ret, src.local());
+            }
             Statement::Unimplemented(_) => (),
         }
         ret
     }
 
-    /// Returns true if the statement may affect locals besides those appearing in the statement.
+    ///// Returns true if the statement may affect locals besides those appearing in the statement.
```
bjorn3, line 611: Accidental change?

bjorn3 (approved):
I took a quick look. LGTM

ykpack/src/types.rs
vext01: Yes, where i have commented and uncommented the block. Thanks.

ykpack/src/types.rs
vext01: For the slice? Yes, good call!

ykcompile/src/lib.rs
```diff
@@ -35,14 +34,13 @@ lazy_static! {
...
-    static ref TEMP_REGS: [u8; 3] = [R9.code(), R10.code(), R11.code()];
-    static ref LOCAL_REGS: [u8; 3] = [R8.code(), RDX.code(), RCX.code()];
-}
+    static ref TEMP_REG: u8 = R11.code();
+    static ref REG_POOL: [u8; 5] = [R10.code(), R9.code(), R8.code(), RDX.code(), RCX.code()];
 
-fn is_temp_reg(reg: u8) -> bool {
-    TEMP_REGS.contains(&reg)
+    static ref TEMP_LOC: Location = Location::Register(*TEMP_REG);
+    static ref PTR_SIZE: u64 = u64::try_from(mem::size_of::<usize>()).unwrap();
```
ltratt, line 43: Having `PTR_SIZE` as a `u64` feels weird. It's never going to be very big, so `u64` feels comically large: `u8` will probably deal with every computer we'll see during our lifetime. But if we want a big number, why not `usize`?

ykcompile/src/lib.rs
```diff
@@ -114,14 +112,25 @@ pub struct RegAndOffset {
...
 #[derive(Debug, Clone, PartialEq)]
 pub enum Location {
     /// A value in a register.
     Register(u8),
     /// A statically known memory location relative to a register.
     Mem(RegAndOffset),
-    /// A runtime memory location stored in a register.
-    Addr(u8),
+    /// A location that contains a pointer to some underlying storage.
+    Indirect { ptr: IndirectLoc, offs: i32 },
```
ltratt, line 131: Can offsets be negative?

ykcompile/src/lib.rs
```diff
@@ -141,17 +151,67 @@ impl Location {
...
         } else {
-            panic!("tried to unwrap a Register location when it wasn't a Register");
+            panic!("tried to unwrap a Mem location when it wasn't a Mem");
+        }
+    }
+
+    /// Returns which register (if any) is used in addressing this location.
+    fn uses_reg(&self) -> Option<u8> {
+        match self {
+            Location::Register(reg) => Some(*reg),
```
ltratt, line 166: You can combine several of these along the lines of:

```rust
match self {
    Location::Register(reg)
  | Location::Mem(RegAndOffset { reg, .. })
  | Location::Indirect {  IndirectLoc::Register(reg), .. }
  <etc>
  => Some(*reg)
```

if it makes things a bit clearer?

ykcompile/src/lib.rs
```diff
@@ -141,17 +151,67 @@ impl Location {
...
+                ptr: IndirectLoc::Mem(RegAndOffset { reg, .. }),
+                ..
+            } => Some(*reg),
+            Location::Const { .. } => None,
+            Location::NotLive => unreachable!(),
+        }
+    }
+
+    /// Apply an offset to the location, returning a new one.
+    fn offset(self, offs: i32) -> Self {
```
ltratt, line 182: `offs` -> `off`? [`offs` is an uncommon spelling, and I briefly wondered what it stood for!]

ykcompile/src/lib.rs
```diff
@@ -860,35 +460,35 @@ impl<TT> TraceCompiler<TT> {
...
-                        }
-                        Location::Addr(_) | Location::NotLive => unreachable!(),
-                    };
-                    self.free_if_temp(loc);
+            // Now load the argument into the correct argument register.
+            match self.iplace_to_location(arg) {
+                Location::Register(reg) => {
+                    // The value *was* in a register before we pushed it with caller_save().
+                    // We load it back from the stack now.
+                    // FIXME assumes that anything in a register was pushed. In reality only caller
```
ltratt, line 472: I'm struggling to parse the `FIXME`.

ykcompile/src/lib.rs
```diff
@@ -860,35 +460,35 @@ impl<TT> TraceCompiler<TT> {
...
+                        ; mov Rq(arg_reg), [rsp + off]
+                    );
                 }
-                Operand::Constant(c) => {
+                Location::Mem(ro) => dynasm!(self.asm
+                    ; mov Rq(arg_reg), [Rq(ro.reg) + ro.offs]
+                ),
+                Location::Indirect { .. } => todo!(),
+                Location::Const { val, .. } => {
+                    // FIXME assumes constant fits in a register.
```
ltratt, line 485: This sounds `assert`able?

ykcompile/src/lib.rs
```diff
@@ -908,113 +508,216 @@ impl<TT> TraceCompiler<TT> {
...
-            Operand::Constant(Constant::Bool(_b)) => todo!(),
-            Operand::Constant(c) => todo!("{}", c),
-        };
+        // FIXME. Check for overflow here. Examine either CF or OF depending on signedness.
 
-        match (&src1, &src2) {
-            (Location::Register(reg1), Location::Register(reg2)) => match ty.size() {
-                1 => {
+        // 3) Move the result to where it is supposed to live.
+        // If it is a checked operation, then we have to build a (value, overflow?) tuple.
```
ltratt, line 554: `overflow?` -> `overflows` maybe? [I wondered initially if the `?` meant "I don't know if this is right or not!]

ykcompile/src/lib.rs
```diff
@@ -908,113 +508,216 @@ impl<TT> TraceCompiler<TT> {
...
-            Operand::Constant(Constant::Bool(_b)) => todo!(),
-            Operand::Constant(c) => todo!("{}", c),
-        };
-        let src2 = match op2 {
-            Operand::Place(p) => self.c_place(&p),
-            Operand::Constant(Constant::Int(ci)) => self.c_constint(&ci),
-            Operand::Constant(Constant::Bool(_b)) => todo!(),
-            Operand::Constant(c) => todo!("{}", c),
-        };
+        // FIXME. Check for overflow here. Examine either CF or OF depending on signedness.
```
ltratt, line 551: It looks to me like we do check for overflow below?

ykcompile/src/lib.rs
```diff
@@ -908,113 +508,216 @@ impl<TT> TraceCompiler<TT> {
...
                     dynasm!(self.asm
-                        ; cmp Rb(reg1), Rb(reg2)
+                        ; add Rq(*TEMP_REG), Rq(r)
                     );
                 }
-                2 => {
-                    dynasm!(self.asm
-                        ; cmp Rw(reg1), Rw(reg2)
-                    );
+                _ => unreachable!(format!("{}", SIR.ty(&dest.ty()))),
```
ltratt, line 578: I think `unreachable!("{}", SIR.ty(...)` works?

ykcompile/src/lib.rs
```diff
@@ -1050,393 +753,637 @@ impl<TT> TraceCompiler<TT> {
...
-                            ; mov [Rq(dest_reg)], Rq(src_reg)
-                        );
-                    }
-                    _ => unreachable!(),
-                }
+    fn c_mkref(&mut self, dest: &IPlace, src: &IPlace) {
+        let src_loc = self.iplace_to_location(src);
+        match src_loc {
+            Location::Register(..) => {
+                // This isn't possible as the allocator explicitely puts things which are
```
ltratt, line 852: "explicitly"

ykpack/src/types.rs
```diff
@@ -38,7 +38,13 @@ pub enum Ty {
     /// A tuple type.
     Tuple(TupleTy),
     /// An array type.
-    Array(TypeId),
+    /// FIXME byte_size and align can be computed from elem_ty and len, but requires some refactoring.
```
ltratt, line 41: Should this line be a doc comment or just `//`?

ykpack/src/types.rs
```diff
@@ -455,34 +376,169 @@ impl BasicBlock {
...
+        write!(f, "{}+{}", self.local, self.offs)
+    }
+}
+
+/// An IR place. This is used in SIR and TIR to describe the (abstract) address of a piece of data.
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
+pub enum IPlace {
+    /// The IPlace describes a value as a Local+offset pair.
+    Val { local: Local, offs: u32, ty: TypeId },
+    /// An indirect place behind a pointer. Ykrustc uses these for deref and (dynamic) index
```
ltratt, line 406: `ykrustc`

ykpack/src/types.rs
```diff
@@ -491,23 +547,18 @@ impl Statement {
...
-                if let Some(dest) = dest {
-                    dest.push_maybe_defined_locals(&mut ret);
-                }
-                for idx in 0..args.len() {
-                    // + 1 to skip return value.
-                    ret.push(Local(start_idx + u32::try_from(idx).unwrap() + 1));
-                }
-            }
-            Statement::Leave => (),
+            Statement::IStore(dest, _src) => Self::maybe_push_local(&mut ret, dest.local()),
```
ltratt, line 550: Isn't `maybe_push_local` just `ret.map(|x| x.push(dest.local())` (or similar) in disguise?

ykpack/src/types.rs
```diff
@@ -516,37 +567,48 @@ impl Statement {
     /// Returns a vector of locals that this SIR statement uses.
     /// A definition is considered a use, so this returns a superset of what
     /// `maybe_defined_locals()` does.
+    /// FIXME above
```
ltratt, line 570: Huh?

yktrace/src/tir.rs
```diff
@@ -262,12 +280,25 @@ impl<'a> TirTrace<'a> {
                     // statements for call arguments. Which mappings we need to remove depends on
                     // the number of arguments the function call had, which we keep track of in
                     // `cur_call_args`.
+                    //let old_offs = rnm.offset;
```
ltratt, line 283: Commented code.

yktrace/src/tir.rs
```diff
@@ -188,7 +208,9 @@ impl<'a> TirTrace<'a> {
                 continue;
             }
 
-            let stmt = match &body.blocks[user_bb_idx_usize].term {
+            // Each SIR terminator becomes zero or more TIR statements.
+            let mut term_stmts = Vec::new();
+            match &body.blocks[user_bb_idx_usize].term {
```
ltratt, line 213: I *think* this should be simplified to just:

```rust
if let Terminator::Call = &body.blocks[user_bb_idx_usize].term {
    ...
}
```

ykcompile/src/lib.rs
bjorn3: In general pointer offsets can be negative. For example `Rc` stores the reference counters at a negative offset from the stored pointer. I don't know if that can leak into here though.

ykcompile/src/lib.rs
vext01: It's because it's passed to `store_raw()`, which is expected to handle arbitrarily large sizes.

There's only one use of PTR_SIZE, so we could just cast if you still feel it is gross.

ltratt:
Unsurprisingly I haven't fully got the high-level view, but I understood more of the low-level details than I expected! A couple of over-arching comments:

* `offs` feels like a weird variable name. Why not the more common `off`?
* There are some `FIXME`s that don't have `assert`s or `panic`s attached. These always worry me, as they're loading a gun for our future selves to blow our future feet off. I think we should have a rule that says if we have a `FIXME` in the middle of code without an `assert` (or something else that aborts compilation), the comment should also explain why we can't have an `assert`. I think we should do that for this PR's changes as a first start.
* Have we run Clippy over this recently?

ykcompile/src/lib.rs
vext01: For indirect locations, in theory, yes. In practise, not sure.

ykcompile/src/lib.rs
ltratt: OK, then it probably doesn't matter too much.

ykcompile/src/lib.rs
ltratt: I think it might be worth thinking about this, because if we can't get negative offsets here, a `u32` (or `usize`?!) will save us a few future bugs. [As an aside, I suspect that `offs` / `off` should be `isize`?]

ykcompile/src/lib.rs
vext01: We use `offs` all over the place already, and I prefer it as it can't be confused with on/off. Unless you feel strongly, I'd like to keep it.

ltratt: What's the convention in other parts of the codebase e.g. rustc? My strong suspicion is that `off` is more common (I thought `offs` might refer to a set / sequence of `off`s!). Either way, we can do this as a mechanical PR subsequent to this.

ykcompile/src/lib.rs
vext01: In light of @bjorn3's comment, since any kind of smart pointer, including `Rc`, will use indirect locations to realise runtime pointers, I think this should stay signed.

ltratt: OK, then I think it should be `isize`?

vext01: Hrm, maybe. The reason it's an i32 is that the ISA doesn't allow such big/small offsets.

```
error[E0308]: mismatched types
 --> src/main.rs:7:27
  |
7 |         ; mov rax, [rsp + offs]
  |                           ^^^^ expected `i32`, found `isize`
  |
```

So using an i32 reduces the amount of annoying casting needed, but then again, `Location` ought to be platform agnostic. I don't know if other chips allow such large/small offsets.

ltratt: Another way of fudging is to make a `type` `OffT` which is an alias for `i32` on x86, but which we could realias on other platforms. Though `type` in Rust is rather underpowered, so it can give an illusion of type safety which it doesn't actually possess. There are probably no perfect answers here.

ykcompile/src/lib.rs
vext01: In rustc they tend to spell it out in full. Same in cranelift.

But at the same time I'm trying to save my hands :P

ykcompile/src/lib.rs
vext01: It seems a decent compromise. One I've considered before actually.

At least a type alias has a name which can convey meaning, even if it isn't type safe.

ykcompile/src/lib.rs
vext01: We pushed all of the caller save registers and the code is assuming that when loading argument registers we will be taking them off the stack. In reality the argument may not have been in a caller save register in the first place.

If that makes sense, I'll paste the above in a comment :P

ykcompile/src/lib.rs
ltratt: OK, I'm going to leave this one to you to think about. It's not worth us spending too much time on.

ykcompile/src/lib.rs
ltratt: OK, let's try it?

ykcompile/src/lib.rs
vext01: `i64_cast()` will crash if it doesn't fit.

ykcompile/src/lib.rs
ltratt: That makes more sense. I think the "code is assuming" bit needs mild rephrasing, but otherwise it's an improvement!

ykcompile/src/lib.rs
vext01: We don't. We just build a `(result, no-it-didnt-overflow-honest-guvnor)` tuple below.

We probably should either fix this.

ykcompile/src/lib.rs
ptersilie: For what it's worth, I've commented on this as well some time ago as I found `offs` a bit odd, but it's not a hill I'm going to die on.

ykcompile/src/lib.rs
vext01: It does. Good catch.

ykcompile/src/lib.rs
vext01: Looks like 2 vs 1. @ptersilie if you are willing to do the change, I will back down.

ykpack/src/types.rs
vext01: Yes. And perhaps it isn't even shorter. I'll kill it.

Sometimes helper functions turn out to be obfuscation functions.

ykpack/src/types.rs
vext01: Wat. I'll kill that. It's bogus.

ykcompile/src/lib.rs
vext01: But I'd wait until I've done the OffT change.

vext01: Pushed. Go ahead with that if you like.

ykcompile/src/lib.rs
ltratt: I'd be fine with a run-time segfault if the overflow check is set.

ykpack/src/types.rs
```diff
@@ -198,7 +199,7 @@ impl Display for UnsignedIntTy {
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Hash)]
 pub struct Fields {
     /// Field offsets.
-    pub offsets: Vec<u64>,
+    pub offsets: Vec<OffT>,
```
ltratt, line 202: I think this is probably a good thing, but notice here we've change a `u64` into a `OffT`, when mostly we've change `u32` intos `OffT`. But I wanted to point it out explicitly!

vext01: Yes, this was deliberate. I've used try_from and try_into where the cast could fail, so we will get crashes if things go wrong.

vext01:
> Clippy

We've never used it here. There's so much on our plate here that this kind of polishing has taken back seat.

ykpack/src/types.rs
vext01: Hrm.

It doesn't work so well because  map is lazy and also returns a value:
```
555 | |             Statement::IStore(dest, _src) => dest.local().map(|l| ret.push(l)),
    | |                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `()`, found enum `std::option::Option`
```

Is there a way to ensure the map is consumed and to (concisely) ignore the return value?

yktrace/src/tir.rs
vext01: I'd agree if there weren't a `Return` case.

ykcompile/src/lib.rs
ptersilie: Fixed in 1ea1bec . Note that this also requires changes in ykrustc: 5b96b1b

ykpack/src/types.rs
vext01: I've merged the cases here at least.

ykpack/src/types.rs
ltratt: Just to be clear: we never use `as`, right?

ltratt:
I suggest doing a Clippy PR after this one. In my experience, 90% of its suggestions are improvements, and it's easy to turn off the other 10%. I've found it easier if I do semi-regular Clippy runs.

ykpack/src/types.rs
vext01: Not for offsets, no.

(We use `as` in some places in the codegen where keeping the underlying bit pattern is correct and desired).

vext01:
Can we do clippy as the immediate next pr?

yktrace/src/tir.rs
ltratt: Oops!

ykpack/src/types.rs
ltratt: OK.

ltratt:
> Can we do clippy as the immediate next pr?

Fine by me!

vext01:
If you are ok with those last two commits, I think we are done here.

ltratt:
Please squash.

vext01:
Splat. Time to look at the companion PR.

ltratt:
I forgot to ask: did we audit all the `FIXME`s?

ptersilie:
I had a look at the FIXMEs. Many of them talk about optimisations. And the remaining few I'm not sure can be sensibly tested for with asserts.

ltratt:
> And the remaining few I'm not sure can be sensibly tested for with asserts.

@vext01 have you had a look at these non-optimisation FIXMEs too?

vext01:
I did a review. There's only one which doesn't already have an assert (or other kind of crash):

> FIXME. Check for overflow here. Examine either CF or OF depending on signedness.

This should be fixed yes.

ltratt:
Should it be fixed in this PR?

vext01:
I'll do it in this PR, yep.

vext01:
I've sneaked in some important fixes found by clippy. The offending code was introduced in this PR so it makes sense to nip them in the bud here. See 0b02db2ca789dc91d0aaf6f8c934484c223f23a2

ptersilie:
Hah, those checks are indeed bogus!

ykcompile/src/lib.rs
```diff
@@ -579,7 +579,7 @@ impl<TT> TraceCompiler<TT> {
             Location::Mem(..) => todo!(),
             Location::Const { val, .. } => {
                 let val = val.i64_cast();
-                if val as u32 > u32::MAX {
+                if val > u32::MAX as i64 {
```
bjorn3, line 582: You could also use

```suggestion
                if val > i64::from(u32::MAX) {
```

vext01: Is there any advantage to doing it that way?

bjorn3: Clippy warns against `as` as it is easy to accidentally write a truncating cast.

vext01: Thanks. Fixed in ac540b6

ltratt: I agree with Clippy and Bjorn: `as` is not a good feature and we should avoid its use whenever possible.

ltratt:
Clippy is useful!

Is it time to prime?

ptersilie:
Do we need another squash?

vext01:
Yes, am I OK to squash?

ltratt:
Please go ahead!

vext01:
splat.

PR #141, opened 2020-11-05
Use macros for binary operations.

ptersilie:
Generate the functions implementing binary operations (add/sub and
mul/div) using macros.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -43,6 +43,160 @@ lazy_static! {
     static ref PTR_SIZE: u64 = u64::try_from(mem::size_of::<usize>()).unwrap();
 }
 
+macro_rules! binop_add_sub {
+    ($name: ident, $op:expr) => {
+        fn $name(&mut self, dest: &IPlace, size: u64, opnd2: &IPlace, temp_reg: u8) {
+            let src_loc = self.iplace_to_location(opnd2);
+            match src_loc {
+                Location::Register(r) => match size {
+                    1 => {
+                        dynasm!(self.asm
+                            ; $op Rb(temp_reg), Rb(r)
+                        );
+                    }
+                    2 => {
+                        dynasm!(self.asm
+                            ; $op Rw(temp_reg), Rw(r)
+                        );
+                    }
+                    4 => {
+                        dynasm!(self.asm
+                            ; $op Rd(temp_reg), Rd(r)
+                        );
+                    }
+                    8 => {
+                        dynasm!(self.asm
+                            ; $op Rq(temp_reg), Rq(r)
+                        );
+                    }
+                    _ => unreachable!(format!("{}", SIR.ty(&dest.ty()))),
+                },
+                Location::Mem(..) => todo!(),
+                Location::Const { val, .. } => {
+                    let val = val.i64_cast();
+                    match size {
+                        1 => {
+                            dynasm!(self.asm
+                                ; $op Rb(temp_reg), val as i8
+                            );
+                        }
+                        2 => {
+                            dynasm!(self.asm
+                                ; $op Rw(temp_reg), val as i16
+                            );
+                        }
+                        4 => {
+                            dynasm!(self.asm
+                                ; $op Rd(temp_reg), val as i32
+                            );
+                        }
+                        8 => {
+                            if i32::try_from(val).is_err() {
+                                // FIXME Work around x86_64 encoding limitations (no imm64 operands).
+                                todo!();
+                            } else {
+                                dynasm!(self.asm
+                                    ; $op Rq(temp_reg), val as i32
+                                );
+                            }
+                        }
+                        _ => unreachable!(format!("{}", SIR.ty(&dest.ty()))),
+                    }
+                }
+                Location::Indirect { .. } => todo!(),
+                Location::NotLive => todo!(),
+            }
+        }
+    }
+}
+
+macro_rules! binop_mul_div {
+    ($name: ident, $op:expr) => {
+        fn $name(&mut self, dest: &IPlace, size: u64, opnd2: &IPlace, temp_reg: u8) {
+            // mul and div overwrite RAX, RDX, so save them first.
+            dynasm!(self.asm
+                ; push rax
+                ; push rdx
+                ; xor rdx, rdx
+            );
+            dynasm!(self.asm
+                ; mov rax, Rq(temp_reg)
+            );
+            let src_loc = self.iplace_to_location(opnd2);
+            match src_loc {
+                Location::Register(r) => match size {
+                    1 => {
+                        dynasm!(self.asm
+                            ; $op Rb(r)
+                        );
+                    }
+                    2 => {
+                        dynasm!(self.asm
+                            ; $op Rw(r)
+                        );
+                    }
+                    4 => {
+                        dynasm!(self.asm
+                            ; $op Rd(r)
+                        );
+                    }
+                    8 => {
+                        dynasm!(self.asm
+                            ; $op Rq(r)
+                        );
+                    }
+                    _ => unreachable!(format!("{}", SIR.ty(&dest.ty()))),
+                },
+                Location::Mem(..) => todo!(),
+                Location::Const { val, .. } => {
+                    let val = val.i64_cast();
+                    match size {
+                        1 => {
+                            dynasm!(self.asm
+                                ; mov Rb(temp_reg), val as i8
+                                ; $op Rb(temp_reg)
+                            );
+                        }
+                        2 => {
+                            dynasm!(self.asm
+                                ; mov Rw(temp_reg), val as i16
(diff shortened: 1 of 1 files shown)


ptersilie:
Just noticed that we are missing some changes from master due to the rebase. Fixing.

ykcompile/src/lib.rs
```diff
@@ -43,6 +43,130 @@ lazy_static! {
     static ref PTR_SIZE: u64 = u64::try_from(mem::size_of::<usize>()).unwrap();
 }
 
+macro_rules! binop_as {
```
vext01, line 46: You should be able to do all of this in one macro where you match over the operation name?

ykcompile/src/lib.rs
```diff
@@ -43,6 +43,130 @@ lazy_static! {
...
+                        );
+                    }
+                    _ => unreachable!(format!("{}", SIR.ty(&dest.ty()))),
+                }
+            }
+            Location::Indirect { .. } => todo!(),
+            Location::NotLive => todo!(),
+        }
+    }
+    }
```
vext01, line 95: Weird indent.

ptersilie: Whoops

ykcompile/src/lib.rs
ptersilie: I guess I could, but thought it might be a bit messy. I can try to see how it looks.

ykcompile/src/lib.rs
ptersilie: Fixed in a4ba0d6

ykcompile/src/lib.rs
ptersilie: I had a quick look and it will make that function messy and hard to read so I think I prefer them being split up.

ykcompile/src/lib.rs
```diff
@@ -43,6 +43,130 @@ lazy_static! {
...
+                        dynasm!(self.asm
+                            ; $op Rq(*TEMP_REG), Rq(r)
+                        );
+                    }
+                    _ => unreachable!(format!("{}", SIR.ty(&dest.ty()))),
+                },
+                Location::Mem(..) => todo!(),
+                Location::Const { val, .. } => {
+                    let val = val.i64_cast();
+                    if val > i64::from(u32::MAX) {
```
bjorn3, line 63:
```suggestion
                    if val > i64::from(u32::MAX) || val < 0 {
```

ykcompile/src/lib.rs
vext01: OK. Can we make the names more descriptive though? `as` and `md` confused me.

vext01:
the macros implicitly work on the temporary register, which is something I accidentally introduced. I think the macro should take as input a target register, thus making it more explicit and flexible.

ykcompile/src/lib.rs
ptersilie: How about `binop_add_sub` and `binop_mul_div`?

vext01: sounds good.

ykcompile/src/lib.rs
ptersilie: I think this would lead to `-3u8` also ending up in this branch. Did you maybe mean:
```rust
if val > i64::from(u32::MAX) || val < -i64::from(u32::MAX) {
```

ptersilie: or just
```rust
if abs(val) > i64::from(u32::MAX) {
```

ptersilie:
> the macros implicitly work on the temporary register, which is something I accidentally introduced. I think the macro should take as input a target register, thus making it more explicit and flexible.

Do you want the macros to take the target register, or the function that is generated by the macro?

vext01:
I think the macro should take a target register argument.

ptersilie:
Huh, that's not what I expected your answer would be. Can you briefly explain your reasoning?

vext01:
Ah, sorry, I misunderstood.

The generated function should accept a target register.

ptersilie:
Okay, shouldn't we, for consistency, then make all functions take the temporary register as an argument then?

vext01:
add/sub/mul/div should accept an arbitrary register number as a u8.

ptersilie:
Yes, I understood that. But shouldn't all the other functions that work with `TEMP_REG` do the same then?

ykcompile/src/lib.rs
ltratt: When it comes to things like `-i64::from(u32::MAX)`, it's worth knowing that `(uX:MAX)` and `uX:MIN` are not symmetric in twos complement.

vext01:
Only if the value in the temp register is still to be used when the call finishes.

Which functions do you have in mind?

ptersilie:
> Only if the value in the temp register is still to be used when the call finishes.
> 
> Which functions do you have in mind?

I got it now. Fixing.

ykcompile/src/lib.rs
vext01: I think what @bjorn3 is saying is that if the number is negative then by definition its 64-bit representation won't fit in a 32-bit register. In signed integers the top bit has to be `1` for negative numbers.

I'm starting to dislike this signed/unsigned malarky here. We don't actually care about the signedness of the data in `val` here, only the bit pattern.

For this check, we just want to know if any higher-order bits are set. We can take signedness out of the picture with:
```
if val >> 32 != 0 { ... }
```

bjorn3: You could also do `u32::try_from(val).is_err()` I think.

ptersilie: @vext01 I understood what @bjorn3 was saying, but in this case `val` can be a negative `i8`, e.g. `-3u8`, which we then cast to a `i64` which is `< 0`. We then end up in this `todo()` branch whereas we can just use the normal instructions below.

@bjorn3 `u32::try_from(val).is_err()` sounds like it would work better.

ltratt: Can I ask a dumb question: what are we trying to achieve in this part of the code?

vext01: We are trying to figure out if the 64-bit constant can fit in a 32-bit register, as in x86_64 you can't move a whole 64-bit constant into a register in one go.

@ptersilie It's right that we hit the todo for -3 as its two's compliment encoding can't fit in 32-bits.

-3i64 is `11111111 11111111 11111111 11111111 11111111 11111111 11111111 11111101`.

Notice that the case for 8-bytes below only actually loads an i32 DWORD. That's because we are sure that the top 32-bits are all zero, so we can load it all in one go.

ltratt: Isn't this something that dynasm does for us? [Obviously I'm a bit clueless!] Otherwise we have to do a lot of bit-checking stuff and it's quite fiddly.

vext01: Afraid not. It doesn't do any instruction selection for you. I don't know of an assembler that does.

ltratt: Personally I'd be totally fine if we currently `todo` for everything that's not an easy case. I would define easy cases as including, for example, `-1` so it's not *totally* easy, but I'm not sure I care about encoding `u64::MIN` as a constant at the moment.

vext01: Yep. That's what the todo is doing. We can implement those later, and it's better to implement them when some test actually touches that case anyway.

ltratt: Makes sense!

vext01: OK, let's see if Lukas agrees.

ptersilie: > @ptersilie It's right that we hit the todo for -3 as its two's compliment encoding can't fit in 32-bits.

But we shouldn't hit it for -3i8, which we would do `if val >> 32 != 0` or `val < 0`, since we cast to i64 first, so `-3i8 = 0xfd` becomes `0xfffffffffffffffd`. But I think @bjorn3 suggestion to use `u32::try_from(val).is_err()` should work.

ptersilie:
Okay, if those last few commits are good we can squash, rebase, and merge.

ykcompile/src/lib.rs
```diff
@@ -45,28 +45,28 @@ lazy_static! {
 
 macro_rules! binop_add_sub {
     ($name: ident, $op:expr) => {
-        fn $name(&mut self, dest: &IPlace, size: u64, opnd2: &IPlace) {
+        fn $name(&mut self, dest: &IPlace, size: u64, opnd2: &IPlace, temp_reg: u8) {
```
vext01, line 48: It seems odd that `dest` is only used for its type.

I have an idea for making this a bit more sane, but I can try in a fresh PR. We really should move on with this one :P

vext01:
Please squash and rebase.

ptersilie:
Done!

PR #143, opened 2020-11-06
Tidy binops

vext01:
This tidies the binary operations and makes them more explicit.

Bonus commit to crash if the operands are not integers.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -43,33 +43,36 @@ lazy_static! {
     static ref PTR_SIZE: u64 = u64::try_from(mem::size_of::<usize>()).unwrap();
 }
 
+/// Generates functions for add/sub-style operations.
+/// The first operand must be in a register.
 macro_rules! binop_add_sub {
     ($name: ident, $op:expr) => {
-        fn $name(&mut self, dest: &IPlace, size: u64, opnd2: &IPlace, temp_reg: u8) {
-            let src_loc = self.iplace_to_location(opnd2);
-            match src_loc {
+        fn $name(&mut self, opnd1_reg: u8, opnd2: &IPlace) {
+            let size = SIR.ty(&opnd2.ty()).size();
+            let opnd2_loc = self.iplace_to_location(opnd2);
+            match opnd2_loc {
                 Location::Register(r) => match size {
                     1 => {
                         dynasm!(self.asm
-                            ; $op Rb(temp_reg), Rb(r)
+                            ; $op Rb(opnd1_reg), Rb(r)
                         );
                     }
                     2 => {
                         dynasm!(self.asm
-                            ; $op Rw(temp_reg), Rw(r)
+                            ; $op Rw(opnd1_reg), Rw(r)
                         );
                     }
                     4 => {
                         dynasm!(self.asm
-                            ; $op Rd(temp_reg), Rd(r)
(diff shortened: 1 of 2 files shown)


ykcompile/src/lib.rs
```diff
@@ -43,51 +43,54 @@ lazy_static! {
     static ref PTR_SIZE: u64 = u64::try_from(mem::size_of::<usize>()).unwrap();
 }
 
+/// Generates functions for add/sub-style operations.
+/// The first operand must be in a register.
```
ltratt, line 47: Can we add a `debug_assert` to this effect?

vext01:
It's (somewhat) enforced by the type of the parameter. I added the comment
for informational value.

ltratt:
> It's (somewhat) enforced by the type of the parameter.

Why "somewhat"?

vext01:
Dynasm doesn't have a dedicated type to identify registers. It uses u8.

But the comment is trying to point out that you can't pass two IPlaces as
the operands. One operand has to be a register (i.e. u8).

ltratt:
OK, makes sense!

ykcompile/src/lib.rs
```diff
@@ -695,12 +703,14 @@ impl<TT> TraceCompiler<TT> {
         // 3) Move the result to where it is supposed to live.
         // If it is a checked operation, then we have to build a (value, overflow-flag) tuple.
         let mut dest_loc = self.iplace_to_location(dest);
+        let size = opnd1_ty.size();
```
ptersilie, line 725: Not quite sure where this comes from and how it compiles.

ykcompile/src/lib.rs
```diff
@@ -669,6 +677,14 @@ impl<TT> TraceCompiler<TT> {
     }
 
     fn c_binop(&mut self, dest: &IPlace, op: BinOp, opnd1: &IPlace, opnd2: &IPlace, checked: bool) {
+        let opnd1_ty = SIR.ty(&opnd1.ty());
```
vext01, line 680: `opnd1_ty` is defined here.

ykcompile/src/lib.rs
```diff
@@ -677,6 +677,14 @@ impl<TT> TraceCompiler<TT> {
...
+        debug_assert!(opnd1_ty == SIR.ty(&opnd2.ty()));
+
+        // For now this whole function assumes we are operating on integers.
```
ptersilie, line 683: Is this true? I think `c_condition` doesn't really care about the types of the operands.

ykcompile/src/lib.rs
ptersilie: Ah, that's the next commit. I assume you'll squash later? Otherwise the first commit won't even compile.

ykcompile/src/lib.rs
vext01: It uses `cmp` which can only be used to compare integers.

For example, to compare floats or doubles, we'd have to use one of the other comparison instructions.

ykcompile/src/lib.rs
vext01: Oops. I didn't realise that. Yes, we can squash it all into one.

vext01:
Can I squash?

ptersilie:
Please squash.

vext01:
splat.

PR #147, opened 2020-11-10
The tags should be in the right-most, not left-most bits.

ltratt:


ltratt:
I can't easily test this locally, so hopefully bors can confirm whether this breaks anything!

bors try

ltratt:
bors try

ltratt:
bors try

ltratt:
bors try

ltratt:
bors try

ykrt/src/mt.rs
```diff
@@ -8,25 +8,24 @@ use std::{
...
-// tag check.
-const PHASE_TAG: u32 = 0b11 << 30; // All of the other PHASE_ tags must fit in this.
-const PHASE_COMPILED: u32 = 0b00 << 30;
-const PHASE_TRACING: u32 = 0b01 << 30;
-const PHASE_COUNTING: u32 = 0b10 << 30; // The value specifies the current hot count.
+// (optionally) a value. We expect the most commonly encountered tag at run-time is PHASE_COMPILED
+// whose value is a pointer to memory. By also making that tag 0b00, we allow that index to be
+// accessed without any further operations after the initial tag check.
+const PHASE_TAG: usize = 0b11; // All of the other PHASE_ tags must fit in this.
+const PHASE_COMPILED: usize = 0b00;
```
vext01, line 26: Should we put an assertion somewhere that checks that the alignment of our trace code buffers guarantees enough space for our tags?

It almost seems silly because dynasm using `mmap`, which page aligns. Page alignment varies, but is usually 4096, which would give us a whopping 12 bits.

ltratt: > Should we put an assertion somewhere that checks that the alignment of our trace code buffers guarantees enough space for our tags?

I don't think that's possible (how do we actually know the alignment? all we can do is document it and hope) or necessary (in practise, the scheme we're using is tried and tested; or, at least, the scheme in this PR :p ).

vext01: You can query the page size at runtime and compute the alignment.

vext01:
I was expecting a change to the portion of code which increments the hot count, as you'd need to either shift and add, or add 4 (instead of one). Wouldn't you?

ptersilie:
Okay, this should work now. @vext01 can you review?

vext01:
That looks good to me.

I suppose it's not convenient to add 4 due to the comparison with the hot threshold. Unless you compare the threshold shifted by 4 too. The latter is probably a bit faster, but at this early stage I'm kind of indifferent.

vext01:
bors try

vext01:
needs `cargo fmt`

ptersilie:
Done.

vext01:
bors try

vext01:
Please squash.

ykrt/src/mt.rs
```diff
@@ -248,7 +248,7 @@ impl MTThread {
             let lp = pack.load(Ordering::Acquire);
             match lp & PHASE_TAG {
                 PHASE_COUNTING => {
-                    let count = lp & !PHASE_TAG;
+                    let count = (lp & !PHASE_TAG) >> 2;
```
ltratt, line 251: `2` should be a constant referenced at the top (e.g. `PHASE_NUM_BITS` or whatever).

vext01:
OK, so don't squash just yet. Address Laurie's comment first.

ptersilie:
Want to check that commit first before squashing?

ltratt:
Works for me!

vext01:
Please rebase and squash.

ptersilie:
Squashed and rebased.

Issue #115, opened 2020-09-04
Give callee save registers to the register allocator.


vext01:
The register allocator could use loads more registers if we included the callee save registers.

To implement this, we'd have to save and restore the callee save registers in our trace code.

PR #149, opened 2020-11-12
Add callee save registers and RAX to the register pool.

### LINKED ISSUE #114: Give RAX back to the register allocator.
RAX is currently reserved as a temporary register, but we now have a better way to acquire temps, and thus we could give RAX back to the allocator.

vext01:
Fixes #114 and #115.

I thought this would be a 15 minute job. How wrong I was. Here be dragons.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -33,12 +33,12 @@ lazy_static! {
     // Registers that are caller-save as per the Sys-V ABI.
     // Note that R11 is also caller-save, but it's our temproary register and we never want to
     // preserve its value across calls.
-    static ref CALLER_SAVE_REGS: [u8; 8] = [RAX.code(), RDI.code(), RSI.code(), RDX.code(),
+    static ref CALLER_SAVED_REGS: [u8; 8] = [RAX.code(), RDI.code(), RSI.code(), RDX.code(),
                                             RCX.code(), R8.code(), R9.code(), R10.code()];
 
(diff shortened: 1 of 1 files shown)


ykcompile/src/lib.rs
```diff
@@ -34,10 +34,15 @@ lazy_static! {
     static ref CALLER_SAVE_REGS: [u8; 8] = [RDI.code(), RSI.code(), RDX.code(), RCX.code(),
                                             R8.code(), R9.code(), R10.code(), R11.code()];
 
+    // Registers that are callee-save as per the Sys-V ABI.
+    // Note that RBP is also callee-save, but is handled specially.
+    static ref CALLEE_SAVE_REGS: [u8; 5] = [RBX.code(), R12.code(), R13.code(),
```
ptersilie, line 41: Should we rename this to `callee_saved_regs` to avoid confusion. Someone might think this means `callee-safe` (as in these registers are safe for the callee to use a.k.a caller-saved).

vext01: Seems sensible!

ptersilie:
Rest looks good.

vext01:
Fixed.

I don't mind that staying as a separate commit, since we already had a const using the `SAVE` idiom.

ptersilie:
Okay, please squash.

vext01:
None required. Each commit is a self-contained unit (for once!).

PR #151, opened 2020-11-16
Implement PHASE_COMPILED in the control point.

ptersilie:


ykcompile/src/lib.rs
```diff
@@ -273,6 +273,10 @@ impl<TT> CompiledTrace<TT> {
     fn exec_trace(&self, t_fn: fn(&mut TT) -> bool, args: &mut TT) -> bool {
         t_fn(args)
     }
+
+    pub fn ptr(&self) -> *const u8 {
```
ltratt, line 277: This might be dodgy unless `self.mc` explicitly guarantees a stable pointer at all points. Does it make such a guarantee?

ykrt/src/mt.rs
```diff
@@ -223,20 +224,31 @@ impl MTThread {
...
         // this thread's tracer.
         if let Some(loc) = loc {
-            self._control_point(loc);
+            if let Some(func) = self._control_point::<I>(loc) {
+                if self.exec_trace(func, inputs) {
+                    // Trace succesfully executed.
+                    return;
+                } else {
+                    // FIXME blackholing
+                    panic!("Guard failure!")
```
ltratt, line 240: `todo!` would be better here?

ykcompile/src/lib.rs
```diff
@@ -1461,18 +1465,25 @@ impl<TT> TraceCompiler<TT> {
                     );
                 }
                 Location::Mem(..) => todo!(),
-                Location::Indirect { ptr, off } => match ptr {
```
vext01, line 1464: It's hard to see what this change was for. Can you elaborate?

ykrt/src/mt.rs
```diff
@@ -283,25 +295,40 @@ impl MTThread {
...
+                            .unwrap()
+                            .0
+                            .stop_tracing()
+                            .unwrap();
+                        let tir_trace = TirTrace::new(&*SIR, &*sir_trace).unwrap();
+                        let ct = TraceCompiler::<I>::compile(tir_trace);
+                        let ptr = Box::into_raw(Box::new(ct)) as usize; //ct.ptr() as usize;
+
+                        let new_pack = PHASE_COMPILED | (ptr << PHASE_NUM_BITS);
+                        if pack.compare_and_swap(lp, new_pack, Ordering::Release) == lp {
```
ltratt, line 311: At this point I think we know we're the only thread that can try writing to this, in which case I think `pack.store` w/ `Ordering::Release` is sufficient (i.e. I don't think the CAS does anything useful apart from doing a store).

ykrt/src/mt.rs
```diff
@@ -283,25 +295,40 @@ impl MTThread {
...
+                            .unwrap()
+                            .tracer
+                            .take()
+                            .unwrap()
+                            .0
+                            .stop_tracing()
+                            .unwrap();
+                        let tir_trace = TirTrace::new(&*SIR, &*sir_trace).unwrap();
+                        let ct = TraceCompiler::<I>::compile(tir_trace);
+                        let ptr = Box::into_raw(Box::new(ct)) as usize; //ct.ptr() as usize;
```
vext01, line 308: commented code.

ykrt/src/mt.rs
```diff
@@ -283,25 +295,40 @@ impl MTThread {
...
+                        if pack.compare_and_swap(lp, new_pack, Ordering::Release) == lp {
                             break;
                         }
                     } else {
                         break;
                     }
                 }
-                PHASE_COMPILED => break, // FIXME call compiled trace, but don't call step_fn().
+                PHASE_COMPILED => {
+                    if pack.load(Ordering::Acquire) == lp {
```
ltratt, line 319: Why do we need to reload `pack` from memory? Can't we just operate on `lp`?

ykrt/src/mt.rs
```diff
@@ -283,25 +295,40 @@ impl MTThread {
...
-                            // FIXME build TIR and compile. Eventually in a background thread?
+                        let sir_trace = Rc::get_mut(&mut self.inner)
+                            .unwrap()
+                            .tracer
+                            .take()
+                            .unwrap()
+                            .0
+                            .stop_tracing()
+                            .unwrap();
+                        let tir_trace = TirTrace::new(&*SIR, &*sir_trace).unwrap();
```
vext01, line 306: I think we should add a comment with a few words explaining what is going on here. About avoiding drops etc.

ykrt/src/mt.rs
```diff
@@ -407,50 +438,69 @@ mod tests {
...
-            prog: Vec<ByteCode>,
+            prog: Vec<u8>,
             pc: usize,
+            count: usize,
         }
 
-        let interp_step = |mut tio: IO| {
+        #[interp_step]
+        fn dumb_interp_step(tio: &mut IO) {
+            // FIXME make `inputs` a struct. Named fields would be much nicer.
```
vext01, line 457: Kill this line. `tio` is already a struct.

ykrt/src/mt.rs
```diff
@@ -407,50 +438,69 @@ mod tests {
...
-            tio
-        };
+        }
 
-        let mut tio = IO { prog, pc: 0 }; // bytecode, program counter.
+        let mut tio = IO {
+            prog,
+            pc: 0,
+            count: 0,
+        }; // bytecode, program counter.
```
vext01, line 472: I think that comment is a leftover from when tio was a tuple also. Let's kill it?

vext01:
Just a few minor comments, otherwise looks good.

ykcompile/src/lib.rs
ptersilie: I'm not sure. The memory is allocated here: https://github.com/danburkert/memmap-rs/blob/master/src/unix.rs#L49
The rest is just a way to get `mc`, which is private, so that I can transmute the pointer in `PHASE_COMPILED`.

ykrt/src/mt.rs
vext01: Good point.

ykcompile/src/lib.rs
ltratt: OK, since this came from `mmap`, this should be fine. But let's add a comment explaining that, otherwise it'll scare future-me :)

ykcompile/src/lib.rs
ptersilie: Sorry, it was difficult to pull those changes apart again so I left them in a single commit. This change is to get the `dumb_interpreter` test running. It implements the `todo` below.

ykrt/src/mt.rs
ptersilie: Agreed.

ykrt/src/mt.rs
ptersilie: Will try.

ykrt/src/mt.rs
ptersilie: You're right. The atomic shouldn't have changed between this and the previous `load`.

ykcompile/src/lib.rs
ptersilie: Fixed in 7dae376.

ykrt/src/mt.rs
ptersilie: Fixed in eef7b2f.

ykrt/src/mt.rs
ptersilie: Fixed in bc9291b.

ykrt/src/mt.rs
ptersilie: Fixed in bc9291b.

ykrt/src/mt.rs
ptersilie: Fixed in bc9291b.

ykrt/src/mt.rs
ptersilie: Fixed in 789a136.

ykrt/src/mt.rs
ptersilie: Fixed in 806a168.

ykrt/src/mt.rs
ptersilie: Fixed in 5638b7f.

ptersilie:
Think that addresses all comments.

ykrt/src/mt.rs
```diff
@@ -305,6 +305,10 @@ impl MTThread {
                             .unwrap();
                         let tir_trace = TirTrace::new(&*SIR, &*sir_trace).unwrap();
                         let ct = TraceCompiler::<I>::compile(tir_trace);
+                        // Now that we've got a compile trace, store it within the tag. However,
```
vext01, line 308: "store it within the *tag*".

Reading the code, it seems that the "pack" is the name we gave the whole `u64` and the "tag" is the two bits reserved to identify the rest of the pack.

"compile" -> "compile*d*".

I think a small rewording would make it easier to read. How about:

> Now that we have a compiled trace, we need to put a pointer to it inside the corresponding Location's pack. To inhibit ExecutableBuffer's Drop, we move it into a box and forget it: the pointer in the pack is now the sole handle on the trace.

ptersilie: Fixed in 4152f09.

vext01:
LGTM. @ltratt are you happy?

ptersilie:
bors try

ptersilie:
> I think I'd like to be more explicit here. Something like `Return a pointer the mmap'd block of memory containing the trace. The underlying data is guaranteed never to move in memory.` or similar perhaps?

Fixed.

ltratt:
I'm happy!

vext01:
Please squash.

ptersilie:
Squashed.

PR #152, opened 2020-11-18
Add a simple trace debugging facility.

vext01:
Companion PR ~~coming~~: https://github.com/softdevteam/ykrustc/pull/144.

We add a special function, yktrace::trace_debug(msg: &'static str).

When the interpreter author calls this special function, the SIR generator
recognises this and, instead of emitting a regular Call terminator, inserts a
special TraceDebugCall terminator. This terminator contains only the bare
minimum for the TIR compiler to later translate the terminator into a Debug
statement, which when printed, looks like a comment in the trace.

For example if the interpreter author sprinkled calls like this in their code:
```
...
trace_debug("Add 10");
...
trace_debug("Multiply 2");
...
```

Then the corresponding stringified trace might look like:
```
  ...
  // Add 10
  $6 = *($1+0)+8 + 10usize (checked)
  $7 = $6
  dead($6)
  guard($7+8, bool(false))
  *($1+0)+8 = $7
  dead($7)
  $0 = ()
  guard(*$1+0, integer(2))
  // Multiply 2
  $2 = *($1+0)+8 * 2usize (checked)
  ...
```

The idea is that interpreter authors could put these debugging statements at key
points in their implementation (e.g. at the start of each bytecode) to aid
debugging. They would want to compile-time guard the calls though, so that the
instrumentation is opt-in.

### DIFF
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -331,8 +331,12 @@ impl Display for Local {
 
 /// Bits in the `flags` bitfield in `Body`.
 pub mod bodyflags {
+    /// This function is annotated #[do_not_trace].
     pub const DO_NOT_TRACE: u8 = 1;
+    /// This function is annotated #[interp_step].
     pub const INTERP_STEP: u8 = 1 << 1;
+    /// This function is yktrace::trace_debug.
+    pub const TRACE_DEBUG: u8 = 1 << 2;
 }
 
 /// The definition of a local variable, including its type.
@@ -539,11 +543,13 @@ pub enum Statement {
     /// Note that locals are implicitly live at first use.
     StorageDead(Local),
     /// A (non-inlined) call from a TIR trace to a binary symbol using the system ABI. This does
-    /// not appear in SIR.
+    /// not appear in SIR. Not to be confused with Terminator::Call in SIR.
     Call(CallOperand, Vec<IPlace>, Option<IPlace>),
     /// Cast a value into another. Since the cast type and the destination type are the same, we
     /// only need the latter.
     Cast(IPlace, IPlace),
+    /// A debug marker. This does not appear in SIR.
+    Debug(String),
     /// Any unimplemented lowering maps to this variant.
     /// The string inside is the stringified MIR statement.
     Unimplemented(String),
@@ -574,7 +580,7 @@ impl Statement {
                     Self::maybe_push_local(&mut ret, dest.local());
                 }
             }
-            Statement::StorageDead(_) | Statement::Unimplemented(_) => (),
+            Statement::StorageDead(_) | Statement::Debug(_) | Statement::Unimplemented(_) => (),
         }
         ret
     }
@@ -622,7 +628,7 @@ impl Statement {
                 Self::maybe_push_local(&mut ret, dest.local());
                 Self::maybe_push_local(&mut ret, src.local());
             }
-            Statement::Unimplemented(_) => (),
+            Statement::Unimplemented(_) | Statement::Debug(_) => (),
         }
         ret
     }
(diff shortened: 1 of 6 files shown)


vext01:
Just noticed a typo in the commit message. Will fix at the end.

yktrace/src/tir.rs
```diff
@@ -617,11 +627,34 @@ impl TirOp {
...
+        let matcher = FMBuilder::new(ptn)
+            .unwrap()
+            .name_matcher(Some((ptn_re, text_re)))
+            .distinct_name_matching(true)
+            .build()
+            .unwrap();
+
+        let res = matcher.matches(&format!("{}", tt));
+        if let Err(e) = res {
+            eprintln!("{}", e); // Visible when tests run with --nocapture.
```
ptersilie, line 653: Doesn't `panic` print this either way, `capture` or `nocapture`? Why do we need the `eprint`?

yktrace/src/tir.rs
```diff
@@ -640,7 +673,67 @@ mod tests {
         black_box(work(&mut io));
         let sir_trace = tracer.stop_tracing().unwrap();
         let tir_trace = TirTrace::new(&*SIR, &*sir_trace).unwrap();
+        println!("{}", tir_trace);
```
ptersilie, line 676: left over print.

ykpack/src/types.rs
```diff
@@ -892,6 +899,12 @@ pub enum Terminator {
         /// The return value and basic block to continue at, if the call converges.
         destination: Option<(IPlace, BasicBlockIndex)>,
     },
+    /// A call to yktrace::trace_debug. This is converted into a Statement::Debug at TIR
+    /// compilation time.
+    TraceDebugCall {
+        msg: String,
+        destination: BasicBlockIndex,
```
ptersilie, line 906: What's the point of `destination`? We never seem to use it.

vext01: You're right that we don't. I included it because I hope that one day we will add a verification step to the TIR compiler which checks that the next location is possible according to the SIR control flow.

We can kill this for now if you like though.

yktrace/src/tir.rs
vext01: I ran some experiments.

the panic argument is always printed, but it doesn't print the error in this case. It says:
```
thread 'tir::tests::trace_debug_tir' panicked at 'Box<Any>', yktrace/src/tir.rs:654:13
```

i.e. not very useful. Will fix.

ykpack/src/types.rs
ptersilie: That's fine then. Leave it in.

ptersilie:
Please squash.

vext01:
splat.

vext01:
And rebased.

yktrace/src/tir.rs
```diff
@@ -619,11 +629,33 @@ impl TirOp {
...
+        let matcher = FMBuilder::new(ptn)
+            .unwrap()
+            .name_matcher(Some((ptn_re, text_re)))
+            .distinct_name_matching(true)
+            .build()
+            .unwrap();
+
+        let res = matcher.matches(&format!("{}", tt));
+        if let Err(e) = res {
+            panic!(format!("{}", e));
```
bjorn3, line 655: This should be `panic!("{}", e)`. The panic! variant you are using will likely be deprecated in the future.

yktrace/src/tir.rs
bjorn3: `panic!(e)` will directly throw `Box::new(e) as Box<Any>` instead of `Box::new(format!("{}", e)) as Box<Any>` like `panic!("{}", e)` does.

vext01: Thanks. Fixed in c38fa973f372c47022b3c66d122b5bc334b2d2af

vext01:
OK to squash again @ptersilie ?

ptersilie:
Yes, please squash.

vext01:
Splat.

PR #157, opened 2020-11-25
Add SIR interpreter.

ptersilie:
This PR adds an early version of our SIR interpreter. There's obviously much we can improve here, and this also doesn't handle nested frames yet. I tried to keep the first version (and PR) of this as small as possible, to get any early design flaws out of the way first before we move on.

One obvious improvement would be the use of vectors for the storage of data. We currently have two alternative ideas:

1) Use arrays instead of vecs which requires some lifetime shenanigans.
2) Allocate one big chunk of memory (i.e. [u8]) for each frame, whose size is equal to the size of all locals in the frame (which we can pre-compute), and then an `Object` is simply a slice into that chunk.

### DIFF
--- a/ykbh/Cargo.toml
+++ b/ykbh/Cargo.toml
@@ -4,9 +4,6 @@ version = "0.1.0"
 authors = ["Lukas Diekmann <lukas.diekmann@gmail.com>"]
 edition = "2018"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
 [dependencies]
 yktrace = { path = "../yktrace" }
 ykpack = { path = "../ykpack" }
-ykcompile = { path = "../ykcompile" }
--- a/ykbh/Cargo.toml
+++ b/ykbh/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "ykbh"
+version = "0.1.0"
+authors = ["Lukas Diekmann <lukas.diekmann@gmail.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+yktrace = { path = "../yktrace" }
+ykpack = { path = "../ykpack" }
+ykcompile = { path = "../ykcompile" }
--- a/ykbh/src/lib.rs
+++ b/ykbh/src/lib.rs
@@ -221,10 +221,7 @@ mod tests {
     use yktrace::sir::SIR;
 
     fn interp(fname: &str, tio: usize) {
-        let body = match SIR.bodies.get(fname) {
-            Some(b) => b,
-            None => panic!("No SIR")
-        };
+        let body = SIR.bodies.get(fname).expect("No SIR");
         let mut vars: Vec<Option<Object>> = vec![None; body.local_decls.len()];
         vars[1] = Some(Object::from_usize(tio));
         let mut si = SIRInterpreter::new(vars);
--- a/ykbh/src/lib.rs
+++ b/ykbh/src/lib.rs
@@ -1,6 +1,6 @@
 use ykpack::{self, Local, Terminator, Statement, IPlace, Constant, ConstantInt, UnsignedInt};
 use yktrace::sir::SIR;
-use std::convert::TryInto;
+use std::convert::{TryInto, TryFrom};
 
 const ZST: Object = Object { data: Vec::new() };
 
@@ -91,7 +91,7 @@ impl SIRInterpreter {
 
     /// Get the Object referenced by a Local.
     fn get_var(&self, local: &Local) -> &Option<Object> {
-        &self.vars[local.0 as usize]
+        &self.vars[usize::try_from(local.0).expect("Can't convert local to usize.")]
     }
 
     /// Implements the Store statement.
@@ -106,7 +106,7 @@ impl SIRInterpreter {
             IPlace::Val { local, off, ty: _ty } => {
                 // Store into a local.
                 if *off == 0 {
-                    self.vars[local.0 as usize] = Some(src);
+                    self.vars[usize::try_from(local.0).expect("Can't convert local to usize.")] = Some(src);
                 } else {
                     todo!()
                 }
@@ -114,7 +114,7 @@ impl SIRInterpreter {
             IPlace::Indirect { ptr, off, ty: _ty } => {
                 // Write to a pointer.
                 if let Some(obj) = self.get_var(&ptr.local) {
-                    let dstptr = obj.to_usize() + ptr.off as usize + *off as usize;
+                    let dstptr = obj.to_usize() + usize::try_from(ptr.off).expect("Can't convert offset to usize.") + usize::try_from(*off).expect("Can't convert offset to usize.");
                     self.store_raw(dstptr, src);
                 }
             }
@@ -141,9 +141,9 @@ impl SIRInterpreter {
                 } else {
                     // Copy data from an offset, e.g. $1 = $2+2.
                     if let Some(obj) = self.get_var(local) {
-                        let ptr = obj.data.as_ptr() as usize + *off as usize;
+                        let ptr = obj.data.as_ptr() as usize + usize::try_from(*off).expect("Can't convert offset to usize.");
                         let size = SIR.ty(ty).size();
-                        Object::from_ptr(ptr, size as usize)
+                        Object::from_ptr(ptr, usize::try_from(size).expect("Can't convert size to usize."))
                     } else {
                         unreachable!()
                     }
@@ -152,9 +152,9 @@ impl SIRInterpreter {
             IPlace::Indirect { ptr, off, ty } => {
                 // Dereference a pointer and copy the data it points to.
                 if let Some(obj) = self.get_var(&ptr.local) {
-                    let dstptr = obj.to_usize() + ptr.off as usize + *off as usize;
+                    let dstptr = obj.to_usize() + usize::try_from(ptr.off).expect("Can't offset to usize.") + usize::try_from(*off).expect("Can't convert offset to usize.");
                     let size = SIR.ty(ty).size();
-                    Object::from_ptr(dstptr, size as usize)
+                    Object::from_ptr(dstptr, usize::try_from(size).expect("Can't convert size to usize."))
                 } else {
                     unreachable!()
                 }
@@ -195,7 +195,7 @@ impl SIRInterpreter {
                     // The pointer to the vector should remain valid, since no changes to the
                     // vector later on will change its size, and so it won't be reallocated.
                     // FIXME: Check this is true!
-                    let ptr = obj.data.as_ptr() as usize + *off as usize;
+                    let ptr = obj.data.as_ptr() as usize + usize::try_from(*off).expect("Can't convert offset to usize.");
                     Object::from_usize(ptr)
                 } else {
                     unreachable!()
@@ -203,7 +203,7 @@ impl SIRInterpreter {
             }
             IPlace::Indirect { ptr, off, ty: _ty } => {
                 if let Some(obj) = self.get_var(&ptr.local) {
-                    let ptr = obj.to_usize() + ptr.off as usize + *off as usize;
+                    let ptr = obj.to_usize() + usize::try_from(ptr.off).expect("Can't convert offset to usize.") + usize::try_from(*off).expect("Can't convert offset to usize.");
                     Object::from_usize(ptr)
                 } else {
                     unreachable!()
--- a/ykbh/src/lib.rs
+++ b/ykbh/src/lib.rs
@@ -1,6 +1,3 @@
-#![feature(test)]
-extern crate test;
-
 use ykpack::{self, Local, Terminator, Statement, IPlace, Constant, ConstantInt, UnsignedInt};
 use yktrace::sir::SIR;
 use std::convert::TryInto;
@@ -17,7 +14,7 @@ impl Object {
 
     /// Creates a new Object from a `usize`.
     fn from_usize(ptr: usize) -> Object {
-        let data = ptr.to_be_bytes();
+        let data = ptr.to_ne_bytes();
         Object {
             data: Vec::from(data)
         }
@@ -32,7 +29,7 @@ impl Object {
 
     /// Converts the objects data back to a `usize`.
     fn to_usize(&self) -> usize {
-        usize::from_be_bytes(self.data.as_slice().try_into().expect("Unexpected data length."))
+        usize::from_ne_bytes(self.data.as_slice().try_into().expect("Unexpected data length."))
     }
 
     /// Given a pointer and size, creates a new object by copying `size` data from `ptr`.
@@ -153,7 +150,6 @@ impl SIRInterpreter {
                 }
             }
             IPlace::Indirect { ptr, off, ty } => {
-                // FIXME copy only depending on type?
                 // Dereference a pointer and copy the data it points to.
                 if let Some(obj) = self.get_var(&ptr.local) {
                     let dstptr = obj.to_usize() + ptr.off as usize + *off as usize;
--- a/ykbh/src/lib.rs
+++ b/ykbh/src/lib.rs
@@ -91,7 +91,7 @@ impl SIRInterpreter {
 
     /// Get the Object referenced by a Local.
     fn get_var(&self, local: &Local) -> &Option<Object> {
-        &self.vars[usize::try_from(local.0).expect("Can't convert local to usize.")]
+        &self.vars[usize::try_from(local.0).unwrap()]
     }
 
     /// Implements the Store statement.
@@ -106,7 +106,7 @@ impl SIRInterpreter {
             IPlace::Val { local, off, ty: _ty } => {
                 // Store into a local.
                 if *off == 0 {
-                    self.vars[usize::try_from(local.0).expect("Can't convert local to usize.")] = Some(src);
+                    self.vars[usize::try_from(local.0).unwrap()] = Some(src);
                 } else {
                     todo!()
                 }
@@ -114,7 +114,7 @@ impl SIRInterpreter {
             IPlace::Indirect { ptr, off, ty: _ty } => {
                 // Write to a pointer.
                 if let Some(obj) = self.get_var(&ptr.local) {
-                    let dstptr = obj.to_usize() + usize::try_from(ptr.off).expect("Can't convert offset to usize.") + usize::try_from(*off).expect("Can't convert offset to usize.");
+                    let dstptr = obj.to_usize() + usize::try_from(ptr.off).unwrap() + usize::try_from(*off).unwrap();
                     self.store_raw(dstptr, src);
                 }
             }
@@ -141,9 +141,9 @@ impl SIRInterpreter {
                 } else {
                     // Copy data from an offset, e.g. $1 = $2+2.
                     if let Some(obj) = self.get_var(local) {
-                        let ptr = obj.data.as_ptr() as usize + usize::try_from(*off).expect("Can't convert offset to usize.");
+                        let ptr = obj.data.as_ptr() as usize + usize::try_from(*off).unwrap();
                         let size = SIR.ty(ty).size();
-                        Object::from_ptr(ptr, usize::try_from(size).expect("Can't convert size to usize."))
+                        Object::from_ptr(ptr, usize::try_from(size).unwrap())
                     } else {
                         unreachable!()
                     }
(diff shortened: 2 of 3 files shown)


ptersilie:
Sorry, forgot to squash.

ykbh/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "ykbh"
+version = "0.1.0"
+authors = ["Lukas Diekmann <lukas.diekmann@gmail.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
vext01, line 7: I'd kill this line.

ykbh/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+
+[dependencies]
+yktrace = { path = "../yktrace" }
+ykpack = { path = "../ykpack" }
+ykcompile = { path = "../ykcompile" }
```
vext01, line 12: Is `ykcompile` used?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+#[derive(Clone, Debug)]
+/// A data object, storing bytes.
+pub struct Object {
+    data: Vec<u8>
+}
+
+impl Object {
+
+    /// Creates a new Object from a `usize`.
+    fn from_usize(ptr: usize) -> Object {
```
vext01, line 19: Are these better as `From` implementations? Maybe too much bikeshedding on my part.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+
+use ykpack::{self, Local, Terminator, Statement, IPlace, Constant, ConstantInt, UnsignedInt};
+use yktrace::sir::SIR;
+use std::convert::TryInto;
+
+const ZST: Object = Object { data: Vec::new() };
+
+#[derive(Clone, Debug)]
+/// A data object, storing bytes.
+pub struct Object {
```
vext01, line 12: So is there one `Object` for each `Local`?

ykbh/Cargo.toml
ptersilie: Not any more. This can go.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+    /// Creates a new Object from a `u8`.
+    fn from_u8(val: u8) -> Object {
+        Object {
+            data: Vec::from([val])
+        }
+    }
+
+    /// Converts the objects data back to a `usize`.
+    fn to_usize(&self) -> usize {
+        usize::from_be_bytes(self.data.as_slice().try_into().expect("Unexpected data length."))
```
vext01, line 35: `be` means big endian? is there a way of asking for native endianess?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+            data.set_len(size);
+        }
+        Object {
+            data
+        }
+    }
+}
+
+pub struct SIRInterpreter {
+    vars: Vec<Option<Object>>,
```
vext01, line 52: A `None` entry here would mean uninitialised?

ykbh/src/lib.rs
ptersilie: Yeah, there used to be more inside the Object. As noted in the PR description, this is not very optimal is gonna change very soon.

ykbh/src/lib.rs
ptersilie: Yeah, we should probably do that.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+            }
+        }
+
+        match &body.blocks[0].term {
+            Terminator::Call {
+                operand: _op,
+                args: _args,
+                destination: _dest
+            } => {
+                todo!()
```
vext01, line 87: Out of interest, what will you do? We want to avoid instantiating a new interpreter for each frame.

ykbh/src/lib.rs
ptersilie: Correct.

ykbh/src/lib.rs
ptersilie: Haven't thought much about it yet. I guess the easiest way (for the current implementation) would be to create a new `vars` vector and restore the old one when returning to this frame.

ykbh/src/lib.rs
vext01: I see. Perhaps later we can use a name which captures the relationship to a local. Maybe `LocalStorage` or something?

ykbh/src/lib.rs
vext01: We should add a comment saying that.

ykbh/src/lib.rs
ptersilie: If it's just a name change, I can update it in this PR.

ykbh/src/lib.rs
ptersilie: No preference. This was the easiest.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+            }
+            Terminator::Return => {
+            }
+            t => todo!("{}", t)
+        }
+    }
+
+    /// Get the Object referenced by a Local.
+    fn get_var(&self, local: &Local) -> &Option<Object> {
```
vext01, line 96: `var` isn't a term we often use (although I know what you mean).

Would a better name be `local_storage()` or `local_bytes()`?

ykbh/src/lib.rs
vext01: Maybe `LocalBytes`?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+use ykpack::{self, Local, Terminator, Statement, IPlace, Constant, ConstantInt, UnsignedInt};
+use yktrace::sir::SIR;
+use std::convert::TryInto;
+
+const ZST: Object = Object { data: Vec::new() };
+
+#[derive(Clone, Debug)]
+/// A data object, storing bytes.
+pub struct Object {
+    data: Vec<u8>
```
vext01, line 13: I think you said you were going to try and use something more efficient than many vectors (for another PR, right?).

We could use one shared (and growing, to account for call frames) vector, and an "Object" is a slice into that?

ykbh/src/lib.rs
ptersilie: Fine with me.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        }
+    }
+
+    pub fn interpret(&mut self, body: &ykpack::Body) {
+        // Ignore yktrace::trace_debug.
+        if body.flags & ykpack::bodyflags::TRACE_DEBUG != 0 {
+            return;
+        }
+
+        for stmt in body.blocks[0].stmts.iter() {
```
vext01, line 68: So we only interpret one hard-coded block for now (hence there is no program counter)?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        }
+    }
+
+    /// Get the Object referenced by a Local.
+    fn get_var(&self, local: &Local) -> &Option<Object> {
+        &self.vars[local.0 as usize]
+    }
+
+    /// Implements the Store statement.
+    fn store(&mut self, dest: &IPlace, src: &IPlace) {
```
vext01, line 101: Since we had `c_store()` in the codegen, should these functions be named like `i_store()`?

ykbh/src/lib.rs
ptersilie: Fine.

ykbh/src/lib.rs
ptersilie: Is this the same ideas as in 2) of the PR description?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        self.store_obj(dest, self.to_obj(src));
+    }
+
+    /// Write an Object to an IPlace. This either creates a new Object if the IPlace is a Val, or
+    /// writes the data to some pointer, if the IPlace is an Indirect.
+    fn store_obj(&mut self, dest: &IPlace, src: Object) {
+        match dest {
+            IPlace::Val { local, off, ty: _ty } => {
+                // Store into a local.
```
vext01, line 110: Interesting. So we don't mutate the existing bytes, but put in new bytes. Is there a reason?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+                        let ptr = obj.data.as_ptr() as usize + *off as usize;
+                        let size = SIR.ty(ty).size();
+                        Object::from_ptr(ptr, size as usize)
+                    } else {
+                        unreachable!()
+                    }
+                }
+            }
+            IPlace::Indirect { ptr, off, ty } => {
+                // FIXME copy only depending on type?
```
vext01, line 156: Can you expand on this FIXME?

ykbh/src/lib.rs
ptersilie: This could be a brand new local, in which case there is nothing to mutate. But you are right, that I could add a check here if the local exists, and if so mutate it instead.

ykbh/src/lib.rs
ptersilie: That's what I did initially, but I didn't like it. I could possibly be convinced to put it back.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        }
+    }
+
+
+    /// Creates a reference to an IPlace.
+    fn mkref(&mut self, dest: &IPlace, src: &IPlace) {
+        let obj = match src {
+            IPlace::Val { local, off, ty: _ty } => {
+                if let Some(obj) = self.get_var(local) {
+                    // The pointer to the vector should remain valid, since no changes to the
```
vext01, line 199: Ah. This is why you can't represent local storage as a slice into one big vector then. That vector will grow as we add new frames and the buffer will move around.

You could store a slice as an `index, len` pair though. That'd be immune.

vext01: But then you'd have to treat pointers given to us from outside specially. Argh!

ykbh/src/lib.rs
ptersilie: Yes, this only interprets one block at the moment. I imagined that later when we do a call, we store the current frame, reinitialise the interpreter, and run it on the other block. Upon returning we just load back the old frame. Though I'm not sure what we'd need the program counter for?

ykbh/src/lib.rs
ptersilie: I don't think I'm a fan of the growing big vector. Though we could have once big vector per frame. That should work.

ptersilie: > But then you'd have to treat pointers given to us from outside specially. Argh!

Yeah, I think my first attempt will be to replace the vecs with arrays and see how far I get with that.

ykbh/src/lib.rs
vext01: It's subtly different. I was proposing one vector for all frames. I don't think that'll work though. See my other comment later about growing/reallocating such a vector.

One vec per-frame would work though. We statically know the size of the frame.

ykbh/src/lib.rs
vext01: We can stick with this for now.

ykbh/src/lib.rs
ptersilie: I'm currently unsure if there are cases where we don't copy the thing this points to, but just return a reference to it instead. I think it depends on the type of the thing, e.g. a tuple is copied, a vec isn't.

ykbh/src/lib.rs
vext01: Well, you kind of have a program counter here. This for loop implicitly is one.

I expected the interpreter to have a stack of frames, then we make a call, we push a new frame (another collection of variables) and update the program counter to the first SIR block of the corresponding function.

So whereas this preliminary version of the interpreter deals with only one-frame's worth of variables, I think the interpreter should maintain a stack. Then we only ever need one instance of the interpreter.

ykbh/src/lib.rs
vext01: Yeah, that sounds good. It's exactly what I expected [here](https://github.com/softdevteam/yk/pull/157/files#r530502603)

ykbh/src/lib.rs
vext01: The duality of it pleases me.

ykbh/src/lib.rs
vext01: > Though we could have once big vector per frame

sounds good.

ykbh/src/lib.rs
ptersilie: Ah ok, I see what you mean now. We need the counter to initialise the interpreter when a guard fails so we can start in the middle of a frame, and also when we come back from a call to remember where we left off.

vext01: Ah yes. That too!

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
+#![feature(test)]
+extern crate test;
```
ltratt, line 2: Do we need these two lines? I vaguely thought they weren't needed in the 2018 edition?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+const ZST: Object = Object { data: Vec::new() };
+
+#[derive(Clone, Debug)]
+/// A data object, storing bytes.
+pub struct Object {
+    data: Vec<u8>
+}
+
+impl Object {
+
```
ltratt, line 17: rustfmt normally removes this particular joy from my life, so I'm glad you let me use my favourite words of old: extraneous blank line.

ykbh/src/lib.rs
ltratt: This is going to be a performance killer, since `Option<Object>` will become `Option<usize>` which requires two machine words of storage.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+impl SIRInterpreter {
+    pub fn new(vars: Vec<Option<Object>>) -> Self {
+        SIRInterpreter {
+            vars,
+        }
+    }
+
+    pub fn interpret(&mut self, body: &ykpack::Body) {
+        // Ignore yktrace::trace_debug.
+        if body.flags & ykpack::bodyflags::TRACE_DEBUG != 0 {
```
ltratt, line 64: ANDing bits to detect something feels quite 1980s: shouldn't `ykpack` be exposing some sort of nice function for this?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    /// Write an Object to an IPlace. This either creates a new Object if the IPlace is a Val, or
+    /// writes the data to some pointer, if the IPlace is an Indirect.
+    fn store_obj(&mut self, dest: &IPlace, src: Object) {
+        match dest {
+            IPlace::Val { local, off, ty: _ty } => {
+                // Store into a local.
+                if *off == 0 {
+                    self.vars[local.0 as usize] = Some(src);
```
ltratt, line 112: As a good rule of thumb we should never ever use `as` unless we document it: `as` is too easy to misuse. `usize::from(...)` is minimally preferable, though `usize::try_from` should be used if there's even the slightest doubt that the cast might not succeed.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{SIRInterpreter, Object};
+    use yktrace::sir::SIR;
+
+    fn interp(fname: &str, tio: usize) {
+        let body = match SIR.bodies.get(fname) {
```
ltratt, line 228: This is `SIR.bodies.get(fname).expect("No SIR")` in disguise, I think.

ykbh/src/lib.rs
ptersilie: Will try.

ykbh/src/lib.rs
ptersilie: Yeah, I should have formatted before opening the PR. I will format later.

ykbh/src/lib.rs
bjorn3: `Object` contains a `Vec`, which has a niche where the storage pointer is a null pointer. This means that `Option<Object>` is exactly as big as `Object`.

ykbh/src/lib.rs
bjorn3:
```suggestion
        usize::from_ne_bytes(self.data.as_slice().try_into().expect("Unexpected data length."))
```

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+    fn from_ptr(ptr: usize, size: usize) -> Object {
+        let mut data = Vec::with_capacity(size);
+        unsafe {
+            std::ptr::copy(ptr as *const u8, data.as_mut_ptr(), size);
+            data.set_len(size);
+        }
+        Object {
+            data
+        }
+    }
```
bjorn3, line 48: This function is unsafe. You can rewrite it to the following to make it safe:
```suggestion
    fn from_slice(slice: &[u8]) -> Object {
        Object {
            data: slice.iter().copied().collect()
        }
    }
```
And then use `std::slice::from_raw_parts` where `from_ptr` is currently called.

ykbh/src/lib.rs
bjorn3: There is the `bitflags` crate.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+            }
+            Terminator::Return => {
+            }
+            t => todo!("{}", t)
+        }
+    }
+
+    /// Get the Object referenced by a Local.
+    fn get_var(&self, local: &Local) -> &Option<Object> {
+        &self.vars[local.0 as usize]
```
bjorn3, line 97:
```suggestion
    fn get_var(&self, local: &Local) -> Option<&Object> {
        self.vars[local.0 as usize].as_ref()
```
This is generally preferred.

ykbh/src/lib.rs
bjorn3: In this case `usize::from` is not possible, as only `From<u8>` and `From<u16>` are implemented for usize.

ykbh/src/lib.rs
bjorn3: A `Vec` value is copied. It is just that the storage pointed to by the internal pointer is not copied.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+                if let Some(obj) = self.get_var(&ptr.local) {
+                    let dstptr = obj.to_usize() + ptr.off as usize + *off as usize;
+                    self.store_raw(dstptr, src);
+                }
+            }
+            _ => unreachable!()
+        };
+    }
+
+    fn store_raw(&self, dest: usize, src: Object) {
```
bjorn3, line 128:
```suggestion
    unsafe fn store_raw(&self, dest: *mut u8, src: Object) {
```
or
```suggestion
    fn store_raw(&self, dest: &[u8], src: Object) {
```

ykbh/src/lib.rs
vext01: I think at the SIR level every store is a copy.

ykbh/src/lib.rs
vext01: I recall looking at the bitflags crate when introducing these flags. I can't remember why I chose not to use it.

ltratt: That's not really what I mean. What I mean is: how we represent flags internally shouldn't be exposed externally. There should almost certainly be a *method* called something like `trace_debug()` so that we can change the internal representation later. Sound like a plan?

vext01: We can do that, but I doubt we will change the representation.

Will raise an issue anyway.

ltratt: No-one thinks they will change a representation... until they do ;)

ykbh/src/lib.rs
ptersilie: Fixed in 2c0a6ed. Looks a bit silly though.

ltratt: As discussed elsewhere, `expect` can be changed to `unwrap`.

ptersilie:
This his gone through a major rewrite so I will raise a new PR.

PR #159, opened 2020-11-26
Add SIR interpreter (Version 2).

ptersilie:
This is the second attempt at our SIR interpreter which now uses a memory map to store the locals. Before you raise your pitchforks keep two things in mind:

1) This currently doesn't support multiple frames and only interprets a single SIR body. We'll implement those together with terminators (calls etc.).
2) The offsets of the locals are currently computed at runtime, which is slow. We will soon make the necessary changes to SIR to do this for us at compile time. We can then retrieve the offsets quickly from SIR. I thought this would be better in a separate PR.

### DIFF
--- a/ykbh/src/lib.rs
+++ b/ykbh/src/lib.rs
@@ -1,0 +1,287 @@
+use std::alloc::{alloc, dealloc, Layout};
+use std::convert::{TryFrom, TryInto};
+use ykpack::{
+    self, BodyFlags, Constant, ConstantInt, IPlace, Local, LocalDecl, Statement, Terminator,
+    TypeId, UnsignedInt,
+};
+use yktrace::sir::SIR;
+
+pub struct SIRInterpreter {
+    /// Pointer to allocated memory containing a frame's locals.
+    locals: *mut u8,
+    /// The offset of each Local into locals.
+    offsets: Vec<usize>,
+    /// The layout of locals. Needed for deallocating locals upon drop.
+    layout: Layout,
+}
+
+impl SIRInterpreter {
+    pub fn new(local_decls: &Vec<LocalDecl>) -> Self {
+        // FIXME Soon this will be pre-computed and handed to us by SIR.
+        let mut offsets = Vec::new();
+        let mut layout = Layout::from_size_align(0, 1).unwrap();
+        for d in local_decls {
+            let align = SIR.ty(&d.ty).align();
+            let size = SIR.ty(&d.ty).size();
+            let l = Layout::from_size_align(size.try_into().unwrap(), align.try_into().unwrap())
+                .unwrap();
+            let (nl, s) = layout.extend(l).unwrap();
+            offsets.push(s);
+            layout = nl;
+        }
+        layout = layout.pad_to_align();
+
+        // Allocate memory for the locals
+        let locals = unsafe { alloc(layout) };
+
+        SIRInterpreter {
+            locals,
+            offsets,
+            layout,
+        }
+    }
+
+    /// Inserts a pointer to the trace inputs into `locals`.
+    pub fn set_trace_inputs(&mut self, tio: *mut u8) {
+        // FIXME Later this also sets other already initialised variables as well as the program
+        // counter of the interpreter.
+        let ptr = self.local_ptr(&Local(1)); // The trace inputs live in $1
+        unsafe {
+            // Write the pointer value of `tio` into locals.
+            std::ptr::write::<*mut u8>(ptr as *mut *mut u8, tio);
+        }
+    }
+
+    pub unsafe fn interpret(&mut self, body: &ykpack::Body) {
+        // Ignore yktrace::trace_debug.
+        if body.flags.contains(BodyFlags::TRACE_DEBUG) {
+            return;
+        }
+
+        for stmt in body.blocks[0].stmts.iter() {
+            match stmt {
+                Statement::MkRef(dest, src) => self.mkref(dest, src),
+                Statement::DynOffs { .. } => todo!(),
+                Statement::Store(dest, src) => self.store(dest, src),
+                Statement::BinaryOp { .. } => todo!(),
+                Statement::Nop => {}
+                Statement::Unimplemented(_) | Statement::Debug(_) => todo!(),
+                Statement::Cast(..) => todo!(),
+                Statement::Call(..) | Statement::StorageDead(_) => unreachable!(),
+            }
+        }
+
+        match &body.blocks[0].term {
+            Terminator::Call {
+                operand: _op,
+                args: _args,
+                destination: _dest,
+            } => todo!(),
+            Terminator::Return => {}
+            t => todo!("{}", t),
+        }
+    }
+
+    /// Get the pointer to a Local.
+    fn local_ptr(&self, local: &Local) -> *mut u8 {
+        let offset = self.offsets[usize::try_from(local.0).unwrap()];
+        unsafe { self.locals.add(offset) }
+    }
+
+    /// Get the pointer for an IPlace, while applying all offsets.
+    fn iplace_to_ptr(&self, place: &IPlace) -> *mut u8 {
+        match place {
+            IPlace::Val {
+                local,
+                off,
+                ty: _ty,
+            } => {
+                // Get a pointer to the Val.
+                let dest_ptr = self.local_ptr(&local);
+                unsafe { dest_ptr.add(usize::try_from(*off).unwrap()) }
+            }
+            IPlace::Indirect { ptr, off, ty: _ty } => {
+                // Get a pointer to the Indirect, which itself points to another pointer.
+                let dest_ptr = self.local_ptr(&ptr.local) as *mut *mut u8;
+                let ptr = unsafe {
+                    // Dereference the pointer, by reading its value.
+                    let mut p = std::ptr::read::<*mut u8>(dest_ptr);
+                    // Add the offsets of the Indirect.
+                    p = p.offset(isize::try_from(ptr.off).unwrap());
+                    p = p.offset(isize::try_from(*off).unwrap());
+                    p
+                };
+                // Now return the value as a pointer.
+                ptr
+            }
+            _ => unreachable!(),
+        }
+    }
+
+    /// Write some bytes to an IPlace. The amount of bytes is determined by the type of the
+    /// destination.
+    fn write(&mut self, dest: &IPlace, src: *const u8) {
+        match dest {
+            IPlace::Val {
+                local: _,
+                off: _,
+                ty,
+            }
+            | IPlace::Indirect { ptr: _, off: _, ty } => {
+                let size = usize::try_from(SIR.ty(ty).size()).unwrap();
+                let ptr = self.iplace_to_ptr(dest);
+                unsafe {
+                    std::ptr::copy(src, ptr, size);
+                }
+            }
+            _ => unreachable!(),
+        }
+    }
+
+    /// Implements the Store statement.
+    fn store(&mut self, dest: &IPlace, src: &IPlace) {
+        match src {
+            IPlace::Val { .. } | IPlace::Indirect { .. } => {
+                let src_ptr = self.iplace_to_ptr(src);
+                self.write(dest, src_ptr);
+            }
+            IPlace::Const { val, ty } => {
+                self.store_const(dest, val, ty);
+            }
+            _ => todo!(),
+        }
+    }
+
+    /// Writes a constant to an IPlace.
+    fn store_const(&mut self, dest: &IPlace, val: &Constant, _ty: &TypeId) {
+        match val {
+            Constant::Int(ci) => match ci {
+                ConstantInt::UnsignedInt(ui) => match ui {
+                    UnsignedInt::U8(v) => self.write(dest, [*v].as_ptr()),
+                    _ => todo!(),
+                },
+                ConstantInt::SignedInt(_si) => todo!(),
+            },
+            Constant::Bool(_b) => todo!(),
+            Constant::Tuple(t) => {
+                if SIR.ty(t).size() == 0 {
+                    // ZST: do nothing.
+                } else {
+                    todo!()
+                }
+            }
+            _ => todo!(),
+        }
+    }
+
+    /// Creates a reference to an IPlace.
+    fn mkref(&mut self, dest: &IPlace, src: &IPlace) {
+        match dest {
+            IPlace::Val { .. } | IPlace::Indirect { .. } => {
+                // Get pointer to src.
+                let src_ptr = self.iplace_to_ptr(src);
+                let dst_ptr = self.iplace_to_ptr(dest);
+                unsafe {
+                    std::ptr::write::<*mut u8>(dst_ptr as *mut *mut u8, src_ptr);
+                }
+            }
+            _ => unreachable!(),
+        }
+    }
+}
+
+impl Drop for SIRInterpreter {
+    fn drop(&mut self) {
+        unsafe { dealloc(self.locals, self.layout) }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::SIRInterpreter;
+    use yktrace::sir::SIR;
+
+    fn interp(fname: &str, tio: *mut u8) {
+        let body = SIR.body(fname).unwrap();
+        let mut si = SIRInterpreter::new(&body.local_decls);
+        // The raw pointer `tio` and the reference it was created from do not alias since we won't
+        // be using the reference until the function `interpret` returns.
+        si.set_trace_inputs(tio);
+        unsafe {
+            si.interpret(&body);
+        }
+    }
+
+    #[test]
+    fn test_simple() {
+        struct IO(u8, u8);
+        #[no_mangle]
+        fn simple(io: &mut IO) {
+            let a = 3;
+            io.1 = a;
+        }
+        let mut tio = IO(0, 0);
+        interp("simple", &mut tio as *mut _ as *mut u8);
+        assert_eq!(tio.1, 3);
+    }
+
+    #[test]
+    fn test_tuple() {
+        struct IO((u8, u8, u8));
+        #[no_mangle]
+        fn func_tuple(io: &mut IO) {
+            let a = io.0;
+            let b = a.2;
+            (io.0).1 = b;
+        }
+
+        let mut tio = IO((1, 2, 3));
+        interp("func_tuple", &mut tio as *mut _ as *mut u8);
+        assert_eq!(tio.0, (1, 3, 3));
+    }
+
+    #[test]
+    fn test_ref() {
+        struct IO(u8, u8);
+        #[no_mangle]
+        fn func_ref(io: &mut IO) {
+            let a = 5u8;
+            let b = &a;
+            io.1 = *b;
+        }
+
+        let mut tio = IO(5, 0);
+        interp("func_ref", &mut tio as *mut _ as *mut u8);
+        assert_eq!(tio.1, 5);
+    }
+
+    #[test]
+    fn test_tupleref() {
+        struct IO((u8, u8));
+        #[no_mangle]
+        fn func_tupleref(io: &mut IO) {
+            let a = io.0;
+            (io.0).1 = 5; // Make sure the line above copies.
+            let b = &a;
+            (io.0).0 = b.1;
+        }
+
+        let mut tio = IO((0, 3));
+        interp("func_tupleref", &mut tio as *mut _ as *mut u8);
+        assert_eq!(tio.0, (3, 5));
+    }
+
+    #[test]
+    fn test_doubleref() {
+        struct IO((u8, u8));
+        #[no_mangle]
+        fn func_doubleref(io: &mut IO) {
+            let a = &io.0;
+            (io.0).0 = a.1;
+        }
+
+        let mut tio = IO((0, 3));
+        interp("func_doubleref", &mut tio as *mut _ as *mut u8);
+        assert_eq!(tio.0, (3, 3));
+    }
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members=[
     "ykrt",
     "yktrace",
     "ykview",
+    "ykbh",
 ]
 
 [profile.dev]
--- a/ykbh/Cargo.toml
+++ b/ykbh/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "ykbh"
+version = "0.1.0"
+authors = ["Lukas Diekmann <lukas.diekmann@gmail.com>"]
+edition = "2018"
+
+[dependencies]
+yktrace = { path = "../yktrace" }
+ykpack = { path = "../ykpack" }


ykbh/src/lib.rs
```diff
@@ -0,0 +1,267 @@
+use std::alloc::{alloc, dealloc, Layout};
+use std::convert::{TryFrom, TryInto};
+use ykpack::{
+    self, Constant, ConstantInt, IPlace, Local, LocalDecl, Statement, Terminator, TypeId,
+    UnsignedInt,
+};
+use yktrace::sir::SIR;
+
+const SUSIZE: usize = std::mem::size_of::<usize>();
```
ltratt, line 9: I think it's better to just do:

```rust
use std::mem::size_of;
```

and then:

```rust
size_of::<usize>()
```

where necessary. It stops people wondering "what is `SUSIZE`?"

ykbh/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+
+pub struct SIRInterpreter {
+    map: *mut u8,
```
ltratt, line 12: What are `map` and `layout`? I was expecting something like:

```rust
struct SIRInterp {
    locals: *mut u8
}
```

but maybe I've misunderstood what's needed!

ykbh/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+    layout: Layout,
+    locals: Vec<usize>,
```
bjorn3, line 14:
```suggestion
    local_offsets: Vec<usize>,
```
It wasn't immediately clear to me that these are offsets into `map`.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+        // FIXME Soon this will be pre-computed and handed to us by SIR.
+        let mut size: usize = 0;
+        let mut locals = Vec::new();
+        for d in local_decls {
+            locals.push(size);
+            size += usize::try_from(SIR.ty(&d.ty).size()).unwrap();
+        }
+
+        // Allocate memory for the locals
+        let layout = Layout::from_size_align(usize::try_from(size).unwrap(), 8).unwrap();
```
bjorn3, line 28: There may be a local that needs to be aligned to a bigger size. I don't know if alignment is really important for the interpreter though. If not, it should be 1 instead.

ykbh/src/lib.rs
bjorn3, line 12: Can this be `Box<[u8]>` or `*mut [u8]`?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+
+        SIRInterpreter {
+            map,
+            layout,
+            locals,
+        }
+    }
+
+    /// Sets the current frame of the interpreter.
+    pub fn set_frame(&mut self, tio: usize) {
```
ltratt, line 39: I'm not sure what `tio` does here, in particular why we we might be calling `to_ne_bytes` on it. Any clues?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+        // FIXME Later this also sets other already initialised variables as well as the program
+        // counter of the interpreter.
+        let ptr = self.get_local(&Local(1)); // The trace inputs live in $1
+        unsafe {
+            std::ptr::copy(usize::to_ne_bytes(tio).as_ptr(), ptr as *mut u8, SUSIZE);
```
ltratt, line 44: Isn't `tio.to_ne_bytes()` accepable over `usize::to_ne_bytes`?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+        }
```
bjorn3, line 45: If `get_local` were to return a `&mut [u8]`, you could use `copy_from_slice` here.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+    pub fn interpret(&mut self, body: &ykpack::Body) {
+        // Ignore yktrace::trace_debug.
+        if body.flags & ykpack::bodyflags::TRACE_DEBUG != 0 {
+            return;
+        }
+
+        for stmt in body.blocks[0].stmts.iter() {
+            match stmt {
+                Statement::MkRef(dest, src) => self.mkref(dest, src),
+                Statement::DynOffs { .. } => todo!("dynoffs"),
```
ltratt, line 57: You can put strings in `todo` if you want, but since it prints the file and line number by default, you might as well save yourself some typing in my opinion!

ykbh/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+            Terminator::Return => {}
+            t => todo!("{}", t),
+        }
+    }
+
+    /// Get the pointer to a Local.
+    fn get_local(&self, local: &Local) -> usize {
+        let offset = self.locals[usize::try_from(local.0).unwrap()];
+        self.map as usize + offset
+    }
```
bjorn3, line 82:
```suggestion
    fn get_local(&self, local: Local) -> *mut u8 {
        let offset = self.locals[usize::try_from(local.0).unwrap()];
        self.map.offset(isize::try_from(offset).unwrap())
    }
```

ykbh/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+                args: _args,
+                destination: _dest,
+            } => todo!(),
+            Terminator::Return => {}
+            t => todo!("{}", t),
+        }
+    }
+
+    /// Get the pointer to a Local.
+    fn get_local(&self, local: &Local) -> usize {
```
ltratt, line 79: I couldn't work out why the return type was `usize` until I realised that you're trying to avoid pointers. I think this method should be called `local_ptr` and return `*mut u8`.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+        let offset = self.locals[usize::try_from(local.0).unwrap()];
+        self.map as usize + offset
+    }
+
+    /// Get the pointer for an IPlace, while applying all offsets.
+    fn iplace_to_ptr(&self, place: &IPlace) -> usize {
```
bjorn3, line 85: Please use `*mut u8` here too.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+                // Add the offsets of the Indirect.
+                val + usize::try_from(ptr.off).unwrap() + usize::try_from(*off).unwrap()
+            }
+            _ => unreachable!(),
+        }
+    }
+
+    /// Write some bytes to an IPlace. The amount of bytes is determined by the type of the
+    /// destination.
+    fn write(&mut self, dest: &IPlace, src: *const u8) {
```
bjorn3, line 113:
```suggestion
    unsafe fn write(&mut self, dest: &IPlace, src: *const u8) {
```
Also prefer `&[u8]`. (It still has to be unsafe, as `dest` could contain a invalid or aliasing pointer.)

ykbh/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+        match dest {
+            IPlace::Val {
+                local: _,
+                off: _,
+                ty,
+            }
+            | IPlace::Indirect { ptr: _, off: _, ty } => {
+                let size = usize::try_from(SIR.ty(ty).size()).unwrap();
```
bjorn3, line 121: I think the SIR should hold `usize` for this kind of things. It isn't portable between architectures anyway.

ltratt:
At a couple of places, this PR returns pointers as `usize`. There are some use cases for that, but it's pretty dangerous in general, as it obscures the fact there's a pointer (I misread the code entirely at first), but without saving us from Rust's onerous pointer rules (see https://www.ralfj.de/blog/2018/07/24/pointers-and-bytes.html for surprising info on this). In general, pointers should be passed around as `*mut T` or `*const T` to make clear what's going on.

bjorn3 (changes requested):
There is way too much unnecessary unsafe here. Miri manages with just a few lines of unsafe code in a very hot function.

ltratt:
[I wonder if you're converting things to `usize` to do pointer arithmetic? If so, Rust's pointers have built-in methods to do this https://doc.rust-lang.org/std/primitive.pointer.html#method.add.]

ykbh/src/lib.rs
ltratt: There could definitely be locals which are aligned to arbitrary sizes, so I think we need to handle that. [In practise, not many things are aligned to `>` machine word, but still...]

ykbh/src/lib.rs
ptersilie: This is probably me trying to be too smart. My thinking was that I don't want to have this call each time we need the size of usize. I forgot the compiler will just optimise away the call anyway.

ykbh/src/lib.rs
ptersilie: We need to hold on to the layout so we can deallocate the memory when the interpreter is dropped. Or will recreating the same layout for the `dealloc` call also work?

ykbh/src/lib.rs
ptersilie: `tio` is a pointer to the trace inputs which I need to store inside the locals map. This will go away (hopefully) once I get rid of the pointer as usize stuff.

ykbh/src/lib.rs
ptersilie: Yeah, looks like some debugging left over.

ykbh/src/lib.rs
ltratt: Yeah, it's not really a call. [Rust has statically evaluated const expressions which, with my Converge CTMP hat on, look incredibly ad-hoc and verbose. But that's a separate rant.]

ykbh/src/lib.rs
ptersilie: Actually, I did mean `1` here. This is a typo.

ykbh/src/lib.rs
ltratt: Once the locals are statically calculated, the `Layout` should be trivially calculable (it should become equivalent to a load of a size & alignment) and thus wouldn't need to be stored here.

ptersilie:
> [I wonder if you're converting things to `usize` to do pointer arithmetic? If so, Rust's pointers have built-in methods to do this https://doc.rust-lang.org/std/primitive.pointer.html#method.add.]

Yes, that was my intention. Missed the pointer methods. Will fix.

ykbh/src/lib.rs
ptersilie: Yeah, I missed that. Will fix.

ptersilie: What are the advantages of this?

ykbh/src/lib.rs
ptersilie: Okay, this is a bit more complicated. We need `to_ne_bytes` as we are interested in the value of pointer itself, not what it points to. So we need to convert the `&IO` into a slice of bytes or a `*const u8`. If there's a better way than going through a `usize` I'm all ears.

ltratt: I don't know what this function is really intending to do, so it's very hard for me to say anything sensible I'm afraid.

ykbh/src/lib.rs
ptersilie: Fixed in b931d23a549829cbd924ec6f1180f7cf7170bb9e.

ykbh/src/lib.rs
ptersilie: Fixed in 9e9ef8f.

ykbh/src/lib.rs
ptersilie: Fixed in 9e9ef8f.

ykbh/src/lib.rs
ptersilie: Fixed in 9e9ef8f.

ykbh/src/lib.rs
ptersilie: Fixed in b931d23.

ykbh/src/lib.rs
ptersilie: Fixed in 01b6575.

ykbh/src/lib.rs
ptersilie: Fixed in b5ac93e.

ykbh/src/lib.rs
ptersilie: At the moment, all it does is pass the pointer to the `trace inputs` to the interpreter. `Trace inputs` is what we also hand to the compiler, which contains stuff like the program counter, etc. If you look at the tests below you can see how `tio` is used.

ykbh/src/lib.rs
ptersilie: Fixed in d48ba09.

ykbh/src/lib.rs
ltratt: I guess what I'm saying is: I don't understand the high-level structure that leads to this decision. My default reaction is that this code is unlikely to be correct under stack borrows, but perhaps it is, and perhaps this is the best way of doing things. I don't know.

ykbh/src/lib.rs
bjorn3: The first one will avoid unsafe allocation and de-allocation code. The second one merely makes bound checking much easier.

ykbh/src/lib.rs
ptersilie: Since we are not writing to `tio` until `interpret` returns, this should be fine under stacked borrows. To be sure I've tested this in MIRI which seems happy: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8a3fd007a16b3db15720382f9b486611

I've added a small comment in 4f9d18c and changed the argument type from `usize` to `*mut usize` to avoid confusion in 1a5048d.

ykbh/src/lib.rs
ptersilie: By aligning the locals I was able to simplify this code: f5cc2d7

ykbh/src/lib.rs
ptersilie: On second thought, I changed this back to `size_of::<usize>()`, which allows us to simplify the code that reads/writes pointers, and should also give us some speedup: f5cc2d7

ykbh/src/lib.rs
ptersilie: I think I prefer `*mut u8` over this, since we'll have to do a lot of pointer arithmetic. Though I could be convinced otherwise if there's a way to easily do that with `[u8]` as well.

ykbh/src/lib.rs
ptersilie: Agreed.

ltratt:
@ptersilie There are a lot of new commits. Do they need reviewing individually?

ptersilie:
After reading up on stacked borrows and playing around with MIRI I have addressed some of the comments. I've also refactored some of the code. So I think this is ready for re-review.

ptersilie:
> @ptersilie There are a lot of new commits. Do they need reviewing individually?

I've commented the commit hash on the individual comments above. So you only need to address the ones addressing your comment. However, the most interesting commits are the last 4, starting at 1a5048d.

ltratt:
I don't see any references to commits?

ptersilie:
> I don't see any references to commits?

https://github.com/softdevteam/yk/pull/159#discussion_r533351143
https://github.com/softdevteam/yk/pull/159#discussion_r533349940
https://github.com/softdevteam/yk/pull/159#discussion_r533349317

ykbh/src/lib.rs
```diff
@@ -35,12 +35,13 @@ impl SIRInterpreter {
     }
 
     /// Sets the current frame of the interpreter.
-    pub fn set_frame(&mut self, tio: usize) {
+    pub fn set_frame(&mut self, tio: *mut usize) {
```
ltratt, line 42: Can you tell me what the model of the "current frame" is here, and what "tio" is? The description "sets the current frame" makes sense, but I can't relate that to the code inside this function. That's probably my bad, but I'm going to need your help to make sense of this!

ykbh/src/lib.rs
```diff
@@ -192,6 +192,8 @@ mod tests {
     fn interp(fname: &str, tio: *mut usize) {
         let body = SIR.bodies.get(fname).expect("No SIR");
         let mut si = SIRInterpreter::new(&body.local_decls);
+        // This does not alias since we won't be reading from or writing to `tio` until the
```
ltratt, line 196: What's "this"?

ykbh/src/lib.rs
ptersilie: At the moment this is pretty much just a stub, since we haven't implemented the relevant part in the SIR yet, which will pass in the `tio` pointer (which points to the same trace inputs used by the yk compiler) and other initialised variables. The goal of this function is to initialise the current frame of the SIR we are going to interpret, which includes setting the program counter to the next instruction to be executed. Maybe I can add the program counter bit in this PR to make more clear what this does?

ykbh/src/lib.rs
ptersilie: See if this is better: 8fe4b9e

ykbh/src/lib.rs
ltratt: OK, so this isn't really about "setting the frame" then? `set_trace_inputs` or something?

ykbh/src/lib.rs
ltratt: Works for me!

ykbh/src/lib.rs
ptersilie: At the moment, no. But soon it will be. So we have two options to fix this:
1) We rename it, and name it back once it actually sets the frame.
2) I add initialisation of the program counter to this function, to make clearer that this will be setting the frame soon.

ltratt: Let's name and document it such that it makes sense in this PR. Future PRs can adjust it as necessary.

ptersilie: Fixed in 801159b.

ptersilie:
I've addressed your comments about alignments, but using Rust's API to calculate the proper alignment. Unfortunately, this meant that we need to store an extra two words for the `Layout` inside `SIRInterpreter` as we need the layout to deallocate the memory when the interpreter is dropped. At least I couldn't work out if recreating a layout using `self.offsets` would result in the same layout (my feeling is no).

ltratt:
> At least I couldn't work out if recreating a layout using self.offsets would result in the same layout (my feeling is no).

Why wouldn't it be deterministic? If:

1. nothing can mutate the offsets
2. you iterate over them in a deterministic order

then there shouldn't be a problem should there?

ptersilie:
> > At least I couldn't work out if recreating a layout using self.offsets would result in the same layout (my feeling is no).
> 
> Why wouldn't it be deterministic? If:
> 
>     1. nothing can mutate the offsets
> 
>     2. you iterate over them in a deterministic order
> 
> 
> then there shouldn't be a problem should there?

Well for a start, I don't know the size of the last element, only it's offset. So when I create a new layout using `from_size_align`, what is its size (if I take the last offset we are missing the size of the last element). And also what do I pass for `align`?

ltratt:
I'm confused. If you don't know that information, how can you lay anything out?

ptersilie:
> I'm confused. If you don't know that information, how can you lay anything out?

I know this information when I create the interpreter, because I'm passing in the locals (`local_decls`). But when we drop the interpreter we won't have access to that information any more.

ltratt:
Ah, hang on, we're not calculating these things statically?

ykbh/src/lib.rs
```diff
@@ -5,36 +5,37 @@ use ykpack::{
...
+        let mut layout = Layout::from_size_align(0, 1).unwrap();
         for d in local_decls {
-            let padding = (ALIGN - (size % ALIGN)) % ALIGN;
-            size += padding;
-            offsets.push(size);
-            size += usize::try_from(SIR.ty(&d.ty).size()).unwrap();
+            let align = SIR.ty(&d.ty).align();
+            let size = SIR.ty(&d.ty).size();
+            let l = Layout::from_size_align(size.try_into().unwrap(), align.try_into().unwrap()).unwrap();
+            let (nl, s) = layout.extend(l).expect("Alignment error");
```
ltratt, line 27: I think `expect` is a) almost entirely pointless and b) clutters code. Let's use `unwrap`.

ykbh/src/lib.rs
```diff
@@ -5,36 +5,37 @@ use ykpack::{
...
+    /// The layout of locals. Needed for deallocating locals upon drop.
+    layout: Layout,
 }
 
 impl SIRInterpreter {
     pub fn new(local_decls: &Vec<LocalDecl>) -> Self {
         // FIXME Soon this will be pre-computed and handed to us by SIR.
-        let mut size: usize = 0;
         let mut offsets = Vec::new();
+        let mut layout = Layout::from_size_align(0, 1).unwrap();
```
ltratt, line 22: This feels a bit evil: would `Layout::new::<u8>()` be a bit clearer as I *think* that's what you're emulating here.

ptersilie:
> Ah, hang on, we're not calculating these things statically?

Not at the moment. My plan was to add that to SIR in a separate PR. However, even if SIR can give us this information statically, we still will have to store the layout. SIR can only give us this information if we tell it which SIR body we want this info for. So we either need to store a function name or the layout within the interpreter somewhere (see https://github.com/softdevteam/yk/pull/159/files#diff-11a1e7f768bc03d887b1e73424972f6621476437c93f785006b0d44582de730fR193 on how we query SIR). Once we do calls, we'll probably need a stack of layouts, so we can deallocate the memory of a SIR body when we are done with it.

ykbh/src/lib.rs
ptersilie: I have no preference. I got this line from one of the examples in the docs: https://doc.rust-lang.org/std/alloc/struct.Layout.html#examples

ltratt:
I'm not sure I understood that, but let's leave that for the next PR.

ykbh/src/lib.rs
ltratt: Let's use `unwrap`.

ykbh/src/lib.rs
ptersilie: Fair.

ykbh/src/lib.rs
ptersilie: You mean use `Layout::new::<u8>()`?

ltratt: Mea culpa, that'll teach me not to reload the GitHub tab! I see what you mean now. We can keep it as-is.

ykbh/src/lib.rs
```diff
@@ -39,8 +39,8 @@ impl SIRInterpreter {
         }
     }
 
-    /// Sets the current frame of the interpreter.
-    pub fn set_frame(&mut self, tio: *mut usize) {
+    /// Inserts a pointer to the trace inputs into `locals`.
+    pub fn set_trace_inputs(&mut self, tio: *mut usize) {
```
ltratt, line 43: I *think* `tio` might better be `*mut u8` to be clearer to readers that it's just a big chunk of bytes. There's not *much* difference between `*mut u8` and `*mut usize`, but the former is more conventional for "random stuff".

ykbh/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+                off,
+                ty: _ty,
+            } => {
+                // Get a pointer to the Val.
+                let dest_ptr = self.local_ptr(&local);
+                unsafe { dest_ptr.add(usize::try_from(*off).unwrap()) }
+            }
+            IPlace::Indirect { ptr, off, ty: _ty } => {
+                // Get a pointer to the Indirect.
+                let dest_ptr = self.local_ptr(&ptr.local) as *mut usize;
```
ltratt, line 103: Why are we casting this `*mut usize`?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+                // Dereference the pointer by reading its value.
+                let val = unsafe { std::ptr::read::<usize>(dest_ptr) };
```
ltratt, line 105: I think this is saying that every local is a `usize`?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+            Constant::Int(ci) => match ci {
+                ConstantInt::UnsignedInt(ui) => match ui {
+                    UnsignedInt::U8(v) => self.write(dest, [*v].as_ptr()),
+                    _ => todo!(),
+                },
+                ConstantInt::SignedInt(_si) => todo!(),
+            },
+            Constant::Bool(_b) => todo!(),
+            Constant::Tuple(t) => {
+                if SIR.ty(t).size() == 0 {
```
ltratt, line 161: Wouldn't this be clearer as `if SIR.ty(t).size() != 0 { todo!(); }`?

ykbh/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+                }
+            }
+            _ => todo!(),
+        }
+    }
+
+    /// Creates a reference to an IPlace.
+    fn mkref(&mut self, dest: &IPlace, src: &IPlace) {
+        // Get pointer to src.
+        let src_ptr = self.iplace_to_ptr(src) as usize;
```
ltratt, line 174: Do we need to use `as` here?

ltratt:
OK, I'm starting to understand more about this PR, so my last round of comments, though not hugely deep, are perhaps a bit more useful than the first round of comments!

ykbh/src/lib.rs
ptersilie: I used `*mut usize` to allude to the fact, that `tio` is a pointer. It also allows me to use `write::<usize>` below, which is little nicer than `copy`. But I can also change this to `*mut u8` if you like, there's not a big difference.

ykbh/src/lib.rs
ptersilie: We are doing this because `local_ptr` returns a pointer which we are going to dereference below. So casting this to `*mut usize` allows me to do:
```rust
let val = read::<usize>(ptr);
```
instead of something like
```rust
let val = 0;
copy(ptr, &val as *mut u8, size_of<usize>());
```

ykbh/src/lib.rs
ptersilie: In this case yes, because this is an odd situation where we have a pointer `dest_ptr` pointing to some value, that happens to also be a pointer. We know this because an `Indirect` local is always a pointer.

ykbh/src/lib.rs
ptersilie: I wasn't sure if we might need to do something someday with ZSTs, but am happy to change it.

ykbh/src/lib.rs
ptersilie: Ah, got call. We can also use the `write::<usize>` here, like we do above. But I'll wait until we've discussed the other comments, because we might also want to change this to use `*mut u8` instead of `*mut usize`.

ykbh/src/lib.rs
ltratt: `*mut T` says "this is is a pointer to zero or more `T`'s". `*mut usize` therefore says "I'm pointing to zero or more `usize`s", which isn't the case here. We really want to say "I'm pointing to zero more bytes".

ptersilie:
I've replied to your comments. See what you think.

ptersilie:
After some discussion offline, we decided that pointers should be cast to `*mut *mut u8`, meaning, this thing is a pointer to a pointer, pointing to some arbitrarily sized data. I've updated the code accordingly.

ykbh/src/lib.rs
```diff
@@ -99,14 +99,18 @@ impl SIRInterpreter {
...
-                // Dereference the pointer by reading its value.
-                let val = unsafe { std::ptr::read::<usize>(dest_ptr) };
-                // Add the offsets of the Indirect.
-                let ptr = val + usize::try_from(ptr.off).unwrap() + usize::try_from(*off).unwrap();
+                // Get a pointer to the Indirect, which itself points to another pointer.
+                let dest_ptr = self.local_ptr(&ptr.local) as *mut *mut u8;
+                let ptr = unsafe {
+                    // Dereference the pointer, by reading its value.
+                    let mut p = std::ptr::read::<*mut u8>(dest_ptr);
+                    // Add the offsets of the Indirect.
```
ltratt, line 107: I think all of the offsets are `usize`s, right? In which case you can do `p.add(ptr.off)`. https://doc.rust-lang.org/std/primitive.pointer.html#method.add

ykbh/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+        let ptr = self.local_ptr(&Local(1)); // The trace inputs live in $1
+        unsafe {
+            // Write the pointer value of `tio` into locals by casting it to an integer first.
+            std::ptr::write::<*mut u8>(ptr as *mut *mut u8, tio);
+        }
+    }
+
+    pub fn interpret(&mut self, body: &ykpack::Body) {
+        // Ignore yktrace::trace_debug.
+        if body.flags & ykpack::bodyflags::TRACE_DEBUG != 0 {
```
ltratt, line 55: This is out of date with the API in master. I suggest you merge master into this PR (don't rebase pretty please :p) and have a look at what's changed. The change is pretty easy.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+            let align = SIR.ty(&d.ty).align();
+            let size = SIR.ty(&d.ty).size();
+            let l = Layout::from_size_align(size.try_into().unwrap(), align.try_into().unwrap()).unwrap();
+            let (nl, s) = layout.extend(l).unwrap();
+            offsets.push(s);
+            layout = nl;
+        }
+
+        // Allocate memory for the locals
+        let locals = unsafe { alloc(layout.pad_to_align()) };
```
bjorn3, line 33:
```suggestion
        layout = layout.pad_to_align();
        let locals = unsafe { alloc(layout) };
```

ykbh/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+        }
+    }
+
+    /// Inserts a pointer to the trace inputs into `locals`.
+    pub fn set_trace_inputs(&mut self, tio: *mut u8) {
+        // FIXME Later this also sets other already initialised variables as well as the program
+        // counter of the interpreter.
+        let ptr = self.local_ptr(&Local(1)); // The trace inputs live in $1
+        unsafe {
+            // Write the pointer value of `tio` into locals by casting it to an integer first.
```
bjorn3, line 48:
```suggestion
```

ykbh/src/lib.rs
ltratt: Agreed.

ykbh/src/lib.rs
ltratt: Agreed.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+            std::ptr::write::<*mut u8>(ptr as *mut *mut u8, tio);
+        }
+    }
+
+    pub fn interpret(&mut self, body: &ykpack::Body) {
```
bjorn3, line 53:
```suggestion
    pub unsafe fn interpret(&mut self, body: &ykpack::Body) {
```

Given the correct `body`, this can dereference arbitrary pointers.

ykbh/src/lib.rs
ptersilie: Unfortunately, the offsets are `i32`s, so `add` doesn't improve things.

ykbh/src/lib.rs
ptersilie: Fixed in d0d1be4.

ykbh/src/lib.rs
ptersilie: Fixed in d0d1be4.

ykbh/src/lib.rs
ptersilie: Fixed in d0d1be4.

ykbh/src/lib.rs
ptersilie: Fixed in 6579b5f.

ptersilie:
Addressed all comments.

ltratt:
@vext01 Can you have a quick look at this? I don't think we want to bother Lukas with small comments at this stage, but just to check that you're happy with the overall direction.

ykbh/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    /// Inserts a pointer to the trace inputs into `locals`.
+    pub fn set_trace_inputs(&mut self, tio: *mut u8) {
+        // FIXME Later this also sets other already initialised variables as well as the program
+        // counter of the interpreter.
+        let ptr = self.local_ptr(&Local(1)); // The trace inputs live in $1
+        unsafe {
+            // Write the pointer value of `tio` into locals.
+            std::ptr::write::<*mut u8>(ptr as *mut *mut u8, tio);
```
vext01, line 50: Why the double pointer here?

vext01:
A couple of small comments.

ykbh/src/lib.rs
ptersilie: We want to store the pointer value of `tio` and not what it points to. This avoids casting `tio` to a `usize` and makes it clearer, that `tio` is a pointer. And since we are storing a pointer `*mut u8`, the destination `ptr` needs to be a pointer to a pointer `*mut *mut u8`.

ptersilie:
> A couple of small comments.

This suggest there are multiple comments, though I only saw one. Was that the only one?

vext01:
My bad. Yes, please go ahead.

Next PR should do support for frames?

ptersilie:
> Next PR should do support for frames?

That's the plan.

ltratt:
Please squash.

ptersilie:
Squashed!

ltratt:
Let's update the commit message so that it gives a bit of context a) about what the SIR interpreter is b) how it works. This is an important PR and people may well look back at it, so we need to be clear with them about what's going on. Feel free to force push an update.

ptersilie:
I've pushed a description. It probably needs changing, so see what you think.

ltratt:
A quick couple of comments:

ptersilie:
>     1. We should explicitly say this for what RPython calls blackholing.

Hah, I had that in there at some point, but it got lost during a rewrite.

>     2. We call them "control points" not "merge points" (IIRC: @vext01 can confirm).

I think we do!

ltratt:
Feel free to force push an update.

ptersilie:
Pushed.

ltratt:
Happy news!

bors r+

ltratt:
@ptersilie You need to rebase against master.

ptersilie:
Rebased!

vext01:
> realloc(): invalid old size

ooer...

ltratt:
@ptersilie Have you seen the CI failure?

ptersilie:
Yes, I'm looking into it.

ptersilie:
bors try

ltratt:
bors try

ptersilie:
Ok to rebase?

vext01:
yup

ptersilie:
Rebased!

PR #160, opened 2020-11-26
Sort Cargo.toml entries.

ltratt:
In my experience it's easier to do things like updating dependencies if the entries are sorted alphabetically.

### DIFF
--- a/ykcompile/Cargo.toml
+++ b/ykcompile/Cargo.toml
@@ -8,12 +8,12 @@ license = "Apache-2.0 OR MIT"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-yktrace = { path = "../yktrace" }
-ykpack = { path = "../ykpack" }
 dynasmrt = "1.0.0"
 hex = "0.4.2"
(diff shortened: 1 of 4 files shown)


ykcompile/Cargo.toml
```diff
@@ -8,12 +8,12 @@ license = "Apache-2.0 OR MIT"
...
 [dependencies]
-yktrace = { path = "../yktrace" }
-ykpack = { path = "../ykpack" }
 dynasmrt = "1.0.0"
 hex = "0.4.2"
-libc = "0.2.80"
 lazy_static = "1.4.0"
+libc = "0.2.80"
+ykpack = { path = "../ykpack" }
+yktrace = { path = "../yktrace" }
```
bjorn3, line 16: I personally like to put all related dependencies together with a newline separating them from the rest. In this case that would be all `yk*` dependencies.

ltratt: The beauty of this approach is that the `yk` dependencies live to next each other alphabetically. Or, put another way, it stops me bikeshedding on newlines :p

vext01:
I have no preference on the ordering, so this is fine for me.

bors r+

PR #161, opened 2020-11-27
Don't expose bit flag operations quite so much to the user.

ltratt:
Using 'bitflags' gives us the same internal representation but slightly more static type safety.

### DIFF
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -1,5 +1,6 @@
 //! Types for the Yorick intermediate language.
 
+use bitflags::bitflags;
 use serde::{Deserialize, Serialize};
 use std::{
     convert::TryFrom,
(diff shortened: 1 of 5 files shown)


ykpack/src/types.rs
```diff
@@ -343,14 +344,17 @@ impl Display for Local {
...
-    /// This function is annotated #[interp_step].
-    pub const INTERP_STEP: u8 = 1 << 1;
-    /// This function is yktrace::trace_debug.
-    pub const TRACE_DEBUG: u8 = 1 << 2;
+bitflags! {
+    /// Bits in the `flags` bitfield in `Body`.
+    #[derive(Serialize, Deserialize)]
+    pub struct BodyFlags: u8 {
+        /// This function is annotated #[do_not_trace].
+        const DO_NOT_TRACE = 0b00000001;
```
vext01, line 352: can we use the `1<<X` idiom? It's easier to read and less error prone.

vext01:
Just an easy comment. Looks good.

ykpack/src/types.rs
ltratt: I used to use `1<<X` but I've made too many mistakes with it to be fond of it any more. To be honest, I didn't think deeply about this:  I think all of the examples you'll find using bitflags use the `0b` notation, which suggests that's what people have gravitated towards.

ltratt:
@vext01 Does this PR need separate `bors`ing?

PR #162, opened 2020-11-28
Force codegen-units=1.

ltratt:
Higher "codegen-units" (in theory at least) improve compile time, but do so at the expense of duplicating compilation. In our case, this means we end up with duplicate SIR elements.

As a rough proxy, I measured how many duplicate elements we end up with in this repository while doing "cargo build --tests":

```
  codegen-units    % of duplicated MIR elements
  0 [i.e. "inf"]   2.93%
  1                0.28%
  2                1.15%
  3                1.62%
```

I couldn't measure any time difference for a fresh "cargo build --tests" between codegen-units={0,1,3}. I'm not sure why.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,4 +8,5 @@ members=[
 ]
 
 [profile.dev]
+codegen-units = 1
 panic = "abort"


vext01:
Can you measure the time for an incremental rebuild? That's what I'd expect
#cgus to impact more.

I think with one Cgu you force rebuild of the entire crate, which wouldn't
be good.

bjorn3:
`-Ccodegen-units=1` also affects release mode more than debug mode, as only optimizations can be done in parallel currently. The rest is waiting for parallel rustc.

ltratt:
At the moment, even in the worst case, `cargo build --tests` from scratch on b13 takes 1 minute, so I don't think we have much to lose. With one lousy test, incremental builds seemed the same in either case (and, TBH, I don't *think* codegen-units affects them much). As Bjorn says, I think for "normal" Rust this mostly affects release mode. Since, at the moment, we can't use `--release` here, we're not going to lose much...

bjorn3:
> (and, TBH, I don't think codegen-units affects them much)

Incremental mode forcefully uses a specific codegen unit count I think.

vext01:
> Incremental mode forcefully uses a specific codegen unit count I think.

Does that mean that Laurie's change is ineffectual?

Assuming it isn't, would forcing one codegen unit here (in the yk repo) help an interpreter author. Wouldn't it be the consumer of `yk` that should set this option?

It may be better to alter the compiler to default to use one CGU (but allow overrides)?

bjorn3:
> Does that mean that Laurie's change is ineffectual?

It does have effect on yorick itself when using it as dependency. Dependencies outside the workspace are never compiled in incremental mode.

> Assuming it isn't, would forcing one codegen unit here (in the yk repo) help an interpreter author. Wouldn't it be the consumer of yk that should set this option?

Yes and yes

> It may be better to alter the compiler to default to use one CGU (but allow overrides)?

That would hurt incremental compilation.

vext01:
> Dependencies outside the workspace are never compiled in incremental mode.

That's interesting, and makes complete sense when I think about it.

So we probably want the proposed change both here and in consumers.

bjorn3:
`[profile]` sections are ignored for dependencies. They only work for the workspace root.

vext01:
Right, so I think that by putting the `[profile]` snippet in both `yk` and consumers, it means the test binaries of `yk` are built with one CGU, and then consumers also get one CGU. In the latter case the `yk` test binaries are not required (or built) at all, so that's fine.

ltratt:
I think the right thing to do is probably to close this PR. It's generated a useful discussion and I think the max 3% saving isn't worth worrying about, at least at the moment. We can come back to this in the future if we need to optimise more things!

PR #163, opened 2020-12-01
Implement lazy SIR loading.

vext01:
Companion PR ~~soon~~ https://github.com/softdevteam/ykrustc/pull/149.

Currently we load all of the SIR into memory the first time we request a
type or a body. By doing so we incur a large (one-time) cost, even if we
don't use 99% of the loaded SIR.

This change implements lazy loading. Each SIR section now starts with a
header which serves as a table of contents, mapping type and body
identifiers to their byte offsets within the section. This gives us
random access to the contents of the SIR, which is both faster and more
memory efficient.

This first version is naive in that it implements no caching (of types,
bodies, or any of the associated decoding mechanisms). It can surely be
optimised further (but later).

Rough benchmarking of a full test suite run before and after this
change:

```
Before:
  real        30.643      0.325       30.246      30.638      31.245
  user        127.624     0.931       126.417     127.238     129.612
  sys         5.106       0.680       3.925       5.270       6.062
After:
  real        22.399      0.614       21.910      22.159      24.056
  user        131.255     0.653       129.918     131.371     132.273
  sys         4.981       0.742       4.145       4.953       6.603
```

The next-most-costly part of the system is now dwarf stuff.

### DIFF
--- a/ykpack/src/lib.rs
+++ b/ykpack/src/lib.rs
@@ -39,19 +39,7 @@ pub const SIR_SECTION_PREFIX: &str = ".yksir_";
 mod tests {
     use super::{BasicBlock, Body, BodyFlags, Decoder, Encoder, Pack, Statement, Terminator};
     use fallible_iterator::{self, FallibleIterator};
-    use std::io::{Cursor, Seek, SeekFrom};
-
-    // Get a cursor to serialise to and deserialise from. For real, we'd be reading from a file,
-    // but for tests we use a vector of bytes.
-    fn get_curs() -> Cursor<Vec<u8>> {
-        let buf: Vec<u8> = Vec::new();
-        Cursor::new(buf)
-    }
-
-    // Rewind a cursor to the beginning.
-    fn rewind_curs(curs: &mut Cursor<Vec<u8>>) {
-        curs.seek(SeekFrom::Start(0)).unwrap();
-    }
+    use std::io::Cursor;
 
     // Makes some sample stuff to round trip test.
     fn get_sample_packs() -> Vec<Pack> {
@@ -90,40 +78,30 @@ mod tests {
         vec![sir1, sir2]
     }
 
-    // Check serialising and deserialising works for zero packs.
-    #[test]
-    fn test_empty() {
-        let mut curs = get_curs();
-
-        let enc = Encoder::from(&mut curs);
-        enc.done().unwrap();
-
-        rewind_curs(&mut curs);
-        let mut dec = Decoder::from(&mut curs);
-        assert!(dec.next().unwrap().is_none());
-    }
-
     // Check a typical serialising and deserialising session.
     #[test]
     fn test_basic() {
         let inputs = get_sample_packs();
-        let mut curs = get_curs();
-
-        let mut enc = Encoder::from(&mut curs);
+        let num_packs = inputs.len();
+        let mut buf = Vec::new();
+        let mut enc = Encoder::from(&mut buf);
         for md in &inputs {
             enc.serialise(md.clone()).unwrap();
         }
-        enc.done().unwrap();
 
-        rewind_curs(&mut curs);
+        let mut curs = Cursor::new(&mut buf);
         let dec = Decoder::from(&mut curs);
 
         // Obtain two fallible iterators, so we can zip them.
         let expect_iter = fallible_iterator::convert(inputs.into_iter().map(|e| Ok(e)));
 
         let mut itr = dec.zip(expect_iter);
-        while let Some((got, expect)) = itr.next().unwrap() {
+        for _ in 0..num_packs {
+            let (got, expect) = itr.next().unwrap().unwrap();
             assert_eq!(expect, got);
         }
+
+        // We've consumed everything, so attempting to decode another pack should fail.
+        assert!(itr.next().is_err());
     }
 }
--- a/ykpack/src/types.rs
+++ b/ykpack/src/types.rs
@@ -1,14 +1,17 @@
 //! Types for the Yorick intermediate language.
 
 use bitflags::bitflags;
+use fxhash::FxHashMap;
 use serde::{Deserialize, Serialize};
 use std::{
     convert::TryFrom,
+    default::Default,
     fmt::{self, Display},
     mem,
 };
 
 // FIXME these should probably all be tuple structs, as type aliases offer little type safety.
+pub type SirOffset = usize;
 pub type DefIndex = u32;
 pub type BasicBlockIndex = u32;
 pub type StatementIndex = usize;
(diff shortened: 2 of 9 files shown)


ltratt:
Those performance numbers suggest that we're now running parallel code more efficiently (as in: not block as often) because the `user` portion is more-or-less unchanged. Does that sound plausible?

ykpack/src/lib.rs
```diff
@@ -90,40 +78,30 @@ mod tests {
...
 
         let mut itr = dec.zip(expect_iter);
-        while let Some((got, expect)) = itr.next().unwrap() {
+        for _ in 0..num_packs {
+            let (got, expect) = itr.next().unwrap().unwrap();
             assert_eq!(expect, got);
         }
+
+        // We've consumed everything, so attempting to decode another pack should fail.
+        assert!(itr.next().is_err());
```
ltratt, line 105: This feels like a `debug_assert` to me. [Similarly the `assert_eq` just above on line 101/126.]

ykpack/src/types.rs
```diff
@@ -1050,32 +1053,49 @@ impl Display for BinOp {
     }
 }
 
+/// The header of a SIR section.
```
ltratt, line 1056: I think this sentence is redundant.

yktrace/src/sir.rs
```diff
@@ -1,92 +1,148 @@
...
+    pub static ref SIR: Sir<'static> = Sir::new(&*EXE_MMAP).unwrap();
 }
 
-lazy_static! {
-    pub static ref SIR: Sir = Sir::read_file(&env::current_exe().unwrap()).unwrap();
+/// An interface to the serialised IR of an executable.
+///
+/// One of these structures is generated in the above `lazy_static` and is then shared immutably
+/// accross all threads. Only the headers of each SIR section are eagerly loaded. For performance
+/// reasons, the actual IR is loaded lazily on-demand.
```
ltratt, line 27: Choose "lazily" or "on-demand" ;)

yktrace/src/sir.rs
```diff
@@ -1,92 +1,148 @@
...
         // SAFETY: Not really, we hope that nobody changes the file underneath our feet.
-        let data = unsafe { Mmap::map(&File::open(file).unwrap()).unwrap() };
-        let object = object::File::parse(&*data).unwrap();
-
-        // We iterate over ELF sections, looking for ones which contain SIR and loading them into
-        // memory.
-        let mut bodies = HashMap::new();
-        let mut types = HashMap::new();
-        for sec in object.sections() {
+        fn read_hdr(curs: &mut Cursor<&[u8]>) -> SirHeader {
```
ltratt, line 40: Why is this a nested function? It only seems to be called once?

yktrace/src/sir.rs
```diff
@@ -1,92 +1,148 @@
...
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        for body in self.bodies.values() {
-            writeln!(f, "{}", body)?;
+    /// Decode a type in a named section, at an absolute offset from the beginning of that section.
+    fn decode_ty(&self, sec_name: &str, off: SirOffset) -> ykpack::Ty {
+        let mut curs = self.cursor_for_section(&sec_name);
+        curs.seek(SeekFrom::Start(u64::try_from(off).unwrap()))
+            .unwrap();
+        let mut dec = Decoder::from(&mut curs);
+        if let Ok(Some(Pack::Type(t))) = dec.next() {
```
ltratt, line 76: What does `next` return here? I wonder if this is just a long version of `dec.next().unwrap().unwrap().0`?

yktrace/src/sir.rs
```diff
@@ -1,92 +1,148 @@
...
+        for (sec_name, hdr, hdr_size) in SIR.hdrs.values() {
+            if let Some(off) = hdr.bodies.get(body_sym) {
+                return Some(self.decode_body(sec_name, hdr_size + off));
             }
         }
+        None
+    }
+}
 
+impl<'m> Display for Sir<'m> {
```
ltratt, line 116: General philosophical point: should we be implementing `Display` or `Debug` for these things? I tend to think the latter but I don't feel strongly about it.

ykpack/src/lib.rs
bjorn3: This is a test, so it should be `assert!`.

ltratt: Every `assert` is a test ;)

ltratt: Mea culpa! I see what you mean.

ptersilie:
After reading up on stacked borrows and playing around with MIRI I have addressed some of the comments. I've also refactored some of the code. So I think this is ready for re-review.

vext01:
> Those performance numbers suggest that we're now running parallel code more efficiently.

I think so, because before all of the test threads had to block for the long-running `lazy_static` to load the SIR. Now the runtime of that `lazy_static` is very short, as we only load a handful of small headers in.

ltratt:
@ptersilie Wrong PR...

yktrace/src/sir.rs
vext01: I felt it looked nicer. I can unroll it.

ptersilie:
Whoops!

yktrace/src/sir.rs
vext01: Good point.

yktrace/src/sir.rs
ltratt: Putting it in a function made me hunt around to see why it was being called multiple times, so I think it's a distraction.

yktrace/src/sir.rs
vext01: The convention we've used so is `Display` for pretty-printing, and  `Debug` for everything else.

vext01:
@ltratt I've addressed your comments in 2e2f34b.

yktrace/src/sir.rs
vext01: I couldn't simplify it quite to that degree, as we need to destructure the header variant of the pack.

If this really bothers us, we can make the header independent of `Pack` and make this new type serde compatible.

I'm not sure it's that' big of an issue though?

ltratt: It's really whatever you think is easiest!

yktrace/src/sir.rs
ltratt: They're both for pretty-printing of sorts, but `Debug` is intended for internal debugging and `Display` for user-facing stuff. Not something to fix in this PR, maybe, but something to think about.

ltratt:
If you're happy with the comment(s) above, then please squash.

yktrace/src/sir.rs
vext01: > They're both for pretty-printing of sorts

I suppose.

Our `Display` implementations try to concise at the cost of not printing everything like `Debug` would. Also our `Display` stuff for SIR is somewhat "user" facing, as `ykview` prints them.

vext01:
All fine by me. Squashed.

vext01:
Compiler-side first, right?

bors r-

ltratt:
Oops!

PR #164, opened 2020-12-02
Remove the `Arc` from inside `Location`.

ltratt:
`Arc`s are bulky -- they malloc memory and, on top of whatever overhead
the malloc library imposes, store two extra machine words. That indirection is
also going to induce some slowdown (though that's harder to quantify).

We want `Location`s to be small and fast. This commit makes a `Location` just a
`AtomicUsize`, which is as small and fast as I think we can possibly make it.

However, I assume one of the reasons we moved to using `Arc` is that it gives us
a way of saying "is the current thread tracing the current Location?" using Arc
pointer equality. Moving to just an AtomicUsize removes that possibility from
us.

The problem then becomes "how do I know if this Location is the one I was
tracing?" We don't want to rely on the address of the Location, since that
imposes a burden on the user to place it somewhere where it can't move (which
they could easily forget, leading to all sorts of pain).

I have therefore borrowed a cunning trick from RPython, whereby whilst -- and
only whilst -- we're `PHASE_TRACING` a Location, we malloc a byte of memory and
use that as a unique identifier for the Location. Even if the `Location` moves
around, the malloc'd address won't move, so we can always identify the
`Location`. We only pay this cost for `Location`s that we're tracing, so it's
fairly cheap.

In order that we don't leak memory, or double free, we can no loner allow
`Location`s to be `Clone`able. I think this is probably a good thing, as it
makes clearer to the user the idea that "`Location`s are sort-of unique." It
does mean tweaking some tests so that `Location`s are stored in a way that
multiple threads can access. In this case, amusingly, the easiest way to solve
the problem is to put `Location`s in tests inside `Arc`s. *However*, there's no
reason why a real, high performance interpreter would do the same: I think
they'd just embed a `Location` (which is just a `usize` after all) directly in
their bytecode.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -1,11 +1,8 @@
-use std::cmp::{Eq, PartialEq};
 #[cfg(test)]
 use std::time::Duration;
 use std::{
     io, mem,
-    ops::Deref,
     panic::{catch_unwind, resume_unwind, UnwindSafe},
-    ptr,
     rc::Rc,
     sync::{
         atomic::{AtomicBool, AtomicUsize, Ordering},
@@ -25,55 +22,29 @@ const PHASE_NUM_BITS: usize = 2;
 // whose value is a pointer to memory. By also making that tag 0b00, we allow that index to be
 // accessed without any further operations after the initial tag check.
 const PHASE_TAG: usize = 0b11; // All of the other PHASE_ tags must fit in this.
-const PHASE_COMPILED: usize = 0b00;
-const PHASE_TRACING: usize = 0b01;
-const PHASE_COUNTING: usize = 0b10; // The value specifies the current hot count.
+const PHASE_COMPILED: usize = 0b00; // Value is a pointer to a chunk of memory containing a
+                                    // compiled trace.
+const PHASE_TRACING: usize = 0b01; // Value is a pointer to a malloc'd block that allows us to
+                                   // precisely identify when we have reached the same Location.
+const PHASE_COUNTING: usize = 0b10; // The value specifies how many times we've seen this Location.
 
-/// A `Location` is a handle on a unique identifier for control point position in the end-user's
-/// program (and is used by the `MT` to store data about that location). In other words, every
-/// position that can be a control point also needs to have one `Location` value associated with
-/// it. Note however that instances of `Location` may be freely cloned and shared across threads.
+/// A `Location` stores state that the meta-tracer needs to identify hot loops and run associated
+/// machine code.
+///
+/// Each position in the end user's program that may be a control point (i.e. the possible start of
+/// a trace) must have an associated `Location`. The `Location` does not need to be at a stable
+/// address in memory and can be freely moved.
 ///
 /// Program positions that can't be control points don't need an associated `Location`. For
 /// interpreters that can't (or don't want) to be as selective, a simple (if moderately wasteful)
 /// mechanism is for every bytecode or AST node to have its own `Location` (even for bytecodes or
 /// nodes that can't be control points).
-#[derive(Debug, Clone)]
-pub struct Location {
-    inner: Arc<LocationInner>,
-}
-
-impl Location {
-    pub fn new() -> Self {
-        Self {
-            inner: Arc::new(LocationInner::new()),
-        }
-    }
-}
-
-impl PartialEq for Location {
-    fn eq(&self, other: &Location) -> bool {
-        ptr::eq(Arc::as_ptr(&self.inner), Arc::as_ptr(&other.inner))
-    }
-}
-
-impl Eq for Location {}
-
-impl Deref for Location {
-    type Target = LocationInner;
-
-    fn deref(&self) -> &Self::Target {
-        &self.inner
-    }
-}
-
 #[derive(Debug)]
-pub struct LocationInner {
+pub struct Location {
     pack: AtomicUsize,
 }
 
-impl LocationInner {
-    /// Create a fresh Location suitable for passing to `MT::control_point`.
+impl Location {
     pub fn new() -> Self {
         Self {
             pack: AtomicUsize::new(PHASE_COUNTING),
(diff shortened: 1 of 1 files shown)


vext01:
Before I do a proper review, let me see if I understand correctly.

We (ab)use the heap allocator as a unique ID generator: addresses to separate allocations are guaranteed to be unique.

We could have used an atomic counter (with relaxed ordering) to issue IDs, but the problem then becomes: what should we do in the (unlikely, but possible) event that the counter reaches its maximum value? You could re-use older, no-longer in use, IDs, but you'd have to maintain some kind of free-list. Gee.. that sounds a lot like a memory allocator :)

bjorn3:
Am I correct that `PHASE_TRACING` can only happen for a single location for each thread? If so you could use the thread id as unique ID. You should probably cache it in a `thread_local!` though.

vext01:
> Am I correct that PHASE_TRACING can only happen for a single location for each thread?

I think we'd only want any given location to be traced at most once at a time. So there needs to be some kind of synchronisation to check that two threads don't trace the same location concurrently.

ltratt:
> We (ab)use the heap allocator as a unique ID generator: addresses to separate allocations are guaranteed to be unique. Am I right?

Correct.

> Am I correct that PHASE_TRACING can only happen for a single location for each thread? If so you could use the thread id as unique ID. You should probably cache it in a thread_local! though.

That's an interesting idea! My initial reaction is that it might work *but* we only have 62 bits for it and the documentation for thread IDs (https://doc.rust-lang.org/std/thread/struct.ThreadId.html#method.as_u64) doesn't seem to allow us to guarantee that the ID is an aligned address. I'm sure it probably *is* an aligned address in practise, but all we need is for one platform to hand out IDs which are an incrementing counter and then we'll have problems :/ Any thoughts?

vext01:
I mis-read @bjorn3's idea totally. My apologies @bjorn3!

ykrt/src/mt.rs
```diff
@@ -25,55 +22,29 @@ const PHASE_NUM_BITS: usize = 2;
...
 const PHASE_TAG: usize = 0b11; // All of the other PHASE_ tags must fit in this.
-const PHASE_COMPILED: usize = 0b00;
-const PHASE_TRACING: usize = 0b01;
-const PHASE_COUNTING: usize = 0b10; // The value specifies the current hot count.
-
-/// A `Location` is a handle on a unique identifier for control point position in the end-user's
-/// program (and is used by the `MT` to store data about that location). In other words, every
-/// position that can be a control point also needs to have one `Location` value associated with
-/// it. Note however that instances of `Location` may be freely cloned and shared across threads.
+const PHASE_COMPILED: usize = 0b00; // Value is a pointer to a malloc'd block containing a
```
vext01, line 25: The traces are `mmap`'d, but it's probably be best to not specify how it was allocated, since that decision isn't ours (it's dynasm's).

ykrt/src/mt.rs
```diff
@@ -25,55 +22,29 @@ const PHASE_NUM_BITS: usize = 2;
...
+const PHASE_TRACING: usize = 0b01; // Value is a pointer to a malloc'd block that allows us to
+                                   // precisely identify when we have reached the same Location.
+const PHASE_COUNTING: usize = 0b10; // The value specifies how many times we've seen this Location.
+
+/// A `Location` stores state that the meta-tracer needs to identify hot loops and run associated
+/// machine code.
+///
+/// Each position in the end user's program that may be a control point (i.e. the possible start of
+/// a trace) must have an associated `Location`. The `Location` does not need to be at a stable
+/// address in memory and can be freely moved.
```
vext01, line 36: and freely cloned? I think we said that we disallow it, but do we need docs?

ykrt/src/mt.rs
```diff
@@ -25,55 +22,29 @@ const PHASE_NUM_BITS: usize = 2;
...
-    type Target = LocationInner;
-
-    fn deref(&self) -> &Self::Target {
-        &self.inner
-    }
-}
-
 #[derive(Debug)]
-pub struct LocationInner {
+pub struct Location {
```
vext01, line 43: I wonder if "pack" is now a bad name. One for another PR maybe.

ykrt/src/mt.rs
```diff
@@ -255,69 +226,69 @@ impl MTThread {
...
+            let lp = loc.pack.load(Ordering::Acquire);
             match lp & PHASE_TAG {
                 PHASE_COUNTING => {
                     let count = (lp & !PHASE_TAG) >> 2;
-                    let new_pack;
                     if count >= self.inner.hot_threshold {
                         if self.inner.tracer.is_some() {
                             // This thread is already tracing. Note that we don't increment the hot
                             // count further.
-                            break;
```
vext01, line 270: This early return is functionally equivalent, right? Why the change?

ykrt/src/mt.rs
```diff
@@ -255,69 +226,69 @@ impl MTThread {
...
+                            return None;
                         }
-                        if pack.compare_and_swap(lp, PHASE_TRACING, Ordering::Release) == lp {
+                        let loc_id = Box::into_raw(Box::new(0u8));
```
vext01, line 241: Should generation and fetching of IDs be encapsulated inside `Location` to allow the implementation to change?

bjorn3:
> That's an interesting idea! My initial reaction is that it might work but we only have 62 bits for it and the documentation for thread IDs (https://doc.rust-lang.org/std/thread/struct.ThreadId.html#method.as_u64) doesn't seem to allow us to guarantee that the ID is an aligned address. I'm sure it probably is an aligned address in practise, but all we need is for one platform to hand out IDs which are an incrementing counter and then we'll have problems :/ Any thoughts?

It isn't an address at all, it is a number. You will have to shift it left by 2 bits. Also if you end up caching it in a tls var, you could have a global counter that gets incremented each time a new thread id is requested. This will have lower overhead as it can skip a system call.

ykrt/src/mt.rs
vext01: We should also document the purpose of the box allocation, as it's non-obvious.

vext01: > We should also document the purpose of the box allocation, as it's non-obvious.

I see, it's documented in `MTThreadInner`. OK!

ykrt/src/mt.rs
ltratt: It doesn't implement `Clone`, so we don't need to be explicit about that.

ykrt/src/mt.rs
ltratt: We think as one :) It is not a good name, and I changed it at one point before deciding I'd do that in a subsequent PR.

ykrt/src/mt.rs
ltratt: The loop is much more than a screen big, so I kept forgetting what `break` would jump to in this case.

ltratt:
> It isn't an address at all, it is a number. You will have to shift it left by 2 bits.

But that's only safe if the 2 left-most bits aren't used, right?

vext01:
> It isn't an address at all, it is a number.

What happens if the thread ID is larger than the space we have to store it in? The beauty of the RPython method is that the IDs are aligned pointers, so they not only always fit, but can automatically be re-used.

bjorn3:
On Linux the task id is 32bit I think.

vext01:
A few minor comments from me.

vext01:
I leave the thread ID decision up to @ltratt :)

ltratt:
@vext01 It would be nice to avoid the `malloc` but I fear that we can't rely on it.

@bjorn3:

> On Linux the task id is 32bit I think.

OK, maybe in the future we can have a per-platform hack where, when we can guarantee the thread ID is small enough, we use it instead of malloc. It would be a nice optimisation and the good news is that the API here means we can make such a change totally transparently!

ykrt/src/mt.rs
vext01: Cool.

"bits", "value" or "data" would all be better IMHO.

ykrt/src/mt.rs
ltratt: Agreed, I think the current vagueness is about right.

vext01:
> Should generation and fetching of IDs be encapsulated...

So we should probably do this now to allow different implementations (neatly).

ltratt:
I did briefly try that, but it makes the performance guarantees that we want much less obvious in the code. TBH, I think if/when we improve things in the future, that's the time to shuffle the code around. If we try it now, we probably won't quite get it right anyway, and either way it's not a user facing change.

vext01:
ok

ltratt:
I think this is ready for merging then?

vext01:
I was expecting the comment we were talking about [here](https://github.com/softdevteam/yk/pull/164#discussion_r535240206) to be reverted?

ltratt:
> I was expecting the comment we were talking about here to be reverted?

I'm not sure I understand?

vext01:
I thought (I guess wrongly) that we had agreed that since the trace is not malloc'd that the change is incorrect.

ltratt:
Sorry, I still don't understand what you're referring to, or what your proposed fix is.

vext01:
So you proposed to change the `PHASE_COMPILED` comment to:
```
const PHASE_COMPILED: usize = 0b00; // Value is a pointer to a malloc'd block...
```

The JIT code isn't malloc'd, it's mmap'd. I'd propose not divulging any information about how it was allocated, just that it's a pointer to it.

ltratt:
I'm with you. See https://github.com/softdevteam/yk/pull/164/commits/dc6f72417369cbb369102e5fab8ee0fec7e18b63.

vext01:
Good job! Please squash.

ltratt:
Squashed.

PR #165, opened 2020-12-04
Audit mt.rs

ltratt:
This PR audits `mt.rs` for issues. Most are fairly harmless bugs (https://github.com/softdevteam/yk/commit/f455452a34857079c109d6fc36780d7820164268, https://github.com/softdevteam/yk/commit/56d0156bb33c3fb79e0785799c017515d8883e11) but https://github.com/softdevteam/yk/commit/b58b6bb5c63185dd5be574bfbbba4183a1e184e9 is, I think, a biggy.

However, with that change our tests now fail, so something else is going on.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -296,7 +296,7 @@ impl MTThread {
                     // FIXME: free up the memory when the trace is no longer used.
                     let ptr = Box::into_raw(Box::new(ct)) as usize;
 
-                    let new_pack = PHASE_COMPILED | (ptr << PHASE_NUM_BITS);
+                    let new_pack = ptr | PHASE_COMPILED;
                     loc.pack.store(new_pack, Ordering::Release);
                     // Free the small block of memory we used as a Location ID.
                     unsafe { Box::from_raw(loc_id) };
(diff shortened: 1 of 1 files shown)


ltratt:
[If nothing else, https://github.com/softdevteam/yk/pull/165/commits/ff795a029ac9d342942dd1fade0bb12537862087 will need squashing.]

ykrt/src/mt.rs
```diff
@@ -244,6 +244,10 @@ impl MTThread {
                             Rc::get_mut(&mut self.inner).unwrap().tracer =
                                 Some((start_tracing(self.inner.tracing_kind), loc_id));
                             return None;
+                        } else {
+                            // We raced with another thread that's also trying to trace this
+                            // Location, so free the malloc'd block.
+                            unsafe { Box::from_raw(loc_id) };
```
vext01, line 250: I wonder if we should just eagerly allocate each thread an ID upon it's creation. Then the ID can be reused across many tracing sessions at zero additional cost.

ltratt: Good idea. Once the current bug is fixed I'll look into that.

vext01: And by extension, is the address of the `MTInner` a better ID?

ltratt: That feature *might* be disappearing at some point. Not

ptersilie:
Looks good to me. Please squash.

ltratt:
Squashed but... the tests don't pass.

bors try

ltratt:
bors try

ltratt:
@ptersilie I suggest I squash this, and we merge it. This does improve some things, even if it doesn't fix the real issue we're after.

ptersilie:
Can do. I can also keep digging for a bit and if I find the bug we can just add it to this PR. Either way is fine by me though.

ltratt:
Let's get this in because we really don't know what the other bug might be.

ltratt:
OK for me to force push?

ptersilie:
Ok. Go ahead!

ltratt:
Force pushed.

PR #166, opened 2020-12-04
We don't need to build an inner loop when we can reuse the outer loop.

ltratt:
This is also safer since it deals with the possibility that another thread has started tracing this location.


ptersilie:
Looks good but needs rebasing.

ltratt:
Rebased.

PR #167, opened 2020-12-04
Stop the PT tracer, when HWTTacer is dropped.

ptersilie:
It's possible for the hardware tracer to be dropped while it is still
tracing, leaving the PT tracer running. This can cause issues with other
threads wanting to trace. Calling `stop_tracing` on the PT tracer when
the hardware tracer is dropped fixes this.

### DIFF
--- a/yktrace/src/hwt/mod.rs
+++ b/yktrace/src/hwt/mod.rs
@@ -38,6 +38,14 @@ impl ThreadTracerImpl for HWTThreadTracer {
     }
 }
 
+impl Drop for HWTThreadTracer {
+    fn drop(&mut self) {
+        // If we haven't stopped the tracer yet, do it now. This might return an error if the
+        // tracer was already stopped, but we don't care as long as it's stopped.
+        let _ = self.ttracer.stop_tracing();
+    }
+}
+
 pub fn start_tracing() -> ThreadTracer {
     let tracer = TracerBuilder::new().build().unwrap();
     let mut ttracer = (*tracer).thread_tracer();


ptersilie:
Typo in the commit. Let me know if ok to force push.

vext01:
Looks good, but why don't we do it one level down, in hwtracer, where we have the ability to know if the tracer is running or not?

ptersilie:
See, that's why I said I'm not very familiar with this code. I didn't know there's a way to check that the tracer is running.

ptersilie:
Mind giving me a hint. There's so many tracers across different files and even repos, and `gd` rarely works on any of them.

ptersilie:
Ah you meant in the `hwtracer` repo. So in `PerfPTThreadTracer` I assume?

vext01:
No worries. To be fair, I had to consult the code a bit myself.

Look [here](https://github.com/softdevteam/hwtracer/blob/7234c02ef870cd6f400af3a5a6a8c89c9fc7b02b/src/backends/perf_pt/mod.rs#L384) and [here](https://github.com/softdevteam/hwtracer/blob/7234c02ef870cd6f400af3a5a6a8c89c9fc7b02b/src/lib.rs#L81)

ptersilie:
Yeah, I already found it.

yktrace/src/hwt/mod.rs
```diff
@@ -38,6 +38,14 @@ impl ThreadTracerImpl for HWTThreadTracer {
     }
 }
 
+impl Drop for HWTThreadTracer {
+    fn drop(&mut self) {
+        // If we haven't stopped the tracer yet, do it now. This might return an error if the
+        // tracer was already stopped, but we don't care as long as it's stopped.
```
bjorn3, line 44: This is not necessary for safety, right? `Drop` is not guaranteed to be called. `mem::forget` is safe.

ptersilie: It's to do with a bug, where the tracer was dropped while it was still tracing, which left the tracer running, and meant other threads weren't allowed to trace. I'm not sure we can do anything about users `mem:forget`ting the tracer, but if they do that, then they shouldn't be surprised if stuff breaks.

ptersilie:
Closed in favour of https://github.com/softdevteam/hwtracer/pull/93.

yktrace/src/hwt/mod.rs
bjorn3: `mem::forget` causing logic bugs is fine, but it must not lead to memory unsafety.

ltratt: If `drop` isn't called, PT will keep on running pointlessly but harmlessly (other than consuming memory + power), so we should be fine here I believe.

PR #168, opened 2020-12-05
Don't create a malloc'd block each time we start tracing.

ltratt:
When we detect a possibly closed loop in the end user's program, we need to determine whether the current thread is the one tracing it or not. We previously allocated a chunk of memory when we started tracing and used its address as a temporary thread ID.

Ideally we would use the system provided thread ID to determine this, but that turns out to be a non-portable minefield. pthreads provides the pthread_t ID type but that's notionally opaque (on OpenBSD, as far as I can tell from a quick search, it's really an opaque struct about which no guarantees are provided, though some programs such as Chrome do seem to rely on it being equivalent to a usize). Some platforms (at least Linux and OpenBSD) use the PID ID type pid_t for threads too and that's guaranteed to be a signed integer though not it seems of a guaranteed size (though on both platforms it's actually a C int, so 32 bits in practise). Rust provides a ThreadID type and an unstable "as_64" function (https://github.com/rust-lang/rust/issues/67939) but that provides no guarantees about how much of the 64-bit integer space is used.

The challenge we have is that whatever ID we use it must fit into (numbits(usize) - numbits(PHASE_TAGS)) i.e. we have 62 bits for the ID on a 64-bit system. If we rely on the system thread IDs it feels like we're storing up a portability time bomb that might explode one day in the dim and distant future. At least for now, the (minimal) extra performance we might get from that doesn't seem worth the danger.

This commit is thus a relatively simple change. Rather than allocating a malloc'd block every time we start tracing a thread, which means we have to be quite careful about freeing memory, we allocate it on thread construction. This simplifies the code slightly and (since it's an aligned address) we can feel fairly safe that it plays nicely with PHASE_TAGS. We could, if we wanted, allocate this block lazily the first time we trace but that feels like an unnecessary optimisation at this point.

This is based on a suggestion from @vext01 and a comment from @bjorn3.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -1,6 +1,7 @@
 #[cfg(test)]
 use std::time::Duration;
 use std::{
+    alloc::{alloc, dealloc, Layout},
     io, mem,
     panic::{catch_unwind, resume_unwind, UnwindSafe},
     rc::Rc,
@@ -238,17 +239,15 @@ impl MTThread {
                             // count further.
                             return None;
                         }
-                        let loc_id = Box::into_raw(Box::new(0u8));
-                        let new_pack = loc_id as usize | PHASE_TRACING;
+                        let new_pack = self.inner.tid as usize | PHASE_TRACING;
                         if loc.pack.compare_and_swap(lp, new_pack, Ordering::Release) == lp {
                             Rc::get_mut(&mut self.inner).unwrap().tracer =
-                                Some((start_tracing(self.inner.tracing_kind), loc_id));
+                                Some((start_tracing(self.inner.tracing_kind), self.inner.tid));
                             return None;
-                        } else {
-                            // We raced with another thread that's also trying to trace this
-                            // Location, so free the malloc'd block.
-                            unsafe { Box::from_raw(loc_id) };
                         }
+                    // We raced with another thread that's (probably) trying to trace this
+                    // Location or (less likely) has already compiled it so we go around the
+                    // loop again to see what we should do.
                     } else {
                         let new_pack = PHASE_COUNTING | ((count + 1) << PHASE_NUM_BITS);
                         if loc.pack.compare_and_swap(lp, new_pack, Ordering::Release) == lp {
@@ -259,18 +258,17 @@ impl MTThread {
                     }
                 }
                 PHASE_TRACING => {
-                    let loc_id = if let Some((_, loc_id)) = self.inner.tracer {
+                    if let Some((_, tid)) = self.inner.tracer {
                         // This thread is tracing something...
-                        if loc_id != ((lp & !PHASE_TAG) as *mut u8) {
+                        if tid != ((lp & !PHASE_TAG) as *mut u8) {
                             // ...but we didn't start at the current Location.
                             return None;
                         }
-                        // ...and we started at this Location, so we've got a complete loop!
-                        loc_id
+                    // ...and we started at this Location, so we've got a complete loop!
                     } else {
                         // Another thread is tracing this location.
                         return None;
-                    };
+                    }
 
                     let sir_trace = Rc::get_mut(&mut self.inner)
                         .unwrap()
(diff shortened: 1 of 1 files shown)


ykrt/src/mt.rs
```diff
@@ -316,6 +311,9 @@ impl MTThread {
 /// The innards of a meta-tracer thread.
 struct MTThreadInner {
     mt: MT,
+    /// A value that uniquely identifies a thread and that is guaranteed not to overlap with
+    /// PHASE_TAG. For simplicities sake this is a pointer to a malloc'd chunk of memory.
```
vext01, line 315: Let's distil some of the wisdom of the PR description into this comment.

Namely, we can briefly mention that most system/language provided thread IDs are non-portable and subject to change. Then it's more accurate to say "For portability, future-proofing and simplicities sake..."

Also I think "overlap" is an odd term to use here. What does it mean to overlap? I think I kind of know what you mean, but it'd be less vague if we again, took some inspiration from the PR description. For example:

> whatever ID we use it must fit into (numbits(usize) - numbits(PHASE_TAGS)) i.e. we have 62 bits for the ID on a 64-bit system.

This explains really well, especially if you mention alignment guarantees of our pointer.

vext01:
As an side-note, I was reading about the 48-bit address space yesterday. The top bits of a pointer in x86_64 (bits 48-63) are "reserved", meaning that you can't use them for your own purposes, but you can throw them away and reconstruct them later: the bits are expected to be either all zeroes or all ones according to the rules of sign-extension.

So in theory, we have even more bits available to us in the "pack" if we compress the pointers in this way, however, this too may be a ticking time bomb. Hasn't there been talk of expanding the 48-bit address space already?

ltratt:
> Hasn't there been talk of expanding the 48-bit address space already?

Yes, I believe upcoming chips will use 56-bits.

ykrt/src/mt.rs
ltratt: I'd prefer not to make the comment *too* long -- see if https://github.com/softdevteam/yk/pull/168/commits/e65104e0a27a693f7f6a9d7a6c916780cc35ab31 works for you.

ltratt:
@vext01 See if https://github.com/softdevteam/yk/pull/168/commits/2ece7cd76309a99519f496c675452770d6a5698c helps.

vext01:
Good, but one question.

> that chunk is aligned to a machine word (or greater)

Why "or greater". Could be 9-byte aligned?

bjorn3:
It could be 16 byte aligned, 32 byte, ... If you generate random 8 byte aligned addresses, some will also be aligned to a higher amount.

Alignment is always a power of two, 9 byte alignment is impossible.

ltratt:
@bjorn3 has hit the nail on the head.

OK to squash?

vext01:
> It could be 16 byte aligned, 32 byte...

Yes, but my point is, if we are asking for 8-byte aligned memory, any higher-aligned pointer that happens to come back (16, 32, ...), is by definition also 8-byte aligned. So to say 8-byte aligned "or greater" is redundant and IMHO actually casts doubt on what is meant.

(My question about 9-byte alignment was intended to be rhetorical to emphasise how "or greater" could be misinterpreted. Alas, it didn't work :P)

So I think it'd be clearer to drop "or greater", and I am probably guilty of highest bikeshedding at this point!

ltratt:
I'm going to be awkward and say that I don't think many people will interpret "greater than 8" as "9" :) It's more to make clear that some platforms will *only* give you multiples of 16 even if you ask for multiples of 8. Anyway, if you would still like the comment change, I'm willing to do it. Let me know!

vext01:
It's not a mole hill I'll die on 😆

Please squash (optionally amending the comment at the same time, if you see fit).

> It's more to make clear that some platforms will only give you multiples of 16.

Now that I didn't know.

ltratt:
OK, I've removed "or greater" -- I see your point. Force pushed.

vext01:
Boris r+

vext01:
bors r+

(Sorry, autocorrect!)

PR #171, opened 2020-12-07
Compile traces in parallel

ltratt:
This PR compiles traces in a thread, allowing the user's interpreter to keep executing while this happens. The major commit is https://github.com/softdevteam/yk/commit/94defd4d52a34c55cd91a3d35169f93c0c17c682 which contains much more information. The core change is less than 20 lines of code, but I have added significant comments, and make a few other changes, which I hope make this part of the code a little less scary than before.

However, one test sometimes (but not always!) fails as follows:

```
running 1 test
test mt::tests::dumb_interpreter ... FAILED

failures:

---- mt::tests::dumb_interpreter stdout ----
thread 'mt::tests::dumb_interpreter' panicked at 'not yet implemented: Guard failure!', ykrt/src/mt.rs:274:21
```

My working guess is that this probably results from us not quite chopping the head/tail of a trace off properly, but I'm not sure. Any thoughts @vext01 or @ptersilie?

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -42,13 +42,13 @@ const PHASE_COUNTING: usize = 0b10; // The value specifies how many times we've 
 /// nodes that can't be control points).
 #[derive(Debug)]
 pub struct Location {
-    pack: AtomicUsize,
+    state: AtomicUsize,
 }
 
 impl Location {
     pub fn new() -> Self {
         Self {
-            pack: AtomicUsize::new(PHASE_COUNTING),
+            state: AtomicUsize::new(PHASE_COUNTING),
         }
     }
 }
@@ -232,7 +232,7 @@ impl MTThread {
         loop {
             // We need Acquire ordering, as PHASE_COMPILED will need to read information written to
             // external data as a result of the PHASE_TRACING -> PHASE_COMPILED transition.
-            let lp = loc.pack.load(Ordering::Acquire);
+            let lp = loc.state.load(Ordering::Acquire);
             match lp & PHASE_TAG {
                 PHASE_COUNTING => {
                     let count = (lp & !PHASE_TAG) >> 2;
@@ -242,8 +242,8 @@ impl MTThread {
                             // count further.
                             return None;
                         }
-                        let new_pack = self.inner.tid as usize | PHASE_TRACING;
-                        if loc.pack.compare_and_swap(lp, new_pack, Ordering::Release) == lp {
+                        let new_state = self.inner.tid as usize | PHASE_TRACING;
+                        if loc.state.compare_and_swap(lp, new_state, Ordering::Release) == lp {
                             Rc::get_mut(&mut self.inner).unwrap().tracer =
                                 Some((start_tracing(self.inner.tracing_kind), self.inner.tid));
                             return None;
@@ -252,8 +252,8 @@ impl MTThread {
                     // Location or (less likely) has already compiled it so we go around the
                     // loop again to see what we should do.
                     } else {
-                        let new_pack = PHASE_COUNTING | ((count + 1) << PHASE_NUM_BITS);
-                        if loc.pack.compare_and_swap(lp, new_pack, Ordering::Release) == lp {
+                        let new_state = PHASE_COUNTING | ((count + 1) << PHASE_NUM_BITS);
+                        if loc.state.compare_and_swap(lp, new_state, Ordering::Release) == lp {
                             return None;
                         }
                         // We raced with another thread, but we'd still like to try incrementing
@@ -284,14 +284,14 @@ impl MTThread {
                     let tir_trace = TirTrace::new(&*SIR, &*sir_trace).unwrap();
                     let ct = TraceCompiler::<I>::compile(tir_trace);
                     // Now that we have a compiled trace, we need to put a pointer to it inside
-                    // the corresponding Location's pack. To inhibit ExecutableBuffer's Drop,
-                    // we move it into a box and forget it: the pointer in the pack is now the
+                    // the corresponding Location's state. To inhibit ExecutableBuffer's Drop,
+                    // we move it into a box and forget it: the pointer in the state is now the
                     // sole handle on the trace.
                     // FIXME: free up the memory when the trace is no longer used.
                     let ptr = Box::into_raw(Box::new(ct)) as usize;
 
-                    let new_pack = ptr | PHASE_COMPILED;
-                    loc.pack.store(new_pack, Ordering::Release);
+                    let new_state = ptr | PHASE_COMPILED;
+                    loc.state.store(new_state, Ordering::Release);
                     Rc::get_mut(&mut self.inner).unwrap().tracer = None;
                     return None;
                 }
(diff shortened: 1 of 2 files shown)


ykrt/src/mt.rs
```diff
@@ -281,28 +348,50 @@ impl MTThread {
...
                     return None;
                 }
+                PHASE_COMPILING => {
+                    let compiling =
+                        unsafe { Box::from_raw((lp & !PHASE_TAG) as *mut CompilingTrace<I>) };
+                    match compiling.rcv.try_recv() {
+                        Ok(ct) => {
+                            // If we've got a result via the channel then we can ignore any
+                            // subsequent errors from the compiling thread.
+                            compiling.handle.join().ok();
```
bjorn3, line 373: I don't think it is necessary to wait on the thread to complete. If you drop the `JoinGuard`, the thread will be detached, so joining is neither required nor allowed to prevent a resource leak.

ltratt: I'm not sure if it's strictly necessary or not and the docs aren't really clear. However, I know from other systems that not calling `join` can create problems. [This blog post](https://matklad.github.io/2019/08/23/join-your-threads.html) suggests one should do it in Rust too, though I'm not entirely sure if the reasoning in there is sound or not. I'd be happy if I definitely don't need to do the `join` as it would help simplify the code a bit more, but I feel I need a bit more documentary evidence to be sure :/

ltratt: I now see what you mean by:

> If you drop the JoinGuard, the thread will be detached

it literally calls `pthread_detach` (or equivalent). Which means you're absolutely right and there's no need to manually `join` things. Excellent! So that simplifies things in https://github.com/softdevteam/yk/pull/171/commits/6e03c452b3c4701656b9981fa41a0f77baded7f8 a little bit, which is always nice. Thanks for the pointer!

bjorn3: > This blog post suggests one should do it in Rust too, though I'm not entirely sure if the reasoning in there is sound or not.

The first point doesn't really apply here as the thread is known to almost exist anyway. Same for the second point. As for the third point, you are ignoring the panic anyway.

ltratt:
Thanks to bug hunting from @vext01 and @ptersilie it turns out the problem I originally thought was present was due to my own idiocy. Fixed in https://github.com/softdevteam/yk/pull/171/commits/893e75d5b2873d6ae88fcd6b3e6a8ceb485f442b.

This PR is now ready for normal, full review IMHO.

ykrt/src/mt.rs
```diff
@@ -242,8 +306,8 @@ impl MTThread {
                             // count further.
                             return None;
                         }
-                        let new_pack = self.inner.tid as usize | PHASE_TRACING;
-                        if loc.pack.compare_and_swap(lp, new_pack, Ordering::Release) == lp {
+                        let new_state = self.inner.tid as usize | PHASE_TRACING;
```
ptersilie, line 309: Did we say we always use `usize::from` from now on, or this is case an exception?

ltratt: This is a pointer -> usize conversion, so I think we have to use `as` unfortunately.

ptersilie: Ah, that's fine then.

ptersilie:
This looks fine to me. If @vext01, feel free to squash.

ykrt/src/mt.rs
```diff
@@ -24,10 +27,11 @@ const PHASE_NUM_BITS: usize = 2;
 // accessed without any further operations after the initial tag check.
 const PHASE_TAG: usize = 0b11; // All of the other PHASE_ tags must fit in this.
 const PHASE_COMPILED: usize = 0b00; // Value is a pointer to a chunk of memory containing a
-                                    // compiled trace.
+                                    // CompiledTrace<I>.
 const PHASE_TRACING: usize = 0b01; // Value is a pointer to a malloc'd block that allows us to
                                    // precisely identify when we have reached the same Location.
-const PHASE_COUNTING: usize = 0b10; // The value specifies how many times we've seen this Location.
+const PHASE_COMPILING: usize = 0b10; // Value is a pointer to an `Box<CompilingTrace<I>>`.
```
vext01, line 33: an -> a?

ykrt/src/mt.rs
```diff
@@ -41,14 +45,62 @@ const PHASE_COUNTING: usize = 0b10; // The value specifies how many times we've
...
 #[derive(Debug)]
-pub struct Location {
-    pack: AtomicUsize,
+pub struct Location<I> {
+    // A Location is a state machine which operates as follows (where PHASE_COUNTING is the start
+    // state):
+    //
+    //               ┌────────────────────────────────────────┐
+    //               │                                        │ ─────────────┐
+    //   reprofile   │             PHASE_COUNTING             │              │
```
vext01, line 54: We don't implement that reprofile stage yet, do we?

ykrt/src/mt.rs
```diff
@@ -41,14 +45,62 @@ const PHASE_COUNTING: usize = 0b10; // The value specifies how many times we've
...
+    //  │              ▼
+    //  │            ┌────────────────────┐
+    //  │            │                    │ ──────┐
+    //  │            │   PHASE_COMPILED   │       │
+    //  └─────────── │                    │ ◀─────┘
+    //               └────────────────────┘
+    //
+    // We hope that a Location soon reaches PHASE_COMPILED (aka "the happy state") and stays there.
+    state: AtomicUsize,
+    phantom: PhantomData<I>,
```
vext01, line 85: The phantom is purely for additional type safety? To stop us using locations for one control point with another?

ykrt/src/mt.rs
ltratt: Not yet, no.

ykrt/src/mt.rs
ltratt: Yes and yes.

ykrt/src/mt.rs
```diff
@@ -41,14 +45,62 @@ const PHASE_COUNTING: usize = 0b10; // The value specifies how many times we've
...
-            pack: AtomicUsize::new(PHASE_COUNTING),
+            state: AtomicUsize::new(PHASE_COUNTING),
+            phantom: PhantomData,
+        }
+    }
+}
+
+impl<I> Drop for Location<I> {
+    fn drop(&mut self) {
+        let lp = self.state.load(Ordering::Acquire);
```
vext01, line 99: Can this be relaxed ordering?

ykrt/src/mt.rs
ltratt: Fixed in https://github.com/softdevteam/yk/pull/171/commits/51c3bb80a3bd6ffc143543d45064abd492be8831.

ykrt/src/mt.rs
ltratt: Good point! Fixed in https://github.com/softdevteam/yk/pull/171/commits/d9bfa2e563763dfcca8dc374af2e52adfbdb9fbe.

ykrt/src/mt.rs
```diff
@@ -378,15 +464,18 @@ mod tests {
...
-        assert_eq!(lp.pack.load(Ordering::Relaxed) & PHASE_TAG, PHASE_COUNTING);
+        assert_eq!(lp.state.load(Ordering::Relaxed) & PHASE_TAG, PHASE_COUNTING);
         mtt.control_point(Some(&lp), dummy_step, &mut io);
-        assert_eq!(lp.pack.load(Ordering::Relaxed) & PHASE_TAG, PHASE_TRACING);
+        assert_eq!(lp.state.load(Ordering::Relaxed) & PHASE_TAG, PHASE_TRACING);
         mtt.control_point(Some(&lp), dummy_step, &mut io);
-        assert_eq!(lp.pack.load(Ordering::Relaxed) & PHASE_TAG, PHASE_COMPILED);
+        assert_eq!(
+            lp.state.load(Ordering::Relaxed) & PHASE_TAG,
+            PHASE_COMPILING
```
vext01, line 477: Could PHASE_COMPILED be observed here too?

vext01:
Please squash.

vext01:
Sorry, there's on more outstanding comment. My bad.

ykrt/src/mt.rs
ltratt: Good point! Fixed in https://github.com/softdevteam/yk/pull/171/commits/a12aac14e0784954d3adb9d350c134e0726c6e4f.

ltratt:
@vext01 I think that's now everything?

vext01:
Yep. Please squash.

ltratt:
Squashed.

PR #172, opened 2020-12-08
Use an interval tree for SIR labels.

vext01:
To make a SIR trace, there's a "mapping" stage which translates PT virtual addresses to SIR locations. Before this change, for each block libipt gave us we had to do a linear search over a list of label addresses, searching for those that coincide. That was slow.

This change puts the labels in a data structure better suited: an interval tree. Then computing the coinciding intervals can be done in log time, rather than linear time.

Rough benchmarking of test suite:

before:
```
1: cargo test
real        9.363       0.104       9.237       9.395       9.480
user        98.112      0.361       97.681      98.081      98.720
sys         6.037       0.553       5.157       6.012       6.804
```

after:
```
real        4.574       0.177       4.310       4.543       4.794
user        19.170      0.180       18.889      19.195      19.382
sys         9.995       0.273       9.666       9.997       10.446
```

### DIFF
--- a/yktrace/Cargo.toml
+++ b/yktrace/Cargo.toml
@@ -11,6 +11,7 @@ fallible-iterator = "0.2.0"
 fxhash = "0.2.1"
 gimli = "0.23.0"
 hwtracer = { git = "https://github.com/softdevteam/hwtracer" }
+intervaltree = "0.2"
 lazy_static = "1.4.0"
 libc = "0.2.80"
 memmap = "0.7.0"
--- a/yktrace/src/hwt/mapper.rs
+++ b/yktrace/src/hwt/mapper.rs
@@ -3,8 +3,9 @@ use phdrs::objects;
 
 use crate::sir::SirLoc;
 use hwtracer::{HWTracerError, Trace};
+use intervaltree::IntervalTree;
 use lazy_static::lazy_static;
-use std::{convert::TryFrom, env, fs};
+use std::{convert::TryFrom, env, fs, iter::FromIterator};
 use ykpack::SirLabel;
 
 lazy_static! {
@@ -17,7 +18,7 @@ lazy_static! {
     /// to be a lazy static, loaded only once and shared.
     ///
     /// FIXME if we want to support dlopen(), we will have to rethink this.
-    static ref LABELS: Vec<SirLabel> = load_labels();
+    static ref LABELS: IntervalTree<usize, SirLabel> = load_labels();
 }
 
 pub struct HWTMapper {
@@ -31,7 +32,11 @@ impl HWTMapper {
     }
 
     /// Maps each entry of a hardware trace to the appropriate SirLoc.
-    pub fn map(&self, trace: Box<dyn Trace>) -> Result<Vec<SirLoc>, HWTracerError> {
+    ///
+    /// For each block in the trace, the interval tree is queried for labels coinciding with the
+    /// block. Label addresses which coincide are therefore contained within the block, and are
+    /// thus part of the SIR trace.
+    pub fn map_trace(&self, trace: Box<dyn Trace>) -> Result<Vec<SirLoc>, HWTracerError> {
         let mut annotrace = Vec::new();
         for block in trace.iter_blocks() {
             let block = block?;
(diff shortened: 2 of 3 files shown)


yktrace/Cargo.toml
```diff
@@ -11,6 +11,7 @@ fallible-iterator = "0.2.0"
 fxhash = "0.2.1"
 gimli = "0.23.0"
 hwtracer = { git = "https://github.com/softdevteam/hwtracer" }
+intervaltree = "0.2.6"
```
ltratt, line 14: In general, I suggest just saying `0.2` (cargo will treat `0.2.6` as a *minimum* not a mandatory requirement, so it suggests a level of precision that doesn't really exist in practise).

yktrace/src/hwt/mapper.rs
```diff
@@ -47,25 +52,20 @@ impl HWTMapper {
...
             // b) `labels` is sorted, so the blocks will be appended to the trace in the
             // correct order.
-            let mut locs = Vec::new();
-            for lbl in &*LABELS {
-                if lbl.off >= start_addr && lbl.off <= end_addr {
-                    // Found matching label.
-                    // Store the virtual address alongside the first basic block, so we can turn
-                    // inlined functions into calls during tracing.
+            let mut locs = LABELS
+                .query(start_addr..(end_addr + 1))
```
ltratt, line 56: Why `+ 1`?

yktrace/src/hwt/mapper.rs
```diff
@@ -31,6 +32,10 @@ impl HWTMapper {
     }
 
     /// Maps each entry of a hardware trace to the appropriate SirLoc.
+    ///
+    /// For each block in the trace, the interval tree is queried for labels coinciding with the
+    /// block. Label addresses which coincide are therefore contained within the block, and are
+    /// thus part of the SIR trace.
     pub fn map(&self, trace: Box<dyn Trace>) -> Result<Vec<SirLoc>, HWTracerError> {
```
ltratt, line 39: I wonder if we can use a slightly more specific name than `map`? Just because `map` has a slightly generic meaning in Rust...

yktrace/src/hwt/mapper.rs
vext01: The library doesn't support inclusive upper-bound ranges for some reason:

```
error[E0308]: mismatched types
  --> yktrace/src/hwt/mapper.rs:56:24
   |
56 |                 .query(start_addr..=end_addr)
   |                        ^^^^^^^^^^^^^^^^^^^^^ expected struct `std::ops::Range`, found struct `RangeInclusive`
   |
   = note: expected struct `std::ops::Range<usize>`
              found struct `RangeInclusive<usize>`

error[E0277]: the trait bound `Element<_, _>: From<(RangeInclusive<usize>, SirLabel)>` is not satisfied
   --> yktrace/src/hwt/mapper.rs:91:5
    |
91  |     IntervalTree::from_iter(vec.into_iter().map(|l| (l.off..=l.off, l)))
    |     ^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<(RangeInclusive<usize>, SirLabel)>` is not implemented for `Element<_, _>`
    | 
   ::: /home/vext01/research/yorick/ykrustc/library/core/src/iter/traits/collect.rs:113:34
    |
113 |     fn from_iter<T: IntoIterator<Item = A>>(iter: T) -> Self;
    |                                  -------- required by this bound in `from_iter`
    |
    = help: the following implementations were found:
              <Element<K, V> as From<(std::ops::Range<K>, V)>>
    = note: required because of the requirements on the impl of `Into<Element<_, _>>` for `(RangeInclusive<usize>, SirLabel)`
    = note: required because of the requirements on the impl of `FromIterator<(RangeInclusive<usize>, SirLabel)>` for `IntervalTree<_, _>`
```

yktrace/Cargo.toml
vext01: I stopped doing that ever since I discovered that `cargo update` inserts a full version number. This way we get less noise from running `cargo update`.

yktrace/src/hwt/mapper.rs
ltratt: Ah fair enough.

yktrace/src/hwt/mapper.rs
vext01: Fixed in 0df753f

yktrace/Cargo.toml
ltratt: Inserts a version number where? I'm not sure how this lessens noise but I'm prepared to be taught!

yktrace/src/hwt/mapper.rs
ltratt: Works for me.

yktrace/Cargo.toml
vext01: I was wrong.

I thought that if I have say "0.2" in the Cargo.toml, that `cargo update` would expand it to e.g. `0.2.1`. Nope.

Fixed in f0643c9

yktrace/src/hwt/mapper.rs
vext01: Raised an issue against the library:
https://github.com/main--/rust-intervaltree/issues/8

ltratt:
Please squash.

vext01:
Splat.

PR #174, opened 2020-12-09
Implement section caching to speed up SIR body decoding.

vext01:
Profiling suggested that quite a bit of time was spent looking up
sections in the executable (by name). This change caches references to
each section's bytes during initialisation (as we load the headers).

Rough benchmarking of the test suite:

Before:
```
real        4.810       0.285       4.463       4.637       5.230
user        22.695      0.245       22.242      22.679      23.145
sys         10.872      0.634       9.409       11.072      11.698
```

After:
```
real        3.874       0.156       3.689       3.869       4.094
user        11.645      0.165       11.279      11.707      11.825
sys         10.790      0.402       9.932       10.759      11.573
```

### DIFF
--- a/yktrace/src/sir.rs
+++ b/yktrace/src/sir.rs
@@ -31,17 +31,23 @@ pub struct Sir<'m> {
     /// Maps a codegen unit hash to a `(section-name, header, header-size)` tuple.
     hdrs: FxHashMap<CguHash, (String, SirHeader, SirOffset)>,
     /// The current executable's ELF information.
-    exe_obj: object::File<'m>
+    exe_obj: object::File<'m>,
+    /// Section cache to avoid expensive `object::File::section_by_name()` calls.
+    sec_cache: FxHashMap<String, &'m [u8]>
 }
 
 impl<'m> Sir<'m> {
     pub fn new(mmap: &'m Mmap) -> Result<Self, ()> {
         // SAFETY: Not really, we hope that nobody changes the file underneath our feet.
         let mut hdrs = FxHashMap::default();
+        let mut sec_cache = FxHashMap::default();
         let exe_obj = object::File::parse(&*mmap).unwrap();
         for sec in exe_obj.sections() {
-            if sec.name().unwrap().starts_with(ykpack::SIR_SECTION_PREFIX) {
-                let mut curs = Cursor::new(sec.data().unwrap());
+            let sec_name = sec.name().unwrap();
+            let sec_data = sec.data().unwrap();
+            sec_cache.insert(sec_name.to_owned(), sec_data);
+            if sec_name.starts_with(ykpack::SIR_SECTION_PREFIX) {
+                let mut curs = Cursor::new(sec_data);
                 let mut dec = Decoder::from(&mut curs);
                 let hdr = if let Pack::Header(hdr) = dec.next().unwrap().unwrap() {
                     hdr
@@ -49,18 +55,18 @@ impl<'m> Sir<'m> {
                     panic!("missing sir header");
                 };
                 let hdr_size = usize::try_from(curs.seek(SeekFrom::Current(0)).unwrap()).unwrap();
-                hdrs.insert(
-                    hdr.cgu_hash,
-                    (sec.name().unwrap().to_owned(), hdr, hdr_size)
-                );
+                hdrs.insert(hdr.cgu_hash, (sec_name.to_owned(), hdr, hdr_size));
             }
         }
-        Ok(Self { hdrs, exe_obj })
+        Ok(Self {
+            hdrs,
+            exe_obj,
+            sec_cache
+        })
     }
 
     fn cursor_for_section(&self, sec_name: &str) -> Cursor<&[u8]> {
-        let sec = self.exe_obj.section_by_name(&sec_name).unwrap();
-        Cursor::new(sec.data().unwrap())
+        Cursor::new(self.sec_cache[sec_name])
     }
 
     /// Decode a type in a named section, at an absolute offset from the beginning of that section.


ltratt:
That's a huge change in `user` time! Can you do me a favour and benchmark this with `--test-threads=1` before and after, as I think that'll give us a better picture of what's really going on (the `sys` time seems hugely high, but maybe that's a consequence of `mmap`?).

vext01:
Before:
```
1: env RUST_TEST_THREADS=1 cargo test
real        14.841      0.059       14.751      14.856      14.931      
user        14.525      0.032       14.487      14.519      14.589      
sys         2.238       0.058       2.141       2.225       2.330       
```

After:
```
1: env RUST_TEST_THREADS=1 cargo test
real        7.104       0.023       7.068       7.104       7.159
user        7.161       0.053       7.080       7.166       7.248
sys         1.058       0.044       0.997       1.061       1.128
```

What are you thinking?

ltratt:
This gives us a much better picture of the speed-up because we're no longer measuring accidental things like how many cores we have. 2x is not bad at all!

yktrace/src/sir.rs
```diff
@@ -31,36 +31,42 @@ pub struct Sir<'m> {
     /// Maps a codegen unit hash to a `(section-name, header, header-size)` tuple.
     hdrs: FxHashMap<CguHash, (String, SirHeader, SirOffset)>,
     /// The current executable's ELF information.
-    exe_obj: object::File<'m>
+    exe_obj: object::File<'m>,
+    /// Section cache to avoid expensive `object::File::section_by_name()` calls.
+    sec_cache: FxHashMap<String, &'m [u8]>
```
ltratt, line 36: Can this be `FxHashMap<&'m [u8], &'m [u8]>` or similar?

vext01:
> 2x is not bad at all!

Better than a kick in the teeth!

yktrace/src/sir.rs
vext01: I'm not sure what you mean. We need to look up the section data by symbol name (i.e. a string).

yktrace/src/sir.rs
```diff
@@ -31,36 +31,42 @@ pub struct Sir<'m> {
...
                     hdr
                 } else {
                     panic!("missing sir header");
                 };
                 let hdr_size = usize::try_from(curs.seek(SeekFrom::Current(0)).unwrap()).unwrap();
-                hdrs.insert(
-                    hdr.cgu_hash,
-                    (sec.name().unwrap().to_owned(), hdr, hdr_size)
-                );
+                hdrs.insert(hdr.cgu_hash, (sec_name.to_owned(), hdr, hdr_size));
```
ltratt, line 58: Can we also do `&'m` to this `to_owned`?

yktrace/src/sir.rs
vext01: Ah, you mean as an `&str`?

ltratt: Yes, but an *owned* string or a *borrowed* string for the key? The value is borrowed (which is cheap), but not the key.

vext01: Yeah, gotcha. I'll try it.

vext01: So I think you mean something like this:
```
diff --git a/yktrace/src/sir.rs b/yktrace/src/sir.rs
index a25e545..a355e04 100644
--- a/yktrace/src/sir.rs
+++ b/yktrace/src/sir.rs
@@ -33,7 +33,7 @@ pub struct Sir<'m> {
     /// The current executable's ELF information.
     exe_obj: object::File<'m>,
     /// Section cache to avoid expensive `object::File::section_by_name()` calls.
-    sec_cache: FxHashMap<String, &'m [u8]>
+    sec_cache: FxHashMap<&'m str, &'m [u8]>
 }

 impl<'m> Sir<'m> {
@@ -45,7 +45,7 @@ impl<'m> Sir<'m> {
         for sec in exe_obj.sections() {
             let sec_name = sec.name().unwrap();
             let sec_data = sec.data().unwrap();
-            sec_cache.insert(sec_name.to_owned(), sec_data);
+            sec_cache.insert(sec_name, sec_data);
             if sec_name.starts_with(ykpack::SIR_SECTION_PREFIX) {
                 let mut curs = Cursor::new(sec_data);
                 let mut dec = Decoder::from(&mut curs);
```

This gives:
```
error[E0597]: `exe_obj` does not live long enough
  --> yktrace/src/sir.rs:45:20
   |
39 | impl<'m> Sir<'m> {
   |      -- lifetime `'m` defined here
...
45 |         for sec in exe_obj.sections() {
   |                    ^^^^^^^-----------
   |                    |
   |                    borrowed value does not live long enough
   |                    argument requires that `exe_obj` is borrowed for `'m`
...
66 |     }
   |     - `exe_obj` dropped here while still borrowed
   
error[E0597]: `sec` does not live long enough
  --> yktrace/src/sir.rs:46:28
   |
39 | impl<'m> Sir<'m> {
   |      -- lifetime `'m` defined here
...
46 |             let sec_name = sec.name().unwrap();
   |                            ^^^-------
   |                            |
   |                            borrowed value does not live long enough
   |                            argument requires that `sec` is borrowed for `'m`
...
60 |         }
   |         - `sec` dropped here while still borrowed
   
error[E0505]: cannot move out of `exe_obj` because it is borrowed
  --> yktrace/src/sir.rs:63:13
   |
39 | impl<'m> Sir<'m> {
   |      -- lifetime `'m` defined here
...
45 |         for sec in exe_obj.sections() {
   |                    ------------------
   |                    |
   |                    borrow of `exe_obj` occurs here
   |                    argument requires that `exe_obj` is borrowed for `'m`
...
63 |             exe_obj,
   |             ^^^^^^^ move out of `exe_obj` occurs here
```

So the section name dies with the section.

I briefly tried making the section hang around longer, but quickly got into lifetime spaghetti.

Let's take the 2x and move on? The clone of the section names happens only once at initialisation time, so I doubt we'd have got a big speed up anyway.

ltratt: OK.

PR #175, opened 2020-12-09
Body and type caches

vext01:
These changes implement body and type caching to avoid repeated SIR decodings.

Below are some measurements. It seems like body caching gives a decent speedup in single-threaded scenarios, but not in multi-threaded scenarios. Type caching however doesn't really offer much of an improvement (kill that commit?).

My guess as to why we don't win much in multi-threaded scenarios is that there's mad contention on the `RwLock` protecting the cache(s). There are also `Arc`s which are necessary to appease the borrow checker.

One idea would be to use a concurrent hashmap library like [this](https://crates.io/crates/dashmap) or [this](https://crates.io/crates/storage-map). The latter takes advantage of the map being monotonic (in the sense that keys are only ever added), which fits our use case (for now).

For now raising this as a draft for discussion.

## Before

Single test thread:
```
real        7.122       0.039       7.072       7.125       7.208
user        7.166       0.065       7.058       7.175       7.247
sys         1.051       0.055       0.969       1.051       1.139
```

Many test threads:
```
real        3.656       0.147       3.432       3.747       3.798
user        12.089      0.248       11.765      11.998      12.563
sys         6.247       0.580       5.451       6.220       7.211
```

## Only Body Cache

Single test thread:
```
real        3.452       0.037       3.396       3.462       3.519
user        2.949       0.042       2.876       2.943       3.024
sys         0.770       0.050       0.691       0.778       0.863
```

Many test threads:
```
real        2.861       0.057       2.782       2.856       2.959
user        6.453       0.104       6.329       6.463       6.640
sys         6.046       0.405       5.414       6.079       6.652
```

## Both Caches

Single test thread:
```
real        3.470       0.029       3.407       3.473       3.511
user        2.984       0.036       2.911       2.990       3.034
sys         0.754       0.042       0.710       0.742       0.823
```

Many test threads:
```
real        2.931       0.049       2.862       2.927       2.999
user        6.304       0.149       6.117       6.278       6.660
sys         6.315       0.389       5.734       6.226       7.133
```

### DIFF
--- a/yktrace/src/sir.rs
+++ b/yktrace/src/sir.rs
@@ -10,7 +10,8 @@ use std::{
     fmt::{self, Debug, Display, Write},
     fs::File,
     io::{Cursor, Seek, SeekFrom},
-    iter::Iterator
+    iter::Iterator,
+    sync::{Arc, RwLock}
 };
 use ykpack::{self, Body, BodyFlags, CguHash, Decoder, Pack, SirHeader, SirOffset};
 
@@ -33,7 +34,9 @@ pub struct Sir<'m> {
     /// The current executable's ELF information.
     exe_obj: object::File<'m>,
     /// Section cache to avoid expensive `object::File::section_by_name()` calls.
-    sec_cache: FxHashMap<String, &'m [u8]>
+    sec_cache: FxHashMap<String, &'m [u8]>,
+    /// Body cache, to avoid repeated decodings.
+    body_cache: RwLock<FxHashMap<String, Option<Arc<Body>>>>
 }
 
 impl<'m> Sir<'m> {
@@ -61,7 +64,8 @@ impl<'m> Sir<'m> {
         Ok(Self {
             hdrs,
             exe_obj,
-            sec_cache
+            sec_cache,
+            body_cache: Default::default()
         })
     }
 
@@ -105,12 +109,31 @@ impl<'m> Sir<'m> {
 
     /// Get the body data for the given symbol name.
     /// Returns None if not found.
-    pub fn body(&self, body_sym: &str) -> Option<Body> {
+    pub fn body(&self, body_sym: &str) -> Option<Arc<Body>> {
+        {
+            let rd = self.body_cache.read().unwrap();
+            if let Some(body) = rd.get(body_sym) {
+                // Cache hit, return a reference to the previously decoded body.
+                return body.clone();
+            }
+        } // Drop the RwLock's read() to prevent deadlocking.
+
+        // Cache miss. Decode the body and update the cache.
         for (sec_name, hdr, hdr_size) in SIR.hdrs.values() {
             if let Some(off) = hdr.bodies.get(body_sym) {
-                return Some(self.decode_body(sec_name, hdr_size + off));
+                let body = self.decode_body(sec_name, hdr_size + off);
+                let mut wr = self.body_cache.write().unwrap();
+                let arc = Arc::new(body);
+                wr.insert(body_sym.to_owned(), Some(arc.clone()));
+                return Some(arc);
             }
         }
+
+        // The body is absent. Update the cache with a `None` to prevent repeated searches.
+        self.body_cache
+            .write()
+            .unwrap()
+            .insert(body_sym.to_owned(), None);
         None
     }
 }
--- a/yktrace/src/sir.rs
+++ b/yktrace/src/sir.rs
@@ -13,7 +13,7 @@ use std::{
     iter::Iterator,
     sync::{Arc, RwLock}
 };
-use ykpack::{self, Body, BodyFlags, CguHash, Decoder, Pack, SirHeader, SirOffset};
+use ykpack::{self, Body, BodyFlags, CguHash, Decoder, Pack, SirHeader, SirOffset, Ty};
 
 lazy_static! {
     pub static ref EXE_MMAP: Mmap =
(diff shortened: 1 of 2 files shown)


ltratt:
Are these things immutable once they've been created? If so we should be able to share them across threads without locking.

vext01:
> Are these things immutable once they've been created?

The maps are not, but the values in the maps are.

So I should implement `Sync` on them?

ltratt:
Another possibility (may or may not work here!). If you know all the keys in advance, you can populate the hashmap so that it's something like `HashMap<&str, Arc<Mutex<Option<X>>>>` so that you end up with a per-SIR-block lock. Obviously this makes more maps, but means multiple threads don't really to share a single "global" lock.

vext01:
Interesting idea.

So, I have a few things to try tomorrow then.

yktrace/src/sir.rs
```diff
@@ -33,7 +34,11 @@ pub struct Sir<'m> {
     /// The current executable's ELF information.
     exe_obj: object::File<'m>,
     /// Section cache to avoid expensive `object::File::section_by_name()` calls.
-    sec_cache: FxHashMap<String, &'m [u8]>
+    sec_cache: FxHashMap<String, &'m [u8]>,
+    /// Body cache, to avoid repeated decodings.
+    body_cache: RwLock<FxHashMap<String, Option<Arc<Body>>>>,
```
bjorn3, line 39: What if instead of having a lock around the map, you have a map that is pre-filled with a lock for each body? While this requires decoding the body names, you only need to decode the actual body when it is used. This could reduce contention a lot.

vext01: I think that's similar to what Laurie proposed above.

I'm going to try this (and a few other things) tomorrow.

bjorn3: Missed that comment.

vext01:
OK, I've tried a handful of different strategies. The below results are for *both* body and type caching.

I included the mutex strategy as a sanity check, and because it ensures that no element is ever decoded twice (the other strategies make it unlikely, but don't guarantee that due to (safe) races). Sadly having that guarantee doesn't really help.

I'm interested to see what you make of this -- it's not as clear-cut as I'd hoped.

## RWLocked FxHashMap (as before)

One thread:
```
real        3.494       0.076       3.427       3.464       3.697
user        3.032       0.126       2.923       3.001       3.395
sys         0.772       0.071       0.688       0.747       0.946
```

Many threads:
```
real        3.156       0.049       3.089       3.149       3.245
user        5.799       0.143       5.529       5.792       5.981
sys         10.949      0.411       10.468      10.896      11.874
```

## Using a Dashmap<FxHash> (566827e49b3fc2aaa87bdbe41337702a018ea5a2)

One thread:
```
real        3.575       0.059       3.534       3.559       3.748
user        3.127       0.140       3.041       3.080       3.538
sys         0.794       0.047       0.700       0.794       0.893
```

Many threads:
```
real        3.272       0.044       3.215       3.274       3.374
user        5.906       0.251       5.532       6.018       6.206
sys         11.058      0.348       10.555      11.060      11.617
```

## Using a StorageMap<FxHash> (55df8f0759d23b9b9bf4712a64e06e6dde60a65e)

CAVEAT: When storing `String` keys in a `StorageMap`, queries require a `&String`, not a `&str`. They don't coerce for some reason, and this causes extra `.to_string()` calls which clone.

One thread:
```
real        3.537       0.028       3.495       3.546       3.573
user        3.061       0.037       3.012       3.055       3.123
sys         0.777       0.040       0.719       0.768       0.846
```

Many threads:
```
real        3.300       0.067       3.199       3.307       3.410
user        5.780       0.253       5.460       5.722       6.119
sys         11.524      0.392       11.072      11.398      12.514
```

## Using a pre-filled immutable FxHashMap with individually RWLocked (so inner-mutable) values (b07f8d1973503dcf642892803ca6f53a6e4cda2e)

One thread:
```
real        3.868       0.037       3.825       3.862       3.923
user        3.372       0.031       3.340       3.358       3.436
sys         0.849       0.040       0.765       0.846       0.902
```

Many threads:
```
real        3.588       0.067       3.491       3.575       3.692
user        5.950       0.192       5.645       5.907       6.242
sys         11.318      0.472       10.442      11.360      12.003
```

## Mutex guarded FxHashMap (b173ee37c9487defa9e273eed291052f8ab57c39)

As a sanity check.

One thread:
```
real        3.577       0.025       3.531       3.578       3.621
user        3.050       0.032       3.013       3.041       3.107
sys         0.847       0.027       0.803       0.844       0.886
```

Many threads:
```
real        3.571       0.045       3.510       3.569       3.656
user        5.035       0.113       4.859       5.005       5.242
sys         10.704      0.585       9.816       10.621      12.142
```

ltratt:
Care to draw a conclusion rather than dump data on me? :p

vext01:
It looks to me like using a simple `RwLock` *may* perform slightly (but not much) better than the other approaches. This does surprise me though.

You'd have thought (e.g.) dashmap would be quicker, as it puts locks on the individual hashmap buckets.

ltratt:
OK, I think it's time for us to move on. In a sense, the test suite is a worse case: real programs are unlikely to contend as much as the test suite. So I think this should be bors'd as is?

vext01:
I think so, yeah :)

Issue #139, opened 2020-11-03
Constant representations.


vext01:
Constants should be a bag of bytes and a type tag. This would allow easier casting and offsetting of constant values.

vext01:
To expand on this, SIR constants are currently a big enum:
```
#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
pub enum Constant {
    Int(ConstantInt),
    Bool(bool),
    Tuple(TypeId),
    Unimplemented(String),
}
```

This works, but is not great in a couple of ways:

 * Every constant is as big as the largest possible constant due to the way Rust enums work.
 * We are required to add a new variant and implement operations on it for every different kind of constant we want to express.

To hammer home the second point: SIR is byte-offset centric. A common operation (once we have optimised MIR and SIR) will be resolving a byte offset into a constant (e.g. a projection into a constant struct or array). Implementing such an operation for all possible types will be laborious since it will differ for each variant in our enum (we are forced to consider the sizes of the fields inside each type).

What rustc does with MIR is much better. Effectively a constant is a handful of bytes and a type tag telling us how to interpret the bytes.

If we adopted this way of doing things, then constants can be truly variable-sized and we don't have this ever-growing enum. Something like:

```
struct Constant {
   bytes: &[u8],
   tyid: TypeId,
}
```

Also finding a byte offset would be the same for every constant regardless of type: `bytes[off]`.

[We could even consider de-duplicating the `bytes` field for further efficiency]

ltratt:
I'm not sure that:

```rust
struct Constant {
   bytes: &[u8],
   tyid: TypeId,
}
```

is ideal as it suggests we have a separate block of bytes for the constant, which might not be great with caching. That said, especially for non-int constants, we can probably put them in a pool, as we'll often be able to reuse them. But perhaps we can do even better without doing anything too complicated?

First, a dumb question: what is the maximum size of a constant? If it's a machine word we don't need to worry about much. If it's more than 2 machine words, we need to worry.

Second, a dumb thought: maybe we can split constants in two (or more). For word-sized constants we could have something like:

```rust
struct WordSizedConstant {
    bytes: usize,
    tyid: TypeId
}
```

where in most cases we can simply copy the word of bytes without being very clever. And then we could have a second case:

```rust
struct BiggerOrSmallerThanWordSizedConstant {
    bytes: &[u8],
    tyid: TypeId
}
```

[...]

PR #176, opened 2020-12-10
Implement call terminators in SIR interpreter.

ptersilie:
This PR implements calls in the SIR interpreter. This requires creating and switching to other stack frames, while copying over any arguments from the previous to the new stack frame.

This PR moves around a lot of code, as most of the interpreters functionality has moved into a new struct `StackFrame`. Because of this, the combined diff isn't very useful, so I recommend looking at the commits separately.

### DIFF
--- a/ykbh/src/lib.rs
+++ b/ykbh/src/lib.rs
@@ -193,7 +193,9 @@ impl SIRInterpreter {
             return;
         }
 
-        loop {
+        let mut bodies = vec![body];
+        let mut returns = Vec::new();
+        while let Some(body) = bodies.last() {
             let bbidx = usize::try_from(self.bbidx).unwrap();
             let block = &body.blocks[bbidx];
             for stmt in block.stmts.iter() {
@@ -227,23 +229,31 @@ impl SIRInterpreter {
                     frame.copy_args(args, self.frame());
                     self.frames.push(frame);
                     self.bbidx = 0;
-
-                    // Interpret the new frame.
-                    self.interpret(body);
-                    // Get pointer to result from current frame.
-                    let ptr = self.frame().local_ptr(&Local(0));
-                    // Restore previous stack frame, but keep the other frame around so the pointer to
-                    // the return value stays valid until we've copied it.
-                    let _oldframe = self.frames.pop().unwrap();
-                    // Write results to destination.
-                    if let Some((dest, bbidx)) = dest {
-                        let dst_ptr = self.frame().iplace_to_ptr(dest);
-                        let size = usize::try_from(SIR.ty(&dest.ty()).size()).unwrap();
-                        self.frame_mut().write_val(dst_ptr, ptr, size);
-                        self.bbidx = *bbidx;
+                    returns.push(dest.as_ref().map(|(p, b)| (p.clone(), *b)));
+                    bodies.push(body);
+                }
+                Terminator::Return => {
+                    // Are we returning from a call?
+                    if let Some(v) = returns.pop() {
+                        // Restore the previous stack frame, but keep the other frame around so we
+                        // can copy over the return value to the destination.
+                        let oldframe = self.frames.pop().unwrap();
+                        if let Some((dest, bbidx)) = v {
+                            // Get a pointer to the return value of the called frame.
+                            let ret_ptr = oldframe.local_ptr(&Local(0));
+                            // Write the return value to the destination in the previous frame.
+                            let dst_ptr = self.frame().iplace_to_ptr(&dest);
+                            let size = usize::try_from(SIR.ty(&dest.ty()).size()).unwrap();
+                            self.frame_mut().write_val(dst_ptr, ret_ptr, size);
+                            self.bbidx = bbidx;
+                        }
+                        // Restore previous body.
+                        bodies.pop();
+                    } else {
+                        // We are returning from the first body, so we are done interpreting.
+                        break;
                     }
                 }
-                Terminator::Return => break,
                 t => todo!("{}", t),
             }
         }
--- a/ykbh/src/lib.rs
+++ b/ykbh/src/lib.rs
@@ -1,12 +1,13 @@
 use std::alloc::{alloc, dealloc, Layout};
 use std::convert::{TryFrom, TryInto};
+use std::sync::Arc;
 use ykpack::{
-    self, BodyFlags, Constant, ConstantInt, IPlace, Local, LocalDecl, Statement, Terminator,
-    TypeId, UnsignedInt,
+    self, BodyFlags, CallOperand, Constant, ConstantInt, IPlace, Local, LocalDecl, Statement,
+    Terminator, TypeId, UnsignedInt,
 };
 use yktrace::sir::SIR;
 
-pub struct SIRInterpreter {
+pub struct StackFrame {
     /// Pointer to allocated memory containing a frame's locals.
     locals: *mut u8,
     /// The offset of each Local into locals.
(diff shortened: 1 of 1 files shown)


ykbh/src/lib.rs
```diff
@@ -117,70 +121,154 @@ impl SIRInterpreter {
...
+                    let fname = if let CallOperand::Fn(sym) = op {
+                        sym
+                    } else {
+                        todo!("unknown call target");
+                    };
+
+                    // Initialise the new stack frame.
+                    let body = SIR.body(fname).unwrap();
+                    let mut frame = SIRInterpreter::allocate_locals(&body.local_decls);
+                    frame.copy_args(args, self.frame()); // Copy arguments to new frame.
```
ltratt, line 224: Redundant comment?

ptersilie: Only if you think it's already clear from the code.

ltratt: `frame.copy_args(...)` is fairly clear to me :)

ptersilie: I just wanted to be clear that we are copying from the old frame to the new frame. But will fix.

ltratt: Descriptive method names are always good. You could say `copy_args_in` or (more accurately?) `move_args_in` or similar?

ykbh/src/lib.rs
```diff
@@ -15,8 +15,27 @@ pub struct SIRInterpreter {
...
 
+impl StackFrame {
+    fn free(&mut self) {
+        unsafe { dealloc(self.locals, self.layout) }
+    }
+}
+
+pub struct SIRInterpreter {
+    frames: Vec<StackFrame>,
+    bbidx: usize,
```
vext01, line 128: Could that be `ykpack::BasicBlockIndex`? Maybe an underscore too? `bb_idx`, or just `bb` even?

ykbh/src/lib.rs
```diff
@@ -58,34 +80,62 @@ impl SIRInterpreter {
...
-                Statement::MkRef(dest, src) => self.mkref(dest, src),
-                Statement::DynOffs { .. } => todo!(),
-                Statement::Store(dest, src) => self.store(dest, src),
-                Statement::BinaryOp { .. } => todo!(),
-                Statement::Nop => {}
-                Statement::Unimplemented(_) | Statement::Debug(_) => todo!(),
-                Statement::Cast(..) => todo!(),
-                Statement::Call(..) | Statement::StorageDead(_) => unreachable!(),
+        loop {
+            for stmt in body.blocks[self.bbidx].stmts.iter() {
```
vext01, line 196: The bbidx won't change in this loop, right? I wonder if rust will be smart enough to lift the `body.blocks[self.bbidx]` outside the loop? Should we do it manually?

ykbh/src/lib.rs
```diff
@@ -58,34 +80,62 @@ impl SIRInterpreter {
...
+                    // Restore previous stack frame, but keep the other frame around so the pointer to
+                    // the return value stays valid until we've copied it.
+                    let mut oldframe = self.frames.pop().unwrap();
+                    // Write results to destination.
+                    if let Some((dest, bbidx)) = dest {
+                        self.write(dest, ptr);
+                        self.bbidx = usize::try_from(*bbidx).unwrap();
+                    }
+                    // We've copied the value, so the old frame can be freed.
+                    oldframe.free();
```
vext01: Could frame have a `Drop` so we don't have to do this explicitly?

ykbh/src/lib.rs
bjorn3: LLVM should be smart enough to apply LICM **if** optimizations were enabled. Because they aren't, I think it won't get hoisted.

ykbh/src/lib.rs
```diff
@@ -16,9 +18,109 @@ pub struct StackFrame {
...
     }
+
+    /// Given a pointer `src` and a size, write its value to the pointer `dst`.
+    pub fn write_val(&mut self, dst: *mut u8, src: *const u8, size: usize) {
+        unsafe {
+            std::ptr::copy(src, dst, size);
+        }
+    }
+    /// Write a constant to the pointer `dst`.
+    pub fn write_const(&mut self, dest: *mut u8, constant: &Constant) {
```
vext01, line 33: Changing the way we represent consts has now moved up my priorities.

With a bag of bytes and a type tag, we'd never need to destructure like this: just copy the bytes.

 We'd probably get a little speedup for it.

See #139

ykbh/src/lib.rs
ptersilie: I thought about this too, but wasn't sure whether Rust would drop `oldframe` before we've written the result to the destination. So I went the safe route.

ykbh/src/lib.rs
vext01: I agree that the comment is redundant.

ykbh/src/lib.rs
ptersilie: Yeah, I'm not a fan of the current representation either.

ykbh/src/lib.rs
vext01: Good point!

vext01:
This is looking pretty good!

A few comments, but mostly superficial.

ykbh/src/lib.rs
ptersilie: Fixed in d6d087a.

ykbh/src/lib.rs
ptersilie: Fixed in 0d5cec2.

ykbh/src/lib.rs
ptersilie: Fixed in 8e04201.

ptersilie:
I think I've addressed all comments (either via replies or commits).

ykbh/src/lib.rs
ltratt: I missed this: manual `free` is an accident waiting to happen. Rust has very predictable (and documented!) semantics for when things are dropped: assuming that works here (which I suspect it does), a `Drop` is far preferable.

ptersilie: Okay, I didn't know the semantics were predictable. I assumed Rust would just drop the value anytime after it's not used anymore, which could be immediately after its last use (which would break this code), or at the end of the block (which would be fine). Let me double check the semantics and then I'll fix it.

ptersilie: I've read up on drops, and you're right that `oldframe` won't be dropped until we've reached the end of the scope. So this is fixed in 0f7c951.

ptersilie:
Addressed the drop vs. manual free situation as well. Ready for re-review.

ltratt:
If @vext01 is OK with this, then I think it's ready for squashing.

vext01:
LGTM. Please squash.

ptersilie:
Squashed and rebased. Unfortunately the rebase required another fix. So this needs squashing again.

ykbh/src/lib.rs
```diff
@@ -187,7 +188,7 @@ impl SIRInterpreter {
         }
     }
 
-    pub unsafe fn interpret(&mut self, body: ykpack::Body) {
+    pub unsafe fn interpret(&mut self, body: Arc<ykpack::Body>) {
```
ltratt, line 191: It *might* be better to have this be `&ykpack::Body`? That might not work for other reasons though.

vext01: Does `&*` at the call site work lukas?

ptersilie: Unfortunately not, since I also need to hold on to those references for a while in a vector: https://github.com/ptersilie/yk/blob/sinterp_calls/ykbh/src/lib.rs#L234

ltratt: OK, maybe not for this PR, but for soon: it feels a bit weird that `interpret` is storing bodies (or, at least, now `Arc<Body>`, which might not actually be so bad). We might want to have a deeper think about that.

ltratt:
Please squash.

ykbh/src/lib.rs
ptersilie: > Does `&*` at the call site work lukas?

This needs to be `body: ykpack::Body`. It can't be a reference. Not sure if I could just move the `body` out like it did before.

vext01: Gah.

That `Arc` is annoying me too. I have a branch which uses a `SyncOnceCell` instead of an `Arc`, but I'm still evaluating it...

ptersilie: > OK, maybe not for this PR, but for soon: it feels a bit weird that `interpret` is storing bodies (or, at least, now `Arc<Body>`, which might not actually be so bad). We might want to have a deeper think about that.

I need to store the bodies to avoid having a recursion in `interpret` when interpreting nested calls. See https://github.com/softdevteam/yk/pull/176/commits/016af9814821733a11030eb15358c17392ea9720

ptersilie: Alternatively, I could store the function name, and then pull body back out of SIR when returning to an older frame.

ltratt: I'm not disputing the usecase, just the mechanism by which we code it up. Anyway, it's not for this PR.

ptersilie:
Squashed!

vext01:
It's just rustfmt.

ptersilie:
Funny it choked on this one, but not a missing newline that I found earlier.

ptersilie:
Try this.

PR #178, opened 2020-12-11
Drop the `test_` prefix on test names.

vext01:
I've started reviewing the test suite, and I'll start with something very superficial: the `test_` prefix on our test function names.

This is something we've inherited from our time with pytest. It's not necessary in Rust, and it's obvious what's a test due to `#[test]`.


ltratt:
Agreed!

bors r+

PR #179, opened 2020-12-11
Location simplifications and exposition

ltratt:
This PR does not fix the two big bugs I know of in `Location` *but* it does a) document what's going on more clearly b) make the code easier to understand. This should help a subsequent PR where we try to fix the actual bugs. Each commit herein is self-contained and should, hopefully, be understandable in isolation.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -299,7 +299,7 @@ impl MTThread {
             let lp = loc.state.load(Ordering::Acquire);
             match lp & PHASE_TAG {
                 PHASE_COUNTING => {
-                    let count = (lp & !PHASE_TAG) >> 2;
+                    let count = (lp & !PHASE_TAG) >> PHASE_NUM_BITS;
                     if count >= self.inner.hot_threshold {
                         if self.inner.tracer.is_some() {
                             // This thread is already tracing. Note that we don't increment the hot
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -379,8 +379,11 @@ impl MTThread {
                             let ptr: *mut CompiledTrace<I> = Box::into_raw(Box::new(ct));
                             let new_state = ptr as usize | PHASE_COMPILED;
                             loc.state.store(new_state, Ordering::Release);
-                            // Go around the loop so that the PHASE_COMPILED case below can extract
-                            // the function.
+
+                            let bct = unsafe { Box::from_raw(ptr) };
+                            let f = unsafe { mem::transmute::<_, fn(&mut I) -> bool>(bct.ptr()) };
+                            mem::forget(bct);
+                            return Some(f);
                         }
                         Err(TryRecvError::Empty) => {
                             // The compiling thread is still operating.
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -302,8 +302,10 @@ impl MTThread {
                     let count = (lp & !PHASE_TAG) >> PHASE_NUM_BITS;
                     if count >= self.inner.hot_threshold {
                         if self.inner.tracer.is_some() {
-                            // This thread is already tracing. Note that we don't increment the hot
-                            // count further.
+                            // This thread is already tracing another Location, so either another
+                            // thread needs to trace this Location or this thread needs to wait
+                            // until the current round of tracing has completed. Either way,
+                            // there's no point incrementing the hot count.
                             return None;
                         }
                         let new_state = self.inner.tid as usize | PHASE_TRACING;
@@ -312,16 +314,21 @@ impl MTThread {
                                 Some((start_tracing(self.inner.tracing_kind), self.inner.tid));
                             return None;
                         }
-                    // We raced with another thread that's (probably) trying to trace this
-                    // Location or (less likely) has already compiled it so we go around the
-                    // loop again to see what we should do.
+                        // We raced with another thread, which probably moved the Location from
+                        // PHASE_COUNTING to PHASE_TRACING. There's a small chance that the other
+                        // thread managed to both trace and compile the Location during the race,
+                        // but that's unlikely enough that it's not worth paying the performance
+                        // penalty of going around the loop again.
+                        return None;
                     } else {
                         let new_state = PHASE_COUNTING | ((count + 1) << PHASE_NUM_BITS);
                         if loc.state.compare_and_swap(lp, new_state, Ordering::Release) == lp {
                             return None;
                         }
-                        // We raced with another thread, but we'd still like to try incrementing
-                        // the count if possible, so go around the loop again.
+                        // We raced with another thread while incrementing the count. It's worth us
+                        // trying to go around the loop again, as unless the count exceeded the hot
+                        // threshold, we'll still be able to increment it, ensuring that we don't
+                        // get weird non-determinism in when a Location is compiled.
                     }
                 }
                 PHASE_TRACING => {
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -293,12 +293,13 @@ impl MTThread {
         // that another thread interrupted us part way through. We therefore use compare_and_swap
         // to update values, allowing us to detect if we were interrupted. If we were interrupted,
         // we simply retry the whole operation.
-        loop {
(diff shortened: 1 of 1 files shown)


ltratt:
Sausages, I forgot there had been a merge while I was working on this: it's actually an annoying one. I'm going to do a rebase now otherwise reviewing this is pointless.

ltratt:
Apologies for the force push: I will repeat to myself 1000 times "remember to rebase against upstream/master before raising a PR"!

This is now ready for review.

ykrt/src/mt.rs
```diff
@@ -312,16 +314,21 @@ impl MTThread {
...
                         let new_state = PHASE_COUNTING | ((count + 1) << PHASE_NUM_BITS);
                         if loc.state.compare_and_swap(lp, new_state, Ordering::Release) == lp {
                             return None;
                         }
-                        // We raced with another thread, but we'd still like to try incrementing
-                        // the count if possible, so go around the loop again.
+                        // We raced with another thread while incrementing the count. It's worth us
+                        // trying to go around the loop again, as unless the count exceeded the hot
+                        // threshold, we'll still be able to increment it, ensuring that we don't
+                        // get weird non-determinism in when a Location is compiled.
```
ptersilie: ~~in~~

ykrt/src/mt.rs
```diff
@@ -379,8 +379,11 @@ impl MTThread {
                             let ptr: *mut CompiledTrace<I> = Box::into_raw(Box::new(ct));
                             let new_state = ptr as usize | PHASE_COMPILED;
                             loc.state.store(new_state, Ordering::Release);
-                            // Go around the loop so that the PHASE_COMPILED case below can extract
-                            // the function.
+
+                            let bct = unsafe { Box::from_raw(ptr) };
+                            let f = unsafe { mem::transmute::<_, fn(&mut I) -> bool>(bct.ptr()) };
+                            mem::forget(bct);
+                            return Some(f);
```
ptersilie: If we return here, can we even still reach `PHASE_COMPILED` below?

ykrt/src/mt.rs
ltratt: Agreed -- though this gets changed again in a follow-up commit so I'm not sure it's worth fixing it here.

ykrt/src/mt.rs
ltratt: Not on this iteration of the loop, but the next time the `control_point` is called we will hit the `PHASE_COMPILED` state.

ykrt/src/mt.rs
vext01: I'd probably remove "weird" too.

ykrt/src/mt.rs
```diff
@@ -379,8 +379,11 @@ impl MTThread {
                             let ptr: *mut CompiledTrace<I> = Box::into_raw(Box::new(ct));
                             let new_state = ptr as usize | PHASE_COMPILED;
                             loc.state.store(new_state, Ordering::Release);
-                            // Go around the loop so that the PHASE_COMPILED case below can extract
-                            // the function.
+
+                            let bct = unsafe { Box::from_raw(ptr) };
```
vext01: Would this be better expressed as an inner function (defined above the loop) with two callsites?

ykrt/src/mt.rs
ptersilie: Yeah, just came back to say I realised that as I continued reviewing. :-)

ykrt/src/mt.rs
ltratt: I'm not sure it makes much of a difference either way to be honest. We'd save 4 lines, but add 6 lines...

ykrt/src/mt.rs
```diff
@@ -321,13 +321,18 @@ impl MTThread {
                     } else {
                         let new_state = PHASE_COUNTING | ((count + 1) << PHASE_NUM_BITS);
                         let old_lp = lp;
-                        lp = loc.state.compare_and_swap(lp, new_state, Ordering::Release);
+                        lp = loc.state.compare_and_swap(lp, new_state, Ordering::Relaxed);
                         if lp == old_lp {
                             return None;
                         }
-                    }
```
vext01: I wasn't expecting this hunk. Only changes to the memory orderings. Is it related?

ltratt: Possibly not -- but given that `PHASE_COUNTING` doesn't need to sync on anything other than the `Location`, `Release` was misleading.

ykrt/src/mt.rs
```diff
@@ -294,8 +294,9 @@ impl MTThread {
         // to update values, allowing us to detect if we were interrupted. If we were interrupted,
         // we simply retry the whole operation.
 
-        // We need Acquire ordering, as PHASE_COMPILED will need to read information written to
-        // external data as a result of the PHASE_TRACING -> PHASE_COMPILED transition.
+        // We need Acquire ordering, as PHASE_COMPILING and PHASE_COMPILED need to read information
+        // written to external data. Alternatively, this load could be Relaxed but we would then
+        // need to place an Acquire fence in PHASE_COMPILING and PHASE_COMPILED.
```
vext01, line 299: I realised that I'd never looked in to how memory fences work.

I'm not sure about this, but it looks to me like they'd perform worse, because they cause *all* memory operations to synchronise, whereas an atomic value only synchronises WRT one specific address?

Looking at the intel manual:

image


Here "weakly ordered memory types" are stuff like acquire and release, I think?

vext01: Wait, what do you mean by "Acquire fence" actually?

ykrt/src/mt.rs
vext01: OK

ykrt/src/mt.rs
vext01: I ask not because I want to save lines, but because then we then have only one place to make concurrency mistakes. I can imagine updating the code in one place, and not in the other -- that's totally the kind of mistake I'd make.

ykrt/src/mt.rs
bjorn3: `SeqCst` atomic operations synchronize with all other memory operations I think. `Acquire` and `Release` atomic operations synchronize with some other memory operations. `Relaxed` atomic operations only synchronize with other atomic operations on the same memory location.

As far as I know fences upgrade other atomic operations to a less weak memory ordering.

According to https://doc.rust-lang.org/stable/std/sync/atomic/fn.fence.html:

> Depending on the specified order, a fence prevents the compiler and CPU from reordering certain types of memory operations around it. That creates synchronizes-with relationships between it and atomic operations or fences in other threads.
>
> A fence 'A' which has (at least) Release ordering semantics, synchronizes with a fence 'B' with (at least) Acquire semantics, if and only if there exist operations X and Y, both operating on some atomic object 'M' such that A is sequenced before X, Y is synchronized before B and Y observes the change to M. This provides a happens-before dependence between A and B.

vext01: > As far as I know fences upgrade other atomic operations to a less weak memory ordering.

Right, so it synchronises *more*, which may be slower, right?

But I'm wondering if @ltratt meant an actual memory fence, or he just used the term "fence" casually when talking about a regular atomic op.

bjorn3: In this case the load would be relaxed, but when the state turned out to be `PHASE_COMPILING` or `PHASE_COMPILED` there would be a fence that as far as I know upgrades the previous store and **only** the previous store to an acquire load.

ltratt: > In this case the load would be relaxed, but when the state turned out to be PHASE_COMPILING or PHASE_COMPILED there would be a fence that as far as I know upgrades the previous store and only the previous store to an acquire load.

Exactly. [I expect this would be slower, overall, but until we benchmark it we can't know for sure.]

vext01: Agreed.

ykrt/src/mt.rs
ltratt: I have tried breaking some of the code (including `type`s out), but it's tended to slightly reduce (or, at least, not improve) readability. I will probably try again later, but perhaps after we've gone through some more PRs, when commonalities in code might become more apparent.

ltratt:
I *think* we've decided not to change anything? If so, this can be merged?

PR #181, opened 2020-12-14
Stop hardware tracing if we haven't closed the loop by the end of a thread

ltratt:
If we start tracing on the last iteration of an interpreter in a given thread, the overall process seems to end up being stuck in some sense (or, at least, perhaps we just can't start tracing in another thread?). This commit explicitly stops the tracer if the thread has completed, which seems to solve the problem.


ltratt:
These should be mergable in any order.

PR #182, opened 2020-12-14
Fix threading bugs

ltratt:
After a minor renaming (https://github.com/softdevteam/yk/commit/addcf1b7aaa9f1dc1e1eccf7d740f7934adb980c), this PR fixes the two major threading bugs I know of (https://github.com/softdevteam/yk/pull/182/commits/4adc5c8d877bad9b2c97f998d31b3dde9e0010d6).

Previously there were two flaws in the Location statemachine, one fundamental, and one based on not reading the Rust docs correctly.

The fundamental flaw relates to the fact that when we move from PHASE_COUNTING to PHASE_COMPILING we have to allocate temporary storage to store the compiled trace in, which PHASE_COMPILING must free when it moves to PHASE_COMPILED. Previously there was a race condition where two threads could observe PHASE_COMPILING as having completed and both would try freeing the temporary storage leading to either a double free being detected or a segfault (I observed both happening!). This commit fixes that by introducing a new state into the state machine PHASE_LOCKED: when we're in PHASE_COMPILING and we want to see if the trace has been compiled, we temporarily flip the state to PHASE_LOCKED, at which point other threads know to leave the Location alone. If the trace has been compiled, we can free the temporary storage without worrying that we're racing to another thread; if it hasn't been compiled we go back to PHASE_COMPILING. The statemachine thus now looks as follows:

```
               ┌────────────────────────────────────────┐
               │                                        │ ─────────────┐
   reprofile   │             PHASE_COUNTING             │              │
  ┌──────────▶ │                                        │ ◀────────────┘
  │            └────────────────────────────────────────┘    increment
  │              │                                 ▲         count
  │              │ start tracing                   │
  │              ▼                                 │
  │            ┌────────────────────┐              │ abort
  │      ┌─────│                    │              |
  │      |     │   PHASE_TRACING    │ ─────────────┤
  │      └────▶│                    │              |
  │            └────────────────────┘              │
  │              │ start compiling trace           │
  │              │ in thread                       │
  │              ▼                                 │
  │            ┌────────────────────┐              │
  │            │                    │              |
  |            |  PHASE_COMPILING   | ─────────────┤
  │            │                    │              |
  │            └────────────────────┘              |
  │              │ trace maybe    ▲                |
  │              │ compiled       | trace not yet  |
  │              ▼                | compiled       |
  │            ┌────────────────────┐              |
  │            │    PHASE_LOCKED    │ ─────────────┘
  │            └────────────────────┘
  │              │ trace definitely
  │              │ compiled
  │              ▼
  │            ┌────────────────────┐
  │            │                    │ ──────┐
  │            │   PHASE_COMPILED   │       │
  └─────────── │                    │ ◀─────┘
               └────────────────────┘
```

The second flaw relates to the fact that we used Rust channels to pass a compiled trace from the compiling thread to an interpreter thread, meaning that multiple threads could simultaneously try reading from the channel. However, while the Receiver type is Send, it is not Sync, so this was not safe: we didn't notice this because we converted the Reciever into a box and passed a tagged pointer around, bypassing Rust's type system. This commit fixes this problem by using an explicit Mutex, which is safe to pass between threads via
our tagged pointers.

The new test "simple_multithreaded_interpreter" starts multiple threads on a small number of locations, with different threads non-deterministically moving amongst the Location statemachine. Because it's inherently non-deterministic, on any given run it doesn't necessarily trigger the problems fixed above, but if run often enough, it is capable of hitting both. We thus run it several times with the aim that neither of the above problems can reappear in the future.

Note that while this commit aims to solve crashing bugs, it now explicitly shows that we have two potential memory leaks, both related to the fact that we can't guarantee that any given Location will ever execute again. At some point we will need to add some sort of timeout that frees traces that haven't recently been executed. We also don't deal with the situation where a loop starts tracing but doesn't actually loop: the Location in question currently ends up stuck in PHASE_TRACING forever. Fixing these things is for the future.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -569,9 +569,9 @@ mod tests {
         let mut mtt = MTBuilder::new().hot_threshold(2).init();
 
         // The program is silly. Do nothing twice, then start again.
-        const NOP: u8 = 0;
+        const INC: u8 = 0;
         const RESTART: u8 = 1;
-        let prog = vec![NOP, NOP, RESTART];
+        let prog = vec![INC, INC, RESTART];
 
         // Suppose the bytecode compiler for this imaginary language knows that the first bytecode
         // is the only place a loop can start.
@@ -586,7 +586,7 @@ mod tests {
         #[interp_step]
         fn dumb_interp_step(tio: &mut IO) {
             match tio.prog[tio.pc] {
-                NOP => {
+                INC => {
                     tio.pc += 1;
                     tio.count += 1;
                 }
@@ -625,7 +625,7 @@ mod tests {
 
         assert_eq!(tio.pc, 0);
 
-        // A trace was just compiled. Running it should execute NOP twice.
+        // A trace was just compiled. Running it should execute INC twice.
         tio.count = 8;
         for i in 0..10 {
             let loc = locs[tio.pc].as_ref();
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -9,8 +9,7 @@ use std::{
     rc::Rc,
     sync::{
         atomic::{AtomicBool, AtomicUsize, Ordering},
-        mpsc::{channel, Receiver, TryRecvError},
-        Arc,
+        Arc, Mutex, TryLockError,
     },
     thread::{self, JoinHandle},
 };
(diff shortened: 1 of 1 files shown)


ykrt/src/mt.rs
```diff
@@ -379,31 +375,46 @@ impl MTThread {
...
-                        // FIXME: free up the memory when the trace is no longer used.
-                        let ptr: *mut CompiledTrace<I> = Box::into_raw(Box::new(ct));
-                        let new_state = ptr as usize | PHASE_COMPILED;
-                        loc.state.store(new_state, Ordering::Release);
-
-                        let bct = unsafe { Box::from_raw(ptr) };
-                        let f = unsafe { mem::transmute::<_, fn(&mut I) -> bool>(bct.ptr()) };
-                        mem::forget(bct);
-                        return Some(f);
+                // We need to free the memory allocated earlier. To ensure we don't race with
```
vext01, line 378: I'm probably being daft, but what memory are we freeing again?

ltratt: `let mtx = Arc::new(Mutex::new(None));` (line 359)

vext01: So it's just the `Arc` wrapper.

Does more than one thread actually need access to what's inside? Could we move the buffer instead of sharing it? Then we wouldn't need another phase?

vext01: Or is this one of the cases where we can't get Rust to believe us that something lives long enough?

ltratt: `Arc` allocates memory underneath, so it's more than just a wrapper. But "yes".

> Does more than one thread actually need access to what's inside?

Yes, one trace can compile location *L* and another thread might observe that *L* has finished compiling and can be used.

ykrt/src/mt.rs
```diff
@@ -94,17 +99,6 @@ impl<I> Location<I> {
     }
 }
 
-impl<I> Drop for Location<I> {
-    fn drop(&mut self) {
-        let lp = self.state.load(Ordering::Relaxed);
-        if lp & PHASE_TAG == PHASE_COMPILING {
-            unsafe {
-                Box::from_raw((lp & !PHASE_TAG) as *mut CompilingTrace<I>);
```
bjorn3, line 102: This is running on the only thread having access to the `Location` so there can be no race-condition. You could even use `self.state.get_mut()` instead of an atomic load.

ltratt: This hunk has been deleted, but, yes, if it's brought back, I agree.

ykrt/src/mt.rs
vext01: Ug. This is hard.

I think I understand. It's a shame we need another phase, but I can't see any other way to do it.

ykrt/src/mt.rs
bjorn3: The removal causes a memory leak, right? Why was it removed? There couldn't be a race-condition.

ltratt: It does indeed (there's a comment on it earlier in the PR). However, there are several ways we leak at the moment, and I need a more principled way to handle them in one go.

vext01:
Ready to go?

ltratt:
If you're happy, I am!

ltratt:
bors was stuck, possibly because is a way that the test might not ever complete in this PR and, at the moment, the best we can do is have a timeout to handle it. It's a bit of a gross hack but https://github.com/softdevteam/yk/pull/182/commits/d6f5bd11c9819503a9b2b3ad3f2741135e8a7660 will at least address that case (assuming that was the cause for bors to get stuck -- there might be others!).

bors try

ltratt:
bors try

vext01:
Is this a deadlock?
```
test mt::tests::simple_multithreaded_interpreter ... test mt::tests::simple_multithreaded_interpreter has been running for over 60 seconds
```

ltratt:
I think it's a livelock, but *why* it's happening, I'm not sure -- it doesn't happen under my normal user on b13! I'll investigate later.

ltratt:
OK, I can observe this happening occasionally under my normal account. For some reason we end up perpetually stuck in `PHASE_COMPILING`. Investigating.

ltratt:
I'm an idiot -- the problem is solely in the test.

The cause is that the compiling of the trace can outlast all the "normal" threads in the test, hence the `Location` was never updated. Thus "all" we need to do is make sure the main thread still calls `control_point`. With https://github.com/softdevteam/yk/pull/182/commits/665a620c3660dc0fb372099fc0290f3f3dca54f1, I can run the test 100s of times without failure (before it would rarely make it 10 times without failing).

ltratt:
bors try

vext01:
OK. Please squash.

ltratt:
Squashed.

vext01:
Bors r+

PR #183, opened 2020-12-14
Use layout and offsets computed at compile time.

ptersilie:
The memory layout and offsets for the locals are now computed at compile
time and stored in the SIR headers. This increases performance when
creating new frames in the SIR interpreter and makes our life easier
during blackholing later on.

### DIFF
--- a/ykbh/src/lib.rs
+++ b/ykbh/src/lib.rs
@@ -1,8 +1,8 @@
 use std::alloc::{alloc, dealloc, Layout};
-use std::convert::{TryFrom, TryInto};
+use std::convert::TryFrom;
 use std::sync::Arc;
 use ykpack::{
-    self, BodyFlags, CallOperand, Constant, ConstantInt, IPlace, Local, LocalDecl, Statement,
+    self, Body, BodyFlags, CallOperand, Constant, ConstantInt, IPlace, Local, Statement,
     Terminator, UnsignedInt,
 };
 use yktrace::sir::SIR;
@@ -132,8 +132,8 @@ pub struct SIRInterpreter {
 }
 
 impl SIRInterpreter {
-    pub fn new(local_decls: &Vec<LocalDecl>) -> Self {
-        let frame = SIRInterpreter::allocate_locals(local_decls);
+    pub fn new(body: Arc<Body>) -> Self {
+        let frame = SIRInterpreter::create_frame(body);
         SIRInterpreter {
             frames: vec![frame],
             bbidx: 0,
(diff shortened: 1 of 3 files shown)


ykbh/src/lib.rs
```diff
@@ -142,21 +142,10 @@ impl SIRInterpreter {
...
-            let size = SIR.ty(&d.ty).size();
-            let l = Layout::from_size_align(size.try_into().unwrap(), align.try_into().unwrap())
-                .unwrap();
-            let (nl, s) = layout.extend(l).unwrap();
-            offsets.push(s);
-            layout = nl;
-        }
-        layout = layout.pad_to_align();
-
+    fn create_frame(body: Arc<Body>) -> StackFrame {
```
vext01, line 145: This is optional, but I think you could make this more general bu getting rid of the `Arc` by making this accept `&body` and then doing `&*` at the call-site?

vext01:
And one small mistake in the commit message:

> stored in the SIR headers

Looks like the layout isn't in the headers, but the data itself?

Otherwise LGTM

ykbh/src/lib.rs
ptersilie: Wouldn't that move the `Body`? I might have had a reason why I did it this way in the last PR, though I can't remember.

ptersilie:
> Looks like the layout isn't in the headers, but the data itself?

We store `size` and `align` in the headers, which I guess describes the layout, but isn't the layout itself.

ykbh/src/lib.rs
vext01: I'm not sure actually. Might be worth a shot?

vext01:
> We store size and align in the headers

Huh? But you didn't change the SIR headers:
https://github.com/ptersilie/yk/blob/985130f7dedc92f66bbae3fff0b679019a83f575/ykpack/src/types.rs#L1058-L1070

You added the layout to the `Body`, right?
https://github.com/softdevteam/yk/pull/183/files#diff-d36972674d06cb36d898cc61a18cfc98abae26114b0ff69887844507cae5a211R387

ptersilie:
Yeah, but aren't the bodies stored in the headers? I might be misusing some terminology here. :-)

ykbh/src/lib.rs
ptersilie: Better? 4b3c667

vext01:
The headers tell you at which byte offset the bodies can be found, but don't store the bodies themselves.

ykbh/src/lib.rs
```diff
@@ -132,7 +132,7 @@ pub struct SIRInterpreter {
 }
 
 impl SIRInterpreter {
-    pub fn new(body: Arc<Body>) -> Self {
+    pub fn new(body: &Body) -> Self {
```
vext01, line 135: Oh cool, it worked :)

ykbh/src/lib.rs
vext01: Great!

vext01:
Please squash (and update commit message).

ptersilie:
No need to squash as the last commit isn't really related to this PR so should stand on its own.

vext01:
OK, so we just need to edit the first commit message:

> The memory layout and offsets for the locals are now computed at compile time and stored in the SIR headers.

Just remove the word "headers" and we are grand.

ptersilie:
Okay to rebase?

vext01:
Yep

ptersilie:
Rebased.

PR #184, opened 2020-12-14
Review Register Allocation Tests

vext01:
See individual commits.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -47,6 +47,10 @@ lazy_static! {
                                      R10.code(), RBX.code(), R12.code(), R13.code(), R14.code(),
                                      R15.code()];
 
+    // The trace inputs/outputs are always allocated to this reserved register.
+    // This register should not appear in REG_POOL.
+    static ref TIO_REG: u8 = RDI.code();
+
     static ref TEMP_LOC: Location = Location::Reg(*TEMP_REG);
     static ref PTR_SIZE: u64 = u64::try_from(mem::size_of::<usize>()).unwrap();
 }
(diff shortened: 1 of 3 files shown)


ykcompile/src/lib.rs
```diff
@@ -47,6 +47,10 @@ lazy_static! {
                                      R10.code(), RBX.code(), R12.code(), R13.code(), R14.code(),
                                      R15.code()];
 
+    // The trace inputs/output are always allocated to this reserved register.
```
vext01, line 50: output*s*

ltratt:
Overall, to the extent I understand this (it's not code I'm very familiar with!) this looks good, and I'm pleased that tests haven't disappeared! My only comment is that when it comes to squashing/rebasing time, can we provide a bit more motivation in some of the commits (so not just "what" but also "why")? For example in https://github.com/softdevteam/yk/pull/184/commits/e005a61a50b5388fce0aa9c39d950a4a3cacbe0d inlining `init` probably just needs something like "Since we only call this once, there's no advantage to having it be a separate function"; in https://github.com/softdevteam/yk/pull/184/commits/177160b1ec7a8d154ce894395480463f0cff2aa7 I presume the implication is that "We didn't previously test this"?

ptersilie:
Looks good to me, though agree with what Laurie said.

vext01:
Sure, I'll expand the commit messages.

OK to fix that typo and squash at the same time?

ltratt:
Please go ahead.

ptersilie:
Sure.

vext01:
Give this a go.

ptersilie:
Looks good to me.

PR #186, opened 2020-12-15
Store size and alignment alongside each type.

ptersilie:
We already did something similar for structs, tuples, etc. Now we are
doing the same for all types and thus no longer need to specify these
values manually.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -21,7 +21,7 @@ use std::ffi::CString;
 use std::fmt::{self, Display, Formatter};
 use std::mem;
 use std::process::Command;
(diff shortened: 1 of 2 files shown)


vext01:
I'm a big fan of this change!

There's a register allocation test with a FIXME that mentions tuple layouts (just merged today). Would you mind adding a "tiny tuple" case to it?

ptersilie:
Sure. Let me have a look.

ptersilie:
> Would you mind adding a "tiny tuple" case to it?

As discussed offline, you were referring to `enum`s, which we haven't lowered yet, so there's nothing to do here for now.

PR #187, opened 2020-12-15
Handle threads that don't finish tracing gracefully.

ltratt:
Previously, if we started tracing but didn't close the loop before the thread finished, the Location set to PHASE_TRACING would be stuck in that phase forever.

This commit changes that. When a thread comes across a PHASE_TRACING Location, it now checks to see if the associated thread is still alive. If it is not, it changes the Location to PHASE_WONT_TRACE which means we'll never try tracing the Location again. This latter phase is a temporary hack: in the longer run, we need to do "retry N times before marking as WONT_TRACE" (as RPython does). Doing that is for a future commit.

The cunning hack in this commit is how we detect if the tracing thread is alive: we just change the thread ID we use to be a pointer to an Arc (where the contents of the Arc aren't important). If that Arc's strong count is above 1, the other thread is alive, otherwise it's finished.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -1,7 +1,6 @@
 #[cfg(test)]
 use std::time::Duration;
 use std::{
-    alloc::{alloc, dealloc, Layout},
     io,
     marker::PhantomData,
     mem,
@@ -27,11 +26,11 @@ const PHASE_NUM_BITS: usize = 3;
 const PHASE_TAG: usize = 0b111; // All of the other PHASE_ tags must fit in this.
 const PHASE_COMPILED: usize = 0b000; // Value is a pointer to a chunk of memory containing a
                                      // CompiledTrace<I>.
-const PHASE_TRACING: usize = 0b001; // Value is a pointer to a malloc'd block that allows us to
-                                    // precisely identify when we have reached the same Location.
+const PHASE_TRACING: usize = 0b001; // Value is a pointer to an Arc<u8> representing a thread ID.
 const PHASE_COMPILING: usize = 0b010; // Value is a pointer to a `Box<CompilingTrace<I>>`.
 const PHASE_COUNTING: usize = 0b011; // Value specifies how many times we've seen this Location.
 const PHASE_LOCKED: usize = 0b100; // No associated value.
+const PHASE_DONT_TRACE: usize = 0b101; // No associated value.
 
 /// A `Location` stores state that the meta-tracer needs to identify hot loops and run associated
 /// machine code.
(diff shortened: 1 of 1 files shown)


ykrt/src/mt.rs
```diff
@@ -27,11 +26,11 @@ const PHASE_NUM_BITS: usize = 3;
 const PHASE_TAG: usize = 0b111; // All of the other PHASE_ tags must fit in this.
 const PHASE_COMPILED: usize = 0b000; // Value is a pointer to a chunk of memory containing a
                                      // CompiledTrace<I>.
-const PHASE_TRACING: usize = 0b001; // Value is a pointer to a malloc'd block that allows us to
-                                    // precisely identify when we have reached the same Location.
+const PHASE_TRACING: usize = 0b001; // Value is a pointer to an Arc<u8> representing a thread ID.
 const PHASE_COMPILING: usize = 0b010; // Value is a pointer to a `Box<CompilingTrace<I>>`.
 const PHASE_COUNTING: usize = 0b011; // Value specifies how many times we've seen this Location.
 const PHASE_LOCKED: usize = 0b100; // No associated value.
+const PHASE_WONT_TRACE: usize = 0b101; // No associated value.
```
vext01, line 33: I'll throw in my obligatory bikeshed comment and say: s/WONT/DONT/g ?

ltratt: I feel not strongly one way nor either the other. Though I'd have written that previous sentence "I don't care one way or t'other" myself :p

ykrt/src/mt.rs
```diff
@@ -332,17 +347,38 @@ impl MTThread {
...
                         // ...and we started at this Location, so we've got a complete loop!
                     }
                 } else {
                     // Another thread is tracing this location.
-                    return None;
+                    if Arc::strong_count(&loc_tid) == 1 {
+                        // The other thread has stopped. Mark this Location as WONT_TRACE.
+                        if loc
+                            .state
+                            .compare_and_swap(lp, PHASE_WONT_TRACE, Ordering::Acquire)
```
vext01, line 366: So this can be a regular store, if we know all racing threads will store the same value `PHASE_WONT_TRACE`?

ltratt: No, because there could be another live (but non-tracing thread) that's carrying out the same check as us, so we might race on the transition to `PHASE_WONT_TRACE`. If we did race, we'd both cause the `Arc` to be `free`d (i.e. a double free).

ykrt/src/mt.rs
```diff
@@ -429,31 +465,26 @@ impl MTThread {
 /// The innards of a meta-tracer thread.
 struct MTThreadInner {
     mt: MT,
-    /// A value that uniquely identifies a thread. Since this ID needs to be ORable with PHASE_TAG,
-    /// we use a pointer to a malloc'd chunk of memory. We guarantee a) that chunk is aligned to a
-    /// machine word b) that it is a non-zero chunk of memory (and thus guaranteed to be a unique
-    /// pointer).
-    tid: *mut u8,
+    /// An Arc whose pointer address uniquely identifies this thread. When a Location is traced,
+    /// this Arc's strong count will be incremented. If, when this thread drops, this Arc's strong
```
vext01, line 469: I'm a bit confused by this bit. When this thread drops what? Itself?

ykrt/src/mt.rs
```diff
@@ -742,6 +765,123 @@ mod tests {
...
+        const RESTART: u8 = 1;
+        let prog = Arc::new(vec![INC, INC, RESTART]);
+
+        struct IO {
+            prog: Arc<Vec<u8>>,
+            count: u64,
+            pc: usize,
+        }
+
+        #[interp_step]
```
vext01, line 784: If this `interp_step` is going to be the same in lots of tests, shall we lift it out and re-use it?

ykrt/src/mt.rs
vext01: Gotcha. Nice catch! I'd have screwed that up!

ykrt/src/mt.rs
ltratt: Would "when this thread is dropped" be clearer?

ykrt/src/mt.rs
ltratt: I agree, there's a growing amount of duplication, but I think that's best left to another PR. At the moment I'm mostly concentrating on making sure we test as much as possible, which is a big effort for my poor little brain!

ykrt/src/mt.rs
vext01: Ah I see.

So when the `MTThreadInner` is dropped, we look at the strong count of the `Arc`. But won't it always be `>=1` at that point because the `MTThreadInner` has a handle on it? Either a typo, or I misunderstand.

ltratt: When a Location goes to `PHASE_TRACING` it increments the strong count. When this thread dies it decrements the strong count. When the Location goes from `PHASE_TRACING` to something else it decrement the strong count. Those two decrements can happen in any order.

vext01: Yes, I get that bit.

Put differently, how do you inspect the strong count of an `Arc` and get a count of zero? Wouldn't that mean the `Arc` doesn't exist?

vext01: And therefore the count has to be at least 1 if you are inspecting the strong count...

I bet I'm confused.

ltratt: The comment is quite careful not to claim that it inspecs the `Arc` to determine a strong count of zero. The strong count can only go to zero (but might not!) after the thread is dropped.

vext01: Right. So, I wonder if this needs a reword, as it confused me.

ltratt: I think there's a limit to how clear this can be without becoming an essay, but https://github.com/softdevteam/yk/pull/187/commits/f07db556763e34c4d91e2a56c6871c590d4f2b27 might help a bit.

vext01: You do yourself a disservice. That small change makes a world of difference!

vext01:
If you are ready. Please squash.

ykrt/src/mt.rs
ltratt: On reflection, I agree. Fixed in https://github.com/softdevteam/yk/pull/187/commits/0ea56e8b8a2a783bf69e9e365587a4c1b1461aa6.

ltratt:
Still OK for me to rebase and squash?

vext01:
By all means.

ltratt:
Squashed.

PR #189, opened 2020-12-16
Only update the TIR compiler's return stack for inlined calls.

### LINKED ISSUE #188: Undefined variable errors in TIR compiler.
```
      #[test]                                                                                                                                                                                                       
      fn recursion() {                                                                                                                                                                                        
          struct IO(usize);                                                                                                                                                                                         
                                                                                                                                                                                                                    
          fn recurse(c: usize) -> usize {                                                                                                                                                                           
              if c == 0 {                                                                                                                                                                                           
                  42                                                                                                                                                                                                
              } else {                                                                                                                                                                                              
                  recurse(c - 1)                                                                                                                                                                                    
              }                                                                                                                                                                                                     
          }                                                                                                                                                                                                         
                                                                                                                                                                                                                    
          #[interp_step]                                                                                                                                                                                            
          fn interp_step(io: &mut IO) {                                                                                                                                                                             
              io.0 = recurse(1);                                                                                                                                                                                    
          }                                                                                                                                                                                                         
                                                                                                                                                                                                                    
          let mut inputs = IO(0);                                                                                                                                                                                   
          let th = start_tracing(TracingKind::HardwareTracing);                                                                                                                                                     
          interp_step(&mut inputs);                                                                                                                                                                                 
          let sir_trace = th.stop_tracing().unwrap();                                                                                                                                                               
    }
```

Gives:
```
thread 'tests::spilling_simple' panicked at 'undefined local: $5 in guard($5+8, bool(false), <_ZN9ykcompile5tests15spilling_simple7recurse17h07a58b090b285788E, 1>, [])', yktrace/src/tir.rs:84:21
```

vext01:
This change fixes a bug where we'd update the return stack for external
calls (for example, things marked `#[do_not_trace]`). This is wrong
because we will never see a corresponding Return terminator for the
callee, and then some poor unrelated (and unsuspecting) Return ends up
popping a garbage value off of the return stack.

The fix is to only push to the return stack if the call is to be
inlined.

Fixes #188.

### DIFF
--- a/yktrace/src/tir.rs
+++ b/yktrace/src/tir.rs
@@ -41,6 +41,10 @@ impl<'a, 'm> TirTrace<'a, 'm> {
         // Maps symbol names to their virtual addresses.
         let mut addr_map: HashMap<String, u64> = HashMap::new();
 
+        // A stack to keep track of where to store return values of inlined calls. When we
+        // encounter `$x = Call(...)` we push `$x` to the stack so that later, when we encounter
+        // the corresponding Return, we can find the correct place to store the return value (by
+        // popping from the stack).
         let mut return_iplaces: Vec<IPlace> = Vec::new();
 
         // As we compile, we are going to check the define-use (DU) chain of our local
@@ -242,7 +246,6 @@ impl<'a, 'm> TirTrace<'a, 'm> {
                         .as_ref()
                         .map(|(ret_val, _)| rnm.rename_iplace(&ret_val, &body))
                         .unwrap();
-                    return_iplaces.push(ret_val.clone());
 
                     if let Some(callee_sym) = op.symbol() {
                         // We know the symbol name of the callee at least.
@@ -257,6 +260,10 @@ impl<'a, 'm> TirTrace<'a, 'm> {
                                 ignore = Some(callee_sym.to_string());
                                 term_stmts.push(Statement::Call(op.clone(), newargs, Some(ret_val)))
                             } else {
+                                // Push the IPlace that the corresponding Return terminator should
+                                // assign the result of the call to.
+                                return_iplaces.push(ret_val.clone());
+
                                 // Inform VarRenamer about this function's offset, which is equal to the
                                 // number of variables assigned in the outer body.
                                 rnm.enter(callbody.local_decls.len());
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -3139,4 +3139,36 @@ mod tests {
         assert_eq!(cr, true);
         assert_eq!(args.cells, vec![1, 4, 3]);
     }
+
+    /// Check that calling a `do_not_trace` annotated function from within a regular (but
+    /// non-interp-step) function works.
+    #[test]
+    fn nested_do_not_trace() {
+        #[do_not_trace]
+        fn one() -> usize {
+            1
+        }
+
+        fn call_one() -> usize {
+            one()
+        }
+
+        struct IO(usize);
+
+        #[interp_step]
+        #[inline(never)]
+        fn interp_step(io: &mut IO) {
+            io.0 = call_one();
+        }
+
+        let mut inputs = IO(0);
+        let th = start_tracing(TracingKind::HardwareTracing);
+        interp_step(&mut inputs);
+        let sir_trace = th.stop_tracing().unwrap();
+        let tir_trace = TirTrace::new(&*SIR, &*sir_trace).unwrap();
+        let ct = TraceCompiler::<IO>::compile(tir_trace);
+        let mut args = IO(0);
+        assert!(ct.execute(&mut args), true);
+        assert_eq!(args.0, 1);
+    }
 }


yktrace/src/tir.rs
```diff
@@ -41,6 +41,11 @@ impl<'a, 'm> TirTrace<'a, 'm> {
         // Maps symbol names to their virtual addresses.
         let mut addr_map: HashMap<String, u64> = HashMap::new();
 
+        // We use a stack of IPlaces to keep track of storage for return values of inlined calls.
+        // When we encounter the Return terminator of a call that's being inlined into the trace,
```
ltratt, line 45: This sentence is very hard to parse (I eventually worked out it out, but it took me a few reads). I'd suggest starting with what it *does* not what it *doesn't do*. Something like "We need to convert `return`s from inlined functions into stores" maybe?

ltratt:
On a related note to my pedantic comment, when we come to merge, can we change the commit title from "Fix a nasty bug in the TIR compiler's return stack" to something like "Convert returns in inlined functions to stores"?

yktrace/src/tir.rs
vext01: Try 968e2a8, is this clearer?

ltratt: I'm not sure what "To avoid unnecessary stores" has to do with anything?

vext01: This is effectively an optimisation that eliminates one store.

Instead of:
```
callee:
  $x = return value
caller:
  $y = $x
```

You get:
```
callee:
   $y = return value
```

ptersilie: What Edd said. This is an optimisation that assigns the return value of an inlined function immediately to the destination of the function call instead of going via the return local (`$0`).

ltratt: Ah, then I misunderstood what's going on. Silly question: why are we doing this optimisation here? Wouldn't it be better to leave this to a generic optimisation pass?

bjorn3: I would guess as it is very common and is much cheaper to do in this case than the general case.

ltratt: I'm asking partly for a generic reason: the problem with "early" optimisations like these is that they tend to introduce bugs (like the one cleared up by this PR). Then we have to ask whether they're worth it, or, at least, whether they're worth doing now. I'm open-minded in this case!

vext01: Since it's implemented, I'd say leave it for now.

ltratt: Are we reasonably confident that a) there aren't other edge cases or b) that we've thought through those edge cases?

ptersilie: At the time (when we didn't have `do_not_trace` yet) it was a very simple optimisation which essentially did this:
```
$0 = $1;
$5 = $0;
```
becomes
```
$5 = $1;
```
So there wasn't too much to think about in terms of edge cases.

ptersilie:
Please squash.

vext01:
If we are not spinning this as an optimisation, I think I should update the comment describing the return stack.

ltratt:
That would be good, thanks @vext01!

vext01:
Try this. This was actually quite hard to put into words!

yktrace/src/tir.rs
```diff
@@ -41,10 +41,10 @@ impl<'a, 'm> TirTrace<'a, 'm> {
         // Maps symbol names to their virtual addresses.
         let mut addr_map: HashMap<String, u64> = HashMap::new();
 
-        // To avoid unnecessary stores, inlined function calls store their return value directly
-        // into its final resting place chosen by the caller. To make this happen, we use a stack
-        // of IPlaces to keep track of storage for our return values. The `Call` pushes to the
-        // stack, and the corresponding `Return` pops.
+        // A stack to keep track of where to store return values of inlined calls. When we
+        // encounter `$x = Call(...)` we push `$x` to the stack so that later, when we encounter
+        // the corresponding Return, we can find the correct place to store the return value to (by
```
ltratt, line 46: Is the `to` stray?

yktrace/src/tir.rs
```diff
@@ -41,10 +41,10 @@ impl<'a, 'm> TirTrace<'a, 'm> {
...
+        // popping from the stack). It has to be a stack to allow for nested inlined calls.
```
ltratt, line 47: I think this last sentence is redundant (it's implied by the first sentence)?

vext01: Fair enough.

yktrace/src/tir.rs
vext01: This one?

> the correct place to store the return value to

I almost went with `into`, would that help?

ltratt: I think we can just drop the "to" entirely? It's a bit hard to parse with the the following `(by`.

vext01:
Give that a go.

ltratt:
Please squash (and I guess you'll want to update the commit message too).

vext01:
splat.

PR #191, opened 2020-12-16
Less exports

bjorn3:
This reduces the api surface of ykcompile and yktrace. If any of the exports turn out to be necessary later on, they can be exported again.


ltratt:
I have to defer to @ptersilie and @vext01 on this one.

vext01:
Looks good. I also compile tested locally. All is well.

bors r+

PR #192, opened 2020-12-16
Less allocations

bjorn3:
This may not matter a lot, but still it should be a small perf win.

Based on #191

### DIFF
--- a/yktrace/src/hwt/mapper.rs
+++ b/yktrace/src/hwt/mapper.rs
@@ -61,7 +61,7 @@ impl HWTMapper {
                     } else {
                         None
                     };
-                    (lbl.off, SirLoc::new(lbl.symbol_name.clone(), lbl.bb, vaddr))
+                    (lbl.off, SirLoc::new(&lbl.symbol_name, lbl.bb, vaddr))
                 })
                 .collect::<Vec<_>>();
             locs.sort_by_key(|l| l.0);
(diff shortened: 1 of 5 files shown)


ltratt:
Are you able to rebase this against master so that it doesn't pull in commits from #191? If not, please feel free to leave as-is.

ltratt:
This isn't part of the code base I'm super familiar with. I guess my main question to @vext01 (or @ptersilie? but I think this is mostly Edd's work) is whether we want to leak this memory for the lifetime of a process or not. If SIR can never be "unloaded" (in some sense of that term) I guess it doesn't matter, though memory leaks do make me mildly queasy, in the sense that they at least make leak checking tools more noisy to use.

vext01:
Do the string slices have to be `'static` lifetime? Could they be `'a` so that we don't have to leak in that one case?

vext01:
>  If SIR can never be "unloaded"

At the moment it's considered eternal.

>  though memory leaks do make me mildly queasy

Same here. I'd rather not, but then again it's only in `ykview` which is just a debugging aid for dumping SIR from a binary. It's not part of the JIT proper.

ltratt:
Yeah, I just realised what's going on -- I forgot that this is referring to something already in the binary! I'm fine with this, because the `ykview` situation is reasonable.

bjorn3:
> Do the string slices have to be `'static` lifetime? Could they be `'a` so that we don't have to leak in that one case?

I tried but ran into lifetime trouble. As a next step I want to replace the `SirTrace` trait with a struct which may make it easier to do this.

bjorn3:
Rebased

PR #193, opened 2020-12-16
Replace SirTrace trait with a Vec wrapper struct

bjorn3:
In practice every implementor of the trait was actually a Vec wrapper. Making SirTrace itself a Vec wrapper reduces code duplication and simplifies some other things.

### DIFF
--- a/yktrace/src/swt.rs
+++ b/yktrace/src/swt.rs
@@ -6,59 +6,32 @@ use libc;
 use std::convert::TryFrom;
 use ykpack::Local;
 
-/// A trace collected via software tracing.
-/// Since the trace is a heap-allocated C buffer, we represent it as a pointer and a length.
-#[derive(Debug)]
-struct SWTSirTrace {
-    locs: Vec<SirLoc>
-}
-
-impl SWTSirTrace {
-    /// Create a SWTSirTrace from a raw buffer and (element) length.
-    ///
-    /// `buf` must have been allocated by malloc(3); this function will free(2) it.
-    fn from_buf(buf: *const SwtLoc, len: usize) -> Self {
-        // When we make a SWTSirTrace, we convert all of the locations from core::SirLoc up to
-        // crate::SirLoc and store them in self so that we can hand out references via raw_loc().
-        let locs = (0..len)
-            .map(|idx| {
-                let idx = isize::try_from(idx).unwrap();
-                let swt_loc = unsafe { &*buf.offset(idx) };
-                let symbol_name = unsafe { std::ffi::CStr::from_ptr(swt_loc.symbol_name) };
-                SirLoc {
-                    symbol_name: symbol_name.to_str().unwrap(),
-                    bb_idx: swt_loc.bb_idx,
-                    addr: None
-                }
-            })
-            .collect();
-
-        unsafe { libc::free(buf as *mut libc::c_void) };
-        Self { locs }
-    }
-}
-
-impl SirTrace for SWTSirTrace {
-    fn raw_len(&self) -> usize {
-        self.locs.len()
-    }
-
-    fn raw_loc(&self, idx: usize) -> &SirLoc {
-        &self.locs[idx]
-    }
-}
-
 /// Softare thread tracer.
 struct SWTThreadTracer;
 
 impl ThreadTracerImpl for SWTThreadTracer {
-    fn stop_tracing(&mut self) -> Result<Box<dyn SirTrace>, InvalidTraceError> {
+    fn stop_tracing(&mut self) -> Result<SirTrace, InvalidTraceError> {
         let mut len = 0;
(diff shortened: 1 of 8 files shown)


ltratt:
I *think* the original reason for this might have been to support both software and hardware tracing. However, since I think both would produce `SIRTrace` and not an implementation of that trait, I don't see any reason why we wouldn't want to merge this PR. @vext01 thoughts?

vext01:
The reason it was implemented this way is that in the beginning I thought I could lazily decode  and map the PT trace (we do only the former at present. The mapper always eagerly does the whole trace).

It didn't work out that way, but it may still be possible if someone put their mind to it. It's not clear to me if it'd be worth the change though...

So if we want to leave that possibility open, then we should leave it a trait, otherwise we can merge this.

ltratt:
What does "lazily decode and map" mean in this context and how would a trait help? I ask this very naively -- I have no preconceptions!

vext01:
We drive the PT decoder in an "incremental" fashion. Asking for the next (native-level) block on-demand without the need to decode the entire trace in one go (and store the result in something like a vector). That's good, because it's less RAM hungry: we only store one block's worth of info at a time.

The code that builds and maps the SIR trace is eager though. It processes the whole trace at once and stores the result in a vector which is then iterated. That uses more ram than is technically necessary. The good news is that the SIR trace is not huge and is short-lived, so it may not matter too much.

ltratt:
Maybe I'm a numpty, but why is a trait necessary to do things lazily? In theory, couldn't one have a sort-of iterator that does the same thing?

vext01:
The trait gives you the ability to store or not store a backing vector.

bjorn3:
An `Iterator` could also do that.

vext01:
Now I'm the one confused. Do we mean to replace `SirTrace` with `Iterator<SirLoc>`?

bjorn3:
This PR replaces the `SirTrace` trait with a `Vec<SirLoc>` wrapper. If lazy decoding is wanted in the future, it can be replaced with `Iterator<Item = SirLoc>`.

vext01:
OK. I'll stop being a fence sitter and say: let's merge this. We can change it later if we need to.

All in favour?

ltratt:
Aye!

yktrace/src/swt.rs
```diff
@@ -6,59 +6,33 @@ use libc;
...
-    fn stop_tracing(&mut self) -> Result<Box<dyn SirTrace>, InvalidTraceError> {
+    fn stop_tracing(&mut self) -> Result<SirTrace, InvalidTraceError> {
         let mut len = 0;
         let buf = unsafe { yk_swt_stop_tracing_impl(&mut len) };
         if buf as usize == 0 {
             Err(InvalidTraceError::InternalError)
         } else {
-            Ok(Box::new(SWTSirTrace::from_buf(buf, len)) as Box<dyn SirTrace>)
+            // When we make a SWTSirTrace, we convert all of the locations from core::SirLoc up to
+            // crate::SirLoc and store them in self so that we can hand out references via raw_loc().
```
vext01, line 20: raw_loc() is gone, right?

vext01:
Just one comment then.

bjorn3:
Fine to squash?

yktrace/src/swt.rs
```diff
@@ -16,8 +16,7 @@ impl ThreadTracerImpl for SWTThreadTracer {
         if buf as usize == 0 {
             Err(InvalidTraceError::InternalError)
         } else {
-            // When we make a SWTSirTrace, we convert all of the locations from core::SirLoc up to
-            // crate::SirLoc and store them in self so that we can hand out references via raw_loc().
+            // When we make a SirTrace, we convert all of the locations from SwtLoc to SirLoc.
```
vext01, line 19: I hadn't heard of `SwtLoc` (I thought it was called `core::SirLoc`). I went searching for both of these names in the compiler and found nothing. My guess is that it's been removed and that this code probably doesn't compile.

I guess that's no surprise. We are not testing the (vastly unimplemented) software tracing backend.

My recommendation is to keep the comment as it was, just remove `via raw_loc()`. Then an interested party can at least grep for `core::SirLoc` in the compiler git history.

bjorn3: `core::SirLoc` hasn't existed anymore for a long time. `SwtLoc` is defined lower in this file. I introduced it when I added software tracing support on the Yorick side again for use with my experiments with cg_clif as compiler a while ago: https://github.com/softdevteam/yk/pull/108

vext01: Gotcha

vext01:
Please squash.

vext01:
(and resolve conflicts)

bjorn3:
Rebased and squashed.

vext01:
Just unused imports.

bjorn3:
Fixed and amended.

PR #194, opened 2020-12-16
Misc cleanups

bjorn3:
Last PR for today. This time I didn't base it on any non-merged PR.

### DIFF
--- a/ykrt/src/lib.rs
+++ b/ykrt/src/lib.rs
@@ -1,4 +1,3 @@
-#![feature(negative_impls)]
 #![cfg_attr(test, feature(test))]
 
 pub mod mt;
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -243,10 +243,9 @@ type CompilingTrace<I> = Mutex<Option<Box<CompiledTrace<I>>>>;
 #[derive(Clone)]
 pub struct MTThread {
     inner: Rc<MTThreadInner>,
+    // Raw pointers are neither send nor sync.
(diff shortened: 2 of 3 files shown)


ltratt:
Other than "feature gates are undesirable" (which I agree with), is there any motivation for https://github.com/softdevteam/yk/pull/194/commits/9146db928ee88937d5ae4aece7323b55716a4de8? The reason I went with `!Sync/!Send` is that it's a well-understood way of documenting that a struct isn't send/sync. Although one gets the same type safety from a phantom type, it doesn't quite have the same consistency of documentation. I can be convinced on this, as I tried both designs, and wasn't 100% convinced I chose the best.

bjorn3:
https://github.com/rust-lang/rust/pull/45198 used `PhantomData<*mut Fn()>` to prevent all OIBITs. That is stronger than we need here, but I believe `PhantomData<...>` is generally used over `impl !... for ...` in the standard library.

ltratt:
OK, I can see some instances of this idiom in the standard library (not many, but a few), so I'm semi-swayed. Let's use their precedent and call the attribute either `marker` or (probably better) `_dont_send_or_sync_me`. I think we can delete the doc test too. If Rust makes raw pointers `Send` or `Sync`, this entire code base will collapse in on itself ;)

bjorn3:
Renamed and dropped doc comment.

ltratt:
Please squash.

bjorn3:
Splat

bjorn3:
I dropped the first commit. Locally it showed an unused import warning and compiled fine with removing the import, but on CI it is necessary.

PR #195, opened 2020-12-16
Make ThreadIdInner a struct.

ltratt:
This makes it slightly harder to accidentally cast the wrong thing.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -26,7 +26,8 @@ const PHASE_NUM_BITS: usize = 3;
 const PHASE_TAG: usize = 0b111; // All of the other PHASE_ tags must fit in this.
 const PHASE_COMPILED: usize = 0b000; // Value is a pointer to a chunk of memory containing a
                                      // CompiledTrace<I>.
-const PHASE_TRACING: usize = 0b001; // Value is a pointer to an Arc<u8> representing a thread ID.
+const PHASE_TRACING: usize = 0b001; // Value is a pointer to an Arc<ThreadIdInner> representing a
+                                    // thread ID.
 const PHASE_COMPILING: usize = 0b010; // Value is a pointer to a `Box<CompilingTrace<I>>`.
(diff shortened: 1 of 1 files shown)


ykrt/src/mt.rs
```diff
@@ -235,6 +236,7 @@ impl MTInner {
 }
 
 type CompilingTrace<I> = Mutex<Option<Box<CompiledTrace<I>>>>;
+struct ThreadIdInner(u8);
```
bjorn3, line 239: I think this can be a unit struct. The `Arc` will have to allocate memory any way to fit the strong and weak reference count.

ltratt: I remember reading that there are some `malloc`s somewhere that don't actually allocate memory if you ask for 0 bytes, so I'm a bit nervous about that. i also half-wondered if I should use a `usize` here to force word-level alignment, but that might be unduly paranoid of me!

bjorn3: The `Arc` itself will allocate a memory locatikn big enough to fit the pointer sized strong and weak reference counters before the actual data. https://github.com/rust-lang/rust/blob/2ba7ca2bbbff6cd424aebc654308febc00b9497a/library/alloc/src/sync.rs#L301

ltratt: Good point! Fixed in https://github.com/softdevteam/yk/pull/195/commits/3d21b29494100c9b7a62c71c1cc59c3fe5685681.

bjorn3 (approved):

vext01:
Can you expand on the PR description? I don't understand how the change makes it harder to accidentally cast the wrong thing.

ltratt:
It's just a standard "new type" thing: I can't confuse a new type with a common type (in other words, if I have another `Arc<u8>` somewhere serving a different purpose, I won't confuse that with a `Arc<ThreadIdInner>`).

vext01:
Ah, gotcha! Yes, that's a mighty fine idea.

Please squash.

ltratt:
Squashed.

PR #196, opened 2020-12-17
Remove some spill tests

vext01:
I spent some time trying to salvage these tests, but alas, it didn't work out.

See commit messages for detailed commentary.

Assigning to @ptersilie as he's familiar with this part of the system.

CC @ltratt


ltratt:
The explanations make sense to me.

PR #198, opened 2020-12-17
Encapsulate tracing state bit manipulation

bjorn3:
This also replaces some `Box::from_raw`/`Arc::from_raw` calls followed by a `mem::forget` after using the result by simply turning the raw pointers into references. Both `Box::into_raw` and `Arc::into_raw` return a pointer that points directly to the stored value. For `Arc::into_raw` this means that it points just after the reference counters.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -34,6 +34,63 @@ const PHASE_COUNTING: usize = 0b100; // Value specifies how many times we've see
 const PHASE_LOCKED: usize = 0b101; // No associated value.
 const PHASE_DONT_TRACE: usize = 0b110; // No associated value.
 
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+struct State(usize);
+
+impl State {
+    fn phase(&self) -> usize {
+        self.0 & PHASE_TAG
+    }
+
+    fn phase_compiled<I>(trace: Box<CompiledTrace<I>>) -> Self {
+        let ptr = Box::into_raw(trace);
+        debug_assert_eq!(ptr as usize & PHASE_TAG, 0);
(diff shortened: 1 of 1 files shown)


bjorn3:
Rebased

ykrt/src/mt.rs
```diff
@@ -34,6 +34,62 @@ const PHASE_COUNTING: usize = 0b100; // Value specifies how many times we've see
...
+
+    fn phase_locked() -> Self {
+        State(PHASE_LOCKED)
+    }
+
+    fn phase_dont_trace() -> Self {
+        State(PHASE_DONT_TRACE)
+    }
+
+    fn number_data(&self) -> usize {
```
ltratt, line 80: I think we can add something like `debug_assert_eq!(self.phase(), PHASE_COUNTING)` inside this function.

bjorn3: Done

ltratt:
Please squash.

bjorn3:
Splat

ltratt:
Thanks! I'm going to be doing some work on this file over the next couple of days, so it might be worth holding back anything which touches this, in case my Cunning Ideas for this file actually work :)

bors r+

bjorn3:
Rustfmt'ed

> I'm going to be doing some work on this file over the next couple of days, so it might be worth holding back anything which touches this, in case my Cunning Ideas for this file actually work :)

Ok

ltratt:
Please squash.

[A minor thing: we prefer not to force push unless the reviewer asks, just because it makes it harder to see what's changed.]

bjorn3:
Sorry. I had squashed the rustfmt run into the first commit. No non-stylistical changes since your last review.

PR #199, opened 2020-12-17
Add PHASE_TRACING_LOCK to prevent accessing freed memory.

ltratt:
I have never observed a crash here, but it's clear that one could happen, because we read from memory that another thread might simultaneously free.

This commit introduces a new PHASE_TRACING_LOCK which is a lock that only PHASE_TRACING uses. It means that we can't read possibly-freed memory. It does mean that we introduce two spinlock phases, both of which will have terrible performance in the case of many threads trying to trace the same location. However, we can fix the performance later, perhaps using a futex or something similar.

It might be tempting to think we can reuse PHASE_LOCKED here, as I spent several hours thinking. However, PHASE_LOCKED allows a thread to continue executing when it observes it, whereas  PHASE_TRACING_LOCK stops all threads that observe it until it changes. The reason for this is that if the current trace is tracing the current location, we can't carry on execution, or the resultant trace will contain more than 1 iteration, with consequent hilarity.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -10,7 +10,7 @@ use std::{
         atomic::{AtomicBool, AtomicUsize, Ordering},
         Arc, Mutex, TryLockError,
     },
-    thread::{self, JoinHandle},
+    thread::{self, yield_now, JoinHandle},
 };
 use ykcompile::{CompiledTrace, TraceCompiler};
 use yktrace::{sir::SIR, start_tracing, tir::TirTrace, ThreadTracer, TracingKind};
@@ -28,10 +28,11 @@ const PHASE_COMPILED: usize = 0b000; // Value is a pointer to a chunk of memory 
                                      // CompiledTrace<I>.
 const PHASE_TRACING: usize = 0b001; // Value is a pointer to an Arc<ThreadIdInner> representing a
                                     // thread ID.
-const PHASE_COMPILING: usize = 0b010; // Value is a pointer to a `Box<CompilingTrace<I>>`.
-const PHASE_COUNTING: usize = 0b011; // Value specifies how many times we've seen this Location.
-const PHASE_LOCKED: usize = 0b100; // No associated value.
-const PHASE_DONT_TRACE: usize = 0b101; // No associated value.
+const PHASE_TRACING_LOCK: usize = 0b010; // No associated value
+const PHASE_COMPILING: usize = 0b011; // Value is a pointer to a `Box<CompilingTrace<I>>`.
+const PHASE_COUNTING: usize = 0b100; // Value specifies how many times we've seen this Location.
+const PHASE_LOCKED: usize = 0b101; // No associated value.
+const PHASE_DONT_TRACE: usize = 0b110; // No associated value.
 
 /// A `Location` stores state that the meta-tracer needs to identify hot loops and run associated
 /// machine code.
@@ -58,8 +59,14 @@ pub struct Location<I> {
     //  │             │ start tracing
     //  │             ▼
     //  │           ┌────────────────────┐
+    //  │           │   PHASE_TRACING    │
+    //  │           └────────────────────┘
+    //  │             │ check for    ▲
+    //  │             │ stuckness    | still
+    //  │             ▼              | active
+    //  │           ┌────────────────────┐
     //  │           │                    │  abort   ┌────────────────────┐
-    //  │           │   PHASE_TRACING    │─────────▶│  PHASE_DONT_TRACE  │
+    //  │           │ PHASE_TRACING_LOCK │─────────▶│  PHASE_DONT_TRACE  │
     //  │           │                    │          └────────────────────┘
     //  │           └────────────────────┘
     //  │             │ start compiling trace
@@ -301,6 +308,7 @@ impl MTThread {
         // written to external data. Alternatively, this load could be Relaxed but we would then
         // need to place an Acquire fence in PHASE_COMPILING and PHASE_COMPILED.
         let mut lp = loc.state.load(Ordering::Acquire);
+
         match lp & PHASE_TAG {
             PHASE_COUNTING => {
                 loop {
@@ -347,12 +355,76 @@ impl MTThread {
                     }
                 }
             }
-            PHASE_TRACING => {
+            PHASE_TRACING | PHASE_TRACING_LOCK => {
+                if lp & PHASE_TAG == PHASE_TRACING_LOCK {
+                    // PHASE_TRACING_LOCK requires special handling, because another thread might
+                    // have grabbed it on this Location even though this thread is tracing this
+                    // location.
+
+                    if self.inner.tracer.is_none() {
+                        // If this thread isn't tracing anything, there's little point in spinning
+                        // as the thread tracing this location is still plausibly alive.
+                        return None;
+                    }
+
+                    // We wait to grab the lock.
+                    // FIXME: this is a terrible spinlock.
+                    while lp & PHASE_TAG == PHASE_TRACING_LOCK {
+                        yield_now();
+                        lp = loc.state.load(Ordering::Acquire);
+                    }
+
+                    if lp & PHASE_TAG != PHASE_TRACING {
+                        // While waiting for the Location to move beyond PHASE_TRACING_LOCK, we
+                        // probably raced with the thread tracing this location.
+                        return None;
+                    }
+                }
+
+                // Before we can do anything with this phase's value, we need to transition to
+                // `PHASE_TRACING_LOCK`.
+                loop {
+                    let old_lp = lp;
+                    lp = loc
+                        .state
+                        .compare_and_swap(lp, PHASE_TRACING_LOCK, Ordering::Acquire);
+                    if lp == old_lp {
+                        break;
+                    }
+                    if self.inner.tracer.is_none() {
+                        // This thread isn't tracing anything, so it's not worth us spinning and
+                        // trying to lock this location.
+                        return None;
+                    }
+                    // We are tracing something, and since it might be the current Location, we
+                    // need to transition to TRACING_LOCK to avoid tracing more than one iteration
+                    // of the loop. However, we might be able to determine that this thread
+                    // couldn't have been tracing this location, at which point we can bail out.
+                    // FIXME: this is a terrible spinlock.
+                    loop {
+                        match lp & PHASE_TAG {
+                            PHASE_TRACING => break,
+                            PHASE_TRACING_LOCK => (),
+                            PHASE_COMPILED | PHASE_COMPILING | PHASE_LOCKED | PHASE_COUNTING
+                            | PHASE_DONT_TRACE => {
+                                // If we observe any of these states, then another thread took
+                                // responsibility for this Location, proving that this thread can't
+                                // be responsible for it.
+                                return None;
+                            }
+                            _ => unreachable!(),
+                        }
+                        yield_now();
+                        lp = loc.state.load(Ordering::Relaxed);
+                    }
+                }
+
                 let loc_tid = unsafe { Arc::from_raw((lp & !PHASE_TAG) as *mut ThreadIdInner) };
                 if let Some((_, ref tid)) = self.inner.tracer {
                     // This thread is tracing something...
                     if !Arc::ptr_eq(tid, &loc_tid) {
                         // ...but we didn't start at the current Location.
+                        loc.state.store(lp, Ordering::Relaxed);
                         mem::forget(loc_tid);
                         return None;
                     } else {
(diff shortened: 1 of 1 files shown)


vext01:
Before I review this, it might be worth coordinating with @bjorn3 who has proposed a change in this section of code too.

ltratt:
Unfortunately this PR has to take priority, as this is a correctness fix.

bjorn3:
No problem, should be easy to adapt my PR.

ykrt/src/mt.rs
```diff
@@ -28,10 +28,11 @@ const PHASE_COMPILED: usize = 0b000; // Value is a pointer to a chunk of memory
                                      // CompiledTrace<I>.
 const PHASE_TRACING: usize = 0b001; // Value is a pointer to an Arc<ThreadIdInner> representing a
                                     // thread ID.
-const PHASE_COMPILING: usize = 0b010; // Value is a pointer to a `Box<CompilingTrace<I>>`.
-const PHASE_COUNTING: usize = 0b011; // Value specifies how many times we've seen this Location.
-const PHASE_LOCKED: usize = 0b100; // No associated value.
-const PHASE_DONT_TRACE: usize = 0b101; // No associated value.
+const PHASE_TRACING_LOCK: usize = 0b010; // No associated value
```
vext01, line 31: Should we make all of these descriptions line up to appease my OCD?

ltratt: `cargo fmt` destroys any such attempt -- I've tried :/

ykrt/src/mt.rs
```diff
@@ -58,8 +59,14 @@ pub struct Location<I> {
     //  │             │ start tracing
     //  │             ▼
     //  │           ┌────────────────────┐
+    //  │           │   PHASE_TRACING    │
+    //  │           └────────────────────┘
+    //  │             │ check for    ▲
+    //  │             │ stuckness    | still
```
vext01, line 65: stuckness :P 

Is that a technical term?

ykrt/src/mt.rs
bjorn3: The alternative is to move the comments before the `const`. That would also allow them to be doc-comments.

ykrt/src/mt.rs
ltratt: Given the space constraints in the diagram "yes" :p

ykrt/src/mt.rs
```diff
@@ -347,12 +355,76 @@ impl MTThread {
...
+                    // We wait to grab the lock.
+                    // FIXME: this is a terrible spinlock.
+                    while lp & PHASE_TAG == PHASE_TRACING_LOCK {
+                        lp = loc.state.load(Ordering::Acquire);
+                        yield_now();
+                    }
+
+                    if lp & PHASE_TAG != PHASE_TRACING {
+                        // While waiting for the Location to move beyond PHASE_TRACING_LOCK, we
+                        // probably raced with the thread tracing this location when.
```
vext01, line 379: when... what? Incomplete comment?

ltratt: Fixed in https://github.com/softdevteam/yk/pull/199/commits/77e2bd236333ecb7a9fe79d44cb854f35ab6f067.

ykrt/src/mt.rs
```diff
@@ -347,12 +355,76 @@ impl MTThread {
                     }
                 }
             }
-            PHASE_TRACING => {
+            PHASE_TRACING | PHASE_TRACING_LOCK => {
```
vext01, line 358: Wouldn't it be easier to split this out into separate cases? It seems the nested `PHASE_TRACING_LOCK` case doesn't re-use any code of the `PHASE_TRACING` case. The former always returns before getting there.

vext01: Sorry, that's not true.

ltratt: I think you've realised this, but the alternative would have to be a loop around the whole PHASE `match` statement, which I think would be harder to understand.

vext01:
I'm starting to get worried about the complexity of the control point. It's fiendishly complicated, to the point where I can barely keep the logic in my head at one time.

I wonder if we've prematurely optimised. The only counter I can think to this is that these internals may have an impact on the user-facing API, which we don't want in flux.

Do we know if all of this lock-free dance is any quicker than using simpler concurrency abstractions?

ykrt/src/mt.rs
vext01: I'm actually really struggling in general. See below.

ltratt:
Yes, it's complicated, and it's going to get worse, but I think it's worth it because we know that is a performance critical part of the pre-JIT interpreter. The alternatives are either:

1. A global mutex. This will have truly awful performance.
2. A mutex per Location. This will consume lots of memory (roughly: 64 bytes per locations *plus* malloc overhead).

The first is out of the question. The second would be easier that what we have, but it's probably going to be a fairly hefty overhead, particularly for the common case. By how much? Well, we'd have to implement the whole thing to know...

vext01:
OK. I just wanted to air my thoughts. As long as we are both in agreement that this code is the dragon's lair.

bjorn3:
On Linux you can use futexes which only require a single int to be stored. This int can freely by moved, so it can directly be stored in the `Location` as `AtomicU32`.

vext01:
> It means that we can't read possibly-freed memory.

What memory is this again?

ykrt/src/mt.rs
```diff
@@ -347,12 +355,76 @@ impl MTThread {
...
+                    }
+                }
+
+                // Before we can do anything with this phase's value, we need to transition to
+                // `PHASE_TRACING_LOCK`.
+                loop {
+                    let old_lp = lp;
+                    lp = loc
+                        .state
+                        .compare_and_swap(lp, PHASE_TRACING_LOCK, Ordering::Acquire);
```
vext01, line 390: Can the location already be in `PHASE_TRACING_LOCK` here then?

ykrt/src/mt.rs
```diff
@@ -347,12 +355,76 @@ impl MTThread {
...
+                    if lp == old_lp {
+                        break;
+                    }
+                    if self.inner.tracer.is_none() {
```
vext01, line 394: Could we have checked that earlier?

ykrt/src/mt.rs
```diff
@@ -347,12 +355,76 @@ impl MTThread {
...
+                        // This thread isn't tracing anything, so it's not worth us spinning and
+                        // trying to lock this location.
+                        return None;
+                    }
+                    // We are tracing something, and since it might be the current Location, we
+                    // need to transition to TRACING_LOCK to avoid tracing more than one iteration
```
vext01, line 400: I thought we'd already transitioned to `PHASE_TRACING_LOCK` above?

bjorn3:
The `parking_lot` crate has mutexes that only use 1 byte without any allocations.

vext01:
I looked at those the other day. It looked like they don't guard data, but instead critical sections. Is that right?

ykrt/src/mt.rs
```diff
@@ -347,12 +355,76 @@ impl MTThread {
...
+
+                    if self.inner.tracer.is_none() {
+                        // If this thread isn't tracing anything, there's little point in spinning
+                        // as the thread tracing this location is still plausibly alive.
+                        return None;
+                    }
+
+                    // We wait to grab the lock.
+                    // FIXME: this is a terrible spinlock.
+                    while lp & PHASE_TAG == PHASE_TRACING_LOCK {
```
vext01, line 372: So here we know that this thread is tracing, but we couldn't have been the thread to put the location into `PHASE_TRACING_LOCK`?

ykrt/src/mt.rs
```diff
@@ -347,12 +355,76 @@ impl MTThread {
...
+                        lp = loc.state.load(Ordering::Acquire);
+                        yield_now();
+                    }
+
+                    if lp & PHASE_TAG != PHASE_TRACING {
```
vext01, line 377: Why would we expect `PHASE_TRACING` here?

ykrt/src/mt.rs
ltratt: Yes -- and we do! But we have to recheck it again after grabbing the lock.

ykrt/src/mt.rs
ltratt: At this point we failed to transition to `PHASE_TRACING_LOCK` so we need to try again (or give up).

ykrt/src/mt.rs
ltratt: > So here we know that this thread is tracing

No, we don't know that: we just know that another thread is doing the same check as us.

ykrt/src/mt.rs
ltratt: Because the loop immediately above just says "spin until != PHASE_TRACING_LOCK". Multiple transitions could occur before we observe the tag again.

ltratt:
We only have 3 bits, so full futexes are too big. My long-term intention is something like https://webkit.org/blog/6161/locking-in-webkit/, but until I fully understand all the cases here, I'm not quite sure if we can get away with that or not.

bjorn3:
The parking_lot crate is based on that algorithm.

vext01:
> > It means that we can't read possibly-freed memory.
> 
> What memory is this again?

I'm still unsure what this change is protecting. What memory could have been freed?

ltratt:
@vext01 Basically what's now line 422 (https://github.com/softdevteam/yk/pull/199/files#diff-9fa85f3ab0ab08c2bb4337dd17d47605f520c2780158e50e66eff7cd9dc1b29bR422) is the key. Previously we dereferenced that, and then potentially read from it *before* we'd obtained a lock. Another thread could have freed it underneath us. The window for that to happen was very small, but there was such a window.

vext01:
OK. That helps.

So do we need to be more specific in one of our comments, so that the next person reading this knows what precisely this new lock state is protecting? I've read this code several times and I wasn't any the wiser.

ltratt:
I think if we try and say everything we're trying to prevent, we'll go mad, because there's an almost infinite amount of ways to do things incorrectly. I think the main thing is that the code now explicitly does "lock then dereference" and hopefully that's all anyone needs to care about going forward.

However, what would make sense is for me to slightly extend the commit message if/when we squash/rebase, so that the historical record is clearer.

ykrt/src/mt.rs
```diff
@@ -347,12 +355,76 @@ impl MTThread {
...
+                    if self.inner.tracer.is_none() {
+                        // If this thread isn't tracing anything, there's little point in spinning
+                        // as the thread tracing this location is still plausibly alive.
+                        return None;
+                    }
+
+                    // We wait to grab the lock.
+                    // FIXME: this is a terrible spinlock.
+                    while lp & PHASE_TAG == PHASE_TRACING_LOCK {
+                        lp = loc.state.load(Ordering::Acquire);
```
vext01, line 373: Should we swap this line with the next? By yielding between reading and checking the value, we are maximising the chance of the information being stale, are we not?

ltratt: True. This doesn't affect correctness, but it's clearly inefficient. Fixed in https://github.com/softdevteam/yk/pull/199/commits/c6f0a220f9d9565164854b81bcf5347c161ebe93.

ykrt/src/mt.rs
vext01: Gah! You're right.

ykrt/src/mt.rs
```diff
@@ -347,12 +355,76 @@ impl MTThread {
...
+                }
+
+                // Before we can do anything with this phase's value, we need to transition to
+                // `PHASE_TRACING_LOCK`.
+                loop {
+                    let old_lp = lp;
+                    lp = loc
+                        .state
+                        .compare_and_swap(lp, PHASE_TRACING_LOCK, Ordering::Acquire);
+                    if lp == old_lp {
```
vext01, line 391: I *think* you've done it like this because `if lp != PHASE_TRACING_LOCK` would be incorrect. Can you confirm my suspicion?

ykrt/src/mt.rs
```diff
@@ -347,12 +355,76 @@ impl MTThread {
...
+                    // We wait to grab the lock.
+                    // FIXME: this is a terrible spinlock.
+                    while lp & PHASE_TAG == PHASE_TRACING_LOCK {
+                        yield_now();
+                        lp = loc.state.load(Ordering::Acquire);
+                    }
+
+                    if lp & PHASE_TAG != PHASE_TRACING {
+                        // While waiting for the Location to move beyond PHASE_TRACING_LOCK, we
+                        // probably raced with the thread tracing this location.
```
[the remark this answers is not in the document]
ltratt: It's only probable: for all we know, zillions of transitions have happened since we last checked and what we see now.

ykrt/src/mt.rs
ltratt: Yes, doing the check in that way would fall prey to the ABA problem.

ykrt/src/mt.rs
vext01: `self.inner.tracer` is local to this thread and I don't see anywhere in the `PHASE_TRACING` or `PHASE_TRACING_LOCK` path that would change this field. This check is also common to both cases. Seems like we should be able to check once the first thing we do?

Going by my track record of this PR though, I will probably be wrong :P

ltratt: In both instances, what's happened is that we've encountered a "we wanted to change state but failed", so the question is "should we spin or give up?" If this thread isn't tracing anything, "give up"is the better answer IMHO. Both instances are independent, so I don't think getting rid of one `self.inner.tracer` check makes much sense.

vext01: I'm not saying get rid of it, I'm saying, unless I'm mistaken, you can check it once before all of this dance. It's local to the thread, so no other thread could change it behind our back.

ltratt: It's not that sort of check: it's only used to decide "a state change failed, should I spin or continue?" Conceptually, if I had 20 state changes in this phase, I'd make the same `self.inner.tracer` check after each.

vext01:
Well, I'm starting to feel like I've spent too much time on this PR. I'm probably less than 50% sure I understand this change. On the upside, I hope my questioning has at least forced us (or rather you) to double-check some of the logic?

ltratt:
Questions never hurt! What do you want to do now?

vext01:
Please rebase and squash.

ltratt:
Squashed (and slightly tweaked the commit message).

PR #201, opened 2020-12-17
Make an error if we see an empty trace.

vext01:


yktrace/src/errors.rs
```diff
@@ -3,11 +3,13 @@ use std::fmt::{self, Display, Formatter};
...
+    /// The trace came back empty.
+    EmptyTrace,
+    /// Something went wrong in the compiler's tracing code
+    InternalError,
     /// There is no SIR for the location in the trace.
     /// The string inside is the binary symbol name in which the location appears.
-    NoSir(String),
-    /// Something went wrong in the compiler's tracing code
-    InternalError
+    NoSir(String)
```
bjorn3, line 12:
```suggestion
    NoSir(String),
```

yktrace/src/errors.rs
```diff
@@ -3,11 +3,13 @@ use std::fmt::{self, Display, Formatter};
 #[derive(Debug)]
 /// Reasons that a trace can be invalidated.
 pub enum InvalidTraceError {
+    /// The trace came back empty.
```
ltratt, line 6: I think it would be better to say something like `We recorded an empty trace.`?

yktrace/src/errors.rs
```diff
@@ -3,11 +3,13 @@ use std::fmt::{self, Display, Formatter};
...
+    EmptyTrace,
+    /// Something went wrong in the compiler's tracing code
```
ltratt, line 8: I think the style here is to end every comment with a `.`?

yktrace/src/errors.rs
```diff
@@ -20,10 +22,11 @@ impl InvalidTraceError {
 impl Display for InvalidTraceError {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         match self {
+            InvalidTraceError::EmptyTrace => write!(f, "The trace is empty"),
```
ltratt, line 25: Might "Empty trace" sound more like an error message?

yktrace/src/errors.rs
vext01: Funny that `rustfmt` doesn't canonicalise that.

vext01: Ah wait, it does. It removes the comma.

vext01:
Give those commits a go.

(They are separate to allow easier squashing)

I also need to rebase, which I can do with squashing.

yktrace/src/errors.rs
bjorn3: That is because rustfmt is configured to remove it. The default configuration adds it.

https://github.com/softdevteam/yk/blob/8d57b6239062ff17cd5eeb0cea70516d47948ddf/yktrace/.rustfmt.toml#L3

ltratt: Ah I thought we'd got rid of the custom `.rustfmt.toml`s but maybe not!

vext01: Oh! Let's kill it.

vext01:
Ok to rebase/squash?

yktrace/src/errors.rs
ltratt: Yes, but that should be a separate PR.

vext01: Ah. I already made the commit with the intention of keeping it separate. Shall I revert?

ltratt:
I think https://github.com/softdevteam/yk/pull/201/commits/2c3b330d2a1db7a9df5f0cd7b0bbf0b8f9da2b66 should be its own PR (which should also check for any other rustfmt configs in this repo -- there are at least two, a quick scan suggests). So can you squash this PR, and then we'll do the rustfmt doo-dah in another PR?

vext01:
OK

> there are at least two, a quick scan suggests

Yes, I used `find` to check. There are two.

vext01:
Reverted. OK to squash?

ltratt:
Please squash.

vext01:
splat.

PR #203, opened 2020-12-18
Split Location out into its own file.

ltratt:
This is a somewhat crude split which we'll probably refine over time, but the mt.rs file was getting too crowded and hard to navigate. This commit contains no functional changes.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -12,177 +12,17 @@ use std::{
     },
     thread::{self, yield_now, JoinHandle},
 };
+
 use ykcompile::{CompiledTrace, TraceCompiler};
 use yktrace::{sir::SIR, start_tracing, tir::TirTrace, ThreadTracer, TracingKind};
 
+use crate::location::{
+    CompilingTrace, Location, State, PHASE_COMPILED, PHASE_COMPILING, PHASE_COUNTING,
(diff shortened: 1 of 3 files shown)


bjorn3 (approved):

ykrt/src/mt.rs
```diff
@@ -678,9 +517,9 @@ mod tests {
                     mtt.control_point(Some(&loc), dummy_step, &mut io);
                     let c4 = loc.load(Ordering::Relaxed);
                     assert_eq!(c4.phase(), PHASE_COUNTING);
-                    assert!(c4.0 > c3.0);
-                    assert!(c3.0 > c2.0);
-                    assert!(c2.0 > c1.0);
+                    assert!(c4.number_data() > c3.number_data());
```
vext01, line 520: So we have something additional here. Looks like `number_data` gets the rest of the word, apart from the tag? In which case would `parameter` or `parm` be better terminology?

ltratt: Maybe, but I don't think that's something we want to do in this PR. I only changed this from `.0` to `number_data` because `.0` is private.

vext01: OK. Perhaps give that some thought for another PR. It's just that `number_data` is a bit wishy-washy.

ltratt: My work on this stuff has only just begun, but it might take me a few days to complete it. I suspect this particular bit might change naturally during the course of that.

PR #204, opened 2020-12-18
Reinstate ignored array slice test.

vext01:
Companion PR ~~coming~~ https://github.com/softdevteam/ykrustc/pull/159

This test was checking that a) MIR `Len` rvalues are correctly lowered
to SIR, and that b) the JIT correctly accesses the length in the second
part of the slice's fat pointer (as part of an array indexing operation).

I've managed to fix the test and simplify it in doing so.

I also found a bug in `copy_memory`, where loading the first operand to
memmove(3) overwrote the place we'd load the second operand from.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -549,8 +549,30 @@ impl<TT> TraceCompiler<TT> {
         dynasm!(self.asm
             ; push rax
             ; xor rax, rax
+        );
+
+        dynasm!(self.asm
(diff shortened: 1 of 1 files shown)


ltratt:
Please squash.

bjorn3:
There is only a single commit at the moment...

vext01:
Rebased.

vext01:
This is good to go now.

PR #205, opened 2020-12-19
Use a C abi interface between ykrt and the rest of yorick

bjorn3:
In the future this will make it possible to separately compile the internals of yorick with optimizations and without tracing support.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -257,31 +257,28 @@ fn local_to_reg_name(loc: &Location) -> &'static str {
 }
 
 /// Compile a TIR trace, returning executable code.
-pub fn compile_trace<TT>(tt: TirTrace) -> CompiledTrace<TT> {
+pub fn compile_trace(tt: TirTrace) -> CompiledTrace {
     CompiledTrace {
-        mc: TraceCompiler::<TT>::_compile(tt, false),
-        _pd: PhantomData,
+        mc: TraceCompiler::_compile(tt, false),
     }
 }
 
 /// A compiled `SIRTrace`.
-pub struct CompiledTrace<TT> {
+pub struct CompiledTrace {
     /// A compiled trace.
     mc: dynasmrt::ExecutableBuffer,
-    _pd: PhantomData<TT>,
 }
 
-impl<TT> CompiledTrace<TT> {
+impl CompiledTrace {
     /// Execute the trace by calling (not jumping to) the first instruction's address.
-    pub fn execute(&self, args: &mut TT) -> bool {
-        let func: fn(&mut TT) -> bool =
-            unsafe { mem::transmute(self.mc.ptr(dynasmrt::AssemblyOffset(0))) };
-        self.exec_trace(func, args)
+    pub unsafe fn execute<TT>(&self, args: &mut TT) -> bool {
+        let func: fn(&mut TT) -> bool = mem::transmute(self.mc.ptr(dynasmrt::AssemblyOffset(0)));
+        self.exec_trace::<TT>(func, args)
     }
 
     /// Actually call the code. This is a separate function making it easier to set a debugger
     /// breakpoint right before entering the trace.
-    fn exec_trace(&self, t_fn: fn(&mut TT) -> bool, args: &mut TT) -> bool {
+    fn exec_trace<TT>(&self, t_fn: fn(&mut TT) -> bool, args: &mut TT) -> bool {
         t_fn(args)
     }
 
@@ -392,11 +389,10 @@ enum RegAlloc {
     Free,
 }
 
-use std::marker::PhantomData;
 use ykpack::LocalDecl;
 
 /// The `TraceCompiler` takes a `SIRTrace` and compiles it to machine code. Returns a `CompiledTrace`.
-struct TraceCompiler<TT> {
+struct TraceCompiler {
     /// The dynasm assembler which will do all of the heavy lifting of the assembly.
     asm: dynasmrt::x64::Assembler,
     /// Stores the content of each register.
@@ -409,19 +405,17 @@ struct TraceCompiler<TT> {
     stack_builder: StackBuilder,
     /// Stores the memory addresses of local functions.
     addr_map: HashMap<String, u64>,
-    _pd: PhantomData<TT>,
 }
 
-impl<TT> TraceCompiler<TT> {
+impl TraceCompiler {
     fn new(local_decls: HashMap<Local, LocalDecl>, addr_map: HashMap<String, u64>) -> Self {
-        let mut tc = TraceCompiler::<TT> {
+        let mut tc = TraceCompiler {
             asm: dynasmrt::x64::Assembler::new().unwrap(),
             register_content_map: REG_POOL.iter().map(|r| (*r, RegAlloc::Free)).collect(),
             variable_location_map: HashMap::new(),
             local_decls,
             stack_builder: StackBuilder::default(),
             addr_map,
-            _pd: PhantomData,
         };
 
         // At the start of the trace, jump to the label that allocates stack space.
@@ -688,7 +682,7 @@ impl<TT> TraceCompiler<TT> {
         let sym_addr = if let Some(addr) = self.addr_map.get(sym) {
             *addr as i64
         } else {
-            TraceCompiler::<TT>::find_symbol(sym)? as i64
+            TraceCompiler::find_symbol(sym)? as i64
         };
         dynasm!(self.asm
             // In Sys-V ABI, `al` is a hidden argument used to specify the number of vector args
@@ -1855,9 +1849,11 @@ mod tests {
         simple(&mut IO(0));
         let sir_trace = th.stop_tracing().unwrap();
         let tir_trace = TirTrace::new(&*SIR, &sir_trace).unwrap();
-        let ct = compile_trace::<IO>(tir_trace);
+        let ct = compile_trace(tir_trace);
         let mut args = IO(0);
-        ct.execute(&mut args);
+        unsafe {
+            ct.execute::<IO>(&mut args);
+        }
         assert_eq!(args.0, 13);
     }
 
@@ -1865,15 +1861,13 @@ mod tests {
     // should not exhaust the allocator.
     #[test]
     fn reg_alloc_same_local() {
-        struct IO(u8);
-
         let types = TestTypes::new();
         let mut local_decls = HashMap::new();
         local_decls.insert(Local(0), LocalDecl::new(types.t_u8, false));
         local_decls.insert(Local(1), LocalDecl::new(types.t_i64, false));
         local_decls.insert(Local(2), LocalDecl::new(types.t_string, false));
 
-        let mut tc = TraceCompiler::<IO>::new(local_decls, Default::default());
+        let mut tc = TraceCompiler::new(local_decls, Default::default());
         let u8_loc = tc.local_to_location(Local(0));
         let i64_loc = tc.local_to_location(Local(1));
         let string_loc = tc.local_to_location(Local(2));
@@ -1890,8 +1884,6 @@ mod tests {
     // Locals should be allocated to different registers.
     #[test]
     fn reg_alloc() {
-        struct IO(u8);
-
         let types = TestTypes::new();
         let mut local_decls = HashMap::new();
         for i in (0..9).step_by(3) {
@@ -1900,7 +1892,7 @@ mod tests {
             local_decls.insert(Local(i + 2), LocalDecl::new(types.t_string, false));
         }
 
-        let mut tc = TraceCompiler::<IO>::new(local_decls, Default::default());
+        let mut tc = TraceCompiler::new(local_decls, Default::default());
         let mut seen: Vec<Location> = Vec::new();
         for l in 0..7 {
             let reg = tc.local_to_location(Local(l));
@@ -1912,8 +1904,6 @@ mod tests {
     // Once registers are full, the allocator should start spilling.
     #[test]
     fn reg_alloc_spills() {
-        struct IO(u8);
-
         let types = TestTypes::new();
         let num_regs = REG_POOL.len() + 1; // Plus one for TIO_REG.
         let num_spills = 16;
@@ -1926,7 +1916,7 @@ mod tests {
             );
         }
 
-        let mut tc = TraceCompiler::<IO>::new(local_decls, Default::default());
+        let mut tc = TraceCompiler::new(local_decls, Default::default());
 
         for l in 0..num_regs {
             assert!(matches!(
@@ -1946,8 +1936,6 @@ mod tests {
     // Freeing registers should allow them to be re-allocated.
     #[test]
     fn reg_alloc_spills_and_frees() {
-        struct IO(u8);
-
         let types = TestTypes::new();
         let num_regs = REG_POOL.len() + 1; // Plus one for TIO_REG.
         let num_decls = num_regs + 4;
@@ -1959,7 +1947,7 @@ mod tests {
             );
         }
 
-        let mut tc = TraceCompiler::<IO>::new(local_decls, Default::default());
+        let mut tc = TraceCompiler::new(local_decls, Default::default());
 
         // Fill registers.
         for l in 0..num_regs {
@@ -1999,7 +1987,6 @@ mod tests {
     // Test cases where a local is allocated on the stack even if registers are available.
     #[test]
     fn reg_alloc_always_on_stack() {
-        struct IO(u8);
         let types = TestTypes::new();
         let mut local_decls = HashMap::new();
 
@@ -2021,7 +2008,7 @@ mod tests {
         local_decls.insert(Local(5), LocalDecl::new(types.t_tiny_array, false));
         local_decls.insert(Local(6), LocalDecl::new(types.t_tiny_tuple, false));
 
-        let mut tc = TraceCompiler::<IO>::new(local_decls, Default::default());
+        let mut tc = TraceCompiler::new(local_decls, Default::default());
 
         // Things larger than a register shouldn't be allocated to a register. Here we are
         // allocating space for a `String`, which at the time of writing, is much larger than the
@@ -2067,9 +2054,11 @@ mod tests {
         fcall(&mut io);
         let sir_trace = th.stop_tracing().unwrap();
         let tir_trace = TirTrace::new(&*SIR, &sir_trace).unwrap();
-        let ct = compile_trace::<IO>(tir_trace);
+        let ct = compile_trace(tir_trace);
         let mut args = IO(0);
-        ct.execute(&mut args);
+        unsafe {
+            ct.execute::<IO>(&mut args);
+        }
         assert_eq!(args.0, 13);
     }
 
@@ -2096,17 +2085,18 @@ mod tests {
         fnested(&mut io);
         let sir_trace = th.stop_tracing().unwrap();
         let tir_trace = TirTrace::new(&*SIR, &sir_trace).unwrap();
-        let ct = compile_trace::<IO>(tir_trace);
+        let ct = compile_trace(tir_trace);
         let mut args = IO(0);
-        ct.execute(&mut args);
+        unsafe {
+            ct.execute::<IO>(&mut args);
+        }
         assert_eq!(args.0, 20);
     }
 
     // Test finding a symbol in a shared object.
     #[test]
     fn find_symbol_shared() {
-        struct IO(());
-        assert!(TraceCompiler::<IO>::find_symbol("printf") == Ok(libc::printf as *mut c_void));
+        assert!(TraceCompiler::find_symbol("printf") == Ok(libc::printf as *mut c_void));
     }
 
     // Test finding a symbol in the main binary.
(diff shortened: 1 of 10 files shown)


bjorn3:
I am not quite sure yet how it should be compiled with a different compiler yet. I tried using a build script, but that will require recompiling everything on every change. In addition it will hide the output of the compilation process unless `-vv` is passed to cargo. I guess in the future it can become part of the rustc sysroot itself. For now maybe a script could be used to compile the internals first and then to pass the right link arguments when compiling the user of yorick?

ltratt:
This looks interesting, and I understand the motivation, but maybe you can spell out any possible consequences too?

bjorn3:
In the future it may be necessary to extend this interface, which could make it a bit harder to develop new features. I don't expect this to be necessary for the foreseeable future though. It also increases the amount of unsafe code.

ltratt:
> In the future this will make it possible to separately compile the internals of yorick with optimizations and without tracing support.

Possibly stupid question: how does this make this feasible? I have a couple of guesses, but I'm not sure which might be correct (maybe none of them!).

bjorn3:
Rust doesn't have a stable abi, so if you want to call code compiled by one version of rustc from code compiled by a different version of rustc, you must use the C abi or another stable abi.

ltratt:
I see. The next question is whether putting the division line in ykrt is necessary and/or sufficient. I don't immediately know the answer to that, I must admit! @vext01 might?

vext01:
So this is all so that we can compile most of the yorick with one set of optimisation flags, and the user's interpreter with another. Am I right?

I'm probably not going to get a chance to do a full review until Monday.

vext01:
> I am not quite sure yet how it should be compiled with a different compiler yet.

A couple of thoughts on this.

Instead of invoking rustc directly, could the bits we want to compile separately be managed by a separate Cargo workspace? Then we would get incremental compilation, I think?

Also, we could use the same ykrustc compiler, just remove the `-C tracer=hw` flags.

Thanks for this PR, I'll review it in full now.

ykcompile/src/lib.rs
```diff
@@ -257,31 +257,28 @@ fn local_to_reg_name(loc: &Location) -> &'static str {
 }
 
 /// Compile a TIR trace, returning executable code.
-pub fn compile_trace<TT>(tt: TirTrace) -> CompiledTrace<TT> {
```
vext01, line 260: It's a shame that we lose the type safety here. I wonder if there's another way we could ensure that the trace for one control point isn't used by another control point? Perhaps it's not worth worrying...

vext01: Would one solution be to make wrappers that have ignored type parameters?

ykrt/src/internal.rs
```diff
@@ -0,0 +1,81 @@
+use std::ffi::{CString, c_void};
```
vext01, line 1: If we choose to merge this, I think we should have a module-level doc comment explaining the design decision.

ykrt_internal/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+
+fn handle_panic<T>(f: impl FnOnce() -> T + UnwindSafe) -> T {
+    std::panic::catch_unwind(|| f()).unwrap_or_else(|_| {
+        // It is not safe for panics to cross through the "C" abi.
+        // FIXME use "C unwind" as abi once it is implemented.
+        std::process::abort();
+    })
+}
+
+/// The different ways by which we can collect a trace.
```
vext01, line 17: This enum is duplicated. Is there a way to share it? Is `include!` a bad idea?

ykrt_internal/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+    SoftwareTracing = 0,
+    /// Hardware tracing via ykrustc + hwtracer.
+    HardwareTracing = 1,
+}
+
+#[no_mangle]
+unsafe extern "C" fn __ykrt_start_tracing(tracing_kind: u8) -> *mut c_void {
+    handle_panic(|| {
+        let tracing_kind = match tracing_kind {
+            0 => yktrace::TracingKind::SoftwareTracing,
```
vext01, line 32: If we enforce the constraint that the tracing kinds are in sync either side of the boundary, we can simplify this I think.

vext01:
I think this is likely to be a (very) worthwhile change, but I'm a little hesitant to merge until a) we know we can get the build system details to work for us, and b) that we see a good speed up from optimising the JIT internals.

I wonder if we should flesh out the build system changes in this PR and then do some quick and dirty benchmarking with multitime?

@ptersilie do you have any thoughts?

ykcompile/src/lib.rs
bjorn3: The wrapper in `ykrt` does have the type parameter.

ykrt_internal/src/lib.rs
bjorn3: If you ever wanted to publish it on crates.io, that won't work.

ykcompile/src/lib.rs
vext01: Cool. I missed that.

ykrt_internal/src/lib.rs
vext01: Fair enough. We can live with the duplication.

ykrt_internal/src/lib.rs
ltratt: How would we enforce that constraint? Is it just a case of documenting it?

vext01: Enforcement via documentation, yeah.

vext01:
I spoke to @ptersilie offline. He raised some other concerns (he's afk, but said I could echo here):

 * Would the new separated crate be statically linked? We don't want to carry around >1 artefact if we can help it.
 * Would this interfere with instantiating the blackhole interpreter? I (Edd), think this should be OK, as the interpreter would too be contained in side the "internal" crate, I think.
 * Does this change our development workflow much?

I hope that covers @ptersilie's concerns. Thanks!

bjorn3:
> Would the new separated crate be statically linked? We don't want to carry around >1 artefact if we can help it.

Using `staticlib` as crate type will cause all libstd symbols to leak. If the libstd is not the same, this isn't a problem for most symbols as they will be mangled differently, but there are some symbols that aren't mangled and thus would still conflict. Using a `cdylib` would solve the symbol leaking issue, but would require the final executable to be shipped with a dynamic library.

On Linux it is possible to link the static library using `ld -r` to a single object file for which you can then hide all symbols except for the ones that need to be exported using objcopy. On macOS this results in a linker error when trying to link against the object file for some reason. This is a problem for cg_clif too as it needs it to compile inline asm using an external assembler and then link both the asm object and the regular object as the incremental system expects a single object file for each codegen unit.

> Would this interfere with instantiating the blackhole interpreter? I (Edd), think this should be OK, as the interpreter would too be contained in side the "internal" crate, I think.

Yes, the interpreter should also be inside the "internal" crate. I think it may be a good idea to make a sketch of the C api before this PR is merged.

> Does this change our development workflow much?

That is a good question. All tests that use tracing to get TIR need to be moved to ykrt or a crate that depends on ykrt. As the rest of the crates will no longer be built with tracing support. In addition the internal crate will somehow need to be built separately from ykrt. Finding a way to do this without hurting development is pretty important. For cg_clif I use a bunch of scripts to do everything from building cg_clif and building the sysroot to running tests. It works, but I don't know if you will like it as much as I do.

vext01:
> Using staticlib as crate type will cause all libstd symbols to leak.

And I suppose we would encounter the same issue if we use a rlib?

I wonder if there is an alternative solution that doesn't require restructuring so much. Could we get ykrustc to recognise the `yktrace` crate (and friends) and optimise all of their items which are not marked `#[optimise(never)`? (It would probably require extending the `optimise` attribute, or adding our own attribute).

bjorn3:
> Could we get ykrustc to recognise the yktrace crate (and friends) and optimise all of their items which are not marked #[optimise(never)?

It is not just yktrace, but all of it's dependencies, ykcompile and ykbh. Basically everything except for ykrt and the user crate/dependencies.

vext01:
I see.

The staticlib problem is concerning. We would like this to run on other systems in the (probably) distant future, so we probably don't want to constrain ourselves too much if we can help it.

vext01:
And does using a shared object solve these issues? The option to install into the compiler's sysroot could still be a path forward?

ptersilie:
> Yes, the interpreter should also be inside the "internal" crate. I think it may be a good idea to make a sketch of the C api before this PR is merged.

I'm currently working on the blackholing stuff and the goal with the interpreter is to be instantiated by the JIT (upon guard failure) and then the interpreter instance is handed to the meta-tracer who runs it. Would this PR interfere with that? There is of course room to change the way the interpreter is instantiated and ran, though conceptually I dislike the idea of the compiler running the interpreter. But I guess that's a price I'm willing to pay to make the compiler/interpreter more performant.

I think sketching out the ABI would be a very good idea at this point, to see how this affects the rest of the system.

bjorn3:
> And does using a shared object solve these issues? The option to install one into the compiler's sysroot could still be a path forward?

Yes, a cdylib works.

> I'm currently working on the blackholing stuff and the goal with the interpreter is to be instantiated by the JIT (upon guard failure) and then the interpreter instance is handed to the meta-tracer who runs it. Would this PR interfere with that? There is of course room to change the way the interpreter is instantiated and ran, though conceptually I dislike the idea of the compiler running the interpreter. But I guess that's a price I'm willing to pay to make the compiler/interpreter more performant.

The interpreter would have to export a C abi function anyway for the compiled trace to switch to it when a guard fails.

> I think sketching out the ABI would be a very good idea at this point, to see how this affects the rest of the system.

I think the first step would be to determine how to switch from the compiled trace to the blackhole interpreter.

vext01:
I think in order to move this PR along, we need to measure the speedup and then decide if it's large enough for us to take on the additional complexity.

@bjorn3 How hard is it to get this branch in a position where we can time runs of the test suite?

ltratt:
@vext01 just so that everyone is clear, you're basically asking for something like `cargo build --tests; multitime -n 10 -s 0 cargo test` before and after this PR?

vext01:
Correct 👍

bjorn3:
> @bjorn3 How hard is it to get this branch in a position where we can time runs of the test suite?

Most tests must be changed to either use ykrt instead of directly the internal code, or rewritten to take fixed SIR and a fixed trace as input.

ltratt:
> Most tests must be changed to either use ykrt instead of directly the internal code, or rewritten to take fixed SIR and a fixed trace as input.

Does that mean that, at the moment, `cargo test` in this branch will fail?

bjorn3:
No, but this branch currently doesn't support building the internal crates or tests without `-Ctracer=hw` and as such nothing is optimized just like on master.

vext01:
> Most tests must be changed to either use ykrt instead of directly the internal code, or rewritten to take fixed SIR and a fixed trace as input.

I prefer the former. Is that something you'd be willing to work on, or shall I take over?

bjorn3:
I am currently doing midterm exams.

vext01:
> I am currently doing midterm exams

No worries. We can take it from here. Hope the exams go well.

vext01:
Is there a reason for the control point itself not being the other side of the optimisation threshold (in ykrt_internal)?

bjorn3:
I had a gut feeling that not calling into ykrt_internal while tracing, but only at the start and end would be better. Maybe in the future `-Ctracer=hw` could allow some optimizations like inlining that could be made to work with tracing?

vext01:
Another thing that crosses my mind: if we enforce that both ykrt_internal and the rest of yk are built using the same compiler (just ykrt_internal is optimised), then can we use the Rust ABI with a cdylib (instead of the C ABI)? I'm guessing that doesn't work...

bjorn3:
ykrt_internal should be compiled with an optimized libstd, while ykrt should be compiled with a `-Ctracer=hw` libstd. This is very likely to cause abi incompatibilities.

vext01:
> ykrt_internal should be compiled with an optimized libstd, while ykrt should be compiled with a -Ctracer=hw libstd. This is very likely to cause abi incompatibilities.

Is the concern here that with `-C tracer=hw` optimisations are disabled? I don't think `-C tracer=hw` changes any ABI beyond that.

In regular Rust, when you do a debug build of a crate, don't you mix unoptimised code with an optimised standard library anyway? As I understand, you can't build a debug standard library and thus have to always use an optimised one.

bjorn3:
> Is the concern here that with -C tracer=hw optimisations are disabled? I don't think -C tracer=hw changes any ABI beyond that.

The different `-Ctracer` value causes the crate hash to differ as it is tracked, this could affect the ABI, but I think it currently doesn't.

> In regular Rust, when you do a debug build of a crate, don't you mix unoptimised code with an optimised standard library anyway? As I understand, you can't build a debug standard library and thus have to always use an optimised one.

In that case you get a single optimized libstd. In this case we want ykrt_internal to use an optimized libstd, while the user code gets an unoptimized libstd. If cargo passes a different `-Cmetadata`, the symbol names will differ and you will get duplicates of certain statics that mustn't be duplicated, while if it doesn't, you will simply end up with a linker error due to duplicated symbols, though in either case you will likely get a linker error anyway due to a duplicate definition of the global allocator methods. The problem is not that you can't mix unoptimized and optimized code, but that you must not link two separately compiled libstd's into a single executable or dylib and must not try to interpret a type from one of the compilations as one from the other compilation as would be necessary when using libstd types in the ffi function signatures.

vext01:
Right, thanks for the info.

I'm going to try and get this into benchmarkable shape this week, probably using a cdylib for now.

vext01:
Status update: I've copied your branch to my fork, rebased it, and got the separate compilation partially working.

[This commit](https://github.com/vext01/yk_pv/commit/5cc555a1faea08c8dd29ec4f0e546d17e75070c0) uses an xtask to invoke the same `cargo` and `rustc` as for the rest of the project, but with optimisations. It's quick and dirty for now.

We currently need to set LD_LIBRARY_PATH too, but we may be able to set an rpath later.

One test fails. Something to do with threads.

> Most tests must be changed to either use ykrt instead of directly the internal code, or rewritten to take fixed SIR and a fixed trace as input.

I now understand this comment better. As it stands, only the ykrt tests will use the new optimised innards.

I'm tempted to move *most* of the system into a new "optimised" workspace, and then all tests into an "unoptimised" workspace and have only a minimal API between. That way no test can inadvertently use unoptimised internals directly accidentally.

vext01:
Hi @bjorn3,

Your idea gives a pretty decent speed up. Enough for us to consider moving this forward.

Here are timings (over 30 runs) for the ykrt test suite on my branch (skipping two tests which crash trying to get a PT file descriptor):
```
Before:
1: env RUST_TEST_THREADS=1 target/debug/deps/ykrt-589c0db448b16d4d
real        2.648       0.061       2.526       2.661       2.766
user        24.545      0.248       23.971      24.612      25.047
sys         0.204       0.022       0.161       0.204       0.248

1: target/debug/deps/ykrt-589c0db448b16d4d
real        2.496       0.112       2.307       2.493       2.818
user        25.704      0.502       24.781      25.825      26.609
sys         0.227       0.056       0.092       0.226       0.401

After:
1: env RUST_TEST_THREADS=1 target/debug/deps/ykrt-b424f141bb012e29
real        1.672       0.148       1.557       1.613       2.104
user        22.127      0.705       21.383      21.959      24.410
sys         0.053       0.012       0.036       0.048       0.088

1: target/debug/deps/ykrt-b424f141bb012e29
real        1.613       0.027       1.565       1.612       1.674
user        22.359      0.103       22.136      22.366      22.539
sys         0.059       0.015       0.036       0.056       0.100
```

Having worked on this, I have fresh questions.

> The problem is not that you can't mix unoptimized and optimized code, but that you must not link two separately compiled libstd's into a single executable or dylib and must not try to interpret a type from one of the compilations as one from the other compilation as would be necessary when using libstd types in the ffi function signatures.

I'm still really keen to use the rust ABI if we can make it at all possible.

So you are saying that we have one libstd for the optimised portion inside the cydlib and another copy of libstd for the unoptimised portion and we can't leak types from one stdlib across the optimisation threshold due to potential incompatibilities introduced by `-C tracer=hw`.

Currently the std lib in the ykrustc sysroot will only ever be compiled *without* `-C tracer=hw`. We don't do anything special for the stdlib right now, as we are unable to trace it anyway. So both the optimised and unoptimised portions will use the same sole stdlib but each will have their own copy. Does that mean that we can relax the constraints and in fact use the rust ABI instead of the C ABI?

There is of course a question of whether we might be able to build an unoptimised stdlib at some time in the future, thus allowing tracing the stdlib. At that point we may link a different stdlib in the optimised and unoptimised portions. At that point, we clearly shouldn't use the rust ABI.

So in fact, I think I've talked myself out of using the Rust ABI :P

Any thoughts on this before I try to get the rest of the tests using the optimised internals?

bjorn3:
> So you are saying that we have one libstd for the optimised portion inside the cydlib and another copy of libstd for the unoptimised portion and we can't leak types from one stdlib across the optimisation threshold due to potential incompatibilities introduced by -C tracer=hw.

It is more so that leaking from one standard library to another is not guaranteed to work unless the SVH matches, which it doesn't due to being compiled with different options.

> Currently the std lib in the ykrustc sysroot will only ever be compiled without -C tracer=hw. We don't do anything special for the stdlib right now, as we are unable to trace it anyway. So both the optimised and unoptimised portions will use the same sole stdlib but each will have their own copy. Does that mean that we can relax the constraints and in fact use the rust ABI instead of the C ABI?

Huh, didn't know that it is compiled without `-Ctracer=hw`. In that case it should technically be possible to use the rust ABI for as long as the same libstd is used and all types outside of libstd that are passed are `#[repr(C)]`. In that case a dylib instead of cdylib should probably be used to make both ykrt_internal and the user program link against the same libstd.

> There is of course a question of whether we might be able to build an unoptimised stdlib at some time in the future, thus allowing tracing the stdlib. At that point we may link a different stdlib in the optimised and unoptimised portions. At that point, we clearly shouldn't use the rust ABI.

Yeah, I actually thought that that was already the case.

vext01:
> Yeah, I actually thought that that was already the case.

Since we want it to be this way, we should stick with the C abi, I think.

vext01:
This morning I've discussed the work-in-progress branch with @ptersilie. The change is obviously good for performance, but we are concerned about the level of complexity we are adding.

To help with our decision, we decided to write a list of pros and cons. Here they are (did I miss anything?):

Pros:
 - Good speedup :+1: 

Cons:
 - Requires use of externs to cross into the optimised portion of code.
   - Which should probably use the C ABI, since in the future we may need two
     different stdlibs.
   - Often means passing raw pointers.
   - Since tests have to be unoptimised, we have to expose (otherwise private)
     internals via externs too.
   - Requires careful boxing of structs passed over the threshold as raw
     pointers.
   - Makes changing the code (which is in flux) more laborious.

 - Requires non-standard build system.
   - Makes dependency management harder.
   - Confuses rust-analyzer (without further config).

Obviously some points carry more weight that others, but in sum

At this time, since the tracer internals are in flux, I'm leaning towards "let's not do this now". In these early stages we probably care more about productivity than performance. However, I could be swayed if anyone disagrees.

[Side-note: We may not have needed the `handle_panic()` stuff, as we build with the abort strategy.]

ltratt:
I think some of those "cons" aren't "cons", they're just "differences". Specifically these don't seem like real cons:

* Which should probably use the C ABI, since in the future we may need two different stdlibs.
* Often means passing raw pointers.
* Requires careful boxing of structs passed over the threshold as raw pointers.
* Requires non-standard build system.
* Confuses rust-analyzer (without further config).

These two are too vague for me to have an opinion on:

* Makes changing the code (which is in flux) more laborious.
* Makes dependency management harder.

However these does seem like a real con:

* Since tests have to be unoptimised, we have to expose (otherwise private) internals via externs too.

Can you expand the "vague" parts somewhat?

ptersilie:
It's difficult to quantify "makes future changes harder", as we don't really know what we might be changing in the future. Just to give an example: While working on blackholing, I had at some point the choice whether the `SIRInterpreter` is instantiated by the trace compiler (ykcompile), or by the meta-tracer (ykrt). Both are valid solutions. With this change, presuming we want the `SIRInterpreter` to be optimised, it needs to be created by the trace compiler. And if we would change our minds, it would be annoying to fix.

That being said, this could also be viewed as a positive, as it forces us to make decisions earlier rather than later. Let's say I had chosen to put the `SIRInterpreter` inside the meta-tracer, and we add this PR later, then this would be an annoying change to do.

vext01:
> Can you expand the "vague" parts somewhat?

Sure!

> Makes changing the code (which is in flux) more laborious.

Suppose that we want to call an optimised function `g` from an unoptimised function. Instead of just calling it, we have to:
 - Add a extern stub for `g` in the unoptimised portion of the code, using appropriate C types as necessary.
 - Implement `g` in the optimised portion of code, but accepting C types, doing boxing as needed.
 - The stub and the implementation need to be kept in sync.

> Makes dependency management harder.

The way it's implemented is as two Rust workspaces instead of one, and deps across the workspaces are implicit. This is only a minor annoyance, we can just always run cargo on the optimised workspace, even if it didn't change.

> Specifically these don't seem like real cons

I think they are real, but minor cons. But their sum is not insignificant.

One minorish con I forgot:

 - The yk (and thus the user's interpreter) will no longer be stand-alone, since it's now a binary and a shared object (it was just a binary before). Consumers of yk would probably also need to do linkage dances to ensure the library is found and comple- and run-time.

vext01:
I spoke to @bjorn3 on Zulip and he concluded (with permission for me to paste this):
> I think this will be necessary in the long term, but I don't know if it should be done right now anymore.

So with this in mind, @ltratt do you think we will need this optimisation when we come to write the paper this year? If not, we can probably put this change off until later. But if we do want to show the best performance we can *this year*, then I'm less sure. I'm already in the zone for this change, but my brain will bitrot over time :P

ltratt:
If we're going to do it at some point, why wouldn't we do it now?

I guess my "big" question is whether, for this:

> Suppose that we want to call an optimised function g from an unoptimised function. Instead of just calling it, we have to:

we could (at least in theory) automate the change?

bjorn3:
> we could (at least in theory) automate the change?

I think it should be possible to have a single file with a macro invocation describing the api between both halves. Both halves can then define this macro in their own way and include the file.

vext01:
OK unless @ptersilie fiercely objects, let's bite the bullet and do it (without automation first). Sometimes you gotta be brave (and I rarely am) 😆

ptersilie:
Fine with me!

ltratt:
@vext01 go for it!

[BTW, you might want to `git merge upstream/master` before going much further?]

vext01:
I'm on it. And I'm already merged. I'm working in a different branch to this PR.

bjorn3:
Closing in favor of #214.

PR #210, opened 2021-01-14
Update dependencies.

vext01:
This change does three things:
 - Update all dependencies to latest versions, including dynasm to silence a security audit warning about the unmaintained `memmap` crate. Dynasm now uses `memmap2`.
 - Also use `memmap2` in our own code, again to silenced the audit.
 - A minimal fix to silence a deprecation warning that would otherwise block CI.

### DIFF
--- a/ykcompile/Cargo.toml
+++ b/ykcompile/Cargo.toml
@@ -8,16 +8,16 @@ license = "Apache-2.0 OR MIT"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-dynasmrt = "1.0.0"
+dynasmrt = "1.0.1"
 hex = "0.4.2"
 lazy_static = "1.4.0"
-libc = "0.2.80"
+libc = "0.2.82"
 ykpack = { path = "../ykpack" }
 yktrace = { path = "../yktrace" }
 
 [build-dependencies]
-cc = "1.0.62"
+cc = "1.0.66"
 
 [dev-dependencies]
-fm = "0.1.4"
-regex = "1.4.2"
+fm = "0.2.0"
+regex = "1.4.3"
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -1445,7 +1445,7 @@ mod tests {
         let text_re = Regex::new(r"\$?.+?\b").unwrap(); // Any word optionally prefixed with `$`.
         let matcher = FMBuilder::new(ptn)
             .unwrap()
-            .name_matcher(Some((ptn_re, text_re)))
+            .name_matcher(ptn_re, text_re)
             .distinct_name_matching(true)
             .build()
             .unwrap();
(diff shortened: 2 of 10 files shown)


ltratt:
There's little point in adding X.Y.V dependencies e.g. `0.2.82`: cargo interprets this as "as any version of 0.2 >= 0.2.82" so just saying "0.2" is more practical. There is one caveat to this: if we *absolutely* need a specific minor version, then putting it in the `Cargo.toml` means that people don't have to do `cargo update` in order to get an update. But, in general, specifying the minor version just leads to unnecessary churn in `Cargo.toml`s. Can we take the opportunity in this PR to remove the not-very-useful minor version numbers (which, I would guess, is probably all of them?).

vext01:
I just used `cargo upgrade`.

ltratt:
> I just used cargo upgrade.

It doesn't matter *how* it was done :) Can you address the meat of my comment: if you disagree with it, let's thrash it out. I might not be right!

vext01:
My objection is that if you have `libc = "0.2"` in your `Cargo.toml` and run `cargo upgrade` it will upgrade it to version `0.2.82`, with an `x.y.z`-style version string. It would be incredibly annoying to have to edit every upgraded dependency by hand.

I looked at the top downloaded crates on crates.io to see what the precedent is. Of the ones i looked at, only one (syn) doesn't use `x.y.z`.

For those reasons, I think we should just expand the version.

ltratt:
Does `cargo upgrade` ever change/suggest a change of a major version? Because if it doesn't do so, then I'm not sure that it does anything useful for us.

vext01:
> Does cargo upgrade ever change/suggest a change of a major version?

Yes, that's how it differs from `cargo update`, it changes `Cargo.toml` instead of `Cargo.lock` and upgrades to the latest version regardless of compatibility.

For example, if I roll back `libc` to `0.1` and run it:
```
Upgrading libc v0.1 -> v0.2.82
```

ltratt:
Hmm, I'm not how sure how keen I am on automatically upgrading major versions: it feels like something that someone should do manually so that we actually know it's safe to do. What do you think?

vext01:
I take your point.

Perhaps in an ideal world, yes, we'd manually check upgrades, but in reality that's quite laborious. We'd have to audit the incompatibilities of all of our dependencies, of which there are many... Tricky.

vext01:
How about this angle:

Supposing I use `cargo upgrade` to see preliminary potential upgrades, which I then verify manually and approve. Should I then be expected to manually strip trailing version numbers? Doesn't that undermine the point of the automation?

vext01:
Give that commit a go.

ltratt:
After offline discussion, I think you should remove that last commit and force push. Then what you should do is comment on each non-minor-version update (just fm in this case) and confirm that you've checked it's OK. Let's see if that process works for all of us!

vext01:
OK!

So with `fm` the `name_matcher` method changed from taking a 2-tuple to taking two arguments.

Good to squash?

ltratt:
Please do!

vext01:
splat.

PR #211, opened 2021-01-15
Downgrade serde a minor version.

vext01:
In the last deps update I broke ykrustc:
```
error: failed to select a version for `serde`.
    ... required by package `ykpack v0.1.0 (https://github.com/softdevteam/yk#6abf57f4)`
    ... which is depended on by `rustc_codegen_llvm v0.0.0 (/home/vext01/research/yorick/ykrustc/compiler/rustc_codegen_llvm)`
    ... which is depended on by `rustc_interface v0.0.0 (/home/vext01/research/yorick/ykrustc/compiler/rustc_interface)`
    ... which is depended on by `rustc_driver v0.0.0 (/home/vext01/research/yorick/ykrustc/compiler/rustc_driver)`
    ... which is depended on by `rustc-main v0.0.0 (/home/vext01/research/yorick/ykrustc/compiler/rustc)`
versions that meet the requirements `^1.0.119` are: 1.0.119

all possible versions conflict with previously selected packages.

  previously selected package `serde v1.0.118`
    ... which is depended on by `bincode v1.3.1`
    ... which is depended on by `rustc_codegen_ssa v0.0.0 (/home/vext01/research/yorick/ykrustc/compiler/rustc_codegen_ssa)`
    ... which is depended on by `rustc-main v0.0.0 (/home/vext01/research/yorick/ykrustc/compiler/rustc)`
```

That's curious because 1.0.118 should be semver compatible with 1.0.119.

If we force the serde version in Cargo.lock we see the problem:
```
error: failed to select a version for `serde_derive`.
    ... required by package `serde v1.0.119`
    ... which is depended on by `bootstrap v0.0.0 (/home/vext01/research/yorick/ykrustc/src/bootstrap)`
versions that meet the requirements `=1.0.119` are: 1.0.119
```

Notice the `=` constraint? I checked that serde-1.0.118 has a similar
constraint, and it does.

In other words, serde minor versions are incompatible.

### DIFF
--- a/ykpack/Cargo.toml
+++ b/ykpack/Cargo.toml
@@ -10,4 +10,4 @@ bincode = "1.3.1"
 bitflags = "1.2.1"
 fallible-iterator = "0.2.0"
 fxhash = "0.2.1"
-serde = { version = "1.0.119", features = ["derive"] }
+serde = { version = "1.0.118", features = ["derive"] }


ltratt:
> In other words, serde minor versions are incompatible.

I think it's a bit more subtle than that: the challenge is that Rust allows multiple versions of the same crate to be included, but that has consequences in some cases. But, whatever the cause, I'm happy with the PR.

bors r+

vext01:
> the challenge is that Rust allows multiple versions of the same crate to be included

I know you can run into that scenario, but I don't think this is it. If that were the case, `cargo` wouldn't be trying to find a unified compatible version.

But anyway, we move on. Thanks!

ltratt:
> I know you can run into that scenario, but I don't think this is it.

I'm not saying this example is caused directly by that problem, but I think it's enabled by it. Anyway, there's an element of angels-on-pins here, so I agree about moving on :)

bjorn3:
If versions are semver compatible, cargo will try to unify them. If they are not, then cargo will keep them separate. So 1.1.1 and 1.2.0 will be unified, =1.1.1 and 1.2.0 will fail as it tries to unify them, but fails. 1.1.1 and 2.0.0 will not be unified.

PR #212, opened 2021-01-15
Use MIR liveness analysis instead of our own.

ptersilie:
We previously ran our own liveness analysis to be able to remove
variables and statements that weren't relevant for the TIR trace.
However, for blackholing we now need to keep the variables and
statements around as they may be needed in the SIR. This means we can go
back to using MIRs liveness analysis to compute live variables for the
guards.

### DIFF
--- a/ykcompile/src/lib.rs
+++ b/ykcompile/src/lib.rs
@@ -512,6 +512,11 @@ impl<TT> TraceCompiler<TT> {
         self.stack_builder.alloc(ty.size(), ty.align())
     }
 
+    fn local_live(&mut self, local: &Local) {
+        // Assign a Location to this Local.
+        self.local_to_location(*local);
+    }
+
     /// Notifies the register allocator that a local has died and that its storage may be freed.
     fn local_dead(&mut self, local: &Local) -> Result<(), CompileError> {
         match self
@@ -984,6 +989,7 @@ impl<TT> TraceCompiler<TT> {
                 idx,
                 scale,
             } => self.c_dynoffs(dest, base, idx, *scale),
+            Statement::StorageLive(l) => self.local_live(l),
             Statement::StorageDead(l) => self.local_dead(l)?,
             Statement::Call(target, args, dest) => self.c_call(target, args, dest)?,
             Statement::Cast(dest, src) => self.c_cast(dest, src),
@@ -1137,6 +1143,17 @@ impl<TT> TraceCompiler<TT> {
                 Location::Reg(reg) => {
                     for c in v {
                         self.cmp_reg_const(reg, *c, SIR.ty(&val.ty()).size());
+                        dynasm!(self.asm
+                            ; je ->guardfail
+                        );
+                    }
+                }
+                Location::Mem(ro) => {
+                    dynasm!(self.asm
+                        ; mov Rq(*TEMP_REG), QWORD [Rq(ro.reg) + ro.off]
+                    );
+                    for c in v {
+                        self.cmp_reg_const(*TEMP_REG, *c, SIR.ty(&val.ty()).size());
                         dynasm!(self.asm
                             ; je ->guardfail
                         );
--- a/yktrace/src/tir.rs
+++ b/yktrace/src/tir.rs
@@ -5,8 +5,7 @@
 use super::SirTrace;
 use crate::{
     errors::InvalidTraceError,
-    sir::{self, Sir},
-    INTERP_STEP_ARG
+    sir::{self, Sir}
 };
 use std::{
     collections::{HashMap, HashSet},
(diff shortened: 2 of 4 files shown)


yktrace/src/tir.rs
```diff
@@ -49,44 +49,13 @@ impl<'a, 'm> TirTrace<'a, 'm> {
         // introducing it via `trace_locals()`.
         let mut defined_locals: HashSet<Local> = HashSet::new();
         let mut def_sites: HashMap<Local, usize> = HashMap::new();
-        let mut last_use_sites = HashMap::new();
+        let mut live_vars: HashSet<Local> = HashSet::new();
```
vext01, line 52: Potential optimisation for later? Use a bitset?

vext01:
LGTM. One small comment.

yktrace/src/tir.rs
ltratt: What's the type of `Local`? I notice `defined_locals` is also a `HashSet<Local>`, for example.

yktrace/src/tir.rs
```diff
@@ -597,7 +544,7 @@ pub struct Guard {
     /// The TIR locals that are live at the time of the guard (in addition to the trace I/O
     /// variable, $1, which is assumed to be live throughout the entirety of the trace). This is
     /// needed so that we can initialise the blackhole interpreter with the correct state.
-    pub live_locals: Vec<Local>
+    pub live_locals: Vec<(Local, Local)>
```
ltratt, line 547: Why a pair?

yktrace/src/tir.rs
ptersilie: I can do it here if you like.

ptersilie:
I also realised I missed some commits when I branched off my other branch. I've put them back. Needs squashing.

ltratt:
Can you quickly explain what those commits are? The commit descriptions are... not very descriptive ;)

yktrace/src/tir.rs
ptersilie: It's a mapping from a TIR local to its SIR equivalent. When we blackhole we need to copy the value from the TIR local to a SIR local. Due to renaming when converting from SIR to TIR, they don't always share the same name.

ptersilie:
> Can you quickly explain what those commits are? The commit descriptions are... not very descriptive ;)

They'll be squashed anyway and were supposed to go into the 1st commit. But if you insist...

yktrace/src/tir.rs
ltratt: OK, let's document that clearly (the doc comment feels a bit out of date?).

yktrace/src/tir.rs
vext01: `Local` is a wrapper around `u32`.

ltratt:
TBH, it's more for Edd's benefit than mine: he's the one who's done a thorough review (unlike my shallow review!).

yktrace/src/tir.rs
vext01: Or make a struct with named fields?

ykcompile/src/lib.rs
```diff
@@ -510,6 +510,20 @@ impl<TT> TraceCompiler<TT> {
         self.stack_builder.alloc(ty.size(), ty.align())
     }
 
+    fn local_live(&mut self, local: &Local) {
+        match self.variable_location_map.get(local) {
+            Some(Location::NotLive) => {
```
vext01, line 515: Do you need to sync? `Location::NotLive` has been removed.

yktrace/src/tir.rs
ltratt: In `yk` we might as well use https://crates.io/crates/vob which is faster for some use case than bitset -- assuming that Vob does work well here! [Using Vob inside ykrustc is something we decided against, but in an arbitrary library, we might as well use the most appropriate library.]

vext01:
A couple more comments from me.

yktrace/src/tir.rs
ltratt: > Or make a struct with named fields?

Good idea: it'll make it harder to mix the two `Locals` up.

yktrace/src/tir.rs
vext01: So make a set and then convert it to a `vob`? We don't want any duplicates.

yktrace/src/tir.rs
ptersilie: Fixed in 8ce69fe

ykcompile/src/lib.rs
ptersilie: I do.

yktrace/src/tir.rs
ltratt: A `Vob` is effectively a set, if you initialise it with a size (this does assume you know the maximum size; if you don't, then you either need a wrapper around vob, or to use bitset).

ptersilie: Up to you.

ltratt: I think you should try bitset and vob and see which works best in your opinion.

vext01: Ah yes. I was thrown by "vector" for a moment. It's just that the set is represented by a vector (of bits).

vext01: I'm not sure if we know the max size ahead of time. I leave that to Lukas to look at :)

ptersilie: I think we should know the size of all locals (though not the live ones) from the SIR. I'll have a look. The live locals are gonna be a subset of all locals though, so not sure if that would waste space when using a vob.

ltratt: If, as is likely, you have 64 or fewer locals, you won't waste any memory at all (since each local will be represented by a single bit).

yktrace/src/tir.rs
vext01: Perhaps we raced? Were we expecting a new struct? I'm not *too* fussed if not.

ptersilie: Fixing.

yktrace/src/tir.rs
```diff
@@ -5,7 +5,12 @@
 use super::SirTrace;
 use crate::{
     errors::InvalidTraceError,
+<<<<<<< HEAD
```
vext01, line 8: Not sure about this change :P

ptersilie: Hah, yeah, cocked that up, but will push the fix together with some other fixes.

yktrace/src/tir.rs
ptersilie: Fixed in f60d82c.

yktrace/src/tir.rs
```diff
@@ -523,6 +523,12 @@ impl Display for GuardBlock {
     }
 }
 
+#[derive(Debug)]
+pub struct LiveLocal {
```
vext01, line 527: Quickly add a doc comment and we are good to go.

yktrace/src/tir.rs
ptersilie: Hmm, doesn't look like we currently have a way to get the total number of used locals (though we could add the in the future). I also looked into `BitSet` which is sort-of unmaintained at the moment: https://crates.io/crates/bit-set. Thoughts?

vext01: Since this is an optimisation, I'd say don't worry about it now, or do it in another PR. I have a PR blocked on this one, so let's get it in.

ptersilie: Agreed!

ptersilie:
Ready to squash?

vext01:
Please go ahead.

ptersilie:
Squashed!

vext01:
cargo fmt

ltratt:
Feel free to force push an update with `cargo fmt` done.

ptersilie:
That should do it.

PR #214, opened 2021-01-21
Optimise Yorick's internals using workspaces.

vext01:
I could happily continue polishing and addressing FIXMEs for another few days, but I have decided to raise an initial PR now, and we can do smaller incremental PRs later.

This is a draft, as I need to check the CI config works, but we can review in the meantime.

From the docstring in `xtask/src/main.rs`:

```
This is required because we need to separately compile parts of the
codebase with different configurations.

 - Interpreter code, that we expect to be able to trace, needs to be
   compiled with `-C tracer=hw`. Crucially, since hardware tracing relies
   on blocks not being reordered by LLVM, this flag disables optimisation.

 - Anything that we don't need to trace, for example the JIT internals
   themselves, can (and should) be optimised.

Note that the above definitions also extend to dependencies, and this
means that we can have two copies of any given crate but compiled with
different flags. This is good, as it means that the dependencies of
interpreter code can be traced, whereas dependencies of the JIT
internals can be optimised.

To this end, we have 2 Rust workspaces in the repo: the "internal"
(optimised) workspace, and the "external" (unoptimised) workspace. The
external workspace then talks to the internal workspace via `extern`
functions defined in the ykshim crate.
```

Note that to build/test we now use xtask. E.g.:
```
cargo xtask build
```
We also no longer need to set `RUSTFLAGS=-C tracer=hw` as xtask will do
it automatically.

Co-authored-by: bjorn3 <bjorn3@users.noreply.github.com>


ykshim_client/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "ykshim_client"
+version = "0.1.0"
+authors = ["Edd Barrett <vext01@gmail.com>"]
```
vext01, line 4: That's wrong. Mea culpa.

ykshim_client/src/lib.rs
```diff
@@ -0,0 +1,245 @@
+//! Client to the ykshim crate in the internal workspace.
```
vext01, line 1: This file is a good place to start a review. It shows the interface to the internal workspace.

xtask/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+//! Note that the above definitions also extend to dependencies, and this means that we can have
+//! two copies of any given crate but compiled with different flags. This is good, as it means that
+//! the dependencies of interpreter code can be traced, whereas dependencies of the JIT internals
+//! can be optimised.
+//!
+//! To this end, we have 2 Rust workspaces in the repo: the "internal" (optimised) workspace, and
+//! the "external" (unoptimised) workspace. The external workspace then talks to the internal
+//! workspace via `extern` functions defined in the ykshim crate.
+//!
+//! There are a few of implementation details to note:
```
vext01, line 22: @bjorn3 would you be able to check I got these details right please?

internal_ws/ykshim/src/test_api.rs
```diff
@@ -0,0 +1,124 @@
...
+/// Returns a string describing the register allocation of the specified local.
+#[no_mangle]
+unsafe extern "C" fn __ykshim_tracecompiler_local_to_location_str(
+    tc: *mut c_void,
+    local: u32,
+) -> *mut c_char {
+    let mut tc = Box::from_raw(tc as *mut TraceCompiler);
+    let rstr = format!("{:?}", tc.local_to_location(Local(local)));
+    let cstr = CString::new(rstr.as_str()).unwrap().into_raw();
+    Box::leak(tc);
```
bjorn3, line 94: Instead of using `Box::from_raw`+`Box::leak` everywhere you should just do `&mut *(tc as *mut TraceCompiler)`.

xtask/src/main.rs
bjorn3: Looks correct.

ykshim_client/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct CguHash(u64);
+// FIMXE make this a repr(C) struct so as to simplify __ykshim_body_ret_ty.
+pub type TypeId = (CguHash, TyIndex);
+pub const RETURN_LOCAL: Local = Local(0);
+
+extern "C" {
+    // The "Production" API.
+    pub fn __ykshim_start_tracing(tracing_kind: u8) -> *mut c_void;
```
bjorn3, line 29: Do these need to be `pub`?

ykshim_client/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+}
+
+pub fn interpret_body<I>(body_name: &str, icx: &mut I) {
+    let body_cstr = CString::new(body_name).unwrap();
+    unsafe { __yktest_interpret_body(body_cstr.into_raw(), icx as *mut _ as *mut u8) };
+}
+
+pub fn reg_pool_size() -> usize {
+    unsafe { __yktest_reg_pool_size() }
+}
```
bjorn3, line 245: It may be a good idea to put the wrappers for the test api in a separate module.

internal_ws/ykshim/src/test_api.rs
vext01: Ah, because the call to these funcs (which don't consume) is inconsequential to the lifetime of the box, right?

ykshim_client/src/lib.rs
vext01: Yes, I was thinking in a follow up PR.

internal_ws/ykshim/src/test_api.rs
bjorn3: Yes

internal_ws/ykshim/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "ykshim"
+version = "0.1.0"
+authors = ["bjorn3 <bjorn3@users.noreply.github.com>"]
```
ptersilie, line 4: Should this also be `YK devs`?

internal_ws/ykshim/src/prod_api.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// Hardware tracing via ykrustc + hwtracer.
+    HardwareTracing = 1,
+}
+
+#[no_mangle]
+unsafe extern "C" fn __ykshim_start_tracing(tracing_kind: u8) -> *mut c_void {
+    let tracing_kind = match tracing_kind {
+        0 => yktrace::TracingKind::SoftwareTracing,
+        1 => yktrace::TracingKind::HardwareTracing,
+        _ => return std::ptr::null_mut(),
```
ptersilie, line 24: Why not `unreachable!()`?

internal_ws/ykshim/src/prod_api.rs
```diff
@@ -0,0 +1,90 @@
...
+            std::ptr::null_mut()
+        }
+    }
+}
+
+#[no_mangle]
+unsafe extern "C" fn __ykshim_compile_trace(
+    sir_trace: *mut c_void,
+    error_msg: *mut *mut c_char,
+) -> *mut c_void {
```
ptersilie, line 55: Can't we use `*mut CompiledTrace` here? Saves you the `as *mut c_void` below.

internal_ws/ykshim/Cargo.toml
vext01: Yes.

internal_ws/ykshim/src/prod_api.rs
```diff
@@ -0,0 +1,90 @@
...
+#[no_mangle]
+unsafe extern "C" fn __ykshim_compiled_trace_get_ptr(
+    compiled_trace: *const c_void,
+) -> *const c_void {
+    let compiled_trace = &*(compiled_trace as *mut CompiledTrace);
+    compiled_trace.ptr() as *const c_void
+}
+
+#[no_mangle]
+unsafe extern "C" fn __ykshim_compiled_trace_drop(compiled_trace: *mut c_void) {
```
ptersilie, line 83: Similarly to the comment above, this could just be `compiled_trace: *mut CompiledTrace`. Unless I'm missing the reason we are using `c_void` here.

internal_ws/ykshim/src/prod_api.rs
vext01: I suppose we could, but I find it easier to have the same function signature on both sides of the API. In the external workspace, it wouldn't be correct to use `*mut CompiledTrace`.

ptersilie: Ah, I've not seen the external workspace yet. It would save you a lot of casting code though.

tests/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "tests"
+version = "0.1.0"
+authors = ["Edd Barrett <vext01@gmail.com>"]
```
ptersilie, line 4: `Yk devs`?

ykshim_client/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+
+extern "C" {
+    // The "Production" API.
+    pub fn __ykshim_start_tracing(tracing_kind: u8) -> *mut c_void;
+    pub fn __ykshim_stop_tracing(tracer: *mut c_void, error_msg: *mut *mut c_char) -> *mut c_void;
+    pub fn __ykshim_compile_trace(
+        sir_trace: *mut c_void,
+        error_msg: *mut *mut c_char,
+    ) -> *mut c_void;
+    pub fn __ykshim_compiled_trace_get_ptr(compiled_trace: *const c_void) -> *const c_void;
```
ptersilie, line 35: Should we comment what these `c_voids` are so one doesn't need to jump to the internal ykshim to find out?

ptersilie:
Left some comments. On a general note, I feel like we need more comments on the shims. If I didn't already know what most of these functions return, I would have no idea what any of the `c_voids` are supposed to be.

bjorn3:
Maybe the `c_void` could become `extern { type Foo; }`?

internal_ws/ykshim/src/prod_api.rs
vext01: Actually, I think returning an error to the client is preferable to crashing. We can deal with this in a later PR where we look at error handling across the whole API.

ykshim_client/src/lib.rs
vext01: We could do, but I always found it obvious from context. Many of the returned pointers are immediately wrapped into a meaningful struct.

ykshim_client/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+
+impl fmt::Display for TirTrace {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let cstr = unsafe { CString::from_raw(__ykshim_tirtrace_display(self.0)) };
+        write!(f, "{}", cstr.into_string().unwrap())
+    }
+}
+
+pub struct CompiledTrace<I> {
+    compiled: *mut c_void,
```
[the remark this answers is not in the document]
vext01: Yes, but the client must not do anything with this pointer other than pass it around. So it must stay as an opaque pointer.

vext01:
> Maybe the c_void could become extern { type Foo; }?

That works for me. We'd have to prefix the name to avoid type clashes.

OK for you @ptersilie ?

vext01:
Replacing the `*mut c_void`s actually removes a lot of casts, so I'm happy to do it. I'll add docs as I go.

ltratt:
Can you provide an outline of what e.g. the `ykshim` crate is doing? [I think we should probably document that somewhere in the repo too, but we can start with comments in the PR.]

vext01:
It's explained here: https://github.com/softdevteam/yk/pull/214/files#diff-9c5e61698d6d8a8ef4a3da7075c31857769134a932ac141d4eb62f68f131aa35R1

But maybe it needs a new home?

ptersilie:
> OK for you @ptersilie ?

Sure. If it removes all those casts and makes it more obvious what is behind that pointer I'm all for it.

vext01:
I've adjusted the types and added docstrings.

After lunch, I'll check the compiler still works. I suspect it won't find ykpack yet.

xtask/src/main.rs
```diff
@@ -0,0 +1,107 @@
+//! Custom build system for the Yorick meta-tracer.
```
ltratt, line 1: I think something a bit like this probably needs to be in a top-level README: for example, without that no-one will know they need to use something other than `cargo build`. It probably needs to be a cut-down, or at least slightly different, version of this text?

ykshim_client/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+#[repr(C)]
+pub struct CguHash(u64);
+// FIMXE make this a repr(C) struct so as to simplify __ykshim_body_ret_ty.
+pub type TypeId = (CguHash, TyIndex);
+pub const RETURN_LOCAL: Local = Local(0);
+
+extern "C" {
+    // The "Production" API.
+    pub fn __ykshim_start_tracing(tracing_kind: u8) -> *mut c_void;
+    pub fn __ykshim_stop_tracing(tracer: *mut c_void, error_msg: *mut *mut c_char) -> *mut c_void;
```
bjorn3, line 30:
```suggestion
extern "C" {
    type RawThreadTracer;
    type RawSirTrace;

    // The "Production" API.
    pub fn __ykshim_start_tracing(tracing_kind: u8) -> *mut RawThreadTracer;
    pub fn __ykshim_stop_tracing(tracer: *mut RawThreadTracer, error_msg: *mut *mut c_char) -> *mut RawSirTrace;
```

With `extern { type Foo; }` I meant something like this.

xtask/src/main.rs
vext01: Arguably this description (minus implementation details) belongs in ykdocs. How about that?

vext01:
@bjorn3 the problem with extern types is that you can't assign to them.

```
error[E0277]: the size for values of type `RawThreadTracer` cannot be known at compilation time
   --> ykshim_client/src/lib.rs:95:18
    |
95  |         self.0 = ptr::null_mut() as _; // consumed.
    |                  ^^^^^^^^^^^^^ doesn't have a size known at compile-time
    | 
   ::: /home/vext01/research/yorick/ykrustc/library/core/src/ptr/mod.rs:219:23
    |
219 | pub const fn null_mut<T>() -> *mut T {
    |                       - required by this bound in `null_mut`
    |
    = help: the trait `Sized` is not implemented for `RawThreadTracer`
```

vext01:
We could use type aliases to `*mut c_void`, but we don't get any type safety from that.

ykshim_client/src/lib.rs
vext01: Good call, they do not!

vext01:
I think everyone's comments are now addressed. OK to squash?

ptersilie:
Ok by me.

README.md
```diff
@@ -8,6 +8,13 @@ This repository houses the non-compiler components of the system.
 
 The compiler repo can be found [here](https://github.com/softdevteam/ykrustc).
 
+## Getting Started
+
+To get started, see
+[this section in the documentation](https://softdevteam.github.io/ykdocs/dev/getting_started.html).
+
+Note that you must use `cargo xtask` commands when dealing with this repo.
```
ltratt, line 16: Can we say something like "`yk` uses `cargo-xtask` instead of normal cargo. You will need to have [xtask](https://github.com/matklad/cargo-xtask/) installed (e.g. with `cargo install cargo-xtask`). Instead of `cargo x` (for some command `x`), you will then need to execute `cargo xtask x`." here?

vext01: Try 121bcc5.

We don't have to install anything, it all works magically with a cargo alias.

ltratt: How does `xtask` get installed? It's not part of cargo by default.

ltratt: [Apart from that, 121bcc5 looks fine to me!]

bjorn3: `xtask` is an alias defined in `.cargo/config`: https://github.com/softdevteam/yk/pull/214/files#diff-f6009bd0d260464389ace37ab2f89adae993e1fa4a47f779e4c9859937005cedR2

vext01: Should we mention that you don't have to install anything to dispel any potential confusion?

ltratt: OK, so the answer to my question is that `cargo run --package` installs xtask. We don't need to document that -- I just worried that you guys had installed xtask and forgotten that you'd done so!

vext01: > cargo run --package installs xtask

It runs the `xtask` package from our workspace, but nothing is installed. Essentially, [this code is executed](https://github.com/vext01/yk_pv/blob/121bcc5691b0fd20d0e7c4cdc6faf3222a1c868e/xtask/src/main.rs#L13).

ltratt:
Please squash.

vext01:
Splat.

vext01:
OK, looks like I didn't test the CI script changes enough. I'll do a forced build behind the scenes and come back when I've re-tested.

vext01:
CI fixed. OK to squash again?

ltratt:
Please squash.

vext01:
splat.

PR #218, opened 2021-01-28
Implement blackholing.

ptersilie:
This PR implements "blackholing", i.e. getting from a guard failure in a compiled trace back to the normal interpreter. This is done using a special interpreter, the SIRInterpreter, which runs on SIR blocks. When a guard fails, we check which block in the SIR execution needs to continue. We then initialise the SIRInterpreter at that point, copying over all live values computed during execution of the compiled trace, and then interpret SIR blocks until we get back to the control point (which happens when we return from the `interp_step`-annotated function). For this each guard stores information about which variables are live up to that point. Guards also need to store information about function calls (even though in the compiled trace those are inlined), so that we can recreate the needed stack frames in the SIRInterpreter.

Let me apologise in advance for this PR. I tried splitting things up into multiple commits, but this wasn't possible for the main change (getting the guards to produce the necessary information for blackholing). This the first commit is a lot to take in. The heart of this PR is in the `ret` function (https://github.com/softdevteam/yk/pull/218/commits/956d84e560ed3907812460d2f0db4008e8321398#diff-a0588be717e433bc24ac399f2b57b188c7b13b56c5a512d4c1404d9fac87bf04R1339), which for each guard generates code that allocates memory for the stack frames and fills it with the live variables, and finally instantiates a `SIRInterpreter`. So I suggest you start with that.

If anything isn't clear, please feel free to leave comments asking for clarification and I'll add and commit more comments.

The second commit needs squashing into the first, too, but I've left them apart for now to make reviewing easier.

### DIFF
--- a/internal_ws/ykbh/src/lib.rs
+++ b/internal_ws/ykbh/src/lib.rs
@@ -1,11 +1,21 @@
 use std::alloc::{alloc, dealloc, Layout};
 use std::convert::TryFrom;
-use std::sync::Arc;
 use ykpack::{
-    self, Body, BodyFlags, CallOperand, Constant, ConstantInt, IPlace, Local, Statement,
-    Terminator, UnsignedInt,
+    self, CallOperand, Constant, ConstantInt, IPlace, Local, Statement, Terminator, TyKind,
+    UnsignedInt, UnsignedIntTy,
 };
 use yktrace::sir::SIR;
+
+/// Stores information needed to recreate stack frames in the SIRInterpreter.
+pub struct FrameInfo {
+    /// The symbol name of this frame.
+    pub sym: String,
+    /// Index of the current basic block we are in. When returning from a function call, the
+    /// terminator of this block is were we continue.
+    pub bbidx: usize,
+    /// Pointer to memory containing the live variables.
+    pub mem: *mut u8,
+}
 
 /// A stack frame for writing and reading locals. Note that the allocated memory this frame points
 /// to needs to be freed manually before the stack frame is destoyed.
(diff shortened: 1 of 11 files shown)


ltratt:
A dumb, quick question: is there just one new test? If so, I really think we need to have more tests, even if they're marked as failure at the moment. For example, if we can't blackhole in a recursive function call (I have no idea if this is the case or not!), there's value in having such a test case and being clear that it's not supported. But, obviously, I'm even keener on test cases which contain different features that pass :)

ptersilie:
You're right. I will try and add more tests. I was maybe a bit too keen to get this out, as Edd's PRs are blocked in the mean time, and this PR is already difficult enough to review.

vext01:
I'm OK with having `#[ignore]` tests in the short term too.

vext01:
[BTW, in the call this morning, our head honcho said that he insists the tests are done in this PR]

ptersilie:
> [BTW, in the call this morning, our head honcho said that he insists the tests are done in this PR]

Hah, I'm already working on them. That's why I haven't been able to address any of your comments yet.

@headhoncho (a.k.a. @ltratt): The recursion test was a good call, as it has exposed a bug in the guards.

ptersilie:
I've addressed all comments. There's also two new commits with tests (and fixes) in b5605c8 and 1898e5f that need reviewing.

vext01:
This is very close. Just a few comments to go.

ptersilie:
This last commit should fix your remaining comments. Ready for re-review.

vext01:
That's much cleaner.

So now we have to resolve conflicts, and we can go!

ptersilie:
I think I will try squashing first. Okay to do both in one go?

vext01:
Yup.

ptersilie:
Okay, that's squashing done. Now the rebase.

ptersilie:
Done. This should be ready for merging now :crossed_fingers:

PR #219, opened 2021-01-29
Remove duplication in test

vext01:
Fix some memory safety issues, and remove duplication of compilation from a test.

This shouldn't get in the way of the black-hole work.

### DIFF
--- a/ykshim_client/src/test_api.rs
+++ b/ykshim_client/src/test_api.rs
@@ -1,6 +1,8 @@
 //! The testing API client to ykshim.
 
-use crate::prod_api::{CompiledTrace, Local, RawCompiledTrace, RawSirTrace, SirTrace, TyIndex};
+use crate::prod_api::{
+    CompiledTrace, Local, RawCompiledTrace, RawSirTrace, RawTirTrace, SirTrace, TyIndex,
+};
 use libc::size_t;
 use std::collections::HashMap;
 use std::ffi::{c_void, CString};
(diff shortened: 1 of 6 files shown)


ykshim_client/src/test_api.rs
```diff
@@ -136,3 +139,12 @@ pub fn interpret_body<I>(body_name: &str, icx: &mut I) {
 pub fn reg_pool_size() -> usize {
     unsafe { __ykshimtest_reg_pool_size() }
 }
+
+pub fn compile_tir_trace<T>(mut tir_trace: TirTrace) -> Result<CompiledTrace<T>, CString> {
+    let compiled = unsafe { __ykshimtest_compile_tir_trace(tir_trace.0) };
+    tir_trace.0 = ptr::null_mut(); // consumed.
```
ptersilie, line 155: Should we expand on this comment a bit. Something like: Since `tir_trace.0` is consumed in `__ykshimtest_compile_tir_trace`, we can set it's pointer to `null`.

ptersilie:
One comment. Rest looks fine.

ykshim_client/src/test_api.rs
vext01: This is an idiom that appears a few times in ykshim. I considered adding a comment to the crate, but it's just standard memory management at the end of the day. Thoughts?

ptersilie: Fine. I'm not too bothered by this. Just thought it might avoid some confusion in the future if the comment says a bit more than consumed. That being said, I guess I immediately knew what was going on despite not having seen this code before. So maybe it's fine as it is.

vext01: If you are happy with this, then please kick bors.

PR #223, opened 2021-02-05
Use underscores more idiomatically.

ltratt:
It's a bit odd to see variables starting with underscores outside of method parameters.

### DIFF
--- a/internal_ws/ykbh/src/lib.rs
+++ b/internal_ws/ykbh/src/lib.rs
@@ -53,7 +53,7 @@ impl LocalMem {
                     }
                     _ => todo!(),
                 },
-                ConstantInt::SignedInt(_si) => todo!(),
+                ConstantInt::SignedInt(_) => todo!(),
             },
             Constant::Bool(b) => self.write_val(dest, [*b as u8].as_ptr(), 1),
             Constant::Tuple(t) => {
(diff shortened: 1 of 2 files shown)


internal_ws/ykcompile/src/lib.rs
```diff
@@ -633,13 +633,6 @@ impl TraceCompiler {
         self.restore_regs(&*CALLER_SAVED_REGS);
     }
 
-    /// Emit a NOP operation.
```
vext01, line 636: This a debugging aid. It can be useful to insert `nop`s as waypoints when starting at traces.

internal_ws/ykbh/src/lib.rs
```diff
@@ -53,7 +53,7 @@ impl LocalMem {
                     }
                     _ => todo!(),
                 },
-                ConstantInt::SignedInt(_si) => todo!(),
```
vext01, line 56: Whilst I agree with this change this time, it can be useful to name unused fields, especially if there are lots of them.

internal_ws/ykcompile/src/lib.rs
ptersilie: While this is true, I just tend to put the `nop`s directly into the other dynasm calls. So I think we can remove this.

ptersilie:
Is this ready for bors?

vext01:
> It's a bit odd to see variables starting with underscores outside of method parameters.

I don't find it all that odd FWIW. The [rust by example book](https://doc.rust-lang.org/rust-by-example/variable_bindings.html) even shows you how to do it:
```
    // The compiler warns about unused variable bindings; these warnings can
    // be silenced by prefixing the variable name with an underscore
    let _unused_variable = 3u32;
```

vext01:
@ptersilie ready if you are.

PR #226, opened 2021-02-05
Random comments

ltratt:
**Do not merge this ever, under any circumstances -- this is not a normal PR!!!**

This is not a PR intended for merge: it's a complete repo code review. Inevitably it's shallow, but it's still thrown up some interesting things. Overall, I'm pretty happy with the code, but we can always improve things :)

There's a mix of things in here. Sometimes I've rephrased a comment; sometimes I've asked a question; less often I have a code change suggestion. So sometimes you can probably just copy an edited thing into your own PR; but sometimes we'll have to have a debate here to decide the way forwards.

What I suggest is that @ptersilie and @vext01 go through each bit of the diff and we use the nomal GitHub comments to decide on things. Then we can open up one or more PRs that address the valid comments in this PR.

Some general comments, in no particular order:

1. We use hard-coded constants a bit more often than I'd like (e.g. `Local(0)` for the interpreter context). We should be careful to weed as many of these out as possible (I won't have found all of them!) and make sure we don't introduce any more ni future.
2. There were some completely redundant comments of the form "// Does x" for `x`. Let's kill those.
3. A few functions and structs could do with documenting and/or better documenting.
4. Similarly, I think we should expect all non-test files to have a doc-comment at the top to explain a bit better what the file's contents are.
5. We've got a curious mix of really-well commented code and sparsely/poorly commented code. I think we need to be a bit more ruthless in future code review at ensuring that complex things have good comments. That's both a quantity (sometimes we just don't say enough!) and quality issue (we have quite a few comments that I could make little sense of).

Overall, the aim of this is to make our future selves happier at what our current selves did :) Bear in mind that I'm not familiar with large chunks of the repo, and so some of my comments will be inane, incorrect, or perhaps plain bonkers.

### DIFF
--- a/internal_ws/ykbh/src/lib.rs
+++ b/internal_ws/ykbh/src/lib.rs
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
+
 use std::alloc::{alloc, dealloc, Layout};
 use std::convert::TryFrom;
 use ykpack::{
@@ -9,6 +11,8 @@ use yktrace::sir::SIR;
 /// Stores information needed to recreate stack frames in the SIRInterpreter.
 pub struct FrameInfo {
     /// The symbol name of this frame.
+    /// i might have misunderstood this, but does every function call in the stopgap interpreter
+    /// require mallocing a String?
     pub sym: String,
     /// Index of the current basic block we are in. When returning from a function call, the
     /// terminator of this block is were we continue.
@@ -199,6 +203,8 @@ impl SIRInterpreter {
     /// Run the SIR interpreter after it has been initialised by a guard failure. Since we start in
     /// the block where the guard failed, we immediately skip to the terminator and interpret it to
     /// see which block we need to start interpretation in.
+    /// XXX we have one function called "interpret" and another "_interpret" -- both public! At
+    /// least one of them needs a better name.
     pub unsafe fn interpret(&mut self, ctx: *mut u8) {
         self.set_interp_ctx(ctx);
         // Jump to the correct basic block by interpreting the terminator.
@@ -272,7 +278,6 @@ impl SIRInterpreter {
             Terminator::Call {
                 operand: op,
                 args,
-                destination: _dest,
             } => {
                 let fname = if let CallOperand::Fn(sym) = op {
                     sym
@@ -286,8 +291,8 @@ impl SIRInterpreter {
                 self.frames.push(frame);
             }
             Terminator::Return => {
-                // Return from current stack frame.
                 let oldframe = self.frames.pop().unwrap();
+                // XXX comment could do with rephrasing
                 // Are we still inside a nested call? Otherwise we are returning from the first
                 // body, so we are done interpreting.
                 if let Some(curframe) = self.frames.last_mut() {
@@ -304,7 +309,8 @@ impl SIRInterpreter {
                         _ => unreachable!(),
                     };
                     // Get a pointer to the return value of the called frame. XXX this needs to be
-                    // a named constant
+                    // a named constant. do a grep over the repo for `Local.[0-9]` to see more of
+                    // these
                     let ret_ptr = oldframe.mem.local_ptr(&Local(0));
                     // Write the return value to the destination in the previous frame.
                     let dst_ptr = curframe.mem.iplace_to_ptr(&dest);
@@ -353,11 +359,14 @@ impl SIRInterpreter {
         }
     }
 
+    /// XXX unhelpful comment
     /// Implements the Store statement.
     fn store(&mut self, dest: &IPlace, src: &IPlace) {
         self.frames.last_mut().unwrap().mem.store(dest, src);
     }
 
+    /// XXX this might be me being an idiot, but i have no idea what this comment means or, indeed,
+    /// what the function actually does!
     /// Creates a reference to an IPlace.
     fn mkref(&mut self, dest: &IPlace, src: &IPlace) {
         match dest {
--- a/internal_ws/ykbh/src/lib.rs
+++ b/internal_ws/ykbh/src/lib.rs
@@ -139,30 +139,34 @@ impl LocalMem {
     }
 }
 
-/// A interpreter stack frame, containing allocated memory for the frames locals, and the function
+/// An interpreter stack frame, containing allocated memory for the frames locals, and the function
 /// symbol name and basic block index needed by the interpreter to continue interpreting after
 /// returning from a function call.
 struct StackFrame {
     /// Allocated memory holding live locals.
     mem: LocalMem,
-    /// The current basic block index of this frame. Upon returning from a function call it is used
+    /// The current basic block index of this frame. XXX i don't understand this sentence. Upon returning from a function call it is used
     /// to look up the previous basic block and check its terminator to decide where to continue
     /// interpreting.
     bbidx: ykpack::BasicBlockIndex,
-    /// Symbol name of this stack frame. Needed to retrieve the SIR body of the function which
+    /// woah, we malloc a `String` per stack frame?! that's going to be very slow. we need an alternative mechanism here. Symbol name of this stack frame. Needed to retrieve the SIR body of the function which
     /// contains the statements we want to interpret.
     func: String,
 }
 
-/// The SIR interpreter, also known as blackholing interpreter, is invoked when a guard fails in a
-/// trace. It is initalised with information from the trace, e.g. live variables, stack frames, and
-/// then run to get us back to a control point from where the normal interpreter can take over.
+/// The SIR interpreter, also known as the blackhole interpreter, is invoked when a guard fails in
+/// a trace. It is initialised with information from the trace (e.g. live variables, stack frames)
+/// and then runs until we are back to the control point, from where the normal interpreter can
+/// take over.
 pub struct SIRInterpreter {
-    /// Keeps track of active stack frames (most recent last).
+    /// Active stack frames (most recent last).
     frames: Vec<StackFrame>,
 }
 
 impl SIRInterpreter {
+    /// XXX i think this test is only needed for tests? if so, we should at least #[cfg(test)] it,
+    /// and probably give it a name that doesn't suggests it's the main way of initialising the
+    /// interpreter.
     pub fn new(sym: String) -> Self {
         let frame = SIRInterpreter::create_frame(&sym);
         SIRInterpreter {
@@ -170,6 +174,7 @@ impl SIRInterpreter {
         }
     }
 
+    /// XXX I think this might better be called `from_frames`?
     /// Initialises the interpreter with information about live variables and stack frames,
     /// received from the failing guard.
     pub fn init_frames(v: Vec<FrameInfo>) -> Self {
@@ -205,13 +210,13 @@ impl SIRInterpreter {
         self._interpret();
     }
 
-    /// Given a vector of local declarations allocate just enough space to hold all of them.
+    /// Given a vector of local declarations allocate just enough XXX "just enough" or "precisely
+    /// the right amount"? space to hold all of them.
     fn create_frame(sym: &String) -> StackFrame {
         let body = SIR.body(&sym).unwrap();
         let (size, align) = body.layout;
         let offsets = body.offsets.clone();
         let layout = Layout::from_size_align(size, align).unwrap();
-        // Allocate memory for the locals
         let locals = unsafe { alloc(layout) };
         let mem = LocalMem {
             locals,
@@ -235,11 +240,11 @@ impl SIRInterpreter {
         // The interpreter context lives in $1
         let ptr = self.frames.first().unwrap().mem.local_ptr(&Local(1));
         unsafe {
-            // Write the pointer value of `tio` into this frames memory.
             std::ptr::write::<*mut u8>(ptr as *mut *mut u8, ctx);
         }
     }
 
+    /// XXX this function shouldn't be prefixed with a `_`
     pub unsafe fn _interpret(&mut self) {
         while let Some(frame) = self.frames.last() {
             let body = SIR.body(&frame.func).unwrap();
@@ -281,7 +286,7 @@ impl SIRInterpreter {
                 self.frames.push(frame);
             }
             Terminator::Return => {
-                // Return from current stackframe.
+                // Return from current stack frame.
                 let oldframe = self.frames.pop().unwrap();
                 // Are we still inside a nested call? Otherwise we are returning from the first
                 // body, so we are done interpreting.
@@ -298,7 +303,8 @@ impl SIRInterpreter {
                         } => dest.as_ref().map(|(p, b)| (p.clone(), *b)).unwrap(),
                         _ => unreachable!(),
                     };
-                    // Get a pointer to the return value of the called frame.
+                    // Get a pointer to the return value of the called frame. XXX this needs to be
+                    // a named constant
                     let ret_ptr = oldframe.mem.local_ptr(&Local(0));
                     // Write the return value to the destination in the previous frame.
                     let dst_ptr = curframe.mem.iplace_to_ptr(&dest);
--- a/internal_ws/ykcompile/src/lib.rs
+++ b/internal_ws/ykcompile/src/lib.rs
@@ -264,7 +264,9 @@ fn local_to_reg_name(loc: &Location) -> &'static str {
 // interpreter for blackholing.
 
 /// Given a vector of `FrameInfo`s `vptr`, instantiates and initialises a SIR interpreter and returns its
-/// pointer. Consumes `vptr`.
+/// pointer. Consumes `vptr`. XXX note this requires a move of `vptr`, so we should document
+/// something like "note that the calling function must fully transfer ownership of vptr
+/// to this function".
 #[no_mangle]
 pub extern "sysv64" fn invoke_sinterp(vptr: *mut Vec<FrameInfo>) -> *mut SIRInterpreter {
     let v = unsafe { Box::from_raw(vptr) };
--- a/internal_ws/ykcompile/src/lib.rs
+++ b/internal_ws/ykcompile/src/lib.rs
@@ -290,7 +290,7 @@ pub extern "sysv64" fn bh_new_vec() -> *mut Vec<FrameInfo> {
 
 /// Pushes a new `FrameInfo` instance onto the vector behind the pointer `vptr`. The `FrameInfo`
 /// instance is creates from a pointer to a symbol name and its length, a basic block index and a
-/// pointer to some allocated memory.
+/// pointer to some allocated memory. XXX need to document that `vptr` is consumed
 pub extern "sysv64" fn bh_push_vec(
     vptr: *mut Vec<FrameInfo>,
     sym_ptr: *const u8,
@@ -310,7 +310,7 @@ pub extern "sysv64" fn bh_push_vec(
     Box::into_raw(v);
 }
 
-/// Compile a TIR trace, returning executable code.
+/// Compile a TIR trace.
 pub fn compile_trace(tt: TirTrace) -> CompiledTrace {
     CompiledTrace {
         mc: TraceCompiler::_compile(tt, false),
@@ -410,7 +410,7 @@ impl Location {
 
     /// Apply an offset to the location, returning a new one.
     fn offset(self, off: OffT) -> Self {
-        if off == 0 {
+        if off == 0 { // Is the special case a performance hack or ... ?
             return self;
         }
         match self {
@@ -454,7 +454,7 @@ pub struct TraceCompiler {
     register_content_map: HashMap<u8, RegAlloc>,
     /// Maps trace locals to their location (register, stack).
     variable_location_map: HashMap<Local, Location>,
-    /// Local decls of the tir trace.
+    /// Local decls of the TIR trace.
     pub local_decls: HashMap<Local, LocalDecl>,
     /// Stack builder for allocating objects on the stack.
     stack_builder: StackBuilder,
@@ -566,6 +566,8 @@ impl TraceCompiler {
         self.stack_builder.alloc(ty.size(), ty.align())
     }
 
+    /// XXX Doc comment needed and maybe a better function name? I'm not sure if it means "make the
+    /// local come alive" or "check the local is alive"
     fn local_live(&mut self, local: &Local) {
         // Assign a Location to this Local.
         self.local_to_location(*local);
@@ -597,7 +599,8 @@ impl TraceCompiler {
     /// Copy bytes from one memory location to another.
     fn copy_memory(&mut self, dest: &RegAndOffset, src: &RegAndOffset, size: u64) {
         // We use memmove(3), as it's not clear if MIR (and therefore SIR) could cause copies
-        // involving overlapping buffers.
+        // involving overlapping buffers. XXX this sounds like a big thing not to be sure about!
+        // what's the underlying problem?
         let sym = Self::find_symbol("memmove").unwrap();
         self.save_regs(&*CALLER_SAVED_REGS);
         dynasm!(self.asm
@@ -642,7 +645,7 @@ impl TraceCompiler {
         );
     }
 
-    /// Push the specified registers to the stack in order.
+    /// Push the specified registers to the stack, in the order they are specified in the array.
     fn save_regs(&mut self, regs: &[u8]) {
         for reg in regs.iter() {
             dynasm!(self.asm
@@ -650,7 +653,9 @@ impl TraceCompiler {
             );
         }
     }
-    /// Pop the specified registers from the stack in reverse order.
+
+    /// Pop the specified registers from the stack in reverse order to how they are specified in
+    /// the array.
     fn restore_regs(&mut self, regs: &[u8]) {
         for reg in regs.iter().rev() {
             dynasm!(self.asm
@@ -1058,7 +1063,7 @@ impl TraceCompiler {
         let src_loc = self.iplace_to_location(src);
         match src_loc {
             Location::Reg(..) => {
-                // This isn't possible as the allocator explicitely puts things which are
+                // This isn't possible as the allocator explicitly puts things which are
                 // referenced onto the stack and never in registers.
                 unreachable!()
             }
@@ -1095,7 +1100,7 @@ impl TraceCompiler {
     fn c_cast(&mut self, dest: &IPlace, src: &IPlace) {
         let src_loc = self.iplace_to_location(src);
         let ty = &*SIR.ty(&src.ty()); // Type of the source.
-        let cty = SIR.ty(&dest.ty()); // Type of the cast (same as dest type).
+        let cty = SIR.ty(&dest.ty()); // Type of the cast (same as dest type XXX isn't this the case by definition?).
         match ty.kind {
             TyKind::UnsignedInt(_) => self.c_cast_uint(src_loc, &ty, &cty),
             _ => todo!(),
@@ -1187,7 +1192,8 @@ impl TraceCompiler {
     /// Compile a guard in the trace, emitting code to abort execution in case the guard fails.
     fn c_guard(&mut self, guard: &Guard, dl: DynamicLabel) {
         // FIXME some of the terminators from which we build these guards can have cleanup blocks.
-        // Currently we don't run any cleanup, but should we?
+        // Currently we don't run any cleanup, but should we? XXX does this mean we don't call
+        // drop? And, if so, what do we not call Drop on?
         match guard {
             Guard {
                 val,
@@ -1370,7 +1376,7 @@ impl TraceCompiler {
         // Reserved stack space for spilled locals during execution.
         let soff = self.stack_builder.size();
         // Reserved memory on the stack to spill live locals to during a guard failure.
-        let live_off = REG_POOL.len() * 8;
+        let live_off = REG_POOL.len() * 8; // Is `8` a machine word? Prefer a calculation/constant name.
 
         // As we'll be making calls to other functions using the Sys-V 64 ABI, we need to make sure
         // the stack is aligned to 16 bytes before the call instruction. Note that the call
@@ -1428,6 +1434,8 @@ impl TraceCompiler {
             for (tirlocal, loc) in live_locations.iter_mut() {
                 match loc {
                     Location::Reg(reg) => {
+                        // XXX I think that "-i32" is potentially dangerous because 2's offset is
+                        // not symmetric. We might need another check for that here?
                         let off = -i32::try_from(soff + u32::from(*reg) * 8).unwrap();
                         let newloc = Location::Mem(RegAndOffset {
                             reg: RBP.code(),
@@ -1442,7 +1450,7 @@ impl TraceCompiler {
                 }
             }
 
-            // Create vector to store the allocated memory pointers in, to be handed to the
+            // Create a vector "of what?" to store the allocated memory pointers in, to be handed to the
             // SIRInterpreter later. We can use a fixed register here, since we've spilled all
             // registers to the stack.
             let frame_vec_reg = R12.code();
@@ -1564,7 +1572,7 @@ impl TraceCompiler {
                 let ptr = buf.ptr(dynasmrt::AssemblyOffset(0)) as *mut libc::c_void;
                 let len = buf.len();
                 let alignment = ptr as usize % libc::sysconf(libc::_SC_PAGESIZE) as usize;
-                let ptr = ptr.offset(-(alignment as isize));
+                let ptr = ptr.offset(-(alignment as isize)); // probably better as ptr.pub?
                 let len = len + alignment;
                 libc::mprotect(ptr, len, libc::PROT_EXEC | libc::PROT_WRITE);
             }
--- a/internal_ws/ykcompile/src/stack_builder.rs
+++ b/internal_ws/ykcompile/src/stack_builder.rs
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
+
 use crate::Location;
 use dynasmrt::{x64::Rq::RBP, Register};
 use std::convert::{TryFrom, TryInto};
@@ -10,10 +12,10 @@ pub(crate) struct StackBuilder {
 
 /// A naive stack builder for allocating objects on the stack.
 ///
-/// Naive because it could allocate less space by caching up the requested allocations and
+/// XXX I can't parse this sentence. Naive because it could allocate less space by caching up the requested allocations and
 /// better-packing them.
 ///
-/// The top of the stack, before any calls to `alloc()` is assumed to be appropriately aligned.
+/// XXX or this sentence. The top of the stack, before any calls to `alloc()` is assumed to be appropriately aligned.
 impl StackBuilder {
     /// Allocate an object of given size and alignment on the stack, returning a `Location::Mem`
     /// describing the position of the allocation. The stack is assumed to grow down.
--- a/internal_ws/ykpack/src/decode.rs
+++ b/internal_ws/ykpack/src/decode.rs
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
+
 use crate::Pack;
 use fallible_iterator::FallibleIterator;
 use std::io::Read;
--- a/internal_ws/ykpack/src/encode.rs
+++ b/internal_ws/ykpack/src/encode.rs
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
+
 use crate::Pack;
 
 /// The pack encoder.
--- a/internal_ws/ykpack/src/types.rs
+++ b/internal_ws/ykpack/src/types.rs
@@ -132,6 +132,9 @@ impl Ty {
         }
     }
 
+    // XXX this function feels a bit evil. surely it should be the caller's job to do this? that
+    // way the need to panic will either a) be clear b) not be possible because the caller has
+    // verified things?
     pub fn unwrap_tuple(&self) -> &TupleTy {
         if let TyKind::Tuple(tty) = &self.kind {
             &tty
@@ -362,12 +365,12 @@ impl Display for Ptr {
     }
 }
 
-/// An IR place. This is used in SIR and TIR to describe the (abstract) address of a piece of data.
+/// An IR place XXX why isn't the struct called "IRPlace"? [I've been wondering for ages what "IPlace" meant! "IRPlace" would be a bit clearer I think.]. This is used in SIR and TIR to describe the (abstract) address of a piece of data.
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
 pub enum IPlace {
     /// The IPlace describes a value as a Local+offset pair.
     Val { local: Local, off: OffT, ty: TypeId },
-    /// An indirect place behind a pointer. ykrustc uses these for deref and (dynamic) index
+    /// An indirect place behind a pointer XXX huh? i'm not sure what that means. ykrustc uses these for deref and (dynamic) index
     /// projections (which cannot be resolved statically and thus depend on a runtime pointer).
     Indirect {
         /// The location of the pointer to be dereferenced at runtime.
@@ -446,6 +449,8 @@ impl IPlace {
                     ty: new_ty,
                 }
             }
+            // are these unreachable or todos? if the former, the strings are unnecessary (and just
+            // lead to a bit of pointless binary bloat)
             Self::Const { .. } => unreachable!("const to indirect"),
             Self::Indirect { .. } => unreachable!("indirect to indirect"),
             Self::Unimplemented(_) => self.clone(),
@@ -467,7 +472,7 @@ pub enum Statement {
         opnd2: IPlace,
         checked: bool,
     },
-    /// Makes a reference.
+    /// Makes a reference. XXX from what to what?
     MkRef(IPlace, IPlace),
     /// Computes a pointer address at runtime.
     DynOffs {
@@ -952,6 +957,7 @@ mod tests {
     }
 }
 
+/// XXX this should be above the tests somewhere!
 /// A SIR mapping label.
 #[derive(Debug, Hash, Eq, PartialEq, Clone, Serialize, Deserialize)]
 pub struct SirLabel {
--- a/internal_ws/ykshim/src/prod_api.rs
+++ b/internal_ws/ykshim/src/prod_api.rs
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
+
 use std::ffi::{c_void, CString};
 use std::os::raw::c_char;
 
--- a/internal_ws/yktrace/src/errors.rs
+++ b/internal_ws/yktrace/src/errors.rs
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
+
 use std::fmt::{self, Display, Formatter};
 
 #[derive(Debug)]
--- a/internal_ws/yktrace/src/hwt/mapper.rs
+++ b/internal_ws/yktrace/src/hwt/mapper.rs
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
+
 use object::{Object, ObjectSection};
 use phdrs::objects;
 
@@ -73,6 +75,7 @@ impl HWTMapper {
 
 /// Extract the program header offset. This offset can be used to translate the address of a trace
 /// block to a program address, allowing us to find the correct SIR location.
+/// XXX if this is an address should it be usize?
 fn get_phdr_offset() -> u64 {
     (&objects()[0]).addr() as u64
 }
--- a/internal_ws/yktrace/src/lib.rs
+++ b/internal_ws/yktrace/src/lib.rs
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
+
 #![cfg_attr(test, feature(test))]
 #![cfg_attr(tracermode = "sw", feature(thread_local))]
 #![cfg_attr(tracermode = "sw", feature(core_intrinsics))]
@@ -23,6 +25,7 @@ use sir::SirTrace;
 use ykpack::Local;
 
 // In TIR traces, the argument to the interp_step is always local #1.
+// XXX this constant could be used in quite a few places throughout the repo!
 pub const INTERP_STEP_ARG: Local = Local(1);
 
 /// The different ways by which we can collect a trace.
--- a/internal_ws/yktrace/src/swt.rs
+++ b/internal_ws/yktrace/src/swt.rs
@@ -77,8 +77,11 @@ static mut TRACE_BUF_CAP: usize = 0;
 static mut __YK_SWT_ACTIVE: bool = false;
 
 // =======================
-// The following code may not call any non `#[do_not_trace]` function to prevent an infinite
-// recursion of `__yk_swt_rec_loc` calling a function calling `__yk_swt_rec_loc`.
+// XXX can we scope "following code"? does it mean "any of the rest of the code in this file?" If
+// so, it might be worth either a) putting this dangerous code in its own file with the warning at
+// the top b) attaching the warning to each function. perhaps the latter is the safest thing to do!
+// The following code must not call any non `#[do_not_trace]` function: if it does, an infinite
+// loop will occur where `__yk_swt_rec_loc` calls `__yk_swt_rec_loc`.
 
 /// Start tracing on the current thread.
 /// A new trace buffer is allocated and MIR locations will be written into it on
--- a/internal_ws/yktrace/src/tir.rs
+++ b/internal_ws/yktrace/src/tir.rs
@@ -82,7 +82,7 @@ impl<'a, 'm> TirTrace<'a, 'm> {
             }
 
             // If a function was annotated with `do_not_trace`, skip all instructions within it as
-            // well. FIXME: recursion.
+            // well. FIXME: recursion. what does "recursion" mean?!
             if let Some(sym) = &ignore {
                 if sym == &loc.symbol_name
                     && body.blocks[user_bb_idx_usize].term == Terminator::Return
@@ -229,9 +229,9 @@ impl<'a, 'm> TirTrace<'a, 'm> {
                     // Rename the return value.
                     //
                     // FIXME It seems that calls always have a destination despite the field being
-                    // `Option`. If this is not always the case, we may want add the `Local` offset
-                    // (`var_len`) to this statement so we can assign the arguments to the correct
-                    // `Local`s during trace compilation.
+                    // `Option`. If this is not always the case, we may want to add the `Local`
+                    // offset (`var_len`) to this statement so we can assign the arguments to the
+                    // correct `Local`s during trace compilation.
                     let ret_val = dest
                         .as_ref()
                         .map(|(ret_val, _)| rnm.rename_iplace(&ret_val, &body))
@@ -280,7 +280,7 @@ impl<'a, 'm> TirTrace<'a, 'm> {
                             term_stmts.push(Statement::Call(op.clone(), newargs, Some(ret_val)))
                         }
                     } else {
-                        todo!("Unknown callee encountered");
+                        todo!();
                     }
                 }
                 Terminator::Return => {
@@ -335,7 +335,7 @@ impl<'a, 'm> TirTrace<'a, 'm> {
                     // Peek at the next block in the trace to see which outgoing edge was taken and
                     // infer which value we must guard upon. We are working on the assumption that
                     // a trace can't end on a SwitchInt. i.e. that another block follows.
-                    let next_blk = itr.peek().expect("no block to peek at").bb_idx;
+                    let next_blk = itr.peek().unwrap().bb_idx;
                     let edge_idx = target_bbs.iter().position(|e| *e == next_blk);
                     match edge_idx {
                         Some(idx) => Some(Guard {
@@ -461,7 +461,7 @@ impl VarRenamer {
         self.offset = self.acc.unwrap();
         self.stack.push(self.offset);
         if let Some(v) = self.acc.as_mut() {
-            *v += num_locals as u32;
+            *v += num_locals as u32; // u32::try_from?
         }
     }
 
@@ -472,6 +472,11 @@ impl VarRenamer {
         if let Some(v) = self.stack.last() {
             self.offset = *v;
         } else {
+            // presumably this fundamentally can't happen? at which point a `debug_assert` might be
+            // more appropriate so that the function is:
+            //   self.stack.pop();
+            //   debug_assert!(!self.stack.is_empty());
+            //   self.offset = self.stack.last().unwrap();
             panic!("Unbalanced enter/leave statements!")
         }
     }
--- a/internal_ws/ykview/src/main.rs
+++ b/internal_ws/ykview/src/main.rs
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
+
 use memmap2::Mmap;
 use std::{fs::File, io::Read};
 
--- a/tests/src/helpers.rs
+++ b/tests/src/helpers.rs
@@ -1,3 +1,6 @@
+// XXX does this file need to be on its own? also, "helpers" is the most unhelpful module name in
+// the history of module names ;)
+
 use fm::FMBuilder;
 use regex::Regex;
 use ykshim_client::TirTrace;
--- a/ykrt/src/lib.rs
+++ b/ykrt/src/lib.rs
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
+
 #![cfg_attr(test, feature(test))]
 
 mod location;
--- a/ykrt/src/location.rs
+++ b/ykrt/src/location.rs
@@ -1,3 +1,5 @@
+//! Trace location: track the state of a program location (counting, tracing, compiled, etc).
+
 use std::{
     marker::PhantomData,
     sync::{
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -1,3 +1,5 @@
+//! Initialise the meta-tracing system.
+
 #[cfg(test)]
 use std::time::Duration;
 use std::{
--- a/ykshim_client/src/prod_api.rs
+++ b/ykshim_client/src/prod_api.rs
@@ -1,4 +1,5 @@
 //! The "Production" API client to ykshim.
+// XXX I'm not really sure what "production" means in this context.
 
 use std::ffi::{c_void, CString};
 use std::marker::PhantomData;
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -150,6 +150,7 @@ impl<'a> WorkspaceAction<'a> {
     }
 }
 
+// `bail` is a weird name. maybe call it `abort`?
 fn bail(err_str: String) -> ! {
     eprintln!("xtask: {}", err_str);
     exit(1);


internal_ws/ykbh/src/lib.rs
```diff
@@ -347,11 +359,14 @@ impl SIRInterpreter {
...
     }
 
+    /// XXX unhelpful comment
     /// Implements the Store statement.
     fn store(&mut self, dest: &IPlace, src: &IPlace) {
         self.frames.last_mut().unwrap().mem.store(dest, src);
     }
 
+    /// XXX this might be me being an idiot, but i have no idea what this comment means or, indeed,
+    /// what the function actually does!
```
bjorn3, line 369: I am pretty sure it implements `dst = &src`.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -595,7 +599,8 @@ impl TraceCompiler {
     /// Copy bytes from one memory location to another.
     fn copy_memory(&mut self, dest: &RegAndOffset, src: &RegAndOffset, size: u64) {
         // We use memmove(3), as it's not clear if MIR (and therefore SIR) could cause copies
-        // involving overlapping buffers.
+        // involving overlapping buffers. XXX this sounds like a big thing not to be sure about!
+        // what's the underlying problem?
```
bjorn3, line 603: Using `memcpy` would require that the equivalent of something like `foo[0..2] = foo[1..3]` in MIR is UB.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -1185,7 +1192,8 @@ impl TraceCompiler {
     /// Compile a guard in the trace, emitting code to abort execution in case the guard fails.
     fn c_guard(&mut self, guard: &Guard, dl: DynamicLabel) {
         // FIXME some of the terminators from which we build these guards can have cleanup blocks.
-        // Currently we don't run any cleanup, but should we?
+        // Currently we don't run any cleanup, but should we? XXX does this mean we don't call
+        // drop? And, if so, what do we not call Drop on?
```
bjorn3, line 1196: The cleanup this is refering to is the unwind path of `-Cpanic=unwind` I think, but I am not sure.

internal_ws/ykpack/src/types.rs
```diff
@@ -132,6 +132,9 @@ impl Ty {
         }
     }
 
+    // XXX this function feels a bit evil. surely it should be the caller's job to do this? that
+    // way the need to panic will either a) be clear b) not be possible because the caller has
+    // verified things?
```
bjorn3, line 137: `unwrap_*` is a commonly used name for a function to call when only a certain variant should be passed, or it would indicate a bug.

internal_ws/yktrace/src/swt.rs
```diff
@@ -77,8 +77,11 @@ static mut TRACE_BUF_CAP: usize = 0;
 static mut __YK_SWT_ACTIVE: bool = false;
 
 // =======================
-// The following code may not call any non `#[do_not_trace]` function to prevent an infinite
-// recursion of `__yk_swt_rec_loc` calling a function calling `__yk_swt_rec_loc`.
+// XXX can we scope "following code"? does it mean "any of the rest of the code in this file?" If
+// so, it might be worth either a) putting this dangerous code in its own file with the warning at
+// the top b) attaching the warning to each function. perhaps the latter is the safest thing to do!
+// The following code must not call any non `#[do_not_trace]` function: if it does, an infinite
+// loop will occur where `__yk_swt_rec_loc` calls `__yk_swt_rec_loc`.
```
bjorn3, line 84: With the split to two workspaces, this has become obsolete. In fact the `#[do_not_trace]` is no longer necessary as this crate is never traced.

internal_ws/yktrace/src/tir.rs
```diff
@@ -472,6 +472,11 @@ impl VarRenamer {
         if let Some(v) = self.stack.last() {
             self.offset = *v;
         } else {
+            // presumably this fundamentally can't happen? at which point a `debug_assert` might be
+            // more appropriate so that the function is:
+            //   self.stack.pop();
+            //   debug_assert!(!self.stack.is_empty());
+            //   self.offset = self.stack.last().unwrap();
```
bjorn3, line 479: It can be `self.offset = self.stack.last().expect("Unbalanced ...")`.

ykshim_client/src/prod_api.rs
```diff
@@ -1,4 +1,5 @@
 //! The "Production" API client to ykshim.
+// XXX I'm not really sure what "production" means in this context.
```
bjorn3, line 2: To be used in production as opposed to only used for testing I guess.

internal_ws/yktrace/src/tir.rs
ltratt: I'd replace the `expect` with `unwrap`, but otherwise I agree!

internal_ws/ykcompile/src/lib.rs
ltratt: And it's not UB, right?

bjorn3: It hasn't been decided yet as far as I know, but I would personally expect it to become UB. At least in most cases. Keeping `memmove` for now in case it doesn't become UB is fine though.

ltratt: If we expect it to become UB I'd be fine with sticking an `assert` in there to catch such cases and moving away from `memmove`. I don't feel strongly about this, though.

bjorn3: It can only be realistically asserted at runtime when executing the compiled trace.

ltratt: Yes, that's the only thing that's making me not be gun-ho about it. On reflection, I think we should probably just link to a Rust github issue / RFC for the relevant UB (if we know one!) in a comment.

bjorn3: https://github.com/rust-lang/rust/issues/68364 is the issue about this.

ltratt: Perfect, thanks!

internal_ws/ykbh/src/lib.rs
vext01: I think that's correct. `mkref` is a SIR and TIR bytecode and this is the implementation of this in the interpreter, right @ptersilie?

ltratt: The fact that three of us aren't 100% sure says that we need, at a minimum, to better document this ;)

internal_ws/ykcompile/src/lib.rs
vext01: A lot of the terminators have "cleanup" blocks, which should be run in certain circumstances. This comment was written before we committed to `abort`, but more recently we've questioned that again...

internal_ws/ykpack/src/types.rs
vext01: Yep. It's just a convenience function to cut down on code duplication.

ltratt: OK. At least it needs a `# Safety` or `#Panics` qualification in the doc.

ykshim_client/src/prod_api.rs
vext01: Yes.

It's implied [here](https://github.com/softdevteam/yk/blob/d167d14fb9e026a45cf1ef77830a49008f3666e5/internal_ws/ykshim/src/lib.rs#L10), but we could make it more explicit.

internal_ws/ykbh/src/lib.rs
```diff
@@ -9,6 +11,8 @@ use yktrace::sir::SIR;
 /// Stores information needed to recreate stack frames in the SIRInterpreter.
 pub struct FrameInfo {
     /// The symbol name of this frame.
+    /// i might have misunderstood this, but does every function call in the stopgap interpreter
+    /// require mallocing a String?
```
vext01, line 15: Lukas and I talked about whether it was possible to borrow a slice directly from the SIR. It might be possible, as (currently) we never unload a SIR body once it's loaded. If we ever want to evict SIR, we must copy.

vext01: @ptersilie why don't we just store the corresponding `Arc` in this struct?

internal_ws/ykbh/src/lib.rs
```diff
@@ -139,37 +143,42 @@ impl LocalMem {
...
 struct StackFrame {
     /// Allocated memory holding live locals.
     mem: LocalMem,
-    /// The current basic block index of this frame. Upon returning from a function call it is used
+    /// The current basic block index of this frame. XXX i don't understand this sentence. Upon returning from a function call it is used
     /// to look up the previous basic block and check its terminator to decide where to continue
     /// interpreting.
     bbidx: ykpack::BasicBlockIndex,
-    /// Symbol name of this stack frame. Needed to retrieve the SIR body of the function which
+    /// woah, we malloc a `String` per stack frame?! that's going to be very slow. we need an alternative mechanism here. Symbol name of this stack frame. Needed to retrieve the SIR body of the function which
```
vext01, line 156: Again, can we store an `Arc<Body>`? Then getting the symbol name is `body.symbol_name`.

internal_ws/ykbh/src/lib.rs
```diff
@@ -139,37 +143,42 @@ impl LocalMem {
...
     /// contains the statements we want to interpret.
     func: String,
 }
 
-/// The SIR interpreter, also known as blackholing interpreter, is invoked when a guard fails in a
-/// trace. It is initalised with information from the trace, e.g. live variables, stack frames, and
-/// then run to get us back to a control point from where the normal interpreter can take over.
+/// The SIR interpreter, also known as the blackhole interpreter, is invoked when a guard fails in
+/// a trace. It is initialised with information from the trace (e.g. live variables, stack frames)
+/// and then runs until we are back to the control point, from where the normal interpreter can
```
vext01, line 163: s/from_where/from which point/

internal_ws/ykbh/src/lib.rs
```diff
@@ -139,37 +143,42 @@ impl LocalMem {
...
 pub struct SIRInterpreter {
-    /// Keeps track of active stack frames (most recent last).
+    /// Active stack frames (most recent last).
     frames: Vec<StackFrame>,
 }
 
 impl SIRInterpreter {
+    /// XXX i think this test is only needed for tests? if so, we should at least #[cfg(test)] it,
+    /// and probably give it a name that doesn't suggests it's the main way of initialising the
+    /// interpreter.
```
vext01, line 173: Be aware that `#[cfg(test)]` [only works intra-crate](https://github.com/rust-lang/cargo/issues/8379). If this is tested from the external workspace, annotating it won't work and will lead to a confusing error message.

internal_ws/ykbh/src/lib.rs
```diff
@@ -194,6 +203,8 @@ impl SIRInterpreter {
     /// Run the SIR interpreter after it has been initialised by a guard failure. Since we start in
     /// the block where the guard failed, we immediately skip to the terminator and interpret it to
     /// see which block we need to start interpretation in.
+    /// XXX we have one function called "interpret" and another "_interpret" -- both public! At
+    /// least one of them needs a better name.
```
vext01, line 207: Does `_interpret` need to be public? Try removing `pub`.

internal_ws/ykbh/src/lib.rs
```diff
@@ -281,8 +291,8 @@ impl SIRInterpreter {
                 self.frames.push(frame);
             }
             Terminator::Return => {
-                // Return from current stackframe.
                 let oldframe = self.frames.pop().unwrap();
+                // XXX comment could do with rephrasing
```
vext01, line 295: Something like "If there are no further frames, then we are back at the control point".

internal_ws/ykcompile/src/lib.rs
```diff
@@ -264,7 +264,9 @@ fn local_to_reg_name(loc: &Location) -> &'static str {
 // interpreter for blackholing.
 
 /// Given a vector of `FrameInfo`s `vptr`, instantiates and initialises a SIR interpreter and returns its
-/// pointer. Consumes `vptr`.
+/// pointer. Consumes `vptr`. XXX note this requires a move of `vptr`, so we should document
+/// something like "note that the calling function must fully transfer ownership of vptr
+/// to this function".
```
vext01, line 269: It says it consumes `vptr`. That's the idiom we've used so far in docs.

ltratt: Yes, this comment is wrong -- though there are other functions which don't document "consume". I must have done this one on auto-pilot though!

internal_ws/ykbh/src/lib.rs
ltratt: Note: we should never have functions beginning with `_`!

ykshim_client/src/prod_api.rs
ltratt: I think "production" is too vague a term to be useful. Can we find something that's clearer?

internal_ws/ykcompile/src/lib.rs
```diff
@@ -408,7 +410,7 @@ impl Location {
 
     /// Apply an offset to the location, returning a new one.
     fn offset(self, off: OffT) -> Self {
-        if off == 0 {
+        if off == 0 { // Is the special case a performance hack or ... ?
```
vext01, line 413: It's not strictly necessary, unless removing it causes us to hit the panic case. But yes, it is a small optimisation too.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -640,15 +645,17 @@ impl TraceCompiler {
         );
     }
 
-    /// Push the specified registers to the stack in order.
+    /// Push the specified registers to the stack, in the order they are specified in the array.
```
vext01, line 648: "order they are specified" -> "order they appear"

internal_ws/ykcompile/src/lib.rs
ltratt: I'm actually slightly worried that it might not be correct, or at least causes a surprise: if I pass `Location::Reg` and `off != 0` I get a panic, but with `off == 0` I don't. I'm not even sure what it means to offset a register.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -1368,7 +1376,7 @@ impl TraceCompiler {
         // Reserved stack space for spilled locals during execution.
         let soff = self.stack_builder.size();
         // Reserved memory on the stack to spill live locals to during a guard failure.
-        let live_off = REG_POOL.len() * 8;
+        let live_off = REG_POOL.len() * 8; // Is `8` a machine word? Prefer a calculation/constant name.
```
vext01, line 1379: It is, and yes that should be a constant.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -1426,6 +1434,8 @@ impl TraceCompiler {
             for (tirlocal, loc) in live_locations.iter_mut() {
                 match loc {
                     Location::Reg(reg) => {
+                        // XXX I think that "-i32" is potentially dangerous because 2's offset is
+                        // not symmetric. We might need another check for that here?
```
vext01, line 1438: Two's compliment?

The negation is always safe because the positive integer range is one less than the negative: from -2^n to (2^n) - 1.

E.g. for `i8` the range is [-128, 127]. There's no positive integer than you could negate and get the wrong result.

ltratt: Yes, but what happens if we negate a negative integer? Maybe that can't happen, but I don't think I could see that from the code.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -1440,7 +1450,7 @@ impl TraceCompiler {
                 }
             }
 
-            // Create vector to store the allocated memory pointers in, to be handed to the
+            // Create a vector "of what?" to store the allocated memory pointers in, to be handed to the
```
vext01, line 1453: It's `Vec<FrameInfo>` (see the signature of `bh_new_vec`).

Agree that this info could be more accessible.

internal_ws/ykcompile/src/stack_builder.rs
```diff
@@ -10,10 +12,10 @@ pub(crate) struct StackBuilder {
 
 /// A naive stack builder for allocating objects on the stack.
 ///
-/// Naive because it could allocate less space by caching up the requested allocations and
+/// XXX I can't parse this sentence. Naive because it could allocate less space by caching up the requested allocations and
 /// better-packing them.
 ///
-/// The top of the stack, before any calls to `alloc()` is assumed to be appropriately aligned.
+/// XXX or this sentence. The top of the stack, before any calls to `alloc()` is assumed to be appropriately aligned.
```
vext01, line 18: It means the stack must be aligned before you allocate using our stack builder. If the stack is misaligned and you use `StackBuilder` the stack will still be misaligned.

internal_ws/ykbh/src/lib.rs
ptersilie: `_interpret` is used by the test suite so it has to be public. But we can certainly do something about the name.

internal_ws/ykpack/src/types.rs
vext01: This is new to me. Surely every function that may panic doesn't need a special doc?

internal_ws/ykbh/src/lib.rs
ptersilie: Haven't tried that. When I wrote this I had other things to worry about. But there's definitely room for optimisations in the blackholing code.

internal_ws/ykpack/src/types.rs
ltratt: It shouldn't be news to you: the Rust std lib documents this carefully :) If you think about it, it's really another way of saying "what are this function's preconditions?" We don't have to document *unexpected* panics, but we should document the expected panics IMHO.

internal_ws/ykbh/src/lib.rs
ptersilie: How about we name rename this to `interpret_with_ctx` and then `_interpret` becomes `interpret`.

ltratt: That's better but I'm not sure what it means to "interpret without context", so that's still maybe not perfect. Perhaps better doc strings can clear this up, however.

internal_ws/ykpack/src/types.rs
```diff
@@ -362,12 +365,12 @@ impl Display for Ptr {
     }
 }
 
-/// An IR place. This is used in SIR and TIR to describe the (abstract) address of a piece of data.
+/// An IR place XXX why isn't the struct called "IRPlace"? [I've been wondering for ages what "IPlace" meant! "IRPlace" would be a bit clearer I think.]. This is used in SIR and TIR to describe the (abstract) address of a piece of data.
```
vext01, line 368: No objections to renaming it. I think we inherited the name from Cranelift.

internal_ws/ykbh/src/lib.rs
ptersilie: Yes, it makes a rust reference to a variable, e.g. `&a`. I thought "creates a reference to an iplace" is quite clear, but we can probably improve it.

internal_ws/ykpack/src/types.rs
```diff
@@ -362,12 +365,12 @@ impl Display for Ptr {
...
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
 pub enum IPlace {
     /// The IPlace describes a value as a Local+offset pair.
     Val { local: Local, off: OffT, ty: TypeId },
-    /// An indirect place behind a pointer. ykrustc uses these for deref and (dynamic) index
+    /// An indirect place behind a pointer XXX huh? i'm not sure what that means. ykrustc uses these for deref and (dynamic) index
```
vext01, line 373: "A pointer to an IPlace".

internal_ws/ykpack/src/types.rs
```diff
@@ -446,6 +449,8 @@ impl IPlace {
                     ty: new_ty,
                 }
             }
+            // are these unreachable or todos? if the former, the strings are unnecessary (and just
+            // lead to a bit of pointless binary bloat)
```
vext01, line 453: I'm not sure if these cases can arise or not.

The strings come from a time before panic would give a line number.

internal_ws/ykcompile/src/lib.rs
ptersilie: I actually assumed this would be `unreachable` for registers and `offset != 0`. At least I've not seen a register/constant being offset, so we need to investigate this.

internal_ws/ykbh/src/lib.rs
ltratt: The problem is that "place" is a very vague term, so it doesn't give much of an intuition. I made a comment on the `IPlace` struct too to this effect IIRC.

internal_ws/yktrace/src/hwt/mapper.rs
```diff
@@ -73,6 +75,7 @@ impl HWTMapper {
 
 /// Extract the program header offset. This offset can be used to translate the address of a trace
 /// block to a program address, allowing us to find the correct SIR location.
+/// XXX if this is an address should it be usize?
```
vext01, line 78: We need to check this, but I think it's because `Elf_Addr` is `uint64_t` on x86_64. This code needs to be platform guarded.

tests/src/helpers.rs
```diff
@@ -1,3 +1,6 @@
+// XXX does this file need to be on its own? also, "helpers" is the most unhelpful module name in
+// the history of module names ;)
+
```
vext01, line 3: This is (or was, need to check) for sharing stuff between different test files.

internal_ws/ykcompile/src/lib.rs
ptersilie: The input for `i32::try_from` is a `u32`, so we can't negate a negative integer here.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -1562,7 +1572,7 @@ impl TraceCompiler {
                 let ptr = buf.ptr(dynasmrt::AssemblyOffset(0)) as *mut libc::c_void;
                 let len = buf.len();
                 let alignment = ptr as usize % libc::sysconf(libc::_SC_PAGESIZE) as usize;
-                let ptr = ptr.offset(-(alignment as isize));
+                let ptr = ptr.offset(-(alignment as isize)); // probably better as ptr.pub?
```
ptersilie, line 1575: I assume you mean `ptr.sub`.

internal_ws/ykbh/src/lib.rs
vext01: FWIW, `Place` has a very specific meaning in MIR.

internal_ws/ykpack/src/types.rs
vext01: A panic here would indeed be unexpected. It's not in a public API :P

ykshim_client/src/prod_api.rs
vext01: We just need to find a term that means "not testing". Production is the best I can think of. Any suggestions?

internal_ws/yktrace/src/tir.rs
```diff
@@ -82,7 +82,7 @@ impl<'a, 'm> TirTrace<'a, 'm> {
             }
 
             // If a function was annotated with `do_not_trace`, skip all instructions within it as
-            // well. FIXME: recursion.
+            // well. FIXME: recursion. what does "recursion" mean?!
```
ptersilie, line 85: I think this used to be recursive so we may have just forgotten to remove this comment. @vext01?

internal_ws/ykbh/src/lib.rs
vext01: @ptersilie  how about `inner_interpret` or `interpret_impl`?

internal_ws/ykcompile/src/lib.rs
vext01: > I'm not even sure what it means to offset a register.

This is currently undefined, so it probably should crash.

internal_ws/yktrace/src/tir.rs
vext01: I thought it meant that we can't handle recursive calls to `do_not_trace` functions?

internal_ws/ykcompile/src/lib.rs
vext01: I look at it from a more functional view. To get to the stuff we've put on the stack, we always have to use a negative index.

internal_ws/ykbh/src/lib.rs
ptersilie: There is no "interpret without context" as `interpret_with_ctx` calls `interpret`. But I agree, this is a bit confusing. `_interpret` only ever used for the stand-alone `SIRInterpreter` tests. Maybe we can update the tests to use the same function. The main difference is the `self.terminator()` line, which isn't needed when testing the Sir interpreter without blackholing.

internal_ws/ykcompile/src/lib.rs
bjorn3: `<*mut T>::sub` uses pointer arithmetic. Aka it offsets in multiples of `size_of::<T>()`.

ltratt: s/pub/sub/ indeed :) But, as @bjorn3 points out, there are different subtraction functions depending on whether you want bytes or object subtraction.

internal_ws/ykcompile/src/lib.rs
ltratt: > The input for i32::try_from is a u32, so we can't negate a negative integer here.

Aha, OK.

internal_ws/yktrace/src/tir.rs
ltratt: > I thought it meant that we can't handle recursive calls to do_not_trace functions?

That would be bad, surely?!

internal_ws/ykcompile/src/lib.rs
ltratt: Right, so I think we should remove the `if` -- the optimisation probably isn't significant, and it potentially hides some bad cases.

ykshim_client/src/prod_api.rs
ltratt: "Non-testing" might indeed be the best term?

internal_ws/ykpack/src/types.rs
ltratt: By "unexpected" meaning "if this panics, I, the function, have got something wrong" vs. "expected" meaning "you the caller did not respect the precondition I require you to respect." The precondition in the latter thus needs documenting.

internal_ws/ykbh/src/lib.rs
ltratt: My impression was that *our* `IPlace` is subtly different?

internal_ws/yktrace/src/hwt/mapper.rs
ltratt: OK.

internal_ws/yktrace/src/tir.rs
ptersilie: Ah, in that case this has been fixed recently in https://github.com/softdevteam/yk/commit/b799c78df1a79a4829ee33e0210cbb1cfa4ac9e2. So this is just a left-over comment.

internal_ws/ykbh/src/lib.rs
vext01: It is, but each MIR place is [lowered to one of our `I[R]Place`s](https://github.com/softdevteam/ykrustc/blob/df02639369734daab3f6505acfc01b8ae783b7be/compiler/rustc_codegen_ssa/src/sir.rs#L497-L502), so it makes sense to maintain that association.

ltratt: But if ours does more, the association implied by the name might be misleading more than it helps?

vext01: I'm not sure it does more. It does different. It works with byte offsets instead of projections, but conceptually they are the same: an abstract storage location whose actual storage location is to be decided (in their case, by LLVM, in ours, by the trace compiler).

ltratt: OK, maybe `IRPlace` is different-enough-but-similar-enough to keep us all happy :)

internal_ws/ykpack/src/types.rs
vext01: I have no issue documenting it, but i'm surprised it's necessary. This idiom is really common :P

ykshim_client/src/prod_api.rs
vext01: Hrm. Not sure about that. Do we really want files named 'non_testing.rs`? That's much less clear than `production` IMHO.

With the aid of a thesaurus, here's some more suggestions for "non testing api":
 - "essential API"
 - "real API"
 - "actual API"
 - "functional API"
 - "true API"

internal_ws/ykpack/src/types.rs
ltratt: I thought the same at one point, but then realised that if we *always* do it, we never have to make the judgment call -- and in the long run, it saves time for me not to think about it.

ykshim_client/src/prod_api.rs
ltratt: They're all worse ;) I suspect the thing to do is to try and think of a different term than "production" entirely and that might then make thesaurus diving useful.

vext01: "Working API" and then I'm spent.

Failing that, let's leave it as `production`. Software developers know that production is the opposite of testing. It makes sense.

ltratt: After an online discussion, I think we've agreed to remove the label "production" entirely on the basis that it's "normal" code. I *think* @vext01 also suggested renaming this to `lib.rs`. Testing modules will remain labelled as such to make clear that they're not "normal" code. @vext01 does that reflect your understanding?

vext01: That's right. Just move the contents of `prod.rs` (in both workspaces) into the (already existing) `lib.rs`.

internal_ws/ykbh/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
```
ptersilie, line 1: Fixed in df14c2a.

internal_ws/ykbh/src/lib.rs
```diff
@@ -139,37 +143,42 @@ impl LocalMem {
     }
 }
 
-/// A interpreter stack frame, containing allocated memory for the frames locals, and the function
+/// An interpreter stack frame, containing allocated memory for the frames locals, and the function
```
ptersilie, line 146: Fixed in df14c2a.

internal_ws/ykbh/src/lib.rs
```diff
@@ -139,37 +143,42 @@ impl LocalMem {
...
 /// symbol name and basic block index needed by the interpreter to continue interpreting after
 /// returning from a function call.
 struct StackFrame {
     /// Allocated memory holding live locals.
     mem: LocalMem,
-    /// The current basic block index of this frame. Upon returning from a function call it is used
+    /// The current basic block index of this frame. XXX i don't understand this sentence. Upon returning from a function call it is used
```
ptersilie, line 152: This sentence doesn't add any information so I removed it in df14c2a.

internal_ws/ykbh/src/lib.rs
ptersilie: Fixed in df14c2a.

internal_ws/ykbh/src/lib.rs
```diff
@@ -139,37 +143,42 @@ impl LocalMem {
...
-/// The SIR interpreter, also known as blackholing interpreter, is invoked when a guard fails in a
-/// trace. It is initalised with information from the trace, e.g. live variables, stack frames, and
-/// then run to get us back to a control point from where the normal interpreter can take over.
+/// The SIR interpreter, also known as the blackhole interpreter, is invoked when a guard fails in
+/// a trace. It is initialised with information from the trace (e.g. live variables, stack frames)
+/// and then runs until we are back to the control point, from where the normal interpreter can
+/// take over.
 pub struct SIRInterpreter {
-    /// Keeps track of active stack frames (most recent last).
+    /// Active stack frames (most recent last).
```
ptersilie, line 166: Fixed in df14c2a.

internal_ws/ykbh/src/lib.rs
ptersilie: Renamed to `from_symbol` in df14c2a.

internal_ws/ykbh/src/lib.rs
```diff
@@ -139,37 +143,42 @@ impl LocalMem {
...
+    /// and probably give it a name that doesn't suggests it's the main way of initialising the
+    /// interpreter.
     pub fn new(sym: String) -> Self {
         let frame = SIRInterpreter::create_frame(&sym);
         SIRInterpreter {
             frames: vec![frame],
         }
     }
 
+    /// XXX I think this might better be called `from_frames`?
```
ptersilie, line 181: Fixed in df14c2a.

internal_ws/ykbh/src/lib.rs
ptersilie: There's a temporary solution in df14c2a, renaming them to `sg_interpret` (stopgap_interpret) and `interpret`. Though I have an idea how to get rid of one of the functions entirely, though its a little more involved and thus for a later PR.

internal_ws/ykbh/src/lib.rs
```diff
@@ -205,13 +216,13 @@ impl SIRInterpreter {
         self._interpret();
     }
 
-    /// Given a vector of local declarations allocate just enough space to hold all of them.
+    /// Given a vector of local declarations allocate just enough XXX "just enough" or "precisely
```
ptersilie, line 219: Fixed in df14c2a.

internal_ws/ykbh/src/lib.rs
ptersilie: Fixed in df14c2a.

internal_ws/ykbh/src/lib.rs
```diff
@@ -298,7 +308,9 @@ impl SIRInterpreter {
                         } => dest.as_ref().map(|(p, b)| (p.clone(), *b)).unwrap(),
                         _ => unreachable!(),
                     };
-                    // Get a pointer to the return value of the called frame.
+                    // Get a pointer to the return value of the called frame. XXX this needs to be
```
ptersilie, line 311: Fixed in 66bd8ff (though I need to check the other files later too).

internal_ws/ykbh/src/lib.rs
```diff
@@ -205,13 +216,13 @@ impl SIRInterpreter {
...
 
-    /// Given a vector of local declarations allocate just enough space to hold all of them.
+    /// Given a vector of local declarations allocate just enough XXX "just enough" or "precisely
+    /// the right amount"? space to hold all of them.
     fn create_frame(sym: &String) -> StackFrame {
         let body = SIR.body(&sym).unwrap();
         let (size, align) = body.layout;
         let offsets = body.offsets.clone();
         let layout = Layout::from_size_align(size, align).unwrap();
-        // Allocate memory for the locals
```
ptersilie, line 214: Fixed in df14c2a.

internal_ws/ykbh/src/lib.rs
```diff
@@ -235,11 +246,11 @@ impl SIRInterpreter {
         // The interpreter context lives in $1
         let ptr = self.frames.first().unwrap().mem.local_ptr(&Local(1));
         unsafe {
-            // Write the pointer value of `tio` into this frames memory.
```
ptersilie, line 238: Fixed in df14c2a.

internal_ws/ykbh/src/lib.rs
```diff
@@ -235,11 +246,11 @@ impl SIRInterpreter {
...
             std::ptr::write::<*mut u8>(ptr as *mut *mut u8, ctx);
         }
     }
 
+    /// XXX this function shouldn't be prefixed with a `_`
```
ptersilie, line 253: Fixed in b5bce55.

internal_ws/ykbh/src/lib.rs
```diff
@@ -267,7 +278,6 @@ impl SIRInterpreter {
             Terminator::Call {
                 operand: op,
                 args,
-                destination: _dest,
```
ptersilie, line 270: We can't just remove this line, so not sure what to do here.

internal_ws/ykbh/src/lib.rs
```diff
@@ -281,8 +291,8 @@ impl SIRInterpreter {
                 self.frames.push(frame);
             }
             Terminator::Return => {
-                // Return from current stackframe.
```
ptersilie, line 284: Fixed in df14c2a.

internal_ws/ykbh/src/lib.rs
```diff
@@ -347,11 +359,14 @@ impl SIRInterpreter {
         }
     }
 
+    /// XXX unhelpful comment
```
ptersilie, line 362: Fixed in df14c2a.

internal_ws/ykbh/src/lib.rs
ptersilie: Improve the doc comment in df14c2a.

internal_ws/ykbh/src/lib.rs
ptersilie: Addressed in 8f5174b.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -288,7 +290,7 @@ pub extern "sysv64" fn bh_new_vec() -> *mut Vec<FrameInfo> {
 
 /// Pushes a new `FrameInfo` instance onto the vector behind the pointer `vptr`. The `FrameInfo`
 /// instance is creates from a pointer to a symbol name and its length, a basic block index and a
-/// pointer to some allocated memory.
+/// pointer to some allocated memory. XXX need to document that `vptr` is consumed
```
ptersilie, line 293: In this case `vptr` is not consumed. Should I point that out in the docs, or is it implied if it's not specifically pointed out?

vext01: I think it would be excessive to point out every time something is not consumed. Then again, the raw pointers make it hard to know.

Perhaps it's OK in this scenario.

ltratt: > In this case vptr is not consumed.

But surely it is at line 306?

```rust
let mut v = unsafe { Box::from_raw(vptr) };
```

internal_ws/ykbh/src/lib.rs
ptersilie: Fixed in 0163c4d.

internal_ws/ykbh/src/lib.rs
ptersilie: Fixed in 0163c4d.

internal_ws/ykcompile/src/lib.rs
ptersilie: Yes, but it's then "deconsumed" (regurgitated?) in 308. So in a way it's consumed, but also the pointer is still valid when the function finishes (and is indeed used in later calls to `bh_push_vec`).

ptersilie: But now that I think about it, I'm not sure Rust guarantees that the pointer will be the same after calling `Box::into_raw`.

bjorn3: It should then use `let mut v = unsafe { &mut *vptr };`.

ptersilie: Will use that then. But won't we still be in trouble if the vector is resized and needs to be moved in memory, invalidating the pointer?

bjorn3: This is a pointer to a vector. This vector contains a pointer to the allocated memory.

ptersilie: Ah, yes of course! Thanks.

ltratt: There's no notion of "deconsume" unfortunately: there's a very good chance this code is UB, even with @bjorn3's suggestion. The question is: where is this called from?

ptersilie: It's called from the compiled trace during a guard failure: https://github.com/softdevteam/yk/blob/master/internal_ws/ykcompile/src/lib.rs#L1479

bjorn3: With my suggestion nothing is consumed. Only borrowed.

ptersilie: Just to give some more information: We create a new vector with `bh_new_vec`, which is then boxed and the pointer stored in a register. We then call `bh_push_vec` one or more times, passing that pointer in as an argument (with @bjorn3 suggesting we then only mutably borrow the pointer to push elements onto the vec). Finally, we pass the vector pointer into `invoke_sinterp`, where it is consumed.

ltratt: The full code is:

```rust
    let mut v = unsafe { Box::from_raw(vptr) };
    v.push(fi);
    Box::into_raw(v);
```

If something else has a mutable reference to that at the same time, we're in UB AFAICS, because we're calling into the vector.

bjorn3: With my suggestion that becomes

```rust
let mut v = unsafe { &mut *vptr };
v.push(fi);
```

This doesn't consume anything.

vext01: Is "that" the vector itself? Can we downgrade the mutability after the push?

bjorn3: `vptr` is a pointer to the vector, so `v` is a mutable reference to the vector.

ltratt: The problem is that if someone else has a mutable reference to the vector and calls this function, we've now got two `&mut` references to the same thing. That is likely (though with stacked borrows there are some exceptions) to be UB. In other words, there *must* be some documented constraints on `vptr` that the caller has to respect.

ptersilie: Isn't the fact that `vptr` is a `mut *` already telling the caller that the pointer is likely to be mutated and thus this function shouldn't be called in parallel. Though I'm happy to add a warning to the doc string.

ltratt: > Isn't the fact that vptr is a mut * already telling the caller that the pointer is likely to be mutated and thus this function shouldn't be called in parallel

Given that some other functions consume the pointer (and thus the `mut *` is quasi-irrelevant) the answer is clearly "no" :) In other words, if every function consumed the pointers, that would be easy; if none of them consumed the pointers that would be a bit more difficult, but at least consistent; but what we've got is a mix.

Basically, raw pointers in Rust are incredibly dangerous -- much, much, more dangerous than in C -- so we have to treat them like unexploded bombs. It's difficult for us to be too cautious!

Let me ask the dangerous question: at the point we get a pointer to this Vec, can we guarantee with 100% confidence that none else holds a `&` or `&mut` reference to it? [It's totally fine if they have a `const *` or `mut *` reference at the same time, but `&` or `&mut` are strictly verboten.]

ptersilie: > at the point we get a pointer to this Vec, can we guarantee with 100% confidence that none else holds a & or &mut reference to it?

With the current code? Yes. The pointer to this vector is only ever passed into two functions: `bh_vec_push` and `invoke_sinterp`, the former is always called before the latter, and the latter consumes the pointer after which it's not used any more (see also my summary comment of this code above).

ltratt: > The pointer to this vector is only ever passed into two functions

But, just to check, does any other code hold a `&`/`&mut` reference to this vec? *If* we convert it once with `Box::into_raw` and this is the only code which converts it back to a `&`/`&mut`, we're safe.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -288,7 +290,7 @@ pub extern "sysv64" fn bh_new_vec() -> *mut Vec<FrameInfo> {
 
 /// Pushes a new `FrameInfo` instance onto the vector behind the pointer `vptr`. The `FrameInfo`
```
ltratt, line 291: Since we know that the pointer is really a `Box` -- and we rely on that by calling `Box::from_raw`! -- we should document that this pointer *must* have been created with `Box::new`. Or, alternatively, we could stop calling `Box::from_raw`, use @bjorn3's `&mut *` suggestion, and then we don't need to specify that the vector must have been created with `Box`. I tend to prefer the latter approach as it's one thing fewer for our future selves to get wrong.

internal_ws/ykcompile/src/lib.rs
ptersilie: No, the vector is created in `bh_new_vec` and then immediately boxed and returned to the assembler code, which only passes it to `bh_vec_push` and `invoke_sinterp`. So at the moment there is no `&`/`&mut` reference, and with @bjorn3's suggestion there will be one.

ltratt: OK, let's do it!

ptersilie: Fixed in 2c47787fbaaf271fdb2fef39104bbf681eee6454.

internal_ws/ykcompile/src/lib.rs
ptersilie: Fixed in 2c47787.

internal_ws/ykcompile/src/lib.rs
ptersilie: Fixed in 2c47787fbaaf271fdb2fef39104bbf681eee6454.

internal_ws/ykcompile/src/lib.rs
ptersilie: Think this was resolved in 9a21d335171ee9c02cf96bc8938609c43acacfed.

internal_ws/yktrace/src/tir.rs
ptersilie: > I thought it meant that we can't handle recursive calls to `do_not_trace` functions?

I misread this before. Now I know what you mean. However, does it matter if the `do_not_trace` function is recursive? The first time we see it, we replace it with a call, and then we'll never know if it is recursive or not. I'll do some testing.

ptersilie: As far as I can tell, this works already, so I think we can mark this as resolved.

ltratt: I am inclined to agree.

ltratt: Er though I can't see an update to this comment in b799c78, but maybe I'm missing something?

ptersilie: Yeah, sorry, I need to remove the FIXME first before resolving this.

ptersilie: Okay, this turned out to not being fixed after all: b88e9fc

internal_ws/ykcompile/src/lib.rs
```diff
@@ -1093,7 +1100,7 @@ impl TraceCompiler {
     fn c_cast(&mut self, dest: &IPlace, src: &IPlace) {
         let src_loc = self.iplace_to_location(src);
         let ty = &*SIR.ty(&src.ty()); // Type of the source.
-        let cty = SIR.ty(&dest.ty()); // Type of the cast (same as dest type).
+        let cty = SIR.ty(&dest.ty()); // Type of the cast (same as dest type XXX isn't this the case by definition?).
```
ptersilie, line 1103: Yes. I think I just wanted to make sure it's clear why we didn't use the explicit type from the cast statement, which we remove during SIR construction for this very reason. But I agree that we can just remove the stuff in the brackets.

internal_ws/ykcompile/src/lib.rs
ptersilie: What are we gonna do about this comment then?

vext01: I think cleanup blocks are not run if the abort strategy is used.

If so, we would only fix this if we moved to unwind strategy.

@bjorn3 may know.

ltratt: "run in certain circumstances" == "only if there's a panic"?

internal_ws/ykcompile/src/lib.rs
```diff
@@ -308,7 +310,7 @@ pub extern "sysv64" fn bh_push_vec(
     Box::into_raw(v);
 }
 
-/// Compile a TIR trace, returning executable code.
+/// Compile a TIR trace.
```
ptersilie, line 313: Fixed in db5a334.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -452,7 +454,7 @@ pub struct TraceCompiler {
     register_content_map: HashMap<u8, RegAlloc>,
     /// Maps trace locals to their location (register, stack).
     variable_location_map: HashMap<Local, Location>,
-    /// Local decls of the tir trace.
+    /// Local decls of the TIR trace.
```
ptersilie, line 457: Fixed in db5a334.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -564,6 +566,8 @@ impl TraceCompiler {
         self.stack_builder.alloc(ty.size(), ty.align())
     }
 
+    /// XXX Doc comment needed and maybe a better function name? I'm not sure if it means "make the
+    /// local come alive" or "check the local is alive"
```
ptersilie, line 570: Fixed in db5a334.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -595,7 +599,8 @@ impl TraceCompiler {
     /// Copy bytes from one memory location to another.
     fn copy_memory(&mut self, dest: &RegAndOffset, src: &RegAndOffset, size: u64) {
         // We use memmove(3), as it's not clear if MIR (and therefore SIR) could cause copies
-        // involving overlapping buffers.
+        // involving overlapping buffers. XXX this sounds like a big thing not to be sure about!
```
ptersilie, line 602: Fixed in db5a334.

internal_ws/ykcompile/src/lib.rs
ptersilie: Fixed in db5a334.

internal_ws/ykcompile/src/lib.rs
ptersilie: Fixed in db5a334.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -640,15 +645,17 @@ impl TraceCompiler {
...
     fn save_regs(&mut self, regs: &[u8]) {
         for reg in regs.iter() {
             dynasm!(self.asm
                 ; push Rq(reg)
             );
         }
     }
-    /// Pop the specified registers from the stack in reverse order.
+
+    /// Pop the specified registers from the stack in reverse order to how they are specified in
```
ptersilie, line 657: Fixed in db5a334.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -1056,7 +1063,7 @@ impl TraceCompiler {
         let src_loc = self.iplace_to_location(src);
         match src_loc {
             Location::Reg(..) => {
-                // This isn't possible as the allocator explicitely puts things which are
+                // This isn't possible as the allocator explicitly puts things which are
```
ptersilie, line 1066: Fixed in db5a334.

internal_ws/ykcompile/src/lib.rs
ptersilie: Fixed in db5a334.

internal_ws/ykcompile/src/lib.rs
ptersilie: Fixed in db5a334.

internal_ws/ykcompile/src/lib.rs
ptersilie: Fixed in db5a334.

internal_ws/ykpack/src/types.rs
ptersilie: Fixed in 396b940.

internal_ws/ykpack/src/types.rs
ptersilie: Fixed in 396b940.

internal_ws/ykpack/src/types.rs
```diff
@@ -467,7 +472,7 @@ pub enum Statement {
         opnd2: IPlace,
         checked: bool,
     },
-    /// Makes a reference.
+    /// Makes a reference. XXX from what to what?
```
ptersilie, line 475: Fixed in 396b940.

internal_ws/ykpack/src/types.rs
```diff
@@ -952,6 +957,7 @@ mod tests {
     }
 }
 
+/// XXX this should be above the tests somewhere!
```
ptersilie, line 960: Fixed in 396b940.

internal_ws/yktrace/src/hwt/mapper.rs
vext01: Yes, so on a 64-bit platform, [ELF represents addresses as a `u64`](https://docs.rs/libc/0.2.86/libc/type.Elf64_Addr.html).

But we should define our own type (say `ElfAddr`) which is either a `u32` or a `u64` depending on platform. Isn't there a `target_pointer_width` `cfg` or something?

internal_ws/yktrace/src/lib.rs
```diff
@@ -23,6 +25,7 @@ use sir::SirTrace;
 use ykpack::Local;
 
 // In TIR traces, the argument to the interp_step is always local #1.
+// XXX this constant could be used in quite a few places throughout the repo!
```
ptersilie, line 28: Fixed in 01a96d0f1a803e5ca4d78eb5ea31bec05847e753.

internal_ws/yktrace/src/hwt/mapper.rs
vext01: [here](https://doc.rust-lang.org/reference/conditional-compilation.html#target_pointer_width)

internal_ws/yktrace/src/lib.rs
ltratt: Has this been merged already?

internal_ws/yktrace/src/hwt/mapper.rs
ptersilie: Looks like libipt is fixed to 64bit for now: https://github.com/intel/libipt/blob/905313ae4f066d0f79535ade8d4bc518bf49913c/libipt/include/intel-pt.h.in#L2231-L2233

internal_ws/yktrace/src/swt.rs
ptersilie: So, can we just remove this whole comment then? I'm scared to touch anything swt related. :smile:

tests/src/helpers.rs
vext01: Fixed by #252.

internal_ws/yktrace/src/hwt/mapper.rs
vext01: This is actually quite fiddly. I think it should be a separate PR.

internal_ws/ykshim/src/prod_api.rs
```diff
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
```
ptersilie, line 1: Fixed in 114a1ca.

internal_ws/yktrace/src/errors.rs
```diff
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
```
ptersilie, line 1: Fixed in 114a1ca.

internal_ws/yktrace/src/hwt/mapper.rs
```diff
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
```
ptersilie, line 1: Fixed in 114a1ca.

internal_ws/yktrace/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
```
ptersilie, line 1: Fixed in 114a1ca.

internal_ws/ykview/src/main.rs
```diff
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
```
ptersilie, line 1: Fixed in 114a1ca.

ykrt/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
```
ptersilie, line 1: Fixed in 114a1ca.

ykrt/src/location.rs
```diff
@@ -1,3 +1,5 @@
+//! Trace location: track the state of a program location (counting, tracing, compiled, etc).
```
ptersilie, line 1: Fixed in 114a1ca.

ykrt/src/mt.rs
```diff
@@ -1,3 +1,5 @@
+//! Initialise the meta-tracing system.
```
ptersilie, line 1: Fixed in 114a1ca.

internal_ws/yktrace/src/hwt/mapper.rs
vext01: https://github.com/softdevteam/yk/issues/256

internal_ws/ykcompile/src/stack_builder.rs
ptersilie: Fixed in 114a1ca8742ee17f5b783c70a197e6a5d2a7d72e.

internal_ws/ykpack/src/encode.rs
```diff
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
```
ptersilie, line 1: Fixed in 114a1ca.

internal_ws/ykpack/src/decode.rs
```diff
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
```
ptersilie, line 1: Fixed in 114a1ca.

internal_ws/ykcompile/src/stack_builder.rs
```diff
@@ -1,3 +1,5 @@
+//! XXX needs a top-level doc comment
```
ptersilie, line 1: Fixed in 114a1ca.

internal_ws/yktrace/src/tir.rs
```diff
@@ -229,9 +229,9 @@ impl<'a, 'm> TirTrace<'a, 'm> {
                     // Rename the return value.
                     //
                     // FIXME It seems that calls always have a destination despite the field being
-                    // `Option`. If this is not always the case, we may want add the `Local` offset
-                    // (`var_len`) to this statement so we can assign the arguments to the correct
-                    // `Local`s during trace compilation.
+                    // `Option`. If this is not always the case, we may want to add the `Local`
```
ptersilie, line 232: Fixed in 232c144.

internal_ws/yktrace/src/tir.rs
```diff
@@ -280,7 +280,7 @@ impl<'a, 'm> TirTrace<'a, 'm> {
                             term_stmts.push(Statement::Call(op.clone(), newargs, Some(ret_val)))
                         }
                     } else {
-                        todo!("Unknown callee encountered");
+                        todo!();
```
ptersilie, line 283: Fixed in 232c144.

internal_ws/yktrace/src/tir.rs
```diff
@@ -335,7 +335,7 @@ impl<'a, 'm> TirTrace<'a, 'm> {
                     // Peek at the next block in the trace to see which outgoing edge was taken and
                     // infer which value we must guard upon. We are working on the assumption that
                     // a trace can't end on a SwitchInt. i.e. that another block follows.
-                    let next_blk = itr.peek().expect("no block to peek at").bb_idx;
+                    let next_blk = itr.peek().unwrap().bb_idx;
```
ptersilie, line 338: Fixed in 232c144.

internal_ws/yktrace/src/tir.rs
```diff
@@ -461,7 +461,7 @@ impl VarRenamer {
         self.offset = self.acc.unwrap();
         self.stack.push(self.offset);
         if let Some(v) = self.acc.as_mut() {
-            *v += num_locals as u32;
+            *v += num_locals as u32; // u32::try_from?
```
ptersilie, line 464: Fixed in 232c144.

internal_ws/yktrace/src/tir.rs
ptersilie: Fixed in 232c144.

ykshim_client/src/prod_api.rs
vext01: https://github.com/softdevteam/yk/pull/260

internal_ws/ykpack/src/types.rs
vext01: https://github.com/softdevteam/yk/pull/261

internal_ws/ykcompile/src/lib.rs
vext01: Yes. I checked in the rustc code to be sure.

Looking at for example a MIR call terminator:
```
         /// Cleanups to be done if the call unwinds.                                                                                                                                                          
        cleanup: Option<BasicBlock>,                                                                                              
```

vext01: Presumably it's mostly drops that happen on the cleanup path. I may be able to write a test which checks that no drop is run (which implies abort strategy) and if we change strategy later it would fail.

Thoughts?

ltratt: OK, perfect, that means that we can clarify this to something like "We assume the abort strategy so we don't call cleanup operations if the stack unwinds." I wonder if we can put that in as an `assert` (maybe even a static assert e.g. https://crates.io/crates/static_assertions ?) *iff* we're not compiled with abort? I wouldn't be surprised if that's visible as a `cfg` flag?

vext01: There's a spanner in the works. As hard as I try, I cannot get tests to use the abort strategy. They appear to always use unwinding.

I had planned for a test like:
```
#[test]                                                                                                                                                                                                   
fn abort_strategy() {                                                                                                                                                                                     
    #[cfg(not(panic = "abort"))]                                                                                                                                                                          
    panic!("not using the abort strategy");                                                                                                                                                               
}                                    
```

This test always fails, regardless of strategy selection in `Cargo.toml` and I've verified this is only a problem with tests. Also verified outside of the Yorick build system. Tests always use unwind. I think the test harness relies on the ability to catch a panic to detect failures.

This may pose a problem for us, as it means our tests are in fact using unwind, but the JIT assumes abort...

I couldn't find much info on this, but maybe [this](https://github.com/rust-lang/rust/issues/67650) looks relevant. It looks like they abort strategy tests to work eventually, but it's not yet fully implemented.

ltratt: > I think the test harness relies on the ability to catch a panic to detect failures.

Oh, that is a good point! Hmm.

It looks like the only roadblock to `-Zpanic-abort-tests` is https://github.com/rust-lang/rust/issues/68936 ?

Alternatively, perhaps we do need to move to unwinding in the long term, in which case we could perhaps encode unwinding information that, for the time being, aborts execution (ironically...) so that we don't actually have to implement it now? I'm not really sure, and I don't have a firm opinion yet on this.

vext01: Suppose that the interpreter (in a test) uses unwinding, and the JIT uses abort. Does it matter? It just means we can't catch panics in the JIT, doesn't it?

If this is still an issue, we should probably move to unwinding.

vext01: I wonder if @bjorn3 has any thoughts?

ltratt: I'm not super familiar with the innards of libunwind (or whatever Rust uses for unwinding), but I assume that it just picks up metadata and does its thing. If so, I don't think you can different parts compiled differently, as such; but we can maybe set a flag that says "if you hit this call frame, abort". I'd be fine with that as a temporary hack.

vext01: In normal rust [only one panic runtime is loaded at a time](https://www.ralfj.de/blog/2019/11/25/how-to-panic-in-rust.html). But our JIT is a separate entity with its own semantics.

A temporary hack is probably best, as unwinding out of jitcode sounds scary.

I'm going to need to investigate what a panic looks like in the JIT, so I don't think this should block closing this audit PR. I've raised an issue so it's not forgotten:
https://github.com/softdevteam/yk/issues/262

bjorn3: I don't think rust-lang/rust#68936 is something to be really worried about. In cg_clif I always enable `-Zpanic-abort-tests`.

internal_ws/ykbh/src/lib.rs
ptersilie: Fixed in #263.

internal_ws/ykpack/src/types.rs
ptersilie: Fixed in #263.

internal_ws/yktrace/src/swt.rs
bjorn3: Yes. In fact I think I already removed it.

ptersilie: Yeah, looks like it's gone.

internal_ws/ykbh/src/lib.rs
ptersilie: Refactored in #266.

bjorn3:
Is there anything not yet fixed?

ltratt:
@ptersilie Any thoughts on @bjorn3's question?

ptersilie:
I've resolved all the issues I've dealt with. There's still 3 other discussions which I think @vext01 has fixed, but not yet marked as resolved. So I think this PR is done, though I'd like @vext01 to confirm as well.

internal_ws/ykbh/src/lib.rs
vext01: What's the question? We don't want a variable name with an `_`?

We can replace the line with `..,`, but I don't think it add's much value.

ptersilie: I think we can just ignore this one. Just mark the remaining open discussions as resolved and I think we can close this.

vext01:
Resolved discussions.

vext01:
Sorry, wrong button.

internal_ws/ykbh/src/lib.rs
ltratt: Or even just `destination: _`.

vext01:
Can we close this?

ltratt:
I believe this is now completed, so I'm closing it.

PR #227, opened 2021-02-09
Allow yk to be consumed like a normal Rust dependency.

vext01:
For the `cargo build` target, we can make `build.rs` recursively run
`cargo build` in the internal workspace and then symlink `libykshim.so`
into the consumer's target directory. This avoids the need for consumers
to have special build logic.

We have to keep xtask around for running other targets
(`test`/`fmt`/...) recursively on both workspaces. I had hoped to have
`build.rs` do this automatically too, but we would need to know which
target has been requested and this information isn't exposed to
`build.rs`.

Tested with `ykbf`, which can now be built with simply:
```
RUSTFLAGS="-C tracer=hw" cargo build
```

Whereas previous attempts had tried using complicated wrapper scripts,
which each consumer in turn would need to duplicate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,18 @@ The compiler repo can be found [here](https://github.com/softdevteam/ykrustc).
 To get started, see
 [this section in the documentation](https://softdevteam.github.io/ykdocs/dev/getting_started.html).
 
-This repo uses xtask instead of regular cargo, so instead of running `cargo
-<target>`, you will instead need to execute `cargo xtask <target>`.
+To work with this repository, instead of running `cargo <target>`, instead
+execute `cargo xtask <target>`.
+
+You will also need to add the `-C tracer=<kind>` flag to `RUSTFLAGS` to tell
+the build system what kind of tracer you want to use (`hw` or `sw`, although
+only `hw` works at the moment).
+
+For example, to test using hardware tracing:
+```
+$ export RUSTFLAGS="-C tracer=hw"
+$ cargo xtask test
+```
 
 ## Contributors
 
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -5,3 +5,4 @@ authors = ["The Yorick Developers"]
 edition = "2018"
 
 [dependencies]
+regex = "1.4.3"
--- a/ykrt/build.rs
+++ b/ykrt/build.rs
@@ -1,6 +1,61 @@
+use regex::Regex;
 use std::env;
+use std::os::unix::fs::symlink;
+use std::path::PathBuf;
+use std::process::Command;
+
+include!("../build_aux.rs");
+
+const YKSHIM_SO: &str = "libykshim.so";
 
 fn main() {
+    let ykrt_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+    let mut internal_dir = ykrt_dir.clone();
+    internal_dir.push("..");
+    internal_dir.push("internal_ws");
+    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let cargo = env::var("CARGO").unwrap();
+    let rust_flags = env::var("RUSTFLAGS").unwrap_or_else(|_| String::new());
+    let tracing_kind = find_tracing_kind(&rust_flags);
+
+    // Only build the internal workspace now if we are not using xtask, i.e. if we are being
+    // consumed as a dependency. This means we see the compiler output for the internal workspace
+    // correctly when working directly with the yk repo (via xtask).
+    if env::var("YK_XTASK").is_err() {
+        let status = Command::new(cargo)
+            .current_dir(&internal_dir)
+            .arg("build")
+            .arg("--release")
+            .env(
+                "RUSTFLAGS",
+                &format!("--cfg tracermode=\"{}\"", tracing_kind),
+            )
+            .spawn()
+            .unwrap()
+            .wait()
+            .unwrap();
+        if !status.success() {
+            eprintln!("building internal workspace failed");
+            std::process::exit(1);
+        }
+    }
+
+    // If we symlink libykshim.so into the target dir, then this is already in the linker path when
+    // run under cargo. In other words, the user won't have to set LD_LIBRARY_PATH.
+    let mut sym_dest = out_dir.clone();
+    sym_dest.push("..");
+    sym_dest.push("..");
+    sym_dest.push("..");
+    sym_dest.push(YKSHIM_SO);
+    if !PathBuf::from(&sym_dest).exists() {
+        let mut sym_src = internal_dir.clone();
+        sym_src.push("target");
+        sym_src.push("release");
+        sym_src.push(YKSHIM_SO);
+        dbg!(&sym_src, &sym_dest);
+        symlink(sym_src, sym_dest).unwrap();
+    }
+
     println!(
         "cargo:rustc-link-search={}/../internal_ws/target/release/",
         env::current_dir().unwrap().to_str().unwrap()
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -14,7 +14,7 @@ RUSTFLAGS="${RUSTFLAGS} -D warnings"
 tar jxf /opt/ykrustc-bin-snapshots/ykrustc-${STD_TRACER_MODE}-stage2-latest.tar.bz2
 export PATH=`pwd`/ykrustc-stage2-latest/bin:${PATH}
 
-cargo xtask test
+RUSTFLAGS="-C tracer=${STD_TRACER_MODE}" cargo xtask test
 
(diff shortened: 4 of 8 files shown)


ptersilie:
> `RUSTFLAGS="-C tracer=hw" cargo build`
At the moment I don't seem to require `RUSTFLAGS`. Is this now needed again? Could be not add this flag to the `build.rs` too, so we can just do `cargo build`, similar to how now I just do `cargo xtask build`?

ykrt/build.rs
```diff
@@ -1,6 +1,59 @@
...
+    let mut internal_dir = ykrt_dir.clone();
+    internal_dir.push("..");
+    internal_dir.push("internal_ws");
+    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let cargo = env::var("CARGO").unwrap();
+    let rust_flags = env::var("RUSTFLAGS").unwrap_or_else(|_| String::new());
+    let tracing_kind = find_tracing_kind(&rust_flags);
+
+    // Build the internal workspace.
+    env::set_current_dir(&internal_dir).unwrap();
```
bjorn3, line 22: Always running `cargo build` inside `build.rs` will consume the output of the build process unless it fails, so you won't see any warnings, nor will you get any progress indication. Maybe `build.rs` could take up the responsibility only if you aren't using `cargo xtask`?

ykrt/build.rs
```diff
@@ -1,6 +1,59 @@
...
+    let status = Command::new(cargo)
+        .arg("build")
+        .arg("--release")
+        .env_remove("RUSTFLAGS")
+        .env(
```
bjorn3, line 27: This is redundant.

vext01:
> At the moment I don't seem to require RUSTFLAGS. Is this now needed again? Could be not add this flag to the build.rs too

Afraid we need to specify this again. `build.rs` [doesn't allow adding of arbitrary rust flags](https://github.com/rust-lang/rfcs/issues/1766).

ykrt/build.rs
vext01: Is there a way to know if we are using xtask?

I also tried having `build.rs` run `xtask` but it deadlocks, as the two cargos fight for the lock file :(

README.md
```diff
@@ -13,8 +13,28 @@ The compiler repo can be found [here](https://github.com/softdevteam/ykrustc).
 To get started, see
 [this section in the documentation](https://softdevteam.github.io/ykdocs/dev/getting_started.html).
 
-This repo uses xtask instead of regular cargo, so instead of running `cargo
-<target>`, you will instead need to execute `cargo xtask <target>`.
+This repo contains two separately-compiled workspaces, and thus the build
+system is a little non-standard. To act recursively on both workspaces, instead
```
ptersilie, line 17: All this talk about "recursion" seems to me like an implementation detail that isn't of much importance to whoever indents to use/build this. Should we simplify these descriptions by removing any mention to recursion and just say something like: "To build both workspaces, use `cargo xtask target` instead of `cargo target`."

ykrt/build.rs
bjorn3: xtask could set an env var.

vext01: We could make warnings become errors in build.rs.

README.md
```diff
@@ -13,8 +13,28 @@ The compiler repo can be found [here](https://github.com/softdevteam/ykrustc).
...
+$ export RUSTFLAGS="-C tracer=hw"
+$ cargo xtask test
+```
+
+## Consuming yk.
+
+This repo can be consumed as a regular dependency, despite the dual-workspace
+model.
+
+(This works because `cargo build` is already recursive via `build.rs`. We were
```
ptersilie, line 35: This also seems more like an internal comment than something that is worth mentioning in a ReadMe, as it is of no consequence to the end user.

xtask/Cargo.toml
```diff
@@ -5,3 +5,4 @@ authors = ["The Yorick Developers"]
 edition = "2018"
 
 [dependencies]
+regex = "1"
```
ptersilie, line 8: Is this dependency really just `1`?

README.md
vext01: I did start with something like that, but I thought people would ask "so how comes it works when I consume yk? Dependencies can't run xtask!".

ykrt/build.rs
bjorn3: That would still not give a progress indication. In addition during development it is often useful to have warnings be warnings while still seeing them.

xtask/Cargo.toml
bjorn3: The changelog doesn't mention any new features used here.

README.md
ptersilie: I don't know why anyone would ask this. The README says this is how you build it so that's what you have to do. ;-) And if they are still interested they can go read the code and its comment. My worry is that it's more likely to confuse people reading the README about what any of this means and why they should know about it, if all they want to do is build the thing.

xtask/Cargo.toml
vext01: Yep!

vext01: Wait, I'm wrong. We agreed to expand the full version numbers, as that's what `cargo upgrade` does. My bad.

ykrt/build.rs
vext01: Why is the output different actually? Does rustc know that it's attached to a pipe, or?

vext01: > xtask could set an env var.

c8823b7c0 implements this. Seems to work.

ykrt/build.rs
```diff
@@ -1,6 +1,62 @@
...
+    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let cargo = env::var("CARGO").unwrap();
+    let rust_flags = env::var("RUSTFLAGS").unwrap_or_else(|_| String::new());
+    let tracing_kind = find_tracing_kind(&rust_flags);
+
+    // Only build the internal workspace now if we are not using xtask, i.e. if we are being
+    // consumed as a dependency. This means we see the compiler output for the internal workspace
+    // correctly when working directly with the yk repo (via xtask).
+    if env::var("YK_XTASK").is_err() {
+        env::set_current_dir(&internal_dir).unwrap();
```
bjorn3, line 25: You can use `cmd.current_dir(&internal_dir)` instead.

ykrt/build.rs
```diff
@@ -1,6 +1,62 @@
...
+            .env(
+                "RUSTFLAGS",
+                &format!("--cfg tracermode=\"{}\"", tracing_kind),
+            )
+            .spawn()
+            .unwrap()
+            .wait()
+            .unwrap();
+        if !status.success() {
+            panic!("building internal workspace failed");
```
bjorn3, line 39:
```suggestion
            eprintln!("building internal workspace failed");
            std::process::exit(1);
```

Avoids a backtrace when `RUST_BACKTRACE=1` is set.

ykrt/build.rs
vext01: Just the `env_remove` is redundant, right?

vext01:
I think this addresses all review comments.

README.md
```diff
@@ -13,9 +13,8 @@ The compiler repo can be found [here](https://github.com/softdevteam/ykrustc).
 To get started, see
 [this section in the documentation](https://softdevteam.github.io/ykdocs/dev/getting_started.html).
 
-This repo contains two separately-compiled workspaces, and thus the build
-system is a little non-standard. To act recursively on both workspaces, instead
-of running `cargo <target>`, instead execute `cargo xtask <target>`.
+To work with this repository, instead of running `cargo <target>`, instead
```
ptersilie, line 16: Didn't you say offline that you need to use `cargo build` without `xtask` if you just want to build? Or can I just always use `xtask` in either case?

README.md
```diff
@@ -13,9 +13,8 @@ The compiler repo can be found [here](https://github.com/softdevteam/ykrustc).
...
+execute `cargo xtask <target>`.
 
 You will also need to add the `-C tracer=<kind>` flag to `RUSTFLAGS` to tell
 the build system what kind of tracer you want to use (`hw` or `sw`, although
```
ptersilie, line 20: There's still a mention of `recursively` below.

README.md
vext01: You can do either, but it's simpler to say "always use xtask".

vext01:
Try this.

ykrt/build.rs
bjorn3: Correct

README.md
ptersilie: Agreed.

ptersilie:
If @bjorn3 is happy, please squash.

bjorn3 (approved):

vext01:
Splat.

PR #228, opened 2021-02-09
Implement binary operations and asserts.

ptersilie:
These changes also uncovered a bug in TIR: we didn't include function
arguments in the live locals, which meant that their values weren't
copied over during a guard failure.

### DIFF
--- a/internal_ws/ykbh/src/lib.rs
+++ b/internal_ws/ykbh/src/lib.rs
@@ -1,7 +1,7 @@
 use std::alloc::{alloc, dealloc, Layout};
 use std::convert::TryFrom;
 use ykpack::{
-    self, CallOperand, Constant, ConstantInt, IPlace, Local, Statement, Terminator, TyKind,
+    self, BinOp, CallOperand, Constant, ConstantInt, IPlace, Local, Statement, Terminator, TyKind,
     UnsignedInt, UnsignedIntTy,
 };
 use yktrace::sir::SIR;
@@ -139,6 +139,43 @@ impl LocalMem {
     }
 }
 
+/// Binary operations macros.
+macro_rules! make_binop {
+    ($name: ident, $type: ident) => {
+        fn $name(
+            &mut self,
+            dest: &IPlace,
+            op: &BinOp,
+            opnd1: &IPlace,
+            opnd2: &IPlace,
+            checked: bool,
+        ) {
+            let a = $type::try_from(self.read_int(opnd1)).unwrap();
+            let b = $type::try_from(self.read_int(opnd2)).unwrap();
+            let locals = self.locals_mut();
+            let ptr = locals.iplace_to_ptr(dest);
+            let (v, of) = match op {
+                BinOp::Add => a.overflowing_add(b),
+                BinOp::Lt => ($type::from(a < b), false),
+                _ => todo!(),
+            };
+            if checked {
+                // Write overflow result into result tuple.
+                let ty = SIR.ty(&dest.ty());
+                let tty = ty.unwrap_tuple();
+                let flag_off = isize::try_from(tty.fields.offsets[1]).unwrap();
+                unsafe {
+                    std::ptr::write::<u8>(ptr.offset(flag_off), u8::from(of));
+                }
+            } else if of {
+                todo!("Raise error.")
+            }
+            let bytes = v.to_ne_bytes();
+            locals.write_val(ptr, bytes.as_ptr(), bytes.len());
+        }
+    };
+}
+
 /// A interpreter stack frame, containing allocated memory for the frames locals, and the function
 /// symbol name and basic block index needed by the interpreter to continue interpreting after
 /// returning from a function call.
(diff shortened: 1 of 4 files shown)


internal_ws/ykbh/src/lib.rs
```diff
@@ -326,11 +374,42 @@ impl SIRInterpreter {
...
+                expected,
+                target_bb,
+            } => {
+                let b = if self.read_int(cond) == 1 {
+                    true
+                } else {
+                    false
+                };
+                if b != *expected {
+                    todo!("Raise error")
```
vext01, line 388: Laurie would ask us to drop the message.

ptersilie: I initially has a comment "FIXME raise error", but that's 2 lines vs 1.

ptersilie: I think there's a difference where it's clear what the "todo" is, whereas here I think it's not immediately obvious.

vext01: You could do: `todo!(); // FIXME ...`?

ptersilie: I don't really follow the reasoning behind this to be honest.

vext01:
This looks fine. Just whether the todos should have a message or not. I don't mind, but I think Laurie prefers without?

internal_ws/ykbh/src/lib.rs
vext01: It just reduces the binary size, as one less static string needs to be interned.

FWIW, I prefer the immediacy of the message.

ptersilie: It also makes it easier to find the `todo!` within a macro, as the line number in the error message isn't helpful.

ptersilie: Funnily enough, I didn't update the `todo!` in the macro. I'm gonna update both: remove the string from this one, but add a string to the one in the macro.

ptersilie: See 77d34dc.

ltratt: In general I think that strings in `todo` are pointless, but there are always exceptions. In this case, I think the *current* string is pointless but @ptersilie makes a good point that debugging macros in Rust is hellish. *If* the string was something like `todo!("{:?}"` then it would make sense, because now it genuinely improves readability.

ltratt:
Would I be an idiot if I said that I couldn't see any tests for any of the possibilities here?

vext01:
Not sure what you mean.

This fixes the test that is now unmarked `#[ignore]`. All other cases remain `todo!` don't they?

ptersilie:
I hope @ltratt doesn't mean he expects to see a test for each new `todo!`.

ltratt:
> This fixes the test that is now unmarked #[ignore].

Yes, that's good, but we should have tests for *all* the binary operations that this now enables, even if they're really simple. That way if somebody comes along in the future and, for example, optimises `+`, there's a test that stands a chance of catching problems.

ltratt:
> I hope @ltratt doesn't mean he expects to see a test for each new todo!.

No, that's not necessary, but we should test all the possibilities (I presume `*`/`-`/`*` etc?) that are now enabled for u8.

ptersilie:
I tend to pair tests and implementations together in the same PR.

ptersilie:
> No, that's not necessary, but we should test all the possibilities (I presume `*`/`-`/`*` etc?) that are now enabled for u8.

Ah, I see what you mean now. This PR doesn't enable any of those operations. At the moment, we can only do `+` and `<`, anything else results in a `todo` error.

ltratt:
> At the moment, we can only do + and <

Aha, yes, I can see now. I'd still like to see explicit tests for these added, because someone might delete the current code that just so happens to use them, and then they'll become untested. Adding explicit tests will also make it easier to add extra tests for the other binary operations going forward.

ptersilie:
> because someone might delete the current code that just so happens to use them

I think this is true for all `todo!`s though, but see if you are ok with the latest commit: 8e22b96

ptersilie:
Okay, now I'm confused. I thought you wanted tests for the `todo` cases, in case someone removes the `todo`. If I remove the non-working ones then I'm not sure what purpose that last commit has, because we were already testing `+` and `<`.

ltratt:
> we were already testing + and <.

I *think* -- but correct me if I'm wrong! -- that `+` and `<` are tested as part of something else (guard failure?). If that "something else" disappears, our tests of `+` and `<` also disappear. So I'm suggesting having stand-alone tests that guarantee that we test them.

ptersilie:
> I _think_ -- but correct me if I'm wrong! -- that `+` and `<` are tested as part of something else (guard failure?). If that "something else" disappears, our tests of `+` and `<` also disappear. So I'm suggesting having stand-alone tests that guarantee that we test them.

Riiiiight! Now I got you. Sorry for being slow. Will remove the non-working cases.

ltratt:
I wasn't clear enough -- sorry!

ptersilie:
Okay, try this: f17e36c

ltratt:
Perfect :) Please squash.

ptersilie:
Ok, need to rebase as well.

ltratt:
Yep!

ptersilie:
Squashed and rebased.

PR #231, opened 2021-02-10
Fix software tracing

bjorn3:
Based on https://github.com/softdevteam/yk/pull/230 for now, but could be merged independently.

PR #232, opened 2021-02-10
Xtask refactor

bjorn3:
Together with a patched cg_clif and registering the custom attributes, this makes testing using cg_clif possible with `YK_EXT_RUSTFLAGS='--cfg tracermode="sw" -Clink-arg=-Wl,--export-dynamic -Cllvm-args=tracer=sw' ../rustc_codegen_cranelift/build/cargo.sh xtask test -p tests`.

Based on #231

### DIFF
--- a/build_aux.rs
+++ b/build_aux.rs
@@ -1,18 +1,29 @@
 // Code shared by `ykrt/build.rs` and `xtask/src/main.rs`.
 
-/// Determine what kind of tracing the user has requested at compile time, by looking at RUSTFLAGS.
-/// Ideally, we'd have the user set another environment variable, and then set RUSTFLAGS
-/// accordingly, but you cant' set arbitrary RUSTFLAGS from build.rs.
+/// Determine what kind of tracing the user has requested at compile time, by looking at the output
+/// of `rustc --print cfg`. Ideally, we'd have the user set a environment variable, and then set
+/// RUSTFLAGS accordingly, but you can't set arbitrary RUSTFLAGS from build.rs.
+/// This doesn't look for `-Ctracer` in RUSTFLAGS as different backends may use different methods
+/// of setting the tracing kind.
 fn find_tracing_kind(rustflags: &str) -> String {
-    let re = Regex::new(r"-C\s*tracer=([a-z]*)").unwrap();
-    let mut cgs = re.captures_iter(&rustflags);
+    let cfgs =
+        std::process::Command::new(std::env::var("RUSTC").unwrap_or_else(|_| "rustc".to_string()))
+            .args(&["--print", "cfg"])
+            .args(rustflags.split(" "))
+            .output()
+            .unwrap()
+            .stdout;
+    let cfgs = String::from_utf8(cfgs).unwrap();
+
+    let re = regex::Regex::new(r#"tracermode="([a-z]*)"\n"#).unwrap();
+    let mut cgs = re.captures_iter(&cfgs);
     let tracing_kind = if let Some(caps) = cgs.next() {
         caps.get(1).unwrap().as_str()
     } else {
         panic!("Please choose a tracer by setting `RUSTFLAGS=\"-C tracer=<kind>\"`.");
     };
     if cgs.next().is_some() {
-        panic!("`-C tracer=<kind>` was specified more than once in $RUSTFLAGS");
+        panic!("Tracer mode was specified more than once in $RUSTFLAGS");
     }
     tracing_kind.to_owned()
 }
@@ -20,6 +31,6 @@ fn find_tracing_kind(rustflags: &str) -> String {
 /// Given the RUSTFLAGS for the external workspace, make flags for the internal one.
 fn make_internal_rustflags(rustflags: &str) -> String {
     // Remove `-C tracer=<kind>`, as this would stifle optimisations.
-    let re = Regex::new(r"-C\s*tracer=[a-z]*").unwrap();
+    let re = regex::Regex::new(r"-C\s*tracer=[a-z]*").unwrap();
     re.replace_all(rustflags, "").to_string()
 }
--- a/build_aux.rs
+++ b/build_aux.rs
@@ -30,7 +30,9 @@ fn find_tracing_kind(rustflags: &str) -> String {
 
 /// Given the RUSTFLAGS for the external workspace, make flags for the internal one.
 fn make_internal_rustflags(rustflags: &str) -> String {
-    // Remove `-C tracer=<kind>`, as this would stifle optimisations.
-    let re = regex::Regex::new(r"-C\s*tracer=[a-z]*").unwrap();
+    // Remove `-C tracer=<kind>` and `-C llvm-args=tracer=<kind>` for ykrustc and cg_clif
+    // respectively, as this would stifle optimisations and in case of software tracing would cause
+    // the tracing callback to recursively call into itself.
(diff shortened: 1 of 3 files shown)


vext01:
If we can rebase / sync this one, I'll take a look.

One thing I was wondering was how much of `YK_EXT_RUSTFLAGS='--cfg tracermode="sw" -Clink-arg=-Wl,--export-dynamic -Cllvm-args=tracer=sw' ../rustc_codegen_cranelift/build/cargo.sh xtask test -p tests` can be hidden from the user.

With the `xtask` stuff, I tried to make it so that the user only has to set `-C tracer=...`. Would it be possible here?

bjorn3:
Stock rustc as used by cg_clif doesn't recognize `-Ctracer`. The exact invocation that is necessary for cg_clif may change in the future. Especially the `-Clink-arg=-Wl,--export-dynamic` is something I want to get rid of.

vext01:
Does -Cllvm-args=tracer=sw mean that you had to modify LLVM btw?

[comment moved from the wrong PR]

bjorn3:
No, I hijacked `-Cllvm-args` for option passing in cg_clif. I effectively used it as a (non-existent) `-Cbackend-args` option.

vext01:
We just merged a branch that conflicts, sorry :P

Is it hard to sync up?

bjorn3:
Since #231 now also changes this code a bit I would prefer to wait with rebasing until that PR is merged.

xtask/src/main.rs
```diff
@@ -31,108 +31,74 @@ impl Workspace {
     }
 }
 
-/// A build step for one of the workspaces.
```
vext01, line 34: This is just a preference with no functional change? Just removed the `WorkspaceAction` struct, right?

What happened to `forced_deps`? IIRC that is needed to make `cargo test` from clean work.

bjorn3: I am recursively calling `run_action` instead of pushing to `forced_deps`. The removal of `WorkspaceAction` makes it easier to build the `Command` without having to pass around all kinds of fields. Instead the respective method can directly be called. IMO this resulted in much easier to read code.

build_aux.rs
```diff
@@ -1,11 +1,21 @@
 // Code shared by `ykrt/build.rs` and `xtask/src/main.rs`.
 
-/// Determine what kind of tracing the user has requested at compile time, by looking at RUSTFLAGS.
-/// Ideally, we'd have the user set another environment variable, and then set RUSTFLAGS
-/// accordingly, but you cant' set arbitrary RUSTFLAGS from build.rs.
+/// Determine what kind of tracing the user has requested at compile time, by looking at the output
+/// of `rustc --print cfg`. Ideally, we'd have the user set another environment variable, and then
```
vext01, line 4: `s/another//` since we are not talking about an environment variable any more.

FWIW I much prefer this to my string parsing. Nice one!

ykrt/build.rs
```diff
@@ -14,8 +14,10 @@ fn main() {
     internal_dir.push("internal_ws");
     let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
     let cargo = env::var("CARGO").unwrap();
-    let rust_flags = env::var("RUSTFLAGS").unwrap_or_else(|_| String::new());
-    let tracing_kind = find_tracing_kind(&rust_flags);
+    let tracing_kind = env::var("YK_TRACING_KIND").unwrap_or_else(|_| {
```
vext01, line 17: Why the extra env var? So as to not duplicate work?

xtask/src/main.rs
```diff
@@ -111,6 +111,7 @@ fn run_action(workspace: Workspace, target: &str, extra_args: &[String]) {
                 }
             } else {
                 cmd.env("YK_TRACING_KIND", tracing_kind);
+                cmd.env("CARGO_TARGET_DIR", "target_ext");
```
vext01, line 114: I see what you are doing, but I'm not certain why it prevents stuff being rebuilt.

ykrt/build.rs
bjorn3: Because of the `--target $HOST`, `RUSTFLAGS` is no longer passed to the build script.

xtask/src/main.rs
bjorn3: By using a different target dir, this cargo invocation will no longer clobber the target dir used when compiling xtask with crates built using a different `RUSTFLAGS`.

vext01: OK, let's add a comment to this effect.

ykrt/build.rs
vext01: Does it hurt to always call `find_tracing_kind`?

bjorn3: It will panic as `-Ctracer=hw` is no longer passed through to rustc and as such `tracermode="hw"` is no longer found in the output of `rustc --print cfg $RUSTFLAGS`.

bjorn3:
I have rebased, but I am not entirely happy with the interaction between the last commit ("Ensure that xtask and build scripts are never traced") and #235, so you may want to hold off reviewing that commit for now. The rest of the commits should be fine.

ykrt/build.rs
vext01: My last question is whether we can fold `YK_XTASK` and `YK_TRACING_KIND` into one `YK_XTASK_TRACING_KIND=hw/sw`?

.buildbot.sh
```diff
@@ -8,7 +8,8 @@ case ${CI_TRACER_KIND} in
        exit 1;;
 esac
 
-export RUSTFLAGS="-C tracer=${CI_TRACER_KIND} -D warnings"
+export RUSTFLAGS="-D warnings"
+export YK_EXT_RUSTFLAGS="-C tracer=${STD_TRACER_MODE}"
```
vext01, line 12: Maybe it's not so bad...

ykrt/build.rs
bjorn3: For ykrustc with hardware tracing that would work as compiling build scripts with tracing support is fine, but for cg_clif compiling them with software tracing won't work. For this reason you need to pass `--target $HOST` to cargo to make it compile build scripts without `RUSTFLAGS` set. If ykrt is used as external dependency, that would mean you would need something like `RUSTFLAGS="..." YK_TRACING_KIND="sw" cg_clif/cargo.sh build --target x86_64-unknown-linux-gnu`. In this case `YK_XTASK` must not be set.

I think I can try to include the software tracer callback as a weak function though. Hopefully weak symbols can be overridden by symbols provided by dylibs. That would make the whole last commit unnecessary.

vext01: OK. I don't fully understand, but let's see what comes of this...

bjorn3:
I was able to define a dummy software tracing callback for all build scripts and xtask. I have dropped the last commit.

vext01:
OK, so is it time to squash?

vext01:
We should also update:
https://softdevteam.github.io/ykdocs/dev/getting_started.html

bjorn3:
Splat. This shouldn't result in any user visible changes when using ykrustc in combination with hardware tracing.

vext01:
> This shouldn't result in any user visible changes when using ykrustc in combination with hardware tracing.

OK. I think I still need to update this page due to my changes.

vext01:
(And we probably should note down somewhere how to use software tracing, once it's in a more mature state)

PR #235, opened 2021-02-11
Pass down rustflags

vext01:
Took a while to get this right, as the flags seem to be handled differently depending on if the toolchain is overridden in rustup or not.

This seems to work for our common scenarios (for developers and CI).

This means we now catch warnings in the internal workspace during CI.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -8,14 +8,20 @@ case ${CI_TRACER_KIND} in
        exit 1;;
 esac
 
-RUSTFLAGS="${RUSTFLAGS} -D warnings"
+export RUSTFLAGS="-C tracer=${CI_TRACER_KIND} -D warnings"
 
 # Use the most recent successful ykrustc build.
 tar jxf /opt/ykrustc-bin-snapshots/ykrustc-${CI_TRACER_KIND}-stage2-latest.tar.bz2
 export PATH=`pwd`/ykrustc-stage2-latest/bin:${PATH}
 
-RUSTFLAGS="-C tracer=${CI_TRACER_KIND}" cargo xtask test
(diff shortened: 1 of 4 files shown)


bjorn3:
https://github.com/softdevteam/yk/pull/232 touches xtask too. It has a separate env var for rustflags to be passed to the external workspace and to the internal workspace. It doesn't change ykrt/build.rs though.

ltratt:
I'm broadly fine with this but I do wonder if we should document that we're doing string processing somewhere, even if just as a doc comment on a function?

vext01:
I think it's implied here:
https://github.com/softdevteam/yk/pull/235/files#diff-6e80a78b6a797cb4a9bc5dc8af2eee7e2ba3c81db2e81dcd6912ddd5943d04d8R20

But we could add something like:
> Unfortunately this involves string matching on $RUSTFLAGS.

ltratt:
I'd be a bit clearer and say something like "Note that this has to perform string processing on the args, which may not always match the expectations of whatever shell you're using."?

vext01:
What's it got to do with the shell though?

ltratt:
I assume that at some point these arguments are passed to a shell for processing/execution?

ltratt:
This is me being paranoid but if someone passes in something like `--some-arg-that-we-dont-care-about '-C tracer=hw'` we'll process that too right?

vext01:
Yes, by design. We do want to pass down any other flags the user gives.

We will see it, but won't touch it, as it won't match the regex.

ltratt:
OK, I'm not totally sure I understand, but you've thought more deeply than I about this, and it's not a major point.

bors r+

PR #236, opened 2021-02-11
Use Arc<Body> instead of allocating strings.

ptersilie:
Instead of cloning the strings we get from the compiler, reference the body directly, which is in an Arc and thus only costs a pointer.

### DIFF
--- a/internal_ws/ykcompile/src/lib.rs
+++ b/internal_ws/ykcompile/src/lib.rs
@@ -298,11 +298,8 @@ pub extern "sysv64" fn bh_push_vec(
 ) {
     let fname =
         unsafe { std::str::from_utf8(std::slice::from_raw_parts(sym_ptr, sym_len)).unwrap() };
-    let fi = FrameInfo {
-        sym: fname.to_string(),
-        bbidx,
-        mem,
-    };
+    let body = SIR.body(fname).unwrap();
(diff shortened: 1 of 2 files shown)


internal_ws/ykcompile/src/lib.rs
```diff
@@ -298,8 +298,9 @@ pub extern "sysv64" fn bh_push_vec(
 ) {
     let fname =
         unsafe { std::str::from_utf8(std::slice::from_raw_parts(sym_ptr, sym_len)).unwrap() };
+    let body = SIR.body(fname).unwrap();
     let fi = FrameInfo {
-        sym: fname.to_string(),
+        body,
```
ptersilie, line 303: I'm not quite sure if we need to clone the Arc here, since due to the boxing below I think the borrow checker wouldn't catch it if we do. Though since `SIR.body` already creates a clone, I think this is fine. Can't hurt to double check with @vext01 though.

bjorn3: `Arc<Body>` has an `'static` lifetime, so this is fine.

ptersilie:
Fixed formatting.

ltratt:
Please squash.

ptersilie:
Squashed.

vext01:
LGTM!

PR #237, opened 2021-02-15
Use constants for register size and stack alignment.

vext01:
(Also add some missing docs while we are here)

This is a draft, because I have outstanding queries for @ptersilie.

### DIFF
--- a/internal_ws/ykcompile/src/arch/x86_64/mod.rs
+++ b/internal_ws/ykcompile/src/arch/x86_64/mod.rs
@@ -42,8 +42,16 @@ lazy_static! {
     static ref ICTX_REG: u8 = RDI.code();
 
     static ref TEMP_LOC: Location = Location::Reg(*TEMP_REG);
+
+    /// The size of a memory address in bytes.
     static ref PTR_SIZE: u64 = u64::try_from(mem::size_of::<usize>()).unwrap();
 }
+
+/// The size of a quad-word register (e.g. RAX) in bytes.
+const QWORD_REG_SIZE: usize = 8;
+
+/// Number of bytes the stack must be aligned to, immediately before a call via the Sys-V ABI.
+const SYSV_CALL_STACK_ALIGN: usize = 16;
 
 /// Generates functions for add/sub-style operations.
 /// The first operand must be in a register.
@@ -470,7 +478,7 @@ impl TraceCompiler {
             // We've also pushed RAX in the meantime.
             rdi_stackpos += 1;
             dynasm!(self.asm
-                ; mov rsi, [rsp + i32::try_from(rdi_stackpos).unwrap() * 8]
+                ; mov rsi, [rsp + i32::try_from(rdi_stackpos).unwrap() * i32::try_from(QWORD_REG_SIZE).unwrap()]
             );
         } else {
             dynasm!(self.asm
@@ -582,7 +590,7 @@ impl TraceCompiler {
                         // have overwritten the value in the meantime. So we should load the value
                         // back from the stack.
                         dynasm!(self.asm
-                            ; mov Rq(arg_reg), [rsp + idx * 8]
+                            ; mov Rq(arg_reg), [rsp + idx * i32::try_from(QWORD_REG_SIZE).unwrap()]
                         );
                     } else {
                         // We didn't save this register, so it remains intact.
@@ -1214,20 +1222,15 @@ impl TraceCompiler {
         // Reserved stack space for spilled locals during execution.
         let soff = self.stack_builder.size();
         // Reserved memory on the stack to spill live locals to during a guard failure.
-        let live_off = REG_POOL.len() * 8;
+        let live_off = REG_POOL.len() * usize::try_from(QWORD_REG_SIZE).unwrap();
 
-        // As we'll be making calls to other functions using the Sys-V 64 ABI, we need to make sure
-        // the stack is aligned to 16 bytes before the call instruction. Note that the call
-        // instruction itself will push the return address onto the stack, so at the beginning of
-        // the trace the stack will have an extra offset of 8 bytes.
-        // Instead of adjusting the alignment before each call instruction, we do this once at the
-        // beginning of the trace, since we won't be misaligning the stack during the trace
-        // (push/pop is allowed as long as there's no call inbetween them). We also need to
-        // consider the extra space reserved for spilled and live locals as well as the callee
-        // saved registers, when calculating the alignment.
-        // While we could use the stack builder to keep track of the stack and the alignment, this
-        // would require us to split up various dynasm! blocks, which is inconvenient.
-        let topalign = 16 - (live_off + soff as usize + CALLEE_SAVED_REGS.len() * 8) % 16;
+        // After allocating stack space for the trace, we pad the stack pointer up to the next
+        // 16-byte alignment boundary. Calls can use this fact when catering for alignment
+        // requirements of callees (if necessary). Interim pushes and pops in trace code are
+        // allowed as long as they are short-lived and correctly restore 16-byte alignment.
+        let topalign = SYSV_CALL_STACK_ALIGN
+            - (live_off + soff as usize + CALLEE_SAVED_REGS.len() * QWORD_REG_SIZE)
+                % SYSV_CALL_STACK_ALIGN;
         // Restore registers.
         dynasm!(self.asm
             ; mov rax, 0 // Signifies that there were no guard failures.
@@ -1272,7 +1275,10 @@ impl TraceCompiler {
             for (tirlocal, loc) in live_locations.iter_mut() {
                 match loc {
                     Location::Reg(reg) => {
-                        let off = -i32::try_from(soff + u32::from(*reg) * 8).unwrap();
+                        let off = -i32::try_from(
+                            soff + u32::from(*reg) * u32::try_from(QWORD_REG_SIZE).unwrap(),
+                        )
+                        .unwrap();
                         let newloc = Location::Mem(RegAndOffset {
                             reg: RBP.code(),
                             off,


internal_ws/ykcompile/src/lib.rs
```diff
@@ -1358,8 +1366,12 @@ impl TraceCompiler {
         // Reserved stack space for spilled locals during execution.
         let soff = self.stack_builder.size();
         // Reserved memory on the stack to spill live locals to during a guard failure.
-        let live_off = REG_POOL.len() * 8;
+        let live_off = REG_POOL.len() * REG_SIZE;
 
+        // FIXME Can we re-review this whole comment, as I'm sure it could be made shorter, and I'm
+        // not sure it's right to say that the stack must be 16-byte aligned *before* the call. I
+        // thought it was right *after* the call (once the RV has been pushed).
+        //
```
vext01, line 1374: To expand on this, here's a snippet from my notes:

```
> The end of the input argument area shall be aligned on a 16 byte boundary.
In other words, the value(%rsp+ 8)is always a multiple of 16 when control is 
transferred to the function entry point. The stack pointer,%rsp, always
points to the end of the latest allocated stack frame.

https://uclibc.org/docs/psABI-x86_64.pdf
```

So this means that once the return value is pushed (once the call instruction has finished, and the rsp points at the first instr of the callee) the stack is 16-byte aligned, but not *before* the call.

The wording is confusing, but the critical detail is (I think) "end of the input argument area" -- i.e. right before the RV is pushed.

internal_ws/ykcompile/src/lib.rs
```diff
@@ -1371,7 +1383,11 @@ impl TraceCompiler {
         // saved registers, when calculating the alignment.
         // While we could use the stack builder to keep track of the stack and the alignment, this
         // would require us to split up various dynasm! blocks, which is inconvenient.
-        let topalign = 16 - (live_off + soff as usize + CALLEE_SAVED_REGS.len() * 8) % 16;
+        //
+        // FIXME Lukas, what's the `16 - ` for here again?
```
bjorn3, line 1387: The `16 -` should be `SYSV_CALL_STACK_ALIGN -` I think. If for example `live_off + soff as usize + CALLEE_SAVED_REGS.len() * 8` is 23, `% 16` you get 7. To re-align the stack, you need 16 - 7 = 9 bytes extra.

vext01: Makes sense. Let's confirm with Lukas (he's AFK today though).

internal_ws/ykcompile/src/lib.rs
```diff
@@ -57,9 +57,17 @@ lazy_static! {
     static ref ICTX_REG: u8 = RDI.code();
 
     static ref TEMP_LOC: Location = Location::Reg(*TEMP_REG);
+
+    /// The size of a memory address in bytes.
     static ref PTR_SIZE: u64 = u64::try_from(mem::size_of::<usize>()).unwrap();
 }
 
+/// The size of a register in bytes.
+const REG_SIZE: usize = 8;
```
ltratt, line 66: Not all registers are the same size, so this name feels a bit confusing?

vext01: we can prefix it `GP` for "general purpose".

ltratt: Maybe a less-worse way would be to have a `const HashMap` from register names to sizes, or simply one constant per register? Or is this a generic "make sure all registers can fit into this" constant or ... ?

vext01: I'm feel uneasy about maintaining tables of registers, as I think it would make the code even harder to read than it already is, and (at least for now) there's no real need to add indirection.

In general CPUs tend to have "classes" of registers, where each register in a class is the same size. But as usual, it's not that simple. In x86_64, general purpose registers are partitioned into 1-, 2- , 4- and 8-byte registers. It's all an upshot of legacy.

In the above case, a better name for my constant would be `QWORD_REG_SIZE`, where `Q` stands for quad-word (because in x86_64, strictly a word is 16-bytes).

ltratt: QWORD_REG_SIZE seems reasonable.

vext01: Fixed in e278c3a

internal_ws/ykcompile/src/lib.rs
```diff
@@ -1386,7 +1386,9 @@ impl TraceCompiler {
         //
         // FIXME Lukas, what's the `16 - ` for here again?
         let topalign = 16
-            - (live_off + soff as usize + CALLEE_SAVED_REGS.len() * REG_SIZE)
+            - (live_off
+                + soff as usize
+                + CALLEE_SAVED_REGS.len() * usize::try_from(QWORD_REG_SIZE).unwrap())
```
ltratt, line 1391: This is a `usize` so the conversion seems pointless?

vext01: Oops.

It's because I toyed with making our constant an `i32` so that it could be used with dynasm directly, but that caused other places to require casting.

I hate juggling integer types in rust. I wish `as` was actually a "try_from().unwrap()".

ltratt: > I hate juggling integer types in rust. I wish as was actually a "try_from().unwrap()".

I wish `as` didn't exist :p

ltratt:
I'm fine with this, so I'll now leave you to the capable hands of @ptersilie.

internal_ws/ykcompile/src/lib.rs
ptersilie: What @bjorn3 said.

internal_ws/ykcompile/src/lib.rs
ptersilie: I thought we have put this topic to rest. But the fact we are still discussing this is a testament to how awful that part of the SYS-V documentation is written. Did you notice that your two paragraphs contradict each other? Your first one says we need to align to 16 bytes after the return value, and your second paragraph says it needs to be before.

How I understood the documentation (and we confirmed this by looking it the assembler of some test programs) is that we need to align to 16 bytes immediately **before** we do the call. This means that the we are aligned to 16 bytes at the end of the input argument area. The call then pushes the return value onto the stack, misaligning it by 8 bytes. However that means that `%rsp + 8` is a multiple of 16 bytes, like the rest of the documentation specifies.

They should definitely fix those docs, e.g. "In other words, the value(%rsp+ 8)is always a multiple of 16 ~~when~~ before control is transferred to the function entry point."

ltratt: Is it possible to propose changes to the sys-v docs or is it hidden behind an impenetrable standards organisation?

ptersilie: Seems it's on Gitlab: https://gitlab.com/x86-psABIs/x86-64-ABI/-/blob/master/x86-64-ABI/low-level-sys-info.tex#L481

Though, before raising a PR they ask you to discuss the changes on their mailing list. I can give that a go.

ltratt: I think it's worth putting a small amount of effort into this, if nothing else to save us heartache in the future :) But if there's much resistance to a change, I'd probably give up: some standards committees are welcoming of changes, some aren't.

vext01: > I think it's worth putting a small amount of effort into this.

Agreed.

vext01:
@ptersilie are you OK with 073ffe6?

internal_ws/ykcompile/src/lib.rs
```diff
@@ -1368,24 +1368,22 @@ impl TraceCompiler {
...
-        // thought it was right *after* the call (once the RV has been pushed).
-        //
         // As we'll be making calls to other functions using the Sys-V 64 ABI, we need to make sure
-        // the stack is aligned to 16 bytes before the call instruction. Note that the call
-        // instruction itself will push the return address onto the stack, so at the beginning of
-        // the trace the stack will have an extra offset of 8 bytes.
+        // the stack is correctly aligned. We must ensure that the stack pointer is 16-byte aligned
+        // immediately before the call instruction is executed.
+        //
+        // FIXME The following is not going to work in the long run, as we will have to push to the
```
ptersilie, line 1375: Do you mean calls we do in `c_call`? The problem with `c_call` is that it calls Rust functions, and there is no nice way to wrap those into sys-v calls, so we've agreed to use the unofficial/unstable Rust ABI for those, so I don't think we need to worry about this. However, we can assume that when we enter `c_call`, the stack is 16-bytes aligned, since it's aligned at the beginning of the trace, and we don't misalign it anywhere (i.e. if we need to push to the stack we always pop immediately after before compiling further instructions).

internal_ws/ykcompile/src/lib.rs
```diff
@@ -57,9 +57,17 @@ lazy_static! {
...
     static ref TEMP_LOC: Location = Location::Reg(*TEMP_REG);
+
+    /// The size of a memory address in bytes.
     static ref PTR_SIZE: u64 = u64::try_from(mem::size_of::<usize>()).unwrap();
 }
 
+/// The size of a quad-word register (e.g. RAX) in bytes.
+const QWORD_REG_SIZE: usize = 8;
+
+/// Number of bytes the stack must be aligned by, immediately after a call via the Sys-V ABI.
```
ptersilie, line 68: This is incorrect. It should be: `Bytes the stack must be aligned to, immediately before a call via the Sys-V ABI.`

vext01: Fixed in 8681961

internal_ws/ykcompile/src/lib.rs
vext01: I would assume that c_call could call anything, via any ABI (although obviously its implementation is incomplete).

It wouldn't surprise me, for example, if there is some sysv function in the stdlib (that we can't trace).

ptersilie: Fair point. But this doesn't change the fact that we will be aligned when we enter the `c_call` function. So if the call requires arguments to be pushed onto the stack, then we just need to count them and make sure we align again before compiling the `call` instruction.

vext01: Good point. I think the wording just needs to be tweaked somehow. Perhaps you can propose something?

internal_ws/ykcompile/src/lib.rs
```diff
@@ -1358,20 +1366,26 @@ impl TraceCompiler {
...
+        // the stack is correctly aligned. We must ensure that the stack pointer is 16-byte aligned
+        // immediately before the call instruction is executed.
+        //
+        // FIXME The following is not going to work in the long run, as we will have to push to the
+        // stack in order to set up arguments for calls to functions with >6 args. We will have to
+        // align at each call site.
+        //
         // Instead of adjusting the alignment before each call instruction, we do this once at the
         // beginning of the trace, since we won't be misaligning the stack during the trace
         // (push/pop is allowed as long as there's no call inbetween them). We also need to
```
ptersilie, line 1381: Maybe add after `there's no call inbetween them).`: `Note however that if we need to compile a call that requires stack arguments the stack needs to be re-aligned`. Alternatively, we can leave a comment in `c_call` like "Note, that even though the trace is already 16-byte aligned, that calls requiring stack arguments need to re-align the stack."

vext01: I'm going to propose something a little more brutal. Watch this space.

vext01: How's e671d1c?

ptersilie: Love it!

ptersilie:
Please squash, rebase, etc.

vext01:
splat.

PR #238, opened 2021-02-15
Start abstracting platforms in ykcompile.

vext01:
**Draft for discussion**

[but note that this works: all tests pass]

This change lays the foundations for platform abstration:

 - Moves the (currently x86_64-centric) `TraceCompiler` struct into
   arch/x86_64/mod.rs.

 - Leaves all generic structs in the top-level to be shared across
   future backends.

 - For now, hard-wires use of the x86_64 backend.

In the long run, we should slice up `TraceCompiler` into some kind of a
modular API (statically dispatched trait?), but this change does not
address this.


ltratt:
This looks like a good, simple, start. If @ptersilie is OK with it, let's merge.

vext01:
@ptersilie what do you think of this one?

ptersilie:
Looks good.

PR #241, opened 2021-02-16
find_symbol() stuff.

vext01:
See commits.

### DIFF
--- a/internal_ws/ykcompile/src/arch/x86_64/mod.rs
+++ b/internal_ws/ykcompile/src/arch/x86_64/mod.rs
@@ -615,6 +615,8 @@ impl TraceCompiler {
         let sym_addr = if let Some(addr) = self.addr_map.get(sym) {
             *addr as i64
         } else {
+            // This path is required for system calls. hwtracer doesn't trace through the kernel,
+            // so system call addresses will never appear in addr_map.
             TraceCompiler::find_symbol(sym)? as i64
         };
         dynasm!(self.asm
@@ -1423,6 +1425,11 @@ impl TraceCompiler {
     }
 
     /// Returns a pointer to the static symbol `sym`, or an error if it cannot be found.
+    ///
+    /// In general, there is no guarantee that a symbol will be found, especially when dealing with
+    /// binary executables where there's no public API to speak of. In such cases the Rust compiler
+    /// is free to (and often does) throw away symbols. Note however that symbols marked
+    /// `#[no_mangle] are always exported.
     pub fn find_symbol(sym: &str) -> Result<*mut c_void, CompileError> {
         let sym_arg = CString::new(sym).unwrap();
         let addr = unsafe { dlsym(RTLD_DEFAULT, sym_arg.into_raw()) };
--- a/internal_ws/ykcompile/src/arch/x86_64/mod.rs
+++ b/internal_ws/ykcompile/src/arch/x86_64/mod.rs
@@ -1,15 +1,13 @@
 //! The Yorick TIR trace compiler.
 
 use crate::{
-    stack_builder::StackBuilder, CompiledTrace, IndirectLoc, Location, RegAlloc, RegAndOffset,
+    find_symbol, stack_builder::StackBuilder, CompileError, CompiledTrace, IndirectLoc, Location,
+    RegAlloc, RegAndOffset,
 };
 use dynasmrt::{x64::Rq::*, DynamicLabel, DynasmApi, DynasmLabelApi, Register};
-use libc::{c_void, dlsym, RTLD_DEFAULT};
 use std::alloc::{alloc, Layout};
 use std::collections::HashMap;
 use std::convert::TryFrom;
-use std::ffi::CString;
-use std::fmt::{self, Display, Formatter};
 use std::mem;
 use std::process::Command;
 use ykbh::{FrameInfo, SIRInterpreter};
@@ -209,20 +207,6 @@ macro_rules! binop_mul_div {
                 ; pop rax
                 ; pop rdx
             );
-        }
-    }
-}
-
-#[derive(Debug, Hash, Eq, PartialEq)]
-pub enum CompileError {
-    /// The binary symbol could not be found.
-    UnknownSymbol(String),
-}
-
-impl Display for CompileError {
-    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        match self {
-            Self::UnknownSymbol(s) => write!(f, "Unknown symbol: {}", s),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -615,7 +599,9 @@ impl TraceCompiler {
         let sym_addr = if let Some(addr) = self.addr_map.get(sym) {
             *addr as i64
         } else {
-            TraceCompiler::find_symbol(sym)? as i64
+            // This path is required for system calls. Currently hwtracer will never trace through
+            // the kernel, so system call addresses will never appear in addr_map.
+            find_symbol(sym)? as i64
```
ltratt, line 604: I know this is the `x86_64` file, but I was curious why `i64` here and not another integer type (`usize` or `u64`).

internal_ws/ykcompile/src/lib.rs
```diff
@@ -146,3 +161,20 @@ impl CompiledTrace {
...
+///
+/// In general, there is no guarantee that a symbol will be found, especially when dealing with
+/// binary executables where there's no public API to speak of. In such cases the Rust compiler is
+/// free to (and often does) throw away symbols. Note however that symbols marked `#[no_mangle] are
+/// always exported.
+pub fn find_symbol(sym: &str) -> Result<*mut c_void, CompileError> {
+    let sym_arg = CString::new(sym).unwrap();
+    let addr = unsafe { dlsym(RTLD_DEFAULT, sym_arg.into_raw()) };
+
+    if addr.is_null() {
```
ltratt, line 175: A minor stylistic issue which *might* also be a minor optimisation: I assume that the `Ok` branch is the common case? In which case I think it's generally clearer to put it in the `true` branch (which can also play better with optimisers, branch predictors, depending on the phase of the moon etc etc).

internal_ws/ykshim/src/test_api.rs
```diff
@@ -104,9 +104,8 @@ unsafe extern "C" fn __ykshimtest_tracecompiler_local_dead(tc: *mut TraceCompile
 
 /// Find a symbol's address in the current memory image. Returns NULL if it can't be found.
```
ltratt, line 105: We need to document explicitly that this consumes a `CString` to maintain the guarantee in https://doc.rust-lang.org/std/ffi/struct.CString.html#method.into_raw. I did wonder if we could even make the input type by `CString` but I guess that doesn't work across the boundary?

internal_ws/ykcompile/src/arch/x86_64/mod.rs
bjorn3: I think dynasm requires it to be signed.

internal_ws/ykcompile/src/lib.rs
bjorn3: LLVM optimizes the following function to the same assembly:

```rust
pub fn foo(a: *const u8) -> Result<*const u8, String> {
    if a.is_null() {
        Err(String::new())
    } else {
        Ok(a)
    }
}

pub fn bar(a: *const u8) -> Result<*const u8, String> {
    if !a.is_null() {
        Ok(a)
    } else {
        Err(String::new())
    }
}
```

In fact it merges the functions into one if they are defined in the same object file.

```asm
playground::bar:
	movq	%rdi, %rax
	testq	%rsi, %rsi
	je	.LBB0_1
	movq	%rsi, 8(%rax)
	xorl	%ecx, %ecx
	movq	%rcx, (%rax)
	retq

.LBB0_1:
	movq	$1, 8(%rax)
	xorps	%xmm0, %xmm0
	movups	%xmm0, 16(%rax)
	movl	$1, %ecx
	movq	%rcx, (%rax)
	retq
.set playground::foo, playground::bar
```

I do agree it is a bit nicer though.

ltratt: At this level of macro-optimisation, what *actually* ends up happening is extremely hard to predict, especially with changing versions of the compiler :) I didn't expect it to merge them into one, however!

bjorn3: Rust uses the `unnamed_addr` flag for all functions, which explicitly allows this optimization. Search for "unnamed_addr" in https://llvm.org/docs/LangRef.html for more info.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
vext01: Yes.

```
  624 |             ; mov Rq(*TEMP_REG), QWORD sym_addr
      |                                        ^^^^^^^^ expected `i64`, found `u64`
```

This is probably leakage from some x86_64 encoding constraint.

internal_ws/ykshim/src/test_api.rs
vext01: Actually, there's no need for this to consume the argument. I'll see if I can make that work.

vext01: bd718ec shows what it would take to make this function (and 2 others) not consume their argument.

I'm not very sure which approach is best. Do you have a preference? How hard should we try to avoid consumption?

vext01: (there's a bug, i missed an `as_ptr()`, but you get the idea -- as_ptr() vs. consumption with documentation).

ltratt: I think this is is better, because not only do we not have to consume things, we don't even have to rely on the various restrictions of `CString`.

ltratt: Oops, "this" is ambiguous: I meant that bd718ec (with your suggested bug fix!) is better.

vext01:
I think that's all comments addressed?

internal_ws/ykcompile/src/lib.rs
ltratt: @vext01 You might have missed this comment?

vext01: I thought you'd come to a conclusion to not adjust this. Shall I change it then?

ltratt: For readability if nothing else, having the common case first seems better?

ptersilie:
Looks good to me.

internal_ws/ykcompile/src/lib.rs
vext01: Fixed in 3272405017.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -615,7 +599,9 @@ impl TraceCompiler {
         let sym_addr = if let Some(addr) = self.addr_map.get(sym) {
             *addr as i64
         } else {
-            TraceCompiler::find_symbol(sym)? as i64
+            // This path is required for system calls. Currently hwtracer will never trace through
```
bjorn3, line 602: > **Currently** hwtracer will never trace through the kernel

Will it ever :)

vext01: Probably not. Fixed in 7398356cd

vext01:
Time to squash?

ltratt:
Please squash.

vext01:
splat.

PR #242, opened 2021-02-17
Fix ykview trace viewer

bjorn3:


vext01:
I wonder why CI didn't catch these?

bjorn3:
ykview doesn't have any tests.

vext01:
Ah, I assumed these were compilation errors.

I was just trying to use ykview, and I'm not certain it's being built by xtask.

bjorn3:
It is built by xtask when using `cargo xtask test`.

PR #245, opened 2021-02-17
Software tracing cleanup

bjorn3:
* Add missing start_tracing(SoftwareTracing)
* Use Vec instead of manual memory management in swt.rs.

### DIFF
--- a/internal_ws/yktrace/src/swt.rs
+++ b/internal_ws/yktrace/src/swt.rs
@@ -2,8 +2,7 @@
 
 use super::{SirTrace, ThreadTracer, ThreadTracerImpl};
 use crate::{errors::InvalidTraceError, sir::SirLoc};
-use libc;
-use std::convert::TryFrom;
+use std::cell::RefCell;
 
 /// Softare thread tracer.
 struct SWTThreadTracer;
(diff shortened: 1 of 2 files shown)


bjorn3:
The second commit was requested in https://github.com/softdevteam/yk/pull/231#discussion_r574527727.

internal_ws/yktrace/src/swt.rs
```diff
@@ -2,74 +2,49 @@
...
+                .collect()
+        });
 
-            // We hand ownership of the trace to the caller. The caller is responsible
-            // for freeing the trace.
-            let ret_trace = TRACE_BUF;
-            len = TRACE_BUF_LEN;
-
-            // Now reset all off the recorder's state.
-            TRACE_BUF = 0 as *mut SwtLoc;
```
vext01, line 35: I think I got this right. The above `drain()` means this stuff isn't needed?

bjorn3: Correct

internal_ws/yktrace/src/swt.rs
```diff
@@ -86,23 +61,10 @@ struct SwtLoc {
...
-/// The number of elements in the trace buffer.
-#[thread_local]
-static mut TRACE_BUF_LEN: usize = 0;
-
-/// The allocation capacity of the trace buffer (in elements).
-#[thread_local]
-static mut TRACE_BUF_CAP: usize = 0;
+thread_local! {
+    /// The trace buffer.
+    static TRACE_BUF: RefCell<Vec<SwtLoc>> = RefCell::new(Vec::with_capacity(1024));
```
vext01, line 66: Is it trivial to prove that two borrows cannot occur? If so, you might be able to be faster without runtime borrow checks? Maybe I'm jumping the gun a bit...

internal_ws/yktrace/src/swt.rs
```diff
@@ -167,35 +129,10 @@ __yk_swt_rec_loc_active:
 /// This is outlined to make the common case of tracing being disabled faster.
 #[no_mangle]
 unsafe extern "C" fn __yk_swt_rec_loc_impl(symbol_name: *const i8, bb_idx: u32) {
-    // Check if we need more space and reallocate if necessary.
```
vext01, line 170: Really great to see this go.

internal_ws/yktrace/src/swt.rs
bjorn3: It shouldn't be possible for two borrows to occur. I could use `UnsafeCell` if you prefer.

vext01:
Supposing that one day we can trace the std library (and Vec::push) for example. What prevents infinite recursion into the location recorder?

bjorn3:
`internal_ws` is not traced and should always keep using a non-traced libstd IMO. If it starts using a traced libstd, that will be disastrous for performance when using the software tracer.

vext01:
OK, I buy that :)

internal_ws/yktrace/src/swt.rs
vext01: Since this is quite performance critical, I think we could allow it, but I'd like to check with @ltratt.

Laurie: For context, this vector is the buffer used to collect a software trace. There will be one push per recorded block. How do you feel about `UnsafeCell`?

ltratt: Is the `RefCell` here just to make things mutable within a thread local?

bjorn3: Yes

ltratt: OK. I agree that the frequency with which we call this might make `RefCell` a noticeable overhead. My suggestion would be that we create a new type that wraps the `Vec` and does the `UnsafeCell` trickery itself, rather than having lots of `unsafe` stuff sprinkled around everywhere.

bjorn3: Done in a76df1ee80d27bbb405905b357df4283dacc3b50.

vext01:
I think this is ready to merge?

bjorn3:
Yes

PR #246, opened 2021-02-17
Address some more comments from #226.

ptersilie:
Keeping those PRs small for now.

### DIFF
--- a/internal_ws/ykcompile/src/arch/x86_64/mod.rs
+++ b/internal_ws/ykcompile/src/arch/x86_64/mod.rs
@@ -264,8 +264,9 @@ pub extern "sysv64" fn bh_new_vec() -> *mut Vec<FrameInfo> {
 }
 
 /// Pushes a new `FrameInfo` instance onto the vector behind the pointer `vptr`. The `FrameInfo`
-/// instance is creates from a pointer to a symbol name and its length, a basic block index and a
-/// pointer to some allocated memory.
+/// instance is created from a pointer to a symbol name and its length, a basic block index and a
+/// pointer to some allocated memory. Note that this function converts the raw pointer `vptr` into
+/// an `&mut` reference.
 pub extern "sysv64" fn bh_push_vec(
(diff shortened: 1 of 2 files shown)


vext01:
LGTM

ptersilie:
Should we wait for @ltratt to approve this first, as these are his comments?

internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -293,9 +293,8 @@ pub extern "sysv64" fn bh_push_vec(
         unsafe { std::str::from_utf8(std::slice::from_raw_parts(sym_ptr, sym_len)).unwrap() };
     let body = SIR.body(fname).unwrap();
     let fi = FrameInfo { body, bbidx, mem };
-    let mut v = unsafe { Box::from_raw(vptr) };
+    let v = unsafe { &mut *vptr };
```
ltratt, line 296: Because we convert the pointer into an `&mut`, I do think that we should document (somehow...) that the `mut *` pointer will be converted into an `&mut`, so that someone who's au fait with stacked borrows can realise the implication of calling this function.

ptersilie: Something like, note this converts the raw pointer into a mutable reference, but there are no other references for this, so we good on stacked borrows. ;-)

ltratt: I think the doc string should lay out the spec only. So "Note that this function converts the raw pointer into an `&mut` reference." is sufficient. It's up to the caller to satisfy that spec re: stacked borrows.

ptersilie: Sorry, I got side-tracked by something else, which I initially wanted to add to this PR, but have decided its better on its own. Try 1b23592.

ltratt:
Please squash.

ptersilie:
Squashed!

Issue #244, opened 2021-02-17
Better #[do_not_trace] handling


bjorn3:
It shouldn't require the first block and return terminator of `#[do_not_trace]` function to be traced. This is important for software tracing. In case of hardware tracing it happens by accident as disabling hardware tracing is relatively costly, but in case of software tracing it did be nice to omit calls into the software tracer for `#[do_not_trace]` functions.

ltratt:
Is this a correctness or a performance issue? I *think* it's just the latter?

bjorn3:
There may be cases where it is really necessary to avoid the software tracing calls, but otherwise it is mostly a performance issue.

ltratt:
Do we have a concrete example of such cases?

bjorn3:
The only example I can think of is that inside interrupt handlers TLS accesses are UB. The software tracing callback uses TLS to determine if tracing is enabled. There may be more examples, but this is the only I can think of right now.

ltratt:
> The only example I can think of is that inside interrupt handlers TLS accesses are UB

Right, I agree that we can't do software tracing inside interrupts / signals. I don't think we would do so, however, because the signal handler would point to the non-software-tracing version and thus would execute normal code. Warning: I might be wrong.

bjorn3:
> I don't think we would do so, however, because the signal handler would point to the non-software-tracing version and thus would execute normal code. Warning: I might be wrong.

Every function is either traced or not. There is only ever one version being created. Without `#[do_not_trace]` all function in a crate are software traced when enabling software tracing support.

I just came up with another case where not tracing is important. When traced code calls a foreign function like from a C library and then this untraced code calls into a traced function, this part of the trace has to be trimmed.

* interp_step@main_exe
    * function_with_callback@some_c_lib
        * callback@main_exe

The TIR trace here would need to only contain the statements of `interp_step@main_exe` and a call to `function_with_callback@some_c_lib`. It must not contain `callback@main_exe`.

vext01:
> When traced code calls a foreign function like from a C library and then this untraced code calls into a traced function.

And this also can occur in just Rust code too, where a `#[do_not_trace]` function then calls a traced function.

ltratt:
Offline @vext01 has gently reminded me that my dreams do not always match reality :) In the long term, we want to have two versions of the interpreter in the binary:

* A fully optimised version in which a program starts execution.
* An unoptimised (or, perhaps, semi-optimised) version which we use for tracing.

[...]

PR #247, opened 2021-02-17
Fix recursion in `do_not_trace` functions.

ptersilie:
Until now there was a bug which prevented us from properly converting
traces to TIR that contained recursive functions that are annotated with
`do_not_trace`. In order to detect when we are done ignoring the content
of such a function, we waited until we observed its return terminator.
However, due to recursion, the return terminator we see might not belong
to the outer-most function, thus aborting the ignore path too early.
This commit fixes this, by counting how often we enter the recursive
function, and only stop ignoring things when we see the outer-most
return.

### DIFF
--- a/internal_ws/yktrace/src/tir.rs
+++ b/internal_ws/yktrace/src/tir.rs
@@ -36,7 +36,9 @@ impl<'a, 'm> TirTrace<'a, 'm> {
         let mut itr = trace.iter().peekable();
         let mut rnm = VarRenamer::new();
         // Symbol name of the function currently being ignored during tracing.
-        let mut ignore: Option<String> = None;
+        let mut dnt_func: Option<String> = None;
+        // Number matching calls and returns of the current `do_not_trace` function.
+        let mut dnt_count: usize = 0;
         // Maps symbol names to their virtual addresses.
         let mut addr_map: HashMap<String, u64> = HashMap::new();
 
@@ -60,7 +62,7 @@ impl<'a, 'm> TirTrace<'a, 'm> {
             };
 
             // Ignore yktrace::trace_debug.
-            // We don't use the 'ignore' machinery below, as that would require the TraceDebugCall
+            // We don't use the 'do_not_trace' machinery below, as that would require the TraceDebugCall
             // terminator to contain the symbol name, which would be wasteful.
             if body.flags.contains(BodyFlags::TRACE_DEBUG) {
                 continue;
@@ -81,14 +83,31 @@ impl<'a, 'm> TirTrace<'a, 'm> {
                 addr_map.insert(loc.symbol_name.to_string(), addr.unwrap());
             }
 
-            // If a function was annotated with `do_not_trace`, skip all instructions within it as
-            // well. FIXME: recursion.
-            if let Some(sym) = &ignore {
-                if sym == &loc.symbol_name
-                    && body.blocks[user_bb_idx_usize].term == Terminator::Return
-                {
-                    ignore = None;
-                }
+            // If a function was annotated with `do_not_trace`, we need to skip all instructions
+            // within it. However, since the tracer will have inlined function calls, the only way
+            // to detect that the `do_not_trace` function has finished, is by observing its return
+            // terminator. Since the `do_not_trace` function can be recursive we count how many
+            // times it is called and returned from and only stop ignoring things until we return
+            // from the outer-most `do_not_trace` function.
+            if let Some(sym) = &dnt_func {
+                match &body.blocks[user_bb_idx_usize].term {
+                    Terminator::Return => {
+                        if sym == loc.symbol_name {
+                            dnt_count -= 1;
+                            if dnt_count == 0 {
+                                dnt_func = None;
+                            }
+                        }
+                    }
+                    Terminator::Call { operand, .. } => {
+                        if let Some(csym) = operand.symbol() {
+                            if csym == sym {
+                                dnt_count += 1;
+                            }
+                        }
+                    }
+                    _ => {}
+                };
                 continue;
             }
 
@@ -247,7 +266,8 @@ impl<'a, 'm> TirTrace<'a, 'm> {
                             // If the function has been annotated with do_not_trace, turn it into a
                             // call.
                             if callbody.flags.contains(BodyFlags::DO_NOT_TRACE) {
-                                ignore = Some(callee_sym.to_string());
+                                dnt_func = Some(callee_sym.to_string());
+                                dnt_count = 1;
                                 term_stmts.push(Statement::Call(op.clone(), newargs, Some(ret_val)))
                             } else {
                                 entered_call = true;
--- a/tests/src/ykcompile.rs
+++ b/tests/src/ykcompile.rs
@@ -1286,3 +1286,71 @@ fn nested_do_not_trace() {
     assert!(unsafe { ct.execute(&mut args).is_null() });
     assert_eq!(args.0, 1);
 }
+
+#[test]
+fn recursive_do_not_trace() {
+    #[do_not_trace]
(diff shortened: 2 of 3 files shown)


ltratt:
Would an alternative (and perhaps even generic?) approach to this be to note the current stack address, and droy out of `do_not_trace` when we go above that address?

ptersilie:
Could you elaborate that thought a bit? I'm not sure where I would get the stack address from.

vext01:
There was a discussion on MM earlier today which is related to this: it's hard for software tracing to see the return terminator of a untraced function. We could do with a better way of implementing #[do_not_trace].

@bjorn3 I wonder if Laurie's suggestion is a good one?

The only spanner in the works is tail recursion optimisation, where the same frame is re-used.

vext01:
> Could you elaborate that thought a bit?

He means, you know when your `do_not_trace` function has returned, when the stack pointer arrives back at its original value.

ptersilie:
What stack pointer? We don't have a stack pointer in SIR, do we? What am I missing?

ltratt:
This is a dynamic check, so you can check, and compare, the address of the call stack.

ptersilie:
> There was a discussion on MM earlier today which is related to this: it's hard for software tracing to see the return terminator of a untraced function. We could do with a better way of implementing #[do_not_trace].

Yeah, I've already linked to this PR in that discussion.

vext01:
Lukas has a point. SIR currently doesn't contain information about the value of the stack pointer during trace collection. I dunno if it could be added or not...

ltratt:
Ah, I might have misunderstood what's going on. Is it the case that the code that's doing the check has got a trace (from hardware or software) and is trying to work out which bits to compile and which bits to leave as an external call (i.e. `do_not_trace`)?

ptersilie:
Exactly! Another problem is, that we don't trace system calls, so we see `Call`s to them, but never see their `Return`, so we can't even just count `Call`s/`Return`s, and stop when we reach 0 again.

vext01:
That's right.

At the time the TIR compiler is running, you only have an vector of basic block locations and nothing about the environmental conditions (concrete register and memory values) under which the trace was collected.

internal_ws/yktrace/src/tir.rs
```diff
@@ -37,6 +37,7 @@ impl<'a, 'm> TirTrace<'a, 'm> {
         let mut rnm = VarRenamer::new();
         // Symbol name of the function currently being ignored during tracing.
         let mut ignore: Option<String> = None;
+        let mut ignore_count: usize = 0;
```
ltratt, line 40: Are there any `ignore` things that aren't `do_not_trace`?

ptersilie: At the moment, no.

ltratt:
@ptersilie @vext01 Right, got it! So please ignore my original suggestion: I had completely misunderstood what part of the code we were in!

internal_ws/yktrace/src/tir.rs
bjorn3: Any function directly or indirectly called from a `#[do_not_trace]` function must also not be traced.

vext01:
I don't think Laurie's idea is dead in the water just yet.

The software trace recorder can easily extract the stack pointer value and include it in the SIR if we wanted.

I'm not certain the same is true of PT though. I'd have to investigate. Should I?

tests/src/ykcompile.rs
```diff
@@ -1286,3 +1286,35 @@ fn nested_do_not_trace() {
...
+#[test]
+fn recursive_do_not_trace() {
+    #[do_not_trace]
+    fn rec(i: u8) -> u8 {
+        let mut j = i;
+        if i < 10 {
+            j = rec(i + 1);
+        }
+        j
+    }
```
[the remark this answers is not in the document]
ptersilie: Got call, will add that.

internal_ws/yktrace/src/tir.rs
ltratt: Then I'd suggest this is the `do_not_trace_count` or similar (personally, I'd probably call it `dnt_trace` and add `# do_not_trace count` or similar).

ltratt:
> I'm not certain the same is true of PT though. I'd have to investigate. Should I?

I don't think this would be useful: it would slow the act of tracing down. The count check being done in this PR should be super fast, and it's done in parallel, so this is the right place to do it.

The only annoying thing is that if we call some large external function that's `do_not_trace`, we end up recording a huge PT trace that we then throw away. In the long run, we might like to temporarily turn off PT for such long cases, but that's not something we necessarily need to do now (I suspect that for short functions, the overhead of turning PT on/off would outweigh any advantages).

vext01:
Agreed, but we still need to find a solution to @bjorn3's software tracing problem #244.

@bjorn3 maybe we just have to live with the fact that we will have to trace the first and return blocks of untraced functions...

tests/src/ykcompile.rs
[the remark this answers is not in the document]
ptersilie: Nice, this immediately uncovered a bug. Fixed in 997cf5d.

internal_ws/yktrace/src/tir.rs
ptersilie: Try dfc8326.

ptersilie:
> `dnt_trace` -- isn't this saying `do_not_trace_trace`? I think just `dnt_` is fine?

Ah, I thought `dnt` stood for `dont`. :smile: Will fix.

> Also, do we need to keep a (clone!) of the function name? I would have thought that the metadata bool "this function is do not trace" is enough.

Generally, this would work. Unfortunately, we currently annotate entire libraries with `do_not_trace`, e.g. `core`. However, some of these functions (e.g. syscalls) aren't traced by the hardware tracer, which means we'll observe a call, but never see its return, so `dnt_count` never becomes 0 again. I'm not sure if there is a way to know which core library functions will be hardware traced, but I can see us changing the way we deal with libraries in the future, which hopefully allows us to "optimise" the `do_not_trace` detection code.

ptersilie:
Think that addresses all comments?

internal_ws/yktrace/src/tir.rs
```diff
@@ -89,20 +89,20 @@ impl<'a, 'm> TirTrace<'a, 'm> {
...
                     Terminator::Return => {
                         if sym == loc.symbol_name {
-                        dnt_trace_count -= 1;
-                        if dnt_trace_count == 0 {
-                            dnt_trace_func = None;
+                        dnt_count -= 1;
+                        if dnt_count == 0 {
+                            dnt_func = None;
                         }
                         }
```
ltratt, line 100: These two same-indent `}` suggest `cargo fmt` is needed?

ltratt:
> Unfortunately, we currently annotate entire libraries with do_not_trace, e.g. core. However, some of these functions (e.g. syscalls) aren't traced by the hardware tracer, which means we'll observe a call, but never see its return, so dnt_count never becomes 0 again. I'm not sure if there is a way to know which core library functions will be hardware traced, but I can see us changing the way we deal with libraries in the future, which hopefully allows us to "optimise" the do_not_trace detection code.

OK, can we raise an issue so that this isn't forgotten. `String` can end up being a noticeable overhead in parts of code like this, and if it's an easy win to get rid of it, we might as well do so soon-ish.

internal_ws/yktrace/src/tir.rs
ptersilie: Fixed in 914b961.

ltratt:
Please squash (and rebase).

ptersilie:
> OK, can we raise an issue so that this isn't forgotten.

New issue: #249

ptersilie:
Squashed!

Issue #243, opened 2021-02-17
Trace verifier


bjorn3:
This should check that a trace doesn't include any block jumps that are impossible. For example returning from a function without `return` terminator, entering a function without `call` terminator or jumping to a block without it being a successor of the current block.

vext01:
Closing, as this is for the old yk design.

PR #253, opened 2021-02-19
Address more comments from #226.

ptersilie:
A quick PR addressing comments from the audit (#226), so I don't loose track of what I've fixed and what not.

### DIFF
--- a/internal_ws/ykcompile/src/arch/x86_64/mod.rs
+++ b/internal_ws/ykcompile/src/arch/x86_64/mod.rs
@@ -282,7 +282,7 @@ pub extern "sysv64" fn bh_push_vec(
     v.push(fi);
 }
 
-/// Compile a TIR trace, returning executable code.
+/// Compile a TIR trace.
 pub fn compile_trace(tt: TirTrace) -> CompiledTrace {
     CompiledTrace {
         mc: TraceCompiler::compile(tt, false),
@@ -297,7 +297,7 @@ pub struct TraceCompiler {
     register_content_map: HashMap<u8, RegAlloc>,
     /// Maps trace locals to their location (register, stack).
     variable_location_map: HashMap<Local, Location>,
-    /// Local decls of the tir trace.
+    /// Local decls of the TIR trace.
     pub local_decls: HashMap<Local, LocalDecl>,
     /// Stack builder for allocating objects on the stack.
     stack_builder: StackBuilder,
@@ -409,7 +409,10 @@ impl TraceCompiler {
         self.stack_builder.alloc(ty.size(), ty.align())
     }
 
+    /// Assign a `Location` to a `Local` turning live. If possible, find it a register for it to
+    /// live in, or failing that, allocate space on the stack.
     fn local_live(&mut self, local: &Local) {
+        debug_assert!(self.variable_location_map.get(local).is_none());
         // Assign a Location to this Local.
         self.local_to_location(*local);
     }
@@ -440,7 +443,7 @@ impl TraceCompiler {
     /// Copy bytes from one memory location to another.
     fn copy_memory(&mut self, dest: &RegAndOffset, src: &RegAndOffset, size: u64) {
         // We use memmove(3), as it's not clear if MIR (and therefore SIR) could cause copies
-        // involving overlapping buffers.
+        // involving overlapping buffers. See https://github.com/rust-lang/rust/issues/68364.
         let sym = find_symbol("memmove").unwrap();
         self.save_regs(&*CALLER_SAVED_REGS);
         dynasm!(self.asm
@@ -478,7 +481,7 @@ impl TraceCompiler {
         self.restore_regs(&*CALLER_SAVED_REGS);
     }
 
-    /// Push the specified registers to the stack in order.
+    /// Push the specified registers to the stack in the order they appear in the array.
     fn save_regs(&mut self, regs: &[u8]) {
         for reg in regs.iter() {
             dynasm!(self.asm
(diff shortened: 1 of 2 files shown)


internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -409,6 +409,7 @@ impl TraceCompiler {
         self.stack_builder.alloc(ty.size(), ty.align())
     }
 
+    /// Implements the StorageLive statement, assigning a Location the local.
```
ltratt, line 412: I'm not sure how to parse this sentence.

vext01: How about:
> Called when a local variable becomes live. The register allocator is invoked to make storage space for the local.

ptersilie: Which part is the problem? This basically tells the compiler a variable is live and assigns it a register or location on the stack. How about: `Implements a local going live, assigning it to a register or, if no registers are free, to a location on the stack.`

ltratt: @vext01 I think, in general, more active descriptions are clearer.

@lukasd Words like "implements" feel a bit like padding. I'm also not sure what "a local going live" exactly means.

Taking those thoughts together, would something like "Assign a Location to a StorageLive (i.e. finding it a register or space on the stack)." be accurate? Maybe too terse?

ptersilie: `StorageLive` is a MIR statement, so saying we assign a location to it, doesn't sound right. We are assigning a location to the local that MIR has decided is live from that point onwards until we see a `StorageDead` for it.

vext01: Tweaking Laurie's prose a bit:
> Assign a `Location` to a `Local` just turning live (i.e. if possible, find it a register to live in, or failing that, space on the stack).

bjorn3: Somewhat related, but https://github.com/rust-lang/rust/pull/78360 if merged will remove `Storage*` after borrowck when not optimizing.

vext01: @ptersilie does commenting out the call to `local_to_location` cause any test to fail?

If my mental model is correct, any later bytecode mentioning a local will already call `local_to_location` and thus lazily allocate space for it as-needed?

vext01: Ah, it might bust stopgapping...

ptersilie: Yeah, I was thinking the same, but seem to remember there was a slight difference, but will have to double check.

ptersilie: Nope, looks like all tests pass, so this is just here to have the counter to `storagedead`. Though it may mean that a local can snatch a register if the live happens before data stored in it.

ptersilie: > Somewhat related, but [rust-lang/rust#78360](https://github.com/rust-lang/rust/pull/78360) if merged will remove `Storage*` after borrowck when not optimizing.

This would be annoying for us (at least until we have our own liveness analysis), so I hope it can be re-enabled in ykrustc.

ltratt: > Nope, looks like all tests pass, so this is just here to have the counter to storagedead.

Does this mean we can delete / simplify code or ... ?

ptersilie: I think we can remove it, though it is a bit odd that `StorageLive` then has no real purpose. Maybe we can use this function as a check that we don't `StorageLive` the same variable twice, without `StorageDead`ing it first. I could imagine a future bug in MIR/SIR/TIR missing a `StorageDead`, or `VarRenamer` not renaming inlined functions properly, which we otherwise wouldn't notice unless it causes programs to behave weirdly.

vext01: It means that we don't strictly have to allocate a local until its first use, and could in fact remove the function in question altogether.

It might result in better allocation results overall. Consider the following:

```
live($1)
if (...) {
   $1 = call foo(...)
   call blah($1)
} else {
  // don't use $1
}
```

If we trace the else branch, then right now, then the trace would allocate space for, but never use $1. In other words, it could potentially block another variable from using a register.

If we allocate on first use, we wouldn't allocate $1 at all (in this trace), leaving it free for other locals.

But, I don't know if MIR would generate code like this...

Another question I then have is, if we don't allocate eagerly, what is the purpose of `StorageLive` in our bytecode?

vext01: Lukas and I raced, but agreed.

So perhaps `StorageDead` is useful for #243. But then again, we could just consider the first use of a variable as its birth?

ltratt: I have no idea so I'm relying on you two to work this out :)

ptersilie: > ```
> live($1)
> if (...) {
>    $1 = call foo(...)
>    call blah($1)
> } else {
>   // don't use $1
> }
> ```

I don't know why MIR wouldn't put the `live($1)` in the same block where it's assigned and used, but you never know. Although a quick test on Rust Playground seems to confirm this suspicion (which again doesn't mean MIR always does this).

ptersilie: Fixed in 88f0a69 and a7b7aee.

ptersilie:
Ready for re-review.

ltratt:
If @vext01 is OK with https://github.com/softdevteam/yk/pull/253/commits/a7b7aeedcf410e97096a1b7d7a007242d7a6067b then please squash.

vext01:
Go ahead.

ptersilie:
Squashed.

PR #257, opened 2021-02-22
Make clippy work.

vext01:
Again, this uses a nightly toolchain version, due to the same rustup bug  that affects `cargo fmt`.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -26,8 +26,9 @@ fn run_action(workspace: Workspace, target: &str, extra_args: &[String]) {
         run_action(Workspace::Internal, target, &[]);
     }
 
-    let mut cmd = if target == "fmt" {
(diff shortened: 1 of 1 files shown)


xtask/src/main.rs
```diff
@@ -117,7 +121,9 @@ fn run_action(workspace: Workspace, target: &str, extra_args: &[String]) {
         .wait()
         .unwrap();
 
-    if !status.success() {
+    // The clippy exception ensures that both workspaces are linted if one workspace fails. The
+    // exit status may be inaccurate, but we can live with this.
+    if !status.success() && (target != "clippy") {
```
bjorn3, line 126: This will cause `cargo xtask clippy` on CI to never fail.

vext01: We don't run clippy in CI, as we don't agree with all of the lints it suggests. Even if we were to build a clippy configuration, we'd probably find many of the lints subjective.

PR #260, opened 2021-02-22
Roll prod.rs into lib.rs in ykshim and ykshim_client.

vext01:
No functional change.

Addresses [this](https://github.com/softdevteam/yk/pull/226/files#r572068599)


internal_ws/ykshim/src/lib.rs
```diff
@@ -3,9 +3,124 @@
...
+    ThreadTracer,
+};
+
+/// The different ways by which we can collect a trace.
+#[derive(Clone, Copy)]
+#[repr(u8)]
+#[allow(dead_code)]
+enum TracingKind {
+    /// Software tracing via ykrustc.
+    SoftwareTracing = 0,
```
ltratt, line 23: Do we need to hardcode these numbers? [I appreciate this is just moved code, but I might as well ask while I'm looking at it!]

internal_ws/ykshim/src/lib.rs
```diff
@@ -3,9 +3,124 @@
...
+    tracer: *mut ThreadTracer,
+    error_msg: *mut *mut c_char,
+) -> *mut SirTrace {
+    let tracer = Box::from_raw(tracer);
+    let sir_trace = tracer.stop_tracing();
+    match sir_trace {
+        Ok(sir_trace) => Box::into_raw(Box::new(sir_trace)),
+        Err(err) => {
+            *error_msg = CString::new(err.to_string())
+                .unwrap_or_else(|err| {
```
ltratt, line 55: I'm not sure if all this code does anything meaningfully useful over just `unwrap`.

internal_ws/ykshim/src/lib.rs
```diff
@@ -3,9 +3,124 @@
...
+unsafe extern "C" fn __ykshim_compile_trace(
+    sir_trace: *mut SirTrace,
+    error_msg: *mut *mut c_char,
+) -> *mut CompiledTrace {
+    let sir_trace = Box::from_raw(sir_trace);
+    let tt = match TirTrace::new(&*SIR, &*sir_trace) {
+        Ok(tt) => tt,
+        Err(err) => {
+            *error_msg = CString::new(err.to_string())
+                .unwrap_or_else(|err| {
```
ltratt, line 77: Ditto.

ykshim_client/src/lib.rs
```diff
@@ -3,17 +3,159 @@
...
-pub use prod_api::*;
+use std::ffi::{c_void, CString};
+use std::marker::PhantomData;
+use std::os::raw::c_char;
+use std::{mem, ptr};
 
 mod test_api;
 pub use test_api::*;
+
+// Opaque pointers.
```
ltratt, line 22: Not sure this doc comment is very helpful.

ykshim_client/src/lib.rs
```diff
@@ -3,17 +3,159 @@
...
+pub(crate) type RawCompiledTrace = c_void;
+pub(crate) type RawSirTrace = c_void;
+type RawThreadTracer = c_void;
+pub(crate) type RawTirTrace = c_void;
+pub type RawStopgapInterpreter = c_void;
+
+// These types and constants must be kept in sync with types of the same name in the internal
```
ltratt, line 29: s/types/the types/ ?

ykshim_client/src/lib.rs
```diff
@@ -3,17 +3,159 @@
...
+    /// Hardware tracing via ykrustc + hwtracer.
+    HardwareTracing = 1,
+}
+
+pub struct ThreadTracer(*mut RawThreadTracer);
+
+/// Start tracing using the specified kind of tracing.
+pub fn start_tracing(tracing_kind: TracingKind) -> ThreadTracer {
+    let tracer = unsafe { __ykshim_start_tracing(tracing_kind as u8) };
+    assert!(!tracer.is_null());
```
ltratt, line 71: `debug_assert` ?

ykshim_client/src/lib.rs
```diff
@@ -3,17 +3,159 @@
...
+    ThreadTracer(tracer)
+}
+
+impl ThreadTracer {
+    pub fn stop_tracing(mut self) -> Result<SirTrace, CString> {
+        let mut err_msg = std::ptr::null_mut();
+        let sir_trace = unsafe { __ykshim_stop_tracing(self.0, &mut err_msg) };
+        self.0 = ptr::null_mut(); // consumed.
```
ltratt, line 79: "consumed" ?

ltratt (commented):
The code moving looks good, but I've done an accidental audit! Whether we address those comments in this PR or a separate one is up to you.

internal_ws/ykshim/src/lib.rs
vext01: It's due to the workspace model. The numbers have to be duplicated in both workspaces whilst also being stable across the C ABI. I don't see any value in adding constants here, as the enum already serves to define constants.

internal_ws/ykshim/src/lib.rs
vext01: The code attempts to give a meaningful error message as to what's gone wrong in the (opaque from Rust's POV) C code. But I think the `unwrap_or_else` might be wrong. Wouldn't it need to be `and_then()`?

If we just let this panic, then we don't make use of the error message the C code has constructed.

vext01: (and we can't use `panic!` on the message since it's not constant)

ykshim_client/src/lib.rs
vext01: `__ykshim_stop_tracing` consumed the pointer, so we set it to null to avoid a double drop.

This is an idiom used throughout.

internal_ws/ykshim/src/lib.rs
vext01: sorry, i misread my own code. Yes, we can panic here.

vext01:
Does c9bd3ce address your comments OK?

ykshim_client/src/lib.rs
```diff
@@ -68,15 +67,15 @@ pub struct ThreadTracer(*mut RawThreadTracer);
...
     let tracer = unsafe { __ykshim_start_tracing(tracing_kind as u8) };
-    assert!(!tracer.is_null());
+    debug_assert!(!tracer.is_null());
     ThreadTracer(tracer)
 }
 
 impl ThreadTracer {
     pub fn stop_tracing(mut self) -> Result<SirTrace, CString> {
         let mut err_msg = std::ptr::null_mut();
         let sir_trace = unsafe { __ykshim_stop_tracing(self.0, &mut err_msg) };
```
ltratt, line 77: Ah! I think what you really want here is to use `mem::replace` which prevents the potential race that you're currently using two lines to avoid.

internal_ws/ykshim/src/lib.rs
ltratt: OK.

ykshim_client/src/lib.rs
vext01: I'm not sure what you mean. There's no race that I'm aware of.

This dance is because the use of raw pointers across the C ABI kills ownership handling and we have to track it ourself:

 - If `self` is dropped without a shim function consuming it, then the client side must drop it to avoid a leak.
 - If `self` is consumed by a shim, then the other side is responsible for dropping it. If the client side also drops it we have a double free.

The null pointer is used to mark `self` consumed, and the check is [here](https://github.com/softdevteam/yk/pull/260/files/8a2c3f62e767a1020d2a10edd2ceaa2559111acc..c9bd3ce8a103045a3aa03882887143d1a5267f3e#diff-01f309b981ab8e667ec6d6662d6d1c8cb17071b932bdb1748dfd9cf869236fdeR88).

(We only use this idiom for types where there exists a shim which consumes)

I'm unsure how `mem::replace` helps, but I may have missed something?

ltratt: There isn't a race here, because you carefully:

1. read the value (line 77).
2. now that it's stale, zero it (line 78).

If you didn't do that second step, you'd have a race.

What I'm suggesting is that I *think* you're doing an explicit version of `mem::replace` and it might be worth using it instead of the two steps above? Something like:

```rust
let v = mem::replace(self.0, ptr::null_mut());
let sir_trace = unsafe { __ykshim_stop_tracing(v, &mut err_msg) };
```

vext01: Is there any worth in moving a raw pointer though? It's already `Copy`.

I think I'm missing something...

ltratt: I think it's worth considering `replace` here because it makes clear that you can't accidentally have a stale value lurking around. Up to you.

vext01: Resolved offline.

vext01:
89f6002 uses `mem::replace`. OK to squash?

ltratt:
Please squash.

vext01:
Splat. I've kept your changes separate so that they don't get swamped by the large movement of code.

PR #266, opened 2021-02-24
Remove sg_interpret.

ptersilie:
This function was needed to properly initialise the stopgap interpreter
when coming from a guard failure. By adding the local holding the
interpreter context to the live locals, the function can be moved into
`from_frames`, which means `sg_interpret` is no longer needed.

There is one thing that makes me a nervous about this, which is why I'm leaving this as a draft for now. Since we store a reference to the interpreter context inside the stopgap interpreter before calling `interpret`, there is a way to get UB here:
```rust
let mut args = InterpCtx(0);
let ptr = ct.execute(&mut args);
let mut si = StopgapInterpreter(ptr);
args = InterpCtx(1); // if we overwrite args here, then si's frame still points to the previous InterpCtx.
si.interpret();
assert_eq!(args.0, 9);
```
At the very least we need to document this, but I'm hoping we can give a stronger guarantee (without having to pass args into `interpret` like before).

### DIFF
--- a/internal_ws/ykcompile/src/lib.rs
+++ b/internal_ws/ykcompile/src/lib.rs
@@ -139,7 +139,10 @@ pub struct CompiledTrace {
 }
 
 impl CompiledTrace {
-    /// Execute the trace by calling (not jumping to) the first instruction's address.
+    /// Execute the trace by calling (not jumping to) the first instruction's address. Returns a
+    /// pointer to an initialised `StopgapInterpreter` if there was a guard failure, or a null
+    /// pointer otherwise. Note that the interpreter holds a `*mut` pointer to `args`, so we need
+    /// to make sure that `args` is still alive when we call the interpreters `interpret` function.
     pub unsafe fn execute<TT>(&self, args: &mut TT) -> *mut StopgapInterpreter {
         let func: extern "sysv64" fn(&mut TT) -> *mut StopgapInterpreter =
             mem::transmute(self.mc.ptr(dynasmrt::AssemblyOffset(0)));
--- a/internal_ws/yksg/src/lib.rs
+++ b/internal_ws/yksg/src/lib.rs
@@ -235,21 +235,15 @@ impl StopgapInterpreter {
             };
             frames.push(frame);
         }
-        StopgapInterpreter { frames }
-    }
-
-    /// Run the SIR interpreter after it has been initialised by a guard failure. Since we start in
-    /// the block where the guard failed, we immediately skip to the terminator and interpret it to
-    /// see which block we need to start interpretation in.
-    pub unsafe fn sg_interpret(&mut self, ctx: *mut u8) {
-        self.set_interp_ctx(ctx);
-        // Jump to the correct basic block by interpreting the terminator.
-        let frame = self.frames.last().unwrap();
+        let mut sg = StopgapInterpreter { frames };
+        let frame = sg.frames.last().unwrap();
+        // Since we start in the block where the guard failed, we immediately skip to the
+        // terminator and interpret it to initialise the block where actual interpretation needs to
+        // start.
         let body = frame.body.clone();
         let bbidx = usize::try_from(frame.bbidx).unwrap();
-        self.terminator(&body.blocks[bbidx].term);
-        // Start interpretation.
-        self.interpret();
+        sg.terminator(&body.blocks[bbidx].term);
+        sg
     }
 
     /// Given the symbol name of a function, generate a `StackFrame` which allocates the precise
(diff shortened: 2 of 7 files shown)


ltratt:
I'm not sure I understand:

```rust
let mut args = InterpCtx(0);
let ptr = ct.execute(&mut args);
let mut si = StopgapInterpreter(ptr);
args = InterpCtx(1); // if we overwrite args here, then si's frame still points to the previous InterpCtx.
si.interpret();
```

Why would I do `args = InterpCtx(1)` if that's not used later?

ptersilie:
> Why would I do `args = InterpCtx(1)` if that's not used later?
It's not about why anyone should, but that you could, and then something unexpected happens. I've updated my example above to use args later.

The problem is, that running `interpret` uses `args`, but by looking at this code you can't tell. To give another example:
```
let mut args = InterpCtx(0);
let ptr = ct.execute(&mut args);
let mut si = StopgapInterpreter(ptr);
mem::forget(args);
si.interpret();
```
One could reasonably assume that args is no longer needed, since it obviously isn't used anymore, but in reality `si` still holds a mutable reference to it.

ltratt:
Does `si` hold an `&mut` reference or a `*mut`? The latter is fine, but the former would be UB.

ptersilie:
It's a `*mut`, but even then wouldn't the above still create undefined behaviour? If we forget/free `args`, and then call interpret, which may read/write to args, we don't know what we are reading from/writing to.

ltratt:
>If we forget/free args, and then call interpret, which may read/write to args, we don't know what we are reading from/writing to.

Well, *yes*, but isn't this a classic case of "if it hurts, don't do it"? In other words, the user can't forget/free `args` can they?

ptersilie:
Yeah, or overwrite them, I guess. I think that should be documented though, that `si` has a hidden reference to `args`.

Another thing I wonder about is, what if `args` goes out of scope before we call `interpret` (note this can't happen in `mt.rs` since there `args` is always in scope).

ltratt:
> Another thing I wonder about is, what if args goes out of scope before we call interpret

We must make sure that doesn't happen.

ptersilie:
Yes, that's my question. To do so, is it sufficient to add a warning comment (either on `interpret` or `execute` or both?), or do we need to make sure that the user can't do this in the first place?

ptersilie:
I've added a doc comment to clarify that the interpreter context needs to stay alive until we call the stopgap interpreters `interpret` function.

ltratt:
I'd like @vext01 to have a look at this too, because he's more familiar with this part of the code, but I'm OK with this based on what I understand now.

vext01:
Does it help to downgrade the mutability of `args` after calling `execute` to ensure the problematic code won't work?

```
let args = args;
```

My gut says the compiler won't allow it, but worth a shot?

vext01:
The other thing I wonder, is whether the stop gap interpreter could be parametrised by the lifetime of the args?

internal_ws/ykcompile/src/lib.rs
```diff
@@ -139,7 +139,10 @@ pub struct CompiledTrace {
 }
 
 impl CompiledTrace {
-    /// Execute the trace by calling (not jumping to) the first instruction's address.
+    /// Execute the trace by calling (not jumping to) the first instruction's address. Returns a
```
vext01, line 142: Can this return `null` if there is no guard failure?

In the future the only way out of the trace is via a guard failure, but I don't think that's the case now?

ptersilie:
> Does it help to downgrade the mutability of `args` after calling `execute` to ensure the problematic code won't work?

Where? Inside the execute function? Not sure how that would help.

ptersilie:
> The other thing I wonder, is whether the stop gap interpreter could be parametrised by the lifetime of the args?

I don't think that would help against code like this:
```rust
let mut si = StopgapInterpreter(ptr);
args = InterpCtx(1); // if we overwrite args here, then si's frame still points to the previous InterpCtx.
si.interpret();
```

internal_ws/ykcompile/src/lib.rs
ptersilie: You're right. Fixed in 2173f3a.

vext01:
> I don't think that would help against code like this

It wouldn't help correctness, but it would prevent UB.

TBH, I think we can move on without my suggestions, perhaps with a comment. This is just an internal property that we as developers have to ensure. It's not user facing as far as I can see.

ptersilie:
Agreed. Is the comment we got now sufficient? https://github.com/softdevteam/yk/pull/266/files#diff-a0588be717e433bc24ac399f2b57b188c7b13b56c5a512d4c1404d9fac87bf04R144

vext01:
Works for me.

ptersilie:
Can I squash?

ltratt:
Please squash.

ptersilie:
Squashed.

PR #267, opened 2021-02-24
Rethink the Location state machine.

ltratt:
Previously we tried to do everything locklessly. This is not only very complicated, but it's very difficult to be sure  that it's correct. There is at least one potential flaw in the existing code, and maybe more that I haven't realised yet.

This commit moves us to a very different scheme, which is probably best described by copying and pasting this comment from the code:

```
A Location is a state machine which operates as follows (where Counting is the
start state):

             ┌──────────────┐
             │              │─────────────┐
  reprofile  │   Counting   │             │
 ┌──────────▶│              │◀────────────┘
 │           └──────────────┘    increment
 │             │                 count
 │             │ start tracing
 │             ▼
 │           ┌──────────────┐
 |           |              | incomplete  ┌─────────────┐
 │           │   Tracing    │────────────▶|  DontTrace  |
 |           |              |             └─────────────┘
 │           └──────────────┘
 │             │ start compiling trace
 │             │ in thread
 │             ▼
 │           ┌──────────────┐
 |           |  Compiling   |
 │           └──────────────┘
 │             │
 │             │ trace compiled
 │             ▼
 │           ┌──────────────┐
 └───────────│   Compiled   |
             └──────────────┘

We hope that a Location soon reaches the Compiled state (aka "the happy state")
and stays there.

The state machine is encoded in a usize in a not-entirely-simple way, as we
don't want to allocate any memory for Locations that do not become hot. The
layout is as follows (on a 64 bit machine):

  bit(s) | 63..3   | 2           | 1         | 0
         | payload | IS_COUNTING | IS_PARKED | IS_LOCKED

In the Counting state, IS_COUNTING is set to 1, and the parking lot mutex bits
are unused and must remain set at 0. The payload representing the count is
incremented locklessly. All other states have IS_COUNTING set to 0 and the
payload is the address of a boxed HotLocation, access to which is controlled by
the two parking lot mutex bits.

The possible combinations of the counting and mutex bits are thus as follows:

  payload       | IS_COUNTING | IS_PARKED | IS_LOCKED | Notes
  --------------+-----------+---------+---------+-----------------------------------
  <count>       | 1           | 0         | 0         | Start state
                | 1           | 0         | 1         | Illegal state
                | 1           | 1         | 0         | Illegal state
                | 1           | 1         | 1         | Illegal state
  <HotLocation> | 0           | 0         | 0         | Not locked, no-one waiting
  <HotLocation> | 0           | 0         | 1         | Locked, no thread(s) waiting
  <HotLocation> | 0           | 1         | 0         | Not locked, thread(s) waiting
  <HotLocation> | 0           | 1         | 1         | Locked, thread(s) waiting

where `<count>` is an integer and `<HotLocation>` is a boxed `HotLocation` enum.

The precise semantics of locking and, in particular, parking are subtle:
interested readers are directed to
https://github.com/Amanieu/parking_lot/blob/master/src/raw_mutex.rs#L33 for a
more precise definition.
```

The parking lot code looks more complicated than it really is: in essence, the approach outlined in
https://github.com/Amanieu/parking_lot/blob/master/src/raw_mutex.rs for `lock` and `unlock` needs only minor adapting to our situation.

Significantly, using parking lot means that we no longer spin incompetently: threads can actually sleep. Intriguingly, our test suite now takes longer to run (on b13 ~3.2s vs. 2.0s before) *but* it consumes less CPU in doing so (user time 14.6s vs. 15.5s before), so it is in some senses "more efficient". Because we now use a mutex much more often, we are unlikely to get back to the performance we had before, but nor do we really need to: the test suite exercises the worst case of performance especially because we do not currently loop in the compiled machine code. Still, it would be nice to reduce some of the unnecessary double loads that we have here.

I believe/hope this code is correct (or, at least, it's less flawed than what we had before), but it is not as well organised as I would like, nor is it as fast as I would like. However, I think that it wouldn't be sensible for me to polish this endlessly in my private tree: it's better to get this in on the basis that it's a correctness fix, and then we can deal with "niceness" and "speed" in-tree.

### DIFF
--- a/ykrt/src/location.rs
+++ b/ykrt/src/location.rs
@@ -4,85 +4,17 @@ use std::{
     marker::PhantomData,
     sync::{
         atomic::{AtomicUsize, Ordering},
-        Arc, Mutex,
+        Arc,
     },
 };
 
-use crate::mt::ThreadIdInner;
-use ykshim_client::CompiledTrace;
+use parking_lot::Mutex;
+use parking_lot_core::{
+    park, unpark_one, ParkResult, SpinWait, UnparkResult, UnparkToken, DEFAULT_PARK_TOKEN,
+};
+use strum::EnumDiscriminants;
 
-// The current meta-tracing phase of a given location in the end-user's code. Consists of a tag and
-// (optionally) a value. We expect the most commonly encountered tag at run-time is PHASE_COMPILED
-// whose value is a pointer to memory. By also making that tag 0b00, we allow that index to be
-// accessed without any further operations after the initial tag check.
-pub(crate) const PHASE_NUM_BITS: usize = 3;
-pub(crate) const PHASE_TAG: usize = 0b111; // All of the other PHASE_ tags must fit in this.
-pub(crate) const PHASE_COMPILED: usize = 0b000; // Value is a pointer to a chunk of memory containing a
-                                                // CompiledTrace<I>.
-pub(crate) const PHASE_TRACING: usize = 0b001; // Value is a pointer to an Arc<ThreadIdInner> representing a
-                                               // thread ID.
-pub(crate) const PHASE_TRACING_LOCK: usize = 0b010; // No associated value
-pub(crate) const PHASE_COMPILING: usize = 0b011; // Value is a pointer to a `Box<CompilingTrace<I>>`.
-pub(crate) const PHASE_COUNTING: usize = 0b100; // Value specifies how many times we've seen this Location.
-pub(crate) const PHASE_LOCKED: usize = 0b101; // No associated value.
-pub(crate) const PHASE_DONT_TRACE: usize = 0b110; // No associated value.
-
-#[derive(Copy, Clone, PartialEq, Eq, Debug)]
-pub(crate) struct State(usize);
-
-impl State {
-    pub(crate) fn phase(&self) -> usize {
-        self.0 & PHASE_TAG
-    }
-
-    pub(crate) fn phase_compiled<I>(trace: Box<CompiledTrace<I>>) -> Self {
-        let ptr = Box::into_raw(trace);
-        debug_assert_eq!(ptr as usize & PHASE_TAG, 0);
-        State(ptr as usize | PHASE_COMPILED)
-    }
-
-    pub(crate) fn phase_tracing(tid: Arc<ThreadIdInner>) -> Self {
-        let ptr = Arc::into_raw(tid);
-        debug_assert_eq!(ptr as usize & PHASE_TAG, 0);
-        State(ptr as usize | PHASE_TRACING)
-    }
-
-    pub(crate) fn phase_tracing_lock() -> Self {
-        State(PHASE_TRACING_LOCK)
-    }
-
-    pub(crate) fn phase_compiling<I>(trace: Arc<CompilingTrace<I>>) -> Self {
-        let ptr: *const CompilingTrace<I> = Arc::into_raw(trace);
-        debug_assert_eq!(ptr as usize & PHASE_TAG, 0);
-        State(ptr as usize | PHASE_COMPILING)
-    }
-
-    pub(crate) fn phase_counting(count: usize) -> Self {
-        debug_assert_eq!(count << PHASE_NUM_BITS >> PHASE_NUM_BITS, count);
-        State(PHASE_COUNTING | (count << PHASE_NUM_BITS))
-    }
-
-    pub(crate) fn phase_locked() -> Self {
-        State(PHASE_LOCKED)
-    }
-
-    pub(crate) fn phase_dont_trace() -> Self {
-        State(PHASE_DONT_TRACE)
-    }
-
-    pub(crate) fn number_data(&self) -> usize {
-        debug_assert_eq!(self.phase(), PHASE_COUNTING);
-        self.0 >> PHASE_NUM_BITS
-    }
-
-    pub(crate) fn pointer_data<T>(&self) -> *const T {
-        (self.0 & !PHASE_TAG) as *const T
-    }
-
-    pub(crate) unsafe fn ref_data<T>(&self) -> &T {
-        &*self.pointer_data()
-    }
-}
+use ykshim_client::{CompiledTrace, ThreadTracer};
 
 /// A `Location` stores state that the meta-tracer needs to identify hot loops and run associated
 /// machine code.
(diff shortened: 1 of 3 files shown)


ykrt/src/location.rs
```diff
@@ -97,88 +29,408 @@ impl State {
...
+                x,
+                marker: PhantomData,
+            }),
+        }
+    }
+
+    /// Locks this `State` with `Acquire` ordering. It is undefined behaviour to call this function
+    /// if this `State` is in the counting state. If the location moves to the Counting state
+    /// during execution, this will return `Err`.
+    pub(crate) fn lock(&self) -> Result<State<I>, ()> {
```
bjorn3, line 141: If it is UB to call this in some cases, this function should be unsafe.

ykrt/src/location.rs
```diff
@@ -97,88 +29,408 @@ impl State {
...
+        {
+            let ls = self.load(Ordering::Relaxed);
+            debug_assert!(!ls.is_counting());
```
bjorn3, line 144: Compared to the rest of this function. Especially the atomic accesses, I think a single check in release mode too is cheap.

ykrt/src/location.rs
```diff
@@ -97,88 +29,408 @@ impl State {
...
+        debug_assert!(!self.is_locked());
+        self.x >> STATE_NUM_BITS
+    }
+
+    /// If this `State` is not counting, return its `HotLocation`. It is undefined behaviour to
+    /// call this function if this `State` is in the counting phase and/or if this `State` is not
+    /// locked.
+    pub(crate) fn hot_location(&self) -> &mut HotLocation<I> {
+        debug_assert!(!self.is_counting());
+        //debug_assert!(self.is_locked());
```
bjorn3, line 365: This is unsound without these assertions being unconditional, so it need to be unsafe.

ykrt/src/location.rs
ltratt: This is a potentially performance critical part of the code. In the longer run, we'll probably follow parking_lot in splitting this function into an inlinable and non-inlinable component but I can't do that until I have benchmarks, which don't currently run. To be honest, this is more a case of "use `debug_assert` as a form of documentation."

ykrt/src/location.rs
ltratt: Oops, the second sentence is out of date (the third sentence is correct). Fixed in dbc4fc27dd44ae6cafd28d34655d7d3389dd4a85.

ykrt/src/location.rs
ltratt: Agreed. Fixed in e2a418384ea6ccda3d9bc3fe6f38103963205989.

vext01:
> In the Counting state, IS_COUNTING is set to 1, and the parking lot mutex bits
are unused and must remain set at 0.

I think we should prefix this sentence with:

> The bits `IS_PARKED` and `IS_LOCKED` are used for a parking lot mutex.

Without this, "parking lot mutex bits" is undefined. It's not hard to infer, but clarity is best.

vext01:
> The payload representing the count is incremented locklessly. All other states have IS_COUNTING set to 0 and the payload is the address of a boxed HotLocation.

If I've understood, I think this would be clearer as something like:
> When IS_COUNTING is set, the payload is an unsigned integer encoding the hot count of the Location. This count is incremented locklessly via atomic operations. When IS_COUNTING cleared, the payload encodes the address of a HotLocation...

Did you say it's pinned, or at least heap allocated? Should we mention this and why?

vext01:
> Did you say it's pinned, or at least heap allocated?

Ah, it says this later. But do we need to say why?

ykrt/src/location.rs
```diff
@@ -97,88 +29,407 @@ impl State {
...
+pub(crate) const STATE_NUM_BITS: usize = 3;
+
+pub(crate) const STATE_IS_LOCKED: usize = 0b001;
+pub(crate) const STATE_IS_PARKED: usize = 0b010;
+pub(crate) const STATE_IS_COUNTING: usize = 0b100;
+
+pub(crate) const TOKEN_NORMAL: UnparkToken = UnparkToken(0);
+pub(crate) const TOKEN_HANDOFF: UnparkToken = UnparkToken(1);
+
+#[derive(PartialEq, Eq, Debug)]
```
vext01, line 309: Please add docstrings. What is `x`?

ykrt/src/location.rs
vext01: Especially since debug assertions are on by default, since we don't (and can't) build the external workspace in release mode.

(This has been bugging me for some time)

ykrt/src/location.rs
```diff
@@ -97,88 +29,407 @@ impl State {
...
+            let ls = self.load(Ordering::Relaxed);
+            debug_assert!(!ls.is_counting());
+            let new_ls = ls.with_lock().with_unparked();
+            if self
+                .state
+                .compare_exchange_weak(
+                    ls.with_unlock().with_unparked().x,
+                    new_ls.x,
+                    Ordering::Acquire,
+                    Ordering::Relaxed,
```
vext01, line 151: It seems you came to the same conclusion I did about the second ordering. If it fails, we don't care what ordering we use.

Perhaps there's a niche use-case where you do care, but ours isn't one.

ykrt/src/location.rs
```diff
@@ -97,88 +29,407 @@ impl State {
...
+            let hr = unsafe { ls.hot_location() };
+            unsafe {
+                Box::from_raw(hr);
+            }
+        }
+    }
+}
+
+#[cfg(target_pointer_width = "64")]
+pub(crate) const STATE_TAG: usize = 0b111; // All of the other tag data must fit in this.
```
vext01, line 298: Does it make more sense to use a u64? Then it at-least works on 32-bit arches.

TBH, I don't think we care much about those chips, but we can at least allow it to work (albeit slowly).

Of course there are other bits of the system that wouldn't work...

ykrt/src/location.rs
```diff
@@ -97,88 +29,407 @@ impl State {
...
+#[cfg(target_pointer_width = "64")]
+pub(crate) const STATE_NUM_BITS: usize = 3;
+
+pub(crate) const STATE_IS_LOCKED: usize = 0b001;
+pub(crate) const STATE_IS_PARKED: usize = 0b010;
+pub(crate) const STATE_IS_COUNTING: usize = 0b100;
+
+pub(crate) const TOKEN_NORMAL: UnparkToken = UnparkToken(0);
```
vext01, line 306: Do these make more sense as an enum with explicit discriminant numbers? I appreciate that the example code does the same... up to you.

ykrt/src/location.rs
```diff
@@ -97,88 +29,407 @@ impl State {
...
+    /// Return a version of this `State` with the parked bit set.
+    pub(crate) fn with_parked(&self) -> State<I> {
+        State {
+            x: self.x | STATE_IS_PARKED,
+            marker: PhantomData,
+        }
+    }
+
+    /// Return a version of this `State` with the parked bit unset.
+    pub(crate) fn with_unparked(&self) -> State<I> {
```
vext01, line 393: Is `without_parked` a better name? The logic seems upside-down to me, but my head is a weird place. U p to you.

ykrt/src/mt.rs
```diff
@@ -219,215 +217,206 @@ impl MTThread {
         &mut self,
         loc: &Location<I>,
     ) -> Option<fn(&mut I) -> *mut RawStopgapInterpreter> {
-        // Since we don't hold an explicit lock, updating a Location is tricky: we might read a
```
vext01, line 222: A few lines above we have `statemachine` which should be `state machine`.

ykrt/src/mt.rs
```diff
@@ -219,215 +217,206 @@ impl MTThread {
...
-        // Location, work out what we'd like to update it to, and try updating it, only to find
-        // that another thread interrupted us part way through. We therefore use compare_and_swap
-        // to update values, allowing us to detect if we were interrupted. If we were interrupted,
-        // we simply retry the whole operation.
-
-        // We need Acquire ordering, as PHASE_COMPILING and PHASE_COMPILED need to read information
+        // XXX We need Acquire ordering, as PHASE_COMPILING and PHASE_COMPILED need to read information
```
vext01, line 220: Is the XXX left in accidentally?

ykrt/src/mt.rs
```diff
@@ -219,215 +217,206 @@ impl MTThread {
...
         // written to external data. Alternatively, this load could be Relaxed but we would then
         // need to place an Acquire fence in PHASE_COMPILING and PHASE_COMPILED.
-        let mut lp = loc.load(Ordering::Acquire);
-
-        match lp.phase() {
-            PHASE_COUNTING => {
+        let mut ls = loc.load(Ordering::Acquire);
```
vext01, line 223: ls is "lock state"?

ykrt/src/mt.rs
```diff
@@ -219,215 +217,206 @@ impl MTThread {
...
-                        lp = loc.compare_and_swap(lp, new_state, Ordering::Relaxed);
-                        if lp == old_lp {
-                            return None;
-                        }
-                        // We raced with another thread.
-                        if lp.phase() != PHASE_COUNTING {
-                            // The other thread probably moved the Location from PHASE_COUNTING to
-                            // PHASE_TRACING.
-                            return None;
+                    match loc.compare_exchange_weak(
```
vext01, line 239: I was about to suggest a simplification involving `fetch_add`, but this could go awry. If another thread transitions us out of counting, we'd end up incrementing an address!

ykrt/src/mt.rs
```diff
@@ -219,215 +217,206 @@ impl MTThread {
...
+                        ls.with_count(count + 1),
+                        Ordering::Relaxed,
+                        Ordering::Relaxed,
+                    ) {
+                        Ok(_) => break,
+                        Err(new_ls) => {
+                            if !new_ls.is_counting() {
+                                // Another thread has probably started tracing this Location.
+                                break;
+                            }
```
vext01, line 250: I wonder if it's worth making this `else if` to make it clear that the two cases are exclusive? Tiny nit, almost not worth thinking about.

ykrt/src/mt.rs
vext01: Here's a weird (and flawed) idea (that you may have already considered).

If we split the counting and address fields into separate atomic integers, then there's no harm in incrementing the count when we've already transitioned to tracing and then this block of code becomes much simpler.

This means the state no longer fits in a usize, which is unacceptable, but I wanted to think out of the box anyway...

This comment is unactionable :P

ykrt/src/mt.rs
```diff
@@ -219,215 +217,206 @@ impl MTThread {
...
-
-                    if lp.phase() != PHASE_TRACING {
-                        // While waiting for the Location to move beyond PHASE_TRACING_LOCK, we
-                        // probably raced with the thread tracing this location.
-                        return None;
-                    }
+                return None;
+            } else {
+                if self.inner.tracing.is_some() {
+                    // This thread is already tracing another Location, so either another
```
vext01, line 269: debug_assert that the HotLocation struct agrees that we are tracing?

ykrt/src/location.rs
```diff
@@ -97,88 +29,407 @@ impl State {
...
+    // there.
+    //
+    // The state machine is encoded in a usize in a not-entirely-simple way, as we don't want to
+    // allocate any memory for Locations that do not become hot. The layout is as follows (on a 64
+    // bit machine):
+    //
+    //   bit(s) | 63..3   | 2           | 1         | 0
+    //          | payload | IS_COUNTING | IS_PARKED | IS_LOCKED
+    //
+    // In the Counting state, IS_COUNTING is set to 1, and the parking lot mutex bits are unused
```
ptersilie, line 71: "and the parking lot mutex bits (IS_PARKED and IS_LOCKED) are unused"

ykrt/src/mt.rs
```diff
@@ -219,215 +217,206 @@ impl MTThread {
...
-
-                // Before we can do anything with this phase's value, we need to transition to
-                // `PHASE_TRACING_LOCK`.
+                // To avoid racing with another thread that may also try starting to trace this
+                // location at the same time, we need to initialise and lock the Location, which we
+                // perform in a single step. Since this is such a critical step, and since we're
+                // prepared to bail out early, there's no point in yielding: either we win the race
+                // by trying repeatedly or we give up entirely.
+                let hl_ptr = Box::into_raw(Box::new(HotLocation::Tracing(None)));
+                let new_ls = State::new().with_hotlocation(hl_ptr).with_lock();
```
vext01, line 281: Is there an alignment guarantee involved here? Why is is safe to mutate the lower bits of the pointer? Do we need to comment?

vext01, line 281: Does it make sense to have a pre-allocated initial hotlocation which we just copy into a box each time? Probably not worth the effort?

ykrt/src/mt.rs
```diff
@@ -219,215 +217,206 @@ impl MTThread {
...
-
-                let loc_tid = unsafe { Arc::from_raw(lp.pointer_data::<ThreadIdInner>()) };
-                if let Some((_, ref tid)) = self.inner.tracer {
-                    // This thread is tracing something...
-                    if !Arc::ptr_eq(tid, &loc_tid) {
-                        // ...but we didn't start at the current Location.
-                        loc.store(lp, Ordering::Relaxed);
-                        mem::forget(loc_tid);
+            }
+        } else {
```
vext01, line 314: This `else` is so far away from it's `if` that it makes it hard to find, especially with all the other nested conditionals in close proximity. Does it make sense to put the bodies of the arms into inlined functions? I think it does.

ykrt/src/mt.rs
```diff
@@ -219,215 +217,206 @@ impl MTThread {
...
+                            // start tracing for real.
+                            let tid = Arc::clone(&self.inner.tid);
+                            let tt = start_tracing(self.inner.tracing_kind);
+                            *unsafe {new_ls.hot_location()} = HotLocation::Tracing(Some((tid, tt)));
+                            Rc::get_mut(&mut self.inner).unwrap().tracing =
+                                Some(hl_ptr as *const ());
+                            loc.unlock();
+                            return None;
+                        }
+                        Err(x) => {
```
ptersilie, line 301: You do love to use `x` as a variable in this PR. :stuck_out_tongue: I can guess it's a `State`, but it's not super clear. Maybe name the variable `state`?

ykrt/src/mt.rs
```diff
@@ -219,215 +217,206 @@ impl MTThread {
...
-                // PHASE_LOCKED so that only this thread will try to free the memory.
-                if loc.compare_and_swap(lp, State::phase_locked(), Ordering::Acquire) != lp {
-                    // We raced with another thread that's probably transitioned this Location to
-                    // PHASE_LOCK.
-                    return None;
+            let hl = unsafe { ls.hot_location() };
+            let hl_ptr = hl as *mut _ as *mut ();
+            match hl {
+                HotLocation::Compiled(tr) => {
+                    // FIXME: we need to refcount the compiled trace to know how many threads are
```
vext01, line 335: Why to we worry about how many threads are executing a trace? Currently we don't discard trace code.

ykrt/src/mt.rs
```diff
@@ -219,215 +217,206 @@ impl MTThread {
...
+                    // currently executing it.
+                    let f = unsafe {
```
vext01, line 337: Do I see it correctly that you need to hold the lock to execute a trace? This would be bad, a bit like Python's GIL.

ykrt/src/mt.rs
```diff
@@ -219,215 +217,206 @@ impl MTThread {
...
-                        if let Some(tr) = (*gd).take() {
-                            let f = unsafe {
-                                mem::transmute::<_, fn(&mut I) -> *mut RawStopgapInterpreter>(
-                                    tr.ptr(),
-                                )
-                            };
-                            loc.store(State::phase_compiled(tr), Ordering::Release);
-                            return Some(f);
+                HotLocation::Compiling(mtx) => {
+                    let tr = {
```
vext01, line 344: I don't understand what is going on here. Comment?

vext01:
> access to which is controlled by the two parking lot mutex bits.

Is that to say, I must hold the lock if I want to read or write the hotlocation?

vext01:
Is the reprofile edge accurate? I don't think we do that, do we?

ykrt/src/location.rs
ltratt: This would be called `0` if it wasn't for the `PhantomType`. I don't think naming this or adding a doc string is very useful.

ykrt/src/location.rs
ltratt: Correct (`parking_lot` does the same thing for the same reason).

ykrt/src/location.rs
ltratt: > Does it make more sense to use a u64?

No, because these need to encode pointers, so you want to use a native word type -- for CHERI-wide pointers, for example, this would need to be 128 bits.

ykrt/src/location.rs
ltratt: These are from `parking_lot_core` so we can't encode them as an `enum`.

ykrt/src/location.rs
ltratt: Good question! I'm not super happy with the name -- let me chew on it.

ykrt/src/mt.rs
ltratt: Huh? I'm not sure what this is talking about. That said, the comment here is clearly incorrect, and the Ordering is unnecessarily strong. Fixed in 80db78f86a108897087f83b958a6ede52671685e.

ykrt/src/mt.rs
ltratt: Fixed in 80db78f86a108897087f83b958a6ede52671685e.

ykrt/src/mt.rs
ltratt: "location state"

ykrt/src/mt.rs
ltratt: Good point! I had vaguely thought about `fetch_add` but decided against it for no deep reason. I think you're right that it would be a bad idea!

ykrt/src/mt.rs
ltratt: That wouldn't be correct: at this point in the code we want to *start* tracing this location, but we don't want to do so if this thread is tracing another location. [One way of thinking of this: we can't start PT twice if we encounter a hot inner loop.]

ykrt/src/location.rs
ltratt: Fixed (slightly differently) in 281452f2191374b4c090a5c2eb93c8da7e894b09.

ykrt/src/mt.rs
ltratt: There is an alignment check, it's enforced in `location.rs`:

```
debug_assert_eq!(hl_ptr & !STATE_TAG, hl_ptr);
```

ltratt: I'm not quite sure I understand. Each location needs a pointer to a new chunk on the heap, otherwise we don't have something we can lock.

ykrt/src/mt.rs
ltratt: I agree, this isn't ideal. However, I think we need to make some other changes in subsequent PRs that *might* negate the need for this -- I'm not really sure. Summary: I think we should wait a bit before trying to make this prettier.

ykrt/src/mt.rs
ltratt: It could be argued this should be named `ls` but that would conflict with the `ls` in an outer scope. Naming is a tar pit!

ykrt/src/mt.rs
ltratt: In the long run we may want to recompile code and at the moment we can't free a trace because we can never be sure if someone's executing in it. This "FIXME" probably won't be addressed for a long time!

ykrt/src/mt.rs
ltratt: No, you need to gain the lock merely to gain the pointer, then you release the lock, then you execute the code at the pointer. [I do have a vague idea for not even needing this lock, but I want to get this correct before I worry about performance.]

ykrt/src/mt.rs
ltratt: Do the `// Compilation is ongoing.` comments within this `if` not explain things a bit?

ltratt:
> Without this, "parking lot mutex bits" is undefined. It's not hard to infer, but clarity is best.

Fixed in 281452f2191374b4c090a5c2eb93c8da7e894b09.

> Did you say it's pinned, or at least heap allocated? Should we mention this and why?

Pointers to heap memory are implicitly pinned in Rust, so we don't need to say that IMHO. I thought "boxed location" would make clear it's on the heap?

> Is that to say, I must hold the lock if I want to read or write the hotlocation?

Yes.

> Is the reprofile edge accurate? I don't think we do that, do we?

Not yet, but I thought it's helpful to remind us that we *will* want to do this at some point. Knowing that does change the way you code the statemachine up.

ykrt/src/location.rs
vext01: Ah, ok.

ykrt/src/location.rs
vext01: Why don't you hide it behind a `Deref`? Both @ptersilie and I were confused by `x`.

ykrt/src/location.rs
vext01: or merge `with_parked` and `with_unparked` into a `with_parked(val: u8)`?

ykrt/src/mt.rs
vext01: Line 215: `/// `Location`s represent a statemachine`

ykrt/src/mt.rs
vext01: Jolly good!

vext01: Never mind. In hindsight it's not worth it.

ykrt/src/mt.rs
vext01: I'm OK with that.

ykrt/src/mt.rs
vext01: Can we qualify the statement?

> // FIXME: If we ever decide to evict/invalidate trace code, then ...`

ykrt/src/mt.rs
vext01: It appears that compilation can be ongoing in two different ways?

ykrt/src/location.rs
vext01: I realised that the `x` Lukas is questioning is a different `x`!

ltratt: In the long run, I'm not very happy with the relationship between `Location` and `State` and I want to try tidying it up. I don't want to get too bogged down in this now though.

ykrt/src/location.rs
vext01: Fair enough.

ykrt/src/mt.rs
ltratt: Good point. Fixed in a3d9eb55c15524789ca14428ab41a73114be3e7e.

ykrt/src/mt.rs
ltratt: Not really: we just have to handle a potential race condition, so we have to do two levels of check (first the mutex; then the `Option`).

ykrt/src/mt.rs
ltratt: Um, and ...? :)

ykrt/src/location.rs
ltratt: Hmm, honestly, I'm not sure. I'm starting to think that this might end up being part of the `State` rethink, and perhaps is best deferred until then.

vext01:
> Pointers to heap memory are implicitly pinned in Rust.

What I'm getting at is that for this locking scheme to work, the allocation must not move. Is that something important to mention? Maybe it's implied by the use of the parking lot...

ltratt:
> What I'm getting at is that for this locking scheme to work, the allocation must not move.

Allocations in Rust fundamentally can't move, so I don't see that stating this buys us anything. If allocations could move, every Rust program with raw pointers would break!

vext01:
Having reviewed this (somewhat superficially I must admit), I have two overall observations:

 - a) The complexity isn't reduced as much as I'd hoped. It remains complicated and I'm certain we will find concurrency bugs. But...
 - b) the complexity follows an established design pattern (i.e. a parking lot mutex), so it may be easier for an outsider to understand by that virtue.

Due to b) I hardly investigated the mutex code. It's very intimidating, but I feel confident that, given time and the parking lot docs, I could figure it out if the bus factor dropped below 1.

@ptersilie how do you feel about this?

ptersilie:
I feel similar. I find both the old cold and the new code in this PR quite overwhelming. Although I agree, that this seems slightly easier to understand from what I can tell with my shallow review and the diagram at the top.

ltratt:
In a sense this PR decreases complexity in one way (less lockless code) but increases complexity in another (threads now actually go to sleep properly). I do hope to slightly reduce the complexity in the future, although I don't think it's going to become much simpler.

Time to squash?

vext01:
I think so... To be clear, the performance hit will be raked back when a trace properly loops?

ltratt:
> To be clear, the performance hit will be raked back when a trace properly loops?

I expect so *largely* but of course I can't actually benchmark that either way yet!

vext01:
@ptersilie ok to squash / merge?

ptersilie:
Yes, please squash.

ltratt:
Before I squash here are the benchmarks (well, all two of them!) before:

```
test mt::tests::bench_multi_threaded_control_point  ... bench:      99,416 ns/iter (+/- 46,737)
test mt::tests::bench_single_threaded_control_point ... bench:   5,909,249 ns/iter (+/- 83,698)
```

and after:

```
test mt::tests::bench_multi_threaded_control_point  ... bench:     197,872 ns/iter (+/- 16,771)
test mt::tests::bench_single_threaded_control_point ... bench:  12,221,330 ns/iter (+/- 232,520)
```

So in wall-clock time, in pretty much the worst case, we're 2x slower. Could be worse.

Squashed.

PR #268, opened 2021-02-25
Make `cargo xtask bench` work.

### LINKED ISSUE #264: xtask prevents us running #[bench] targets?
If I try `cargo xtask bench` I am told:

```
xtask: the build system does not support the bench target
```

Is this insurmountable, or is there a way to get our existing `#[bench]` targets back up and running?

vext01:
CC @bjorn3.

Fixes #264

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -78,7 +78,7 @@ fn run_action(workspace: Workspace, target: &str, extra_args: &[String]) {
                 cmd.arg("--release".to_string());
             }
         }
-        "build" | "check" | "clippy" | "test" => {
+        "bench" | "build" | "check" | "clippy" | "test" => {
             // Ensure that the whole workspace is tested and not just the base crate in the
             // workspace.
             if target == "test" {
@@ -89,8 +89,11 @@ fn run_action(workspace: Workspace, target: &str, extra_args: &[String]) {
                 let tracing_kind = find_tracing_kind(&rust_flags);
                 rust_flags = make_internal_rustflags(&rust_flags);
 
-                // Optimise the internal workspace.
-                cmd.arg("--release");
+                // Optimise the internal workspace. Don't pass `--release` for `cargo bench` however as it is
+                // already implied (and explicitly passing it is illegal).
+                if target != "bench" {
+                    cmd.arg("--release");
+                }
                 // Set the tracermode cfg macro, but without changing anything relating to code
                 // generation. We can't use `-C tracer=hw` as this would turn off optimisations
                 // and emit SIR for stuff we will never trace.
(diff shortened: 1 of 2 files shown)


vext01:
Is now a good time to revisit https://github.com/softdevteam/yk/issues/197 by the way?

xtask/src/main.rs
```diff
@@ -89,8 +89,10 @@ fn run_action(workspace: Workspace, target: &str, extra_args: &[String]) {
                 let tracing_kind = find_tracing_kind(&rust_flags);
                 rust_flags = make_internal_rustflags(&rust_flags);
 
-                // Optimise the internal workspace.
-                cmd.arg("--release");
+                if target != "bench" {
+                    // Optimise the internal workspace.
+                    cmd.arg("--release");
+                }
```
bjorn3, line 95: Why is this necessary? Shouldn't the internal workspace always be optimized?

vext01:
```
$ cargo bench --release
error: Found argument '--release' which wasn't expected, or isn't valid in this context
```

I guess the implication is that benchmarks are always optimised.

bjorn3: Can you add a comment?

ltratt: I think we can just tweak the existing comment to something like "`--bench` implies `--release`; otherwise we need to add it"?

vext01: Try f1eb1c9

xtask/src/main.rs
```diff
@@ -89,8 +89,11 @@ fn run_action(workspace: Workspace, target: &str, extra_args: &[String]) {
                 let tracing_kind = find_tracing_kind(&rust_flags);
                 rust_flags = make_internal_rustflags(&rust_flags);
 
-                // Optimise the internal workspace.
-                cmd.arg("--release");
+                // Optimise the internal workspace but, not for `cargo bench` where `--release` is
+                // already implied (and explicitly passing it is illegal).
```
bjorn3, line 93: How about

```suggestion
                // Optimise the internal workspace. Don't pass `--release` for `cargo bench` however as it is
                // already implied (and explicitly passing it is illegal).
```

vext01:
OK to squash?

ltratt:
Please squash.

vext01:
splat.

xtask/src/main.rs
```diff
@@ -89,8 +89,11 @@ fn run_action(workspace: Workspace, target: &str, extra_args: &[String]) {
                 let tracing_kind = find_tracing_kind(&rust_flags);
                 rust_flags = make_internal_rustflags(&rust_flags);
 
-                // Optimise the internal workspace.
-                cmd.arg("--release");
+                // Optimise the internal workspace. Don't pass `--release` for `cargo bench` however as it is
```
vext01, line 92: CI should have failed, as this line is too long :\

More xtask debugging...

PR #269, opened 2021-02-25
Correctly handle `cargo xtask fmt` args.

vext01:
We just need to pass down extra args to the internal workspace.

[I also learned that by default `rustfmt` will not wrap long comments]

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -23,7 +23,7 @@ enum Workspace {
 fn run_action(workspace: Workspace, target: &str, extra_args: &[String]) {
     // The external workspace depends on libykshim.so produced by the internal workspace
     if workspace == Workspace::External {
-        run_action(Workspace::Internal, target, &[]);
+        run_action(Workspace::Internal, target, extra_args);
     }
 
     let mut cmd = if ["fmt", "clippy"].contains(&target) {
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -89,8 +89,7 @@ fn run_action(workspace: Workspace, target: &str, extra_args: &[String]) {
                 let tracing_kind = find_tracing_kind(&rust_flags);
                 rust_flags = make_internal_rustflags(&rust_flags);
 
-                // Optimise the internal workspace. Don't pass `--release` for `cargo bench`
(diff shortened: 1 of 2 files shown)


xtask/src/main.rs
```diff
@@ -89,8 +89,8 @@ fn run_action(workspace: Workspace, target: &str, extra_args: &[String]) {
                 let tracing_kind = find_tracing_kind(&rust_flags);
                 rust_flags = make_internal_rustflags(&rust_flags);
 
-                // Optimise the internal workspace. Don't pass `--release` for `cargo bench` however as it is
-                // already implied (and explicitly passing it is illegal).
+                // Optimise the internal workspace. Don't pass `--release` for `cargo bench`
```
ltratt, line 92: On reflection, I think it's fairly obvious that `--release` implies optimisation, so maybe we can simplify this to something like "`cargo bench` errors if you pass `--release`"?

vext01: > On reflection, I think it's fairly obvious that --release implies optimisation

Your proposed text is fine, but not that the subtlety is that `cargo bench` implies `--release`, not that `--release` implies optimisation.

[Maybe you meant that?]

ltratt: "cargo bench implies --release" yes "not that --release implies optimisation" um no?

vext01: I think you misunderstood.

The fact that --release implies optimisation, whilst true, is irrelevant.

What matters here is that `cargo bench` implies optimisation, and this is the root cause of why `--release` is invalid in that context.

vext01: We can be more concise with:
```
// Optimise the internal workspace. `--release` is not valid for `cargo bench`.
```

ltratt: I think you are misunderstanding me :) But since I'm happy with your proposed new text (it's entirely compatible with my understanding), I don't think we should worry about it. Let's use your newest suggestion.

vext01: OK, so e9556f2 fixes that.

ltratt:
Please squash.

vext01:
There's no squashing to do (these all address separate things), but I did reword a commit message. Please go ahead.

PR #271, opened 2021-02-28
Fixes suggested by Clippy

ltratt:
This PR does a semi-sweep of Clippy's suggestions for the repository (there are some others that I haven't tackled yet). I've broken https://github.com/softdevteam/yk/commit/2220368c161af55829f6a38f3f152337e453a2c8 into its own commit, because it's a slightly odd case of pointless variable assignment, where the "seemingly correct" fix goes beyond what Clippy is capable of suggesting. https://github.com/softdevteam/yk/commit/4082770ead57966fcc58d5390f2b62c4752e50b1 is, mostly, more mechanical.

### DIFF
--- a/internal_ws/yksg/src/lib.rs
+++ b/internal_ws/yksg/src/lib.rs
@@ -95,7 +95,7 @@ impl LocalMem {
     }
 
     /// Copy over the call arguments from another frame.
-    pub fn copy_args(&mut self, args: &Vec<IRPlace>, frame: &LocalMem) {
+    pub fn copy_args(&mut self, args: &[IRPlace], frame: &LocalMem) {
         for (i, arg) in args.iter().enumerate() {
             let dst = self.local_ptr(&Local(u32::try_from(i + 1).unwrap()));
             match arg {
@@ -133,16 +133,13 @@ impl LocalMem {
             IRPlace::Indirect { ptr, off, ty: _ty } => {
                 // Get a pointer to the Indirect, which itself points to another pointer.
                 let dest_ptr = self.local_ptr(&ptr.local) as *mut *mut u8;
-                let ptr = unsafe {
+                unsafe {
                     // Dereference the pointer, by reading its value.
                     let mut p = std::ptr::read::<*mut u8>(dest_ptr);
                     // Add the offsets of the Indirect.
                     p = p.offset(isize::try_from(ptr.off).unwrap());
-                    p = p.offset(isize::try_from(*off).unwrap());
-                    p
-                };
-                // Now return the value as a pointer.
-                ptr
+                    p.offset(isize::try_from(*off).unwrap())
+                }
             }
             _ => unreachable!(),
         }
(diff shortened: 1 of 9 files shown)


ykrt/build.rs
```diff
@@ -41,18 +39,17 @@ fn main() {
...
     sym_dest.push("..");
     sym_dest.push(YKSHIM_SO);
     if !PathBuf::from(&sym_dest).exists() {
-        let mut sym_src = internal_dir.clone();
-        sym_src.push("target");
-        sym_src.push("release");
-        sym_src.push(YKSHIM_SO);
-        dbg!(&sym_src, &sym_dest);
-        symlink(sym_src, sym_dest).unwrap();
+        internal_dir.push("target");
```
bjorn3, line 48: I think keeping `let mut sym_src = internal_dir;` would be nicer as after these pushes it is no longer contains the internal dir. Just remove the `.clone()`.

internal_ws/yksg/src/lib.rs
```diff
@@ -133,16 +133,13 @@ impl LocalMem {
             IRPlace::Indirect { ptr, off, ty: _ty } => {
                 // Get a pointer to the Indirect, which itself points to another pointer.
                 let dest_ptr = self.local_ptr(&ptr.local) as *mut *mut u8;
-                let ptr = unsafe {
+                unsafe {
                     // Dereference the pointer, by reading its value.
                     let mut p = std::ptr::read::<*mut u8>(dest_ptr);
                     // Add the offsets of the Indirect.
                     p = p.offset(isize::try_from(ptr.off).unwrap());
```
bjorn3, line 140: Maybe
```suggestion
                    let p = std::ptr::read::<*mut u8>(dest_ptr);
                    // Add the offsets of the Indirect.
                    let p = p.offset(isize::try_from(ptr.off).unwrap());
```

internal_ws/yktrace/src/sir.rs
```diff
@@ -47,7 +48,7 @@ pub struct Sir<'m> {
 }
 
 impl<'m> Sir<'m> {
-    pub fn new(mmap: &'m Mmap) -> Result<Self, ()> {
+    pub fn new(mmap: &'m Mmap) -> Result<Self, Box<dyn Error>> {
```
bjorn3, line 51: Can you also replace all unwraps with `?`?

ltratt: I think so, but I need to take a higher level view of the API to understand that. I think that can be done in a subsequent PR.

ykrt/build.rs
ltratt: Makes sense. Fixed in e134a8216b2188fb00ac6ed9f3f6545de7016bf6.

internal_ws/yksg/src/lib.rs
ltratt: I'm neutral on this either way.

vext01:
This all looks OK to me. If you're happy, please squash.

ltratt:
Squashed.

PR #273, opened 2021-03-01
Don't run step_fn after stopgapping.

ptersilie:
After executing the stopgap interpreter we are back at the beginning of
the control point, so we must not execute `step_fn` immediately
afterwards.

Otherwise this leads to incorrect program behaviour as after stopgapping, `step_fn` is executed once more than it should be.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -195,6 +195,7 @@ impl MTThread {
                     unsafe {
                         let mut si = StopgapInterpreter(ptr);
                         si.interpret();
+                        return;
                     }
                 }
             }
@@ -912,8 +913,8 @@ mod tests {
             match ctx.prog[ctx.pc] {
                 INC => {
                     ctx.pc += 1;
-                    if ctx.run == 0 {
-                        ctx.run = 99;
+                    if ctx.run > 0 {
+                        ctx.run += 1;
                     }
                 }
                 RESTART => ctx.pc = 0,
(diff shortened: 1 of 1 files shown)


ptersilie:
Adding @vext01 since I briefly mention this earlier.

ykrt/src/mt.rs
```diff
@@ -912,8 +913,8 @@ mod tests {
             match ctx.prog[ctx.pc] {
                 INC => {
                     ctx.pc += 1;
-                    if ctx.run == 0 {
-                        ctx.run = 99;
+                    if ctx.run > 0 {
```
ltratt, line 916: This feels a bit weird: why `>` and why `+= 99`?

ptersilie: I needed to adjust the test to catch the bug. The `+=` is so that I can tell if `step_fn` ran twice. If the value is 100 it ran once, if its 199 it ran twice. That means I also had to change the condition so that the `if` triggers on both executions of `step_fn`.

ltratt: Silly question: are there smaller values which also highlight the bug, or is this the smallest (or close to the smallest) value?

ptersilie: Any number will do. I don't know why I chose `99`. Want me to pick something smaller?

ltratt: Yes, that would probably be a good plan :) Also when we come to squash, I think it would be good to add a bit more information to the commit message about what the bug was and why this commit fixes it (but we can worry about this in the squashing phase).

ptersilie:
Try this commit. Will update the commit message during squashing.

ltratt:
Please squash.

ptersilie:
Squashed and reworded.

PR #274, opened 2021-03-01
Start tweaking the unsafe code

ltratt:
This PR is a very lightweight pass over the stopgap interpreter's unsafe code. It marks a few more functions as `unsafe` (https://github.com/softdevteam/yk/commit/aa05f0503cec32c002ea07cce111081f97c768a0), reduces the visibility of unsafe code (https://github.com/softdevteam/yk/commit/72f039ad05024a8f3eb24a09ea76639cfb706740), and simplifies various code doing unsafe things (https://github.com/softdevteam/yk/commit/72f039ad05024a8f3eb24a09ea76639cfb706740). I also couldn't help myself noticing that we were wildly inconsistent in spelling `dst` and `dest`: the former makes more sense with `src`, so I did a `srep` renaming with that (https://github.com/softdevteam/yk/commit/4bca1ea7eb474f4b973ce18ec8f2726adadca983).

In one sense this PR doesn't do very much that's hugely useful: it doesn't properly document safety constraints, for example. That's because I haven't really got my head around some of that yet. But, since this seems a fairly safe (pun intended) small step, I thought I might as well raise this now, rather than storing it up for a later monster PR.

### DIFF
--- a/internal_ws/yksg/src/lib.rs
+++ b/internal_ws/yksg/src/lib.rs
@@ -62,34 +62,34 @@ impl LocalMem {
     }
 
     /// Stores one IRPlace into another.
-    fn store(&mut self, dst: &IRPlace, src: &IRPlace) {
+    unsafe fn store(&mut self, dst: &IRPlace, src: &IRPlace) {
         match src {
             IRPlace::Val { .. } | IRPlace::Indirect { .. } => {
                 let src_ptr = self.iplace_to_ptr(src);
                 let dst_ptr = self.iplace_to_ptr(dst);
                 let size = usize::try_from(SIR.ty(&src.ty()).size()).unwrap();
-                unsafe { std::ptr::copy(src_ptr, dst_ptr, size); }
+                std::ptr::copy(src_ptr, dst_ptr, size);
             }
             IRPlace::Const { val, ty: _ty } => {
                 let dst_ptr = self.iplace_to_ptr(dst);
-                unsafe { self.write_const(dst_ptr, val); }
+                self.write_const(dst_ptr, val);
             }
             _ => todo!(),
         }
     }
 
     /// Copy over the call arguments from another frame.
-    fn copy_args(&mut self, args: &[IRPlace], frame: &LocalMem) {
+    unsafe fn copy_args(&mut self, args: &[IRPlace], frame: &LocalMem) {
         for (i, arg) in args.iter().enumerate() {
             let dst = self.local_ptr(&Local(u32::try_from(i + 1).unwrap()));
             match arg {
                 IRPlace::Val { .. } | IRPlace::Indirect { .. } => {
                     let src = frame.iplace_to_ptr(arg);
                     let size = usize::try_from(SIR.ty(&arg.ty()).size()).unwrap();
-                    unsafe { std::ptr::copy(src, dst, size); }
+                    std::ptr::copy(src, dst, size);
                 }
                 IRPlace::Const { val, .. } => {
-                    unsafe { self.write_const(dst, val); }
+                    self.write_const(dst, val);
                 }
                 _ => unreachable!(),
             }
@@ -225,7 +225,9 @@ impl StopgapInterpreter {
         // start.
         let body = frame.body.clone();
         let bbidx = usize::try_from(frame.bbidx).unwrap();
-        sg.terminator(&body.blocks[bbidx].term);
+        unsafe {
+            sg.terminator(&body.blocks[bbidx].term);
+        }
         sg
     }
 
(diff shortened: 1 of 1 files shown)


ltratt:
In retrospect, I think https://github.com/softdevteam/yk/pull/274/commits/4bca1ea7eb474f4b973ce18ec8f2726adadca983 shouldn't be part of this PR. I also should probably rebase against master before review. @ptersilie OK for me to force push an update before review?

ptersilie:
Sure. Please go ahead.

ltratt:
Rebased and ready for review.

internal_ws/yksg/src/lib.rs
```diff
@@ -44,65 +44,53 @@ impl Drop for LocalMem {
...
-                if SIR.ty(t).size() == 0 {
-                    // ZST: do nothing.
-                } else {
-                    todo!()
-                }
-            }
+            Constant::Bool(b) => {
+                // XXX are we sure that we're always supposed to copy 1 byte, and not instead to
+                // set 1 bit?
+                std::ptr::copy(b as *const bool, dst as *mut bool, 1)
```
ptersilie, line 61: This just stores a bool in a variable, e.g. `$1 = true`. So, I'm fairly certain we are not supposed to set only a bit here.

ltratt: I *think* you're right, but I'm not 100% sure. Where should we check to be 100% sure?

ptersilie:
This looks fine and I agree with the removal of the unnecessary `pub`s.

I'm not sure how I feel about the `unsafe`s though. After yesterday's arduous discussion about proper use of `unsafe`, the summary was that `unsafe` should be used if there's a precondition that needs to be checked by the caller. There is certainly a precondition here in that the caller of `write_const` needs to make sure that the destination is a valid pointer and can hold the constant. However, since the stopgap interpreter runs on SIR blocks, it will only run code that has already been (type) checked by `rustc`, so there is no way the caller of `write_const` can actually verify that the precondition is satisfied.

ltratt:
My aim in this PR -- a fairly shallow aim, I admit -- was to bring out how much `unsafe` code we actually have. There's less than I thought: what's pleasingly surprising to me is that we don't have to add `unsafe` all over the place. My personal opinion, though this is something reasonable people can differ on, is that this is a useful stepping stone to rethinking the `unsafe` story entirely (which might, perhaps, end up with us getting rid of the `unsafe` functions entirely).

internal_ws/yksg/src/lib.rs
ptersilie: Setting bits, e.g. `a &= 0b0001` will result in a `BinOp:BitAnd` operation, which we haven't implemented yet. So far the interpreter has only implemented `BinOp::Add` and `BinOp::Lt`.

`write_const` is only used in `store` (and `copy_args`) which just assigns values to a local and doesn't do any binary operations.

ltratt: Does that mean setting a byte is definitely OK?

ptersilie:
Will we be "rethinking the `unsafe` story` in this PR, or will the thinking be done in another one? In other words, I understand your reasoning and am happy to merge this, if we don't wait too long with the rethinking.

ltratt:
I am already trying to think about the next step :) I hope this PR is a small step on that path.

internal_ws/yksg/src/lib.rs
ptersilie: Since `write_const` is called from `store`, rust will have ensured that the constant and the destination have the same type. So assuming `bool`'s in Rust are 1 byte, copying 1 byte is fine here.

I hope this answers your question, since I wasn't quite sure what exactly you were asking.

ltratt: I checked, and I agree https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=edb602f2e4628e4d3a6a553ad174aee2

ptersilie: We chould change `write_const` to take a size as an argument, which is passed by `store`, so that we don't have to hardcode sizes in. Although, `bool` might be the only case where this is necessary, or can you think of any others?

ltratt: I don't think we need to worry further about this.

ptersilie:
Anything needs squashing?

ltratt:
I *think* it's OK as-is.

ltratt:
I forgot to remove the `XXX` comment (https://github.com/softdevteam/yk/pull/274/commits/5a7c8b6fd75a9816ba9870e04c1efea9553115b1) and I've also realised that we still have *some* custom rustfmt settings, so my local `fmt` doesn't work (reformatted in https://github.com/softdevteam/yk/pull/274/commits/aa4d365c883d2c90da6ac9bab52e9a08612f49ca). OK to squash?

ptersilie:
Ah, lucky it failed then. Please squash.

ptersilie:
You may also need to rebase.

ltratt:
Rebased and squashed.

PR #275, opened 2021-03-02
More thoroughly test binary operations.

vext01:
This will probably be tricky to review. It's all details.

In the end I opted *not* to split the codegen for MUL and DIV. Although they have different semantics, under the subset of conditions that we use these instructions, they are in fact very similar. I've added a big comment attempting to explain this.

We use macros to help us generate a lot of tests.

Several nasty bugs found and fixed along the way.

### DIFF
--- a/internal_ws/ykcompile/src/arch/x86_64/mod.rs
+++ b/internal_ws/ykcompile/src/arch/x86_64/mod.rs
@@ -123,45 +123,125 @@ macro_rules! binop_add_sub {
 /// Generates functions for mul/div-style operations.
 /// The first operand must be in a register.
 macro_rules! binop_mul_div {
+    // The addressing modes and semantics of x86_64 MUL and DIV are as follows:
+    //
+    //   Encoding       Semantics
+    //   --------------------------------------------------------------------
+    //   MUL r/m8:      AX * r/m8          AX <- result
+    //   MUL r/m16:     AX * r/m16         DX:AX <- result
+    //   MUL r/m32:     EAX * r/m32        EDX:EAX <- result
+    //   MUL r/m64:     RAX * r/m64        RDX:RAX <- result
+    //
+    //   DIV r/m8:      AX / r/m8          AL <- quotient,  AH <- remainder
+    //   DIV r/m16:     DX:AX / r/m16      AX <- quotient,  DX <- remainder
+    //   DIV r/m32:     EDX:EAX * r/m32    EAX <- quotient, EDX <- remainder
+    //   DIV r/m64:     RDX:RAX * r/m64    RAX <- quotient, RDX <- remainder
+    //
+    // A couple of thing to note:
+    //   - Both MUL and DIV's first (implicit) operand is twice as large as the second (explicit)
+    //     operand. For all but r/m8 variants, the operand extends into RDX.
+    //   - Similarly, the result of both MUL and DIV is twice as large as the second operand.
+    //
+    // SIR only deals with same-sized operands and doesn't require us to capture the remainder for
+    // division, so we don't use this extended operand/result space. This makes MUL and DIV very
+    // similar from our point of view, hence this macro is used to generate both implementations.
+    //
+    // We are however, required to zero the unused extended operand space to get the correct
+    // result, and we must be careful to save and restore anything that would be clobbered.
+    // Specifically:
+    //   - RAX is always clobbered, so we must always save and restore it.
+    //   - For all but the r/m8 variations, RDX is clobbered. So if the size of our SIR operands is
+    //     >1 byte, then we must also save and restore RDX.
+    //   - If our register allocator has put the second SIR operand in a register that will be
+    //     clobbered, then we cannot use this register as the r/m operand. In this case we borrow
+    //     RCX temporarily.
     ($name: ident, $op:expr) => {
         fn $name(&mut self, opnd1_reg: u8, opnd2: &IRPlace) {
-            // mul and div overwrite RAX, RDX, so save them first.
+            let size = SIR.ty(&opnd2.ty()).size();
+            // For all but r/m8 variants, RDX is clobbered.
+            if size > 1 {
+                dynasm!(self.asm
+                    ; push rdx
+                );
+            }
+            // RAX is always clobbered.
             dynasm!(self.asm
                 ; push rax
-                ; push rdx
-                ; xor rdx, rdx
             );
+            // Set up first operand.
             dynasm!(self.asm
                 ; mov rax, Rq(opnd1_reg)
             );
-            let size = SIR.ty(&opnd2.ty()).size();
+            // Set up second operand.
             let src_loc = self.iplace_to_location(opnd2);
             match src_loc {
-                Location::Reg(r) => match size {
-                    1 => {
+                Location::Reg(src_r) => {
+                    // Handle cases where our input operands clash with a clobbered register.
+                    let (r, borrow_rcx) = if src_r == RAX.code() {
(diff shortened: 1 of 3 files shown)


internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -123,45 +123,121 @@ macro_rules! binop_add_sub {
...
                         dynasm!(self.asm
-                            ; $op Rq(r)
+                            ; pop rcx
                         );
                     }
-                    _ => unreachable!(format!("{}", SIR.ty(&opnd2.ty()))),
                 },
                 Location::Mem(..) => todo!(),
                 Location::Const { val, .. } => {
+                    dynasm!(self.asm // FIXME not always.
```
ltratt, line 238: "not always" is gnomic! What does it mean?

internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -123,45 +123,121 @@ macro_rules! binop_add_sub {
...
                         dynasm!(self.asm
-                            ; $op Rd(r)
+                            ; xor rdx, rdx
                         );
                     }
-                    8 => {
+                    match size {
+                        1 => {
+                            dynasm!(self.asm
+                                ; and ax, 0xff // Zero unused extended operand space.
```
ltratt, line 208: Didn't we do this on line 199 (or, at least, the comment matches)?

vext01: We did it for size > 1.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
vext01: That's left in by accident.

ltratt: OK, I shall look forward to the fix commit :)

internal_ws/ykcompile/src/arch/x86_64/mod.rs
vext01: We could move it up in an else branch if you like?

ltratt: Dumb question: if we did it for size >1 on line 199, why do we then do it for size==1 here? Couldn't we handle both cases on line 199?

vext01: We raced with the same idea.

I only sneaked the `and` in there to make the code shorter, but at the cost of readability I admit.

1fadfad fixes it.

ltratt:
I'm happy and will leave it to @ptersilie to comment.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -123,45 +123,121 @@ macro_rules! binop_add_sub {
...
         fn $name(&mut self, opnd1_reg: u8, opnd2: &IRPlace) {
-            // mul and div overwrite RAX, RDX, so save them first.
+            let size = SIR.ty(&opnd2.ty()).size();
+            // For all but r/m8 variants, RDX is clobbered.
+            if size > 1 {
+                dynasm!(self.asm
+                    ; push rdx
+                );
+            }
+            // RAX is an always clobbered.
```
ptersilie, line 168: s/an//

internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -123,45 +123,121 @@ macro_rules! binop_add_sub {
...
+                        let rax_off = i32::try_from(QWORD_REG_SIZE).unwrap();
                         dynasm!(self.asm
-                            ; $op Rb(r)
+                            ; push rcx
+                            ; mov rcx, [rsp + rax_off]
                         );
-                    }
-                    2 => {
+                        (RCX.code(), true)
+                    } else if size > 1 && src_r == RDX.code() {
```
ptersilie, line 189: Why do we need to do this? `rdx` hasn't been clobbered yet. Do we need to do this because we are `xor`ing `rdx` further below. In any case can we just do `mov rcx, rdx` here, as it looks like `rdx` is still intact at this point.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -123,45 +123,121 @@ macro_rules! binop_add_sub {
...
+                            ; push rcx
+                            ; mov rcx, [rsp + rdx_off]
                         );
-                    }
-                    4 => {
+                        (RCX.code(), true)
+                    } else {
+                        (src_r, false)
+                    };
+                    // Zero unused extended operand space.
```
ptersilie, line 199: Until I looked at the code below I couldn't parse this sentence as I read `zero` as a noun. Maybe say `Set unused extended operand space to zero`?

internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -123,45 +123,121 @@ macro_rules! binop_add_sub {
...
-                            ; $op Rq(r)
+                            ; pop rcx
                         );
                     }
-                    _ => unreachable!(format!("{}", SIR.ty(&opnd2.ty()))),
                 },
                 Location::Mem(..) => todo!(),
                 Location::Const { val, .. } => {
+                    dynasm!(self.asm // FIXME not always.
+                        ; xor rdx, rdx
```
ptersilie, line 242: This doesn't look right. If the operand is a constant, we zero `rdx`, but we only `push/pop rdx` if the operand size is 1. What happens with constants of bigger sizes? Do we then clobber `rdx` without restoring it?

internal_ws/ykcompile/src/arch/x86_64/mod.rs
vext01: Fixed in 0214bfc9.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
vext01: You're right. Fixed in Fixed in 0214bfc9.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
vext01: Fixed in 0214bfc9.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
vext01: You're right, it should be guarded by `if size >= 1`.

I've not yet fixed this because I'd like to find a test which fails under the current code.

ptersilie: `>=` or just `>`?

vext01: yes, `>`.

I'm struggling to get a test that hits this bug. The problem is that our unoptimised MIR copies a lot, and it's always an old unused copy that gets corrupted...

I think we should move on, but we could fix the bug and import the passing test anyway?

vext01: Well 7e87b29 fixes the bug and adds a test which **might** have triggered the bug, had our MIR and SIR been more optimal.

vext01:
@ptersilie OK to squash?

ptersilie:
Yes, please squash.

vext01:
splat.

PR #278, opened 2021-03-03
Add a 'testing' feature.

ltratt:
**This is a draft PR -- do not merge yet**

What we would like to do is mark some things (e.g. yksg's `set_interp_ctx` function) as `#[cfg(test)]` but since we have a cross-crate testing dependency (which is exacerbated, though not entirely caused by, the internal/external
workspace split), that doesn't work.

This commit adds a "testing" feature which is intended as yk's cross-crate substitute for "#[cfg(test)]". It enables us to a) clearly mark some parts of the code as only being necessary for testing b) only build those when the "testing" feature is enabled.

This does lead to some ugliness e.g. because we can't, as far as I know, conditionally compile a module, we have to stuff the contents of the "test_api.rs" modules into sub-modules. We also have to fiddle with our xtask build so that it passes down the "testing" feature. But, overall, I think the ugliness is probably worth the gain.

Addresses the spirit of (but not always the original text of!) https://github.com/softdevteam/yk/issues/277

### DIFF
--- a/internal_ws/ykshim/Cargo.toml
+++ b/internal_ws/ykshim/Cargo.toml
@@ -14,3 +14,6 @@ yksg = { path = "../yksg" }
 ykcompile = { path = "../ykcompile" }
 ykpack = { path = "../ykpack" }
 yktrace = { path = "../yktrace" }
+
+[features]
+testing = ["yksg/testing"]
--- a/internal_ws/ykshim/src/test_api.rs
+++ b/internal_ws/ykshim/src/test_api.rs
@@ -1,6 +1,7 @@
 //! The ykshim testing API.
 //!
 //! These functions are only exposed to allow testing from the external workspace.
+#![cfg(feature = "testing")]
 
 use libc::size_t;
 use std::convert::TryFrom;
@@ -113,6 +114,7 @@ unsafe extern "C" fn __ykshimtest_find_symbol(sym: *const c_char) -> *mut c_void
 
 /// Interpret a SIR body with the specified interpreter context.
 #[no_mangle]
+#[cfg(feature = "testing")]
 unsafe extern "C" fn __ykshimtest_interpret_body(body_name: *const c_char, ctx: *mut u8) {
     let fname = CStr::from_ptr(body_name).to_str().unwrap().to_string();
     let mut si = StopgapInterpreter::from_symbol(fname);
--- a/ykshim_client/src/test_api.rs
+++ b/ykshim_client/src/test_api.rs
@@ -1,4 +1,5 @@
 //! The testing API client to ykshim.
+#![cfg(feature = "testing")]
 
 use crate::{CompiledTrace, Local, RawCompiledTrace, RawSirTrace, RawTirTrace, SirTrace, TyIndex};
 use libc::size_t;
--- a/internal_ws/yksg/Cargo.toml
+++ b/internal_ws/yksg/Cargo.toml
@@ -8,3 +8,6 @@ edition = "2018"
 libc = "0.2.82"
 yktrace = { path = "../yktrace" }
 ykpack = { path = "../ykpack" }
+
+[features]
+testing = []
--- a/internal_ws/yksg/src/lib.rs
+++ b/internal_ws/yksg/src/lib.rs
@@ -13,7 +13,7 @@ use ykpack::{
     self, BinOp, Body, CallOperand, Constant, ConstantInt, IRPlace, Local, Statement, Terminator,
     TyKind, UnsignedInt, UnsignedIntTy,
 };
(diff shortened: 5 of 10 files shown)


internal_ws/ykshim/src/test_api.rs
```diff
@@ -2,135 +2,146 @@
...
-/// Consumes and compiles the given TIR trace to native code, returning an opaque pointer to the
-/// compiled trace.
-#[no_mangle]
-fn __ykshimtest_compile_tir_trace(tir_trace: *mut TirTrace) -> *mut CompiledTrace {
-    let tir_trace = unsafe { Box::from_raw(tir_trace) };
-    let compiled_trace = ykcompile::compile_trace(*tir_trace);
-    Box::into_raw(Box::new(compiled_trace))
+// To avoid marking each individual import/struct/function/etc. as "#[cfg(feature="testing")]", we
+// stuff everything inside this conditionally compiled "testing" module.
+#[cfg(feature = "testing")]
```
bjorn3, line 7: Why not use `#![cfg(feature = "testing")]` on at the top of this file?

vext01, line 7: why don't we do this annotation one level up in `lib.rs`?

```
#[cfg(testing)]
mod test_api;
```

ykshim_client/src/test_api.rs
```diff
@@ -1,162 +1,169 @@
...
-pub struct CguHash(u64);
-#[derive(Debug, Copy, Clone)]
-#[repr(C)]
-pub struct TypeId {
-    pub cgu: CguHash,
-    pub idx: TyIndex,
-}
+// To avoid marking each individual import/struct/function/etc. as "#[cfg(feature="testing")]", we
+// stuff everything inside this conditionally compiled "testing" module.
+#[cfg(feature = "testing")]
```
bjorn3, line 5: Same here.

vext01:
This looks OK to me, but I don't think the large indent changes (putting things in a nested module) are necessary. I proposed one way to fix this, and @bjorn3 another. I'm not sure which is best.

internal_ws/ykshim/src/test_api.rs
ltratt: I might not have understood correctly, but would doing so be exactly equivalent to "don't build the contents of the file unless this feature is turned on"?

ltratt: AFAICS `cargo` still builds files even if they're not `mod`ed, so you end up with compilation errors.

bjorn3: yes

bjorn3: No, it doesn't. There were proposals to make `mod` implicit in the 2018 edition, but those were not accepted.

ltratt:
https://github.com/softdevteam/yk/pull/278/commits/eef7c7a5ca9269f1ccfba2154ab3ae014d4120e6 uses @bjorn3's fix -- it seems to work fine! Any other thoughts?

internal_ws/ykshim/Cargo.toml
```diff
@@ -14,3 +14,6 @@ yksg = { path = "../yksg" }
 ykcompile = { path = "../ykcompile" }
 ykpack = { path = "../ykpack" }
 yktrace = { path = "../yktrace" }
+
+[features]
+testing = ["yksg/testing"]
```
vext01, line 19: What's the significance of "yksg/testing" here?

ltratt: It means "when the testing flag for this crate is on, also build yksg with the testing flag."

vext01: I see. I hadn't seen this before. The `/` makes it look like a filesystem path, but it's not.

ltratt: I'm not even sure if it's properly documented (from memory it's mentioned, but in a confusing manner), but I saw it somewhere, guessed what it did, and verified that it seems to match my guess.

vext01:
LGTM. If all of the existing xtask functionality still works, I think we can put this in.

ltratt:
OK to squash?

ltratt:
@vext01 ping.

vext01:
Sorry, please squash.

ltratt:
Squashed.

vext01:
Just rustfmt.

ltratt:
Gah, I was caught out by own attempts at rustfmt sanitising! Fixed in https://github.com/softdevteam/yk/pull/278/commits/31b8407e3239f1e0da78c47c3230bc4da791c6ce.

vext01:
Please squash.

ltratt:
Squashed.

PR #280, opened 2021-03-03
Consistently spell 'dest'/'dst' as 'dst'.

ltratt:
Previously we had both 'dest' and 'dst' as spellings, but we only ever use 'src' (not 'sorc' or 'surc' or similar). 'dst' seems much more symmetric to 'src', so I've consistently preferred that over 'dest'.


vext01:
fine by me.

bors r+

ltratt:
bors r-

We should wait for @ptersilie to comment.

ptersilie:
Looks good.

bors r+

PR #281, opened 2021-03-03
Add tests for binary operations with constant arguments.

vext01:
Shakes out a couple of bugs:
 - Something wrong with loading 64-bit constants into registers.
 - Forgot to zero unused argument space for binops on constant bytes.

[We now have 287 binop tests!]

### DIFF
--- a/internal_ws/ykcompile/src/arch/x86_64/mod.rs
+++ b/internal_ws/ykcompile/src/arch/x86_64/mod.rs
@@ -127,7 +127,7 @@ macro_rules! binop_mul_div {
     //
     //   Encoding       Semantics
     //   --------------------------------------------------------------------
-    //   MUL r/m8:      AX * r/m8          AX <- result
+    //   MUL r/m8:      AL * r/m8          AX <- result
     //   MUL r/m16:     AX * r/m16         DX:AX <- result
     //   MUL r/m32:     EAX * r/m32        EDX:EAX <- result
     //   MUL r/m64:     RAX * r/m64        RDX:RAX <- result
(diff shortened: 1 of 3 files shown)


internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -237,7 +237,12 @@ macro_rules! binop_mul_div {
                 },
                 Location::Mem(..) => todo!(),
                 Location::Const { val, .. } => {
-                    if size > 1 {
+                    // Set unused extended operand space to zero.
+                    if size == 1 {
+                        dynasm!(self.asm
+                            ; and ax, 0xff
```
ptersilie, line 243: How is this setting `ax` to zero? Doesn't this just leave `ax` as is?

internal_ws/ykcompile/src/arch/x86_64/store.rs
```diff
@@ -100,21 +100,9 @@ impl TraceCompiler {
...
-                    let i64_c = c_val.i64_cast();
-                    let hi_word = (i64_c >> 32) as i32;
-                    let lo_word = (i64_c & 0xffffffff) as i32;
-                    dynasm!(self.asm
-                        ; mov Rq(dst_reg), hi_word
-                        ; shl Rq(dst_reg), 32
-                        ; or Rq(dst_reg), lo_word
-                    );
-                }
+                dynasm!(self.asm
```
ptersilie, line 103: Huh, I wonder why we thought this wasn't possible and went through so much trouble.

vext01: I think we didn't realise that we had to `QWORD` the second operand.

I stared at the Intel manual and the dynasm reference and concluded that it *must* be possible somehow.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
vext01: I was about to say:

> It sets the top byte of AX to zero. In the `r/m8` variants, the first operand is all of AX, but we only care for the lower byte.

But, that's only true of `DIV` I now realise...

image

image

Let me see if I can find a way to only do this for DIV without a string comparison at runtime.

vext01: Of course, the obvious fix causes test failures...

vext01: Fixed in 549d394cd.

vext01:
Ready to squash?

ptersilie:
Looks good, even though you didn't add the zeroes as promised. :wink: Please squash.

vext01:
Splat. I added the zeros in doing so. Sorry I forgot!

PR #283, opened 2021-03-04
Implement trace stitching.

ptersilie:
Currently, we return after each trace execution back to the meta-tracer,
only for it to run the same trace again. Instead, we can just stitch
together the compiled trace, so that it loops indefinitely until a guard
fails.

The API for this very basic. The `interp_step` function is now required
to return a bool. If a traced path through the function returns `true`,
this means the interpreter is finished and we return; if a path returns
`false`, this means trace stitching is enabled for that trace and it can
only be exited via a guard.

Typically this means interpreter authors want to do something similar to
this:

```rust
fn interp_step(io: &mut Ctx) -> bool {
    // do interpreter stuff
    if some_abort_condition {
        return true;
    }
    false // loop forever
}
```

### DIFF
--- a/internal_ws/ykcompile/src/arch/x86_64/mod.rs
+++ b/internal_ws/ykcompile/src/arch/x86_64/mod.rs
@@ -1286,7 +1286,11 @@ impl TraceCompiler {
     }
 
     /// Emit a return instruction.
-    fn ret(&mut self, gl: Vec<(&Guard, HashMap<&Local, Location>, DynamicLabel)>) {
+    fn ret(
+        &mut self,
+        gl: Vec<(&Guard, HashMap<&Local, Location>, DynamicLabel)>,
+        stitch_trace: bool,
+    ) {
         // Reset the stack/base pointers and return from the trace. We also need to generate the
         // code that reserves stack space for spilled locals here, since we don't know at the
         // beginning of the trace how many locals are going to be spilled.
@@ -1310,11 +1314,18 @@ impl TraceCompiler {
         let topalign = SYSV_CALL_STACK_ALIGN
             - (live_off + soff as usize + CALLEE_SAVED_REGS.len() * QWORD_REG_SIZE)
                 % SYSV_CALL_STACK_ALIGN;
-        // Restore registers.
+
+        if stitch_trace {
+            // Loop trace indefinitely until a guard fails.
+            dynasm!(self.asm
+                ; jmp ->main
+            );
+        }
         dynasm!(self.asm
             ; mov rax, 0 // Signifies that there were no guard failures.
             ; ->cleanup:
         );
+        // Restore registers.
         self.restore_regs(&*CALLEE_SAVED_REGS);
         // Restore the stack and return.
         dynasm!(self.asm
@@ -1483,7 +1494,7 @@ impl TraceCompiler {
                 Err(e) => tc.crash_dump(Some(e)),
             }
         }
-        tc.ret(gl);
+        tc.ret(gl, tt.stitch);
         let buf = tc.asm.finalize().unwrap();
         if debug {
             // In debug mode the memory section which contains the compiled trace is marked as
--- a/internal_ws/yksg/src/lib.rs
+++ b/internal_ws/yksg/src/lib.rs
@@ -150,6 +150,7 @@ macro_rules! make_binop {
             let (v, of) = match op {
                 BinOp::Add => a.overflowing_add(b),
                 BinOp::Lt => ($type::from(a < b), false),
+                BinOp::Gt => ($type::from(a > b), false),
                 _ => todo!(),
             };
             if checked {
@@ -191,6 +192,7 @@ struct StackFrame {
 pub struct StopgapInterpreter {
     /// Active stack frames (most recent last).
     frames: Vec<StackFrame>,
+    returnval: bool,
 }
 
 impl StopgapInterpreter {
@@ -199,6 +201,7 @@ impl StopgapInterpreter {
         let frame = StopgapInterpreter::create_frame(&sym);
         StopgapInterpreter {
             frames: vec![frame],
+            returnval: true,
         }
     }
 
@@ -220,7 +223,10 @@ impl StopgapInterpreter {
             };
             frames.push(frame);
         }
-        let mut sg = StopgapInterpreter { frames };
+        let mut sg = StopgapInterpreter {
+            frames,
+            returnval: true,
+        };
         let frame = sg.frames.last().unwrap();
         // Since we start in the block where the guard failed, we immediately skip to the
         // terminator and interpret it to initialise the block where actual interpretation needs to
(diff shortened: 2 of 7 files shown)


ltratt:
I'll do a full review of this later, but gut reaction: we need several more tests. e.g. what happens if we have multiple `return false`/`return true` in an `interp_step`? I wouldn't be surprised if we can start sharing code between tests to make it easier to write more of them.

ptersilie:
It doesn't really matter how many `return true/false` you have in your `interp_step` function, as you'll only ever see one of them inside the trace. But I'm happy to add more tests anyway.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -1286,7 +1286,11 @@ impl TraceCompiler {
     }
 
     /// Emit a return instruction.
-    fn ret(&mut self, gl: Vec<(&Guard, HashMap<&Local, Location>, DynamicLabel)>) {
+    fn ret(
+        &mut self,
+        gl: Vec<(&Guard, HashMap<&Local, Location>, DynamicLabel)>,
+        stitch_trace: bool,
```
ltratt, line 1292: This function is already very overloaded AFAICS in that it does more than just "emit a return instruction". I'm nervous about bloating it further with `stitch_trace` especially as I'm not sure why it's a bool -- can you clarify why it might be true/false?

internal_ws/yksg/src/lib.rs
```diff
@@ -191,6 +192,7 @@ struct StackFrame {
 pub struct StopgapInterpreter {
     /// Active stack frames (most recent last).
     frames: Vec<StackFrame>,
+    returnval: bool,
```
ltratt, line 195: I think we need to document what this means.

internal_ws/yksg/src/lib.rs
```diff
@@ -345,6 +352,11 @@ impl StopgapInterpreter {
                     let size = usize::try_from(SIR.ty(&dst.ty()).size()).unwrap();
                     std::ptr::copy(ret_ptr, dst_ptr, size);
                     curframe.bbidx = bbidx;
+                } else {
+                    // The return value of `interp_step` tells the meta-tracer whether to stop or
```
ltratt, line 356: This string (or some variant of it) might better off be on the attribute in the `struct`?

ltratt:
> The API for this very basic. The interp_step function is now required
to return a bool. If a traced path through the function returns true,
this means the interpreter is finished and we return; if a path returns
false, this means trace stitching is enabled for that trace and it can
only be exited via a guard.

I'm not entirely sure if this will work in our situation, but I'm not quite sure. Imagine I have this program:

```rust
x = 0
for i in range(len10)):
  x += 1
print(x)
```

Let's assume I have a hot threshold of 5. I want to end up with a trace which (at a high level) looks like:

```
let tmp = next_int_in_range();
if tmp == 10 { return true; }
x += 1
return false;
```

What I'm not immediately sure is how naturally that maps onto a classical interpreter where the above program might have bytecode such as:

```
INT 0
SETVAR x
INT 1
SETVAR tmp
L1:
GETVAR tmp
JEQ 10 L1
GETVAR x
INT 1
ADD
SETVAR x
JMP L1
L2:
```

Where in the interpreter would I put the `return false`/`return true` that correspond to this? It seems to me that you can't know that unless you know which `Location` the tracing started at but, at the moment, I don't think that an interp_step can know that information.

Warning: I haven't fully thought this through and I might be very wrong.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
ptersilie: This is a restriction of the current API (where API is used very loosely), which we'll likely change in the future. The user needs to return `true/false` in `interp_step`, where `false` means loop forever. But currently nothing keeps them from only return true, e.g.:

```
fn interp_step() -> bool {
    // do interpreter stuff
    true
}
```

Without a condition that returns `false` in one case, we would loop forever with no way of exiting the trace. So in such a case we do not want to stitch the trace. Although, similarly if the function only returns `false` we would also loop forever, so maybe we do require some extra check. My initial plan was to do this when we turn this into a proper API in a future PR.

ptersilie:
I don't think I follow this example. What is this program? Is it something the end-user writes, or is this the interpreter loop?

vext01:
I think what @ltratt  means, if a guard fails, how do you specify if the trace exited with true or false?

Assuming that's what he means (sorry if not), a guard failure always implies `false`, i.e. the program isn't done, just we've diverged from the trace.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
ltratt: I thought we weren't going to offer any solution for "I don't want stitching", because we couldn't think of a useful use case where someone might reasonably want to do that?

vext01:
I must admit I'm a little confused by this wording:

> If a traced path through the function returns true,
this means the interpreter is finished and we return; if a path returns
false, this means trace stitching is enabled for that trace and it can
only be exited via a guard.

Let me re-word it and see if it's still right:

 * If `interp_step` returns false, then the control point continues, by running another interpreter step.
 * If `interp_step` returns true, then the control point returns.
 * If a guard fails, the trace doesn't make it as far as a return from the `interp_step`.

Is that right?

internal_ws/ykcompile/src/arch/x86_64/mod.rs
vext01: [We've discussed splitting `ret` a week or so ago. I think we need this more now than ever]

ptersilie:
A guard failure returns neither true nor false. It returns a pointer to the stopgap interpreter. The interpreter then interprets the guarded branch which tells us what we need to return:
```
fn inter_step() -> bool {
    if abort_condition {
        return true;
    }
    if some_other_condition {
        return false;
    }
    false // loop
}
```

If this trace is stitched the only way to leave it is via one of the guards. When then invoke the stopgap interpreter to interpret the guarded branch, e.g. for `abort_condition` we'd interpret `return true`, which tells us if we are done. If the `some_other_condition` guard failed, it would return `false` telling us to keep running the interpreter.

internal_ws/yksg/src/lib.rs
vext01: s/returnval/return_val/ ?

Or just shorten it to `rv`?

ptersilie:
> If a guard fails, the trace doesn't make it as far as a return from the interp_step.

Note, that `control_point` will always run the `interp_step` to the end, either via `exec_trace`, `step_fn`, or `sg.interpret`. So we'll always end up with a `bool` telling us to continue/stop.

vext01:
> A guard failure returns neither true nor false. It returns a pointer to the stopgap interpreter.

Ah. I just remembered a detail. A reader might say, "but I thought interp_step must return a bool?"

There's what interp_step returns at the rust level, and then there's what the trace code returns.

I wonder if we need to clarify the levels in our docstrings?

ptersilie:
Yes, `interp_step` returns a `bool`, but a compiled trace doesn't. It'll return a pointer to the stopgap interpreter. And if stitching is always enabled, the compiled trace won't even be able to return `null` anymore.

ptersilie:
After offline discussion it was decided to move the stitching into the TIR level (i.e. added labels and gotos), instead of adding the jump during compilation. We will be needed `Goto`s later on anyway to implement "hoisting" and stitching of different traces (e.g. when a guard failure always moves to another compiled trace).

PR #285, opened 2021-03-04
Try to clarify the documentation around the stopgap FramesVec.

ltratt:
I'm not 100% sure my updates are correct. I also think we leak the memory in FrameInfo::mem -- we `alloc` memory but I can't see a corresponding `dealloc`. Indeed, I think we'd need to pass size/align around for that to work but that's rather tricky at the moment as with the x86_64 SysV ABI we only have 1 "easy" paramater register left...

### DIFF
--- a/internal_ws/ykcompile/src/arch/x86_64/mod.rs
+++ b/internal_ws/ykcompile/src/arch/x86_64/mod.rs
@@ -333,9 +333,9 @@ fn local_to_reg_name(loc: &Location) -> &'static str {
 // Collection of functions required during a guard failure to instantiate and initialise the
 // stopgap interpreter.
 
-/// Given a vector of `FrameInfo`s `vptr`, instantiates and initialises a SIR interpreter and returns its
-/// pointer. Consumes `vptr`.
-extern "sysv64" fn invoke_sinterp(vptr: *mut Vec<FrameInfo>) -> *mut StopgapInterpreter {
+/// Given a pointer to a vector of `FrameInfo`s, creates and returns a boxed StopgapInterpreter.
+/// Consumes `vptr`.
+extern "sysv64" fn new_stopgap(vptr: *mut Vec<FrameInfo>) -> *mut StopgapInterpreter {
     let v = unsafe { Box::from_raw(vptr) };
     let si = StopgapInterpreter::from_frames(*v);
     Box::into_raw(Box::new(si))
@@ -1338,8 +1338,8 @@ impl TraceCompiler {
         // Add guard failure labels. When a guard fails we jump to one of these labels. For each
         // stack frame, we need to allocate memory using `allocate_layout`. We then copy the live
         // variables into this memory and store it, together with other information, inside a
-        // Vec<FrameInfo>. Finally, we call `invoke_sinterp` passing in this vector, which
-        // initialises a SIR interpreter and runs it.
+        // Vec<FrameInfo>. Finally, we call `new_stopgap` passing in this vector, which
+        // returns a StopgapInterpreter that we can then run.
         for (guard, mut live_locations, label) in gl {
             // Symbol names of the functions called while executing the trace. Needed to recreate
             // the stack frames in the StopgapInterpreter.
@@ -1437,7 +1437,7 @@ impl TraceCompiler {
                 // Create and initialise SIR interpreter, then return it.
                 ; mov rdi, Rq(frame_vec_reg)
                 ; xor rax, rax
-                ; mov r11, QWORD invoke_sinterp as i64
+                ; mov r11, QWORD new_stopgap as i64
                 ; call r11
                 ; jmp ->cleanup
             );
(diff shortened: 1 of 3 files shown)


ptersilie:
I don't think it's a leak. `FrameInfo::mem` is passed into `LocalMem` (https://github.com/softdevteam/yk/blob/master/internal_ws/yksg/src/lib.rs#L211), which implements Drop where it deallocates the memory (https://github.com/softdevteam/yk/blob/master/internal_ws/yksg/src/lib.rs#L42).

ltratt:
> FrameInfo::mem is passed into LocalMem (https://github.com/softdevteam/yk/blob/master/internal_ws/yksg/src/lib.rs#L211), which implements Drop where it deallocates the memory (https://github.com/softdevteam/yk/blob/master/internal_ws/yksg/src/lib.rs#L42).

Ah. OK, dumb question: given that a trace will inline functions, how is it the case that the live variables at a guard failure seem to have a 1:1 mapping to a MIR function's locals?

ptersilie:
Do you have an example? They shouldn't, but of course some could happen to have the same local in both IRs.

ltratt:
Something like:

```rust
fn interp_step(...) {
   let x: u64 = ...;
   f();
}

fn f() {
   if <before tracing> { ... } else { ... }
}
```

and let's assume that we trace the "true" branch of the `if` and after a guard failure we have to go to the stopgap interpreter. At that point the inlined trace has one live variable `x` but the function frame for `f` has no local variables at all.

ltratt:
I now see that we do free the memory so I've fixed the comment in https://github.com/softdevteam/yk/pull/285/commits/c5acb8cf9438e538485c96a8b975740968c85c8b.

ltratt:
@ptersilie See what you think of those last two commits.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -333,9 +333,9 @@ fn local_to_reg_name(loc: &Location) -> &'static str {
 // Collection of functions required during a guard failure to instantiate and initialise the
 // stopgap interpreter.
 
-/// Given a vector of `FrameInfo`s `vptr`, instantiates and initialises a SIR interpreter and returns its
-/// pointer. Consumes `vptr`.
-extern "sysv64" fn invoke_sinterp(vptr: *mut Vec<FrameInfo>) -> *mut StopgapInterpreter {
+/// Given a pointer to a vector of `FrameInfo`s, creates and returns a boxed SIR interpreter.
```
ptersilie, line 336: s/SIR interpreter/Stopgap interpreter

ltratt: Fixed in f1ec182ad469bce8ec189a2c012f6ba8c5be9ca2.

ltratt:
I've also fixed the "block" attribute in https://github.com/softdevteam/yk/pull/285/commits/1634f506bd1e2c8439b2d75b2aa0749f0b76fb47. I think that's enough stuff for one PR!

ptersilie:
Looking good. Please squash.

ltratt:
Squashed.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -1378,28 +1376,26 @@ impl TraceCompiler {
...
 
-            // Create a vector of `FrameInfo`s in which we store the allocated memory pointers,
-            // body, and basic block index for each active stack frame during the guard failure.
-            // This vector is later handed to the StopgapInterpreter, in order to initialise the
-            // stack frames needed to interpret SIR. We can use a fixed register here, since we've
-            // spilled all registers to the stack.
+            // The trace inlines several functions, so we now have to split the "sort of single
+            // frame" trace into a multi-frame vector so that the StopgapInterpreter can execute
+            // normal SIR. Fortunately all the variables we care about are on the stack at this
+            // point which simplifies this process.
```
ptersilie, line 1382: If I was being super nitpicky, I would like to specify why this simplifies the process, i.e. because we can use registers freely without worrying about clobbering stuff. But I'm not going to insist on it.

ltratt: Maybe deleting "which simplifies this process" is the easy fix?

ltratt: I've done a slightly bigger fix than I first intended, but I hope it addresses this -- and more fully than I first intended! https://github.com/softdevteam/yk/pull/285/commits/bb3a1fda28b0acd72d7150c4fe53f1a3717c5dec

ptersilie:
This looks good. Please squash.

ltratt:
Squashed.

ptersilie:
Looks like this needs a `fmt`.

ltratt:
Hmph, my local `cargo fmt` is giving subtly different formatting than the yk fmt! Fixed in https://github.com/softdevteam/yk/pull/285/commits/e2edefc3bc790222c76750d3c73c7f2c98467a45. OK to squash?

ptersilie:
Please squash.

ltratt:
Squashed.

PR #289, opened 2021-03-15
Trace looping

ptersilie:
This is essentially #283, but the stitching has been moved to the TIR level. I've also added some additional tests.

The stitching API still uses the bool method, but we''ll be changing this to an enum very soon, once we have implemented enums in both compiler and stopgap interpreter.

You've already reviewed a55e313 in the previous PR, so it can be skipped. Obviously needs squashing later.

Companion PR: https://github.com/softdevteam/ykrustc/pull/177

### DIFF
--- a/internal_ws/ykcompile/src/arch/x86_64/mod.rs
+++ b/internal_ws/ykcompile/src/arch/x86_64/mod.rs
@@ -994,6 +994,16 @@ impl TraceCompiler {
             Statement::StorageDead(l) => self.local_dead(l)?,
             Statement::Call(target, args, dst) => self.c_call(target, args, dst)?,
             Statement::Cast(dst, src) => self.c_cast(dst, src),
+            Statement::LoopStart => {
+                dynasm!(self.asm
+                    ; ->loop_start:
+                );
+            }
+            Statement::LoopEnd => {
+                dynasm!(self.asm
+                    ; jmp ->loop_start
+                );
+            }
             Statement::Nop | Statement::Debug(..) => {}
             Statement::Unimplemented(s) => todo!("{:?}", s),
         }
@@ -1329,6 +1339,7 @@ impl TraceCompiler {
         let topalign = SYSV_CALL_STACK_ALIGN
             - (live_off + soff as usize + CALLEE_SAVED_REGS.len() * QWORD_REG_SIZE)
                 % SYSV_CALL_STACK_ALIGN;
+
         dynasm!(self.asm
             ; mov rax, 0 // Signifies that there were no guard failures.
             ; ->cleanup:
--- a/internal_ws/ykpack/src/types.rs
+++ b/internal_ws/ykpack/src/types.rs
@@ -492,6 +492,10 @@ pub enum Statement {
     /// Cast a value into another. Since the cast type and the destination type are the same, we
     /// only need the latter.
     Cast(IRPlace, IRPlace),
+    /// Marks the spot where we want the trace to loop.
+    LoopStart,
+    /// Jumps to the position marked by `LoopStart`.
+    LoopEnd,
     /// A debug marker. This does not appear in SIR.
     Debug(String),
     /// Any unimplemented lowering maps to this variant.
@@ -538,6 +542,8 @@ impl Display for Statement {
                 write!(f, "{} = call({}, [{}])", dst_s, op, args_s)
             }
             Statement::Cast(d, s) => write!(f, "Cast({}, {})", d, s),
+            Statement::LoopStart => write!(f, "LoopStart"),
+            Statement::LoopEnd => write!(f, "LoopEnd"),
             Statement::Debug(s) => write!(f, "// {}", s),
             Statement::Unimplemented(mir_stmt) => write!(f, "unimplemented_stmt: {}", mir_stmt),
         }
--- a/internal_ws/yksg/src/lib.rs
+++ b/internal_ws/yksg/src/lib.rs
@@ -150,6 +150,7 @@ macro_rules! make_binop {
             let (v, of) = match op {
                 BinOp::Add => a.overflowing_add(b),
                 BinOp::Lt => ($type::from(a < b), false),
+                BinOp::Gt => ($type::from(a > b), false),
                 _ => todo!(),
             };
             if checked {
@@ -191,6 +192,8 @@ struct StackFrame {
 pub struct StopgapInterpreter {
     /// Active stack frames (most recent last).
     frames: Vec<StackFrame>,
+    /// Value to be returned by the interpreter.
+    rv: bool,
 }
 
 impl StopgapInterpreter {
@@ -199,6 +202,7 @@ impl StopgapInterpreter {
         let frame = StopgapInterpreter::create_frame(&sym);
         StopgapInterpreter {
             frames: vec![frame],
+            rv: true,
         }
     }
 
@@ -220,7 +224,7 @@ impl StopgapInterpreter {
             };
             frames.push(frame);
         }
-        let mut sg = StopgapInterpreter { frames };
+        let mut sg = StopgapInterpreter { frames, rv: true };
         let frame = sg.frames.last().unwrap();
         // Since we start in the block where the guard failed, we immediately skip to the
         // terminator and interpret it to initialise the block where actual interpretation needs to
@@ -276,7 +280,7 @@ impl StopgapInterpreter {
         std::ptr::write::<*mut u8>(ptr as *mut *mut u8, ctx);
     }
 
-    pub unsafe fn interpret(&mut self) {
+    pub unsafe fn interpret(&mut self) -> bool {
         while let Some(frame) = self.frames.last() {
             let body = frame.body.clone();
             let block = &body.blocks[usize::try_from(frame.bbidx).unwrap()];
@@ -296,11 +300,14 @@ impl StopgapInterpreter {
                     Statement::Unimplemented(_) | Statement::Debug(_) => todo!(),
                     Statement::Cast(..) => todo!(),
                     Statement::StorageLive(_) | Statement::StorageDead(_) => {}
-                    Statement::Call(..) => unreachable!(),
+                    Statement::Call(..) | Statement::LoopStart | Statement::LoopEnd => {
+                        unreachable!()
+                    }
                 }
             }
             self.terminator(&block.term);
         }
+        self.rv
     }
 
     unsafe fn terminator(&mut self, term: &Terminator) {
@@ -345,6 +352,11 @@ impl StopgapInterpreter {
                     let size = usize::try_from(SIR.ty(&dst.ty()).size()).unwrap();
                     std::ptr::copy(ret_ptr, dst_ptr, size);
                     curframe.bbidx = bbidx;
+                } else {
+                    // The return value of `interp_step` tells the meta-tracer whether to stop or
+                    // continue running the interpreter.
+                    let ret_ptr = oldframe.mem.local_ptr(&RETURN_LOCAL);
+                    self.rv = std::ptr::read::<bool>(ret_ptr as *const bool);
                 }
             }
             Terminator::SwitchInt {
--- a/internal_ws/yktrace/src/tir.rs
+++ b/internal_ws/yktrace/src/tir.rs
@@ -51,6 +51,7 @@ impl<'a, 'm> TirTrace<'a, 'm> {
         let mut live_locals: Vec<HashSet<Local>> = Vec::new();
         let mut guard_blocks: Vec<GuardBlock> = Vec::new();
 
+        let mut loop_trace = false;
         let mut in_interp_step = false;
         let mut entered_call = false;
         while let Some(loc) = itr.next() {
@@ -153,10 +154,33 @@ impl<'a, 'm> TirTrace<'a, 'm> {
                             idx: rnm.rename_irplace(idx, &body),
                             scale: *scale,
                         },
-                        Statement::Store(dst, src) => Statement::Store(
-                            rnm.rename_irplace(dst, &body),
-                            rnm.rename_irplace(src, &body),
-                        ),
+                        Statement::Store(dst, src) => {
+                            if matches!(dst, IRPlace::Val { local, .. } if local == &sir::RETURN_LOCAL)
+                            {
+                                if body.flags.contains(BodyFlags::INTERP_STEP) {
+                                    match src {
+                                        IRPlace::Const {
+                                            val: Constant::Bool(b),
+                                            ..
+                                        } => {
+                                            // If the `interp_step` function returns false, we
+                                            // enabled trace looping which loops the trace
+                                            // indefinitely until a guard fails.
+                                            if *b == false {
+                                                loop_trace = true;
+                                            }
+                                        }
+                                        _ => {
+                                            panic!("Return value of `interp_step` function must be a constant bool.")
+                                        }
+                                    }
+                                }
+                            }
+                            Statement::Store(
+                                rnm.rename_irplace(dst, &body),
+                                rnm.rename_irplace(src, &body),
+                            )
+                        }
                         Statement::BinaryOp {
                             dst,
                             op,
@@ -187,7 +211,9 @@ impl<'a, 'm> TirTrace<'a, 'm> {
                             Statement::StorageDead(l)
                         }
                         // The following statements are specific to TIR and cannot appear in SIR.
-                        Statement::Call(..) => unreachable!(),
+                        Statement::Call(..) | Statement::LoopStart | Statement::LoopEnd => {
+                            unreachable!()
+                        }
                     };
 
                     // In TIR, stores to local number zero are always to the return value of the
@@ -418,6 +444,11 @@ impl<'a, 'm> TirTrace<'a, 'm> {
         }
 
         let local_decls = rnm.done();
+
+        if loop_trace {
+            ops.insert(0, TirOp::Statement(Statement::LoopStart));
+            ops.push(TirOp::Statement(Statement::LoopEnd));
+        }
 
         Ok(Self {
             ops,
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -180,8 +180,9 @@ impl MTThread {
         loc: Option<&Location<I>>,
         step_fn: S,
         ctx: &mut I,
-    ) where
-        S: Fn(&mut I),
+    ) -> bool
+    where
+        S: Fn(&mut I) -> bool,
     {
         // If a loop can start at this position then update the location and potentially start/stop
         // this thread's tracer.
@@ -193,13 +194,13 @@ impl MTThread {
                 let ptr = self.exec_trace(func, ctx);
 
                 if ptr.is_null() {
-                    // Trace succesfully executed.
-                    return;
+                    // Trace finished executing, which means the user did not specify that
+                    // `interp_step` should run in a loop, by adding a `return false`.
+                    return true;
                 } else {
                     unsafe {
                         let mut si = StopgapInterpreter(ptr);
-                        si.interpret();
-                        return;
+                        return si.interpret();
                     }
                 }
             }
@@ -483,7 +484,9 @@ mod tests {
     struct EmptyInterpCtx {}
 
     #[interp_step]
-    fn empty_step(_: &mut EmptyInterpCtx) {}
+    fn empty_step(_: &mut EmptyInterpCtx) -> bool {
+        true
+    }
 
     #[test]
     fn threshold_passed() {
@@ -606,7 +609,7 @@ mod tests {
         }
 
         #[interp_step]
-        fn simple_interp_step(ctx: &mut InterpCtx) {
+        fn simple_interp_step(ctx: &mut InterpCtx) -> bool {
             match ctx.prog[ctx.pc] {
                 INC => {
                     ctx.pc += 1;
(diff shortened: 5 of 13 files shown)


ykrt/src/mt.rs
```diff
@@ -949,4 +957,41 @@ mod tests {
...
+
+        #[interp_step]
+        fn simple_interp_step(ctx: &mut InterpCtx) -> bool {
+            let a = ctx.counter;
+            if ctx.counter < 100 {
+                ctx.counter = a + 1;
+            } else {
+                ctx.counter = a + 2;
+            }
+            if ctx.counter > 110 {
```
ltratt, line 980: This can just be `ctx.counter > 110` without the `if`/`return` dance.

ykrt/src/mt.rs
```diff
@@ -949,4 +957,41 @@ mod tests {
...
+            false
+        }
+
+        let mut ctx = InterpCtx { counter: 0 };
+        loop {
+            if mtt.control_point(Some(&loc), simple_interp_step, &mut ctx) {
+                break;
+            }
+            // Slow down the interpreter so we get a chance to compile stuff.
+            thread::sleep(time::Duration::from_millis(10));
```
ltratt, line 992: Sleeping isn't going to work in general (all we need is a faster computer...). The only way to do this is to copy other tests which check what the state of the `Location` is.

ykrt/src/mt.rs
ptersilie: I'll have to test first that this still generates `return false` in SIR, which then triggers the rewrite to `LoopEnd` in TIR.

ykrt/src/mt.rs
ptersilie: Okay, will fix.

internal_ws/ykcompile/src/arch/x86_64/mod.rs
```diff
@@ -1329,6 +1339,7 @@ impl TraceCompiler {
         let topalign = SYSV_CALL_STACK_ALIGN
             - (live_off + soff as usize + CALLEE_SAVED_REGS.len() * QWORD_REG_SIZE)
                 % SYSV_CALL_STACK_ALIGN;
+
```
vext01, line 1342: Did we mean to add this blank line?

internal_ws/ykpack/src/types.rs
```diff
@@ -492,6 +492,10 @@ pub enum Statement {
     /// Cast a value into another. Since the cast type and the destination type are the same, we
     /// only need the latter.
     Cast(IRPlace, IRPlace),
+    /// Marks the spot where we want trace stitching to start.
```
vext01, line 495: Have we defined "stitching" anywhere? Should we just say explicitly what this does instead?

Something like:
> The position we jump to when we encounter a `LoopEnd` bytecode. There must be only one `LoopStart` bytecode per trace.

Similarly for `LoopEnd`.

internal_ws/ykpack/src/types.rs
```diff
@@ -492,6 +492,10 @@ pub enum Statement {
...
+    LoopStart,
+    /// Jumps to the trace stitching marker.
+    LoopEnd,
```
vext01, line 498: Is there any scenario where we can have >1 `LoopEnd` I wonder?

internal_ws/ykpack/src/types.rs
ltratt: I'm actually not sure I even like the term "stitching", which to me suggests putting multiple things together, while this is about adding a loop.

internal_ws/yksg/src/lib.rs
```diff
@@ -191,6 +192,7 @@ struct StackFrame {
 pub struct StopgapInterpreter {
     /// Active stack frames (most recent last).
     frames: Vec<StackFrame>,
+    returnval: bool,
```
vext01, line 195: Did we agree on `rv`for a field elsewhere? I can't remember, but seems to ring a bell.

internal_ws/yktrace/src/tir.rs
```diff
@@ -153,10 +154,23 @@ impl<'a, 'm> TirTrace<'a, 'm> {
                             idx: rnm.rename_irplace(idx, &body),
                             scale: *scale,
                         },
-                        Statement::Store(dst, src) => Statement::Store(
-                            rnm.rename_irplace(dst, &body),
-                            rnm.rename_irplace(src, &body),
-                        ),
+                        Statement::Store(dst, src) => {
+                            if matches!(dst, IRPlace::Val { local, .. } if local == &sir::RETURN_LOCAL)
```
vext01, line 158: I think you can reuse the `in_interp_step` variable instead of querying the body again?

Does the return value have to be constant? If so, does the compiler check that?

internal_ws/yktrace/src/tir.rs
```diff
@@ -153,10 +154,23 @@ impl<'a, 'm> TirTrace<'a, 'm> {
...
+                            {
+                                if matches!(src, IRPlace::Const { val: Constant::Bool(false), ..}) {
+                                    if body.flags.contains(BodyFlags::INTERP_STEP) {
+                                        // If the `interp_step` function returns false, we enabled
+                                        // trace stitching which loops the trace indefinitely until
+                                        // a guard fails.
+                                        stitch_trace = true;
+                                    }
+                                }
```
vext01, line 167: does there need to be an `else { unreachable! }` around here somewhere for the case that we are in the intep_step, but the type isn't what we expect?

ykrt/src/mt.rs
```diff
@@ -949,4 +957,41 @@ mod tests {
         mtt.control_point(loc, simple_interp_step, &mut ctx);
         assert_eq!(ctx.run, 2);
     }
+
+    #[test]
```
vext01, line 961: Is it worth adding doc comments to new tests, saying what they test at a glance? Unless it's super-obvious?

vext01:
This looks good.

Did you check that we got the performance increase we expect? Perhaps bump the loop count of one of your stitching tests and measure with and without stitching?

internal_ws/ykpack/src/types.rs
vext01: I could (in real life, using a needle) stitch a straight line of fabric with a loop half way up it's length, and that'd be quite a lot like out trace code. So the term doesn't bother me, but it should at the least be defined.

ptersilie: I was thinking the same as @ltratt that stitching implies to me stitching together multiple _different_ traces. I've just used it in lack of a better alternative.

internal_ws/ykpack/src/types.rs
ptersilie: I don't think so. That would imply that there are multiple returns in a single trace, which can't happen since by definition a trace only contains a single path.

internal_ws/yksg/src/lib.rs
ptersilie: I don't remember this.

internal_ws/yktrace/src/tir.rs
ptersilie: 1) I think `in_interp_step` means we tracing, it doesn't mean we are processing statements of the `interp_step` function, i.e. we could also be in another function that is being inlined. So we need to check the body again, since we only want to turn `return false` into `LoopEnd` if we are looking at the return of `interp_step` function.

internal_ws/yktrace/src/tir.rs
ptersilie: I guess we can. Though we already check that the return type is `bool` at compile time, so another check here isn't really needed.

ykrt/src/mt.rs
ptersilie: Yeah, I've just done that since I couldn't remember what I was testing here myself. :wink:

ykrt/src/mt.rs
ptersilie: This seems to work, although we'll soon replace the return type with an enum, in which case the if/else will come back. So we might as well leave it. Unless you feel strongly about it.

internal_ws/ykpack/src/types.rs
vext01: I never thought of it from that POV, but we can still consider other options.

How about "trace folding", as it implies a transform on a single entity?

internal_ws/yksg/src/lib.rs
vext01: OK. Well I prefer "rv" or "return_val", but don't feel all that strongly.

internal_ws/yktrace/src/tir.rs
vext01: > I think in_interp_step means we tracing, it doesn't mean we are processing statements of the interp_step function.

I just checked. I think you are right. It's true between the call and return (including callees) to `interp_step`.

vext01: About 2:

```
if matches!(src, IRPlace::Const { val: Constant::Bool(false), ..})
```

What if the boolean being returned is not a constant? Do we allow this?

```
fn interp_step(...) -> bool {
    let ret = false;
    ...
    if (somethig-dynamic) {
        ret = true;
    }
    ret;
}
```

internal_ws/yktrace/src/tir.rs
vext01: I think we should, as a sanity check. In theory no "unreachable!" should ever be needed, but we insert them to check our understanding of things.

internal_ws/ykpack/src/types.rs
ptersilie: I could just change the doc string to "Mark the spot where we want the trace to loop [on itself]".

internal_ws/yktrace/src/tir.rs
ptersilie: Fair doos. I hadn't thought about that. The question is: can we get the value out of the local at TIR time?

internal_ws/ykcompile/src/arch/x86_64/mod.rs
ptersilie: Yes I did. Otherwise it suggest this and the following line are somehow related, even though the following line doesn't even use `topalign`.

internal_ws/yktrace/src/tir.rs
ptersilie: After thinking about this a bit more: We could do some very basic static analysis to see what value the returned local has, but it's not a solution for the general case. As a very stupid example: the user could read the `bool` from a file or calculate it randomly in which case we'll never know. So maybe the right thing to do here is to enforce that they must return a constant.

ptersilie: Then we can resolve the comment below by causing a panic or similar which gives the user a hint that the return value must be a constant.

ykrt/src/mt.rs
ptersilie: Fixed in 7e160bbd13c050265ff1e5a8292f471e209dffd6.

ykrt/src/mt.rs
ptersilie: Fixed in 7e160bbd13c050265ff1e5a8292f471e209dffd6.

internal_ws/yksg/src/lib.rs
ptersilie: Fixed in 7e160bbd13c050265ff1e5a8292f471e209dffd6.

ptersilie:
> This looks good.
> 
> Did you check that we got the performance increase we expect? Perhaps bump the loop count of one of your stitching tests and measure with and without stitching?

I have not, since the stopgap interpreter doesn't have `usize` operations implemented yet. So I can't bump the iterations high enough to make a measurable difference. But the stitching is easy enough to disable that we can do those measurements at a later stage.

ykrt/src/mt.rs
ltratt: OK.

internal_ws/ykpack/src/types.rs
ltratt: > I could just change the doc string to "Mark the spot where we want the trace to loop [on itself]".

That seems better than "stitching" to me.

internal_ws/ykpack/src/types.rs
ltratt: > That would imply that there are multiple returns in a single trace, which can't happen since by definition a trace only contains a single path.

Once we have side traces, we will have a tree of traces, and then in a sense one does end up with multiple loop ends.

internal_ws/yktrace/src/tir.rs
ltratt: I tend to agree with @vext01 -- it means if/when we change another bit of code elsewhere we are more likely to have a nice warning that we've broken a long-forgotten invariant.

ltratt:
I tend to agree with Edd that we should get into the habit of doing some sort of performance measurements for these sorts of PRs, even if it's really Noddy-level measurements at the moment. But I'm not sure if this means:

>I have not, since the stopgap interpreter doesn't have usize operations implemented yet.

that trying to do that in this case will hit a `todo!` or not?

internal_ws/ykpack/src/types.rs
ptersilie: I guess from that perspective that's true, side-traces also reintroduce control-flow, making it possible to have multiple returns.

internal_ws/ykpack/src/types.rs
vext01: Works for me too.

internal_ws/ykpack/src/types.rs
vext01: At that point we can revise our bytecode names.

ptersilie: If my memory serves me well, then we've briefly discussed at some point, that we'll need to give `LoopStart` an id, so we can jump to it from another trace. But I think we decided to leave that for a future PR unless there are strong reasons why we should do it now.

ltratt: It's not just the names I'm worried about, it's whether they embed other assumptions that we might soon invalidate, and that it might be easiest to `todo!()` now.

internal_ws/yktrace/src/tir.rs
vext01: I agree that we should enforce a constant return value.

vext01:
Can you nest loops over smaller bounds?

internal_ws/yktrace/src/tir.rs
ltratt: I'm not sure I understand.

ltratt:
> Can you nest loops over smaller bounds?

I'm not sure I understand.

vext01:
Lukas is saying he can't do: `for i in 0..<some_usize>` due to limitations in the sg interp.

But we can probably do:
```
for i in 0..<some_u32> {
   for j in 0..<some_u32> {
       for k in 0..<some_u32> {
```

ltratt:
I'm not sure that we even need something that long / complicated? What I think we need is a rough idea of how long the second loop in https://github.com/softdevteam/yk/pull/289/commits/7e160bbd13c050265ff1e5a8292f471e209dffd6 takes before/after this PR. [i.e. once we've compiled a loop to machine code, how long does the machine code take to execute? This PR *should* be much faster, as we should execute the loop entirely in the JITed machine code, whereas before we'd keep dropping back to the interpreter.]

vext01:
Maybe I misunderstood, but it sounded like lukas was having trouble making the loop run long enough:

> So I can't bump the iterations high enough to make a measurable difference

ltratt:
Yes, but I *think* that's the wrong way of looking at it. I *think* my suggestion measures what we actually care about for this PR, but I am very willing to be proven wrong!

ptersilie:
I think I know what @ltratt means and I think I agree. I should be able to do a quick performance measure with that test.

internal_ws/ykpack/src/types.rs
ptersilie: I'm not sure what kind of `todo` you expect here. The only thing we can do here is give LoopStart/LoopEnd an id parameter so we can distinguish between different loops.

ltratt: Ah, I think I understand. OK let's not worry about this for the time being.

internal_ws/ykpack/src/types.rs
ptersilie: Fixed in 655129f.

internal_ws/yktrace/src/tir.rs
ptersilie: Fixed in d7f02a4.

ptersilie: The only problem with this fix is that it's only checked at JIT compile time. But I'm not sure how easy it will be to check this at rustc compile time.

internal_ws/yktrace/src/tir.rs
ptersilie: Fixed in d7f02a4.

ptersilie:
This is ready for re-view now (and there are still some unanswered comments as well).

I've also cobbled together a quick performance test which looks like this:
```rust
fn interp_step(io: &mut InterpCtx) {
    if io.0 < 100000000 {
        io.0 += 1;
        return false; // loop
    }
    true
}
```
Using stitching/looping we roughly get a 2.5x speed up (714ms vs 1.8s). This would likely be even more if I had done this on a test that uses the `MTThread` since we then have an additional call to `control_point` there. Although we did discuss at some point moving the loop inside the control point which remove that additional call again.

vext01:
> Using stitching/looping we roughly get a 2.5x speed up

Nice!

ykrt/src/mt.rs
```diff
@@ -978,18 +978,29 @@ mod tests {
                 ctx.counter = a + 2;
             }
             if ctx.counter > 110 {
-                return true; // Exit the interpreter.
+                return true; // Exit the interpreter;
```
vext01, line 981: Accidental change?

internal_ws/yktrace/src/tir.rs
vext01: You'd have to walk the MIR looking for returns. I think it's OK to check at runtime for now. Perhaps raise an issue?

ltratt:
> Return value of interp_step function must be a constant bool

I think this message is subtly wrong: a constant bool is `true` or `false`. I think we mean "The return value of an interp_step function must be of type `bool`."

vext01:
No, we literally require the user to return a `true` or `false` literal. We can't easily deal with a dynamically computed return value.

ltratt:
Oh OK! @vext01 I'm happy with this PR, so I'll leave you to shepherd it through until merging.

ltratt:
[A minor aside: a "literal" is subtly different than a "constant". I don't know which one of those we actually need.]

vext01:
I think literal is a syntactic term?

ykrt/src/mt.rs
ptersilie: Whoops. Fixed in 0b7857a.

ptersilie:
Think this is ready for squashing. I also need to update the commit message a bit.

internal_ws/yktrace/src/tir.rs
```diff
@@ -153,10 +154,30 @@ impl<'a, 'm> TirTrace<'a, 'm> {
...
+                                        IRPlace::Const { val: Constant::Bool(b), ..} => {
+                                            // If the `interp_step` function returns false, we
+                                            // enabled trace looping which loops the trace
+                                            // indefinitely until a guard fails.
+                                            if *b == false {
+                                                loop_trace = true;
+                                            }
+                                        }
+                                        _ => {
+                                            panic!("Return value of `interp_step` function must be constant.")
```
vext01, line 171: We still need to revise this comment. See [here](https://github.com/softdevteam/yk/pull/289#commitcomment-48317368).

ptersilie: Fixed in 388b3c3.

vext01:
Please squash.

ptersilie:
Squashed.

ptersilie:
Forgot to `fmt`.

vext01:
Please squash.

ptersilie:
Squashed.

vext01:
Wait, don't we have to merge the compiler PR first?

ptersilie:
Yeah, we do. Completely forgot about that PR!

ptersilie:
Also, why are there still format errors. I did `cargo xtask fmt`. Is that not enough?

vext01:
It should have been...

ptersilie:
Had to manually apply the format, since `cargo xtask fmt --all` wouldn't do it on its own. Weird.

vext01:
I think I know the source of that bug.

But anyway, please squash.

ptersilie:
Squashed. Though we have to merge https://github.com/softdevteam/ykrustc/pull/177 first.

vext01:
Is this ready to go now?

ptersilie:
Should be.

vext01:
I think the problem with this is that `.buildbot.sh` should patch the ykrustc dependency on ykpack. Without that it'll use the master branch of `yk` which contains a potentially stale `ykpack`.

This was my mistake. It should have been done in #290.

ltratt:
OK so do we need another PR to fix that or do we do it here or ... ?

vext01:
It would probably be best to apply the fix on top of this, since we have a scenario where the ykpack ABI has changed. Testing in a new PR wouldn't test the breakage case.

I'm testing a fix as we speak.

vext01:
[Then again, I probably can't push to Lukas' branch]

bjorn3:
You should be able to push to the branch. The checkbox for allowing this is enabled by default.

vext01:
It worked.

@ltratt if you are OK with this, please r+ it.

vext01:
OK it worked!

PR #290, opened 2021-03-15
Build ykrustc instead of using a binary snapshot.

vext01:
This means that we can detect changes in yk which would break the compiler build. A couple instances of that happening have arisen recently.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -8,21 +8,7 @@ case ${CI_TRACER_KIND} in
        exit 1;;
 esac
 
-export RUSTFLAGS="-C tracer=${CI_TRACER_KIND} -D warnings"
-
-# Use the most recent successful ykrustc build.
-tar jxf /opt/ykrustc-bin-snapshots/ykrustc-${CI_TRACER_KIND}-stage2-latest.tar.bz2
-export PATH=`pwd`/ykrustc-stage2-latest/bin:${PATH}
-
-# Test both workspaces.
-cargo xtask test
-cargo xtask bench
-cargo xtask clean
-
-# Also test the build without xtask, as that's what consumers will do.
-cargo build
-
-unset RUSTFLAGS
+# Install rustup.
 export CARGO_HOME="`pwd`/.cargo"
 export RUSTUP_HOME="`pwd`/.rustup"
 export RUSTUP_INIT_SKIP_PATH_CHECK="yes"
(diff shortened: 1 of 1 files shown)


.buildbot.sh
```diff
@@ -2,27 +2,15 @@
 
 set -e
 
+YKRUSTC_REPO=https://github.com/softdevteam/ykrustc
```
ltratt, line 5: I must admit, this makes the script harder to read IMHO: it's only used in one place, and the code that comes after it heavily depends on the repo being called `ykrustc`. I'd suggest inlining this variable TBH.

.buildbot.sh
```diff
@@ -33,5 +21,27 @@ sh rustup.sh --default-host x86_64-unknown-linux-gnu \
...
+rustup override set ykrustc-stage1
+
+# Test both workspaces using the compiler we just built.
+export RUSTFLAGS="-C tracer=${CI_TRACER_KIND} -D warnings"
+cargo xtask test
+cargo xtask bench
+cargo xtask clean
+
+# Also test the build without xtask, as that's what consumers will do.
+cargo build
```
ltratt, line 41: Does this work -- I thought we *had* to use `xtask`?

.buildbot.sh
```diff
@@ -33,5 +21,27 @@ sh rustup.sh --default-host x86_64-unknown-linux-gnu \
...
+
+# Run rustfmt.
```
ltratt, line 43: Given how lengthy the preceding stages are, should we do `fmt` first? We also won't have to unset `RUSTFLAGS` that way!

.buildbot.sh
bjorn3: Yes it does work for `cargo build`, but all warnings and build progress for the internal workspace is hidden as build script output is only shown when it fails.

vext01: We tested the xtask case a few lines above. This is testing what crates consuming us will do (we have no way to have consumers run xtask, so we had to duplicate some code in `build.rs` -- it's complicated).

.buildbot.sh
vext01: I wondered that myself. But can we do it as a separate PR, as testing these things properly it time consuming.

ltratt: TBH, I'm happy with you raising the commit and doing `bors try`1

vext01:
bors try

vext01:
bors try

vext01:
OK to squash?

ltratt:
Please squash.

vext01:
splat.

ltratt:
Sorry, I confused myself as to which PR is which! This one is fine.

bors r+

PR #291, opened 2021-03-16
Make debugging trace code in GDB easier.

vext01:
**DRAFT DRAFT DRAFT**

We used to be able to set a break point for all entry points to trace
code with `b exec_trace` in gdb. Somewhere along the line this was changed and
it makes debugging harder than it needs to be (you have to step over the
call to get the trace function pointer, which is irritating).

This change fixes this and adds documentation in attempt to prevent this
regressing again.

I'm raising this as a draft because this has a tiny performance overhead of a call. I had planned to mark the separated functions `#[inline(always)]`, but then gdb doesn't find them (presumably the symbols are removed).

Does anyone have any ideas how to have them inlined *and* have gdb find them? We can't `#[no_mangle]` due to type parameters.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -187,7 +187,11 @@ impl MTThread {
         // this thread's tracer.
         if let Some(loc) = loc {
             if let Some(func) = self.transition_location::<I>(loc) {
+                #[cfg(not(debug_assertions))]
+                let ptr = func(ctx);
+                #[cfg(debug_assertions)]
                 let ptr = self.exec_trace(func, ctx);
+
                 if ptr.is_null() {
                     // Trace succesfully executed.
                     return;
@@ -203,6 +207,11 @@ impl MTThread {
         step_fn(ctx)
     }
 
+    /// Call the compiled trace code.
+    /// This is separate from the control point so as to make it easy to get a GDB break point
+    /// immediately before our trace code. It is also named in such a way that `b exec_trace` will
+    /// break at all possible entry points to trace code (there is another one in `CompiledTrace`).
+    #[cfg(debug_assertions)]
     fn exec_trace<I>(
         &mut self,
         func: fn(&mut I) -> *mut RawStopgapInterpreter,
--- a/ykshim_client/src/lib.rs
+++ b/ykshim_client/src/lib.rs
@@ -151,6 +151,21 @@ impl<I> CompiledTrace<I> {
     /// Execute the trace with the given interpreter context.
     pub unsafe fn execute(&self, ctx: &mut I) -> *mut RawStopgapInterpreter {
         let f = mem::transmute::<_, fn(&mut I) -> *mut RawStopgapInterpreter>(self.ptr());
+        #[cfg(not(debug_assertions))]
+        return f(ctx);
+        #[cfg(debug_assertions)]
+        Self::exec_trace(f, ctx)
+    }
+
+    /// Call the compiled trace code.
+    /// This is separate from `execute()` so as to make it easy to get a GDB break point
+    /// immediately before our trace code. It is also named in such a way that `b exec_trace` will
+    /// break at all possible entry points to trace code (there is another one in `MTThread`).
+    #[cfg(debug_assertions)]
+    fn exec_trace(
+        f: fn(&mut I) -> *mut RawStopgapInterpreter,
+        ctx: &mut I,
+    ) -> *mut RawStopgapInterpreter {
         f(ctx)
     }
 }


ykshim_client/src/lib.rs
```diff
@@ -151,6 +151,17 @@ impl<I> CompiledTrace<I> {
...
     pub unsafe fn execute(&self, ctx: &mut I) -> *mut RawStopgapInterpreter {
         let f = mem::transmute::<_, fn(&mut I) -> *mut RawStopgapInterpreter>(self.ptr());
+        Self::exec_trace(f, ctx)
+    }
+
+    /// Call the compiled trace code.
+    /// This is separate from `execute()` so as to make it easy to get a GDB break point
+    /// immediately before our trace code. It is also named in such a way that `b exec_trace` will
+    /// break at all possilbe entry points to trace code (there is another one in `MTThread`).
+    fn exec_trace(
```
bjorn3, line 161:
```suggestion
    #[inline(never)]
    fn exec_trace(
```

ykshim_client/src/lib.rs
```diff
@@ -151,6 +151,17 @@ impl<I> CompiledTrace<I> {
     /// Execute the trace with the given interpreter context.
     pub unsafe fn execute(&self, ctx: &mut I) -> *mut RawStopgapInterpreter {
         let f = mem::transmute::<_, fn(&mut I) -> *mut RawStopgapInterpreter>(self.ptr());
+        Self::exec_trace(f, ctx)
+    }
+
+    /// Call the compiled trace code.
+    /// This is separate from `execute()` so as to make it easy to get a GDB break point
+    /// immediately before our trace code. It is also named in such a way that `b exec_trace` will
+    /// break at all possilbe entry points to trace code (there is another one in `MTThread`).
```
bjorn3, line 160:
```suggestion
    /// break at all possible entry points to trace code (there is another one in `MTThread`).
```

ykrt/src/mt.rs
```diff
@@ -203,6 +203,10 @@ impl MTThread {
         step_fn(ctx)
     }
 
+    /// Call the compiled trace code.
+    /// This is separate from the control point so as to make it easy to get a GDB break point
+    /// immediately before our trace code. It is also named in such a way that `b exec_trace` will
+    /// break at all possible entry points to trace code (there is another one in `CompiledTrace`).
     fn exec_trace<I>(
```
bjorn3, line 210:
```suggestion
    #[inline(never)]
    fn exec_trace<I>(
```

ykshim_client/src/lib.rs
vext01: I think that may help to prevent the symbol disappearing, but ensures the performance hit?

ykshim_client/src/lib.rs
vext01: Oops :P Thanks.

ykshim_client/src/lib.rs
bjorn3: Alternatively it could be

```suggestion
    #[cfg_attr(debug_assertions, inline(never))]
    fn exec_trace(
```

vext01: Perhaps it would be better as: `#[cfg_attr(not(debug_assertions), inline(always)]`.

The symbol only goes missing if it's annotated `#[inline(always)]`.

vext01: If I duplicate the function, I can have both.

ltratt:
What sort of overhead are we talking about here?

ykshim_client/src/lib.rs
bjorn3: Without `#[inline(always)]` it is still possible to inline it, which loses the symbol.

bjorn3:
A couple of instructions including a call and return instruction likely.

ltratt:
Why don't we make this `cfg`able e.g. in debug mode? That way it's zero cost in production.

bjorn3:
That was my suggestion in https://github.com/softdevteam/yk/pull/291#discussion_r595185789.

vext01:
> Why don't we make this cfgable e.g. in debug mode?

Yep, that's exactly what we are doing. It won't help to debug in release mode, but it's better than nothing.

See 900bf6b.

ltratt:
Are we ready to squash? I *think* @bjorn3's comments have been addressed, but please tell me if I'm wrong!

vext01:
I believe we are ready.

ltratt:
Please squash.

vext01:
Splat.

PR #292, opened 2021-03-22
Avoid shell argument parsing.

vext01:
This solves the issue discussed [here](https://github.com/softdevteam/ykbf/pull/12#issuecomment-802979045).

I'm not super-proud of this: passing user controlled stuff to a shell is often asking for trouble, but needs must...


ltratt:
Fine with me if it's OK with @bjorn3.

bjorn3 (approved):

PR #295, opened 2021-03-23
Only reserve space for registers actually in use.

ptersilie:
The first step in our guard failure is to spill all live locals that are
in registers to the stack. This allows us to use any register for
stopgap initialisation without having to worry about overwriting live
values.

Out of simplicity (or rather lazyness) we reserved space for each
register that is available to the register allocater, no matter if it
actually stored a live local or not. We thus sometimes reserved more
space than needed. To make matters worse, this code also had a bug,
which meant we were sometimes actually reserving less space than
required, since the code assumed `REG_POOL` contained all 15 registers
(it doesn't) and then used the register's integer value as an offset
into the the stack offset.

This commit kills two birds with one stone by only reserving space for
registers storing live locals, and then mapping the stack offset to the
register, to avoid this bug from happening in the future.

### DIFF
--- a/internal_ws/ykcompile/src/arch/x86_64/mod.rs
+++ b/internal_ws/ykcompile/src/arch/x86_64/mod.rs
@@ -1330,7 +1330,16 @@ impl TraceCompiler {
         // Reserved stack space for spilled locals during execution.
         let soff = self.stack_builder.size();
         // Reserved memory on the stack to spill live locals to during a guard failure.
-        let live_off = REG_POOL.len() * QWORD_REG_SIZE;
+        let mut llmap = HashMap::new();
+        let mut live_off = 0;
+        for (_, live_locations, _) in &gl {
+            for (_, loc) in live_locations.iter() {
+                if let Location::Reg(reg) = loc {
+                    live_off += QWORD_REG_SIZE;
+                    llmap.insert(*reg, usize::try_from(soff).unwrap() + live_off);
+                }
+            }
+        }
 
         // After allocating stack space for the trace, we pad the stack pointer up to the next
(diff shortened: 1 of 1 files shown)


ptersilie:
Immediately noticed some typos. @vext01 Okay for me to force push a fix before you start reviewing?

vext01:
Go for it.

ptersilie:
Fixed.

vext01:
s/lazyness/laziness/ in commit message.

Other than that, looks good. Did you test with optimised MIR?

vext01:
One (future?) optimisation might be to use a Vec instead of a HashMap. Since we have so few registers, we can probably waste a little memory in exchange for constant-time lookup?

(we'd have to use a special number to mean "not spilled").

ptersilie:
> Did you test with optimised MIR?

I did.

> One (future?) optimisation might be to use a Vec instead of a HashMap.

I'm not sure this is really the bottleneck in the TIR conversion, but I'm also not against that change.

vext01:
Let's just get this fix in now to solve UB in the current code.

Please force push to fix the typo.

ptersilie:
Fixed.

vext01:
```
Diff in /home/buildbot-worker3/worker3/yorick-buildbot-build-script-hw/build/internal_ws/ykcompile/src/arch/x86_64/mod.rs at line 1385:
     fn compile_guards(
         &mut self,
         gl: Vec<(&Guard, HashMap<&Local, Location>, DynamicLabel)>,
-        llmap: HashMap<u8, usize>
+        llmap: HashMap<u8, usize>,
     ) {
```

vext01:
Please fix and squash.

ptersilie:
Whoops. Fixed.

PR #297, opened 2021-03-23
Remove unused code.

ltratt:
It seemed rather odd to me that a constant bool would end up being converted to a usize, but I couldn't see it causing problems. That appears to be because this function isn't actually called from yk or ykrustc.


ltratt:
bors try

vext01:
Probably a leftover from an earlier mode of operation.

ltratt:
The `try` suggests we're OK to merge this (I wanted to be doubly sure!).

PR #298, opened 2021-03-24
Fixes suggested by Clippy.

ltratt:
There are actually two fixes in here, although one is slightly obscured by the indentation change caused by the other.


vext01:
Can you point out the second lint?

ltratt:
`if *b == false`  ->  `if !(*b)`

PR #299, opened 2021-03-24
Move away from a generic, and easy to get wrong, read_int function.

ltratt:
`read_int` is a type-mangling function that expects to convert any constant
integer, or an integer read from memory, to a u128. This effectively turns any
code that calls the function into dynamic typing mode, and is likely to cause
various problems (e.g. with bitwise operations).

This commit breaks out read_bool as a separate function, as it's clearly
different than "normal" integer operations. However this doesn't currently work
as, on b13 at least, it "reads" values of 210 (not 0 or 1) from a byte in one
test! This suggests that the `Terminator::Assert` code only previously worked by
accident because we effectively tested "is <0 or 1> != 210" which of course
always produced `false`.

### DIFF
--- a/internal_ws/yksg/src/lib.rs
+++ b/internal_ws/yksg/src/lib.rs
@@ -383,13 +383,23 @@ impl StopgapInterpreter {
                 expected,
                 target_bb,
             } => {
-                let b = self.read_int(cond) == 1;
-                if b != *expected {
-                    todo!() // FIXME raise error
+                if self.read_bool(cond) != *expected {
+                    todo!();
                 }
                 self.frames.last_mut().unwrap().bbidx = *target_bb;
             }
             t => todo!("{}", t),
+        }
+    }
+
+    fn read_bool(&self, src: &IRPlace) -> bool {
+        if let IRPlace::Const { val: Constant::Int(ConstantInt::UnsignedInt(UnsignedInt::U8(v))), ty: _ } = src {
+            *v != 0
+        } else if let TyKind::Bool = SIR.ty(&src.ty()).kind {
+            let ptr = self.locals().irplace_to_ptr(src);
+            unsafe { std::ptr::read::<u8>(ptr) != 0 }
+        } else {
+            unreachable!();
         }
     }
 


ltratt:
https://github.com/rust-lang/unsafe-code-guidelines/issues/53 and https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/layout/scalars.md, in conjunction [with the C++ spec](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf) (I assume C17 says the same thing -- I couldn't immediately find a version online), are a bit of a mess, with a lot of people being scared to commit to precise details, *but* my conclusion from them is the following:

1. size_of::<bool>() is not fixed -- we incorrectly assume it's `u8`. [It probably always will be `u8`, but we shouldn't assume that.]
2. The valid bit patterns of a bool in memory are 0x0 (false) and 0x1 (true).
3. When converting an integer to a bool, 0 means false (and translates to 0x0) and any other values means true (and translates to 0x1).

bjorn3:
`bool` is defined to use exactly one byte of memory. In addition C uses false = 0 and true = 1 on all non-exotic systems as far as I know. Rust has to match this for `#[repr(C)]` and `extern "C"`. https://doc.rust-lang.org/stable/reference/types/boolean.html

ltratt:
> bool is defined to use exactly one byte of memory.

I don't believe that's true: [this page](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/layout/scalars.md) says "Rust's bool has the same layout as C17's _Bool, that is, its size and alignment are implementation-defined."

> Any value other than 0 and 1 is UB for bools in Rust.

Yes.

internal_ws/yksg/src/lib.rs
```diff
@@ -383,16 +383,26 @@ impl StopgapInterpreter {
...
             t => todo!("{}", t),
         }
     }
 
+    fn read_bool(&self, src: &IRPlace) -> bool {
+        if let IRPlace::Const { val: Constant::Int(ConstantInt::UnsignedInt(UnsignedInt::U8(v))), ty: _ } = src {
+            *v != 0
+        } else if let TyKind::Bool = SIR.ty(&src.ty()).kind {
+            let ptr = self.locals().irplace_to_ptr(src);
+            unsafe { std::ptr::read::<u8>(ptr) != 0 }
```
ptersilie, line 400: We could change this to `unsafe { std::ptr::read::<bool>(ptr as *const bool) != 0 }` to address your concerns about assuming `bool`s will always be `u8`.

ltratt: Yes, I've got that in a local branch, though the assumption is embedded in several other places. However, I still end up reading bytes whose value is > 1 (i.e. we've caused undefined behaviour somewhere).

vext01:
What's the status of this PR?

ltratt:
It is currently waiting on the stoargelive/dead problem we're dealing with elsewhere. I may close it, or carry on with this, or not, depending on how long that takes and the state of the code base at that point.

ltratt:
It's now fairly clear that this PR will soon be out of date and will need starting from scratch. Closing.

PR #308, opened 2021-03-27
Simple tidy ups

ltratt:
This PR contains a few manually determined tidy ups: each commit has a bit more detail.

PR #309, opened 2021-03-28
Depub

ltratt:
Needs https://github.com/softdevteam/yk/pull/308 to be merged first. I'm raising this as useful context for https://github.com/softdevteam/depub/pull/1.

This PR uses https://github.com/softdevteam/depub/ to reduce the visibility of quite a lot more items in yk. In some cases, we only slightly reduce visibility (`pub(crate)` to `pub(super)`) but there are a reasonable number where we get rid of `pub` entirely. We're also able to gate a small number of items behind `#[cfg(feature="testing")]` to make clear when something is only needed for our (unusual) test setup. Overall, this PR has no functional change, but should make it a tiny little bit easier to understand the code base.

### DIFF
--- a/internal_ws/yktrace/src/swt.rs
+++ b/internal_ws/yktrace/src/swt.rs
@@ -29,7 +29,7 @@ impl ThreadTracerImpl for SWTThreadTracer {
     }
 }
 
-pub(crate) fn start_tracing() -> ThreadTracer {
+pub(super) fn start_tracing() -> ThreadTracer {
     TRACE_BUF.with(|trace_buf| {
         assert!(trace_buf.is_empty());
     });
--- a/internal_ws/ykcompile/src/arch/mod.rs
+++ b/internal_ws/ykcompile/src/arch/mod.rs
@@ -4,4 +4,4 @@
 compile_error!("Currently only linux x86_64 is supported.");
 
 #[cfg(all(target_arch = "x86_64", target_os = "linux"))]
-pub(crate) mod x86_64;
+pub(super) mod x86_64;
--- a/internal_ws/ykcompile/src/arch/x86_64/store.rs
+++ b/internal_ws/ykcompile/src/arch/x86_64/store.rs
@@ -7,7 +7,7 @@ use yktrace::sir::SIR;
 
 impl TraceCompiler {
     /// Store the value in `src_loc` into `dst_loc`.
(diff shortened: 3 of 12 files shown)


ltratt:
If you're OK with me doing so, I suggest I rebase this against master and force push when #308 has merged before you review?

internal_ws/yktrace/src/swt.rs
```diff
@@ -29,7 +29,8 @@ impl ThreadTracerImpl for SWTThreadTracer {
     }
 }
 
-pub(crate) fn start_tracing() -> ThreadTracer {
```
vext01, line 32: I'm not sure if this is right. Why would starting tracing be a testing feature?

It's probably because your tool build with hardware tracing rustflags (and not software tracing)?

vext01: Similarly for the other bits in this file.

ltratt: > It's probably because your tool build with hardware tracing rustflags (and not software tracing)?

Yes, I used `hw`. I didn't realise that there'd be a difference here with `sw`? Would a bors try flush this out?

vext01: > I didn't realise that there'd be a difference here with sw?

My guess is that compile-time guards like [this](https://github.com/softdevteam/yk/blob/6023c7c091f2483f5a22344a73be4575ad2be254/internal_ws/yktrace/src/lib.rs#L74) would give inaccurate results if the oracle uses only `-C tracer=hw`.

Ideally your oracle would include both kinds of tracing, but read on...

> Would a bors try flush this out?

No, because we don't yet test software tracing with cranelift. I'm not sure it's ready yet, but @bjorn3 is the person to ask.

ltratt: Having looked at this again, I agree that applying this to `swt.rs` is incorrect. Fixed in https://github.com/softdevteam/yk/pull/309/commits/0703d9473e51a224d08fed953e70b32638b284f0.

vext01:
I think this is ready for squashing (and may as well sync the branch while we are here).

ltratt:
OK, as soon as https://github.com/softdevteam/yk/pull/308 merges, I'll sync and force push.

ltratt:
Rebased and squashed.

PR #314, opened 2021-04-02
Add a job queueing system to cap the number of worker threads

ltratt:
Previously yk could spin up an arbitrary number of compilation threads, bounded only by how many loops are present in the end users program. For more complex compilation jobs -- particularly for multi-threaded programs -- we would probably end up making the system grind to a near-halt as we would spend nearly all our time bouncing between partially complete compilation threads, while interpreter threads make very little progress.

This PR adds the concept of "worker threads" (where "compilation" is our only current form of "worker") and bounds their number. A global queue is used when there is more work to be done than threads. When there is no work to be done, threads block without consuming CPU time. A "job" is a `Send`able function (this is a very obvious design in retrospect, but it took me 3 attempts to realise this is the right model for us).

The PR isn't huge, but I've kept it as a small, complete, commits as I think that's easier to understand. https://github.com/softdevteam/yk/commit/669a0a00f4b86299e7eff41a9e72f5916f67ecb7 adds the necessary config bits but does nothing with them. https://github.com/softdevteam/yk/commit/6af51af484f2c3b69212159dad5eaa6977a4c2db adds a basic queue but allows worker threads to spin endlessly when there is no work to do. Those two commits use `std::sync::Mutex` because our use of `catch_unwind` prevents us using `parking_lot`: when I realised that I couldn't even use `std::sync::CondVar` it became clear that `catch_unwind` is not worth the bother so... off with its head (https://github.com/softdevteam/yk/commit/432d456c867bf85e9d2eea2f3ae5f774a8f1b946)! The final major commit uses parking_lot's `Condvar` to ensure that worker threads block gracefully when there is no work for them to do (https://github.com/softdevteam/yk/commit/b3c483dcb0817efd121ed4f613db881c3d3f8090).

That might sound frightening, but the core part of the PR is perhaps 50LoC: the remainder is roughly evenly split between configuration and testing code.

### DIFF
--- a/traced/ykrt/Cargo.toml
+++ b/traced/ykrt/Cargo.toml
@@ -6,6 +6,7 @@ edition = "2018"
 license = "Apache-2.0 OR MIT"
 
 [dependencies]
+num_cpus = "1"
 parking_lot = "0.11"
 parking_lot_core = "0.8"
 strum = { version = "0.20", features = ["derive"] }
--- a/traced/ykrt/src/mt.rs
+++ b/traced/ykrt/src/mt.rs
@@ -4,6 +4,7 @@
 use std::time::Duration;
 use std::{
     cmp,
+    collections::VecDeque,
     error::Error,
     io,
     marker::PhantomData,
@@ -12,6 +13,7 @@ use std::{
     ptr,
     rc::Rc,
     sync::{
+        self,
         atomic::{AtomicBool, AtomicU32, AtomicUsize, Ordering},
         Arc,
     },
@@ -24,7 +26,8 @@ use parking_lot_core::SpinWait;
 
 use crate::location::{HotLocation, Location, State, ThreadIdInner};
 use untraced_api::{
-    compile_trace, start_tracing, RawStopgapInterpreter, StopgapInterpreter, TracingKind,
+    compile_trace, start_tracing, CompiledTrace, RawStopgapInterpreter, SirTrace,
+    StopgapInterpreter, TracingKind,
 };
 
 // The HotThreshold must be less than a machine word wide for [`Location::Location`] to do its
@@ -139,6 +142,37 @@ impl MT {
             }
         })
     }
+
+    /// Queue `job` to be run on a worker thread.
+    fn queue_job(&self, job: Box<dyn FnOnce() + Send>) {
+        // We have a very simple model of worker threads. Each time a job is queued, we spin up a
+        // new worker thread iff we aren't already running the maximum number of worker threads.
+        // Once started, a worker thread never dies, and spins endlessly waiting for work.
+
+        let inner = Arc::clone(&self.inner);
+        inner.job_queue.lock().unwrap().push_back(job);
+
+        let max_jobs = inner.max_worker_threads.load(Ordering::Relaxed);
+        if inner.active_worker_threads.load(Ordering::Relaxed) < max_jobs {
+            // At the point of the `load` on the previous line, we weren't running the maximum
+            // number of worker threads. There is now a possible race condition where multiple
+            // threads calling `queue_job` could try creating multiple worker threads and push us
+            // over the maximum worker thread limit.
+            if inner.active_worker_threads.fetch_add(1, Ordering::Relaxed) > max_jobs {
+                // Another thread(s) is also spinning up another worker thread and they won the
+                // race.
+                inner.active_worker_threads.fetch_sub(1, Ordering::Relaxed);
+                return;
+            }
+
+            let inner_cl = Arc::clone(&self.inner);
+            thread::spawn(move || loop {
+                if let Some(x) = inner_cl.job_queue.lock().unwrap().pop_front() {
+                    x();
+                }
+            });
+        }
+    }
 }
 
 impl Drop for MT {
@@ -153,6 +187,8 @@ struct MTInner {
     /// The number of threads currently running the user's interpreter (directly or via JITed
     /// code).
     active_user_threads: AtomicUsize,
+    /// The ordered queue of compilation worker functions.
+    job_queue: sync::Mutex<VecDeque<Box<dyn FnOnce() + Send>>>,
     /// The hard cap on the number of worker threads.
     max_worker_threads: AtomicUsize,
     /// How many worker threads are currently running. Note that this may temporarily be `>`
@@ -195,6 +231,7 @@ impl MTInner {
 
         let mtc = Self {
             hot_threshold: AtomicHotThreshold::new(hot_threshold),
+            job_queue: sync::Mutex::new(VecDeque::new()),
             active_user_threads: AtomicUsize::new(1),
             max_worker_threads: AtomicUsize::new(max_worker_threads),
             active_worker_threads: AtomicUsize::new(0),
@@ -451,25 +488,15 @@ impl MTThread {
                     // while tracing.
                     match opt.take().unwrap().1.stop_tracing() {
                         Ok(sir) => {
-                            // Start a compilation thread.
                             let mtx = Arc::new(Mutex::new(None));
                             let mtx_cl = Arc::clone(&mtx);
                             *hl = HotLocation::Compiling(mtx);
                             loc.unlock();
-
-                            Rc::get_mut(&mut self.inner).unwrap().tracing = None;
-                            thread::spawn(move || {
-                                let compiled = compile_trace::<I>(sir).unwrap();
-                                *mtx_cl.lock() = Some(Box::new(compiled));
-                                // FIXME: although we've now put the compiled trace into the mutex, there's no
-                                // guarantee that the Location for which we're compiling will ever be executed
-                                // again. In such a case, the memory has, in essence, leaked.
-                            });
-
-                            return None;
+                            self.queue_compile_job(sir, mtx_cl);
                         }
                         Err(_) => todo!(),
                     }
+                    return None;
                 }
                 HotLocation::DontTrace => {
                     loc.unlock();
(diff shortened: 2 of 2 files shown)


vext01:
Before I review this:

There are a lot of thread pool crates on crates.io. Is there something about our system that prevents us from using them?

I'm not against implementing our own thread pool, but at the same time if there's no need then it's probably going to be easier to use an existing library?

ltratt:
The "thread pool" bit is, in a sense, about 10LoC: most of the rest of the PR is the plumbing we'd have to add no matter what library we used. Does this mean that we shouldn't use a library? Honestly, I don't know. This PR is really a minimal attempt to stop yk grinding to a halt if we fire up too many compilation threads. I'm fairly sure we will want something more sophisticated later, but I don't know exactly what it should like. I'm inclined to wait and see before trying to do anything more clever than this PR. [My first attempt at this was more clever, and a) it was too complex b) I realised that I don't know what the right trade-offs in our setting will end up looking like.]

vext01:
Let's move ahead with this, and if we start to encounter previously unforeseen issues, we can revisit the question.

There's a good chance this is just fine (tm).

I'll do my review now.

traced/ykrt/src/mt.rs
```diff
@@ -123,7 +129,7 @@ impl Drop for MT {
 /// The innards of a meta-tracer.
 struct MTInner {
     hot_threshold: AtomicHotThreshold,
-    active_threads: AtomicUsize,
+    active_user_threads: AtomicUsize,
```
vext01, line 187: What's a user thread?

Broader question: should comment all of our struct and enum fields no matter how obvious it seems?

traced/ykrt/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2018"
 license = "Apache-2.0 OR MIT"
 
 [dependencies]
+num_cpus = "1"
```
vext01, line 9: I just checked if this copes with offline CPUs due to cgroups etc. It does:
https://docs.rs/num_cpus/1.13.0/num_cpus/fn.get.html

traced/ykrt/src/mt.rs
```diff
@@ -47,6 +51,7 @@ impl MTBuilder {
     pub fn new() -> Result<Self, Box<dyn Error>> {
         Ok(Self {
             hot_threshold: DEFAULT_HOT_THRESHOLD,
+            max_worker_threads: cmp::max(1, num_cpus::get() - 1),
```
vext01, line 54: So if I see it right, `num_cpus::get()` return different values over time? For example if in Linux I do:

```
# echo 0 > /sys/devices/system/cpu/cpu6/online
```

Then I can disable a core.

So should we delay this query to the latest moment?

This is a pretty spacious bikeshed.

ltratt: > This is a pretty spacious bikeshed.

You said it ;)

bjorn3: Shouldn't this use `cmp::min`? Otherwise there will be either 0 or 1 worker threads.

ltratt: Using `max` guarantees we get at least 1 worker thread; but if you have, say, 12 cores, you'll get 11 worker threads.

bjorn3: Got it mixed up.

traced/ykrt/src/mt.rs
```diff
@@ -139,6 +142,37 @@ impl MT {
...
+        let inner = Arc::clone(&self.inner);
+        inner.job_queue.lock().unwrap().push_back(job);
+
+        let max_jobs = inner.max_worker_threads.load(Ordering::Relaxed);
+        if inner.active_worker_threads.load(Ordering::Relaxed) < max_jobs {
+            // At the point of the `load` on the previous line, we weren't running the maximum
+            // number of worker threads. There is now a possible race condition where multiple
+            // threads calling `queue_job` could try creating multiple worker threads and push us
+            // over the maximum worker thread limit.
+            if inner.active_worker_threads.fetch_add(1, Ordering::Relaxed) > max_jobs {
```
vext01, line 156: Why don't we only do this check and not the one at line 156 as well? I suppose you are trying to minimise the chance of having to do the correctional subtract?

traced/ykrt/src/mt.rs
```diff
@@ -139,6 +142,37 @@ impl MT {
...
+                // Another thread(s) is also spinning up another worker thread and they won the
+                // race.
+                inner.active_worker_threads.fetch_sub(1, Ordering::Relaxed);
+                return;
+            }
+
+            let inner_cl = Arc::clone(&self.inner);
+            thread::spawn(move || loop {
```
vext01: Do we only check if we can assign a job to a worker thread at the time that a job is queued then?

vext01: Wait... It spins forever even if there is no work?

So if we have at some time in the past had 8 threads compiling simultaneously, then we will forever have 8 threads spinning looking for work?

traced/ykrt/src/mt.rs
ltratt: In the average case, we do one load and a compare, which is quicker than the load/sub/add of `fetch_add`.

traced/ykrt/src/mt.rs
ltratt: > Do we only check if we can assign a job to a worker thread at the time that a job is queued then?

No, we spin worker threads up each time a new job is added, but it's possible that by the time they've started working that another worker thread has picked the job up.

> Wait... It spins forever even if there is no work?

No, they gracefully wait (https://github.com/softdevteam/yk/pull/314/commits/b3c483dcb0817efd121ed4f613db881c3d3f8090).

vext01:
Having read more, I'm getting more nervous about introducing more hairy concurrency code. 
Event loops are notoriously hard to write and I'm almost certain we are going to get it wrong. Are we sure that we want a hand-rolled solution?

vext01:
[the mere mention of conditional variables strikes me with fear -- i've never managed to understand]

ltratt:
> Event loops are notoriously hard to write and I'm almost certain we are going to get it wrong. Are we sure that we want a hand-rolled solution?

The loop is 6LoC. It's not that scary.

traced/ykrt/src/mt.rs
```diff
@@ -179,7 +186,7 @@ struct MTInner {
     hot_threshold: AtomicHotThreshold,
     active_user_threads: AtomicUsize,
     /// The ordered queue of compilation worker functions.
-    job_queue: Mutex<VecDeque<Box<dyn FnOnce() + Send>>>,
+    job_queue: (Condvar, Mutex<VecDeque<Box<dyn FnOnce() + Send>>>),
```
vext01, line 189: The cond var is guarded by the mutex, so should we put it inside the mutex alongside the vecdeque?

They don't do this in the [rust docs](https://doc.rust-lang.org/beta/std/sync/struct.Condvar.html), so I expect to be wrong, but wouldn't this stop a thread from accessing the var without holding the lock?

ltratt: Note we're using `parking_lot` here https://docs.rs/parking_lot/0.11.1/parking_lot/struct.Condvar.html -- if we put the condvar in the mutex, I think that we wouldn't ever be able to unlock it to put a thread to sleep.

vext01:
When a worker thread thread crashes do we lose the slot in the thread pool? Is there a way to keep hold of the thread handle and re-increment the number of active threads?

ltratt:
At the moment, if a worker or user thread crashes, that thread is taken out. At one point I thought we shuld use `catch_unwind` but now I'm not sure. Frankly, if a thread panics, I think all bets are off in all other threads too. I have vaguely wondered if the worker function should return `Result` so that the job runner can catch things in that way. Any thoughts?

vext01:
>  I have vaguely wondered if the worker function should return Result so that the job runner can catch things in that way. Any thoughts?

I think all that you'd care about it whether a thread exits (regardless of any potential result). If you retain the thread handle `spawn()` gives you, then you could try and join the thread. If it does join (which it shouldn't) then something went wrong.

I see what you mean about the correctness of the system at that point though...

vext01:
I've done about the best I can with reviewing this. I'm not totally confident in the concurrency -- I find `CondVar` hard to understand, but I *think* I've convinced myself that it's not obviously incorrect.

There's an outstanding question about commenting up our fields.

I wonder if more tests might be wise -- assuming there are any that can be practically written.

ltratt:
I am not sure how to write any other meaningful tests TBH because with this sort of thing you can't force all possible interleavings -- you just have to hope they happen.

traced/ykrt/src/mt.rs
ltratt: Good point. Fixed in https://github.com/softdevteam/yk/pull/314/commits/a465e5ff4a3e3c48eed4934d64e72622fae5a01f.

vext01:
> I am not sure how to write any other meaningful tests TBH because with this sort of thing you can't force all possible interleavings -- you just have to hope they happen.

Should we put the test in a loop to encourage other interleavings?

ltratt:
> Should we put the test in a loop to encourage other interleavings?

That's more-or-less what the existing test does. It runs 1 user thread per CPU core then a further (num cores)/2 worker threads, so the user and worker threads end up contending a lot and so, hopefully, we test at least many of the possible interleavings. We could run things for even longer, but it's hard to know if we're just burning electricity for no use at some point. I don't have a firm opinion on this.

traced/ykrt/src/mt.rs
```diff
@@ -184,6 +184,8 @@ impl Drop for MT {
 /// The innards of a meta-tracer.
 struct MTInner {
     hot_threshold: AtomicHotThreshold,
+    /// How many threads running user code (i.e. their interpreter, whether directly, or JITed) are
+    /// there at the moment?
```
vext01, line 188: One more bike shed and I will leave you alone I promise.

Let's make this shorter and more affirmative:

> The number of threads running the user's interpreter (directly, or via JITted code).

ltratt: Try .

vext01:
Please squash.

ltratt:
Squashed.

PR #317, opened 2021-04-06
ykpack tidy

ltratt:
This is a mostly-shallow tidy-up of ykpack::types, mostly adding constructors, and adding methods instead of having code access attribute directly. The main aim here is to make it a bit easier to change internal representations in the future. https://github.com/softdevteam/yk/commit/b95a75b614cc720033f497c4de721c3a2a5dbfe4 is a little example of how this can help.

This PR raises one interesting possibility: at the moment `CguHash` is a `u64`. If we convert that to a `u32`, it leads to a saving of 5-10% in the size of the serialised output from ykpack across a `yk` build. That does raise a few more issues than I want to think about in this PR, however.

Note that merging should happen via https://github.com/softdevteam/ykrustc/pull/179.

### DIFF
--- a/untraced/ykcompile/src/stack_builder.rs
+++ b/untraced/ykcompile/src/stack_builder.rs
@@ -13,26 +13,26 @@ use std::convert::{TryFrom, TryInto};
 #[derive(Default, Debug)]
 pub(super) struct StackBuilder {
     /// Keeps track of how many bytes have been allocated.
-    stack_top: u64,
+    stack_top: usize,
 }
 
 impl StackBuilder {
     /// Allocate an object of given size and alignment on the stack, returning a `Location::Mem`
     /// describing the position of the allocation. The stack is assumed to grow down.
-    pub(super) fn alloc(&mut self, size: u64, align: u64) -> Location {
+    pub(super) fn alloc(&mut self, size: usize, align: usize) -> Location {
         self.align(align);
         self.stack_top += size;
         Location::new_mem(RBP.code(), -i32::try_from(self.stack_top).unwrap())
     }
 
     /// Aligns `offset` to `align` bytes.
-    fn align(&mut self, align: u64) {
+    fn align(&mut self, align: usize) {
         let mask = align - 1;
         self.stack_top = (self.stack_top + mask) & !mask
     }
 
     /// Total allocated stack size in bytes.
-    pub(super) fn size(&self) -> u32 {
+    pub(super) fn size(&self) -> usize {
         self.stack_top.try_into().unwrap()
     }
 }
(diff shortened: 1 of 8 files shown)


untraced/ykcompile/src/stack_builder.rs
```diff
@@ -13,26 +13,26 @@ use std::convert::{TryFrom, TryInto};
 #[derive(Default, Debug)]
 pub(super) struct StackBuilder {
     /// Keeps track of how many bytes have been allocated.
-    stack_top: u64,
+    stack_top: usize,
```
vext01, line 16: I'm surprised this works without additional `try_from` changes elsewhere. I thought there was some DWARF-related nonsense, but maybe we evolved that out already..

ltratt: I was pleasantly surprised too!

untraced/ykpack/src/types.rs
```diff
@@ -136,6 +124,30 @@ impl Ty {
     }
 }
 
+#[cfg(target_pointer_width = "64")]
+impl Ty {
+    // Pack size & align into 64 bits (48 bits for size, 16 bits for align).
```
vext01, line 129: I agree that this is good for performance, but it'd be irresponsible of me to not ask: how future proof is the 48-bit address space model?

ltratt: It's not an address: it's a size. If you manage to make a static type which exceeds 48 bits (~131072GiB) in size, I will be very impressed. More to the point, you'll get a compile-time error as ykpack will refuse to encode it, so there isn't a correctness issue (though there is a usability issue).

vext01: As long as we are not going to become the next "no-one will need more than 640KB of RAM" :)

It should be easily changed later I hope anyway.

ltratt: I think we're a little way off from machines that can have 48 bits worth of RAM in, and we'll probably have some advanced notice of that fact :) Fortunately we can change the representation transparently: that's the beauty of using methods not attributes :)

untraced/ykpack/src/build.rs
```diff
@@ -104,13 +98,13 @@ impl SirBuilder {
     /// Appends a statement to the specified basic block.
     pub fn push_stmt(&mut self, bb: BasicBlockIndex, stmt: Statement) {
         self.func.blocks[usize::try_from(bb).unwrap()]
-            .stmts
+            .mut_stmts()
```
vext01, line 101: Would `stmts_mut()` be more idiomatic? E.g. `null_mut()`.

vext01:
I can't really argue with this. Just one comment remains.

untraced/ykpack/src/build.rs
ltratt: Good point. Fixed in https://github.com/softdevteam/yk/pull/317/commits/23f0d81fbc410d9828411328b66e64002acb226b.

ltratt:
If you're OK with that change, I think we're ready to squash.

vext01:
Please squasharoo.

ltratt:
Squashed.

ltratt:
bors r-

You need to do this in the ykrustc accompanying PR I think?

ltratt:
I think this is ready for merging now the ykrustc PR has merged.

PR #319, opened 2021-04-06
Remove opt-level hacks from cargo configs.

vext01:
Requires https://github.com/softdevteam/ykrustc/pull/180/ to be merged first.

Now that ykrustc automatically turns off optimisation for hardware
tracing, we can simplify our build system a little.


vext01:
This is good to go now.

PR #322, opened 2021-04-08
Make ykpack::Body's attributes private.

ltratt:
This was a bit more involved than I expected as ykrustc gets mutable references in more than one place.

See also https://github.com/softdevteam/ykrustc/pull/181.


vext01:
This seems ok to me. We need an override directive in the PR descr, right?

ltratt:
There is one in the ykrustc side of this PR.

vext01:
Ah yes. Sorry.

PR #323, opened 2021-04-08
Tidy and refactor the stopgap interpreter

ltratt:
This PR does a fair chunk of mostly-not-very-interesting work to the stopgap interpreter e.g. to make it reflect standard interpreter terminology (https://github.com/softdevteam/yk/commit/15805b0bef0ad1c60b1838d5ea5e6b23dfda84dc) or allocate and clone less memory (https://github.com/softdevteam/yk/commit/6d5157cd8ad4331cd30ff651b4165a09df1dace9). Each commit has, hopefully, adequate motivation attached to it. There's more that could and should be done here, but this feels like enough work to be worth reviewing as-is.

[I tried including a benchmark in https://github.com/softdevteam/yk/commit/dd1f20b8e7496e165a88368c0cf799d472d9951c. I'm not sure if it's measuring what I hoped or not. Either way, it shows no difference before and after this PR, which is a bit disappointing as this PR *should* allocate less memory, but it may be that's an insignificant part of the overall execution time. Profiling and optimising properly are for a future PR.]

### DIFF
--- a/traced/tests/src/stopgap/guard_fail.rs
+++ b/traced/tests/src/stopgap/guard_fail.rs
@@ -1,5 +1,7 @@
 //! Testing the stopgap interpreter via guard failures.
 
+extern crate test;
+use self::test::Bencher;
 use to_untraced::{compile_trace, start_tracing, StopgapInterpreter, TracingKind};
 
 #[test]
@@ -217,4 +219,64 @@ fn trace_looping_nested() {
     let mut args = InterpCtx(0);
     unsafe { ct.execute(&mut args) };
     assert_eq!(args.0, 100);
+}
+
+#[bench]
+fn bench_recursion(b: &mut Bencher) {
+    struct InterpCtx {
+        fail_guard: bool,
+        x: u8,
+    }
+
+    // Test that the SIR interpreter can deal with new recursions after a guard failure.
+    fn rec(i: u8) -> u8 {
+        if i < 100 {
+            return rec(i + 1);
+        } else {
+            return i;
+        }
+    }
+
+    #[interp_step]
+    fn interp_step(ctx: &mut InterpCtx) -> bool {
+        if ctx.fail_guard {
+            ctx.x = rec(ctx.x);
+        } else {
+            ctx.x = rec(ctx.x - 1);
+        }
+        true
+    }
+
+    let mut ctx = InterpCtx {
+        fail_guard: false,
+        x: 101,
+    };
+    #[cfg(tracermode = "hw")]
+    let th = start_tracing(TracingKind::HardwareTracing);
+    #[cfg(tracermode = "sw")]
+    let th = start_tracing(TracingKind::SoftwareTracing);
+    interp_step(&mut ctx);
+    let sir_trace = th.stop_tracing().unwrap();
+    let ct = compile_trace(sir_trace).unwrap();
+    ctx = InterpCtx {
+        fail_guard: false,
+        x: 101,
+    };
+    assert!(unsafe { ct.execute(&mut ctx).is_null() });
+    assert_eq!(ctx.x, 100);
+    b.iter(|| {
+        for _ in 0..10 {
+            // Execute the trace with a context that causes a guard to fail.
+            let mut ctx = InterpCtx {
+                fail_guard: true,
+                x: 0,
+            };
+            let ptr = unsafe { ct.execute(&mut ctx) };
+            assert!(!ptr.is_null());
+            // Check that running the interpreter gets us the correct result.
+            let mut si: StopgapInterpreter = StopgapInterpreter(ptr);
+            unsafe { si.interpret() };
+            assert_eq!(ctx.x, 100);
+        }
+    });
 }
--- a/untraced/yksg/src/lib.rs
+++ b/untraced/yksg/src/lib.rs
@@ -184,14 +184,12 @@ macro_rules! make_binop {
     };
 }
 
-/// An interpreter stack frame, containing allocated memory for the frames locals, and the function
-/// symbol name and basic block index needed by the interpreter to continue interpreting after
-/// returning from a function call.
+/// An interpreter function frame representing the current state of an executing function.
 #[derive(Debug)]
 struct Frame {
-    /// Allocated memory holding live locals.
+    /// This frame's local variables.
     mem: LocalMem,
-    /// The current basic block index of this frame.
+    /// This frame's program counter (which always increments in terms of basic blocks).
     bbidx: ykpack::BasicBlockIndex,
     /// Body of this stack frame.
     body: Arc<Body>,
--- a/untraced/yksg/src/lib.rs
+++ b/untraced/yksg/src/lib.rs
@@ -185,7 +185,7 @@ impl StopgapInterpreter {
             frames.push(frame);
         }
         let mut sg = StopgapInterpreter { frames, rv: true };
-        let frame = sg.frames.last().unwrap();
+        let frame = sg.peek_frame();
         // Since we start in the block where the guard failed, we immediately skip to the
         // terminator and interpret it to initialise the block where actual interpretation needs to
         // start.
@@ -195,6 +195,16 @@ impl StopgapInterpreter {
             sg.terminator(&body.blocks()[pc].term());
         }
         sg
+    }
+
+    // Return a reference to the most recent frame on the stack.
+    fn peek_frame(&self) -> &Frame {
+        &self.frames.last().unwrap()
+    }
+
+    // Return a mutable reference to the most recent frame on the stack.
+    fn peek_frame_mut(&mut self) -> &mut Frame {
+        self.frames.last_mut().unwrap()
     }
 
     /// Given the symbol name of a function, generate a `Frame` which allocates the precise
@@ -267,7 +277,7 @@ impl StopgapInterpreter {
                 };
 
                 // Initialise the new stack frame.
-                let oldframe = self.frames.last().unwrap();
+                let oldframe = self.peek_frame();
                 let mut newframe = StopgapInterpreter::create_frame(&fname);
                 newframe.copy_args(args, oldframe);
                 self.frames.push(newframe);
@@ -310,7 +320,7 @@ impl StopgapInterpreter {
                 otherwise_bb,
             } => {
                 let val = self.read_int(discr);
-                let frame = self.frames.last_mut().unwrap();
+                let frame = self.peek_frame_mut();
                 frame.pc = *otherwise_bb;
                 for (i, v) in values.iter().enumerate() {
                     if val == *v {
@@ -320,7 +330,7 @@ impl StopgapInterpreter {
                 }
             }
             Terminator::Goto(bb) => {
-                self.frames.last_mut().unwrap().pc = *bb;
+                self.peek_frame_mut().pc = *bb;
             }
             Terminator::Assert {
                 cond,
@@ -331,7 +341,7 @@ impl StopgapInterpreter {
                 if b != *expected {
                     todo!() // FIXME raise error
                 }
-                self.frames.last_mut().unwrap().pc = *target_bb;
+                self.peek_frame_mut().pc = *target_bb;
             }
             t => todo!("{}", t),
         }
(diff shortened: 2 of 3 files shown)


ltratt:
I tried benchmarking what I *think* we actually want to benchmark in https://github.com/softdevteam/yk/pull/323/commits/3fd514dff48f33883c102bffde282c50599ab27b *but* I get a test failure when doing so. I wonder if this might be due to the same thing you're fixing @ptersilie?

untraced/yksg/src/lib.rs
```diff
@@ -18,7 +18,7 @@ use ykpack::{
 use yktrace::sir::{RETURN_LOCAL, SIR};
 
 /// Stores information needed to recreate stack frames in the StopgapInterpreter.
-pub struct FrameInfo {
+pub struct IncomingFrame {
```
ptersilie, line 23: Not sure about the name `IncomingFrame`, but honestly don't have a better idea either. @vext01 is good at naming stuff.

untraced/yksg/src/lib.rs
```diff
@@ -203,7 +203,7 @@ struct StackFrame {
 /// over.
 pub struct StopgapInterpreter {
     /// Active stack frames (most recent last).
-    frames: Vec<StackFrame>,
+    frames: Vec<Frame>,
```
ptersilie, line 159: I think the reasoning was that from the StopgapInterpreter's point of view `frames` is the stack on which is stores the stack frames, but I agree with this change nonetheless.

untraced/yksg/src/lib.rs
vext01: When the stop-gap interpreter pushes a new frame, will it also be of this same type?

In that context, the frame isn't "incoming", and in fact I find the new name just raises questions.

Was there  a reason this can't just be `Frame`?

vext01: I see, we already have a `Frame` below.

If this struct is only used during initialisation, then `IncomingFrame` makes sense, but how about `InitFrame`?

traced/tests/src/stopgap/guard_fail.rs
```diff
@@ -218,3 +220,45 @@ fn trace_looping_nested() {
...
+    let th = start_tracing(TracingKind::SoftwareTracing);
+    interp_step(&mut ctx);
+    let sir_trace = th.stop_tracing().unwrap();
+    let ct = compile_trace(sir_trace).unwrap();
+
+    b.iter(|| {
+        let mut args = InterpCtx{compiled: true, count: 0};
+        unsafe { ct.execute(&mut args) };
+        assert_eq!(args.count, 255);
+    });
```
vext01, line 263: FWIW, this benchmark looks sensible to me.

untraced/yksg/src/lib.rs
```diff
@@ -81,12 +87,12 @@ impl LocalMem {
     }
 
     /// Copy over the call arguments from another frame.
-    unsafe fn copy_args(&mut self, args: &[IRPlace], frame: &LocalMem) {
+    unsafe fn copy_args(&mut self, args: &[IRPlace], oframe: &Frame) {
```
vext01, line 90: what is the `o` in `oframe` for?

untraced/yksg/src/lib.rs
```diff
@@ -277,15 +229,14 @@ impl StopgapInterpreter {
...
-            .mem
             .local_ptr(&yktrace::sir::INTERP_STEP_ARG);
         std::ptr::write::<*mut u8>(ptr as *mut *mut u8, ctx);
     }
 
     pub unsafe fn interpret(&mut self) -> bool {
         while let Some(frame) = self.frames.last() {
             let body = frame.body.clone();
-            let block = &body.blocks()[usize::try_from(frame.bbidx).unwrap()];
+            let block = &body.blocks()[usize::try_from(frame.pc).unwrap()];
```
vext01, line 239: Make an accessor for `pc`?

traced/tests/src/stopgap/guard_fail.rs
ptersilie: This does not benchmark the stopgap interpreter though. It's actually never run. This particular test `trace_looping_nested` just makes sure that "trace stitching" still works even if other functions return `true`.

The `recursion1/2` tests actually execute the stopgap interpreter, but it probably doesn't run long enough to bench it. This test is a better candidate for benchmarking the stopgap interpreter as it will fail a guard multiple times over: https://github.com/softdevteam/yk/blob/master/traced/ykrt/src/mt.rs#L1172.

untraced/yksg/src/lib.rs
```diff
@@ -183,17 +177,10 @@ impl StopgapInterpreter {
...
-                locals: fi.locals,
-                offsets: body.offsets().clone(),
-                layout: Layout::from_size_align(layout.0, layout.1).unwrap(),
-            };
             let frame = Frame {
-                mem,
+                locals: fi.locals,
                 pc: u32::try_from(fi.bbidx).unwrap(),
-                body: fi.body.clone(),
+                body: Arc::clone(&fi.body),
```
ptersilie, line 183: Is this just to make clearer that `fi.body` is an `Arc`, or does this do something different?

traced/tests/src/stopgap/guard_fail.rs
```diff
@@ -222,8 +222,43 @@ fn trace_looping_nested() {
...
-fn bench_looping_nested(b: &mut Bencher) {
+fn bench_looping(b: &mut Bencher) {
+    struct InterpCtx {
+        compiled: bool,
+        count: u8
+    }
+
+    #[interp_step]
+    fn interp_step(io: &mut InterpCtx) -> bool {
+        if io.compiled {
```
ptersilie, line 233: The reason why this benchmark doesn't work, is that the stopgrap interpreter is never run. You compiled `interp_step` with `io.compiled = false`, so it compiles the `else` branch on this function. Then you set `io.compiled = true` which I assume you did to immediately invoke the stopgap interpreter to benchmark it. However, `ct.execute` doesn't automatically run the stopgap interpreter, rather it returns a pointer to the stopgap interpreter which we can then cast to a `StopgapInterpreter` and run it (see https://github.com/softdevteam/yk/blob/master/traced/ykrt/src/mt.rs#L281 how this is done).

Keep in mind that the stopgap interpreter only finishes a single instruction, in this case `if io.compiled`, and then stops, so even if you run the stopgap interpreter, the `assert` below will never pass.

untraced/yksg/src/lib.rs
ptersilie: I'm just guessing but I think it might be `original frame`.

untraced/yksg/src/lib.rs
ltratt: It's "incoming" in the sense that it's passed to us from a compiled trace. We then convert it to a different (internal) format.

untraced/yksg/src/lib.rs
ltratt: "Other" or "Original" -- I just wanted something that isn't `frame` since that (mostly) means "the current frame".

untraced/yksg/src/lib.rs
ltratt: This is all in the same module, so we don't get much benefit from doing that.

untraced/yksg/src/lib.rs
ltratt: > Is this just to make clearer that fi.body is an Arc

Correct.

traced/tests/src/stopgap/guard_fail.rs
ltratt: Dumb question: which tests in `tests/src/stopgap` do execute the stopgap interpreter?

traced/tests/src/stopgap/guard_fail.rs
ltratt: Aha, I see! OK, I think I can turn those into benchmarks.

traced/tests/src/stopgap/guard_fail.rs
ptersilie: `recursion1` and `recursion2` test that the stopgap interpreter executes. The other tests check the initialisation of the stopgap interpreter (which is kinda hidden in `ct.execute`). There's also `tests/src/stopgap/interp.rs` which have more implementation specific tests of the stopgap interpreter. And then of course there's the tests in `ykrt/src/mt.rs` which test the whole meta-tracing chain, i.e. interpret->compile->guardfail->stopgap->interpret.

untraced/yksg/src/lib.rs
vext01: Yeah, I eventually got there. I'm OK with `IncomingFrame` but prefer `InitFrame`. You guys fight it out :)

ltratt: I'm not a fan of `InitFrame`: it sounds like it's part of this module. I'm not promising that `IncomingFrame` is the best name ever, however!

ptersilie: I'm fine with `IncomingFrame`, at least it's very verbose about what it does. I was hoping to find something snappier, but if we can't it's not the end of the world.

vext01: I see.

In that case, the only other suggestion I have is "FailureFrame", but I'm not sure it's any better than your suggestion.

ltratt:
I've made another stab at a benchmark in cd823ecdd53579ee1a4048206b450a7743a7d7d4. I'm still not sure if it's doing what I expect, and I can't measure any performance difference before/after this PR with it.

ltratt:
953cee6f3f2abc620b5e7338cb553d0c5de6dd19 is probably my final stab at a benchmark. It does do something, but I still can't measure any performance difference. However, at the very least, it provides some sort of metric going forwards. I think we might be ready to squash?

traced/tests/src/stopgap/guard_fail.rs
```diff
@@ -230,7 +230,7 @@ fn bench_recursion(b: &mut Bencher) {
 
     // Test that the SIR interpreter can deal with new recursions after a guard failure.
     fn rec(i: u8) -> u8 {
-        if i < 5 {
+        if i < 100 {
```
ptersilie, line 233: I think this does make little (if not no) difference to the runtime of the stopgap interpreter, since it only ever executes one bytecode before the normal interpreter takes over again. That's probably why you see so little measurable difference. I think one of your previous commits where you had something like
```
for i in 0..1000 {
    si.interpret()
}
```
might be better to measure the stopgap interpreter so it isn't downed out by `ct.execute` calls. But then again, this will only ever test runtime of a single bytecode, and thus isn't even a good representation of the stopgap interpreter in general. I don't immediately have a better idea of testing the stopgap interpreter but I'll think about it in the back of my head.

ptersilie:
Just one more comment, but in lack of a better idea at this moment I'd say we can squash this.

traced/tests/src/stopgap/guard_fail.rs
ltratt: Since `rec` is a recursive call, it effectively is the `for` loop (in disguise) *but* continually calling itself?

ptersilie: I don't think this makes a difference to the runtime of the stopgap interpreter though, since each guard failure kicks us out of the interp_step and back to [here](https://github.com/softdevteam/yk/pull/323/files/847091e5b424c57abf64d88f83991fa14598399b..953cee6f3f2abc620b5e7338cb553d0c5de6dd19#diff-e40b2230bb45a5d0ec3f9ea9a7c503d99052f4da3b2cc179d3fe56415f7e79f8R269).

That being said, I think I talked nonsense above. The stopgap interpreter runs only one bytecode of the user interpreter, but that could involve running multiple MIR bytecodes, so `sg.interpret` tests more than I claimed above.

ltratt: > The stopgap interpreter runs only one bytecode of the user interpreter, but that could involve running multiple MIR bytecodes, so sg.interpret tests more than I claimed above.

Yes, my test is designed to test that (although whether it does so successfully is perhaps another matter!).

ltratt:
OK for me to squash then?

ptersilie:
Please squash.

ltratt:
Squashed.

ltratt:
Weird. Apparently my local `fmt` isn't formatting all the files. I've applied the changes by hand. Let's see if that's enough...

bors try

ltratt:
bors try

ltratt:
@ptersilie OK this has passed `fmt`. I suggest I squash and you retry borsing.

bors try-

ptersilie:
Please squash.

ltratt:
bors try-

ltratt:
Squashed.

PR #326, opened 2021-04-26
Remove unused deps and set hwtracer to the right repo.

vext01:


vext01:
[used `cargo udeps` to find these: https://crates.io/crates/cargo-udeps]

PR #328, opened 2021-04-27
None of the tests currently run but make at least one semi-runnable.

ltratt:
The test this commit fiddles with now compiles (even if it's uncommented!) but fails because ykrt currently throws a `todo!()`. However, it still seems useful to slightly unbreak this, so that when ykrt does work, we can quickly get at
least one test up and running.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -512,660 +512,649 @@ impl MTThreadInner {
 // FIXME These tests were broken during the big code deletion event of 2021 where we moved to LLVM
 // IR. I've left them in (commented), because although they are busted, the ykrt API isn't going to
 // change much and I think we can recover these test at some point.
-//#[cfg(test)]
-//mod tests {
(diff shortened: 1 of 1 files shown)


vext01:
If I see it correctly, you've made a test compile, but left it commented?

Would it be better to uncomment it and mark in `#[ignore]` so that it is at least "compile tested"?

ltratt:
Good idea! Fixed in https://github.com/softdevteam/yk/pull/328/commits/a6f948addfe7b108105c9a94bbffab77876d28aa.

vext01:
Please squash.

ltratt:
It seems that rustfmt has changed a lot (?!) so the `cargo fmt` is a bit brutal https://github.com/softdevteam/yk/pull/328/commits/1df27a33876fd01b948f03e7289d7f8e8dcac51c.

vext01:
please squash.

ltratt:
Squashed.

PR #329, opened 2021-04-28
Start work on the mapper.

vext01:
This change:
 - builds my LLVM fork (required for post-LTO blockmaps).
 - adds an LLVM blockmap parser.
 - tests it from a C program.

The C test infrastructure is rough (for example it isn't run
automatically by cargo and it has no clean target). It may be replaced
by a lang_tester or py.test solution, but this gets us going.

I've had to make a new crate called yktestapi which is responsible for
exposing bits that we want to test from C code (I guess a "ykapi" crate
will follow which exposes only the "real" (non-test) API).

Ideally we'd have had each individual crate export its own interfaces to
C, but this is problematic. ykrt needs to reference yktrace, and any
crate that we want C code to dynamically link needs to be of crate-type
either 'dylib' or 'cdylib', but:

 - Linking >1 "dylib" crate together causes Rust symbol clashing.
   ```
   error: cannot satisfy dependencies so `std` only shows up once
   ```

 - Linking >1 "cdylib" crate toghether causes rust crates to not see
   each other:
   ```
   17 | use yktrace::{CompiledTrace, ThreadTracer};
      |     ^^^^^^^ use of undeclared crate or module `yktrace`
   ```

   I guess the module system can't work between cdylibs, only bare
   function calls via the C ABI.

This of course means that we have to mark more things "public" so that
yktestapi can see them. We can probably live with that.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -23,6 +23,34 @@ export PATH=${CARGO_HOME}/bin/:$PATH
 rustup toolchain install nightly --allow-downgrade --component rustfmt
 
 cargo fmt --all -- --check
+
+# Run the Rust tests.
 cargo test
 cargo test --release
 cargo bench
+
+# Build LLVM for the C tests.
+#
+# This is required because we have an un-upstreamed patch to get the post-LTO
+# blockmap section into the end binaries.
+#
+# Also note that this is a fork of Rust's fork, as we hope to get all of this
+# working for Rust binaries one day. Blocker:
+# https://github.com/rust-lang/rust/issues/84395
+cd target
+git clone -b yk/12.0-2021-04-15 https://github.com/vext01/llvm-project
+cd llvm-project
+mkdir build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=`pwd`/inst \
+    -DLLVM_INSTALL_UTILS=On \
+    -DCMAKE_BUILD_TYPE=release \
+    -DLLVM_ENABLE_ASSERTIONS=On \
+    -DLLVM_ENABLE_PROJECTS="lld;clang" \
+    ../llvm
+make -j `nproc` install
+export PATH=`pwd`/inst/bin:${PATH}
+cd ../../..
+
+# Run the C tests.
+make
--- a/c_tests/blockmap/Makefile
+++ b/c_tests/blockmap/Makefile
@@ -1,0 +1,5 @@
+all: blockmap
+	env LD_LIBRARY_PATH=${SYSROOT_LIBDIR} ./blockmap
+
+blockmap:
+	clang ${CFLAGS} ${LDFLAGS} -o blockmap blockmap.c
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+rustflags = ["-Cprefer-dynamic"]
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,3 @@
+all:
+	${MAKE} -C c_tests test-debug
+	${MAKE} -C c_tests test-release
--- a/c_tests/Makefile
+++ b/c_tests/Makefile
@@ -1,0 +1,29 @@
+LDFLAGS = -Wl,--plugin-opt=-lto-embed-bitcode=optimized -lyktrace
+CFLAGS = -fuse-ld=lld -flto
+
+DEBUG_TARGET_DIR = ../../target/debug
+RELEASE_TARGET_DIR = ../../target/release
+
+# We can't use an rpath for the rust sysroot shared objects as the system rust is found first.
+SYSROOT_LIBDIR = $(shell rustc --print sysroot)/lib
+
+all:
+	@echo "valid targets are:"
+	@echo "  test-release"
+	@echo "  test-debug"
+	exit 1
+
+test-debug:
+	cargo build --features c_testing
+	@for i in `ls -d */`; do \
+		echo "==> Running C test $$i" && ${MAKE} -C $$i \
+		LDFLAGS="${LDFLAGS} -L${DEBUG_TARGET_DIR} -Wl,-rpath=${DEBUG_TARGET_DIR}" \
+		CFLAGS="${CFLAGS}" SYSROOT_LIBDIR=${SYSROOT_LIBDIR}; done
+
+test-release:
+	cargo build --release --features c_testing
+	@for i in `ls -d */`; do \
+		echo "==> Running C test $$i" && ${MAKE} -C $$i \
+		LDFLAGS="${LDFLAGS} -L${RELEASE_TARGET_DIR} -Wl,-rpath=${RELEASE_TARGET_DIR}" \
+		CFLAGS="${CFLAGS}" SYSROOT_LIBDIR=${SYSROOT_LIBDIR}; done
+
--- a/c_tests/blockmap/.gitignore
+++ b/c_tests/blockmap/.gitignore
@@ -1,0 +1,1 @@
+blockmap
--- a/c_tests/blockmap/blockmap.c
+++ b/c_tests/blockmap/blockmap.c
@@ -1,0 +1,19 @@
+// Check the blockmap for this test program contains blocks.
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <assert.h>
+
+// FIXME find a way to auto-generate these protos.
+void *yktrace_hwt_mapper_blockmap_new(void);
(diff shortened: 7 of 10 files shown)


.buildbot.sh
```diff
@@ -23,6 +23,34 @@ export PATH=${CARGO_HOME}/bin/:$PATH
...
+cd llvm-project
+mkdir build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=`pwd`/inst \
+    -DLLVM_INSTALL_UTILS=On \
+    -DCMAKE_BUILD_TYPE=release \
+    -DLLVM_ENABLE_ASSERTIONS=On \
+    -DLLVM_ENABLE_PROJECTS="lld;clang" \
+    ../llvm
+make -j 36 install
```
ltratt, line 51: Let's do:

```
make -j `nproc` install
```

so that we don't hard code the number of cores in.

.buildbot.sh
```diff
@@ -23,6 +23,34 @@ export PATH=${CARGO_HOME}/bin/:$PATH
...
+export PATH=`pwd`/inst/bin:${PATH}
+cd ../../..
+
+# Run the C tests.
+make
```
ltratt, line 56: We can do the `nproc` trick here too.

.buildbot.sh
vext01: Nice, I didn't know about that command!

c_tests/blockmap/Makefile
```diff
@@ -0,0 +1,3 @@
+all:
+	clang ${CFLAGS} ${LDFLAGS} -o blockmap blockmap.c
```
ltratt, line 2: Shouldn't this be:

```make
all: blockmap
    ./blockmap

blockmap: blockmap.c
    clang ${CFLAGS} ${LDFLAGS} -o blockmap blockmap.c
```

I'm not even sure we need the long command line if we set `CC` outside of the Makefile?

vext01: That would prevent recompilation if the binary is already on disk, yeah.

> I'm not even sure we need the long command line if we set CC outside of the Makefile?

I'm not sure what you mean here though? Do you mean rely on an implicit suffix rule to do the compilation step?

ltratt:
This all looks good except:

> This of course means that we have to mark more things "public" so that yktestapi can see them.

Past experience suggests we quickly lose track of *why* things are public, so I'd like to avoid us having to wrench this stuff out again in the future. We could e.g. use a feature flag (`testable` was what we used before) to avoid having blanket-`pub`s.

vext01:
> We could e.g. use a feature flag (testable was what we used before) to avoid having blanket-pubs.

Yeah, also kind of messy, but I can't think of another way off the top of my head.

ltratt:
> Yeah, also kind of messy, but I can't think of another way off the top of my head.

Messy internally, but `pub` commits us to an external API that's messy for other people :) Are there things in this commit that we should put behind a feature flag? [I assume the answer is "yes" but I don't know!]

.buildbot.sh
vext01: It wouldn't have any effect here, as the targets are all sequential.

Bear in mind that I'm likely to try and convert all of the make gubbins to lang_tester soon.

ltratt: OK.

vext01:
> Messy internally, but pub commits us to an external API that's messy for other people

I agree, but it's only partially true in our circumstances, as we will have to explicitly choose what is exported via the C ABI (as I have done on `yktestapi`). Anything `pub` is only public to the Rust ABI, and our system is not be intended to be consumed via the Rust ABI.

bjorn3:
> Linking >1 "dylib" crate together causes Rust symbol clashing.

Compiling with `-Cprefer-dynamic` will likely fix this.

vext01:
> Compiling with -Cprefer-dynamic will likely fix this.

Trying this...

vext01:
It looks like that may work. Let me see if I can get rid of yktestapi :)

vext01:
What's odd is that by turning `ykrt` and `yktrace` into dylibs, the loader rejects `libyktrace.so` when running the test C program. Here's an strace:

```
...
openat(AT_FDCWD, "../../target/debug/x86_64/libc.so.6", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)                                                                                                
openat(AT_FDCWD, "../../target/debug/libc.so.6", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)                                                                                                       
openat(AT_FDCWD, "/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3                                                                                                                                                       
fstat(3, {st_mode=S_IFREG|0644, st_size=63196, ...}) = 0                                                                                                                                                           
mmap(NULL, 63196, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7f3bd0a81000                                                                                                                                                   
close(3)                                = 0                                                                                                                                                                        
openat(AT_FDCWD, "/lib/x86_64-linux-gnu/libc.so.6", O_RDONLY|O_CLOEXEC) = 3                                                                                                                                        
read(3, "\177ELF\2\1\1\3\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0\260A\2\0\0\0\0\0"..., 832) = 832                                                                                                                           
fstat(3, {st_mode=S_IFREG|0755, st_size=1824496, ...}) = 0                                                    
mmap(NULL, 1837056, PROT_READ, MAP_PRIVATE|MAP_DENYWRITE, 3, 0) = 0x7f3bd08c0000                                                                                                                                   
mprotect(0x7f3bd08e2000, 1658880, PROT_NONE) = 0                                                                                                                                                                   
mmap(0x7f3bd08e2000, 1343488, PROT_READ|PROT_EXEC, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x22000) = 0x7f3bd08e2000                                                                                               
mmap(0x7f3bd0a2a000, 311296, PROT_READ, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x16a000) = 0x7f3bd0a2a000                                                                                                         
mmap(0x7f3bd0a77000, 24576, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x1b6000) = 0x7f3bd0a77000                       
mmap(0x7f3bd0a7d000, 14336, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x7f3bd0a7d000                                                                                                     
close(3)                                = 0                                                                                                                                                                        
openat(AT_FDCWD, "/lib/x86_64-linux-gnu/tls/haswell/avx512_1/x86_64/libyktrace.so", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)                                                                    
stat("/lib/x86_64-linux-gnu/tls/haswell/avx512_1/x86_64", 0x7ffd1bd51710) = -1 ENOENT (No such file or directory)                 
openat(AT_FDCWD, "/lib/x86_64-linux-gnu/tls/haswell/avx512_1/libyktrace.so", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)         
...
```

Any ideas?

c_tests/blockmap/Makefile
vext01: Suggest we don't worry about this, as i want to kill the `make`-based system.

That was just to get something going quickly.

vext01:
That last commit should address all outstanding comments.

We do now have to dynamically link the rust std lib. Is that an issue? It'll mean the user has to set a LD_LIBRARY_PATH.

ltratt:
> We do now have to dynamically link the rust std lib. Is that an issue? It'll mean the user has to set a LD_LIBRARY_PATH.

Does that mean they also have to build their own rustc?

vext01:
> Does that mean they also have to build their own rustc?

No, all of the shared objects will come down with their rust install. We just need them to point us to the sysroot. It can be done semi-automatically with `rustc --print sysroot`, but of course if we are talking about existing interpreters then we have to start talking about build systems.

ltratt:
I can live with that, at least for the time being.

Please squash.

vext01:
[I tried to put the sysroot path in an rpath, but the system rustc in /usr/bin is found first and things explode]

vext01:
squashed, but realised i can make something private again. OK to squash again?

ltratt:
Please squash.

vext01:
splat.

PR #330, opened 2021-04-28
Remove the external notion of an MTThread.

ltratt:
Before this commit we required users to create threads through the MT struct, so that we could pass those threads an `MTThread` structure. This works well if you know from the very beginning that you're creating a yk interpreter, but will not work if you e.g. embed a yk regex interpreter inside a wider program that's otherwise ignorant about yk (and may create threads via another API).

The "obvious" solution is to make `MTThread` a thread_local, but if we do that, we also have to make MT a global. And that's more-or-less what this commit does, with the minor tweak that it hides `MTThread` from the outside world.

Users can access the MT struct via `MT::global()` at any point in their program: `MTThread` is a thread local that is used by `MT` internally but doesn't need to be exposed to the user.

This makes the API simpler to use, perhaps slightly icky in some ways (we've all been brought up to believe that globals are bad!), but probably powerful enough to carry us for a decent distance with, hopefully, only minor tweaks.

Internally, this clearly needs work: `MT`s are initialised eagerly but `MTThread`s lazily; and the horrible `with` hack needed to get an `MTThread` reference to `hot_location` is best not mentioned in polite company. Those can, I believe, be finessed in future PRs without changing the external API. Put another way: this commit is as close to "minimal change" as I think I can get.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -6,24 +6,21 @@
 // Rust's POV.
 #![allow(unreachable_code, unused_variables, dead_code)]
 
-#[cfg(test)]
-use std::time::Duration;
 use std::{
+    cell::Cell,
     cmp,
     collections::VecDeque,
-    error::Error,
-    io,
     marker::PhantomData,
     ptr,
-    rc::Rc,
     sync::{
-        atomic::{AtomicBool, AtomicU32, AtomicUsize, Ordering},
+        atomic::{AtomicU32, AtomicUsize, Ordering},
         Arc,
     },
-    thread::{self, JoinHandle},
+    thread,
 };
 
 use num_cpus;
+use once_cell::sync::Lazy;
 use parking_lot::{Condvar, Mutex, MutexGuard};
 use parking_lot_core::SpinWait;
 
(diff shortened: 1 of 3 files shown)


ykrt/src/mt.rs
```diff
@@ -99,6 +48,24 @@ pub struct MT {
 }
 
 impl MT {
+    pub fn global() -> &'static Self {
```
vext01, line 51: I don't understand what this is for.

It returns self in a really convoluted way?

ltratt: Notice: there is no `&self`!

vext01: Is this a singleton in disguise?

ltratt: Yes.

vext01: OK, I think I get it now. I've done this sort of thing with `lazy_static` before.

What's the advantage of `once_cell` over `lazy_static` or even `std::sync::Once`?

vext01: Or `std::lazy::OnceCell` (same thing, just in std?)

ltratt: `sync::Once` is awkward to use; `once_cell` is a layer over `lazy_static`. I wouldn't be surprised if we change this particular aspect at some point, but I don't think it's a big deal either way right now.

vext01: How about `std::lazy::OnceCell`? It's not stabilized, but would mean we don't have to pull any deps in.

ltratt: I'm not sure that an unstable API over a stable external package is a win TBH.

vext01: OK, let's keep an eye on it in the future.

ltratt: Agreed.

vext01:
So the user never sees the `MTThread` even. That's nice.

The `with` "hack" is fine. That's how you access a thread local :P

I suppose there's no meaningful test that can be done?

ltratt:
> I suppose there's no meaningful test that can be done?

Alas, not until the mapper does something.

PR #331, opened 2021-04-29
Convert the C test suite to lang_tester.

vext01:


c_tests/tests/blockmap.c
```diff
@@ -1,3 +1,6 @@
+// Compiler:
```
ltratt, line 1: Shouldn't we at least check for `status: success`? At the moment these tests don't test anything at all.

vext01: Oops. Yes, that was my intention!

vext01: Docs say:

> If not specified, defaults to success.

Assuming that's correct?

vext01: I put an `assert(0)` in the test and it flagged as a failure, so I think the docs are correct.

ltratt: I stand corrected!

vext01:
bors try

vext01:
OK to squash?

ltratt:
Please squash.

vext01:
splat

PR #333, opened 2021-05-04
Add a thin wrapper around LLVMSymbolizer.

vext01:
This allows us to do virtual addresses to symbol name mappings, which is fundamental to mapping PT block addresses to LLVM IR blocks.

This requires calling to C++, so I've added a `ykllvm` crate to contain
LLVM C++ wrappers. I'm sure this won't be the last.

[I notice that in the meantime @ltratt has bagsied the name `ykllvm` for a possible llvm fork, so perhaps we want to rename the new crate for C++ LLVM wrappers here? `ykcpp` or `ykll`?]

### DIFF
--- a/ykllvmwrap/src/symbolizer.rs
+++ b/ykllvmwrap/src/symbolizer.rs
@@ -1,0 +1,97 @@
+//! A wrapper around llvm::symbolizer::LLVMSymbolizer.
+
+use libc::free;
+use std::{
+    ffi::{c_void, CStr},
+    ops::Drop,
+    os::raw::c_char,
+};
+
+extern "C" {
+    fn __yk_symbolizer_new() -> *mut c_void;
+    fn __yk_symbolizer_free(symbolizer: *mut c_void);
+    fn __yk_symbolizer_find_code_sym(symbolizer: *mut c_void, addr: usize) -> *mut c_char;
+}
+
+pub struct Symbolizer(*mut c_void);
+
+impl Symbolizer {
+    pub fn new() -> Self {
+        Self(unsafe { __yk_symbolizer_new() })
+    }
+
+    /// Returns the name of the symbol at the provided virtual address.
+    pub fn find_code_sym(&self, vaddr: usize) -> Option<String> {
+        let ptr = unsafe { __yk_symbolizer_find_code_sym(self.0, vaddr) };
+        if ptr.is_null() {
+            None
+        } else {
+            let ret = {
+                let sym = unsafe { CStr::from_ptr(ptr) };
+                String::from(sym.to_str().unwrap())
+            };
+            unsafe { free(ptr as *mut _) };
+            Some(ret)
+        }
+    }
+}
+
+impl Drop for Symbolizer {
+    fn drop(&mut self) {
+        unsafe { __yk_symbolizer_free(self.0) }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::Symbolizer;
+
+    extern "C" {
+        fn getuid();
+    }
+
+    #[inline(never)]
+    fn symbolize_me_mangled() {}
+
+    // This function has a different signature to the one above to prevent LLVM from merging the
+    // functions (and their symbols) when optimising in --release mode.
+    #[inline(never)]
+    #[no_mangle]
+    fn symbolize_me_unmangled() -> u8 {
+        1
+    }
+
+    #[test]
+    fn find_code_sym_mangled() {
+        let f_addr = symbolize_me_mangled as *const fn() as *const _ as usize;
+        let s = Symbolizer::new();
+        let sym = s.find_code_sym(f_addr).unwrap();
+        // The symbol will be suffixed with an auto-generated module name, e.g.:
+        // ykllvmwrap::symbolizer::tests::symbolize_me_mangled::hc7a76ddceae6f9c4
+        assert!(sym.starts_with("ykllvmwrap::symbolizer::tests::symbolize_me_mangled::"));
+        let elems = sym.split("::");
+        assert_eq!(elems.count(), 5);
+    }
+
+    #[test]
+    fn find_code_sym_unmangled() {
+        let f_addr = symbolize_me_unmangled as *const fn() as *const _ as usize;
+        let s = Symbolizer::new();
+        let sym = s.find_code_sym(f_addr).unwrap();
+        assert_eq!(sym, "symbolize_me_unmangled");
+    }
+
+    #[test]
+    fn find_code_sym_libc() {
+        let f_addr = getuid as *const fn() as *const _ as usize;
+        let s = Symbolizer::new();
+        let sym = s.find_code_sym(f_addr).unwrap();
+        assert_eq!(sym, "getuid");
+    }
+
+    #[test]
+    fn find_code_sym_not_found() {
+        // Virtual address 0x1 is obviously nonsense.
+        assert!(Symbolizer::new().find_code_sym(1).is_none());
+    }
+}
--- a/ykllvmwrap/src/ykllvmwrap.cc
+++ b/ykllvmwrap/src/ykllvmwrap.cc
@@ -1,0 +1,46 @@
+// LLVM-related C++ code wrapped in the C ABI for calling from Rust.
+
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+
+#include <dlfcn.h>
+#include <llvm/DebugInfo/Symbolize/Symbolize.h>
+#include <string.h>
+
+using namespace llvm;
+using namespace llvm::symbolize;
+
+extern "C" LLVMSymbolizer *__yk_symbolizer_new() {
+    return new LLVMSymbolizer;
+}
+
+extern "C" void __yk_symbolizer_free(LLVMSymbolizer *Symbolizer) {
+    delete Symbolizer;
+}
+
+// Finds the name of a code symbol from a virtual address.
+// The caller is responsible for freeing the returned (heap-allocated) C string.
+extern "C" char *__yk_symbolizer_find_code_sym(LLVMSymbolizer *Symbolizer, void *Vaddr) {
+    // Find which of the loaded objects holds this virtual address.
+    Dl_info Info;
+    if (dladdr(Vaddr, &Info) == 0 ) {
+        return NULL;
+    }
+
+    // Find the corresponding offset of vaddr from the start of the object.
+    auto Offs = (uintptr_t) Vaddr - (uintptr_t) Info.dli_fbase;
+
+    // Use the LLVM symbolizer to find the symbol name. Note that we don't
+    // simply rely on `Info.dli_sname`, as `dl_addr()` can only find a subset
+    // of symbols -- those exported -- and we don't want users to have to link
+    // with --export-dynamic.
+    auto LineInfo = Symbolizer->symbolizeCode(Info.dli_fname,
+        {Offs, object::SectionedAddress::UndefSection});
+    if (auto err = LineInfo.takeError()) {
+        return NULL;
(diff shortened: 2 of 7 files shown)


vext01:
Typo in the commit message that I will fix at the end.

ykllvm/src/ykllvm.cc
```diff
@@ -0,0 +1,41 @@
...
+
+// Finds the name of a code symbol from a virtual address.
+extern "C" char *ykllvm_symbolzer_find_code_sym(LLVMSymbolizer *Symbolizer, void *Vaddr) {
+	// Find which of the loaded objects holds this virtual address.
+    Dl_info Info;
+    if (dladdr(Vaddr, &Info) == 0 ) {
+        return NULL;
+    }
+
+	// Find the corresponding offset of vaddr from the start of the object.
```
ltratt, line 30: Indent broken?

ykllvm/src/ykllvm.cc
```diff
@@ -0,0 +1,41 @@
...
+    auto Offs = (uintptr_t) Vaddr - (uintptr_t) Info.dli_fbase;
+
+    // Use the LLVM symbolizer to find the symbol name. Note that we don't
+    // simply rely on `Info.dli_sname`, as `dl_addr()` can only find a subset
+    // of symbols -- those exported -- and we don't want users to have to link
+    // with --export-dynamic.
+    auto LineInfo = Symbolizer->symbolizeCode(Info.dli_fname,
+        {Offs, object::SectionedAddress::UndefSection});
+
+    return strdup(LineInfo->FunctionName.c_str());
```
ltratt, line 40: Do we really want to `strdup` here? That means we're allocating memory for something that I presume is stored in the static section of the binary?

ltratt:
> This requires calling to C++, so I've added a ykllvm crate to contain LLVM C++ wrappers. I'm sure this won't be the last.

I see why the crate is `ykllvm` now. I suppose it might be called `ykllvmhelper` or something like that, but then it becomes rather a mouthful... I don't have any good ideas!

ykllvm/src/ykllvm.cc
vext01: Yeah, maybe. LLVM doesn't document the lifetime of the returned pointer. Let me read the source and see if I can figure out what it does.

vext01: I'm not having much luck figuring this out. The details are hidden *somewhere* amongst the C++ :P

My worry is that the field we are copying bytes out of is a `std::String`, which is a bit like an owned string for C++. It's possible that the underlying string data is dropped along with the C++ frame, so taking a pointer may not be valid.

One thing we could do is wrap `std::String` (providing an accessor too) and return the string itself.

Thoughts?

ltratt: My dumb suggestion is to print out the address of one `FunctionName` and see if it's in the static part of the binary or not. If it isn't, then clearly the answer is "we need `strdup`". If it *is* in the static part, then I suggest we *assume* that `strdup` isn't needed, leave a comment in the source code to indicate our assumption, then plough on. [Amongst the reasons not to use `strdup` is that this function's API now requires the caller to `free` the string memory, which we would have to document etc etc.]

vext01: > Amongst the reasons not to use strdup is that this function's API now requires the caller to free

Which I've also forgotten to do :P

ltratt: > Which I've also forgotten to do :P

Exactly -- that was what prompted my question in the first place :)

vext01: > print out the address of one FunctionName and see if it's in the static part of the binary or not

The pointer (`0x7ffc817ed2a0`) falls under:
```
 80 7ffc817cf000-7ffc817f1000 rw-p 00000000 00:00 0                          [stack]
```

Which I assume is not static.

This fits with my experience. Usually a pointer beginning `0x7f` is stack.

ltratt: You're suggest it's copying a variable length string to the stack? That doesn't seem very likely?

vext01: Another option is to reserve (say) 256 bytes (or a `MAX_SYM` if one exists) on the parent stack and pass a pointer which we copy into.

vext01: > That doesn't seem very likely?

I don't know what else to say. Yes?

ltratt: I can't see how a function can return a pointer to a variable sized stack unless we passed a pointer to that part of the stack in. Maybe I'm missing out on something!

vext01: To use an analogy in Rust, we have:
```
struct X {
    s: String,
}
...
fn f() {
    x: X = some_call();
}
```

Unless there are pointer gymnastics going on, the string in `x` dies with `f()`.

ltratt: A Rust `String` is a constant sized `struct` pointing to variable sized backing heap (for the bytes of the string).

vext01: Maybe my method was wrong. I printed the pointer address, suspended the process and looked in `/proc/pid/maps`.

Is that what you had in mind?

ltratt: If this was Rust, I think you're printing the address of the *outer* `String`, not the inner bytes of actual text.

vext01: I'm not sure if `c_str()` has to allocate a new buffer so that it can append the NUL terminator.

Will dig more after lunch.

vext01: I think I've gotten to the bottom of this.

Firstly, `c_str()` doesn't specify exactly how the pointer to the char array is derived, but from what I can extract from informal conversations found onthe internet, `std::string` often stores the string nul terminated so that `c_str()` is a matter of returning a pointer to the backing store. Makes sense.

Next, [here's the bit of LLVM code](https://llvm.org/doxygen/SymbolizableObjectFile_8cpp_source.html#l00240) that populates the `FunctionName` member in the `LineInfo` struct that I was copying from. The code assigns to a `&std::string` and `std::string` is really a `basic_string`. So [that means it's a string copy](https://en.cppreference.com/w/cpp/string/basic_string/operator%3D) (first case in the table).

Regardless of how the backing storage is allocated, the string's destructor ["destroys the string, deallocating internal storage if used"](https://en.cppreference.com/w/cpp/string/basic_string).

Further, there's [nothing special about how the parent `LineInfo` struct is allocated](https://code.woboq.org/llvm/llvm/lib/DebugInfo/DWARF/DWARFContext.cpp.html#1176) -- plain stack allocation (I've just found that code search website -- should come in very useful!).

So if my reasoning is correct, all of this means that we can't hold on to our `char *` pointer past the end of our C++ wrapper.

ltratt: OK. Should we maybe give ourselves the ability to return a potentially cached / ref counted string so that we can later do a lookup rather than malloc each time or should we worry about that later? If you just want to document this as "the caller is responsible for freeing the string", I'm fine with that.

vext01: > I think you're printing the address of the outer String, not the inner bytes of actual text.

Nah:

```
 43     auto ptr = LineInfo->FunctionName.c_str();                                                                                                                                                                  
 44     printf("%p\n", ptr);                   
```

vext01: > If you just want to document this as "the caller is responsible for freeing the string", I'm fine with that.

Looking at how my past self used the returned pointer (Friday was a long time ago apparently, and I've forgotten already), I pass it to [`CString::new()`](https://doc.rust-lang.org/std/ffi/struct.CString.html) which:

> will consume the provided data and use the underlying bytes to construct a new string, ensuring that there is a trailing 0 byte.

So that sounds like it will copy the data into a new buffer discarding (and in this case leaking) the old buffer? I'm not clear on this.

Is there a way of "moving" the `strdup()`ed buffer into the backing store of a `CString`? Then it would be automagically deallocated.

vext01: I have an idea. Leave it with me.

vext01:
I think that's all comments addressed.

ykllvmwrap/src/symbolizer.rs
```diff
@@ -8,21 +8,21 @@ use std::{
 };
 
 extern "C" {
-    fn ykllvm_symbolzer_new() -> *mut c_void;
-    fn ykllvm_symbolzer_free(symbolizer: *mut c_void);
-    fn ykllvm_symbolzer_find_code_sym(symbolizer: *mut c_void, addr: usize) -> *mut c_char;
+    fn ykllvmwrap_symbolzer_new() -> *mut c_void;
```
ltratt, line 11: I assume this is meant to be "symbolizer" with a full fat "i"?

Also, I'm not sure whether we need to use the whole prefix here: I'd be OK with just `yk_` as a prefix (we can trust ourselves to keep these unique across the yk packages) if you'd prefer it.

ykllvmwrap/src/symbolizer.rs
```diff
@@ -68,8 +68,8 @@ mod tests {
         let sym = s.find_code_sym(f_addr).unwrap();
         //let sym_str = sym.to_str().unwrap();
         // The symbol will be suffixed with an identifier for the tests module, e.g.:
-        // ykllvm::symbolizer::tests::symbolize_me_mangled::hc7a76ddceae6f9c4
-        assert!(sym.starts_with("ykllvm::symbolizer::tests::symbolize_me_mangled::"));
+        // ykllvmwrap::symbolizer::tests::symbolize_me_mangled::hc7a76ddceae6f9c4
```
ltratt, line 70: I'm not sure I understand the comment.

ykllvmwrap/src/symbolizer.rs
vext01: I prefer to "namespace" them so that they can't be confused with our API.

I'd be OK with `__yk_` though. That makes is pretty clear it's internal.

ltratt: Fine with me.

ykllvmwrap/src/symbolizer.rs
vext01: What I'm trying to convry is that the `hc7a76ddceae6f9c4` junk is an auto-generated module name out of our control.

ltratt: Ah, I get it. Ignore me -- I was being an idiot!

vext01: I'll reword it a little nonetheless.

vext01:
Hopefully that's everything now.

ltratt:
Please squash.

vext01:
squashed and updated commit msg.

vext01:
bors try

vext01:
@ltratt OK to squash?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -37,6 +37,9 @@ extern "C" char *__yk_symbolizer_find_code_sym(LLVMSymbolizer *Symbolizer, void
     // with --export-dynamic.
     auto LineInfo = Symbolizer->symbolizeCode(Info.dli_fname,
         {Offs, object::SectionedAddress::UndefSection});
+	if (auto err = LineInfo.takeError()) {
```
ltratt, line 40: The indentation seems off here?

vext01: Fixed in 
0deee82

ltratt:
Please squash.

vext01:
splat.

PR #334, opened 2021-05-07
Allow C code to interface, at a basic level, to code in ykrt.

ltratt:
This is crude, and will need substantial expanding (and probably tweaking), but it's a start.

### DIFF
--- a/c_tests/run.rs
+++ b/c_tests/run.rs
@@ -12,22 +12,29 @@ const COMMENT: &str = "//";
 /// Make a compiler command that compiles `src` to `exe`.
 fn mk_compiler(exe: &Path, src: &Path) -> Command {
     let mut compiler = Command::new("clang");
+
     let mut lib_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
     lib_dir.push("..");
     lib_dir.push("target");
-
     #[cfg(cargo_profile = "debug")]
     lib_dir.push("debug");
     #[cfg(cargo_profile = "release")]
     lib_dir.push("release");
 
+    let mut ykrt_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+    ykrt_dir.push("..");
+    ykrt_dir.push("ykrt");
+
     compiler.args(&[
         "-fuse-ld=lld",
         "-flto",
         "-Wl,--plugin-opt=-lto-embed-bitcode=optimized",
+        "-I",
+        ykrt_dir.to_str().unwrap(),
         "-L",
         lib_dir.to_str().unwrap(),
         "-lyktrace",
+        "-lykrt",
         "-o",
         exe.to_str().unwrap(),
         src.to_str().unwrap(),
--- a/ykrt/ykrt.h
+++ b/ykrt/ykrt.h
@@ -1,0 +1,55 @@
+#include <stdint.h>
+
+/// A meta-tracer.
+struct MT;
+typedef struct MT MT;
+
+// A `Location` stores state that the meta-tracer needs to identify hot loops
+// and run associated machine code.
+//
+// Each position in the end user's program that may be a control point (i.e.
+// the possible start of a trace) must have an associated `Location`. The
+// `Location` does not need to be at a stable address in memory and can be
+// freely moved.
+//
+// Program positions that can't be control points don't need an associated
+// `Location`. For interpreters that can't (or don't want) to be as selective,
+// a simple (if moderately wasteful) mechanism is for every bytecode or AST
+// node to have its own `Location` (even for bytecodes or nodes that can't be
+// control points).
+typedef struct {
+    uintptr_t state;
+} Location;
+
+#if defined(__x86_64)
+typedef uint32_t HotThreshold;
+#else
+#error Unable to determine type of HotThreshold
+#endif
+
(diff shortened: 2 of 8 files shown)


c_tests/run.rs
```diff
@@ -12,22 +12,29 @@ const COMMENT: &str = "//";
 /// Make a compiler command that compiles `src` to `exe`.
 fn mk_compiler(exe: &Path, src: &Path) -> Command {
     let mut compiler = Command::new("clang");
+
```
vext01, line 15: Some unrelated whitespace shuffling here.

ykrt/ykrt.h
```diff
@@ -0,0 +1,55 @@
...
+// Program positions that can't be control points don't need an associated
+// `Location`. For interpreters that can't (or don't want) to be as selective,
+// a simple (if moderately wasteful) mechanism is for every bytecode or AST
+// node to have its own `Location` (even for bytecodes or nodes that can't be
+// control points).
+typedef struct {
+    uintptr_t state;
+} Location;
+
+#if defined(__x86_64)
```
vext01, line 24: Would it be better to do the test based on pointer size rather than platform?

ykrt/ykrt.h
```diff
@@ -0,0 +1,55 @@
...
+
+// Return a reference to the global `MT` instance: at any point, there is at
+// most one of these per process and an instance will be created if it does not
+// already exist.
+MT *yk_mt();
+
+// Return this `MT` instance's current hot threshold. Notice that this value
+// can be changed by other threads and is thus potentially stale as soon as it
+// is read.
+HotThreshold yk_mt_hot_threshold(MT *);
```
vext01, line 38: Is it important to expose the current hot threshold?

ykrt/ykrt.h
```diff
@@ -0,0 +1,55 @@
...
+//
+// Note that a `Location` created by this call must not simply be discarded: if
+// no longer wanted, it must be passed to `yk_drop_location` to allow
+// appropriate clean-up.
+Location yk_new_location();
+
+// Clean-up a `Location` previously created by `yk_new_location`. The
+// `Location` must not be further used after this call or undefined behaviour
+// will occur.
+void yk_drop_location(Location *);
```
vext01, line 55: The convention we've been using to free things from C is `free` (not `drop`). We should probably be consistent.

vext01:
Mostly LGTM. Some of the changes are unrelated and they'd have been better suited to separate commits. I won't make you unpick it all though :P

c_tests/run.rs
ltratt: Agreed.

ykrt/ykrt.h
```diff
@@ -0,0 +1,55 @@
...
+// `Location` does not need to be at a stable address in memory and can be
+// freely moved.
+//
+// Program positions that can't be control points don't need an associated
+// `Location`. For interpreters that can't (or don't want) to be as selective,
+// a simple (if moderately wasteful) mechanism is for every bytecode or AST
+// node to have its own `Location` (even for bytecodes or nodes that can't be
+// control points).
+typedef struct {
+    uintptr_t state;
```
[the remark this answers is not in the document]
ltratt: Rust, unfortunately, conflates "the size of an int" with "the size of a pointer". Modern C differentiates them as `size_t` and `uintptr_t`: since `Location`s encode a pointer, the latter is the right thing for us.

ykrt/ykrt.h
ltratt: I couldn't work out how to do that portably :/ This was my hack on the basis that it fails fast: if `__X86_64` isn't defined you get a compile-time error.

ykrt/ykrt.h
ltratt: Yes and no: it's not *vital* but without it you can't write tests which exceed the hot threshold reliably, so it is fairly useful.

ykrt/ykrt.h
ltratt: This isn't `free`ing the memory being pointed to though: `free` and `drop` are subtly different IMHO.

vext01: I see it as the same thing just in different languages terminology.

I'm OK with drop if we go and change all of the other instances in the repo.

ykrt/ykrt.h
vext01: ok

ykrt/ykrt.h
vext01: Lets move on and come back to it.

ykrt/ykrt.h
vext01: You're right: https://scaryreasoner.wordpress.com/2009/02/28/checking-sizeof-at-compile-time/

I guess we leave it as-is then.

ykrt/ykrt.h
ltratt: Agreed.

ykrt/ykrt.h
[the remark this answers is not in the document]
ltratt: > I thought that Location may encode a pointer.

Yes... but I'm not sure what deeper question you're asking?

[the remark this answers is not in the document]
vext01: Yeah, I was confused.

PR #335, opened 2021-05-11
Don't use dynamic libraries everywhere.

vext01:
In a feature branch I've recently hit this error:
```
error: cannot satisfy dependencies so `libc` only shows up once
  |
  = help: having upstream crates all available in one format will likely make this go away
```

What the error means to say is that two unrelated crates in our
workspace depend on the `libc` crate, and this would cause code
duplication and ultimately symbol clashing at link time (had the
compiler not caught this earlier).

One solution is to make one crate "responsible" for importing libc, and
then other crates would import it via a mere `extern crate`. Whilst
this would work, it would contort the repo into an odd structure, and
further, we may have to do this dance later for other deps other than
`libc`.

A more palatable solution is to use a dedicated `cdylib` crate for
exporting to C and then all other crates are regular `rlibs`. This way
there can be no duplicate deps.

This change therefore introduces `ykapi` as our sole `cdylib` and
reverts all other workspace members to the default crate type. We also
remove our cargo config which gave preference to dynamic libs.

Thank you @bjorn3 and @jonas-schievink for the Zulip discussion on this.
I love that the Rust community is open to these kinds of informal
discussions.

**Additional to commit message:** I implemented this in my mapper branch to check that it solves my problem (it does) and then back-ported this to `master`. The only thing I'm not certain on here is the name `ykapi`. The name of this crate dictates the name of the library that interpreters will need to link (i.e. `cc -lykapi ...`). Arguably the crate should just be called `yk` so that we have `-lyk`, but I thought it may be confusing to have a crate called `yk` in our `yk` workspace. FWIW, I *did* call the header `yk.h` not `ykapi.h`. Thoughts?

For transparency, Zulip discussion [here](
https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/cannot.20satisfy.20dependencies.20so.20.60libc.60.20only.20shows.20up.20once).


vext01:
I've spotted a typo in the commit message that I'll need to fix before merge.

vext01:
> ykapi is super generic: API to whom and what? I wonder if ykext or something like that might make it clearer that this is the external world's interface to yk.

`ykext` is even worse, as it would mean to build a C interpreter the user would have to do `cc -lykext ...` which makes little sense from the POV of the interpreter author. The name of this crate will dictate the name of the shared object. Did you see the bit in the PR description about possibly using just `yk`?

> I don't think a single header is a good idea

There is actually a second `yk_testing.h` but it appears I forgot to `git add` it. Sorry, my bad. See bf496e3.

> For a good first approximation, each Rust crate could have its own header file?

I don't expect the interpreter author to have to use anything other than symbols from `ykrt` and supposing they did, why burden them with having to know which internal crate the function they want comes from. We should make it as easy as possible for the user, don't you agree?

vext01:
How about call the directory `ykextapi` (if you dislike `ykcapi`) and the crate `yk`, then later if/when Cargo can produce a `.so` with a different name then it's a no-op for consumers.

I'd rather we incur the nit, rather than palming it off to (possibly external, non-softdev) consumers.

ltratt:
`ykcapi` works for me. I think I'd rather keep things simple in the repo: to be honest, not many people are going to have a problem editing their Makefile from `ykcapi.so` to `yk.so` if/when we're able to actually do so (and there's no guarantee that Rust will ever allow us to do so!).

vext01:
OK. `ykcapi` makes sense in the context of both us and a consumer, so I can live with that.

vext01:
Let's keep it `yk.h`. It's a C header, so it's obvious that it's a C API :P

vext01:
I think we are ready to squash?

vext01:
Squashed, fixed typo and renamed crate name in commit msg.

vext01:
Odd...

> fatal: Could not parse object 'a6d477ca73f3da9dae50b6ab56eabf752cecb5b9'.

Looks like an intermittent git issue? bors hasn't realised, so I'll try to cancel the job.

bors r-

vext01:
Try again?

vext01:
bors try

vext01:
Looks like it will work if we try once more.

I'm wondering if the OOM killer has been doing its dirty work. One of my mosh sessions also died suspiciously...

vext01:
Oops. My apologies. I didn't realise that was happening. If I have to rebase again, I will remove the `@` to hopefully stop the madness.

vext01:
bors try

vext01:
I've just killed the @ mentions so as to eliminate the possibility of any further annoying pings.

The last build worked. What I did was change the commit message slightly and use a newer version of git to do a force push. I think we hit a git bug that corrupted the commit(?).

So let's pray that a r+ will now work @ltratt 🙏

vext01:
The squashed commit seemed to have been corrupted, so I needed to replace it with a commit pushed from a newer git, but git won't push if the hash is identical. So I made a tiny change to the commit message wording to have it push.

ltratt:
OK. Let's check before we do force pushes next time though, even for a case like this -- it just keeps the process simpler.

bors r+

PR #336, opened 2021-05-12
Add the ability to map PT blocks to LLVM IR blocks.

vext01:
This change allows us to translate the block addresses found in a PT
trace into block identifiers in LLVM IR.

We don't yet look up the IR block in the LLVM bitcode, only collect the
information necessary to do so.

In doing so I've also created a little crate called `ykutil` which
currently only shares a routine between both `ykllvmwrap` and `yktrace`
in a sensible and dependency-cycle-free manner.

Observations:
 - LLVM blocks are not terminated by calls.
 - One PT block may map to many LLVM blocks (see doc string on
   HWTMapper::map_block() for more info).

### DIFF
--- a/c_tests/tests/mapper_funcs_branches.c
+++ b/c_tests/tests/mapper_funcs_branches.c
@@ -1,0 +1,51 @@
+// Compiler:
+// Run-time:
+
+// Check that inter-procedural tracing works.
+
+#include <assert.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk_testing.h>
+
+__attribute__((noinline))
+int
+add_one_or_two(int arg)
+{
+    if (arg % 2 == 0) {
+        arg++;
+    } else {
+        arg += 2;
+    }
+    return arg;
+}
+
+int
+main(int argc, char **argv)
+{
+    void *tt = __yktrace_start_tracing(HW_TRACING);
+    argc = add_one_or_two(argc);
+    void *tr = __yktrace_stop_tracing(tt);
+
+    size_t len = __yktrace_irtrace_len(tr);
+    assert(len >= 4); // At least one block for `main`, at least 3 blocks for `add_one_or_two`.
+    for (int i = 0; i < len; i++) {
+        char *func_name = NULL;
+        size_t bb = 0;
+        __yktrace_irtrace_get(tr, i, &func_name, &bb);
+
+        // We expect blocks from `add_one_or_two` to be book-ended by `main` bb0.
+        // Note that in LLVM, a call does not terminate a block.
+        if ((i == 0) || (i == len - 1)) {
+            assert((strcmp(func_name, "main") == 0) && (bb == 0));
+        } else {
+            assert(strcmp(func_name, "add_one_or_two") == 0);
+        }
+        free(func_name);
+    }
+
+    __yktrace_drop_irtrace(tr);
+    return (EXIT_SUCCESS);
+}
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -42,7 +42,11 @@ pub extern "C" fn yk_drop_location(loc: *mut Location) {
 /// These symbols are not shipped as part of the main API.
 #[cfg(feature = "c_testing")]
 mod c_testing {
-    use yktrace::BlockMap;
+    use std::{ffi::CString, mem, os::raw::c_char};
+    use yktrace::{start_tracing, BlockMap, IRTrace, ThreadTracer, TracingKind};
+
+    const SW_TRACING: usize = 0;
+    const HW_TRACING: usize = 1;
 
     #[no_mangle]
     pub extern "C" fn __yktrace_hwt_mapper_blockmap_new() -> *mut BlockMap {
@@ -58,7 +62,55 @@ mod c_testing {
     pub extern "C" fn __yktrace_hwt_mapper_blockmap_len(bm: *mut BlockMap) -> usize {
         let bm = unsafe { Box::from_raw(bm) };
         let ret = bm.len();
-        Box::leak(bm);
+        mem::forget(bm);
         ret
+    }
+
+    #[no_mangle]
+    pub extern "C" fn __yktrace_start_tracing(kind: usize) -> *mut ThreadTracer {
+        let kind: TracingKind = match kind {
+            SW_TRACING => TracingKind::SoftwareTracing,
+            HW_TRACING => TracingKind::HardwareTracing,
+            _ => panic!(),
+        };
+        Box::into_raw(Box::new(start_tracing(kind)))
+    }
+
+    #[no_mangle]
+    pub extern "C" fn __yktrace_stop_tracing(tt: *mut ThreadTracer) -> *mut IRTrace {
+        let tt = unsafe { Box::from_raw(tt) };
+        Box::into_raw(Box::new(tt.stop_tracing().unwrap())) as *mut _
+    }
+
+    #[no_mangle]
+    pub extern "C" fn __yktrace_irtrace_len(trace: *mut IRTrace) -> usize {
+        let trace = unsafe { Box::from_raw(trace) };
+        let ret = trace.len();
+        mem::forget(trace);
+        ret
+    }
+
+    /// Fetches the function name (`res_func`) and the block index (`res_bb`) at position `idx` in
+    /// `trace`. The caller must free() the function name.
+    #[no_mangle]
+    pub extern "C" fn __yktrace_irtrace_get(
+        trace: *mut IRTrace,
+        idx: usize,
+        res_func: *mut *mut c_char,
+        res_bb: *mut usize,
+    ) {
+        let trace = unsafe { Box::from_raw(trace) };
+        let blk = trace.get(idx).unwrap();
+        let c_func = CString::new(blk.func_name()).unwrap();
+        unsafe {
+            *res_func = c_func.into_raw();
+            *res_bb = blk.bb();
+        }
+        mem::forget(trace);
+    }
+
+    #[no_mangle]
+    pub extern "C" fn __yktrace_drop_irtrace(trace: *mut IRTrace) {
+        unsafe { Box::from_raw(trace) };
     }
 }
--- a/ykllvmwrap/src/symbolizer.rs
+++ b/ykllvmwrap/src/symbolizer.rs
@@ -10,7 +10,11 @@ use std::{
 extern "C" {
     fn __yk_symbolizer_new() -> *mut c_void;
     fn __yk_symbolizer_free(symbolizer: *mut c_void);
-    fn __yk_symbolizer_find_code_sym(symbolizer: *mut c_void, addr: usize) -> *mut c_char;
+    fn __yk_symbolizer_find_code_sym(
+        symbolizer: *mut c_void,
+        obj: *const c_char,
+        off: u64,
+    ) -> *mut c_char;
 }
 
 pub struct Symbolizer(*mut c_void);
@@ -20,15 +24,20 @@ impl Symbolizer {
         Self(unsafe { __yk_symbolizer_new() })
     }
 
-    /// Returns the name of the symbol at the provided virtual address.
-    pub fn find_code_sym(&self, vaddr: usize) -> Option<String> {
-        let ptr = unsafe { __yk_symbolizer_find_code_sym(self.0, vaddr) };
+    /// Returns the name of the symbol at byte offset `off` in the object file `obj`,
+    /// or `None` if the symbol couldn't be found.
+    pub fn find_code_sym(&self, obj: &CStr, off: u64) -> Option<String> {
+        let ptr = unsafe { __yk_symbolizer_find_code_sym(self.0, obj.as_ptr(), off) };
         if ptr.is_null() {
             None
         } else {
             let ret = {
                 let sym = unsafe { CStr::from_ptr(ptr) };
-                String::from(sym.to_str().unwrap())
+                let sym = String::from(sym.to_str().unwrap());
+                if sym == "<invalid>" {
+                    return None;
+                }
+                sym
             };
             unsafe { free(ptr as *mut _) };
             Some(ret)
@@ -45,6 +54,7 @@ impl Drop for Symbolizer {
 #[cfg(test)]
 mod tests {
     use super::Symbolizer;
+    use ykutil::addr::code_vaddr_to_off;
 
     extern "C" {
         fn getuid();
@@ -63,9 +73,10 @@ mod tests {
 
     #[test]
     fn find_code_sym_mangled() {
-        let f_addr = symbolize_me_mangled as *const fn() as *const _ as usize;
+        let f_vaddr = symbolize_me_mangled as *const fn() as *const _ as usize;
+        let (obj, f_off) = code_vaddr_to_off(f_vaddr).unwrap();
         let s = Symbolizer::new();
-        let sym = s.find_code_sym(f_addr).unwrap();
+        let sym = s.find_code_sym(obj, f_off).unwrap();
         // The symbol will be suffixed with an auto-generated module name, e.g.:
         // ykllvmwrap::symbolizer::tests::symbolize_me_mangled::hc7a76ddceae6f9c4
         assert!(sym.starts_with("ykllvmwrap::symbolizer::tests::symbolize_me_mangled::"));
@@ -75,23 +86,19 @@ mod tests {
 
     #[test]
     fn find_code_sym_unmangled() {
-        let f_addr = symbolize_me_unmangled as *const fn() as *const _ as usize;
+        let f_vaddr = symbolize_me_unmangled as *const fn() as *const _ as usize;
+        let (obj, f_off) = code_vaddr_to_off(f_vaddr).unwrap();
         let s = Symbolizer::new();
-        let sym = s.find_code_sym(f_addr).unwrap();
+        let sym = s.find_code_sym(obj, f_off).unwrap();
         assert_eq!(sym, "symbolize_me_unmangled");
     }
 
     #[test]
     fn find_code_sym_libc() {
-        let f_addr = getuid as *const fn() as *const _ as usize;
+        let f_vaddr = getuid as *const fn() as *const _ as usize;
+        let (obj, f_off) = code_vaddr_to_off(f_vaddr).unwrap();
         let s = Symbolizer::new();
-        let sym = s.find_code_sym(f_addr).unwrap();
+        let sym = s.find_code_sym(obj, f_off).unwrap();
         assert_eq!(sym, "getuid");
-    }
-
-    #[test]
-    fn find_code_sym_not_found() {
-        // Virtual address 0x1 is obviously nonsense.
-        assert!(Symbolizer::new().find_code_sym(1).is_none());
     }
 }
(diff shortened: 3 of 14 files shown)


vext01:
> In doing so I've also created a little crate called `ykutil` which currently only shares a routine between both `ykllvmwrap` and `yktrace` in a sensible and dependency-cycle-free manner.

To expand on this, the routine that needs to be shared maps virtual addresses to file offsets. It's needed by both the `ykllvmwap` tests and `yktrace`.

It would work to put the routine in `ykllvmwrap`, although the routine has nothing to do with llvm. On the other hand, we could put it in `yktrace`, but then we have a circular dep: `ykllvmwrap <-> yktrace`. Hence `ykutil` is born.

ltratt:
I'm not a big fan of `ykutil` as a name because it's such a generic term: "what utils and for whom?" is always my question. If it's just testing utils `yktestutils` maybe?

ykcapi/src/lib.rs
```diff
@@ -61,4 +66,52 @@ mod c_testing {
         Box::leak(bm);
         ret
     }
+
+    #[no_mangle]
+    pub extern "C" fn __yktrace_start_tracing(kind: usize) -> *mut ThreadTracer {
+        let kind: TracingKind = match kind {
+            SW_TRACING => TracingKind::SoftwareTracing,
+            HW_TRACING => TracingKind::HardwareTracing,
+            _ => panic!(),
```
ltratt, line 75: Isn't this turning a compile-time error into a run-time error? If you remove the `_=>panic!()` line, the compiler will refuse to compile this if we add a new `TracingKind` and don't expand the `match` accordingly.

ykcapi/src/lib.rs
```diff
@@ -61,4 +66,52 @@ mod c_testing {
         Box::leak(bm);
         ret
     }
+
+    #[no_mangle]
+    pub extern "C" fn __yktrace_start_tracing(kind: usize) -> *mut ThreadTracer {
```
ltratt, line 71: We presumably need to document somewhere that this returns a `Box`ed thing? Even if it's just a comment that says "if you don't follow this up with `__yktrace_stop_tracing` you're in undefined behaviour"?

ykcapi/src/lib.rs
```diff
@@ -61,4 +66,52 @@ mod c_testing {
...
+        res_bb: *mut usize,
+    ) {
+        let trace = unsafe { Box::from_raw(trace) };
+        let blk = trace.get(idx).unwrap();
+        let c_func = CString::new(blk.func_name()).unwrap();
+        unsafe {
+            *res_func = c_func.into_raw();
+            *res_bb = blk.bb();
+        }
+        Box::leak(trace);
```
ltratt, line 110: I think this might better be `mem::forget(trace)`? Basically you just don't want it to be `drop`ed, right?

ykcapi/src/lib.rs
```diff
@@ -61,4 +66,52 @@ mod c_testing {
...
+    /// Fetches the function name (`res_func`) and the block index (`res_bb`) at position `idx` in
+    /// `trace`. The caller must free() the function name.
+    #[no_mangle]
+    pub extern "C" fn __yktrace_irtrace_get(
+        trace: *mut IRTrace,
+        idx: usize,
+        res_func: *mut *mut c_char,
+        res_bb: *mut usize,
+    ) {
+        let trace = unsafe { Box::from_raw(trace) };
```
ltratt, line 103: Notice the implicit requirement that `trace: *mut IRTrace` was a `Box`. We should document that IMHO.

ykcapi/src/lib.rs
```diff
@@ -61,4 +66,52 @@ mod c_testing {
...
+        let kind: TracingKind = match kind {
+            SW_TRACING => TracingKind::SoftwareTracing,
+            HW_TRACING => TracingKind::HardwareTracing,
+            _ => panic!(),
+        };
+        Box::into_raw(Box::new(start_tracing(kind)))
+    }
+
+    #[no_mangle]
+    pub extern "C" fn __yktrace_stop_tracing(tt: *mut ThreadTracer) -> *mut IRTrace {
```
ltratt, line 81: This returns a `Box` which users are expected to pass repeatedly to `__yktrcae_irtrace_get` before finally calling `__yktrace_drop_irtrace` (I think)?

ykllvmwrap/src/symbolizer.rs
```diff
@@ -63,9 +73,10 @@ mod tests {
 
     #[test]
     fn find_code_sym_mangled() {
-        let f_addr = symbolize_me_mangled as *const fn() as *const _ as usize;
+        let f_vaddr = symbolize_me_mangled as *const fn() as *const _ as usize;
+        let (obj, f_off) = code_vaddr_to_off(f_vaddr).unwrap();
         let s = Symbolizer::new();
-        let sym = s.find_code_sym(f_addr).unwrap();
+        let sym = s.find_code_sym(obj, f_off as usize).unwrap();
```
ltratt, line 79: Can the `as` be a `try_from` or similar? Ditto for the `as`'s below.

yktrace/src/hwt/mapper.rs
```diff
@@ -2,39 +2,43 @@
...
     f_off: u64,
     /// Basic block number.
-    bb: u8,
+    bb: usize,
 }
 
 /// Maps (unrelocated) block offsets to their corresponding block map entry.
 pub struct BlockMap {
-    map: FxHashMap<usize, BlockMapEntry>,
+    tree: IntervalTree<u64, BlockMapEntry>,
```
ltratt, line 34: I notice that the block index here is a `u64` but above we have `bb: usize`. Are they / should they be the same integer types?

yktrace/src/hwt/mapper.rs
```diff
@@ -46,54 +50,124 @@ impl BlockMap {
...
+        debug_assert!(itr.all(|block| self.map_block(&block.unwrap()).is_empty()));
+
+        Ok(ret_irblocks)
     }
-}
 
-/// Extract the program header offset.
-fn get_phdr_offset() -> u64 {
-    (&objects()[0]).addr() as u64
+    /// Maps one PT block to one or many LLVM IR blocks.
```
ltratt, line 123: This isn't quite what I was expecting. I *think* the example below suggests that LLVM optimised away the implicit `JMP` at the end of `bbx` so that PT sees only one basic block?

If so, I think the example might be more clear if we spell this out. Something like:

```
Imagine we have two optimised LLVM blocks:

bb0:
  x
  y
  JMP bb1
bb1:
  z

If those two basic blocks are compiled adjacent to each other the `JMP bb1` is unneeded so we end up with machine code:

bb0:
  x
  y
bb1:
  z

If the machine code happens to JMP from non-bb0 to bb1 then PT will see a block starting at bb1. However, if execution flows straight from bb0 to bb1 PT will only see bb0. We thus have to expand out the implicit jmp.
```

That's not phrased very well, but hopefully something like that might work?

c_tests/tests/mapper_funcs_branches.c
```diff
@@ -0,0 +1,51 @@
+// Compiler:
```
ptersilie, line 1: What's this?

c_tests/tests/mapper_funcs_branches.c
```diff
@@ -0,0 +1,51 @@
...
+// Run-time:
+
+// Check the blockmap for this test program contains blocks.
```
ptersilie, line 4: "Check that"

c_tests/tests/mapper_funcs_branches.c
ltratt: It's a `lang_tester` test. [It basically just tests that the compiler and run-time don't error at the moment.]

yktrace/src/hwt/mapper.rs
vext01: The interval tree is mapping ranges of file offsets to block map entries, so the `u64` there is actually a file offset.

vext01:
> If it's just testing utils yktestutils maybe?

Open to new names, but it's not just for testing. The mapper uses this code to do it's thing.

yktrace/src/hwt/mapper.rs
ltratt: OK.

ykcapi/src/lib.rs
vext01: We can do yeah, but I've not been too worried about overly documenting code in the testing interface.

ykcapi/src/lib.rs
vext01: I'm not sure what you mean. The match has to be exhaustive, yes.

ltratt: My proposal is that you remove the line `_ => panic!()`.

ykcapi/src/lib.rs
ltratt: Hmm, good point. OK I'm less worried if it's purely used for testing.

ykcapi/src/lib.rs
vext01: That's the general gist, but users won't see this. It's for the testing code only. Users will use ykrt, which internally will just use the Rust code.

yktrace/src/hwt/mapper.rs
vext01: Yes, LLVM has lowered a jump to a fall-thru during code-gen. But apparently that's not how the blockmap section sees it.

During development I had an assert that checked that each blockmap query only returned one block. The assert was failing due to the reason outlined in the docstring.

[Note that in LLVM-IR there is no such concept as fall-thru. All blocks must be properly terminated.]

yktrace/src/hwt/mapper.rs
```diff
@@ -46,54 +50,124 @@ impl BlockMap {
...
+    ///
+    /// The reason that there may be many corresponding blocks is due to this scenario:
+    ///
+    ///  |   bbx:
+    ///  |      mov ...
+    ///  |      mov ...
+    ///  \-> bby:
+    ///         mov ...
+    ///
+    /// Here a PT trace from bbx falling into bby wil be considered a single block (there were no
```
ptersilie, line 133: s/wil/will
s/no no/no

yktrace/src/hwt/mapper.rs
ltratt: OK, so I think we need to clarify the comment, perhaps as I suggested above: it took me quite a while to work out what's going on. [Possibly because I'm an idiot, but idiots like me do deserve sympathy! The arrow in the comment was a bit misleading IMHO.]

ykcapi/src/lib.rs
vext01: That won't work. The match is over an integer (due to C ABI reasons).

```
72 |         let kind: TracingKind = match kind {
   |                                       ^^^^ pattern `_` not covered
```

ltratt: Aha!

ykcapi/src/lib.rs
vext01: You sir, are correct.

Bjorn had a more concise way of getting a reference without any Box dances (thus effectively borrowing the content), but I forgot how it was done :P

`mem::forget` is fine.

ykllvmwrap/src/symbolizer.rs
vext01: Hrm, arguably `find_code_sym` should take a `u64` as it is dealing with a file offset, not a virtual address. Let me reshuffle some types...

ykcapi/src/lib.rs
vext01: Fixed in 2a1af1a2

ltratt: You can convert the `*mut` into a `&` with `unsafe { &*trace }` and that won't be dropped (it's just a reference).

ykllvmwrap/src/symbolizer.rs
vext01: Fixed in 9433cc07425 along with some out of date commentary.

yktrace/src/hwt/mapper.rs
vext01: I eventually understood the distinction between my comment and yours.

I went back and re-checked the exact reason for the block split.

In one of the mapper tests we have this C func:
```
__attribute__((noinline))                                                                                                       
int                                                                                                                             
add_one_or_two(int arg)                                                                                                         
{                                                                                                                               
    if (arg % 2 == 0) {                                                                                                         
        arg++;                                                                                                                  
    } else {                                                                                                                    
        arg += 2;                                                                                                               
    }                                                                                                                           
    return arg;                                                                                                                 
}         
```

And here's the LLVM IR:
```
; Function Attrs: noinline nounwind optnone uwtable                                                                                                                                                                 
define internal fastcc i32 @add_one_or_two(i32 %0) unnamed_addr #0 {                                                                                                                                                
  %2 = alloca i32, align 4                                                                                                                                                                                          
  store i32 %0, i32* %2, align 4                                                                                                                                                                                    
  %3 = load i32, i32* %2, align 4                                                                                                                                                                                   
  %4 = srem i32 %3, 2                                                                                                                                                                                               
  %5 = icmp eq i32 %4, 0                                                                                                                                                                                            
  br i1 %5, label %6, label %9                                                                                                                                                                                      
                                                                                                                                                                                                                    
6:                                                ; preds = %1                                                                                                                                                      
  %7 = load i32, i32* %2, align 4                                                                                                                                                                                   
  %8 = add nsw i32 %7, 1                                                                                                                                                                                            
  store i32 %8, i32* %2, align 4                                                                                                                                                                                    
  br label %12                                                                                                                                                                                                      
                                                                                                                                                                                                                    
9:                                                ; preds = %1                                                                                                                                                      
  %10 = load i32, i32* %2, align 4                                                                                                                                                                                  
  %11 = add nsw i32 %10, 2                                                                                                                                                                                          
  store i32 %11, i32* %2, align 4                                                                                                                                                                                   
  br label %12                                                                                                                                                                                                      
                                                                                                                                                                                                                    
12:                                               ; preds = %9, %6                                                                                                                                                  
  %13 = load i32, i32* %2, align 4                                                                                                                                                                                  
  ret i32 %13                                                                                                                                                                                                       
}
```

Which is compiled to:
```
            ;-- add_one_or_two:                                                                                                 
            0x00201a10      55             push rbp                                                                             
            0x00201a11      4889e5         mov rbp, rsp                                                                         
            0x00201a14      897dfc         mov dword [rbp - 4], edi                                                             
            0x00201a17      8b45fc         mov eax, dword [rbp - 4]                                                             
            0x00201a1a      b902000000     mov ecx, 2                                                                           
            0x00201a1f      99             cdq                                                                                  
            0x00201a20      f7f9           idiv ecx                                                                             
            0x00201a22      83fa00         cmp edx, 0                                                                           
        ┌─< 0x00201a25      750b           jne 0x201a32                                                                         
        │   0x00201a27      8b45fc         mov eax, dword [rbp - 4]                                                             
        │   0x00201a2a      83c001         add eax, 1                                                                           
        │   0x00201a2d      8945fc         mov dword [rbp - 4], eax                                                             
       ┌──< 0x00201a30      eb09           jmp 0x201a3b                                                                         
       │└─> 0x00201a32      8b45fc         mov eax, dword [rbp - 4]    ; first hit                                 
       │    0x00201a35      83c002         add eax, 2                                                                           
       │    0x00201a38      8945fc         mov dword [rbp - 4], eax                                                             
       └──> 0x00201a3b      8b45fc         mov eax, dword [rbp - 4]    ; second hit                                  
            0x00201a3e      5d             pop rbp                                                                              
            0x00201a3f      c3             ret                                                                                  
```

And:
 * PT saw a block `201a32..=201a3f`.
 * The blockmaper returned 2 hits (which I've annotated in the asm above):
    * `201a32`
    * `201a3b`

So it's actually a mix of both of our reasoning:

 * Yes, there's a fallthru at `201a38` into `201a3b`.
 * The fallthru came from an LLVM `br` (unconditional jump)
 * But the fallthru only exists because there's another incoming edge from `201a30` to `201a3b` and this splits the block.

So both of our comments are correct.

Does that make any sense? I'll try to clarify the comment a bit tomorrow.

ltratt: > But the fallthru only exists because there's another incoming edge from 201a30 to 201a3b and this splits the block.

This doesn't seem like a necessary part of the logic to me. But maybe I'm missing something?

vext01: Put differently, if there hadn't been that incoming edge, then those two blocks could have been one and talk of fallthru is then moot.

You are still correct though because, although pointless, LLVM could keep two mergable blocks separate.

I think we are agreeing, but we see the situation in different, but equivalent, ways.

ltratt: > Put differently, if there hadn't been that incoming edge, then those two blocks could have been one block and talk of fallthru is then moot.

I don't think that's true. Because even though we only see 1 PT block, when the code-gen removes a branch to the immediately-following basic block, we still need to realise that it represents two LLVM IR blocks.

vext01: You can see it that way.

If the initial LLVM IR was:
```
bb1:
  ...
  br bb2
bb2:
  ...
```

And the `br` was removed for fallthru, then yes, we'd map one PT block to two IR blocks.

What I'm saying is, I bet even with `-O0` LLVM would merge those.

Regardless, we should be able to handle the case where the blocks are separate.

ltratt: > What I'm saying is, I bet even with -O0 LLVM would merge those.

I don't think it matters about optimisation levels :) I *think* the only important point is that something subsequent to IR optimisation (i.e. the machine code generation part) can remove branches for adjacent blocks. My claim is that whether that's `-O0` or `-O3` or whether there is an incoming edge is irrelevant.

vext01: Having sat on this overnight, I think I agree that the incoming edge, whilst related, is a distraction. I'll revise the comment.

vext01: Try 6e8c6111

yktrace/src/hwt/mapper.rs
vext01: In a later PR we could wrap `u64` into a `Off` struct, and `uszie` into a `Vaddr` struct for clarity/added type safety?

yktrace/src/hwt/mapper.rs
ltratt: Works for me.

yktrace/src/hwt/mapper.rs
ltratt: Up to you.

ltratt:
I think we're ready to squash?

vext01:
We are if you are OK with those last two commits.

ltratt:
I am! Please squash.

vext01:
splat.

vext01:
bors ping

bjorn3:
Rust-analyzer also has trouble with bors at the moment.

vext01:
still down?

bors ping

vext01:
Once bors starts working again, I expect this PR will fail to test as an LLVM assertion will abort the `find_code_sym_not_found` test (turns out I developed without assertions).

If that's the case, I think we can remove that test. It's not important.

bors ping

ltratt:
Let's try.

bors r+

vext01:
As expected, LLVM assertion failure:
```
Program aborted due to an unhandled Error: No such file or directory
```

Fixed in c546488. OK to squash?

ltratt:
Please squash.

vext01:
splat.

vext01:
OK to squash?

ltratt:
Please squash.

vext01:
splat.

PR #337, opened 2021-05-13
Tweak ykrt's C API a little bit.

ltratt:
There is one non-cosmetic change:

  * We now pass a YkLocation (not a *YkLocation) to `drop` to make a slight parallel with a Rust "move" (i.e. trying to be a bit clearer that "after this call, you can't do anything with this location").

Then two cosmetic changes:

  * Adding Yk prefixes.
  * Adding a "don't include this header multiple times" #define.

### DIFF
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -29,12 +29,12 @@ pub extern "C" fn yk_control_point(mt: *mut MT, loc: *mut Location) {
 }
 
 #[no_mangle]
-pub extern "C" fn yk_new_location() -> Location {
+pub extern "C" fn yk_location_new() -> Location {
     Location::new()
 }
 
 #[no_mangle]
-pub extern "C" fn yk_drop_location(loc: *mut Location) {
+pub extern "C" fn yk_location_drop(loc: Location) {
     drop(loc)
 }
 
--- a/c_tests/tests/basic_mt.c
+++ b/c_tests/tests/basic_mt.c
@@ -7,10 +7,10 @@ int
 main(int argc, char **argv)
 {
     MT *mt = yk_mt();
-    Location loc = yk_new_location();
+    YkLocation loc = yk_location_new();
     for (int i = 0; i < yk_mt_hot_threshold(mt); i++) {
         yk_control_point(mt, &loc);
     }
-    yk_drop_location(&loc);
+    yk_location_drop(loc);
     return 0;
 }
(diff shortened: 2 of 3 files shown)


ykcapi/src/lib.rs
```diff
@@ -34,7 +34,7 @@ pub extern "C" fn yk_new_location() -> Location {
 }
 
 #[no_mangle]
-pub extern "C" fn yk_drop_location(loc: *mut Location) {
+pub extern "C" fn yk_drop_location(loc: Location) {
```
vext01, line 37: Begs the question: what's the point of this function now? If locations are stack alloc'd then they are automatically dropped, in both C and Rust.

Also, didn't the address of the location have some meaning? Maybe not any more, but if so, copying the location around by value will make its identity change.

vext01: I see, there is drop code.

Identity question still holds.

ltratt: > Also, didn't the address of the location have some meaning?

No it deliberately doesn't. Internally, the malloc'd block the `Location` points to is significant, but it's a deliberate choice that users can move `Location`s around freely.

vext01: OK, that's fine.

vext01:
unrelated question:

Do we prefer `yk_drop_location` or `yk_location_drop`? I'd been using the latter convention in the testing api.

i.e.

```
[__]yk_<type>_<func>
```

ltratt:
I'm neutral.

vext01:
I think the latter is more general. 

Suppose we want to expose a rust function called `Blah::do_the_thing()`.

`yk_do_the_thing_blah()` is odd and I'd prefer `yk_blah_do_the_thing()`.

ltratt:
Fixed in ce7f0ab.

vext01:
Thanks, and sorry for being picky.

Please squash.

ltratt:
Squashed.

ltratt:
That's embarrassing. OK to squash https://github.com/softdevteam/yk/pull/337/commits/9e65cc4f28a01f6acca1790ba183f144d2dd62ca?

vext01:
yup

ltratt:
Squashed.

ltratt:
And another fallout from the renaming in https://github.com/softdevteam/yk/pull/337/commits/38877fd431aa72b816de3cc1ffcafe0f168013f0. OK to squash?

vext01:
Please squash.

ltratt:
Squashed.

PR #340, opened 2021-05-18
Add a yk_stopgap_safe marker.

ltratt:
When adjusting an existing C interpreter, we often want to mark multiple points in the interpreter as being safe for the stopgap interpreter to transition back to normal "C" interpretation. This commit adds a function that we can use as such as a marker in the future (at the moment it's a no-op). Exactly how we teach clang/rustc/LLVM about what's special about this function is something we can worry about in the future. Similarly, the precise name for the function is something we can tweak in the future.

Why do we need this marker at all? Well, previously we assumed this was synonymous with control points, but in the presence of computed gotos this isn't really enough: interpreters rarely go through control points. So while it should always be safe for the stopgap interpreter to transition back to normal execution at a control point, that potentially means we execute *lots* of code in the stopgap interpreter unnecessarily. "yk_stopgap_safe" allows an interpreter author to add extra places in their interpreter where it's safe to transition back to normal execution. The precise rules of "safe" are deeply unclear, but I believe a safe, conservative approximation is "add it either
directly before a `goto` (if you're using computed gotos for your main interpreter loop) or at the very end or beginning of your interpreter's main loop". We might be able to relax these requirements later, but I don't think it's something we should worry about for now.

### DIFF
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -29,6 +29,12 @@ pub extern "C" fn yk_control_point(mt: *mut MT, loc: *mut Location) {
 }
 
 #[no_mangle]
+pub extern "C" fn yk_stopgap_safe() {
+    // This function is a marker for the trace compiler. It might need to become an LLVM builtin,
+    // or we may need to add an LLVM annotation or ...
+}
+
+#[no_mangle]
 pub extern "C" fn yk_location_new() -> Location {
     Location::new()
 }
--- a/c_tests/tests/basic_mt.c
+++ b/c_tests/tests/basic_mt.c
@@ -10,6 +10,7 @@ main(int argc, char **argv)
     YkLocation loc = yk_location_new();
     for (int i = 0; i < yk_mt_hot_threshold(mt); i++) {
         yk_control_point(mt, &loc);
+        yk_stopgap_safe();
     }
     yk_location_drop(loc);
     return 0;
--- a/ykcapi/yk.h
+++ b/ykcapi/yk.h
@@ -45,6 +45,13 @@ YkHotThreshold yk_mt_hot_threshold(MT *);
 // ever be the beginning of a trace.
 void yk_control_point(MT *, YkLocation *);
 
+// Mark this point in the program as being a safe point for the stopgap
+// interpreter to transition back to normal program execution. Typical stopgap
+// safe points are at the backwards jump(s) of an interpreter's main for/while
+// loop. If, for example, your interpreter uses interpreted gotos, you may want
+// to mark multiple locations as stopgap safe.
+void yk_stopgap_safe();
+
 // Create a new `Location`.
 //
 // Note that a `Location` created by this call must not simply be discarded: if


ykcapi/src/lib.rs
```diff
@@ -28,6 +28,12 @@ pub extern "C" fn yk_control_point(mt: *mut MT, loc: *mut Location) {
     }
 }
 
+#[no_mangle]
+pub extern "C" fn yk_stopgap_safe() {
+    // This function is a marker for the trace compiler. It might need to become an LLVM builtin,
+    // or we may need to add an LLVM annotation or ...
+}
```
bjorn3, line 35: LLVM has ways to go from compiled code to an interpreter, but I can't find anything to go from an interpreter to compiled code.

https://llvm.org/docs/StackMaps.html, https://llvm.org/docs/LangRef.html#llvm-experimental-deoptimize-intrinsic

ptersilie:
Looks good to me, but let's address bjorn's comment before merging.

ykcapi/src/lib.rs
ltratt: The "new" yk (which, unlike "old" yk isn't fundamentally Rust based) is, in a sense, the missing bit that you're referring to. It (will) use things like stackmaps and `deoptimize` to help it do so. That does raise some interesting challenges, and one thing that's implicit in this PR is that I think we'll need to extend clang/LLVM to support function attributes that yk will need. This is all a bit sketch-y at the moment, so the precise details will probably alter!

bjorn3: I mean that LLVM currently doesn't have a way to setup the stackframe and jump into the middle of a function AFAICT. It only has support for extracting info out of the stackframe and jumping out of the middle of a function into an interpreter. Stackmaps store where variables can be read, but they may very well also be held in registers at the same time with the same value or derived values may be held in registers (or on the stack without the stackmap mentioning how to compute them or mentioning them at all). Jumping into the middle of a function would require computing the derived values and storing the right value in registers.

```rust
fn foo(a: u8) -> u8 {
    safepoint();
    a * 3
}
```

could be transformed into for example

```rust
fn foo(a: u8) -> u8 {
    let tmp = a << 1;
    safepoint();
    a + tmp
}
```

The stackmap would only record `a` at the safepoint, however to safely jump into `foo` at the safepoint, you have to store `tmp` at the right location. Another example would be

```rust
fn foo(a: u8) -> u8 {
    foo(&a);
    safepoint();
    a
}
```

could be codegened as

```rust
fn foo(a: u8) -> u8 { // a is stored in a register
    // store a on the stack here
    foo(&a); // reference to the a on the stack is passed to foo
    safepoint(); // stackmap references a on the stack
    a // return the version of a in a register
}
```

here only the a on the stack is marked in the stackmap, but the a in a register is actually used for the return value and as such needs to be written when jumping into foo at the safepoint.

Stackmaps are sufficient for reading the value of all locals, but not for writing it as required to jump into the middle of a function from the interpreter.

ltratt: That's my understanding to. Basically we expect control flow to look like:

```c
pc = 0;
while (true) {
   control_point(...);
   if opcodes[pc] == PUSH {...}
   else if opcodes[pc] == BACKWARDS_JUMP {
      pc -= val(opvalue[pc]);
      yk_stopgap_safe();
      continue;
  }
}
```

As a good approximation of the truth, you go into compiled code at `control_point` and come out of it at `stopgap_safe`: the latter thus needs to put the frame in the "right state" for `control_point` to take back over. This is far beyond what any C spec is likely to declare as defined behaviour, but there's no alternative that I can think of.

vext01:
If I understand Laurie's reasoning, due to unstructured control flow in interpreter implementations, a bytecode may not simply be executed at the top of an interpreter loop, but in many places.

Assuming the above is true, wouldn't you still want to profile all bytecodes using a control point? Presumably a loop may start at any given bytecode in the language interpreter. So should we just have many invocations to the control point instead of this extra annotation?

ltratt:
> If I understand Laurie's reasoning, due to unstructured control flow in interpreter implementations, a bytecode may not simply be executed at the top of an interpreter loop, but in many places.

I'm not sure I understand this.

>Assuming the above is true, wouldn't you still want to profile all bytecodes using a control point?

I've been wondering this on and off for ages and my thinking has flip-flopped more than once. I *think* the answer is "no". I base this partly on the fact that the Converge RPython interpreter profiles all opcodes, and that was mostly a waste -- admittedly that's partly because control points in RPython are more expensive than in yk. But control points aren't free in yk either

What we do in yklua is only profile opcodes which are definitely a) the beginning of a function b) the beginning of a loop [OK, OK, we also profile opcodes that have something to do with Lua coroutines, but that might or might not be a good thing -- I'm not sure.]

What I *do* think we might want to do is have a third annotation which people can put at the start of every bytecode to say "don't bother profiling anything, but if there is machine code for this `Location` please execute it". However, I don't think we need to worry too much about this yet: at worst we go around the interpreter once for each loop before executing machine code. It will definitely get us a noticeable speed-up, but I think it won't be vast.

vext01:
OK, i'm not too worried about this, as I think we are likely to get it wrong first time whatever we do. It's going to evolve over time as our understanding of what we need improves. Agreed?

ltratt:
There's something I left implicit earlier. In a computed goto interpreter we have something like:

```c
int pc = ...;
int next_instr;
int opcode = opcodes[pc];
while (true) {
  switch opcode {
    OP_POP: { <do pop>;  pc += 1; next_instr = next_instr_pointer(); goto next_instr; }
    OP_BACKWARDS_JMP: { pc -= ...; next_instr = next_instr_pointer(); goto next_instr; }
  }
}
```

If all we have is `control_point` then we can do:

```
```c
int pc = ...;
int next_instr;
int opcode = opcodes[pc];
while (true) {
  switch opcode {
    OP_POP: { control_point(); <do pop>;  pc += 1; next_instr = next_instr_pointer(); goto next_instr; }
    OP_BACKWARDS_JMP: { control_point(); pc -= ...; next_instr = next_instr_pointer(); goto next_instr; }
  }
}
```

but that means that we do control point stuff for every bytecode, even those that can't possibly be loop/function headers -- I believe that will have a noticeable effect on the interpreter's performance (I'll guess ~10%).

So, like my first attempt in ykcpython, we can try avoid putting control points at opcodes that can't possibly be loop/function headers e.g.:

```c
int pc = ...;
int next_instr;
int opcode = opcodes[pc];
while (true) {
  switch opcode {
    OP_POP: { <do pop>;  pc += 1; next_instr = next_instr_pointer(); goto next_instr; }
    OP_BACKWARDS_JMP: { pc -= ...; next_instr = next_instr_pointer(); control_point(); goto next_instr; }
  }
}
```

However, that means that we potentially have to execute many opcodes in the stopgap interpreter. yk_stopgap_safe means we can stop that happening with:

```c
int pc = ...;
int next_instr;
int opcode = opcodes[pc];
while (true) {
  control_point();
  switch opcode {
    OP_POP: { <do pop>;  pc += 1; next_instr = next_instr_pointer(); stopgap_safe(); goto next_instr; }
    OP_BACKWARDS_JMP: { pc -= ...; next_instr = next_instr_pointer(); continue; }
  }
}
```

With our mythical third annotation, however, we can do:

```c
int pc = ...;
int next_instr;
int opcode = opcodes[pc];
while (true) {
  control_point();
  switch opcode {
    OP_POP: { if !tracer_switch() { continue; } 
      <do pop>;  pc += 1; next_instr = next_instr_pointer(); stopgap_safe(); goto next_instr;
    }
    OP_BACKWARDS_JMP: { pc -= ...; next_instr = next_instr_pointer(); continue; }
  }
}
```

ltratt:
All that said, stopgap_safe and tracer_switch are, arguably, optimisations to allow one not to put control points on all opcodes. But I'm now wondering if I'm prematurely optimising here.

ptersilie:
`stopgap_safe()` means that it's safe at this point to leave the stopgap interpreter and continue running the the compiled interpreter, right? And if I understood bjorn's comment correctly, then he was saying that this
```
next_instr = next_instr_pointer(); stopgap_safe(); goto next_instr;
```
means we need to reconstruct a compiled stackframe to continue running compiled code after `stopgap_safe`, e.g. we somehow need to know what `next_instr` is, but `next_instr_pointer` was interpreted by the stopgap interpreter. Although, if we could actually do that, then we wouldn't need a stopgap interpreter in the first place. I feel like there's something I'm missing here.

bjorn3:
> means we need to reconstruct a compiled stackframe to continue running compiled code after stopgap_safe

Correct

> e.g. we somehow need to know what next_instr is, but next_instr_pointer was interpreted by the stopgap interpreter.

You should already know that as it is just another local and you should be able to get the value of all necessary locals to even get to the stopgap interpreter. The problem here is that there may be new locals introduced that need to be restored, but are not included in the stackmap or locals are included in the stackmap at a single location, but need to be restored to multiple locations (eg a stackslot and a register).

At least that is my understanding of how much LLVM stackmaps guarantee. Maybe you could ask someone who works on LLVM if stackmaps are also suitable for returning to the middle of a function?

If not maybe you could have a pass after LTO that for each function with this kind of marker duplicates it and in the duplicate adds extra code to restore all necessary locals and then jump to the location of the marker. You can then return to this function from the stopgap interpreter.

ltratt:
I'm going to close this. For the time being, it makes our lives easier if we only have to think about `control_point`s. I think we might come back to something like `stopgap_safe` when we have more evidence about how best to structure interpreters in our brave new world.

PR #341, opened 2021-05-18
Make IRBlock store a CString, not a String.

vext01:
Currently when we hand out an IRBlock's function name to C/C++ we have
to convert the internal Rust string into a CString, before passing it to
the caller. The caller is then responsible for freeing the string. This is
unergonomic and slower than it could be.

We didn't care before as we only do this in testing, but soon we will
have some C++ code on the critical path which will need to walk an
IRTrace querying IRBlocks. Therefore we should fix this now.

Storing a CString instead of a String means that we can hand out
borrowed `char *` pointers (via CStr) without having to allocate each
time and without the caller having to free the result.

### DIFF
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -42,7 +42,7 @@ pub extern "C" fn yk_location_drop(loc: Location) {
 /// These symbols are not shipped as part of the main API.
 #[cfg(feature = "c_testing")]
 mod c_testing {
-    use std::{ffi::CString, mem, os::raw::c_char};
+    use std::{mem, os::raw::c_char};
     use yktrace::{start_tracing, BlockMap, IRTrace, ThreadTracer, TracingKind};
 
     const SW_TRACING: usize = 0;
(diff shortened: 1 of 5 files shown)


ltratt:
Dumb question: do we *have* to pass `CStr`/`CString` around? What if we were to wrap it in a new type of our devising so that we don't have to copy the string so often?

ykllvmwrap/src/symbolizer.rs
```diff
@@ -26,19 +29,18 @@ impl Symbolizer {
...
-            let ret = {
-                let sym = unsafe { CStr::from_ptr(ptr) };
-                let sym = String::from(sym.to_str().unwrap());
-                if sym == "<invalid>" {
-                    return None;
-                }
-                sym
-            };
+            let sym = unsafe { CStr::from_ptr(ptr) };
+            if sym == INVALID_SYM.as_c_str() {
```
bjorn3, line 38:
```suggestion
            if sym.to_bytes() == b"<invalid>" {
```

vext01:
> Dumb question: do we have to pass CStr/CString around? What if we were to wrap it in a new type of our devising so that we don't have to copy the string so often?

Ultimately we need a C++ String to query the LLVM module with, but due to ABI and C stuff, the lowest common denom is going to be a C string. I don't think we can improve the situation with our own type.

Later we will however implement a CString pool, as many of the strings are duplicated. But that's for later.

ykcapi/src/lib.rs
```diff
@@ -91,19 +91,18 @@ mod c_testing {
...
-        res_func: *mut *mut c_char,
+        res_func: *mut *const c_char,
         res_bb: *mut usize,
     ) {
         let trace = unsafe { Box::from_raw(trace) };
         let blk = trace.get(idx).unwrap();
-        let c_func = CString::new(blk.func_name()).unwrap();
         unsafe {
-            *res_func = c_func.into_raw();
+            *res_func = blk.func_name().as_ptr();
```
ptersilie, line 105: `func_name` returns a `&CStr`, whose pointer you then assign to `res_func`. Isn't the string still owned by `blk`, which goes out of scope at the end of this function. Or is `trace` where `blk` comes from still sticking around? It is hard to see from this code that the `CStr` pointer is still valid after `__yktrace_irtrace_get` returns.

vext01: The string is indeed owned by `blk` who in turn is owned by the (borrowed) `trace`. Thus the string lives on.

ykllvmwrap/src/symbolizer.rs
vext01: Fixed in 87f21bd09615, thanks!

vext01:
OK to squash?

ptersilie:
Please do.

vext01:
splat.

PR #342, opened 2021-05-19
Stub out the C++ trace compilation routine.

vext01:
**Do not merge, this isn't finished**

This is an attempt to:
 * Make accessors for getting a trace's length and its blocks available to C++.
 * Stubbing out a C++ function that Lukas can drop his LLVM code into to compile a trace.

I'm getting this error when a C test is compiled:
```
/tmp/.tmpDLh635/blockmap: symbol lookup error: /home/vext01/research/yorick/yk/target/debug/deps/libykcapi.so: undefined symbol: _Z21__yktrace_irtrace_lenPv
```

I've never seen this kind of error before.

And I've narrowed it down to this (seemingly unrelated) function:
```
#[no_mangle]                                                                                                                                                                                              
pub extern "C" fn __yktrace_irtrace_compile(trace: *mut IRTrace) {                                                                                                                                        
    unsafe {&*trace}.compile(); // XXX                                                                                                                                                           
}
```

Comment out the line marked `XXX` and the error goes away.

I'm baffled. Does anyone see anything wrong?

### DIFF
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -112,4 +112,9 @@ mod c_testing {
     pub extern "C" fn __yktrace_drop_irtrace(trace: *mut IRTrace) {
         unsafe { Box::from_raw(trace) };
     }
+
+    #[no_mangle]
+    pub extern "C" fn __yktrace_irtrace_compile(trace: *mut IRTrace) {
+        unsafe { &*trace }.compile();
+    }
 }
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -60,10 +60,7 @@ mod c_testing {
 
     #[no_mangle]
     pub extern "C" fn __yktrace_hwt_mapper_blockmap_len(bm: *mut BlockMap) -> usize {
-        let bm = unsafe { Box::from_raw(bm) };
-        let ret = bm.len();
-        mem::forget(bm);
-        ret
+        unsafe { &*bm }.len()
     }
 
     #[no_mangle]
@@ -84,10 +81,7 @@ mod c_testing {
 
     #[no_mangle]
     pub extern "C" fn __yktrace_irtrace_len(trace: *mut IRTrace) -> usize {
-        let trace = unsafe { Box::from_raw(trace) };
-        let ret = trace.len();
-        mem::forget(trace);
-        ret
+        unsafe { &*trace }.len()
     }
 
     /// Fetches the function name (`res_func`) and the block index (`res_bb`) at position `idx` in
@@ -99,13 +93,12 @@ mod c_testing {
         res_func: *mut *const c_char,
         res_bb: *mut usize,
     ) {
-        let trace = unsafe { Box::from_raw(trace) };
+        let trace = unsafe { &*trace };
         let blk = trace.get(idx).unwrap();
         unsafe {
             *res_func = blk.func_name().as_ptr();
             *res_bb = blk.bb();
         }
-        mem::forget(trace);
     }
 
     #[no_mangle]
--- a/yktrace/src/lib.rs
+++ b/yktrace/src/lib.rs
@@ -64,6 +64,17 @@ impl IRTrace {
     pub fn get(&self, idx: usize) -> Option<&IRBlock> {
         self.blocks.get(idx)
     }
+
+    pub fn compile(&self) {
+        let len = self.len();
+        let mut func_names = Vec::with_capacity(len);
+        let mut bbs = Vec::with_capacity(len);
+        for blk in &self.blocks {
+            func_names.push(blk.func_name().as_ptr());
+            bbs.push(blk.bb());
+        }
+        unsafe { ykllvmwrap::__ykllvmwrap_irtrace_compile(func_names.as_ptr(), bbs.as_ptr(), len) }
+    }
 }
 
 /// Binary executable trace code.
--- a/c_tests/tests/compile_simple.c
+++ b/c_tests/tests/compile_simple.c
@@ -1,0 +1,25 @@
+// Compiler:
+// Run-time:
+
+// Check that basic trace compilation works.
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk_testing.h>
+
+int
+main(int argc, char **argv)
+{
+    void *tt = __yktrace_start_tracing(HW_TRACING);
+    int res = argc + 1;
+    void *tr = __yktrace_stop_tracing(tt);
+
+    assert(res == 2);
+    __yktrace_irtrace_compile(tr); // FIXME test something.
+
+    __yktrace_drop_irtrace(tr);
+
+    return (EXIT_SUCCESS);
+}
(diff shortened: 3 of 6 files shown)


ltratt:
The error might be a bit misleading in the sense that it could be something *reachable* from `compile()` that's throwing the error. For example if you comennt out the `ykllvmwrap::__ykllvmwrap_irtrace_compile` call, what happens? [It also feels a bit odd to me that we seem to go C -> Rust -> C at that point? But I might have misunderstood.]

vext01:
> For example if you comennt out the ykllvmwrap::__ykllvmwrap_irtrace_compile call, what happens?

Commenting that out (in `IRTrace::compile()`) also makes the error disappear. Hrm...

> It also feels a bit odd to me that we seem to go C -> Rust -> C at that point? But I might have misunderstood.

All of the `c_tests` are `C -> Rust -> C++` if that's what you mean.

ltratt:
> Commenting that out (in IRTrace::compile()) also makes the error disappear. Hrm...

My suggestion is to follow this through the control flow graph (i.e. comment out parts of functions being called etc etc): what's possibly happening is that some indirect dependency is causing the linker error.

ykcapi/yk_internal.h
```diff
@@ -0,0 +1,5 @@
+// C ABI functions used internally by the system to call from C++ back to Rust.
+// They may also be used by C tests.
+
+size_t __yktrace_irtrace_len(void *trace);
+void __yktrace_irtrace_get(void *trace, size_t idx, char **res_func, size_t *res_bb);
```
bjorn3, line 5: You need to wrap this in an `extern "C"` block to avoid C++ mangling the names.

vext01: damn, I didn't realise that you had to do it on the headers, but of course!

vext01:
OK, well 9203931 is the fix I devised prior to seeing @bjorn3's comment.

Which approach do we prefer?

Arguably 9203931 is simpler, but it does require [looping over the `IRTrace` and converting the `IRBlock`s to something FFI compatible](https://github.com/ykjit/yk/pull/342/commits/92039318fe5ddabc96cb787f2c3cd75e5ba9b250#diff-05da6bebe23e3531cbbc156d93e8acf76445526b8ccf30530c300df2b664bf6aR70).

(I've just noticed some unnecessary churn in that diff, sorry)

I did consider trying to make `IRBlock` `repr(C)`, but the `CString` is then problematic.

ltratt:
I have no preference -- choose whatever you think is best.

vext01:
In some sense it doesn't matter, as this isn't user-facing and we can optimise later.

I also prefer to *not* introduce another header, so let's go with 9203931.

I'll tidy it up now.

vext01:
Fixed.

So... This change adds `IRTrace::Compile` which calls a dummy C++ function `ykllvmwrap::__ykllvmwrap_irtrace_compile` for me and Lukas to populate shortly with code to actually compile a trace.

It also adds a dummy test `compile_simple.c` which we can use to drive trace compilation code.

[If people prefer, we can stack the compilation code on top of this PR -- I don't care which]

yktrace/src/lib.rs
```diff
@@ -67,12 +67,13 @@ impl IRTrace {
...
         let len = self.len();
-        let func_names = self
-            .blocks
-            .iter()
-            .map(|b| b.func_name().as_ptr())
-            .collect::<Vec<_>>();
-        let bbs = self.blocks.iter().map(|b| b.bb()).collect::<Vec<_>>();
+        let mut func_names = Vec::with_capacity(len);
+        let mut bbs = Vec::with_capacity(len);
+        for i in 0..len {
```
ltratt, line 72: Is this `for blk in &self.blocks` in disguise?

ykcapi/src/lib.rs
```diff
@@ -83,33 +116,12 @@ mod c_testing {
...
-        mem::forget(trace);
+    pub extern "C" fn __yktrace_drop_irtrace(trace: *mut IRTrace) {
+        unsafe { Box::from_raw(trace) };
     }
 
     #[no_mangle]
-    pub extern "C" fn __yktrace_drop_irtrace(trace: *mut IRTrace) {
-        unsafe { Box::from_raw(trace) };
+    pub extern "C" fn __yktrace_irtrace_compile(trace: *mut IRTrace) {
+        unsafe {&*trace}.compile();
```
ptersilie: Is this a way to get around the `Box::new()/mem::forget()` dance? If so, why don't we use it in the other places too?

ptersilie: I'm not sure that's what's going on here, but it looks to me like Edd uses `{&*trace}.compile()` instead of doing
```rust
let t = Box::from_raw(trace);
t.compile();
mem::forget(t);
```
as he's done in the other functions. If that is the case, then I'm wondering why we are not consistent everywhere. Unless there's something else going on here.

ltratt: `&*` converts a `*mut` into a `&`: since it's a reference, there is nothing to drop (you can't drop references). It's much more elegant than the `from_raw`/`forget` dance -- and more flexible, because it doesn't require the pointer to have been created from a `Box`.

ptersilie: Yes, hence my question why we don't use this in the other places too instead of doing `from_raw/forget`, like [here](https://github.com/ykjit/yk/pull/342/commits/b408c8a0ced85e7c0c994682f5c50f46e21bbf1f#diff-e609c9c537d75440ea3db6f14fa49100e4e511dab07bfb704cc4e8199fb1fae3R49) and [here](https://github.com/ykjit/yk/pull/342/commits/b408c8a0ced85e7c0c994682f5c50f46e21bbf1f#diff-e609c9c537d75440ea3db6f14fa49100e4e511dab07bfb704cc4e8199fb1fae3R64).

ltratt: I don't think Edd's introduced any new `from_raw`'s in this PR. I think fixing those in a separate PR would be a good thing.

ptersilie: What do you mean? The two links I posted above are functions introduced in this PR, where he can also use `&*` instead of `from_raw/forget`.

ltratt: I *think* later commits have removed that code.

ptersilie: Huh, I guess I'm the numpty here. I didn't look at the [overall changes](https://github.com/ykjit/yk/pull/342/files) and just the separate commits. I only saw that this code was moved around [herehttps://github.com/ykjit/yk/pull/342/commits/92039318fe5ddabc96cb787f2c3cd75e5ba9b250#diff-e609c9c537d75440ea3db6f14fa49100e4e511dab07bfb704cc4e8199fb1fae3R87) and then never touched again. So I assumed it must be still there. What I didn't realise is that this code was just moved somewhere else in the first commit, and then moved back to where it was initially in the second commit, negating those changes.

Anyway, I think it's still worth changing. Either here or in a separate PR.

vext01: There are indeed some other places where I could use `&*` instead of all of the boxing fun. I only recently remembered this trick.

vext01: Fixed in a commit that I'll leave separate: 241f73d

yktrace/src/lib.rs
vext01: Fixed in 7b4c878

vext01:
I think that's all the comments addressed.

ltratt:
Please squash.

vext01:
splat.

PR #344, opened 2021-05-26
Implement compilation of simple traces.

ptersilie:
This is the first stab at our API which takes a PT trace and compiles it
using LLVM. There's still a lot missing here, so the traces we can
successfully compile are very simple.  For example, we can't yet compile
function calls like `printf` or pass values into or out of a trace.

As this may suggest, this limits us in our ability to test that the
compiled function does the right thing, so for now we only test that an
executed trace doesn't segfault.

### DIFF
--- a/c_tests/tests/compile_simple.c
+++ b/c_tests/tests/compile_simple.c
@@ -2,6 +2,8 @@
 // Run-time:
 
 // Check that basic trace compilation works.
+// FIXME An optimising compiler can remove all of the code between start/stop
+// tracing.
 
 #include <assert.h>
 #include <stdio.h>
@@ -13,13 +15,15 @@ int
 main(int argc, char **argv)
 {
     void *tt = __yktrace_start_tracing(HW_TRACING);
-    int res = argc + 1;
+    int res = 1 + 1;
     void *tr = __yktrace_stop_tracing(tt);
 
     assert(res == 2);
-    __yktrace_irtrace_compile(tr); // FIXME test something.
 
+    void *ptr = __yktrace_irtrace_compile(tr);
     __yktrace_drop_irtrace(tr);
+    void(*func)() = (void(*)())ptr;
+    func();
 
     return (EXIT_SUCCESS);
 }
--- a/ykllvmwrap/src/memman.cc
+++ b/ykllvmwrap/src/memman.cc
@@ -1,0 +1,77 @@
+#include <llvm/ExecutionEngine/RTDyldMemoryManager.h>
+#include <sys/mman.h>
+
+using namespace llvm;
+
+struct AllocMem {
+  uint8_t *Ptr;
+  uintptr_t Size;
+};
+
+class MemMan : public RTDyldMemoryManager {
+public:
+  MemMan();
+  ~MemMan() override;
+
+  uint8_t *allocateCodeSection(uintptr_t Size, unsigned Alignment,
+                               unsigned SectionID,
+                               StringRef SectionName) override;
+
+  uint8_t *allocateDataSection(uintptr_t Size, unsigned Alignment,
+                               unsigned SectionID, StringRef SectionName,
+                               bool isReadOnly) override;
+
+  bool finalizeMemory(std::string *ErrMsg) override;
+  void freeMemory();
+
+private:
+  std::vector<AllocMem> code;
+  std::vector<AllocMem> data;
+};
+
+MemMan::MemMan() {}
+MemMan::~MemMan() {}
+
+uint8_t *alloc_mem(uintptr_t Size, unsigned Alignment,
+                   std::vector<AllocMem> *Vec) {
+  uintptr_t RequiredSize = Alignment * ((Size + Alignment - 1) / Alignment + 1);
+  auto Ptr = (unsigned char *)mmap(0, RequiredSize, PROT_WRITE,
+                                   MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+  assert(Ptr != MAP_FAILED);
+  Vec->push_back({Ptr, RequiredSize});
+  return Ptr;
+}
+
+uint8_t *MemMan::allocateCodeSection(uintptr_t Size, unsigned Alignment,
+                                     unsigned SectionID,
+                                     StringRef SectionName) {
+  return alloc_mem(Size, Alignment, &code);
+}
+
+uint8_t *MemMan::allocateDataSection(uintptr_t Size, unsigned Alignment,
+                                     unsigned SectionID, StringRef SectionName,
+                                     bool isReadOnly) {
+  return alloc_mem(Size, Alignment, &data);
+}
+
+bool MemMan::finalizeMemory(std::string *ErrMsg) {
+  for (const AllocMem &Value : code) {
+    if (mprotect(Value.Ptr, Value.Size, PROT_READ | PROT_EXEC) == -1) {
+      errx(EXIT_FAILURE, "Can't make allocated memory executable.");
+    }
+  }
+  return true;
+}
+
+void MemMan::freeMemory() {
+  for (const AllocMem &Value : code) {
+    if (munmap(Value.Ptr, Value.Size) == -1) {
+      errx(EXIT_FAILURE, "Failed to unmap memory.");
+    }
+  }
+  for (const AllocMem &Value : data) {
+    if (munmap(Value.Ptr, Value.Size) == -1) {
+      errx(EXIT_FAILURE, "Failed to unmap memory.");
+    }
+  }
+}
--- a/ykllvmwrap/src/ykllvmwrap.cc
+++ b/ykllvmwrap/src/ykllvmwrap.cc
@@ -22,6 +22,8 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include "memman.cc"
+
 using namespace llvm;
 using namespace llvm::symbolize;
 
@@ -70,15 +72,22 @@ extern "C" void *compile_module(Module *M) {
   InitializeNativeTargetAsmPrinter();
   InitializeNativeTargetAsmParser();
 
+  // FIXME Remember memman or allocated memory pointers so we can free the
+  // latter when we're done with the trace.
+  auto memman = new MemMan();
+
   auto MPtr = std::unique_ptr<Module>(M);
-  ExecutionEngine *EE = EngineBuilder(std::move(MPtr)).create();
+  ExecutionEngine *EE =
+      EngineBuilder(std::move(MPtr))
+          .setMemoryManager(std::unique_ptr<MCJITMemoryManager>(memman))
+          .create();
   EE->finalizeObject();
 
   if (EE->hasError())
     errx(EXIT_FAILURE, "Couldn't compile trace: %s",
          EE->getErrorMessage().c_str());
 
-    return (void *) EE->getFunctionAddress(TRACE_FUNC_NAME);
+  return (void *)EE->getFunctionAddress(TRACE_FUNC_NAME);
 }
 
 // Compile an IRTrace to executable code in memory.
--- a/ykllvmwrap/src/ykllvmwrap.cc
+++ b/ykllvmwrap/src/ykllvmwrap.cc
@@ -4,48 +4,162 @@
 #define _GNU_SOURCE
 #endif
 
-#include <dlfcn.h>
 #include <llvm/DebugInfo/Symbolize/Symbolize.h>
-#include <string.h>
+#include <llvm/ExecutionEngine/ExecutionEngine.h>
+#include <llvm/ExecutionEngine/MCJIT.h>
+#include <llvm/IR/IRBuilder.h>
+#include <llvm/IR/LLVMContext.h>
+#include <llvm/IR/Module.h>
+#include <llvm/IR/Verifier.h>
+#include <llvm/IRReader/IRReader.h>
+#include <llvm/Support/SourceMgr.h>
+#include <llvm/Support/TargetSelect.h>
+#include <llvm/Transforms/Utils/ValueMapper.h>
+
+#include <dlfcn.h>
+#include <err.h>
 #include <link.h>
+#include <stdlib.h>
+#include <string.h>
 
 using namespace llvm;
 using namespace llvm::symbolize;
 
+#define TRACE_FUNC_NAME "__yk_compiled_trace"
+#define YKTRACE_START "__yktrace_start_tracing"
+#define YKTRACE_STOP "__yktrace_stop_tracing"
+
 extern "C" LLVMSymbolizer *__yk_llvmwrap_symbolizer_new() {
-    return new LLVMSymbolizer;
+  return new LLVMSymbolizer;
 }
 
 extern "C" void __yk_llvmwrap_symbolizer_free(LLVMSymbolizer *Symbolizer) {
-    delete Symbolizer;
+  delete Symbolizer;
 }
 
 // Finds the name of a code symbol from a virtual address.
 // The caller is responsible for freeing the returned (heap-allocated) C string.
-extern "C" char *__yk_llvmwrap_symbolizer_find_code_sym(
-    LLVMSymbolizer *Symbolizer,
-    const char *Obj,
-    uint64_t Off)
-{
-    object::SectionedAddress Mod{Off, object::SectionedAddress::UndefSection};
-    auto LineInfo = Symbolizer->symbolizeCode(Obj, Mod);
-    if (auto Err = LineInfo.takeError()) {
-        return NULL;
-    }
+extern "C" char *
+__yk_llvmwrap_symbolizer_find_code_sym(LLVMSymbolizer *Symbolizer,
+                                       const char *Obj, uint64_t Off) {
+  object::SectionedAddress Mod{Off, object::SectionedAddress::UndefSection};
+  auto LineInfo = Symbolizer->symbolizeCode(Obj, Mod);
+  if (auto Err = LineInfo.takeError()) {
+    return NULL;
+  }
 
-    // OPTIMISE_ME: get rid of heap allocation.
-    return strdup(LineInfo->FunctionName.c_str());
+  // OPTIMISE_ME: get rid of heap allocation.
+  return strdup(LineInfo->FunctionName.c_str());
 }
 
-/// Compiles an IRTrace to executable code in memory.
+// Load an LLVM module from an address.
+std::unique_ptr<Module> load_module(LLVMContext &Context, char *Ptr,
+                                    size_t Len) {
+  auto Sf = StringRef(Ptr, Len);
+  auto Mb = MemoryBufferRef(Sf, "");
+  SMDiagnostic Error;
+  auto M = parseIR(Mb, Error, Context);
+  if (!M)
+    errx(EXIT_FAILURE, "Can't load module.");
+  return M;
+}
+
+// Compile a module in-memory and return a pointer to its function.
+extern "C" void *compile_module(Module *M) {
+  InitializeNativeTarget();
+  InitializeNativeTargetAsmPrinter();
+  InitializeNativeTargetAsmParser();
+
+  auto MPtr = std::unique_ptr<Module>(M);
+  ExecutionEngine *EE = EngineBuilder(std::move(MPtr)).create();
+  EE->finalizeObject();
+
+  if (EE->hasError())
+    errx(EXIT_FAILURE, "Couldn't compile trace: %s",
+         EE->getErrorMessage().c_str());
+
+    return (void *) EE->getFunctionAddress(TRACE_FUNC_NAME);
+}
+
+// Compile an IRTrace to executable code in memory.
 //
 // The trace to compile is passed in as two arrays of length Len. Then each
 // (FuncName[I], BBs[I]) pair identifies the LLVM block at position `I` in the
 // trace.
-extern "C" void __ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[], size_t Len) {
-    for (size_t Idx = 0; Idx < Len; Idx++) {
-        // FIXME populate.
-        //auto FuncName = FuncNames[Idx];
-        //auto BB = BBs[Idx];
+//
+// Returns a pointer to the compiled function.
+extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
+                                              size_t Len, char *SecPtr,
+                                              size_t SecSize) {
+  LLVMContext Context;
+  auto DstMod = new Module("", Context);
+
+  // Set up new module for the trace.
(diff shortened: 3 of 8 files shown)


ptersilie:
I probably want to `clang-format` all the C stuff first, but that might mess up @vext01's branch which is based on this. So `ykllvmwrap/src/ykllvmwrap.cc` is currently not formatted, but `ykllvmwrap/src/memman.cc` is.

ptersilie:
@vext01 The second commit is new so you can have a look too.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -37,15 +52,107 @@ extern "C" char *__yk_llvmwrap_symbolizer_find_code_sym(
...
-extern "C" void __ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[], size_t Len) {
+//
+// Returns a pointer to the compiled function.
+extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[], size_t Len, char *SecPtr, size_t SecSize) {
+    LLVMContext Context;
+    auto DstMod = new Module("", Context);
+
+    // Set up new module for the trace.
+    llvm::FunctionType *FType = llvm::FunctionType::get(Type::getVoidTy(Context), false);
+	llvm::Function *DstFunc = llvm::Function::Create(FType, Function::InternalLinkage, TRACE_FUNC_NAME, DstMod);
```
vext01, line 103: Odd indent here.

vext01: And elsewhere.

ptersilie: Yeah, noticed that. But probably not worth fixing manually if we run `clang-format` over it.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -37,15 +52,107 @@ extern "C" char *__yk_llvmwrap_symbolizer_find_code_sym(
...
+        if (!F)
+            errx(EXIT_FAILURE, "can't find function %s", FuncName);
+
+        // Skip to the correct block.
+        auto It = F->begin();
+        std::advance(It, BBs[Idx]);
+        BasicBlock *BB = &*It;
+        // Iterate over all instructions within this block and copy them over
+        // to our new module.
+        for (auto I = BB->begin(), EI = BB->end(); I != EI; I++) {
```
vext01, line 130: Can this just be: ` for (auto I = BB->begin(); I != BB->end(); I++) {`?

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Which reminds me, we also want to setup bors to check the formatting. Maybe we should do that before merging this PR.

c_tests/tests/compile_simple.c
```diff
@@ -13,13 +13,15 @@ int
 main(int argc, char **argv)
 {
     void *tt = __yktrace_start_tracing(HW_TRACING);
-    int res = argc + 1;
+    int res = 1 + 1;
```
ltratt, line 16: Do we compile these tests with optimisations? If so, this will probably be compiled to:

```
void *tt = __yktrace_start_tracing(HW_TRACING);
void *tr = __yktrace_stop_tracing(tt);
assert(2 == 2);
```

where the `assert(2==2)` will also be removed. So this trace will effectively be a no-op.

ykllvmwrap/src/memman.cc
```diff
@@ -0,0 +1,78 @@
...
+  std::vector<AllocMem> code;
+  std::vector<AllocMem> data;
+};
+
+MemMan::MemMan() {}
+MemMan::~MemMan() {}
+
+uint8_t *alloc_mem(uintptr_t Size, unsigned Alignment,
+                   std::vector<AllocMem> *Vec) {
+  uintptr_t RequiredSize = Alignment * ((Size + Alignment - 1) / Alignment + 1);
```
vext01, line 37: Odd indent?

vext01: Or maybe the formatter is using 2 spaces? Which do we want?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -37,15 +52,107 @@ extern "C" char *__yk_llvmwrap_symbolizer_find_code_sym(
...
+    auto Sf = StringRef(Ptr, Len);
+    auto Mb = MemoryBufferRef(Sf, "");
+    SMDiagnostic Error;
+    auto M = parseIR(Mb, Error, Context);
+    if (!M)
+        errx(EXIT_FAILURE, "Can't load module.");
+    return M;
+}
+
+// Compile a module in-memory and return a pointer to its function.
```
ltratt, line 68: Is `void *` the best return type here?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -37,15 +52,107 @@ extern "C" char *__yk_llvmwrap_symbolizer_find_code_sym(
...
+    ExecutionEngine *EE = EngineBuilder(std::move(MPtr)).create();
+	EE->finalizeObject();
+
+    if (EE->hasError())
+        errx(EXIT_FAILURE, "Couldn't compile trace: %s", EE->getErrorMessage().c_str());
+
+    return (void *) EE->getFunctionAddress(TRACE_FUNC_NAME);
+}
+
+// Compiles an IRTrace to executable code in memory.
```
ltratt, line 90: s/Compiles/Compile/

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -37,15 +52,107 @@ extern "C" char *__yk_llvmwrap_symbolizer_find_code_sym(
...
+
+    // Set up new module for the trace.
+    llvm::FunctionType *FType = llvm::FunctionType::get(Type::getVoidTy(Context), false);
+	llvm::Function *DstFunc = llvm::Function::Create(FType, Function::InternalLinkage, TRACE_FUNC_NAME, DstMod);
+    DstFunc->setCallingConv(CallingConv::C);
+    auto DstBB = BasicBlock::Create(Context, "bb0", DstFunc);
+	llvm::IRBuilder<> Builder(Context);
+	Builder.SetInsertPoint(DstBB);
+
+    // Load the binary's bitcode.
```
ltratt, line 109: I think this comment is a bit redundant ;)

ykllvmwrap/src/memman.cc
```diff
@@ -0,0 +1,78 @@
...
+  for (const AllocMem &Value : code) {
+    if (mprotect(Value.Ptr, Value.Size, PROT_READ | PROT_EXEC) == -1) {
+      errx(EXIT_FAILURE, "Can't make allocated memory executable.");
+    }
+  }
+  return true;
+}
+
+void MemMan::freeMemory() {
+  for (const AllocMem &Value : code) {
```
vext01, line 68: Do we want to free the memory? I thought we'd keep the compiled code around?

c_tests/tests/compile_simple.c
ptersilie: When we checked the compiled code it still had a `mov reg, 2`, so it wasn't completely compiled away. So I guess optimisations are not on. I think Edd was experimenting with optimisations but hopefully by then we have more complicated tests that are not so easy to optimise away.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -37,15 +52,107 @@ extern "C" char *__yk_llvmwrap_symbolizer_find_code_sym(
...
+        // Skip to the correct block.
+        auto It = F->begin();
+        std::advance(It, BBs[Idx]);
+        BasicBlock *BB = &*It;
+        // Iterate over all instructions within this block and copy them over
+        // to our new module.
+        for (auto I = BB->begin(), EI = BB->end(); I != EI; I++) {
+            if (isa<CallInst>(I)) {
+                Function *CF = cast<CallInst>(&*I)->getCalledFunction();
+                if (CF->getName() == "__yktrace_start_tracing") {
```
ltratt, line 133: I think these strings should be constants at the top of the module.

ykllvmwrap/src/ykllvmwrap.cc
vext01: Let's leave it until after my branches are in to avoid merge conflicts?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -37,15 +52,107 @@ extern "C" char *__yk_llvmwrap_symbolizer_find_code_sym(
...
+            // operands still reference values in the original bitcode, remap
+            // the operands to point to new values within the IR trace.
+            auto NewInst = &*I->clone();
+            llvm::RemapInstruction(NewInst, VMap, RF_NoModuleLevelChanges);
+            VMap[&*I] = NewInst;
+            Builder.Insert(NewInst);
+        }
     }
+    Builder.CreateRetVoid();
+    // FIXME Only do this for debug builds.
```
ltratt, line 161: I think we should fix this in this PR (it's a one line fix).

yktrace/src/lib.rs
```diff
@@ -65,15 +69,33 @@ impl IRTrace {
...
             bbs.push(blk.bb());
         }
-        unsafe { ykllvmwrap::__ykllvmwrap_irtrace_compile(func_names.as_ptr(), bbs.as_ptr(), len) }
+
+        // Find bitcode section in executable.
+        let pathb = env::current_exe().unwrap();
+        let file = fs::File::open(&pathb.as_path()).unwrap();
+        let mmap = unsafe { memmap2::Mmap::map(&file).unwrap() };
+        let object = object::File::parse(&*mmap).unwrap();
+        let sec = object.section_by_name(".llvmbc").unwrap();
```
ltratt, line 86: Given all these `unwrap`s, I suspect the function signature should be `compile(&self) -> Result<*const c_void, Box<dyn Error>>` and the errors percolated upwards.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: This used to be `uintptr_t`, but we thought a pointer is nicer. Suggestions on what other type we could use here (we don't really know the type of this function yet).

vext01: The only other reasonable type would be a function pointer.

ykllvmwrap/src/memman.cc
```diff
@@ -0,0 +1,78 @@
...
+  std::vector<AllocMem> data;
+};
+
+MemMan::MemMan() {}
+MemMan::~MemMan() {}
+
+uint8_t *alloc_mem(uintptr_t Size, unsigned Alignment,
+                   std::vector<AllocMem> *Vec) {
+  uintptr_t RequiredSize = Alignment * ((Size + Alignment - 1) / Alignment + 1);
+  auto Ptr = (unsigned char *)mmap(0, RequiredSize, PROT_WRITE,
```
ltratt, line 38: We should add some sort of check here maybe just `assert(Ptr != MAP_FAILED)`.

ykllvmwrap/src/memman.cc
```diff
@@ -0,0 +1,78 @@
...
+                                   MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+  Vec->push_back({Ptr, RequiredSize});
+  return Ptr;
+}
+
+uint8_t *MemMan::allocateCodeSection(uintptr_t Size, unsigned Alignment,
+                                     unsigned SectionID,
+                                     StringRef SectionName) {
+  auto Ptr = alloc_mem(Size, Alignment, &code);
```
ltratt, line 47: Is the assignment necessary here rather than just `return alloc_mem(...)`?

ykllvmwrap/src/ykllvmwrap.cc
vext01: I don't think it is. We'd need to pass down the cargo profile to C++ somehow.

ykllvmwrap/src/memman.cc
```diff
@@ -0,0 +1,78 @@
...
+    if (mprotect(Value.Ptr, Value.Size, PROT_READ | PROT_EXEC) == -1) {
+      errx(EXIT_FAILURE, "Can't make allocated memory executable.");
+    }
+  }
+  return true;
+}
+
+void MemMan::freeMemory() {
+  for (const AllocMem &Value : code) {
+    if (munmap(Value.Ptr, Value.Size) != 0) {
```
ltratt, line 69: I think `!= 0` should be `== -1`.

yktrace/src/lib.rs
vext01: I'm torn. Part of me thinks that at this early stage unwrapping is ok.

ykllvmwrap/src/memman.cc
vext01: agreed.

yktrace/src/lib.rs
ltratt: Borking on file system accesses is pretty drastic IMHO.

ykllvmwrap/src/memman.cc
vext01: Agreed:
```
RETURN VALUES
     Upon successful completion, the value 0 is returned; otherwise the
     value -1 is returned and the global variable errno is set to indicate the
     error.
```

vext01: Well, actually either of you is correct.

ykllvmwrap/src/ykllvmwrap.cc
ltratt: It's fine to do something like `#ifdef DEBUG` or whatever we think we *will* be doing. I can be convinced otherwise, but this doesn't feel like it should be too difficult?

ykllvmwrap/src/ykllvmwrap.cc
ltratt: A function pointer sounds like it would make sense although in fairness I guess we don't know the actual signature i.e. the arguments it takes?

ykllvmwrap/src/memman.cc
ptersilie: At some point we'll want to free it. If we decide to keep the memory manager around (i.e. by putting it in a `unique_ptr`) then we'll need this function later. If we only keep the allocated memory vectors, then I guess it's not needed.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -37,15 +56,113 @@ extern "C" char *__yk_llvmwrap_symbolizer_find_code_sym(
...
+//
+// Returns a pointer to the compiled function.
+extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[], size_t Len, char *SecPtr, size_t SecSize) {
+    LLVMContext Context;
+    auto DstMod = new Module("", Context);
+
+    // Set up new module for the trace.
+    llvm::FunctionType *FType = llvm::FunctionType::get(Type::getVoidTy(Context), false);
+	llvm::Function *DstFunc = llvm::Function::Create(FType, Function::InternalLinkage, TRACE_FUNC_NAME, DstMod);
+    DstFunc->setCallingConv(CallingConv::C);
```
bjorn3, line 106: I think this will need a fixme for matching the calling convention of the original function.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -37,15 +56,113 @@ extern "C" char *__yk_llvmwrap_symbolizer_find_code_sym(
...
+                    Tracing = true;
+                    I++; // FIXME Better way of removing result of __yktrace_start_tracing.
+                    continue;
+                } else if (CF->getName() == YKTRACE_STOP) {
+                    // FIXME Remove argument setup before __yktrace_stop_tracing call.
+                    Tracing = false;
+                }
+            }
+            if (!Tracing && !llvm::isa<AllocaInst>(I)) {
+                // FIXME Remove allocas not needed in traced code.
```
bjorn3, line 145: Probably done by LLVM itself already.

yktrace/src/lib.rs
ptersilie: This code is gonna go anyway once we load this stuff lazily (and ideally from memory and not the binary).

ykllvmwrap/src/ykllvmwrap.cc
vext01: Ultimately the pointer has to become a func ptr (with static type) somewhere in order to be called. Currently the test does a cast.

c_tests/tests/compile_simple.c
ltratt: My point is that the test really isn't robust -- the compiler can do whatever it wants to this code -- and if we don't fix it in this PR, we'll probably never fix it. Yes, it's annoying :) One way to do this is to have a function which either is (or looks like!) it's doing side effects being called between start/stop tracing. [We could use a trick along the lines of https://stackoverflow.com/questions/37786547/enforcing-statement-order-in-c/38025837#38025837 but there are other ways too.]

ykllvmwrap/src/ykllvmwrap.cc
ltratt: There's not a single function pointer type and I don't know what the casting rules are for punning between different function pointer types. It is probably (but I'm not sure) undefined behaviour to cast between different (incorrect) function pointer types.

ykllvmwrap/src/memman.cc
ptersilie: Fixed in 43872de.

ykllvmwrap/src/memman.cc
ptersilie: Fixed in 43872de.

ykllvmwrap/src/memman.cc
ptersilie: Fixed in 43872de.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Fixed in 926f02c.

ykllvmwrap/src/memman.cc
ptersilie: Yeah, C++ uses 2 spaces instead of 4 according to the style guide. We can change, it maybe sticking to what `clang-format` does by default causes us less grief in the long term.

ykllvmwrap/src/ykllvmwrap.cc
vext01: I'd imagined the pointer being varargs accepting only pointers.

c_tests/tests/compile_simple.c
vext01: We already had the idea to print to stdout and use lang tester to match it, but this was problematic as we can't yet do handle calls.

ykllvmwrap/src/memman.cc
vext01: I like to stick to defaults so that we don't have to configure anything.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: This is our newly created function which contains our trace, and will be run by the meta-tracer. So I think we do want C or SysV here. I could be wrong though.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Fixed in daa611b.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Fixed in daa611b.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Sounds to me like we should leave it until we have decided what this function is going to look like (which I assume we'll figure out soon when we make traces to a bit more things, e.g. pass in values for reading/writing).

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: I guess this is better than a comment, even if `DEBUG` isn't set anywhere yet: 208fcb6.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Possibly. But it looks like we'll be generating all allocas ourselves anyway soon, so this will go away.

ykllvmwrap/src/ykllvmwrap.cc
vext01: ok

ykllvmwrap/src/ykllvmwrap.cc
ltratt: Works for me.

ykllvmwrap/src/ykllvmwrap.cc
ltratt: I agree with waiting.

c_tests/tests/compile_simple.c
ltratt: Ah, we can't handle *any* function calls? If so, let's mark the test as "broken" or "fixme" and document why the test isn't robust.

ptersilie: Yeah, I alluded a bit to this in the PR description. Something like 7c7aea5 will do?

ltratt:
If @vext01 is OK with this, I think we're ready to squash.

ptersilie:
@vext01 still needs to tell me if he's okay with me `clang-format`ting `ykllvmwrap.cc` or if this is gonna mess up his branch. We can do it a separate PR.

vext01:
Shall we just do it? I'm a big boy, I can deal with merge conflicts :)

ltratt:
[I'm afk. @vext01 please bors this when @ptersilie has squashed.]

ptersilie:
> Shall we just do it? I'm a big boy, I can deal with merge conflicts :)

Done. Tbf you can probably avoid most conflicts by doing a `clang-format` on your end and squashing it down before rebasing.

ptersilie:
Ready to squash?

vext01:
Are we going to check the formatting in CI, or is that a separate PR?

ptersilie:
Not sure. Either way is fine with me.

ptersilie:
Actually, let's merge this first so you can get your stuff going and I'll look at buildbot in the meantime.

vext01:
Good. Please squash.

ptersilie:
Squashed. Though I was undecided whether to squash the last two commits. There's an argument to be made either way.

PR #346, opened 2021-05-26
Make dummy storage for vars that appear in, but are allocated outside of, traces.

vext01:


vext01:
Will need squashing.

c_tests/tests/compile_trace_not_entry_bb.c
```diff
@@ -0,0 +1,34 @@
+// Compiler:
+// Run-time:
+
+// Check that trace compilation works in the non-entry block.
+//
+// Why? Because LLVM allocas typically appear in the entry block, so tracing in
+// another block will means they are not seen in the trace.
```
ptersilie: s/means/mean.

But maybe change this description to something like:

Since LLVM allocas only seem to appear in the entry block of a function, we will miss the allocas if tracing starts in a later block.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -146,10 +146,27 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
...
+
+      // FIXME For any variable defined outside of the trace, we realise
+      // dummy storage space. Eventually, such variables should be inputs
+      // to the trace. This hack prevents the remapper from choking when
+      // it can't find the definition of an instruction operand.
+      if ((llvm::isa<llvm::StoreInst>(I)) && (VMap[&*I] == nullptr)) {
+        Value *DestOp = I->getOperand(1);
+        Value *SrcOp = I->getOperand(0);
+        Instruction *Alloca =
+            Builder.CreateAlloca(SrcOp->getType(), (unsigned)0);
```
ptersilie, line 161: Does it make a difference that these allocas are spread through-out the block[s]. Not sure if it's worth it moving them all into the beginning of first one, similar to what LLVM does.

vext01: The module verified, so I think it's ok. Since this isn't going to be the final design, I don't think it's worth worrying about.

ptersilie:
Please squash.

vext01:
Hold on, i need to address a comment.

vext01:
OK to squash

ptersilie:
Please go ahead.

c_tests/tests/compile_trace_not_entry_bb.c
```diff
@@ -0,0 +1,34 @@
...
+
+int main(int argc, char **argv) {
+  // Causes the traced block to NOT be the entry block.
+  if (argc == -1)
+    abort();
+
+  int res;
+  void *tt = __yktrace_start_tracing(HW_TRACING);
+  // Causes both a load and a store to things defined outside the trace.
+  res = 1 + argc;
```
ltratt, line 23: This test needs a FIXME as, when optimised, this will quite probably turn into:

```
start_tracing();
stop_tracing();
assert(argc == 1);
```

vext01:
splat.

c_tests/tests/compile_trace_not_entry_bb.c
vext01: We can't go on like this. Do we have a solution to stop things floating around?

I have a branch which checks all optimisation levels, but i'd prefer a proper fix.

ykllvmwrap/src/ykllvmwrap.cc
bjorn3: Alloca's outside of the first block will dynamically grow the stackframe each time they are encountered and grow the stackframe multiple times if they are hit multiple times. Alloca's inside the first block will all be grouped together to grow the stackframe in the prologue once.

c_tests/tests/compile_trace_not_entry_bb.c
vext01: is the asm volatile trick acceptable?

ykllvmwrap/src/ykllvmwrap.cc
vext01: I'd expect optimisation passes to push allocations up?

c_tests/tests/compile_trace_not_entry_bb.c
ltratt: > Do we have a solution to stop things floating around?

There is no guaranteed solution. *But* you're *much* less likely to have problems if things are structured along the lines of:

```
fn bench() {
  start_tracing();
  f();
  stop_tracing();
}

#[inline(never)]
fn f() {
  // the benchmark
}
```

perhaps using `asm` tricks and so on. Put another way: having benchmark code in the same function as the timing loop all but guarantees that the optimising compiler will optimise things away.

ptersilie: This is not going to work, because that's exactly what this test is supposed to check. If we put the benchmark into a function, then we will trace the allocas.

vext01: I've recently noticed that LTO won't listen to requests to not inline.

> There is no guaranteed solution

I think we can find platform specific solutions, like using a `MFENCE` and an compiler barrier (i.e. mfence in a volatile asm block).

image

ltratt: Good point. Perhaps this PR is solving the wrong problem in a sense? *Perhaps* what we should actually do is drop the `start_tracing`/`stop_tracing` model and move to a `trace(f)` model which means "give me back a trace of function `f`". That way we don't have to worry about this alloca problem at all, and we can also (probably) stop the compiler optimising `f` into `trace`.

vext01: What are the implications for the VM author?

ltratt: Ouch, yes, sometimes my suggestion would make their life miserable (though sometimes it would work). Scratch that (at least until I've thought more about it).

MFENCE etc. stops *some* optimisations (side effecting things) but not all. You're also right that nothing guarantees to stop inlining, LTO or otherwise. There are no great solutions here. That said, `start_tracing` and `stop_tracing` can easily have compiler barriers added such that side effecting optimisations won't move. However, notice that this test does not have side effects, so compiler barriers won't help.

ykllvmwrap/src/ykllvmwrap.cc
bjorn3: Nope, even with `-O3` I get:

```llvm
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

define void @first_alloca() unnamed_addr #0 {
  %baz = alloca i8, align 1
  br label %bb0

bb0:                       
  store i8 0, i8* %baz, align 1
  call void @foo(i8* noalias readonly align 1 dereferenceable(1) %baz)
  br label %bb1

bb1:                                              ; preds = %bb0
  br label %bb0
}

define void @second_alloca() unnamed_addr #0 {
  br label %bb0

bb0:                       
  %baz = alloca i8, align 1
  store i8 0, i8* %baz, align 1
  call void @foo(i8* noalias readonly align 1 dereferenceable(1) %baz)
  br label %bb1

bb1:                                              ; preds = %bb0
  br label %bb0
}

declare void @foo(i8* noalias readonly align 1 dereferenceable(1)) unnamed_addr

attributes #0 = { nonlazybind uwtable "probe-stack"="__rust_probestack" "target-cpu"="x86-64" }
```

```asm
first_alloca:                           # @first_alloca
        push    rbx
        sub     rsp, 16
        lea     rbx, [rsp + 15]
.LBB0_1:                                # %bb0
        mov     byte ptr [rsp + 15], 0
        mov     rdi, rbx
        call    foo@PLT
        jmp     .LBB0_1
second_alloca:                          # @second_alloca
        push    rbp
        mov     rbp, rsp
.LBB1_1:                                # %bb0
        mov     eax, 16
        call    __rust_probestack
        sub     rsp, rax
        mov     rdi, rsp
        mov     byte ptr [rdi], 0
        call    foo@PLT
        jmp     .LBB1_1
```

c_tests/tests/compile_trace_not_entry_bb.c
ltratt: For this PR, let's just add a FIXME to this test.

ykllvmwrap/src/ykllvmwrap.cc
vext01: Huh! Surprising!

bjorn3: LLVM is optimized for the code patterns clang uses. Clang places alloca's at the first block for regular variables, so I am not too surprised that LLVM doesn't optimize alloca's in other blocks.

vext01:
FIXME added. OK to squash?

ptersilie:
Please do.

vext01:
splat.

PR #347, opened 2021-06-01
Run the C tests with different optimisation levels.

vext01:
We test all of clang's different optimization levels to try and shake
out bugs, which it already has!

Namely:

 * One of our tests was checking an outcome specific to -O0 IR and had
   to have its assertions weakened.

 * Higher optimisation levels exposed bugs in the code that allocates dummy
   storage space for initialisations prior to starting tracing. It assumed
   that: a) we only need to do this for stores and loads (not true); and
   b) that all values are stack allocated (not true). I've fixed this
   and revised the comment to be more descriptive too.

All tests now passing for all optimisation levels.

### DIFF
--- a/c_tests/run.rs
+++ b/c_tests/run.rs
@@ -9,8 +9,8 @@ use tempfile::TempDir;
 
 const COMMENT: &str = "//";
 
-/// Make a compiler command that compiles `src` to `exe`.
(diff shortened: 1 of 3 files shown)


ltratt:
I like it!

PR #348, opened 2021-06-01
Integrate C++ debugging with Cargo.

vext01:
We set NDEBUG for release builds, thus only running assertions and LLVM
module verification in cargo debug builds.

### DIFF
--- a/ykllvmwrap/build.rs
+++ b/ykllvmwrap/build.rs
@@ -1,4 +1,4 @@
-use std::process::Command;
+use std::{env, process::Command};
 
 fn main() {
     // Compile our wrappers with the right LLVM C++ flags.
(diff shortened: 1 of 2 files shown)


ykllvmwrap/build.rs
```diff
@@ -19,6 +19,12 @@ fn main() {
         // Lots of unused parameters in the LLVM headers.
         .flag("-Wno-unused-parameter")
         .cpp(true);
+
+    // Set the C NDEBUG macro if Cargo is building in release mode. This ensures that assertions
+    // (and other things we guard with NDEBUG) only happen in debug builds.
+    if env::var("PROFILE").unwrap() == "release" {
```
ptersilie, line 25: Any reason why we don't call it `DEBUG` and do `!=` here to avoid the double negative `ifndef NDEBUG` below?

vext01: It's annoying, but this is the C way. `NDEBUG` controls libc assertions, so it makes sense to adopt that name, rather than have two knobs.

From `assert(3)`:

>      The assert() macro may be removed at compile time with the cc(1) option -DNDEBUG.

ptersilie: Well, isn't that silly. But if it's the C way, then I guess we should do the same.

PR #349, opened 2021-06-01
Name trace functions uniquely.

vext01:
The first commit is the main change, and I've sneaked in a small (unrelated) second commit.

See individual commit messages.

### DIFF
--- a/ykllvmwrap/src/ykllvmwrap.cc
+++ b/ykllvmwrap/src/ykllvmwrap.cc
@@ -119,7 +119,7 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
   llvm::Function *DstFunc = llvm::Function::Create(
       FType, Function::InternalLinkage, TraceName, DstMod);
   DstFunc->setCallingConv(CallingConv::C);
-  auto DstBB = BasicBlock::Create(Context, "bb0", DstFunc);
+  auto DstBB = BasicBlock::Create(Context, "", DstFunc);
   llvm::IRBuilder<> Builder(Context);
   Builder.SetInsertPoint(DstBB);
 
--- a/ykllvmwrap/src/ykllvmwrap.cc
+++ b/ykllvmwrap/src/ykllvmwrap.cc
@@ -16,8 +16,10 @@
 #include <llvm/Support/TargetSelect.h>
 #include <llvm/Transforms/Utils/ValueMapper.h>
 
+#include <atomic>
 #include <dlfcn.h>
 #include <err.h>
+#include <limits>
 #include <link.h>
 #include <stdlib.h>
 #include <string.h>
@@ -26,10 +28,14 @@
 
 using namespace llvm;
 using namespace llvm::symbolize;
+using namespace std;
 
-#define TRACE_FUNC_NAME "__yk_compiled_trace"
+#define TRACE_FUNC_PREFIX "__yk_compiled_trace_"
 #define YKTRACE_START "__yktrace_start_tracing"
 #define YKTRACE_STOP "__yktrace_stop_tracing"
+
+// An atomic counter used to issue compiled traces with unique names.
+atomic<uint64_t> NextTraceIdx(0);
 
 extern "C" LLVMSymbolizer *__yk_llvmwrap_symbolizer_new() {
   return new LLVMSymbolizer;
@@ -67,7 +73,7 @@ std::unique_ptr<Module> load_module(LLVMContext &Context, char *Ptr,
 }
 
 // Compile a module in-memory and return a pointer to its function.
-extern "C" void *compile_module(Module *M) {
+extern "C" void *compile_module(string TraceName, Module *M) {
   InitializeNativeTarget();
   InitializeNativeTargetAsmPrinter();
   InitializeNativeTargetAsmParser();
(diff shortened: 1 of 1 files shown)


ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -26,11 +28,15 @@
...
 using namespace llvm::symbolize;
+using namespace std;
 
-#define TRACE_FUNC_NAME "__yk_compiled_trace"
+#define TRACE_FUNC_PREFIX "__yk_compiled_trace_"
 #define YKTRACE_START "__yktrace_start_tracing"
 #define YKTRACE_STOP "__yktrace_stop_tracing"
 
+// An atomic counter used to issue compiled traces with unique names.
+atomic<uint64_t> NextTraceIdx(0);
```
ltratt, line 39: Instead of using a counter (when one always has to worry "what happens if this process runs for 70 years and the integer wraps), could we use the address of the trace in memory? That's guaranteed to be stable.

vext01: Good idea.

Since raising the PR, I'm also wondering if there is any need to have a unique name except for when debugging (when you need to set a breakpoint on a specific trace -- lukas and I realised that LLVM cleverly inserts a symbol to help with this). We might choose to only name things uniquely for debug builds...

Thoughts?

vext01: > We might choose to only name things uniquely for debug builds

Unless that causes symbol clashing later, that is. At present I'm not sure. We will find out soon.

ltratt: I have no opinion on whether we need unique names or not. That's up to you!

vext01: I may also shorted `__yk_compiled_trace_$ptr` to `__yk_jitted_$ptr`, where `$ptr` is the hex address of the pointer (as you suggested). OK with that?

vext01: Hold on, I don't know the address of the trace until codegen is done, at which point I'm unable to name the trace. We can allocate a byte just for the purpose of having a unique ID perhaps?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -119,7 +120,7 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
   llvm::Function *DstFunc = llvm::Function::Create(
       FType, Function::InternalLinkage, TraceName, DstMod);
   DstFunc->setCallingConv(CallingConv::C);
-  auto DstBB = BasicBlock::Create(Context, "bb0", DstFunc);
+  auto DstBB = BasicBlock::Create(Context, ENTRY_BLOCK, DstFunc);
```
ptersilie, line 123: I think we could also not give it a name at all by passing `""`. I'm not sure how much more helpful the name for debugging is anyway since the entry block isn't really hard to find.

vext01: The second commit is just for sake of convention. As Laurie said to me last week: "when in Rome" (buy a good sun hat).

ykllvmwrap/src/ykllvmwrap.cc
vext01: Also found this: https://llvm.discourse.group/t/duplicate-symbols-in-separate-modules-no-longer-allowed/1805

So naming our functions uniquely is probably a good idea.

ltratt: If we don't have a simple memory address, then let's just use a counter and add something like:

```
let new_counter = old_counter.fetch_add(1);
if new_counter < old_counter { todo!(); }
```

so we catch the "process has been running for 70 years" case. It's a good habit :)

vext01: We have a similar check [here](https://github.com/ykjit/yk/pull/349/files#diff-28de043b71aaedbd9af7218078abd71f6216cff0ce9dbb53e6518efc62817021R114).

Works for you?

vext01: [I think your solution allows one more trace than mine!]

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Not sure what you are saying. I'm arguing that we don't really need to pass anything here. If this was really just a convention fix, then you shouldn't have changed `bb0` to `entry` either.

vext01: You are right that we don't have to name the block. I only do because I see it named "entry" everywhere. E.g. https://llvm.org/docs/tutorial/MyFirstLanguageFrontend/LangImpl04.html

ptersilie: The few times I inspected LLVM IR, the first block never seemed to have a name.

Anyway, I just wanted to point it out. The decision is ultimately yours.

ykllvmwrap/src/ykllvmwrap.cc
ltratt: Erp, that's a good point. I think the current check you have is correct (and mine is wrong).

vext01:
OK to squash then?

ptersilie:
Please do.

vext01:
splat.

PR #350, opened 2021-06-01
Add clang-format as an xtask.

ptersilie:
Use `cargo xtask cfmt" to format all header, C, and C++ files with clang-format.

Personally I think this is a bit overkill for something that could have been a single line in bash. But apparently this is more "rusty" and I guess would also work in Windows if `clang-format` is installed.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -1,0 +1,48 @@
+use std::env;
+use std::process::Command;
+use walkdir::{DirEntry, WalkDir};
+
+fn ignore_dir(entry: &DirEntry) -> bool {
+    if entry.path().starts_with("./target")
+        || entry.path().starts_with("./.cargo")
+        || entry.path().starts_with("./.git")
+    {
+        return false;
+    }
+    true
+}
+
+fn clang_format() {
+    for entry in WalkDir::new(".")
+        .into_iter()
+        .filter_entry(|e| ignore_dir(e))
+        .filter_map(|e| e.ok())
+    {
+        if !entry.file_type().is_file() {
+            continue;
+        }
+        if let Some(ext) = entry.path().extension() {
+            match ext.to_str().unwrap() {
+                "h" | "c" | "cpp" | "cc" => {
+                    Command::new("clang-format")
+                        .arg("-i")
+                        .arg(entry.path())
+                        .output()
+                        .expect("Failed to execute xtask: cfmt");
+                }
+                _ => {}
+            }
+        }
+    }
+}
+
+fn main() {
+    let task = env::args().nth(1);
+    match task.as_deref() {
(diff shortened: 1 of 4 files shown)


ltratt:
In `.buildbot.sh` can we check that the formatting is stable (i.e. that it doesn't actually change anything)?

vext01:
Oh wow, you actually did the `find` in Rust code! I thought you would do `Command::new("/bin/sh").args(&["-c", "find", ...])`!

Well, you are right, this is more portable!

vext01:
> In .buildbot.sh can we check that the formatting is stable (i.e. that it doesn't actually change anything)?

We already [check the formatting in CI](https://github.com/ykjit/yk/blob/1e81d70f5dfa490fde89164a4944f29350d60b98/.buildbot.sh#L46), but I complained that I wanted a way for developers to format the C code manually.

We could update the CI instructions to use this new cargo command though.

xtask/src/main.rs
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+}
+
+fn main() {
+    let task = env::args().nth(1);
+    match task.as_deref() {
+        Some("cfmt") => clang_format(),
+        _ => println!(
+            "Please specify task to run:
```
vext01, line 44: Is that a newline in the middle of a string? Would it be better to use `\n`?

vext01:
> We could update the CI instructions to use this new cargo command though.

Hrm, maybe that's annoying. We'd need cargo to exit with the exit status of `clang-format`. Is it possible?

ptersilie:
The problem is that this code changes the files in place. Though maybe we could add a switch here that doesn't change the files and just reports the diff?

xtask/src/main.rs
ptersilie: Is this not allowed?

vext01: I dunno! I've never seen it done! @ltratt what do you think?

vext01:
> The problem is that this code changes the files in place.

I guess that's fine as long as a) clang-format returns non-zero if it changed a file, and b) you have the ability to percolate that exit status up so that cargo also exits with this status.

If it's a pain, I'd say just leave the check duplicated. We have bigger fish to fry :)

ptersilie:
> I'd say just leave the check duplicated. We have bigger fish to fry :)

This. Think this is ready for bors.

ltratt:
Ah! Yes, I'd suggest we update `.buildbot.sh` to use this command: we can just do a `git diff --quiet` which will naturally return `1` if there are changes and abort the build. It means that CI and developers will both be formatting the same things in the same way. Given this PR has merged, I guess this should be done in another PR :p

vext01:
Good idea.

PR #351, opened 2021-06-01
Check C formatting in bors using xtask cfmt.

ptersilie:


ptersilie:
bors try

ptersilie:
Okay, this works. But now I'm wondering it's not better to use `git diff --exit-code` instead, so we can see the diff in the bors output.

ltratt:
> But now I'm wondering it's not better to use git diff --exit-code instead, so we can see the diff in the bors output.

Good idea -- let's try it!

ptersilie:
This is ready for review now.

vext01:
Please squash.

ptersilie:
I'm afraid that if I squash any more there won't be any commits left. ;-)

vext01:
Outrageous!

bors r+

PR #352, opened 2021-06-02
Make the trace compilation routine thread safe.

vext01:
`LLVMContext` isn't thread safe and since `Module` holds a reference to a
context, neither is `Module`.

This change uses `ThreadSafeContext`, `ThreadSafeModule` and `std::call_once`
to: a) ensure that the in-binary bitcode is loaded exactly once; b)
share the resulting IR between threads safely, thus allowing parallel
trace compilation.

Added a test to check it works, and that commenting out locking of the
`ThreadSafeContext` causes tests to fail as expected.

Relevant docs:
 - https://llvm.org/docs/ORCv2.html#how-to-use-threadsafemodule-and-threadsafecontext
 - https://en.cppreference.com/w/cpp/thread/call_once

### DIFF
--- a/c_tests/tests/compile_many.c
+++ b/c_tests/tests/compile_many.c
@@ -1,0 +1,32 @@
+// Compiler:
+// Run-time:
+
+// Check that compiling and running multiple traces in sequence works.
+//
+// FIXME An optimising compiler can remove all of the code between start/stop
+// tracing.
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk_testing.h>
+
+void trace(void) {
+  void *tt = __yktrace_start_tracing(HW_TRACING);
+  int res = 1 + 1;
+  void *tr = __yktrace_stop_tracing(tt);
+  assert(res == 2);
+
+  void *ptr = __yktrace_irtrace_compile(tr);
+  __yktrace_drop_irtrace(tr);
+  void (*func)() = (void (*)())ptr;
+  func();
+}
+
+int main(int argc, char **argv) {
+  for (int i = 0; i < 3; i++)
+    trace();
+
+  return (EXIT_SUCCESS);
+}
--- a/ykllvmwrap/src/ykllvmwrap.cc
+++ b/ykllvmwrap/src/ykllvmwrap.cc
@@ -4,6 +4,7 @@
 #define _GNU_SOURCE
 #endif
 
+#include "llvm/ExecutionEngine/Orc/ThreadSafeModule.h"
 #include <llvm/DebugInfo/Symbolize/Symbolize.h>
 #include <llvm/ExecutionEngine/ExecutionEngine.h>
 #include <llvm/ExecutionEngine/MCJIT.h>
@@ -21,14 +22,39 @@
 #include <err.h>
 #include <limits>
 #include <link.h>
+#include <mutex>
 #include <stdlib.h>
 #include <string.h>
 
 #include "memman.cc"
 
 using namespace llvm;
+using namespace llvm::orc;
 using namespace llvm::symbolize;
 using namespace std;
+
+// The bitcode module loaded from the .llvmbc section of the currently-running
+// binary. This cannot be shared across threads and used concurrently without
+// acquiring a lock, and since we do want to allow parallel compilation, each
+// thread takes a copy of this module.
+ThreadSafeModule GlobalAOTMod;
+
+// Flag used to ensure that GlobalAOTMod is loaded only once.
+once_flag GlobalAOTModLoaded;
+
+// A copy of GlobalAOTMod for use by a single thread.
+//
+// A thread should never access this directly, but should instead go via
+// getThreadAOTMod() which deals with the necessary lazy initialisation.
+//
+// OPTIMISE_ME Copying GlobalAOTMod is quite expensive (cloneToNewContext()
+// serialises and deserializes). When a compilation thread dies, we should
+// return its ThreadAOTMod to a pool and transfer ownership to the next thread
+// that needs its own copy of GlobalAOTMod.
+thread_local ThreadSafeModule ThreadAOTMod;
+
+// A flag indicating whether GlobalAOTMod has been copied into the thread yet.
+thread_local bool ThreadAOTModInitialized = false;
 
 #define TRACE_FUNC_PREFIX "__yk_compiled_trace_"
 #define YKTRACE_START "__yktrace_start_tracing"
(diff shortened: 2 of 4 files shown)


c_tests/tests/compile_many.c
```diff
@@ -0,0 +1,29 @@
...
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk_testing.h>
+
+void trace(void) {
+  void *tt = __yktrace_start_tracing(HW_TRACING);
+  int res = 1 + 1;
```
ltratt, line 14: At the risk of flogging a dead horse, this line is likely to be removed entirely by the optimiser.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -21,15 +22,27 @@
 #include <err.h>
 #include <limits>
 #include <link.h>
+#include <mutex>
```
ltratt, line 25: Where does this library (no `.h`?) live?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -61,15 +74,24 @@ __yk_llvmwrap_symbolizer_find_code_sym(LLVMSymbolizer *Symbolizer,
...
+  SharedCtx = std::make_unique<LLVMContext>();
+  auto M = parseIR(Mb, Error, *SharedCtx.getContext());
   if (!M)
     errx(EXIT_FAILURE, "Can't load module.");
-  return M;
+  SharedMod = ThreadSafeModule(std::move(M), SharedCtx);
+}
+
+// Load and return the in-binary LLVM module (or return the cached module if we
+// have already loaded it).
```
ltratt, line 91: Possibly dumb question: is caching this module safe or should we create a new one each time?

ykllvmwrap/src/ykllvmwrap.cc
vext01: It's C++, so no .h suffix.

We can separate out the C and C++ headers if you like?

ykllvmwrap/src/ykllvmwrap.cc
ltratt: Ah, so it's not "an LLVM module in a binary" but "a binary version of the LLVM module"? Should we not use different terminology here (I don't know what the right thing is: bitcode or ... ?)?

ykllvmwrap/src/ykllvmwrap.cc
ltratt: Ah! I don't feel strongly.

ykllvmwrap/src/ykllvmwrap.cc
vext01: It's the bitcode we load from the `.llvmbc` section of the interpreter binary.

So far @ptersilie and I have been using the term `SrcMod` and `DstMod` for the module we load from `.llvmbc` and the module we build and compile for the trace.

ltratt: Aha so we need to bikeshed the names :) "src" and "dst" suggest copying, but we're doing more than that here. How about "AOTMod" and "JITMod"?

c_tests/tests/compile_many.c
ptersilie: Once we can pass arguments into the trace we will go around and update all tests. Before then we don't have much choice here.

ltratt: OK, but let's at least add FIXME to each one that needs fixing.

vext01: Will do.

ykllvmwrap/src/ykllvmwrap.cc
vext01: > "AOTMod" and "JITMod"

I like it!

ptersilie: Works for me.

vext01:
Give this a go. I've realised an inefficiency that I can fix soon -- added a FIXME for that.

ltratt:
Works for me! If @ptersilie is OK with this, you could squash?

ptersilie:
I am. Please squash.

vext01:
splat.

vext01:
OK, well I don't get that locally, but I'm willing to bet it's a bug in my threading logic.

vext01:
OK, so those last two commits fix this PR.

In short: In order to allow parallel compilation, each thread needs a copy of the AOT module within a different context.

LGTY?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -34,12 +34,28 @@ using namespace llvm::symbolize;
...
+// binary. This cannot be shared across threads and used concurrently without
+// acquiring a lock, and since we do want to allow parallel compilation, each
+// thread takes a copy of this module.
 ThreadSafeModule GlobalAOTMod;
 
-// Flag used to ensure that bitcode is loaded only once.
+// Flag used to ensure that GlobalAOTMod is loaded only once.
 once_flag GlobalAOTModLoaded;
 
+// The thread's copy of the GlobalAOTMod.
```
ltratt, line 45: s/The thread's/A thread local/ ?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -34,12 +34,28 @@ using namespace llvm::symbolize;
...
+//
+// A thread should never access this directly, but should instead go via
+// getThreadAOTMod() which deals with the necessary lazy initialisation.
+//
+// OPTIMISE_ME Copying GlobalAOTMod is quite expensive (cloneToNewContext()
```
ltratt, line 50: `OPTIMISE_ME` isn't really standard: `FIXME` is probably better?

vext01: I thought we agreed on this name early on? I prefer to have a separate name so that we know where to look when we want to speed things up.

ltratt: I don't remember it. Also, `OPTIMISE` is a bit problematic because half the world spells it `OPTIMIZE`. Personally I think it's a `FIXME` but I'm not going to worry too much over it: it's not a big deal.

vext01: @ptersilie what to you think?

ykllvmwrap/src/ykllvmwrap.cc
vext01: It should be obvious that it's a thread local from `thread_local`. Crucially, it's a copy of a global variable, which is subtly different.

ltratt: The comment doesn't parse to me at the moment, hence my suggestion. I'm prepared to admit that my suggestion might not be perfect, but I think we need to do do something to clarify matters.

ykllvmwrap/src/ykllvmwrap.cc
bjorn3: [Bevy](https://github.com/bevyengine/bevy) uses `PERF` for this kind of comments.

ykllvmwrap/src/ykllvmwrap.cc
vext01: I'm confused. It parses fine, doesn't it?

A few lines above we have a comment which explains that each thread needs a copy of the `GlobalAOTMod`.

Then here:
> The thread's copy of the GlobalAOTMod.

i.e. The copy of the GlobalAOTMod as owned by the thread.

Call me out if I'm being silly :P

ltratt: It doesn't parse to me.

ykllvmwrap/src/ykllvmwrap.cc
vext01: That works for me too.

ltratt: And I!

ykllvmwrap/src/ykllvmwrap.cc
vext01: Is f5deac clearer?

ltratt: Yes, it's a bit clearer, but it still seems less clear than my suggestion. However, I don't want to fight on this one, so let's go with it.

ltratt:
I suggest we do the `OPTIMISE_ME` -> `PERF` thing (or should we do a new PR for that?) but then we're ready for squashing.

vext01:
Let's do it as a separate commit in this PR.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: `PERF` it is.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -184,6 +235,8 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
             VMap[Op] = Alloca;
           } else {
             // Value is not a stack allocation, so make a dummy default value.
+            OpTy->dump();
```
ptersilie, line 238: Forgotten print.

vext01:
OK to squash?

ltratt:
Please squash.

vext01:
splat.

vext01:
The build didn't start for this. Can we retry?

vext01:
image

I don't know if this has to do with the fastly outages, but I guess we should wait a bit and try again later.

bors r-

vext01:
(that's a bors.tech error btw)

vext01:
[bors should be back now](https://forum.bors.tech/t/bors-down/577/13?u=vext01).

I don't know if we need an r- or not before retry though...

PR #353, opened 2021-06-08
Test compiled traces

ptersilie:
Depends on: https://github.com/ykjit/ykllvm/pull/2

Until now a compiled trace was a bit of a black box as we had no means
to observe what it was doing when run. This change allows us to pass
pointers into a compiled trace which we can then write to, in order to
test that the trace is executing correctly.

### DIFF
--- a/c_tests/tests/compile_multiblocks.c
+++ b/c_tests/tests/compile_multiblocks.c
@@ -12,18 +12,26 @@
 #include <yk_testing.h>
 
 int main(int argc, char **argv) {
+  int cond = argc;
+  void *tt = __yktrace_start_tracing(HW_TRACING, &cond);
   int res = 0;
-  void *tt = __yktrace_start_tracing(HW_TRACING, &res);
-  res = 2;
+  if (cond) {
+    res = 2;
+    cond = 3;
+  } else {
+    res = 4;
+  }
   void *tr = __yktrace_stop_tracing(tt);
+
+  assert(cond == 3);
   assert(res == 2);
 
   void *ptr = __yktrace_irtrace_compile(tr);
   __yktrace_drop_irtrace(tr);
-  void (*func)(int *) = (void (*)(int *))ptr;
-  int res2 = 0;
-  func(&res2);
-  assert(res2 == 2);
+  void (*func)(void *) = (void (*)(void *))ptr;
+  int output = 0;
+  func(&output);
+  assert(output == 3);
 
   return (EXIT_SUCCESS);
 }
--- a/ykllvmwrap/src/ykllvmwrap.cc
+++ b/ykllvmwrap/src/ykllvmwrap.cc
@@ -122,6 +122,28 @@ ThreadSafeModule *getThreadAOTMod(char *Ptr, size_t Len) {
   return &ThreadAOTMod;
 }
 
+std::vector<Value *> get_trace_inputs(Function *F, uintptr_t BBIdx) {
+  std::vector<Value *> Vec;
+  auto It = F->begin();
+  // Skip to the first block in the trace which contains the `start_tracing`
+  // call.
+  std::advance(It, BBIdx);
+  BasicBlock *BB = &*It;
+  for (auto I = BB->begin(); I != BB->end(); I++) {
+    if (isa<CallInst>(I)) {
+      CallInst *CI = cast<CallInst>(&*I);
+      if (CI->getCalledFunction()->getName() == YKTRACE_START) {
+        // Skip first argument to start_tracing.
+        for (auto Arg = CI->arg_begin() + 1; Arg != CI->arg_end(); Arg++) {
+          Vec.push_back(Arg->get());
+        }
+        break;
+      }
+    }
+  }
+  return Vec;
+}
+
 // Compile a module in-memory and return a pointer to its function.
 extern "C" void *compile_module(string TraceName, Module *M) {
   InitializeNativeTarget();
@@ -166,17 +188,39 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
   if (TraceIdx == numeric_limits<uint64_t>::max())
     errx(EXIT_FAILURE, "trace index counter overflowed");
 
+  // Get var args from start_tracing call.
+  auto Inputs = get_trace_inputs(AOTMod->getFunction(FuncNames[0]), BBs[0]);
+
+  std::vector<Type *> InputTypes;
+  for (auto Val : Inputs) {
+    InputTypes.push_back(Val->getType());
+  }
+
+  // Create function to store compiled trace.
   string TraceName = string(TRACE_FUNC_PREFIX) + to_string(TraceIdx);
   llvm::FunctionType *FType =
-      llvm::FunctionType::get(Type::getVoidTy(JITContext), false);
+      llvm::FunctionType::get(Type::getVoidTy(JITContext), InputTypes, false);
   llvm::Function *DstFunc = llvm::Function::Create(
       FType, Function::InternalLinkage, TraceName, JITMod);
   DstFunc->setCallingConv(CallingConv::C);
+
+  // Create entry block and setup builder.
   auto DstBB = BasicBlock::Create(JITContext, "", DstFunc);
   llvm::IRBuilder<> Builder(JITContext);
   Builder.SetInsertPoint(DstBB);
 
   llvm::ValueToValueMapTy VMap;
+  // Variables that are used (but not defined) inbetween start and stop tracing
+  // need to be replaced with function arguments which the user passes into the
+  // compiled trace. This loop creates a mapping from those original variables
+  // to the function arguments of the compiled trace function.
+  for (size_t Idx = 0; Idx != Inputs.size(); Idx++) {
+    Value *OldVal = Inputs[Idx];
+    Value *NewVal = DstFunc->getArg(Idx);
+    assert(NewVal->getType()->isPointerTy());
+    VMap[OldVal] = NewVal;
+  }
+
   bool Tracing = false;
 
   // Iterate over the PT trace and stitch together all traced blocks.
@@ -214,6 +258,11 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
       if (!Tracing)
         continue;
 
+      if (llvm::isa<llvm::BranchInst>(I)) {
+        // FIXME Replace all branch instruction with guards.
+        continue;
+      }
+
       // If execution reaches here, then the instruction I is to be copied into
       // JITMod. We now scan the instruction's operands checking that each is
       // defined in JITMod. Any variable not defined means that the
@@ -234,17 +283,21 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
                 OpTy->getPointerElementType(), OpTy->getPointerAddressSpace());
             VMap[Op] = Alloca;
           } else {
-            // Value is not a stack allocation, so make a dummy default value.
-            // FIXME fails for meta-data types (when you build with -g).
-            Value *NullVal = Constant::getNullValue(OpTy);
-            VMap[Op] = NullVal;
+            if (OpTy->isIntegerTy()) {
+              // Value is an integer constant, so leave it as is.
+              // FIXME Extend this with other types as needed to get new
+              // tests to run. Ultimately, find a better way to do this.
+              VMap[Op] = Op;
+              continue;
+            } else {
+              // Value is not a stack allocation or constant, so make a dummy
+              // default value.
+              // FIXME fails for meta-data types (when you build with -g).
+              Value *NullVal = Constant::getNullValue(OpTy);
+              VMap[Op] = NullVal;
+            }
           }
         }
-      }
-
-      if (llvm::isa<llvm::BranchInst>(I)) {
-        // FIXME Replace all branch instruction with guards.
-        continue;
       }
 
       // Copy instruction over into the IR trace. Since the instruction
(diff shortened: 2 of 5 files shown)


yktrace/src/hwt/mapper.rs
```diff
@@ -51,20 +51,21 @@ impl BlockMap {
...
             let f_off = crsr.read_u64::<NativeEndian>().unwrap();
             let n_blks = leb128::read::unsigned(&mut crsr).unwrap();
-            for bb in 0..n_blks {
+            for _ in 0..n_blks {
                 let b_off = leb128::read::unsigned(&mut crsr).unwrap();
                 // Skip the block size. We still have to parse the field, as it's variable-size.
                 let b_sz = leb128::read::unsigned(&mut crsr).unwrap();
                 // Skip over block meta-data.
                 crsr.seek(SeekFrom::Current(1)).unwrap();
+                let b_idx = leb128::read::unsigned(&mut crsr).unwrap();
```
vext01, line 60: So do we definitely have a strategy for dealing with the case where a `BasicBlock` lowers to many `MachineBasicBlock`s?

Can you think of any scenario where it could be ambiguous?

For example, if a trace reports the blocks `<1, 2, 1, 1, 4>`, do we always know how to interpret `1, 1`? This could either be a block with a call, or a block which lowered to 2 MBBs.

ptersilie: > For example, if a trace reports the blocks `<1, 2, 1, 1, 4>`, do we always know how to interpret `1, 1`? This could either be a block with a call, or a block which lowered to 2 MBBs.

Good point. I think we might be able to encode this into the address map, i.e. when we a machine block maps to an IR block we've already seen, then don't emit the index. This way we know when we see an MBB without an index, that it comes from a split BB. Though I'm not sure if there are other surprises with split up BBs but we won't know until we see some in action.

c_tests/tests/compile_simple.c
```diff
@@ -12,16 +12,18 @@
 #include <yk_testing.h>
 
 int main(int argc, char **argv) {
-  void *tt = __yktrace_start_tracing(HW_TRACING);
-  int res = 1 + 1;
+  int res = 0;
+  void *tt = __yktrace_start_tracing(HW_TRACING, &res);
+  res = 1 + 1;
```
ltratt, line 17: I suspect that even in `-O0` this will become `res = 2`.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -72,6 +72,29 @@ std::unique_ptr<Module> load_module(LLVMContext &Context, char *Ptr,
...
+  std::advance(It, BBIdx);
+  BasicBlock *BB = &*It;
+  for (auto I = BB->begin(); I != BB->end(); I++) {
+    if (isa<CallInst>(I)) {
+      CallInst *CI = cast<CallInst>(&*I);
+      if (CI->getCalledFunction()->getName() == YKTRACE_START) {
+          auto Arg = CI->arg_begin();
+          // Skip first argument to start_tracing.
+          Arg++;
+          for (; Arg != CI->arg_end(); Arg++) {
```
ltratt, line 89: Would this be simpler as:

```c++
for (auto Arg = CI->arg_begin() + 1; ...)
```

?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -72,6 +72,29 @@ std::unique_ptr<Module> load_module(LLVMContext &Context, char *Ptr,
...
+  auto It = F->begin();
+  // Skip to the first block in the trace which contains the `start_tracing`
+  // call.
+  std::advance(It, BBIdx);
+  BasicBlock *BB = &*It;
+  for (auto I = BB->begin(); I != BB->end(); I++) {
+    if (isa<CallInst>(I)) {
+      CallInst *CI = cast<CallInst>(&*I);
+      if (CI->getCalledFunction()->getName() == YKTRACE_START) {
+          auto Arg = CI->arg_begin();
```
ltratt, line 86: We seem to have moved from 2-indent to 4-indent here?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -183,18 +233,21 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
                 OpTy->getPointerElementType(), OpTy->getPointerAddressSpace());
             VMap[Op] = Alloca;
           } else {
-            // Value is not a stack allocation, so make a dummy default value.
-            Value *NullVal = Constant::getNullValue(OpTy);
-            VMap[Op] = NullVal;
+              if (OpTy->isIntegerTy()) {
+                  // Value is an integer constant, so leave it as is.
```
ltratt, line 237: Weird indent.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -113,19 +136,41 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
...
+
+  // Create entry block and setup builder.
   auto DstBB = BasicBlock::Create(Context, "", DstFunc);
   llvm::IRBuilder<> Builder(Context);
   Builder.SetInsertPoint(DstBB);
 
-  auto SrcMod = load_module(Context, SecPtr, SecSize);
-
   llvm::ValueToValueMapTy VMap;
+  // Map trace inputs to their appropriates variables in the trace.
```
ltratt, line 163: s/appropriates/appropriate/

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -113,19 +136,41 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
...
+  for (size_t Idx = 0; Idx != Inputs.size(); Idx++) {
+    Value* OldVal = Inputs[Idx];
+    Value* NewVal = DstFunc->getArg(Idx);
+    if (NewVal->getType()->isPointerTy()) {
```
ltratt, line 167: I assumed this would be correct by construction? In other words we probably want:

```c++
assert(NewVal->getType()->isPointerTy());
VMap[OldVal] = NewVal;
```
```

File: ykllvmwrap/src/ykllvmwrap.cc [hunk 8]
```diff
@@ -113,19 +136,41 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
...
   auto DstBB = BasicBlock::Create(Context, "", DstFunc);
   llvm::IRBuilder<> Builder(Context);
   Builder.SetInsertPoint(DstBB);
 
-  auto SrcMod = load_module(Context, SecPtr, SecSize);
-
   llvm::ValueToValueMapTy VMap;
+  // Map trace inputs to their appropriates variables in the trace.
+  for (size_t Idx = 0; Idx != Inputs.size(); Idx++) {
+    Value* OldVal = Inputs[Idx];
```
ltratt, line 165: In C (I don't know about C++) the convention is that the `*` goes next to the variable name so `Value *OldVal` to avoid the problem that `X* y, z` defines `y` as type `X*` and `z` as `X` (seriously!).

ykllvmwrap/src/ykllvmwrap.cc
ltratt: Also I'm not sure if "appropriate variables" is correct? Aren't they being mapped to function parameters?

ykllvmwrap/src/ykllvmwrap.cc
vext01: I prefer putting the star next to the variable name (not the type) too.

yktrace/src/hwt/mapper.rs
vext01: OK, sounds like we will have to cross that bridge if and when we come to it.

c_tests/tests/compile_simple.c
bjorn3: Probably. LLVM does mandatory const folding during ir generation, so if no alloca are used for the temp results, this will always be const folded.

ptersilie: Yeah, it does, but that doesn't change the outcome of this test. But I can change it to `res = 2` if you like.

ltratt: We should change it because `1+1` is kinda misleading in the context of what we care about (we'll probably only ever see IR for `2`).

yktrace/src/hwt/mapper.rs
ltratt: Can we not test for this at the moment because we don't support function calls?

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: In the original code we have variables. When we construct the trace we then replace those variables with the trace inputs (which are passed as function arguments). I guess this could be reworded.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: You're right. I had at some point considered allowing the user to pass things by value, but then realised that just muddies the waters as it allows users to write to a variable in their original code, but then the trace can't write to it, which is just confusing. So I agree we should enforce that users can only pass pointers.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Whoops. That's a typo. Although this is the first time I hear a good reason why this is the standard, as it previously didn't make any sense to me (semantically the pointer is part of the type).

yktrace/src/hwt/mapper.rs
ptersilie: Yeah, we can't do function calls atm.

c_tests/tests/compile_simple.c
ptersilie: Fixed in 7731d53.

yktrace/src/hwt/mapper.rs
ltratt: OK.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: I think I initially tried `for (auto Arg = CI->arg_begin()++, ...` which didn't work. Your suggestion works though.
Fixed in 7b5d9d9.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: I'll do a formatting path after squashing if that's ok. That'll make my life easier.

ltratt: If you mean "after everything else but before squashing", that's fine ;)

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Try bf1a44c.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Fixed in 815ed73.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Fixed in 18b0420.

ptersilie:
I've also just added a new test, that I forgot about. This was what initially caused these changes, as I was trying to see if tests with multiple blocks work. Since it works out of a box I don't think we need a separate PR, so can just add it along with this one.

ptersilie:
> I'm not sure the last sentence ("the mapping...") adds anything? Otherwise this looks OK to me.

Fixed in 9ccc730 .

ptersilie:
Okay to squash/fmt/rebase?

ltratt:
Please push a format commit. I also suggest you do a merge commit against master to deal with the conflict now.

c_tests/tests/compile_multiblocks.c
```diff
@@ -0,0 +1,37 @@
...
+// tracing.
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk_testing.h>
+
+int main(int argc, char **argv) {
+  int cond = 2;
```
ltratt, line 15: If, for example, you make this test depend on `argc` rather than the constant `cond` you'll make it harder (not impossible!) for the compiler to optimise things away (though it will still potentially move them around).

ptersilie:
> Please push a format commit. I also suggest you do a merge commit against master to deal with the conflict now.

I can do, but after squashing I'll just remove the commit and then format each commit on their own, so adding it now feels a bit pointless. I will likely do the same for the merge commit.

c_tests/tests/compile_multiblocks.c
ptersilie: I'm not quite sure how we can pass things into those tests within our test suite. There currently doesn't seen to be an option for this.

ltratt:
We don't do changes when we squash, so you have to add a format commit now. You can then split it back up (e.g. with `git absorb` later) later.

c_tests/tests/compile_multiblocks.c
ltratt: You don't have to pass anything in: `argc` is a "true" variable that the compiler can't guess is a constant.

ptersilie: Fixed in 47f0ad0.

ptersilie:
Okay to squash?

ltratt:
Please squash.

ptersilie:
Squashed, though we have to wait for https://github.com/ykjit/ykllvm/pull/2 before we can merge.

ptersilie:
This should be good to go now.

PR #354, opened 2021-06-09
Simplify reading in the (raw) .llvmbc section.

vext01:
61cfd3a is the main commit. See commit messages.

### DIFF
--- a/c_tests/run.rs
+++ b/c_tests/run.rs
@@ -28,6 +28,8 @@ fn mk_compiler(exe: &Path, src: &Path, opt: &str) -> Command {
 
     compiler.args(&[
         opt,
+        "-Werror",
+        "-Wall",
         "-fuse-ld=lld",
         "-flto",
         "-Wl,--plugin-opt=-lto-embed-bitcode=optimized",
--- a/ykcapi/Cargo.toml
+++ b/ykcapi/Cargo.toml
@@ -13,6 +13,7 @@ libc = "0.2.94"
 ykllvmwrap = { path = "../ykllvmwrap" }
 ykrt = { path = "../ykrt" }
 yktrace = { path = "../yktrace" }
+ykutil = { path = "../ykutil" }
 
 [features]
 c_testing = []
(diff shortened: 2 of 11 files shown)


ltratt:
This may be a very silly question but I had always vaguely thought (without ever bothering to check) that the whole executable is in a sense mmap'd into memory by default? Or does the linker actually only load in the sections that "do something useful"?

vext01:
The linker will only load in the segments marked `PT_LOAD`, so not the whole binary.

Lukas and I spent a while (a month or so ago) trying to get the bitcode to be marked in such a way, but ended up doing it this way to just get going.

We could add a `PERF` comment actually?

vext01:
Any further comments on this one?

ltratt:
> We could add a PERF comment actually?

Good idea!

vext01:
I've added a couple of comments, as there are two places where we could apply a similar optimisation.

vext01:
OK to squash?

ltratt:
Please squash.

vext01:
Splat.

vext01:
Looks like our threading still isn't right.

vext01:
I'm unable to reproduce this, even when bumping the loop bounds in the test and having it run for hours.

What do you think we should do? Try again and keep a close eye on this?

ltratt:
Oh hang on.

bors r-

ltratt:
Sorry, I thought you meant this was some sort of weird OS error. But it's one of our tests failing. I fear that we need to work out what's going on here. Let's try again to see if at least happens again under bors, which might help give you some clues?

bors try

vext01:
19ae71a seems to fix the threading issue.

9a5e3d0 is mostly unrelated, but it might have made debugging easier.

I need to sync the branch before this will merge due to Lukas' block map change. OK to do that with the squash?

ltratt:
Please squash.

vext01:
2e492ea deals with the braces. OK to squash and sync?

ltratt:
Please squash.

vext01:
splat.

PR #355, opened 2021-06-11
Correctly stitch together PT traces containing function calls.

ptersilie:
This PR ensures that blocks in PT traces are stitched together correctly. This is necessary because calls in LLVM IR are not terminators and thus PT traces an extra block that isn't needed. For example if we have two functions A and B, where A calls B, a PT trace would be A B A.

Luckily, using LLVM has made implementing this a lot easier than in our old system. The "hardest" of these commits is 2e47b79 which implements the stitching together of functions inlined by the trace. We have to be careful here since calls are no longer terminators in LLVM bitcode, which means we see the same block twice after a call. See the commit message for more information. The remaining commits implement inlining of function returns (25ff397) and arguments (1828eef).

### DIFF
--- a/c_tests/tests/compile_call_args.c
+++ b/c_tests/tests/compile_call_args.c
@@ -1,5 +1,11 @@
 // Compiler:
 // Run-time:
+//   stderr:
+//     define internal void @__yk_compiled_trace_0(i32* %0) {
+//       store i32 5, i32* %0, align 4, !tbaa !0
+//       ret void
+//     }
+//     ...
 
 // Check that basic trace compilation works.
 // FIXME An optimising compiler can remove all of the code between start/stop
--- a/c_tests/tests/compile_call_args.c
+++ b/c_tests/tests/compile_call_args.c
@@ -1,0 +1,34 @@
+// Compiler:
+// Run-time:
+
+// Check that basic trace compilation works.
+// FIXME An optimising compiler can remove all of the code between start/stop
+// tracing.
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk_testing.h>
+
+__attribute__((noinline)) int f(int a, int b) {
+  int c = a + b;
+  return c;
+}
+
+int main(int argc, char **argv) {
+  int res = 0;
+  void *tt = __yktrace_start_tracing(HW_TRACING, &res);
+  res = f(2, 3);
+  void *tr = __yktrace_stop_tracing(tt);
+  assert(res == 5);
+
+  void *ptr = __yktrace_irtrace_compile(tr);
+  __yktrace_drop_irtrace(tr);
+  void (*func)(void *) = (void (*)(void *))ptr;
+  int output = 0;
+  func(&output);
+  assert(output == 5);
+
+  return (EXIT_SUCCESS);
+}
--- a/c_tests/tests/compile_calls.c
+++ b/c_tests/tests/compile_calls.c
@@ -1,0 +1,31 @@
+// Compiler:
+// Run-time:
+
+// Check that basic trace compilation works.
+// FIXME An optimising compiler can remove all of the code between start/stop
+// tracing.
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk_testing.h>
+
+void f() { int b = 1; }
+
+int main(int argc, char **argv) {
+  int res = 0;
+  void *tt = __yktrace_start_tracing(HW_TRACING, &res);
+  f();
+  void *tr = __yktrace_stop_tracing(tt);
+  assert(res == 0);
+
+  void *ptr = __yktrace_irtrace_compile(tr);
+  __yktrace_drop_irtrace(tr);
+  void (*func)(void *) = (void (*)(void *))ptr;
+  int output = 0;
+  func(&output);
+  assert(output == 0);
+
+  return (EXIT_SUCCESS);
+}
--- a/c_tests/tests/compile_calls.c
+++ b/c_tests/tests/compile_calls.c
@@ -11,21 +11,24 @@
 #include <string.h>
 #include <yk_testing.h>
 
-void f() { int b = 1; }
+__attribute__((noinline)) int f() {
+  int b = 2;
+  return b;
+}
 
 int main(int argc, char **argv) {
   int res = 0;
   void *tt = __yktrace_start_tracing(HW_TRACING, &res);
-  f();
+  res = f();
   void *tr = __yktrace_stop_tracing(tt);
-  assert(res == 0);
+  assert(res == 2);
 
   void *ptr = __yktrace_irtrace_compile(tr);
   __yktrace_drop_irtrace(tr);
   void (*func)(void *) = (void (*)(void *))ptr;
   int output = 0;
   func(&output);
-  assert(output == 0);
+  assert(output == 2);
 
   return (EXIT_SUCCESS);
 }
--- a/ykllvmwrap/src/ykllvmwrap.cc
+++ b/ykllvmwrap/src/ykllvmwrap.cc
@@ -251,7 +251,8 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
         continue;
       }
       if (isa<CallInst>(I)) {
-        Function *CF = cast<CallInst>(&*I)->getCalledFunction();
+        CallInst *CI = cast<CallInst>(&*I);
+        Function *CF = CI->getCalledFunction();
         if (CF == nullptr)
           continue;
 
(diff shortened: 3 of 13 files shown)


vext01:
LGTM

c_tests/tests/compile_calls.c
```diff
@@ -0,0 +1,31 @@
...
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk_testing.h>
+
+void f() { int b = 1; }
+
+int main(int argc, char **argv) {
+  int res = 0;
```
ltratt, line 20: This variable is pointless as it's not assigned to. Either we should remove it, or have `f` return a value which we assign to `res`?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -236,9 +238,18 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
     auto It = F->begin();
     std::advance(It, BBs[Idx]);
     BasicBlock *BB = &*It;
+
     // Iterate over all instructions within this block and copy them over
     // to our new module.
     for (auto I = BB->begin(); I != BB->end(); I++) {
+      // If we've returned from a call skip ahead to the instruction where we
+      // left off.
```
ltratt, line 246: If we have multiple calls to the same function in the same basic block, does this handle them correctly?

c_tests/tests/compile_calls.c
```diff
@@ -11,21 +11,24 @@
 #include <string.h>
 #include <yk_testing.h>
 
-void f() { int b = 1; }
+int f() {
```
ltratt, line 14: It's probably worth marking this function as "don't align" -- it's not *guaranteed* to work, but it does make it harder for the optimiser to inline it!

c_tests/tests/compile_calls.c
vext01: it's passed by reference to `__yktrace_start_tracing`.

vext01: (which mutates it)

c_tests/tests/compile_call_args.c
```diff
@@ -0,0 +1,34 @@
...
+// FIXME An optimising compiler can remove all of the code between start/stop
+// tracing.
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk_testing.h>
+
+int f(int a, int b) {
```
ltratt, line 14: Again, probably worth marking as "don't inline".

c_tests/tests/compile_calls.c
ltratt: So it does... but now I'm not sure why!?

vext01: I think this is the way lukas is getting stuff into and out of the trace (for now).

But lukas, if this does mutate `res` (to 0) we should initialise it to something other than 0? Otherwise how do we know it has been mutated?

ptersilie: Ah, in this test `res` could go. It's not needed as this test only checks that an inlined function doesn't crash the trace compiler. In the next commit however, we alter this test to write the return value of `f()` to it, so we might as well leave it here too.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: It should, but I will add another test to make sure.

c_tests/tests/compile_calls.c
ltratt: OK (not ideal, but probably not worth fixing). Can we also say why `res` is passed to `__yktrace_start_tracing`?

ptersilie: Ah, yes, that definitely deserves a clarifying comment which probably should have gone into #353.

ltratt: OK, can we add a comment commit into this PR because I must admit that I missed it there!

c_tests/tests/compile_calls.c
ptersilie: Fixed in 14dca0be.

c_tests/tests/compile_call_args.c
ptersilie: Fixed in 373a19d.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Added another test in 3a513a3.

ltratt:
I've just realised something that I should have realised earlier: the tests for this PR don't seem to meet the functionality. The PR says "implement function inlining" but the tests simply test the outcome of tracing: that might, or might not, have inlined anything AFAICT?

I *think* what we need here are tests which print out the IR of the programs being compiled so that we can check that inlining can occur. In a sense, that sort of test is what https://crates.io/crates/fm was designed to make a bit easier.

vext01:
Can @ptersilie do that in a separate PR perhaps?

ptersilie:
> I've just realised something that I should have realised earlier: the tests for this PR don't seem to meet the functionality. The PR says "implement function inlining" but the tests simply test the outcome of tracing: that might, or might not, have inlined anything AFAICT?
> 
> I _think_ what we need here are tests which print out the IR of the programs being compiled so that we can check that inlining can occur. In a sense, that sort of test is what https://crates.io/crates/fm was designed to make a bit easier.

I guess the description "Implement inlining" isn't really accurate as we are not implementing inlining ourselves (the tracer inlines stuff for us), but we make sure that when the tracer traced across multiple functions, we stitch the traces together correctly.

But I agree, that it won't hurt to add an FM test to check that the IR reflects this. On the other hand I'm not sure if/how the IR would look different when we do the inlining after tracing, or when LLVM has inlined things already before tracing.

ltratt:
I think we might be using "inlining" ambiguously. I *think* that what this PR is doing is creating "TIR" which inlines things (as tracing naturally does) and then handing that inlined thing off to LLVM -- which may itself perform more inlining. We don't want to try and test what LLVM does, as that will change very often -- what we want to do is test the IR *we* pass to LLVM.

vext01:
I think what Lukas is saying, is that there's no guarantee that LLVM hasn't already inlined things in the AOT module?

vext01:
My counter is that the `-O0` variant of the suite is unlikely to have inlined anything. The code LLVM emits at `-O0` is so naive it hurts :P

ptersilie:
I agree that if LLVM has already inlined things in the AOT module, then the current tests won't show us that our TIR code is working as there are no functions to inline during TIR. However, while FM is more likely to show discrepancies between our inlined code and LLVMs inlined code, there's no guarantee both approaches produce the same IR, so even there may be cases that even FM can't reliably test.

vext01:
I think we can live with that.

ptersilie:
See 35ec6ef for an attempt to match the IR of compiled traces using lang_tester.

ltratt:
Here's an idea which might or might not work. For e.g. `compile_simple.c.O0` you could avoid duplicating the C code by `#include "compile_simple.c"` I think?

ptersilie:
> Here's an idea which might or might not work. For e.g. `compile_simple.c.O0` you could avoid duplicating the C code by `#include "compile_simple.c"` I think?

Whoops, the code doesn't need to be duplicated actually. I just forgot to delete it for this file. lang_tester only reads `compile_simple.c.O0` for the pattern, but executes `compile_simple.c`. See the other files for examples. Will fix.

ptersilie:
This should do the trick.

ltratt:
Please squash.

ptersilie:
Squashed.

ptersilie:
Whoops. Formatting fixed.

ltratt:
Please squash.

ptersilie:
Squashed.

ptersilie:
That's hopefully it!

ltratt:
Please squash.

ptersilie:
:crossed_fingers:

ptersilie:
Damn, well at least it's not a formatting error. Seems the environment variable is not working on bors. Looking into it.

ptersilie:
Ah, I see what's going on. We are running tests on bors in release mode, but I put the `YKDEBUG_PRINT_IR` check behind a `NDEBUG` pp check (https://github.com/ykjit/yk/pull/355/commits/898c9efa65d31ee61ce602a81eab5cae9660b5e6#diff-28de043b71aaedbd9af7218078abd71f6216cff0ce9dbb53e6518efc62817021R353) as I didn't want to do this environment check in a release version. Thoughts?

ltratt:
Good question. TBH, in the long run we probably want to have some sort of internal boolean controlling IR printing (which the env var would set to true, but you can maybe set/unset it other ways too) which suggests it shouldn't be behind `NDEBUG`. I think we should therefore take it out of `NDEBUG` and make our lives easier here.

ptersilie:
bors try

ltratt:
Please squash.

ptersilie:
Squashed.

PR #358, opened 2021-06-15
 Copy over global variables if needed.

ptersilie:
If a trace references global variables they need to be copied over from
 the AOTModule. This happens in two steps: we first recreate any global
 variable we encounter inside the trace and add it to the JITModule;
 then once the JITModule is finished, we set their initialisers.
 Typically, cloning a module also involves copying a global variable's
 linkage, metadata, and comdat, but until we know for certain that it's
 needed we skip this step.


ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -327,6 +328,18 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
             Value *Alloca = Builder.CreateAlloca(
                 OpTy->getPointerElementType(), OpTy->getPointerAddressSpace());
             VMap[Op] = Alloca;
+          } else if (isa<GlobalVariable>(Op)) {
+            // If there's a reference to a GlobalVariable, copy it over to the
+            // new module.
+            GlobalVariable *OldGV = cast<GlobalVariable>(Op);
+            GlobalVariable *GV = new GlobalVariable(
```
vext01, line 335: Should we check if the global is already defined and skip this step if so? Perhaps we need a test where we trace two separate functions that reference the same global.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -355,6 +368,18 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
     }
   }
   Builder.CreateRetVoid();
+
+  // Fix initialisers/referrers for copied global variables.
+  // FIXME Do we also need to copy Linkage, MetaData, Comdat?
+  for (GlobalVariable *G : cloned_globals) {
+    GlobalVariable *NewGV = cast<GlobalVariable>(VMap[G]);
+    if (G->isDeclaration())
```
vext01, line 376: What's this line about?

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Very good point. We definitely need a test for that. I'll try to construct one.

bjorn3:
I presume that at least mutable global variables need to be declared but not defined. Otherwise you would get inconsistencies between what the baseline and jitted code reads from globals.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Not entirely sure (this is mostly copied from LLVM's `cloneModule` function). But I assume that if it's a declaration you don't need to initialise it.

ltratt:
Would I be an idiot if I said that I couldn't see any global variables being used in the test?

ptersilie:
> Would I be an idiot if I said that I couldn't see any global variables being used in the test?

You wouldn't. It's kind of hidden. LLVM turns those `if`s into a switch table (which is a global variable) and reads the return values from there using `getelementptr`.

vext01:
The `if`s are translated to a jump table stored in a global var.

ptersilie:
> I presume that at least mutable global variables need to be declared but not defined. Otherwise you would get inconsistencies between what the baseline and jitted code reads from globals.

Not sure if that's what you meant, but your comment made me realise: what if we write to a mutable global variable. If we write to it inside a compiled trace, we won't see these changes once we exit the compiled trace via a guard failure. So we either need to make sure, that we write those changes back to the AOTModule after a guard failure, or somehow get the compiled trace to write directly to the global inside the AOTModule, without cloning it.

ykllvmwrap/src/ykllvmwrap.cc
vext01: Ah. I vaguely remember being confused by "define" vs "declare" when I was working in rustc. @bjorn3's comment reminded me.

We should double check that we understand what's going on there.

ltratt:
OK. I'm fine with having a test with a jump table but I think:

1. We need to document in the test what we're expecting to happen.
2. We need to check the LLVM IR to check that a global jump table really was generated (they could change this at any point).
3. Why don't we have a second test that uses a normal global variable :)

Also I notice that the PR says "copy over global variables", but isn't that going to do the wrong thing with mutable variables?

vext01:
> If we write to it inside a compiled trace, we won't see these changes once we exit the compiled trace via a guard failure.

Good point...

This is reminiscent of how you already ensure that locals are passed in. Should the global be passed by pointer into the trace?

bjorn3:
> Not sure if that's what you meant, but your comment made me realise: what if we write to a mutable global variable. If we write to it inside a compiled trace, we won't see these changes once we exit the compiled trace via a guard failure.

Exactly!

> So we either need to make sure, that we write those changes back to the AOTModule after a guard failure, or somehow get the compiled trace to write directly to the global inside the AOTModule, without cloning it.

The second option here was my suggestion. By declaring but not defining the global, it should resolve to the global in the main program executable I think. (if the global is not private)

ptersilie:
Okay, loads of good comments. There's clearly some extra investigation and work needed here.

vext01:
>  By declaring but not defining the global, it should resolve to the global in the main program executable I think. (if the global is not private)

Ah! So `declare` is like `extern` in C? Makes sense. That would make things easier.

ltratt:
It sounds like reckon we want at least three separate tests:

1. Reading from a global variable.
2. Reading and writing from a global variable.
3. Jump tables.

ptersilie:
> The second option here was my suggestion. By declaring but not defining the global, it should resolve to the global in the main program executable I think. (if the global is not private)

I think I know what you mean and will experiment a bit. Though even if this works, we still need a solution for private, mutable globals.

ptersilie:
I think this is what you meant @bjorn3:
```
@switch.table.foo = external global [3 x i32]*

define internal void @__yk_compiled_trace_0(i32* %0) {
  %2 = icmp ult i32 2, 3
  %3 = sext i32 2 to i64
  %4 = getelementptr inbounds [3 x i32], [3 x i32]** @switch.table.foo, i64 0, i64 %3
  %5 = load i32, i32* %4, align 4
  store i32 %5, i32* %0, align 4, !tbaa !0
  ret void
}
```
But it doesn't work: `Symbol not found: switch.table.foo`. I wonder if that's because the compiled trace isn't linked against the main program executable.

ptersilie:
As discussed offline we will deal with mutable globals some other time, so 18827af makes sure we clone constants into the trace module so we can keep going, and errors if the global isn't a constant. It also adds an ignored test, that we can enable when we implement supoprt for mutable globals.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: 5d22c88 should address this. There was no changes needed, as we already make sure that any operand/instruction only gets cloned once into the `VMap`.

c_tests/tests/compile_globals.c
```diff
@@ -1,5 +1,6 @@
 // Compiler:
```
ltratt, line 1: I suspect this file should be called something like `mutable_global`?

c_tests/tests/compile_globals.c
```diff
@@ -1,5 +1,6 @@
...
 // Run-time:
+// ignore: Mutable global variables not supported.
```
ltratt, line 3: In general I reckon `ignore` tags should go at the top of the file so that people can't miss them.

c_tests/tests/compile_globals.c
ptersilie: Agreed.

ptersilie: Fixed in a19c0c8.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -332,14 +332,21 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
...
-                *JITMod, OldGV->getValueType(), OldGV->isConstant(),
-                OldGV->getLinkage(), (Constant *)nullptr, OldGV->getName(),
-                (GlobalVariable *)nullptr, OldGV->getThreadLocalMode(),
-                OldGV->getType()->getAddressSpace());
-            GV->copyAttributesFrom(&*OldGV);
-            VMap[&*OldGV] = GV;
-            cloned_globals.push_back(OldGV);
+            if (OldGV->isConstant()) {
+              // Global variable is a constant so just copy it into the trace.
+              GlobalVariable *GV = new GlobalVariable(
```
ltratt, line 340: It might not be a problem here, but in C (dunno about C++) the evaluation order of paramaters is undefined i.e. "OldGV->isConstant()` might be called before `OldGV->getValueType()`. My sort-of hack for this is that if there's more than one function call in the parameters (and I think there *might* be mutation in one of them), I tend to assign everything to temporary variables before the function call. Possibly more for reference here than anything else.

c_tests/tests/compile_globals.c
ptersilie: Fixed in 274a5b9.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: I'm not saying that LLVM is the poster child of code quality, but I've copied this almost verbatim from https://github.com/llvm/llvm-project/blob/d480f968ad8b56d3ee4a6b6df5532d485b0ad01e/llvm/lib/Transforms/Utils/CloneModule.cpp#L63, so I think this should be fine.

I also don't quite understand why it should make any difference if the parameters are evaluated in a different order in this case.

ltratt: If none of the functions does any mutation, it should be safe.

ltratt:
Please squash (and I guess commit messages need to be clear about what we have and haven't done in this PR).

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: As far as I can tell these are all just reads.

ptersilie:
Squashed.

PR #359, opened 2021-06-16
Fix returning a constant from a traced function.

vext01:
Prior to this change, we would miscompile the JIT module and the verifier would crash.

We were looking to the variable map for a mapped variable, which doesn't make sense for a constant.

[@ptersilie helped me understand what was going on here]

### DIFF
--- a/c_tests/tests/compile_constant_ret.c
+++ b/c_tests/tests/compile_constant_ret.c
@@ -1,0 +1,37 @@
+// Compiler:
+// Run-time:
+//   stderr:
+//     define internal void @__yk_compiled_trace_0(i32* %0) {
+//       ...
+//       store i32 30, i32* %0, align 4...
+//       ...
+
+// Check that returning a constant value from a traced function works.
+//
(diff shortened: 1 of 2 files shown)


ltratt:
Is it worth adding an IR test here? [@ptersilie might have an opinion on this.]

ptersilie:
While we should always get a crash in these cases, I think a IR test can't hurt. Just run the test with `YKDEBUG_PRINT_IR=1` and then copy output to the top. See https://github.com/ykjit/yk/blob/master/c_tests/tests/compile_simple.c for an example. If one of the opt-levels produces different code, create a new file with just the lang_tester stuff at the top, e.g. https://github.com/ykjit/yk/blob/master/c_tests/tests/compile_simple.c.O0

vext01:
20ac6e0 checks that the IR stores the constant `30` into the first trace argument. I believe that's correct.

ltratt:
Please squash.

vext01:
Splat.

PR #361, opened 2021-06-17
Add the ability to dump an annotated trace to stderr.

vext01:
This is a debugging aid which makes it easier to:
 - see the inlining thresholds in our JIT module.
 - see the variable mapping between the JIT and AOT module.

This feature is available only in debug builds and the dump is enabled by
setting `YKDEBUG_PRINT_IR_ANNOTATED=1` in the environment.

Here's a sample dump:

```
 -- Begin annotated trace dump for __yk_compiled_trace_0 ---
 JIT Module                                 | AOT Module
 # main()
 %3 = load i32, i32* %1, align 4, !tbaa !0  |  %8 = load i32, i32* %3, align 4, !tbaa !5
 # f()
 %4 = icmp eq i32 %3, 1                     |  %2 = icmp eq i32 %0, 1
 %5 = select i1 %4, i32 47, i32 52          |  %3 = select i1 %2, i32 47, i32 52
 %6 = icmp eq i32 %3, 0                     |  %4 = icmp eq i32 %0, 0
 %7 = select i1 %6, i32 30, i32 %5          |  %5 = select i1 %4, i32 30, i32 %3
 # main()
 store i32 %7, i32* %0, align 4, !tbaa !0   |  store i32 %9, i32* %4, align 4, !tbaa !5
 ret void
 --- End annotated trace dump for __yk_compiled_trace_0 ---
```

This could be improved by showing the inlined call sites (and their arguments)
in the AOT column, but one step at a time.

Co-authored-by: Lukas Diekmann <lukas.diekmann@gmail.com>

### DIFF
--- a/ykllvmwrap/src/ykllvmwrap.cc
+++ b/ykllvmwrap/src/ykllvmwrap.cc
@@ -174,6 +174,79 @@ extern "C" void *compile_module(string TraceName, Module *M) {
   return (void *)EE->getFunctionAddress(TraceName);
 }
 
+#ifndef NDEBUG
+// Left trim (in-place) the character `C` from the string `S`.
+void lTrim(string &S, const char C) {
+  S.erase(0, std::min(S.find_first_not_of(C), S.size() - 1));
+}
+
+// Dumps an LLVM Value to a string and trims leading whitespace.
+void dumpValueToString(Value *V, string &S) {
+  raw_string_ostream RSO(S);
+  V->print(RSO);
+  lTrim(S, ' ');
+}
+
+// Print a trace's instructions "side-by-side" with the instructions from
+// which they were derived in the AOT module.
+void printSBS(Module *AOTMod, Module *JITMod, ValueToValueMapTy &RevVMap) {
+  assert(JITMod->size() == 1);
+  Function *JITFunc = &*JITMod->begin();
+
+  // Find the longest instruction from the JITMod so that we can align the
+  // second column.
+  size_t LongestJITLine = 0;
+  for (auto &JITBlock : *JITFunc) {
+    for (auto &JITInst : JITBlock) {
+      string Line;
+      dumpValueToString(&JITInst, Line);
+      auto Len = Line.length();
+      if (Len > LongestJITLine)
+        LongestJITLine = Len;
+    }
+  }
+
+  const string JITHeader = string("Trace");
+  string Padding = string(LongestJITLine - JITHeader.length(), ' ');
+  errs() << "\n\n--- Begin trace dump for " << JITFunc->getName() << " ---\n";
+  errs() << JITHeader << Padding << "  | AOT\n";
+
+  // Keep track of the AOT function we are currently in so that we can print
+  // inlined function thresholds in the dumped trace.
+  StringRef LastAOTFunc;
+  for (auto &JITBlock : *JITFunc) {
+    for (auto &JITInst : JITBlock) {
+      auto V = RevVMap[&JITInst];
+      if (V == nullptr) {
+        // The instruction wasn't cloned from the AOTMod, so print it only in
+        // the JIT column and carry on.
+        std::string Line;
+        dumpValueToString((Value *)&JITInst, Line);
+        errs() << Line << "\n";
+        continue;
+      }
+      Instruction *AOTInst = (Instruction *)&*V;
+      assert(AOTInst != nullptr);
+      Function *AOTFunc = AOTInst->getFunction();
+      assert(AOTFunc != nullptr);
+      StringRef AOTFuncName = AOTFunc->getName();
+      if (AOTFuncName != LastAOTFunc) {
+        // Print an inlining threshold.
+        errs() << "# " << AOTFuncName << "()\n";
+        LastAOTFunc = AOTFuncName;
+      }
+      string JITStr;
+      dumpValueToString((Value *)&JITInst, JITStr);
+      string Padding = string(LongestJITLine - JITStr.length(), ' ');
+      string AOTStr;
+      dumpValueToString((Value *)AOTInst, AOTStr);
+      errs() << JITStr << Padding << "  |  " << AOTStr << "\n";
+    }
+  }
+  errs() << "--- End trace dump for " << JITFunc->getName() << " ---\n";
+}
+#endif
+
 // Compile an IRTrace to executable code in memory.
 //
 // The trace to compile is passed in as two arrays of length Len. Then each
@@ -215,6 +288,9 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
   Builder.SetInsertPoint(DstBB);
 
   llvm::ValueToValueMapTy VMap;
+#ifndef NDEBUG
+  llvm::ValueToValueMapTy RevVMap;
+#endif
   // Variables that are used (but not defined) inbetween start and stop tracing
   // need to be replaced with function arguments which the user passes into the
   // compiled trace. This loop creates a mapping from those original variables
@@ -382,6 +458,9 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
       auto NewInst = &*I->clone();
       llvm::RemapInstruction(NewInst, VMap, RF_NoModuleLevelChanges);
       VMap[&*I] = NewInst;
+#ifndef NDEBUG
+      RevVMap[NewInst] = &*I;
+#endif
       Builder.Insert(NewInst);
     }
   }
@@ -399,9 +478,14 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
   }
 
 #ifndef NDEBUG
+  char *SBS = getenv("YK_PRINT_IR_SBS");
+  if ((SBS != nullptr) && (strcmp(SBS, "1") == 0)) {
+    printSBS(AOTMod, JITMod, RevVMap);
+  }
   llvm::verifyModule(*JITMod, &llvm::errs());
 #endif
-  auto PrintIR = std::getenv("YKDEBUG_PRINT_IR");
+
+  auto PrintIR = std::getenv("YK_PRINT_IR");
   if (PrintIR != nullptr) {
     if (strcmp(PrintIR, "1") == 0) {
       // Print out the compiled trace's IR to stderr.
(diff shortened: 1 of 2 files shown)


ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -174,6 +174,91 @@ extern "C" void *compile_module(string TraceName, Module *M) {
...
+      dumpValueToString(&JITInst, Line);
+      auto Len = Line.length();
+      if (Len > LongestJITLine)
+        LongestJITLine = Len;
+    }
+  }
+
+  const string JITHeader = string("JIT Module");
+  string Padding = string(LongestJITLine - JITHeader.length(), ' ');
+  errs() << "\n\n--- Begin annotated trace dump for " << JITFunc->getName()
```
ltratt, line 221: "annotated trace dump" feels a mouthful. Just "annotated trace" or even just "trace" maybe?

ltratt:
Dumb question: when we inline across AOT modules, what happens to the RHS of the printout? And I now realise that "annotated trace dump" might perhaps be better something like "side-by-side trace"? I guess not many people will want this side-by-side style except us for internal debugging purposes?

ykllvmwrap/src/ykllvmwrap.cc
vext01: "annotated" is to distinguish this from the other kind of trace dump that we already have (for use by lang_tester IR tests).

"annotated trace" works for me.

vext01:
> what happens to the RHS of the printout?

The instructions in the AOT module currently live as long as the compilation thread, but we have a FIXME to return the cloned module to a pool for reuse.

I'm not sure if that answers your question?

vext01:
> I guess not many people will want this side-by-side style except us for internal debugging purposes?

Correct. All of this is guarded by `NDEBUG`.

ltratt:
> The instructions in the AOT module currently live as long as the compilation thread, but we have a FIXME to return the cloned module to a pool for reuse.

That's not what I mean :) Imagine I have a function `f` that calls a function `g` and I trace `f`: what does the RHS show? A mix of `f` and `g`? How do I know when I've flipped from one to another?

ykllvmwrap/src/ykllvmwrap.cc
ltratt: I think "side-by-side" might be better ("annotated" is a very generic term that doesn't really say much)?

ptersilie:
We don't trace across AOT modules. There's only one (are they ever going to be more?).

> Imagine I have a function f that calls a function g and I trace f: what does the RHS show?
It just shows the instructions from f and g in the order they were cloned over to the traced module. In a way, Edds example above already has such a case with `main` and `f`.

ltratt:
I'm confused: AOT means "Ahead Of Time" right? So how can we have inlined those things ahead of time? Is the RHS of this debugging output maybe not named correctly? or ... ?

ptersilie:
To be fair, I think this side-by-side thing is only really useful for small traces, i.e. our tests. In an actual program, the AOT module is going to be so huge, that it will be almost impossible to manually lookup the RHS in the AOT module to see what the trace did. 

Maybe this needs to become interactive at some point, where you can just select an instruction in the JIT module, and then it jumps to the corresponding instruction in the AOT module.

ltratt:
That still doesn't answer my "what does AOT mean here" question though :) I'm genuinely curious / confused about the terminology!

ptersilie:
> I'm confused: AOT means "Ahead Of Time" right? So how can we have inlined those things ahead of time? Is the RHS of this debugging output maybe not named correctly? or ... ?

The RHS is the original instruction in the AOT module. The LHS is the cloned instruction in the JIT module. This is just an aid to see where the instructions inside a trace originally came from.

ptersilie:
> That still doesn't answer my "what does AOT mean here" question though :) I'm genuinely curious / confused about the terminology!

I'm not so fast! :-) That comment was still replying to your comments further above.

ltratt:
Are you saying, I think, that we're likely to rip this out at some point in the not-too distant future because a) it doesn't scale b) it catches a small class of bugs that (hopefully) we soon won't encounter any more?

vext01:
> Are you saying

Neither a nor b. I want this to be a useful debugging aid.

> what does AOT mean here

To build a trace we copy instructions out of the blocks of the AOT module (which is a full CFG) into the JIT module (which is a trace). In doing so new variables have to be instantiated, as you can't share variables across modules. But this means that each time we copy an instruction into the trace, we can't "simply copy it" (Tm) as the operands must reference the new variables in the JIT module, not the old variables from the AOT module.

So the AOT column shows the instructions from the AOT module, and the return values and operands of those statements are variables from the AOT module. The JIT coumn shows the same instructions, but with their newly instantiated variables (inevitably the variable numbers differ, as tracing typically only observes a subset of all variables).

Seeing the two columns in this way is useful when debugging, as trying to keep the mapping in your head is really difficult. I certainly struggle.

> Imagine I have a function f that calls a function g and I trace f: what does the RHS show? 

You will see mostly the same instructions on the right hand side, but with different variable numbers to those on the left hand side.

[Note also that a %1 on the left-hand side is distinct from a %1 on the right. The variable numbers are only unique to a module].

[The `#` lines are semi-related. They show the point at which the function from which the instructions were cloned has changed. This is useful for locating problematic portions of the trace]

Does it make sense? I think it'd make more sense once we include the calls in the right-hand column. At present we don't do that, but I would like to.

ltratt:
> I think it'd make more sense once we include the calls in the right-hand column.

Yes, that might be the point at which I start to see that this is potentially useful long-term.

I think part of my problem is that the LHS is (I think accurately) called "JIT Module" but the RHS is (I think inaccurately) called "AOT Module". I *think* the RHS is more like "Snippets from AOT Modules"?

ptersilie:
> I _think_ the RHS is more like "Snippets from AOT Modules"?

Yes, that's a more accurate description.

> I think it'd make more sense once we include the calls in the right-hand column.

We can tell already where the calls are, even though we don't show the actual call instruction. In a way the RHS will always be snippets, since we can't show the entire AOT module as that would create too much bloat. However, it would be good if we could see which instructions were turned into guards in the future.

ltratt:
OK let's call the left-hand side "Trace" or "JIT" and the RHS "AOT" and drop "module" entirely then?

vext01:
> OK let's call the left-hand side "Trace" or "JIT" and the RHS "AOT" and drop "module" entirely then?

That's fine.

ykllvmwrap/src/ykllvmwrap.cc
vext01: I don't like side-by-side.

Annotated feels right here. We are marking (annotating) the inlining boundaries.

@ptersilie any thoughts?

ltratt: Won't we have other types of annotated trace things? Basically I'm saying that "annotated" and "dump" are not adding value here (but they are adding a (rather small, I admit) cognitive load).

vext01: I'd still like to hear what Lukas has to say, but note that "dump" is a convention used in LLVM for printing debugging info the stderr. You can invoke `->dump()` on pretty much anything and get a textual version to eyeball.

ptersilie: I'm not fussed about the name. Just call it `Begin trace dump`, then we don't need to worry about whether `annotated` or `side-by-side` is better.

vext01: The problem is, we already have the other trace dump used by tests, so we need to differentiate them with some term.

ptersilie: The other dump doesn't print any headers though. I don't think it's an issue.

ptersilie:
I have two more comments:
1) Can we dump this entire thing into a separate file, e.g. `debug.cc` that we then include. I don't want to unnecessarily bloat `ykllvmwrap.cc` 
2) This currently doesn't work if you have the same function follow each other (and I assume the same is true for recursion), since this check (https://github.com/ykjit/yk/pull/361/files#diff-28de043b71aaedbd9af7218078abd71f6216cff0ce9dbb53e6518efc62817021R244) can't tell the difference between two instructions from the same function but different calls.

ltratt:
>Can we dump this entire thing into a separate file, e.g. debug.cc that we then include. I don't want to unnecessarily bloat ykllvmwrap.cc

Good idea!

ykllvmwrap/src/ykllvmwrap.cc
vext01: What env var should control this? Currently you turn it on with `YKDEBUG_PRINT_IR_ANNOTATED=1`.

The tests set `YKDEBUG_PRINT_IR=1`.

ltratt: Personally I think everything should be `YK_` and that `DEBUG_PRINT` just says the same thing twice. So I think something like `YK_PRINT_IR` and `YK_PRINT_SBS` for side-by-side might be right. I know I'm repeating myself, but I'm convinced we'll have other "annotated" print thingies, so "annotated" just doesn't feel like a useful discriminator.

vext01:
I've now pushed fixes for all outstanding comments.

Lukas found a bug which i'll now fix.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -484,14 +481,14 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
   }
 
 #ifndef NDEBUG
-  char *DumpAnn = getenv("YKDEBUG_PRINT_IR_ANNOTATED");
-  if ((DumpAnn != nullptr) && (strcmp(DumpAnn, "1") == 0)) {
-    dumpAnnotatedTrace(AOTMod, JITMod, &VMap);
+  char *SBS = getenv("YK_PRINT_IR_SBS");
```
ltratt, line 484: Ah, yes, this is a better name than my suggestion!

vext01:
@ptersilie try 8316a42

ptersilie:
Looks good. Though I'm not sure how this messes with our plan to put this debug stuff into its own file.

vext01:
That's one for later PR.

OK to squash?

ltratt:
Please squash.

vext01:
splat.

ltratt:
Minor typo of the year in the commit title: should be "A side-by-side" (not "AN"). Please feel free to force push an update.

vext01:
fixed.

PR #362, opened 2021-06-18
Update a stale comment.

vext01:


ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -394,14 +394,17 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
       }
 
       // If execution reaches here, then the instruction I is to be copied into
-      // JITMod. We now scan the instruction's operands checking that each is
+      // JITMod.
+      //
+      // FIXME: We now scan the instruction's operands checking that each is
```
ltratt, line 399: "each is corresponding variable" doesn't parse.

vext01: There's an unchanged line in-between, so the line reads:

> FIXME: We now scan the instruction's operands checking that each is defined in JITMod.

ltratt: Mea culpa!

PR #363, opened 2021-06-18
Use the new version of lang_tester to set env vars on a per-test basis.

ltratt:
This means we no longer have to set environment variables for *every* test, as before.


vext01:
I like it.

bors r+

PR #364, opened 2021-06-21
Simplifications

vext01:
Whilst working on something else, I've gained new insights which I think allow us to simplify some of the operand scanning code.

Each commit is free-standing. See individual commit messages.

All tests passing.


ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -437,20 +437,17 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
...
-              // default value.
-              // FIXME fails for meta-data types (when you build with -g).
-              Value *NullVal = Constant::getNullValue(OpTy);
-              VMap[Op] = NullVal;
-            }
+          } else if (isa<Constant>(Op)) {
+            // The operand is a constant, so leave it as is.
+            VMap[Op] = Op;
+            continue;
+          } else if (Op == ThreadTracer) {
```
ltratt, line 444: I *think* `ThreadTrace` can be `null` at this point. Does that cause issues?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -394,24 +398,20 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
...
+          // The operand is undefined in JITMod.
           Type *OpTy = Op->getType();
           if (isa<llvm::AllocaInst>(Op)) {
-            // Value is a stack allocation, so make a dummy stack slot.
+            // The operand is alocated on the stack with an `alloca`.
+            //
+            // LLVM has a tendency to move allocas up to the first block of a
+            // function, and if we didn't trace that block, then we will have
+            // missed those allocations. In these cases we materialise the
+            // allocations as we see them used in code that *was* traced.
```
ltratt, line 414: Does this mean that we duplicate some allocas if they were included in the trace?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -403,10 +403,10 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
       for (unsigned OpIdx = 0; OpIdx < I->getNumOperands(); OpIdx++) {
         Value *Op = I->getOperand(OpIdx);
         if (VMap[Op] == nullptr) {
-          // Value is undefined in JITMod.
+          // The operand is undefined in JITMod.
           Type *OpTy = Op->getType();
           if (isa<llvm::AllocaInst>(Op)) {
-            // Value is a stack allocation.
+            // The operand is alocated on the stack with an `alloca`.
```
ptersilie, line 409: s/alocated/allocated

ptersilie:
Looks good to me.

ykllvmwrap/src/ykllvmwrap.cc
vext01: It shouldn't be, as we only flip `Tracing` on when we see the instruction that generates `ThreadTracer`.

ykllvmwrap/src/ykllvmwrap.cc
vext01: No. Each variable has exactly one alloca.

ykllvmwrap/src/ykllvmwrap.cc
ltratt: Do you mean `ThreadTracer` can't be null? If so, maybe add a (debug) assert after the `for` loop to check this?

ykllvmwrap/src/ykllvmwrap.cc
ltratt: How do we guarantee that?

ykllvmwrap/src/ykllvmwrap.cc
vext01: Fixed in 24be293.

ykllvmwrap/src/ykllvmwrap.cc
vext01: The LLVM verifier would have rejected the module at AOT compile-time.

vext01: And in fact, since each alloca is a pointer address, they can't be duplicated.

ltratt: Let me change my question: do we explicitly check "ah, this var already has an alloca in the trace so we don't need to add another"?

ykllvmwrap/src/ykllvmwrap.cc
vext01: ThreadTracer cannot be null if `Tracing` is set.

When we see the call that turns on tracing:
```                                                                                                                                                                                                      
        if (CF->getName() == YKTRACE_START) {                                                                                                                                                                      
          ThreadTracer = &*I;                                                                                                                                                                                      
          Tracing = true;        
```

It occurs to me that we don't need the `Tracing` flag, as we can just check the nullness of this pointer. Would that be preferential / clearer?

ykllvmwrap/src/ykllvmwrap.cc
vext01: Indirectly, yes. We only reach this code if the variable is [as-yet undefined](https://github.com/ykjit/yk/blob/4d6276ced71842c0495379e3763c5cdf17161b4e/ykllvmwrap/src/ykllvmwrap.cc#L405-L416).

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: I think we should keep `Tracing` as it's much clearer what it's for than comparing to some pointer.

ykllvmwrap/src/ykllvmwrap.cc
ltratt: OK, maybe we should add that to the comment (or, even better, add an assert along those lines, if that's possible)?

ykllvmwrap/src/ykllvmwrap.cc
ltratt: Having two variables implicitly track the same thing seems like it's more confusing, at least to me.

vext01: @ptersilie How about if we renamed `ThreadTracer` to `StartTracingInstr`? That way it makes some sense in both contexts. We are tracing if we've seen the `StartTracingInstr` and the value produced is the `ThreadTracer`.

ykllvmwrap/src/ykllvmwrap.cc
vext01: Well, we can, but the line where we do the check is only a few lines above. Is it really necessary?

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: I'm happy with that.

vext01: Have a look at 3708c9f2da0. I'm on the fence as to whether it's better or not.

ltratt: This reads better to me.

vext01:
I think the last remaining discussion is: https://github.com/ykjit/yk/pull/364#discussion_r655381220

ykllvmwrap/src/ykllvmwrap.cc
vext01: [this one](https://github.com/ykjit/yk/blob/4d6276ced71842c0495379e3763c5cdf17161b4e/ykllvmwrap/src/ykllvmwrap.cc#L405)

vext01: @ltratt ping

ltratt: OK, I see. I think a more clear comment that this is what happening at line 405ish would really help (at least it would help me).

vext01: The comment at line 406 is about as good as it gets:

> // The operand is undefined in JITMod.

But maybe it helps to reiterate this a few lines later? Try b59cf77008?

ltratt:
Please squash.

vext01:
splat.

PR #365, opened 2021-06-22
Allow us to trace binaries compiled with -g.

vext01:
The first commit is the main body of work. The second is tidying.

### DIFF
--- a/c_tests/run.rs
+++ b/c_tests/run.rs
@@ -29,6 +29,8 @@ fn mk_compiler(exe: &Path, src: &Path, opt: &str) -> Command {
 
     compiler.args(&[
         opt,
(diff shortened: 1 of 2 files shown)


ptersilie:
Just a typo in in the second commit message: s/it's/its. Did you also run `cargo xtask cfmt` already. I saw something that I'm not sure would be caught by the formatter. You can squash any formatting right away.

vext01:
`cargo xtask cfmt` doesn't change anything. Which line is suspect?

ptersilie:
Nevermind then. :-D Just the typo to fix then.

vext01:
okie dokie. Commit message fixed.

PR #366, opened 2021-06-22
Dealing with recursive traces.

ptersilie:
This is a work-in-progress PR attempting to deal with traces of recursive functions. The problem with recursion in traces lies in the way we copy over instructions from the AOT module to the JIT module. In short: Inlining the same function twice overwrites the mapping in the ValueToValueMap, which is used to update references in cloned instructions with their cloned counterparts. Here's an example:

```
f() AOTMod                    JITMod                VMAP
  %1 = i32 1                  %1 = i32 1            %1 => %1
  %2 = add 1, %1              %2 = add 1, %1        %2 => %2
  # call f()                      
      %1 = i32 1              %3 = i32, 1           %1 => %3
      %2 = add 1, %1          %4 = add 1, %3        %2 => %4
      # call f() skipped        
      %3 = add %2, %2         %5 = add %4, %4       %3 => %5
  %3 = add %2, %2             %6 = add %4, %4       %3 => %6
```

On the LHS are the traced instructions from the AOT module (minus calls/returns). In the middle is the JIT module after we've cloned over each traced instruction into the JIT module. On the RHS is shown how we update the VMap after each instruction. Note, how when we clone the same instruction a second time (e.g. %2 = add 1, %1), we overwrite its VMap entry, which previously mapped `%2` to `%2`, but now maps to `%4`. This means, that when update the references in instruction `%6`, we map to the wrong values, i.e. instead of `%6 = add %4, %4`, it should read `%6 = add %2, %2`.

This PR solves this by having a separate VMap for each function boundary that we traced, i.e. each inlined function gets its own VMap during construction of the JIT module. This way instructions can't clash and won't be overwritten. However, there's a few things I dislike about this approach:
1) Due to VMaps being non-movable we need to jump through some hoops, to switch out the active VMap when reaching a new function boundary or returning from it, see the `(*VMap)[V] = V` stuff. This might just be my lacking knowledge of C++ though.
2) We need to copy over all globals, each time we create a new VMap for another function.

I had an idea for another approach, which involves only using one VMap, but remembering each time another function overwrites an entry, and then after returning from that function, restores those overwritten values back to their previous version. But this solution isn't much better, since we'd need to keep a stack of hashmaps around (one per nested function).

I'd be grateful for any ideas/comments.

### DIFF
--- a/ykllvmwrap/src/ykllvmwrap.cc
+++ b/ykllvmwrap/src/ykllvmwrap.cc
@@ -287,7 +287,8 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
   llvm::IRBuilder<> Builder(JITContext);
   Builder.SetInsertPoint(DstBB);
 
-  llvm::ValueToValueMapTy VMap;
+  std::vector<ValueToValueMapTy*> vmaps;
+  vmaps.push_back(new llvm::ValueToValueMapTy());
 #ifndef NDEBUG
   llvm::ValueToValueMapTy RevVMap;
 #endif
@@ -295,11 +296,12 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
   // need to be replaced with function arguments which the user passes into the
   // compiled trace. This loop creates a mapping from those original variables
   // to the function arguments of the compiled trace function.
+  auto VMap = vmaps.back();
   for (size_t Idx = 0; Idx != Inputs.size(); Idx++) {
     Value *OldVal = Inputs[Idx];
     Value *NewVal = DstFunc->getArg(Idx);
     assert(NewVal->getType()->isPointerTy());
-    VMap[OldVal] = NewVal;
+    (*VMap)[OldVal] = NewVal;
   }
 
   bool Tracing = false;
@@ -353,6 +355,9 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
           // FIXME Deal with calls we cannot or don't want to inline.
           if (Tracing) {
             inlined_calls.push_back(CI);
+            auto PrevVMap = vmaps.back();
+            vmaps.push_back(new llvm::ValueToValueMapTy());
+            VMap = vmaps.back();
             // During inlining, remap function arguments to the variables
             // passed in by the caller.
             for (unsigned int i = 0; i < CI->arg_size(); i++) {
@@ -360,9 +365,9 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
               Value *Arg = CF->getArg(i);
               // If the operand has already been cloned into JITMod then we need
               // to use the cloned value in the VMap.
-              if (VMap[Var] != nullptr)
-                Var = VMap[Var];
-              VMap[Arg] = Var;
+              if ((*PrevVMap)[Var] != nullptr)
+                Var = (*PrevVMap)[Var];
+              (*VMap)[Arg] = Var;
             }
             break;
           }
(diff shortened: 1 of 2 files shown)


ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -287,19 +287,21 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
   llvm::IRBuilder<> Builder(JITContext);
   Builder.SetInsertPoint(DstBB);
 
-  llvm::ValueToValueMapTy VMap;
+  std::vector<ValueToValueMapTy*> vmaps;
+  vmaps.push_back(new llvm::ValueToValueMapTy());
```
vext01, line 291: I think you said that we were unable to allocate the VMaps directly into the vector. Is that still the case?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -353,16 +355,24 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
           // FIXME Deal with calls we cannot or don't want to inline.
           if (Tracing) {
             inlined_calls.push_back(CI);
+            auto PrevVMap = vmaps.back();
+            vmaps.push_back(new llvm::ValueToValueMapTy());
+            VMap = vmaps.back();
+            // Copy over globals
+            auto OutermostVMap = vmaps.front();
```
vext01, line 362: Does this assume that all globals are pulled in my the first frame?

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Ah, yes, I forgot to mention this as number 3 why I don't like this approach. We need to allocate the VMaps on the heap. Though, again, that could just be me being bad at C++. ;-) But I couldn't find another way.

ptersilie: It did not.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Define "pulled in"? I'm adding all globals to the first frame, because we need to be able to find them once the trace is done in order to initialise them.

vext01: So that we don't have to iterate all VMaps collecting a set? Ah, I see. I think a comment is required here.

ltratt:
I think we need to define our terms really carefully here. Does "recursive trace" mean "recursive PT (or software) trace in the sense that a function `f` in the trace calls itself (possibly indirectly) again during the trace"? If so, the PT trace is *unrolling* the recursion. I suggest that we want to not do that in the JIT trace we compile: we should convert the first "recursive call" to a real call to `f` rather than unroll things.

vext01:
We can do that in the short term, but as you know from your experience with RPython, unrolling recursion to a bounded depth can be beneficial. We will need to be able to deal with inlining recursive calls at some point.

Whether that's now or not, I don't know.

ptersilie:
> I think we need to define our terms really carefully here. Does "recursive trace" mean "recursive PT (or software) trace in the sense that a function `f` in the trace calls itself (possibly indirectly) again during the trace"? If so, the PT trace is _unrolling_ the recursion. I suggest that we want to not do that in the JIT trace we compile: we should convert the first "recursive call" to a real call to `f` rather than unroll things.

Is this only for direct recursion or also for something like `f` calling `g`, and `g` calling `f`. At what point do we stop inlining?

ltratt:
>  but as you know from your experience with RPython, unrolling recursion to a bounded depth can be beneficial.

On the contrary, I strongly suspect that it's a bad idea in general to unroll recursive function calls. Imagine I have something like:

```rust
fn f(x: u64) {
  if x = 0 {
    return;
  }
  print(...);
  f(x - 1);
}
```

And let's imagine when we run PT `x` starts at 5 so we get several unrollings. I would like the compiled trace to look (roughly) as follows:

```
guard(x > 0);
print(...);
f(x - 1);
```

i.e. the "first" instance of `f` appears in the trace but the first recursive call of `f` turns into an actual call.

> Is this only for direct recursion or also for something like f calling g, and g calling f. At what point do we stop inlining?

We need to catch direct and indirect recursion.

vext01:
So I think we are going to close this draft and start by just emitting calls if we detect recursion?

ptersilie:
Yes, we can close it.

PR #367, opened 2021-06-24
Canonicalise object paths.

vext01:
See individual commits.

### DIFF
--- a/c_tests/tests/rel_path.c
+++ b/c_tests/tests/rel_path.c
@@ -1,0 +1,58 @@
+// Compiler:
+// Run-time:
+//   env-var: YK_PRINT_IR=1
+//   stderr:
+//   stderr:
+//     define internal void @__yk_compiled_trace_0(i32* %0) {
+//        store i32 2, i32* %0, align 4, !tbaa !0
+//        ret void
+//     }
+//     ...
+
+// Check that running a traced binary via a relative path works.
+//
+// FIXME An optimising compiler can remove all of the code between start/stop
+// tracing.
+
+#include <assert.h>
+#include <err.h>
+#include <libgen.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <yk_testing.h>
+
+int main(int argc, char **argv) {
+  if (*argv[0] == '/') {
+    // Reinvoke ourself with a relative path.
+    char *base = basename(argv[0]);
+    if (base == NULL)
+      err(EXIT_FAILURE, "basename");
+
+    char *dir = dirname(argv[0]);
+    if (dir == NULL)
+      err(EXIT_FAILURE, "dirname");
+    if (chdir(dir) != 0)
+      err(EXIT_FAILURE, "chdir");
+
+    if (execl(base, base, NULL) == -1)
+      err(EXIT_FAILURE, "execl");
+    // NOREACH
+  }
+
+  int res = 0;
+  void *tt = __yktrace_start_tracing(HW_TRACING, &res);
+  res = 2;
+  void *tr = __yktrace_stop_tracing(tt);
+  assert(res == 2);
+
+  void *ptr = __yktrace_irtrace_compile(tr);
+  __yktrace_drop_irtrace(tr);
+  void (*func)(int *) = (void (*)(int *))ptr;
+  int res2 = 0;
+  func(&res2);
+  assert(res2 == 2);
+
+  return (EXIT_SUCCESS);
+}
--- a/ykutil/src/addr.rs
+++ b/ykutil/src/addr.rs
@@ -43,3 +43,26 @@ pub fn code_vaddr_to_off(vaddr: usize) -> Option<(&'static CStr, u64)> {
     }
     None
 }
+
+#[cfg(test)]
+mod tests {
+    use super::code_vaddr_to_off;
+    use libc::dlsym;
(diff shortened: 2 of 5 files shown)


ykutil/src/addr.rs
```diff
@@ -43,3 +43,25 @@ pub fn code_vaddr_to_off(vaddr: usize) -> Option<(&'static CStr, u64)> {
...
+
+#[cfg(test)]
+mod tests {
+    use super::code_vaddr_to_off;
+    use libc::dlsym;
+    use std::ptr;
+
+    #[test]
+    fn map_libc() {
+        let vaddr = unsafe { dlsym(ptr::null_mut(), "getuid".as_ptr() as *const i8) };
```
ltratt, line 71: Is `"getuid".as_ptr()` equivalent to C string? I think we should use `CString` to ensure a null terminator?

vext01: You are right. This is incorrect.

c_tests/tests/rel_path.c
```diff
@@ -0,0 +1,58 @@
...
+    if (base == NULL)
+      err(EXIT_FAILURE, "basename");
+
+    char *dir = dirname(argv[0]);
+    if (dir == NULL)
+      err(EXIT_FAILURE, "dirname");
+    if (chdir(dir) != 0)
+      err(EXIT_FAILURE, "chdir");
+
+    if (execl(base, base, NULL) != 0)
```
ltratt, line 39: I think that this should be `if (execl(...) == -1` to be clear about the range of return codes?

c_tests/tests/rel_path.c
```diff
@@ -0,0 +1,58 @@
...
+#include <libgen.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <yk_testing.h>
+
+int main(int argc, char **argv) {
+  if (*argv[0] == '/') {
+    // Reinvoke ourself with a relative path.
```
ltratt, line 28: This is wonderfully evil :p I don't know if I'd have thought of this!

vext01: It took a while to invent :P

ykutil/src/addr.rs
```diff
@@ -1,21 +1,35 @@
...
     ptr::{null, null_mut},
 };
 
-static CUR_EXE_C: Lazy<CString> =
-    Lazy::new(|| CString::new(env::current_exe().unwrap().to_str().unwrap()).unwrap());
+// Make the path `p` a canonical absolute path.
+//
+// PERF: Cache the results to avoid file-system usage.
+//
+// PERF: A lot of the canonicalisation work we do is for the object names in the program headers.
```
ltratt, line 18: I think this is the right `PERF` and we should delete the "cache" `PERF` above.

ykutil/src/addr.rs
```diff
@@ -26,16 +40,18 @@ pub fn code_vaddr_to_off(vaddr: usize) -> Option<(&'static CStr, u64)> {
...
     for obj in &objects() {
-        let mut obj_name = obj.name();
-        let obj_name_p = obj_name.as_ptr();
-        if unsafe { *obj_name_p } == 0 {
+        let obj_name = obj.name();
+        let obj_name: PathBuf = if unsafe { *obj_name.as_ptr() } == 0 {
             // On some systems, the empty string indicates the main binary.
-            obj_name = CUR_EXE_C.as_c_str();
-        }
+            // Note that the result of current_exe() is already canonical.
```
ltratt, line 50: Instead of this comment how about something like `debug_assert_eq!(env::current_exe(), canonicalise_path(env::curent_exe());`?

ykutil/src/addr.rs
vext01: The second `PERF` doesn't cover all canonicalizations, hence both comments.

(The other case is [here](https://github.com/ykjit/yk/pull/367/files/03b373c8bdb38bcbcf1af170a213bfadb42028b2..d7dcbe3f667db477b1291f74d61be7745ce2aa6f#diff-aa389a21efb030fbbaf6bba425c697d7076980ec84b58822881b2acc78d3778cR43))

ltratt: I'm actually pretty nervous about this function because caching can give different results than non-caching (if someone moves files around). I worry that this function will bite us on the bottom if we ever make it cache. Can we fix the second case another way and/or maybe it's not actually a bottleneck?

vext01: I don't want to get too bogged down in optimisation now, but I think {re,}moving binaries and libraries will likely be fatal for our system anyway. For example the PT decoder relies on on-disk objects to construct a trace, and we need to query objects for their symbols etc.

For now shall we can weaken our `PERF` comment to something like `PERF: Repeated canonicalisation is likely to be slow. Is caching beneficial?`. Then our future selves can have fun profiling.

ltratt: I'd rather not mention caching at all and I'd rather not implement caching in the future unless this turns out to be absolutely killer to performance.

vext01: Fair enough then. Removed.

vext01:
I think that's all comments addressed in 6b68914.

ptersilie:
In @ltratt's absence I will taker over.

Please squash.

vext01:
splat.

PR #369, opened 2021-06-29
Add the ability to express holes in a trace.

vext01:
Currently we stop mapping a trace at the first (non-prefix) unmappable
block. This is incorrect in the long run, as the trace may not be
complete. For example, it could be that there's a call to external code
for which we have no IR.

This change allows us to express holes in a trace, which we then leave
up to the trace compiler to interpret as it sees fit.

This will be refined with follow-up PRs. Namely:
 - Use a more compact representation for holes in traces.
 - Strip "holes" from the beginning and end of the trace, as they are
   never helpful.

### DIFF
--- a/ykllvmwrap/src/ykllvmwrap.cc
+++ b/ykllvmwrap/src/ykllvmwrap.cc
@@ -315,6 +315,9 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
   for (size_t Idx = 0; Idx < Len; Idx++) {
     auto FuncName = FuncNames[Idx];
 
+    // FIXME: Deal with holes in the trace, e.g. calls to libc.
+    assert(FuncName != nullptr);
+
     // Get a traced function so we can extract blocks from it.
     Function *F = AOTMod->getFunction(FuncName);
     if (!F)
--- a/yktrace/src/hwt/mapper.rs
+++ b/yktrace/src/hwt/mapper.rs
@@ -100,24 +100,32 @@ impl HWTMapper {
         }
     }
 
-    /// Maps each entry of a hardware trace back the IR block from whence it was compiled.
-    pub(super) fn map_trace(&self, trace: Box<dyn Trace>) -> Result<Vec<IRBlock>, HWTracerError> {
+    /// Maps each entry of a hardware trace back to the IR block from which it was compiled.
+    pub(super) fn map_trace(
+        &self,
+        trace: Box<dyn Trace>,
+    ) -> Result<Vec<Option<IRBlock>>, HWTracerError> {
         let mut ret_irblocks = Vec::new();
         let mut itr = trace.iter_blocks();
         while let Some(block) = itr.next() {
             let block = block?;
             let irblocks = self.map_block(&block);
-            if !ret_irblocks.is_empty() && irblocks.is_empty() {
-                // Once we have seen the last block that can be mapped we are done.
-                break;
+            if irblocks.is_empty() {
+                // The block is unmappable. Indicate the hole in the track with a None.
+                if ret_irblocks.last() != Some(&None) && !ret_irblocks.is_empty() {
+                    // Collapse repeated unmappable blocks to a single None.
+                    ret_irblocks.push(None);
+                }
             } else {
-                ret_irblocks.extend(irblocks);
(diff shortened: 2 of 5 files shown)


ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -267,7 +276,8 @@ extern "C" void *__ykllvmwrap_irtrace_compile(char *FuncNames[], size_t BBs[],
     errx(EXIT_FAILURE, "trace index counter overflowed");
 
   // Get var args from start_tracing call.
-  auto Inputs = get_trace_inputs(AOTMod->getFunction(FuncNames[0]), BBs[0]);
+  // auto Inputs = get_trace_inputs(AOTMod->getFunction(FuncNames[0]), BBs[0]);
```
ptersilie, line 279: Left-over comment.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -126,26 +126,35 @@ ThreadSafeModule *getThreadAOTMod(void) {
...
-      if (CI->getCalledFunction()->getName() == YKTRACE_START) {
-        // Skip first argument to start_tracing.
-        for (auto Arg = CI->arg_begin() + 1; Arg != CI->arg_end(); Arg++) {
-          Vec.push_back(Arg->get());
+
+  // Search for the YKTRACE_START call and extract its arguments.
+  for (size_t Idx = 0; Idx < Len; Idx++) {
+    Function *F = AOTMod->getFunction(FuncNames[Idx]);
+    if (F == nullptr) {
+      // unmappable blocks can't contain YKTRACE_START.
```
ptersilie, line 137: I'm not sure how we can end up with a trace where the first block does not contain `start_tracing`. I assumed that's always the first block in the trace. Do you have an example?

yktrace/src/hwt/mapper.rs
```diff
@@ -101,23 +101,25 @@ impl HWTMapper {
     }
 
     /// Maps each entry of a hardware trace back the IR block from whence it was compiled.
```
ptersilie, line 103: ...back **to** the IR block...

Also, I would prefer saying "from where" or rather even "from which" is was compiled. Whence just sounds odd.

ykllvmwrap/src/ykllvmwrap.cc
vext01: We are searching the functions in the AOTMod that the trace touches here.

I'm working on a change which might mean we can revert this hunk back to it's simpler form.

yktrace/src/lib.rs
```diff
@@ -50,22 +50,30 @@ impl IRBlock {
 
 /// An LLVM IR trace.
 pub struct IRTrace {
-    blocks: Vec<IRBlock>,
+    // The blocks of the trace. A None element represents an arbitrarily sized portion of unknown
+    // code which we were unable to determine LLVM IR for, e.g. external library code. The length
+    // of the unknown code is unknown.
```
ptersilie, line 55: I think we've established the code is unknown so we can drop one of the "unknowns".

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Okay, but why don't we just search for `start_tracing` in block 0 of the main function, where I think it's guaranteed to be?

vext01: It was due to a) trace holes, and b) misunderstanding.

3d85f6e5357a reverts this function.

ptersilie: :+1:

yktrace/src/hwt/mapper.rs
vext01: It sound's quite old fashioned, but does have precedent in programming.

e.g. fseek:
```
     int
     fseek(FILE *stream, long offset, int whence);
```

I'll fix it anyway :)

ptersilie: I think with the renaming to yk we should continue to cut our ties to Shakespeare. ;-)

vext01:
That should be all remaining comments addressed. OK to squash?

ptersilie:
Yep. Please squash.

vext01:
splat

PR #371, opened 2021-06-29
Refactor ykllvmwrap.

ptersilie:
This PR refactors ykllvmwrap, in particular the function `__ykllvmwrap_irtrace_compiler` which has gotten rather big. The main reason for this refactoring is that it will make it easier for us to deal with recursive functions that we don't want to inline.

I'm sure there's some more code that can be moved into separate functions, but we can do that incrementally when it's needed. So this should be a good starting point for now. Though I'm open for suggestions.


ykllvmwrap/src/jitbuilder.cc
```diff
@@ -0,0 +1,306 @@
+using namespace llvm;
```
vext01, line 1: Can we have a comment at the top of the file explaining what this does.

ykllvmwrap/src/jitbuilder.cc
```diff
@@ -0,0 +1,306 @@
...
+using namespace std;
+// An atomic counter used to issue compiled traces with unique names.
```
vext01, line 3: probably need a new line between `using` and the following line.

ykllvmwrap/src/jitbuilder.cc
```diff
@@ -0,0 +1,306 @@
...
+
+class JITBuilder {
+  vector<GlobalVariable *> cloned_globals;
+  Module *JITMod;
+  // A pointer to the call to YKTRACE_START in the AOT module (once
+  // encountered). When this changes from NULL to non-NULL, then we start
+  // copying instructions from the AOT module into the JIT module.
+  Instruction *StartTracingInstr = nullptr;
+
+public:
```
vext01, line 44: Can we add comments to each field?

ykllvmwrap/src/jitbuilder.cc
```diff
@@ -0,0 +1,298 @@
...
+  // Skip to the first block in the trace which contains the `start_tracing`
+  // call.
+  std::advance(It, BBIdx);
+  BasicBlock *BB = &*It;
+  bool found = false;
+  for (auto I = BB->begin(); I != BB->end(); I++) {
+    if (isa<CallInst>(I)) {
+      CallInst *CI = cast<CallInst>(&*I);
+      if (CI->getCalledFunction()->getName() == YKTRACE_START) {
+        // Skip first argument to start_tracing.
```
ltratt, line 22: Why do we skip it?

ykllvmwrap/src/jitbuilder.cc
```diff
@@ -0,0 +1,298 @@
...
+    // Iterate over the PT trace and stitch together all traced blocks.
+    for (size_t Idx = 0; Idx < Len; Idx++) {
+      auto FuncName = FuncNames[Idx];
+
+      // FIXME: Deal with holes in the trace, e.g. calls to libc.
+      assert(FuncName != nullptr);
+
+      // Get a traced function so we can extract blocks from it.
+      Function *F = AOTMod->getFunction(FuncName);
+      if (!F)
```
ltratt, line 103: Would it be easier to use `assert`s for things like these that, I assume, we don't think can ever go wrong?

ykllvmwrap/src/jitbuilder.cc
```diff
@@ -0,0 +1,298 @@
...
+      // Iterate over all instructions within this block and copy them over
+      // to our new module.
+      for (auto I = BB->begin(); I != BB->end(); I++) {
+        // Skip calls to debug intrinsics (e.g. @llvm.dbg.value). We don't
+        // currently handle debug info and these "pseudo-calls" cause our blocks
+        // to be prematurely terminated.
+        if (isa<DbgInfoIntrinsic>(I))
+          continue;
+
+        // If we've returned from a call skip ahead to the instruction where we
```
ltratt, line 120: `,` in between "call" and "skip"?

ykllvmwrap/src/jitbuilder.cc
vext01: It's just the tracing kind (hardware of software). We never want to include that stuff.

ltratt:
Can we add some justification to commits like https://github.com/ykjit/yk/pull/371/commits/4413fce6ccc0ba765ea97bdfdb3576ee97f313b5 and https://github.com/ykjit/yk/pull/371/commits/2d32cdd686a0660e69e75d1f53e16479de953f32? In other words, not just *what* we've done, but *why* we've done it.

ykllvmwrap/src/jitbuilder.cc
ltratt: OK, let's explain that (perhaps in another PR). [I can tell the first argument is skipped from the "+1" in the code; but I don't know why!]

ykllvmwrap/src/jitbuilder.cc
vext01: It depends if you want a good error message or not in release mode (where C++ assertions are not compiled in). If `getFunction()` returned null and we'd used an assertion to check for this, in release mode we'd probably get a null pointer dereference somewhere down the line.

Personally, I think an explicit check is good here.

ltratt: This sounds like a debug assert sort of thing to me. I don't think we want to pay this penalty all the time?

[I do agree that we need a non-debug assert thing too. I don't know if one's available or whether we have to make one ourselves.]

vext01: The problem is, in C++ there's no such thing as a debug assert. There's one kind of assertion and they are either on (debug builds) or off (release builds with `-DNDEBUG`).

We could roll our own debug assertion code, but I'd say that's out of scope of Lukas' PR.

ltratt: Agreed it's out of scope here, but why don't we fix this in a PR in the next day or two? [Before we end up with loads of not-very-useful `errx` calls, slowing down release mode, and probably not making the code clearer either.]

ykllvmwrap/src/jitbuilder.cc
ptersilie: Why do I feel like this isn't the first time Laurie has commented on this? ;-)

ptersilie:
The reason for these changes is explained in the PR description. In essence, the function has become very difficult to work with, so moving things around and into separate functions will make out life easier down the line. I didn't think adding a message along those lines to each commit adds anything but I'm happy to do so.

ltratt:
PR descriptions don't end up in the commit history, so each commit really needs to stand on its own.

ykllvmwrap/src/jitbuilder.cc
vext01: I realised that I got my logic all wrong. Let me try again.

C assertions are enabled only in Cargo debug builds. Therefore an `assert()` in C++ code is in fact like a `debug_assert!`. What we don't have in our setup is *release assertions* for C++ code.

Whether or not we try to implement release assertions, this particular line of code can be an `assert()`.

Sorry for the screw-up!

ykllvmwrap/src/jitbuilder.cc
ptersilie: Fixed in 9f14d25.

ykllvmwrap/src/jitbuilder.cc
ptersilie: Fixed in 9f14d25.

ykllvmwrap/src/jitbuilder.cc
ptersilie: Fixed in 9f14d25.

ykllvmwrap/src/jitbuilder.cc
ptersilie: Fixed in 9f14d25.

ptersilie:
That should address most comments. The rest will be done in future PRs. I also need to squash and change the commit messages.

ykllvmwrap/src/jitbuilder.cc
ltratt: I meant the same as you, so we are now on the same page (though I assumed we were on the same page before ;)).

ykllvmwrap/src/jitbuilder.cc
```diff
@@ -0,0 +1,315 @@
...
+      Value *OldVal = Inputs[Idx];
+      Value *NewVal = JITFunc->getArg(Idx);
+      assert(NewVal->getType()->isPointerTy());
+      VMap[OldVal] = NewVal;
+    }
+
+    std::vector<CallInst *> inlined_calls;
+    CallInst *last_call = nullptr;
+
+    // Iterate over the PT trace and stitch together all traced blocks.
```
vext01, line 103: We shouldn't mention PT here, as the software tracer would share the same code, right?

ykllvmwrap/src/jitbuilder.cc
```diff
@@ -0,0 +1,315 @@
...
+
+          if (CF->getName() == YKTRACE_START) {
+            StartTracingInstr = &*I;
+            continue;
+          } else if (CF->getName() == YKTRACE_STOP) {
+            finalise(&Builder);
+            return JITMod;
+          } else {
+            // Skip remainder of this block and remember where we stopped so we
+            // can continue tracing from this position after returning frome the
```
vext01, line 151: s/tracing// ?

vext01: frome

ykllvmwrap/src/jitbuilder.cc
ptersilie: Fixed in 9ff6b8f.

ykllvmwrap/src/jitbuilder.cc
ltratt: @vext01 is correct. Proper nouns should be capitalised. https://en.wikipedia.org/wiki/Frome

ptersilie: Fixed in cdc0a33.

vext01:
This looks good to me. My only remaining comment is the name "jit builder". Should it be "trace builder", "module builder" or "inliner" maybe?

ptersilie:
I think my problem with trace builder is that we already got the trace. This turns the trace into a module that we then (just in time) compile. Technically this should be called `JITModuleBuilder`, but that was a bit long. Maybe `JITModBuilder`?

vext01:
What did you think of "inliner"?

ptersilie:
Why inliner?

vext01:
Oh. Well the motivation was that we are taking inter-procedural blocks from the AOT module and putting them in-line. But never mind.

`JITModBuilder` is fine then.

ptersilie:
Done.

vext01:
We should also rename the `.cc` file.

ptersilie:
I feared you would say this. I purposely didn't do this as it will be a huge pain to squash away.

vext01:
Do it as a separate commit if you like.

ptersilie:
Let's try.

ptersilie:
Okay to squash?

vext01:
Go ahead.

ptersilie:
Done.

ptersilie:
Forgot to update an include. Ok to squash?

vext01:
go ahead.

ptersilie:
Squashed.

PR #373, opened 2021-07-07
Turn inlined recursive functions into explicit calls.

ptersilie:
When tracing a program, PT automatically inlines any function call it encounters. Recursive functions are, unfortunately, bad candidates for inlining as they can quickly blow up the trace, which can significantly increase JIT compilation times.

This PR detects inlined recursive functions and turns them back into explicit calls.

Depends on https://github.com/ykjit/ykllvm/pull/4.

### DIFF
--- a/c_tests/tests/compile_fib.c
+++ b/c_tests/tests/compile_fib.c
@@ -1,0 +1,62 @@
+// Compiler:
+// Run-time:
+//   env-var: YK_PRINT_IR=1
+//   stderr:
+//     ...
+//     define internal void @__yk_compiled_trace_0(i32* %0, i32* %1) {
+//       %3 = load i32, i32* %1, align 4, !tbaa !0
+//       %4 = shl nsw i32 %3, 3
+//       %5 = icmp ult i32 %4, 3
+//       %6 = add nsw i32 %4, -2
+//       %7 = tail call fastcc i32 @fib(i32 %6, i32* %1)...
+//       %8 = add nsw i32 %4, -1
+//       %9 = tail call fastcc i32 @fib(i32 %8, i32* %1)...
+//       %10 = add nsw i32 %9, %7
+//       store i32 %10, i32* %1, align 4, !tbaa !0
+//       store i32 %10, i32* %0, align 4, !tbaa !0
+//       ret void
+//     }
+//     ...
+
+// Check that recursive function calls are not unrolled.
+//
+// FIXME An optimising compiler can remove all of the code between start/stop
+// tracing.
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk_testing.h>
+
+__attribute__((noinline)) int fib(int num, int *tcp) {
+  if (num == 0)
+    return 0;
+  if (num == 1)
+    return 1;
+  if (num == 2)
+    return 1;
+  int a = fib(num - 2, tcp);
+  int b = fib(num - 1, tcp);
+  int c = a + b;
+  *tcp = c; // Prevent tail call optimisation.
+  return c;
+}
+
+int main(int argc, char **argv) {
+  int res = 0;
+  void *tt = __yktrace_start_tracing(HW_TRACING, &res, &argc);
+  res = fib(argc * 8, &argc);
+  void *tr = __yktrace_stop_tracing(tt);
+  assert(res == 21);
+
+  void *ptr = __yktrace_irtrace_compile(tr);
+  __yktrace_drop_irtrace(tr);
+  void (*func)(void *, void *) = (void (*)(void *, void *))ptr;
+  int output = 0;
+  argc = 1;
+  func(&output, &argc);
+  assert(output == 21);
+
+  return (EXIT_SUCCESS);
+}
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -48,6 +48,8 @@ class JITModBuilder {
   Instruction *StartTracingInstr = nullptr;
 
 public:
+  // Store virtual addresses for called functions.
+  std::map<StringRef, uint64_t> globalMappings;
   // The function name of this trace.
   string TraceName;
   // Mapping from AOT instructions to JIT instructions.
@@ -57,8 +59,18 @@ public:
   ValueToValueMapTy RevVMap;
 #endif
 
+  Value *getMappedValue(Value *V) {
+    if (isa<Constant>(V)) {
+      return V;
+    } else {
+      auto NV = VMap[V];
+      return NV;
+    }
+  }
+
   Module *createModule(char *FuncNames[], size_t BBs[], size_t Len,
-                       Module *AOTMod) {
+                       Module *AOTMod, char *FAddrKeys[], uint64_t FAddrVals[],
+                       size_t FAddrLen) {
     LLVMContext &JITContext = AOTMod->getContext();
     JITMod = new Module("", JITContext);
     uint64_t TraceIdx = NextTraceIdx.fetch_add(1);
@@ -99,6 +111,8 @@ public:
 
     std::vector<CallInst *> inlined_calls;
     CallInst *last_call = nullptr;
+    size_t call_stack = 0;
+    CallInst *noinline_func = nullptr;
 
     // Iterate over the trace and stitch together all traced blocks.
     for (size_t Idx = 0; Idx < Len; Idx++) {
@@ -134,9 +148,11 @@ public:
           }
           continue;
         }
+
         if (isa<CallInst>(I)) {
           CallInst *CI = cast<CallInst>(&*I);
           Function *CF = CI->getCalledFunction();
+          StringRef CFName = CF->getName();
           if (CF == nullptr)
             continue;
 
@@ -147,22 +163,58 @@ public:
             finalise(&Builder);
             return JITMod;
           } else {
+            if (AOTMod->getFunction(CFName) != nullptr && call_stack > 0) {
+              // When ignoring an inlined function, we need to count other
+              // inlined function calls so we know when we left the initial
+              // function call.
+              call_stack += 1;
+              inlined_calls.push_back(CI);
+              break;
+            }
+            // If this is a recursive call that has been inlined, remove the
+            // inlined code and turn it into a normal call.
+            for (CallInst *cinst : inlined_calls) {
+              // Have we inlined this call already? Then this is recursion.
+              if (cinst->getCalledFunction() == CF) {
+                if (VMap[CF] == nullptr) {
+                  // Declare function.
+                  auto DeclFunc = llvm::Function::Create(
+                      CF->getFunctionType(), GlobalValue::ExternalLinkage,
+                      CFName, JITMod);
+                  VMap[CF] = DeclFunc;
+                  for (size_t i = 0; i < FAddrLen; i++) {
+                    char *FName = FAddrKeys[i];
+                    uint64_t FAddr = FAddrVals[i];
+                    if (strcmp(FName, CFName.data()) == 0) {
+                      globalMappings.insert(
+                          pair<StringRef, uint64_t>(CFName, FAddr));
+                      break;
+                    }
+                  }
+                }
+                copyInstruction(&Builder, (Instruction *)&*I);
+                noinline_func = CI;
+                call_stack = 1;
+                break;
+              }
+            }
             // Skip remainder of this block and remember where we stopped so we
             // can continue from this position after returning from the inlined
             // call.
-            // FIXME Deal with calls we cannot or don't want to inline.
             if (StartTracingInstr != nullptr) {
               inlined_calls.push_back(CI);
               // During inlining, remap function arguments to the variables
               // passed in by the caller.
-              for (unsigned int i = 0; i < CI->arg_size(); i++) {
-                Value *Var = CI->getArgOperand(i);
-                Value *Arg = CF->getArg(i);
-                // If the operand has already been cloned into JITMod then we
-                // need to use the cloned value in the VMap.
-                if (VMap[Var] != nullptr)
-                  Var = VMap[Var];
-                VMap[Arg] = Var;
+              if (call_stack == 0) {
+                for (unsigned int i = 0; i < CI->arg_size(); i++) {
+                  Value *Var = CI->getArgOperand(i);
+                  Value *Arg = CF->getArg(i);
+                  // If the operand has already been cloned into JITMod then we
+                  // need to use the cloned value in the VMap.
+                  if (VMap[Var] != nullptr)
+                    Var = VMap[Var];
+                  VMap[Arg] = Var;
+                }
               }
               break;
             }
@@ -182,17 +234,34 @@ public:
         if (isa<ReturnInst>(I)) {
           last_call = inlined_calls.back();
           inlined_calls.pop_back();
+          if (call_stack > 0) {
+            call_stack -= 1;
+            if (call_stack == 0) {
+              last_call = noinline_func;
+            }
+            continue;
+          }
           // Replace the return variable of the call with its return value.
           // Since the return value will have already been copied over to the
           // JITModule, make sure we look up the copy.
           auto OldRetVal = ((ReturnInst *)&*I)->getReturnValue();
-          if (isa<Constant>(OldRetVal)) {
-            VMap[last_call] = OldRetVal;
-          } else {
-            auto NewRetVal = VMap[OldRetVal];
-            VMap[last_call] = NewRetVal;
-          }
+          VMap[last_call] = getMappedValue(OldRetVal);
           break;
+        }
+
+        if (call_stack > 0) {
+          // We are currently ignoring an inlined function.
+          continue;
+        }
+
+        if (isa<PHINode>(I)) {
+          assert(Idx > 0);
+          auto LBIt = F->begin();
+          std::advance(LBIt, BBs[Idx - 1]);
+          BasicBlock *LastBlock = &*LBIt;
+          Value *V = ((PHINode *)&*I)->getIncomingValueForBlock(LastBlock);
+          VMap[&*I] = getMappedValue(V);
+          continue;
         }
 
         // If execution reaches here, then the instruction I is to be copied
--- a/yktrace/src/hwt/mapper.rs
+++ b/yktrace/src/hwt/mapper.rs
@@ -8,8 +8,11 @@ use memmap2;
 use object::{Object, ObjectSection};
 use once_cell::sync::Lazy;
 use std::{
+    collections::HashMap,
     convert::TryFrom,
-    env, fs,
+    env,
+    ffi::CString,
+    fs,
     io::{prelude::*, Cursor, SeekFrom},
 };
 use ykllvmwrap::symbolizer::Symbolizer;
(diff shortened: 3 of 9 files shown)


ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -48,6 +48,8 @@ class JITModBuilder {
   Instruction *StartTracingInstr = nullptr;
 
 public:
+  // Store virtual addresses for called functions.
+  std::map<StringRef, uint64_t> globalMappings;
```
vext01, line 52: Should that be `GlobalMappings`, with a capital `G`?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -99,6 +111,8 @@ class JITModBuilder {
 
     std::vector<CallInst *> inlined_calls;
     CallInst *last_call = nullptr;
+    size_t call_stack = 0;
```
vext01, line 114: Would `StackDepth` or `CallDepth` be better? Either way, it needs to be C++ cased.

vext01: (There are lots of places where the variable casing isn't correct)

yktrace/src/hwt/mapper.rs
```diff
@@ -91,17 +96,22 @@ impl BlockMap {
 
 pub struct HWTMapper {
     symb: Symbolizer,
+    faddrs: HashMap<CString, u64>,
```
vext01, line 99: We should comment the members here.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -57,8 +59,18 @@ class JITModBuilder {
   ValueToValueMapTy RevVMap;
 #endif
 
+  Value *getMappedValue(Value *V) {
```
ltratt, line 62: Silly question perhaps: *which* Map? I'd have imagined there can be more than one, at least realistically in the future?

ltratt:
We'll definitely need some IR-printing tests for this one, because it's something where we can't tell if the PR is doing the right thing when observing the program as a black box.

One thing I'm unsure about is how difficult "turn a recursive call into a normal call" becomes. Imagine I have something like:

```rust
fn f(x: Y) {
  <blah}
  f(z);
}
```

when it's unrolled to something like:

```
  blah
  blah
```

how do we get the argument `z` (and there may be multiple arguments) out of the trace and into a parameter (so to speak).

ykllvmwrap/src/jitmodbuilder.cc
vext01: I see what you are saying, but the LLVM terminology is (for better or probably worse) `ValueMapper`, as in it maps `Value *` to `Value *` :P

> #include <llvm/Transforms/Utils/ValueMapper.h>

ltratt: Yes, but we can use a value mapper to do different things I assume? Is there a way of giving *this* valuemapper a more descriptive name?

vext01: Well, we have a second map for debug builds, which we called the "reverse vmap" as opposed to the (forward) vmap.

As far as I can see, we won't need more, but let's see what @ptersilie thinks.

ltratt: Even so, if we can give it a better name, we should.

vext01:
I think since we inline `f` to depth one, `z` will be handled as-per-usual in the first inlined instance.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Under protest, yes. I just can't get used to this weird way of naming things. We should have never adopted LLVM's way here.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: You mean LLVM cased. `CallDepth` works for me.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: I think we only need multiple maps if we want to inline recursive calls. At least that was one case where I experimented with multiple maps (which frankly was a nightmare so I hope I won't have to do that again).

So for now there's only one map. If you want to future proof this, we would have to pass the map to use into the `getMappedValue` function, which for now would serve no purpose.

ptersilie:
Yes, the first function call of a recursive function is inlined. Only subsequent calls are skipped. Though, I'm not sure that was what @ltratt was asking.

Also, I realised I forgot to add the Fibonacci test. Maybe that will put Laurie's mind at ease?

ltratt:
I'm fairly convinced that we need an IR-printing part of the test, otherwise how do we know that the IR we create doesn't unroll recursion?

ptersilie:
> I'm fairly convinced that we need an IR-printing part of the test, otherwise how do we know that the IR we create doesn't unroll recursion?

If it would unroll the recursive call the trace wouldn't compile as we need to do some extra work to deal with this. But I agree that an IR test won't hurt.

ltratt:
> If it would unroll the recursive call the trace wouldn't compile as we need to do some extra work to deal with this.

Yes, but at some point we'll probably accidentally fix that :p Anyway, it sounds like we agree about an IR test :)

ptersilie:
@ltratt See if this works.

vext01:
I worry that our IR tests are too specific. For example we are checking that in the optimised cases, that recursion has been sequentialised. But LLVM may not have, and that'd have been valid too. Also all of the variable numbers are checked verbatim, but you could increment them all by one and the IR would remain valid.

Silly, but honest question: What are the goals of our IR tests?

 - To check a property of something LLVM does, e.g. optimisation?
 - To check a general property about the structure of the generated IR?

Too far toward the former and you get very fragile tests.

Yet the latter can't be done without an element of the former...

ltratt:
@ptersilie Silly question perhaps: why are there so many `tail call fastcc i32 @fib(`s? I expected 1, but perhaps I got the wrong end of the stick? Maybe this is because (I think) it's printing the compile-time IR and I was expected the trace IR to be printed?

@vext01 The IR tests are definitely fragile, and there's absolutely nothing we can do about that. We are at the mercy of how LLVM compiles our code -- if it changes our tests break. `fm` tries to ameliorate that by providing ways of abstracting away from things like variable names, which will help, but is still imperfect. But, all that said, the only thing worse than IR tests would not be having IR tests.

vext01:
> why are there so many tail call fastcc i32 @fib

I think that's what lukas was saying about LLVM sequentialising the recursion.

> fm tries to ameliorate that by providing ways of abstracting away from things like variable names

Right, which we are not yet using. But should we be? In the tests so far, all variables are verbatim.

ptersilie:
> I worry that our IR tests are too specific.

I have thought about this when I wrote the first test. And my conclusion was that I rather like to know if LLVM changes in a way that changes our variable numbers, so I can then check why it has changed and whether this affects other parts. In our cases a variable name change could also mean that an instruction has moved somewhere else or the order has changed. I'd like to catch when that happens. To be fair I don't expect this to happen too often that I'm okay with adapting the tests when it does.

> Silly question perhaps: why are there so many tail call fastcc i32 @fib

I think this has to do with the way LLVM optimises recursive calls. Though now that I think about it, technically I would expect to only see a single call to fib, since LLVM "inlines" (or sequentialises) the first recursive call. Let me look into it to make sure I understand what's happening.

ltratt:
@vext01 At the moment we don't really know how LLVM changes will effect our tests: perhaps variables will never be renumbered, perhaps they will regularly. It's probably best to start off with concrete variable numbers and see how much pain that causes us over the coming months.

@ptersilie 

> Though now that I think about it, technically I would expect to only see a single call to fib, since LLVM "inlines" (or sequentialises) the first recursive call. Let me look into it to make sure I understand what's happening.

Thanks!

ptersilie:
Okay, I think I got it now. If we the recursive loop would be something like:
```
def fib(num):
   if num < 1:
      return 1
   fib(num - 1)
```
Then we would only see a single call to fib (for all `a`), because LLVM turns the recursive call to `fib` into a branch instruction, so it effectively turns the recursion into a loop. Now in our example, we have, however, two calls to fib, e.g. `fib(num - 2)` and `fib(num - 1)`. LLVM can only sequentialise one of the those calls, so `fib(num - 2)`, is being kept as a `fast call` whereas `fib(num - 1)` is turned into a loop. So for each iteration of that loop, see another call to `fib(num - 2)`. So we see 6 calls into total:
```
main
  fib(8)
    fib(6)
    // fib(7) done sequentially
    fib(5)
    // fib(6) done sequentially
    fib(4)
    // fib(5) done sequentially
    fib(3)
    // fib(3) done sequentially
    fib(1)
    // fib(2) done sequentially
```

vext01:
Probably because only one of the two recursive calls can be tail calls? Anyway, impressive!

ltratt:
I must admit that I am very confused! Is that IR the post-LLVM-optimised IR? Can we print out the IR of the trace we put into LLVM?

vext01:
Lukas is AFK, so I'll take over as far as I can.

Here's Lukas' Fib implementation from the test:
```
__attribute__((noinline)) int fib(int num) {                                                                                                                                                                       
  if (num == 0)                                                                                                                                                                                                    
    return 0;                                                                                                                                                                                                      
  if (num == 1)                                                                                                                                                                                                    
    return 1;                                                                                                                                                                                                      
  if (num == 2)                                                                                                                                                                                                    
    return 1;                                                                                                                                                                                                      
  int a = fib(num - 2);                                                                                                                                                                                            
  int b = fib(num - 1);                                                                                                                                                                                            
  int c = a + b;                                                                                                                                                                                                   
  return c;                                                                                                                                                                                                        
}
```

Here's the AOT `fib` function:
```
define internal fastcc i32 @fib(i32 %0) unnamed_addr #0 !dbg !15 {
  br label %2, !dbg !25

2:                                                ; preds = %6, %1
  %3 = phi i32 [ 0, %1 ], [ %10, %6 ]
  %4 = phi i32 [ %0, %1 ], [ %9, %6 ]
  call void @llvm.dbg.value(metadata i32 %4, metadata !21, metadata !DIExpression()), !dbg !26
  %5 = icmp ult i32 %4, 3, !dbg !27
  br i1 %5, label %11, label %6, !dbg !27

6:                                                ; preds = %2
  %7 = add nsw i32 %4, -2, !dbg !28
  %8 = tail call fastcc i32 @fib(i32 %7), !dbg !29
  call void @llvm.dbg.value(metadata i32 %8, metadata !22, metadata !DIExpression()), !dbg !26
  %9 = add nsw i32 %4, -1, !dbg !30
  call void @llvm.dbg.value(metadata i32 undef, metadata !23, metadata !DIExpression()), !dbg !26
  %10 = add nsw i32 %8, %3, !dbg !31
  call void @llvm.dbg.value(metadata i32 %10, metadata !24, metadata !DIExpression()), !dbg !26
  br label %2, !dbg !25

11:                                               ; preds = %2
  %12 = icmp ult i32 %0, 2, !dbg !25
  %13 = select i1 %12, i32 %0, i32 2, !dbg !25
  %14 = zext i32 %13 to i64, !dbg !27
  %15 = getelementptr inbounds [3 x i32], [3 x i32]* @switch.table.fib, i64 0, i64 %14, !dbg !27
  %16 = load i32, i32* %15, align 4, !dbg !27
  %17 = add nsw i32 %16, %3, !dbg !31
  ret i32 %17, !dbg !32
} 
```

And here's the IR trace we have assembled:
```
define internal void @__yk_compiled_trace_0(i32* %0) {                                                                                                                                                             
  %2 = icmp ult i32 8, 3                                                                                                                                                                                           
  %3 = add nsw i32 8, -2                                                                                                                                                                                           
  %4 = tail call fastcc i32 @fib(i32 %3)                                                                                                                                                                           
  %5 = add nsw i32 8, -1                                                                                                                                                                                           
  %6 = add nsw i32 %4, 0                                                                                                                                                                                           
  %7 = icmp ult i32 %5, 3                                                                                                                                                                                          
  %8 = add nsw i32 %5, -2                                                                                                                                                                                          
  %9 = tail call fastcc i32 @fib(i32 %8)                                                                                                                                                                           
  %10 = add nsw i32 %5, -1                                                                                                                                                                                         
  %11 = add nsw i32 %9, %6                                                                                                                                                                                         
  %12 = icmp ult i32 %10, 3                                                                                                                                                                                        
  %13 = add nsw i32 %10, -2                                                                                                                                                                                        
  %14 = tail call fastcc i32 @fib(i32 %13)                                                                                                                                                                         
  %15 = add nsw i32 %10, -1                                                                                                                                                                                        
  %16 = add nsw i32 %14, %11                                                                                                                                                                                       
  %17 = icmp ult i32 %15, 3                                                                                                                                                                                        
  %18 = add nsw i32 %15, -2                                                                                                                                                                                        
  %19 = tail call fastcc i32 @fib(i32 %18)                                                                                                                                                                         
  %20 = add nsw i32 %15, -1                                                                                                                                                                                        
  %21 = add nsw i32 %19, %16                                                                                                                                                                                       
  %22 = icmp ult i32 %20, 3                                                                                                                                                                                        
  %23 = add nsw i32 %20, -2                                                                                                                                                                                        
  %24 = tail call fastcc i32 @fib(i32 %23)                                                                                                                                                                         
  %25 = add nsw i32 %20, -1                                                                                                                                                                                        
  %26 = add nsw i32 %24, %21                                                                                                                                                                                       
  %27 = icmp ult i32 %25, 3                                                                                                                                                                                        
  %28 = add nsw i32 %25, -2                                                                                                                                                                                        
  %29 = tail call fastcc i32 @fib(i32 %28)                                                                                                                                                                         
  %30 = add nsw i32 %25, -1                                                                                                                                                                                        
  %31 = add nsw i32 %29, %26                                                                                                                                                                                       
  %32 = icmp ult i32 %30, 3                                                                                                                                                                                        
  %33 = icmp ult i32 8, 2                                                                                                                                                                                          
  %34 = select i1 %33, i32 8, i32 2                                                                                                                                                                                
  %35 = zext i32 %34 to i64                                                                                                                                                                                        
  %36 = getelementptr inbounds [3 x i32], [3 x i32]* @switch.table.fib, i64 0, i64 %35                                                                                                                             
  %37 = load i32, i32* %36, align 4                                                                                                                                                                                
  %38 = add nsw i32 %37, %31                                                                                                                                                                                       
  store i32 %38, i32* %0, align 4, !tbaa !0                                                                                                                                                                        
  ret void                                                   
}
```

Notice how the AOT module already replaced the second recursive call to fib with a lookup table? I think that's because it has figured out that in `main` we call fib with the constant `8`, so it can memoise all of the potential outcomes. Not sure why it didn't do that for the first recursive call mind.

vext01:
So why all the calls to `fib` in the trace?

vext01:
Ah bb6 jumps back to bb2.

So the first recursive call is in a loop, for some reason. Compiler magic?

vext01:
Do I see it correctly that it uses the loop until the final input value for the fib computation is less than three, then looks up the the last fib result from a lookup table? How bizarre :P

ptersilie:
No, it doesn't memoise anything. The lookup table is for `0, 1, 2`. But it turned one of the recursive calls into a loop by cleverly using phi nodes. `%9 = add nsw i32 %4, -1` is what we pass to `fib (num - 1)`. But watch how it then just loops back to `bb2`, where the phi node reads out `%9` into `%4` and then it goes round and round until `%4` is smaller than `3` at which point the lookup table comes into play. The result is accumulated in `%3`.

vext01:
Yeah agreed. I just took me a while to see what was going on :P

So that explains the repeated calls to fib anyway.

ptersilie:
However, since our fib function has two calls to `fib`, LLVM can only turn one into a loop. So on each iteration we see the other call to fib, and that's why we see 6 of them.

ltratt:
If we call `fib(n)` does this lead to (roughly) `n/2` tail calls to `fib` in our trace? Or is 5 or 6 the maximum possible number?

ptersilie:
I'm afraid that for bigger `n`s we would get bigger and bigger traces, since we are essentially unrolling a loop here.

vext01:
> we are essentially unrolling a loop here.

Agreed. That is now separate from recursive calls.

A future PR would therefore be: unrolling loops to a bound, or maybe only if they are annotated?

ptersilie:
Yes, I believe by looking at the jumps we should be able to turn the unrolled trace back into a loop.

ltratt:
OK, so just to make sure I understand: our PR to stop unrolling for recursive function calls either a) doesn't work at all b) only works for some cases? I think the latter and the "fib" example is an example of a case it doesn't work for?

ptersilie:
No it works. We never inline a recursive call. It's just that in the fib example, LLVM has turned one of the recursive calls into a loop, which the PT trace has automatically unrolled. So in a separate PR we need to detect loop unrolling and revert it.

ltratt:
Ah, right, our terminology is broken (I think). What you're saying is that the *AOT IR* is unrolled, right?

An easy way to stop this happening would be to not use a constant in `compile_fib`. I would probably do `res = fib(argc);` and then in the lang_test call it with 8 arguments.

vext01:
> res = fib(argc)

fine by me.

ptersilie:
No, that is not quite what I am saying. Let's imagine a recursive function with only one recursive call:
```
def foo(num):
    if num == 1:
        return num
    return foo(num - 1)
```
-O2 IR would then be something like:
```
0:
  br 1
1:
  if num == 1:
     br 3
  br 2
2:
  num = num - 1
  br 1
3:
  return
```
So LLVM has turned the recursive call into a loop using branch instructions (and phi nodes - not seen here). The loop itself is not unrolled in the AOTMod. However, when we execute that program, the PR tracer unrolls that loop, so we see a trace like: `0 1 2 1 2 1 2 1 2 1 3`.

So I don't think your above trick would help against this. We need to detect the unrolling when assembling the JITModule from that trace and turn it back into a proper loop.

ltratt:
> So I don't think your above trick would help against this

Can we try? Because if it doesn't, then I have totally misunderstood what you've said earlier in the PR :p

ptersilie:
s/I don't think/I know.

The AOTMod will look exactly the same no matter what I pass into fib. But I've double checked just for you and replaced it with `argc` and no surprises there.

ltratt:
Hmm, OK. I have a theory. What happens if we pass (say) 256 to `argc`?

ltratt:
OK, offline discussion has resolved this: the problem with our fib example is that one call becomes a tail call, which becomes a loop, which PT unrolls; but the other call does not become a tail call and is not unrolled. Seeing both in example blew my little mind!

c_tests/tests/compile_fib.c
```diff
@@ -0,0 +1,59 @@
...
+//       %14 = tail call fastcc i32 @fib(...
+//       ...
+//       %19 = tail call fastcc i32 @fib(...
+//       ...
+//       %24 = tail call fastcc i32 @fib(...
+//       ...
+//       %29 = tail call fastcc i32 @fib(...
+//       ...
+
+// Check that basic trace compilation works.
```
vext01, line 21: This comment should be updated btw.

ptersilie:
@ltratt See if that last commit works for you. @vext01 and I made sure that tail call optimisation is prevented now. Even if LLVM might still be able to optimise this at some point, we would then notice, since the `-O2` and higher tests, explicitly check that there are two calls to `fib`.

ltratt:
I like it! Please squash.

ptersilie:
Squashed! Requires https://github.com/ykjit/ykllvm/pull/4 before merging.

PR #374, opened 2021-07-09
Handle tracing over inline asm.

vext01:
In LLVM inline asm manifests as a call to a special InlineAsm value. The
value can be treated as a constant and copied straight into the JITMod.

### DIFF
--- a/c_tests/tests/inline_asm.c
+++ b/c_tests/tests/inline_asm.c
@@ -1,0 +1,42 @@
+// Compiler:
+// Run-time:
+//   env-var: YKD_PRINT_IR=1
+//   stderr:
+//     ...
+//     ...call i32 asm "mov $$5, $0"...
+//     ...
+
+// Check that we can handle inline asm properly.
+//
+// FIXME An optimising compiler can remove all of the code between start/stop
+// tracing.
+
+#include <assert.h>
+#include <stdlib.h>
+#include <yk_testing.h>
+
+int main(int argc, char **argv) {
+  int res = 0;
+  void *tt = __yktrace_start_tracing(HW_TRACING, &res);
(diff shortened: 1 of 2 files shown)


c_tests/tests/inline_asm.c
```diff
@@ -0,0 +1,38 @@
...
+
+#include <assert.h>
+#include <stdlib.h>
+#include <yk_testing.h>
+
+int main(int argc, char **argv) {
+  int res = 0;
+  void *tt = __yktrace_start_tracing(HW_TRACING, &res);
+  // Stores the constant 5 into `res`.
+  asm("mov $5, %0"
```
ltratt, line 22: I think we should guard this with something like:

```c
#if <x86>
asm(...)
#else
#error asm statement not tested on this platform
```

just so that if/when we port to another platform we still test this properly.

vext01: Fixed in a0738bf

vext01:
OK to squash?

ltratt:
Please squash.

vext01:
splat.

vext01:
I can't reproduce this, even when running again under CI :\

ltratt:
Let's try again and see what happens.

bors r+

vext01:
bors try

vext01:
bors try

vext01:
OK, so by inspecting the worker's left-over sandbox I can see what's up.

The test uses `YK_PRINT_IR` instead of `YKD_PRINT_IR`. This all works locally because the branch is out of sync.

I suspect this will now work. If it does, should I sync the branch?

bors try

ltratt:
Please do!

PR #376, opened 2021-07-13
Add a NOOPT macro to try and block optimisations in tests.

vext01:
**This is a draft PR for discussion for now. I'll adapt the existing tests if this is the right approach.**

For context, see:
https://stackoverflow.com/questions/37786547/enforcing-statement-order-in-c/38025837#38025837

We use (the latest version of) the method proposed there.

In the new test, I checked that without using NOOPT, the compiler was
able to optimise `x = 2; y = 3; res = x + y` to `res = 5`. And that with
NOOPT, the addition remains intact.

### DIFF
--- a/c_tests/tests/noopts.c
+++ b/c_tests/tests/noopts.c
@@ -1,0 +1,58 @@
+// Compiler:
+// Run-time:
+//   env-var: YKD_PRINT_IR=aot,jit-pre-opt,jit-post-opt
+//   stderr:
+//     ...
+//     --- Begin aot ---
+//     ...
+//     store i32 2...
+//     ...
+//     ...add...
+//     ...
+//     --- End aot ---
+//     --- Begin jit-pre-opt ---
+//     ...
+//     store i32 2...
+//     ...
+//     ...add...
+//     ...
+//     --- End jit-pre-opt ---
+//     --- Begin jit-post-opt ---
+//     ...
+//     store i32 2...
+//     ...
+//     ...add...
+//     ...
+//     --- End jit-post-opt ---
+//     ...
+
+// Check that our NOOPT_VAL macro blocks optimisations.
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk_testing.h>
+
+int main(int argc, char **argv) {
+  int x, res;
+  void *tt = __yktrace_start_tracing(HW_TRACING, &res, &x);
+  // __yktrace_start_tracing() will already inhibit optimisations on `x` since
+  // its address is passed as an argument and the compiler must assume that it
+  // may be mutated. Therefore to properly test NOOPT_VAL we need to initialise
+  // `x` after the call to __yktrace_start_tracing..
+  x = 2;
+  NOOPT_VAL(x);
+  res = x + 3; // We don't want this operation to be optimised away.
+  void *tr = __yktrace_stop_tracing(tt);
+  assert(res == 5);
+
+  void *ptr = __yktrace_irtrace_compile(tr);
+  __yktrace_drop_irtrace(tr);
+  void (*func)(int *, int *) = (void (*)(int *, int *))ptr;
+  int res2 = 0;
+  func(&res2, &x);
+  assert(res2 == 5);
+
+  return (EXIT_SUCCESS);
+}
--- a/ykcapi/yk_testing.h
+++ b/ykcapi/yk_testing.h
@@ -20,3 +20,19 @@ void __yktrace_irtrace_get(void *trace, size_t idx, char **res_func,
                            size_t *res_bb);
 void *__yktrace_irtrace_compile(void *trace);
 void __yktrace_drop_irtrace(void *trace);
+
+// Blocks the compiler from optimising the specified value or expression.
+//
+// This is similar to the non-const variant borrowed from Google benchmark:
+// https://github.com/google/benchmark/blob/e451e50e9b8af453f076dec10bd6890847f1624e/include/benchmark/benchmark.h#L350
+//
+// Our version works on a value, rather than a pointer.
+//
+// Note that Google Benchmark also defines a variant for constant data. At the
+// time of writing, NOOPT_VAL seems to suffice (even for constants), but we may
+// need to consider using the const version later.
+#ifdef __clang__
+#define NOOPT_VAL(X) asm volatile("" : "+r,m"(X) : : "memory");
+#else
+#error non-clang compilers are not supported.
+#endif
--- a/ykllvmwrap/src/ykllvmwrap.cc
+++ b/ykllvmwrap/src/ykllvmwrap.cc
@@ -5,11 +5,13 @@
 #endif
 
 #include "llvm/ExecutionEngine/Orc/ThreadSafeModule.h"
+#include "llvm/Transforms/IPO/PassManagerBuilder.h"
 #include <llvm/DebugInfo/Symbolize/Symbolize.h>
 #include <llvm/ExecutionEngine/ExecutionEngine.h>
 #include <llvm/ExecutionEngine/MCJIT.h>
 #include <llvm/IR/IRBuilder.h>
 #include <llvm/IR/LLVMContext.h>
+#include <llvm/IR/LegacyPassManager.h>
 #include <llvm/IR/Module.h>
 #include <llvm/IR/Verifier.h>
 #include <llvm/IRReader/IRReader.h>
@@ -61,6 +63,146 @@ thread_local bool ThreadAOTModInitialized = false;
 
 // Flag used to ensure that LLVM is initialised only once.
 once_flag LLVMInitialised;
+
+#ifndef NDEBUG
+// Left trim (in-place) the character `C` from the string `S`.
+void lTrim(string &S, const char C) {
+  S.erase(0, std::min(S.find_first_not_of(C), S.size() - 1));
+}
+
+// Dumps an LLVM Value to a string and trims leading whitespace.
+void dumpValueToString(Value *V, string &S) {
+  raw_string_ostream RSO(S);
+  V->print(RSO);
+  lTrim(S, ' ');
+}
+#endif
+
+enum DebugIR {
+  AOT,
+  JITPreOpt,
+  JITPreOptSBS,
+  JITPostOpt,
+};
+
+class DebugIRPrinter {
+private:
+  bitset<4> toPrint;
+
+  const char *debugIRStr(DebugIR IR) {
+    switch (IR) {
+    case DebugIR::AOT:
+      return "aot";
+    case DebugIR::JITPreOpt:
+      return "jit-pre-opt";
+    case DebugIR::JITPreOptSBS:
+      return "jit-pre-opt-sbs";
+    case DebugIR::JITPostOpt:
+      return "jit-post-opt";
+    default:
+      errx(EXIT_FAILURE, "unreachable");
+    }
+  }
+
+public:
+  DebugIRPrinter() {
+    char *Env = std::getenv("YKD_PRINT_IR");
+    char *Val;
+    while ((Val = strsep(&Env, ",")) != nullptr) {
+      if (strcmp(Val, "aot") == 0)
+        toPrint.set(DebugIR::AOT);
+      else if (strcmp(Val, "jit-pre-opt") == 0)
+        toPrint.set(DebugIR::JITPreOpt);
+#ifndef NDEBUG
+      else if (strcmp(Val, "jit-pre-opt-sbs") == 0)
+        toPrint.set(DebugIR::JITPreOptSBS);
+#endif
+      else if (strcmp(Val, "jit-post-opt") == 0)
+        toPrint.set(DebugIR::JITPostOpt);
+      else
+        errx(EXIT_FAILURE, "invalid parameter for YKD_PRINT_IR: '%s'", Val);
+    }
+  }
+
+#ifndef NDEBUG
+  // Print a trace's instructions "side-by-side" with the instructions from
+  // which they were derived in the AOT module.
+  void printSBS(Module *AOTMod, Module *JITMod, ValueToValueMapTy &RevVMap) {
+    assert(JITMod->size() == 1);
+    Function *JITFunc = &*JITMod->begin();
+
+    // Find the longest instruction from the JITMod so that we can align the
+    // second column.
+    size_t LongestJITLine = 0;
+    for (auto &JITBlock : *JITFunc) {
+      for (auto &JITInst : JITBlock) {
+        string Line;
+        dumpValueToString(&JITInst, Line);
+        auto Len = Line.length();
+        if (Len > LongestJITLine)
+          LongestJITLine = Len;
+      }
+    }
+
+    const string JITHeader = string("Trace");
+    string Padding = string(LongestJITLine - JITHeader.length(), ' ');
+    errs() << "\n\n--- Begin trace dump for " << JITFunc->getName() << " ---\n";
+    errs() << JITHeader << Padding << "  | AOT\n";
+
+    // Keep track of the AOT function we are currently in so that we can print
+    // inlined function thresholds in the dumped trace.
+    StringRef LastAOTFunc;
+    for (auto &JITBlock : *JITFunc) {
+      for (auto &JITInst : JITBlock) {
+        auto V = RevVMap[&JITInst];
+        if (V == nullptr) {
+          // The instruction wasn't cloned from the AOTMod, so print it only in
+          // the JIT column and carry on.
+          std::string Line;
+          dumpValueToString((Value *)&JITInst, Line);
+          errs() << Line << "\n";
+          continue;
+        }
+        Instruction *AOTInst = (Instruction *)&*V;
+        assert(AOTInst != nullptr);
+        Function *AOTFunc = AOTInst->getFunction();
+        assert(AOTFunc != nullptr);
+        StringRef AOTFuncName = AOTFunc->getName();
+        if (AOTFuncName != LastAOTFunc) {
+          // Print an inlining threshold.
+          errs() << "# " << AOTFuncName << "()\n";
+          LastAOTFunc = AOTFuncName;
+        }
+        string JITStr;
+        dumpValueToString((Value *)&JITInst, JITStr);
+        string Padding = string(LongestJITLine - JITStr.length(), ' ');
+        string AOTStr;
+        dumpValueToString((Value *)AOTInst, AOTStr);
+        errs() << JITStr << Padding << "  |  " << AOTStr << "\n";
+      }
+    }
+    errs() << "--- End trace dump for " << JITFunc->getName() << " ---\n";
+  }
+#endif
+
+  void print(enum DebugIR IR, Module *M, Module *SBSAOTMod = nullptr,
+             ValueToValueMapTy *SBSRevVMap = nullptr) {
+    if (toPrint[IR]) {
+      if (IR == DebugIR::JITPreOptSBS) {
+#ifndef NDEBUG
+        printSBS(SBSAOTMod, M, *SBSRevVMap);
+#endif
+      } else {
+        // We print begin/end markers so that we can more test the IR at
+        // specific stages in the JIT pipeline (by anchoring matches to the
+        // begin/end markers).
+        errs() << "--- Begin " << DebugIRPrinter::debugIRStr(IR) << " ---\n";
+        M->dump();
+        errs() << "--- End " << DebugIRPrinter::debugIRStr(IR) << " ---\n";
+      }
+    }
+  }
+};
 
 // Initialise LLVM for JIT compilation. This must be executed exactly once.
 void initLLVM(void *Unused) {
@@ -133,6 +275,7 @@ extern "C" void *compileModule(string TraceName, Module *M,
   string ErrStr;
   ExecutionEngine *EE =
       EngineBuilder(std::move(MPtr))
+          .setEngineKind(EngineKind::JIT)
           .setMemoryManager(std::unique_ptr<MCJITMemoryManager>(memman))
           .setErrorStr(&ErrStr)
           .create();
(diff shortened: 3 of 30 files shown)


c_tests/tests/noopts.c
```diff
@@ -0,0 +1,50 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <yk_testing.h>
+
+int add_them(int x, int y) { return x + y; }
+
+int main(int argc, char **argv) {
+  int x, y, res;
+  void *tt = __yktrace_start_tracing(HW_TRACING, &res, &x, &y);
+  // The __yktrace_start_tracing() call has almost certainly already blocked
```
ltratt, line 26: This paragraph isn't necessarily true *if* `__yktrace_start_tracing` can be inlined. I'm not sure the comment is very helpful TBH.

ykcapi/yk_testing.h
```diff
@@ -20,3 +20,10 @@ void __yktrace_irtrace_get(void *trace, size_t idx, char **res_func,
                            size_t *res_bb);
 void *__yktrace_irtrace_compile(void *trace);
 void __yktrace_drop_irtrace(void *trace);
+
+// Force sequentialisation of instructions surrounding a computation.
```
ltratt, line 24: This is hard to parse, and it's not correct, because we can't stop *all* computations being moved around: we can just stop *some* computations being moved around (see e.g. http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4455.html). Rather it's preventing the compiler optimising away this variable at this point *and* putting a sort-of barrier that prevents computation on this variable being moved around.

[Also we can't "borrow" code without copyright. But I don't think we're borrowing from Google here. We're just implementing the same well-known technique -- one can find this technique used in many places. FWIW, I think the SO post is a better explanation than the header file.]

c_tests/tests/noopts.c
```diff
@@ -0,0 +1,50 @@
...
+  // optimisation of x and y, as pointers to them are passed as arguments and
+  // the compiler must assume that it may mutate them. Therefore we initialise
+  // x and y after the call.
+  x = 2;
+  NOOPT(x);
+  y = 3;
+  NOOPT(y);
+  // An optimising compiler would inline this call and (without our NOOPTs)
+  // would then use constant prorogation to replace the whole call with the
```
ltratt, line 35: I wasn't expecting parliament to make its effects felt here :) s/prorogation/propogation/

c_tests/tests/noopts.c
```diff
@@ -0,0 +1,50 @@
...
+  // value 5. We check that this doesn't happen.
+  res = add_them(x, y);
```
ltratt, line 37: I am not convinced that `add_them` can't be inlined: I am convinced that the `asm` trick prevents this turning into a constant however.

c_tests/tests/noopts.c
vext01: `__yktrace_start_tracing()` is in an external shared object, so it can't be inlined.

The reason I included this comment is because the first version of the test initialised `x` and `y` outside of the traced section of code and the compiler was already unable (without `NOOPT`) to replace the call with a constant due to `__yktrace_start_tracing`.

ltratt: OK. I think I'd just remove the comment to be honest.

ykcapi/yk_testing.h
vext01: > preventing the compiler optimising away this variable at this point.

That doesn't capture what's going on here either.

How about "Tries to trick the compiler into thinking that the asm block may mutate the value, thus blocking certain optimisations". This is my understanding.

> one can find this technique used in many places

The problem is that, although it is an old trick, there are lots of different variants of it.

There's the one from the stack overflow post:
```
asm volatile("" : "+m"(const_cast<T &>(value)));
```

which is not used any more it seems.

Then variations in the current google benchmark tree are as follows:

 * The const variant:  `asm volatile("" : : "r,m"(value) : "memory");`
 * The clang non-const variant: `asm volatile("" : "+r,m"(value) : : "memory");`
 * The non-clang non-const variant: `asm volatile("" : "+m,r"(value) : : "memory");`

And then I've seen other variants too like
 * `asm("":::"memory");` -- the one you see a lot in C circles.
 * Rust's black box: `llvm_asm!("" : : "r"(&mut dummy) : "memory" : "volatile");` (equiv to `asm volatile("": "r"(value), "memory");` I think) which is yet another variant.

So actually I think citing the source of our block is quite important here.

c_tests/tests/noopts.c
vext01: I don't often push back, but I think this is a critical comment.

Someone may be tempted to move the initialisation up outside of the traced section, at which point we no longer test `NOOPT`. At that point we'd be testing if `__yk_start_tracing()` blocks optimisations, which is not the intention of the test.

ltratt: Surely all I have to do is something like:

```c
int x = 1;
start_tracing(&x);
x = 2;
NOOPT(x);
...
```

? In other words, I don't think *initialisation* is very special per se.

ykcapi/yk_testing.h
ltratt: > How about "Tries to trick the compiler into thinking that the asm block may mutate the value, thus blocking certain optimisations".

This is closer, but note that we're only preventing optimisations *on that variable* (or an expression if you pass in an expression). In other words, this has sort-of local effects on a variable, not sort-of global effects on every part of the function.

> So actually I think citing the source of our block is quite important here.

Dumb question: why are we using this rather than one of the others? Is it a bit of a random guess?

c_tests/tests/noopts.c
vext01: OK, you got me.

s/initialisation/assignment/

We can't expect the compiler to constant fold an *assignment* that happened before the call to `start_tracing()`, as it doesn't know if it was mutated.

What I don't want is for someone to change the test to the (shorter):
```
int x = 1, y = 2;
NOOPT(X);
NOOPT(Y);
start_tracing()
...
```

As here, whether or not `NOOPT` blocks optimisations is moot, as we know that (currently) `start_tracing` already blocks optimisations. Thus we wouldn't be testing what we thought we were. `NOOPT` could be `#define NOOPT(x) x` and the test would still pass.

Does that make sense?

ykcapi/yk_testing.h
vext01: I chose the google non-const clang variant, because:
 - we use clang
 - the value may not be constant.
 - the google implementations are likely to be correct (?)

What do you reckon?

c_tests/tests/noopts.c
vext01: Just you wait until you make a typo ;)

c_tests/tests/noopts.c
vext01: You are right, inlining is irrelevant.

c_tests/tests/noopts.c
ltratt: I now realise that (once again!) there are two levels of IR we're thinking about here: `NOOPT` is meant to inhibit optimisations in both the AOT and JIT IRs. [I must get used to thinking at both levels!] I think our test is meant to test only that it does so in the JIT IR? If that is the case (and it's probably worth being clear about that in a comment), then it's obvious that the "IR part of the test" can only happen between stop/start tracing.

vext01: I think here we'd be checking that it doesn't happen in the AOTMod, since `YKD_PRINT_IR` prints the `IR` before it's submitted to the execution engine.

TBH, I don't care where we test if `NOOPT` works, I just want to know that it does work!

ltratt: In our case I think we do want to know that it works at both IR levels! It would be confusing if it worked in the AOT IR but not the JIT IR. It wouldn't hurt to test both IRs, but at the moment we only have the JIT IR debugging as output -- so we might as well test that. In fairness, if it works in the JIT IR, it almost certainly works in the AOT IR!

vext01: OK. Testing the post-opt JIT module would require a larger change (and an additional test). 

Let's put this on hold until @ptersilie can figure out the blockmap bug.

ykcapi/yk_testing.h
ltratt: So the "real" reference is actually the talk at https://youtu.be/nXaxk27zwlk?t=2441 -- it's worth watching! He defines `escape` and `clobber` which correspond to `DoNotOptimize` (which due to C++ overloading comes in several flavours) and `ClobberMemory` in the "Google code". Note that they have both:

```c++
inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value)
```

and:

```c++
inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp& value) {
```

I am not sure why they need to overload `const&`, but @vext01, your better knowledge of x86 asm might help here?

ltratt:
One minor thing that occurred to me: I think we should guard this with something like `#if defined(LLVM)` just to emphasise that we've only tested this with LLVM. Just in case someone comes along one day, ports all the other LLVM stuff to (say) the MS compiler and then this bit silently doesn't quite do the right thing.

ykcapi/yk_testing.h
vext01: In light of new info -- that in a commit to Google benchmark, the definition shown in the talk was deemed buggy -- what shall we do with this?

I still think I'd like to cite Google. They've clearly though hard about this.

vext01:
> just to emphasise that we've only tested this with LLVM.

Is it necessary? The whole approach is underpinned by LLVM IR :P

ykcapi/yk_testing.h
ltratt: > I still think I'd like to cite Google. They've clearly though hard about this.

OK.

Do we have an answer for the `const&` thing?

ltratt:
I think this is more general than just this repo, and I wouldn't be surprised if at some future point someone takes this stuff out and tries it in other compilers. Guarding it for LLVM costs us little, but will help our future selves.

vext01:
OK, but it's the clients of this repo that need to be built with ykllvm, so such guards would need to go in them, not us.

In theory you could build yk with gcc and it would still work so long as the client use clang.

ltratt:
The compile-time guards need to go around our `NOOPT` stuff.

vext01:
I'm adding the GCC variant for the reason above :)

ltratt:
Let's not add what we've not tested.

ykcapi/yk_testing.h
vext01: > In other words, this has sort-of local effects on a variable, not sort-of global effects on every part of the function.

Notice that the whole memory space is still clobbered.

So in fact, the `clobber` operation is weaker than the `escape` operation. I didn't expect that.

So how about:

> Blocks the compiler from optimising the specified value or expression.

vext01:
89c6d67bd9 addresses comments and allows us to test all of the various IR stages in the pipeline. It looks worse than it is because I had to move some functions around.

I've changed our environment variables so that you can do things like:
```
YKD_PRINT_IR=aot,jit-post-opt
```

Which would print the IR at those two stages in the pipeline.

Each IR print has a header which can then be anchored in the lang test to check that specific IR.

Only the NOOPT test has been updated to use this new facility. If you like what you see, I will update the others.

vext01:
@ptersilie you might like to take a look too.

c_tests/tests/noopts.c
```diff
@@ -18,22 +40,24 @@
 #include <string.h>
 #include <yk_testing.h>
 
-int add_them(int x, int y) { return x + y; }
+int add_them(int x, int y) {
+  // At the call-site x and y are constant, so an optimising compiler would try
```
ltratt, line 44: Putting the comment here feels weird given that this isn't where the noopt effect happens.

ykcapi/yk_testing.h
ltratt: I'm not going to let go of the `const` question so let's start with that :)

ptersilie:
Looks good. Do we know why the `ExecutionEngine` doesn't optimise even when given an opt level via the `EngineBuilder`?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -62,6 +64,146 @@ thread_local bool ThreadAOTModInitialized = false;
...
+// Dumps an LLVM Value to a string and trims leading whitespace.
+void dumpValueToString(Value *V, string &S) {
+  raw_string_ostream RSO(S);
+  V->print(RSO);
+  lTrim(S, ' ');
+}
+#endif
+
+enum DebugIR {
+  AOT,
```
ltratt, line 82: Is it guaranteed that these are given the numbers 0,1,2,3?

vext01: I believe so:

>  If the first enumerator does not have an initializer, the associated value is zero. For any other enumerator whose definition does not have an initializer, the associated value is the value of the previous enumerator plus one. 

https://en.cppreference.com/w/cpp/language/enum

vext01:
>  Do we know why the ExecutionEngine doesn't optimise even when given an opt level via the EngineBuilder?

Afraid not. I could dig further, but I'm starting to grow tired of this PR :P

c_tests/tests/noopts.c
vext01: But it is where the optimisation is blocked. Removal of the plus operation is blocked.

ykllvmwrap/src/ykllvmwrap.cc
ltratt: Perfect!

c_tests/tests/noopts.c
ltratt: Sorry I wasn't very clear. I don't think this function needs to be broken out. I think it's equivalent to something like:

```
x = NOOPT(...);
y = NOOPT(...);
z = x + y;
```

i.e. breaking the function out isn't necessary (and is confusing).

ykcapi/yk_testing.h
vext01: Comparing the const and non-const version:
```
    const: asm volatile("" :               : "r,m"(value) : "memory");
non-const: asm volatile("" : "+r,m"(value) :              : "memory");
```

And the syntax is:


Where the meaning of the input and output strings is:
>      OutputOperands:  A comma-separated list of the C variables modified by the instructions in the AssemblerTemplate. An empty list is permitted. See OutputOperands.
InputOperands
>
>    A comma-separated list of C expressions read by the instructions in the AssemblerTemplate. An empty list is permitted. See InputOperands.

And the `+` (allowed only for output operands) means:
> Output constraints must begin with either ‘=’ (a variable overwriting an existing value) or ‘+’ (when reading and writing).

So putting this all together, the const version asks the compiler to put put the constant value in either memory or a register and informs that the block **reads** the constant.

The non-const version asks the compiler to do the same, but tells the compiler that the block **both reads and writes** to the variable.

My guess is that it's like this because you can't write to a constant.

vext01: This begs the question of whether we need a constant version.

I get the impression that this isn't an exact science, and that these funky asm blocks are just maximising the chances of hosing optimisation (with no absolute guarantee).

Note that the non-const version when applied to a constant in our test, **did* block optimisation.

I want to move on. What would you like to do?

ltratt: I'm happy with moving on and a) not having the const variant b) but documenting that it could exist (so that we know if/when we need it).

vext01: OK. 3e476be30 addresses that.

vext01:
So @ltratt, I take it the new IR printing idea was OK? If so I will now update the existing tests.

ltratt:
> I take it the new IR printing idea was OK?

Works for me!

vext01:
@ltratt see those last commits.

The second one is a bit frustrating. See the commit message. I'd be interested in your thoughts.

I'm also now questioning the correctness of `NOOPT`, as I just realised that both Carruth and Google always work on a pointer (or a reference) to the data they want to block optimisations on. We are working directly on the values.

So whilst our current `NOOPT` sems to work, it not precisely what Google do in their benchmark repo.

I'm going to try and fix that but it requires more fiddling because:

 a) It can't be done with hygene in a C macro.
 b) our tracer bombs out if it encounters the obvious alternative: a `static inline` function.

Bear with me...

ltratt:
You're right, according to the video at least, you have to do things like:

```c
x = ...;
NOOPT(&x);
```

in order to keep `x` on the stack.

ltratt:
I'm not sure what the problem in the second commit is?

vext01:
> you have to do things like NOOPT(&x)

Indeed, but inline asm won't take `&` to constraints, so we'd have to use an intermediate variable, and you can't do that hygienically. I'm trying to make it a function, but the trace compiler is now choking on something.

> I'm not sure what the problem in the second commit is?

Instead of a lengthy discussion, let me ask: are you OK using `black_box()` to block optimisations in rust code?

ltratt:
> Indeed, but inline asm won't take & to constraints, so we'd have to use an intermediate variable, and you can't do that hygienically with a macro.

To be honest, I don't understand what this means!

> are you OK using black_box() to block optimisations in rust code?

Yes, I assumed we would use `black_box` rather than trying to invent our own thing that does the same!

vext01:
> Yes, I assumed we would use black_box

OK, so there is no problem there at least. I was trying to translate the C definition to Rust, which it turns out is quite hard.

vext01:
So now, silly question: why not expose `black_box()` to C via the yk testing API?

ltratt:
>  So now, silly question: why not expose black_box() to C via the yk testing API?

Good idea! Why not tyr?

vext01:
@ltratt @lukasd @jhughes The plot thickens yet further. This is going to be a bit of a write-up I'm afraid. And it's full of nitty-gritty detail. Brace yourself and I'll try my best to explain what I think is going on.

Consider we trace this:
```
  int res;
  void *tt = __yktrace_start_tracing(HW_TRACING, &res);
  res = 0;
  NOOPT(&res);
  f(&res);
  void *tr = __yktrace_stop_tracing(tt);
```

And that `NOOPT` is now a function accepting a pointer, as discussed offline (clang checks elided for clarity):
```
static inline void NOOPT(void *x) {
  asm volatile("" : "+r,m"(x) : : "memory");
}
```

The AOT IR then looks like this:
```
%5 = bitcast i32* %3 to i8*, !dbg !44
...
%6 = call i8* (i64, ...) @__yktrace_start_tracing(i64 1, i32* nonnull %3) #7, !dbg !45
...
call fastcc void @NOOPT(i8* nonnull %5), !dbg !46
...
%7 = call i8* @__yktrace_stop_tracing(i8* %6) #7, !dbg !50
...
```

This is problematic, as the generation of the pointer to `res` has been moved outside of the traced section, despite the use of `NOOPT`. This means that we end up with an undefined variable (`%5`) in our pre-opt JIT code. We need the bitcast operation to stay in side the traced section.

Why can this happen? Due to `NOOPT(&res)`, the compiler must assume that the pointer itself (`&res`) may be mutated **and** that the pointee (`*&res`) too may be mutated. But note that:

 - Mutations to the pointer itself would be local to the frame of `NOOPT()`, as the pointer is passed by value, and thus copied in.
 - Mutating the pointee does not alter its address on the stack at the call-site of `NOOPT`.

This gives the compiler license to move the computation of `&res` up and out of the traced section, as the definition of `NOOPT` can't possibly alter the address of `res` on the stack at the call-site of NOOPT.

In other words, we have blocked optimisations to `res`, but not to `&res`. By introducing `&res` we introduce a second value that we'd need to block optimisations on. But you can't just do `NOOPT(&&res)` as then you'd just make a third thing (`&&res`) that you'd then need to block optimisations on, and so on.

[This also shows that the call to black_box() in start_tracing() is not sufficient to block that particular optimisation either]

What is interesting though, is that if we use our old definition of `NOOPT`:

```
#define NOOPT(X) asm volatile("" : "+r,m"(X) : : "memory");
```

And pass the `NOOPT()` arguments by value, not pointer (and not that there is no actual call, as it's a macro function -- the asm block is just inlined), then the problem goes away. It goes away because there is no longer any "address of" operation, and thus no bitcast to "move up".

The question remains though, does doing `NOOPT(x)` -- i.e. not `NOOPT(&x)` -- do what we intend.

What I can say with full confidence is:

 - a) passing a value to the macro version of NOOPT macro does not do quite the same as Google's implementation of `DoNotOptimize`. I disassembled a usage ofDoNotOptimise to check that a pointer is passed at the binary level. It is. That means that they too would block optimisations on `res`, but not `&res`.

 - b) passing a value `x` to the macro version of NOOPT does block optimisations on `x` (today on a Linux system).

Once you've understood a), you realise that our requirements are *stronger* than Google's, and that's why they can get away with passing a pointer.
They require only that the **pointee** isn't optimised; we require than neither the **pointee** nor the **pointer** be optimised. We can't block optimisations to the **pointer**, but we can work around the problem by ensuring that no pointer exists to call NOOPT.

What's actionable then? Given b), I suggest we use the current macro version of NOOPT and let tests guide us. If at any point our NOOPT test fails, we that we got it wrong, and that we have to try again.

 I hope this makes sense. It's been very hard to explain.

vext01:
[I realise that the example isn't the best -- will try to update]

vext01:
> [I realise that the example isn't the best -- will try to update]

Updated, please refresh.

ptersilie:
> Lukasd

You pinged some random dude here. ;-)

vext01:
> We can't block optimisations to the pointer, but we can work around the problem by ensuring that no pointer exists to call NOOPT.

I now realise that this is a more general problem that extends beyond the scope of `NOOPT`.

If the user wants to trace a call to any function that accepts a pointer argument, and that call is in the top-level frame directly between start_tracing() and stop_tracing(), and LLVM deems that a bitcast is required, then will hit this same problem: the bitcast can be hoisted out of the traced section.

vext01:
> You pinged some random dude here. ;-)

Oops, sorry!

ptersilie:
We can't even easily unhoist the bitcast back inbetween start/stoptracing, because the arguments we pass to `start_tracing` still reference the pre-bitcast value.

However! We've always planned to automate the passing values into the trace part, i.e. we automatically figure out which values are needed inside the trace, but were defined outside of it. In that case we would then figure out, that `%7` is actually needed, and not `%3` and automatically pass in `%7`. Wouldn't that solve this issue?

jacob-hughes:
> > Lukasd
> 
> You pinged some random dude here. ;-)

Same here. Nothing to see here Jon ;)

vext01:
> Wouldn't that solve this issue?

Lukas is right. This would solve the issue.

OK, so here's the facts:
 - Our NOOPT macro does block the optmisations that we want, even if it's different to Google's implementation.
 - Google's implementation would require a function that accepts a pointer.
 - In general we can't inline functions that accept pointers without implementing the auto-trace-inputs Lukas mentions.
 - auto-trace-inputs is no small undertaking: it likely requires all manner of low-level juggling including stack maps and a fair chunk of inline asm.

Let's implement an imperfect solution using our current `NOOPT` and adjust it later when we have auto-trace inputs.

I'll let this settle, and update the PR on Monday if all is still well.

ltratt:
I'm fine with moving on, with one suggestion: if our `NOOPT` works on values (not pointers) let's call it `NOOPT_VAL` (or similar) so that if we want a pointer version later we don't have to disentangle it.

vext01:
> NOOPT_VAL

Good idea.

vext01:
3005599f2 does the rename and some other tidy-ups. A separate commit updates the gitignore file.

OK to squash?

c_tests/tests/noopts.c
```diff
@@ -7,67 +7,52 @@
...
-  int x, y, res;
-  void *tt = __yktrace_start_tracing(HW_TRACING, &res, &x, &y);
-  // The __yktrace_start_tracing() call has almost certainly already blocked
-  // optimisation of x and y, as pointers to them are passed as arguments and
-  // the compiler must assume that it may mutate them. Therefore to properly
-  // test NOOPT we need initialise x and y after the call.
+  int x, res;
+  void *tt = __yktrace_start_tracing(HW_TRACING, &res, &x);
+  // __yktrace_start_tracing() will already inhibit optimisations on x since
+  // it's address is passed as an argument and the compiler must assume that it
```
ltratt, line 41: s/it's/its/

c_tests/tests/noopts.c
```diff
@@ -7,67 +7,52 @@
...
+  // may be mutated. Therefore to properly test NOOPT_VAL we need initialise x
```
ltratt, line 42: s/we need initialise/we need to initialise/

[I also think that backticks around the variable names would make this easier to read!]

c_tests/tests/noopts.c
```diff
@@ -7,67 +7,52 @@
...
+  // and y after the call to __yktrace_start_tracing..
   x = 2;
-  NOOPT(&x);
-  y = 3;
-  NOOPT(&y);
-  res = add_them(x, y);
+  NOOPT_VAL(x);
+  res = x + 3; // We don't want this operation to be optimised away.
```
ltratt, line 46: I claim that `NOOPT_VAL(res)` is not necessary in this example because in `res = x + 3`, `x` must be opaque.

ykcapi/yk_testing.h
```diff
@@ -23,16 +23,16 @@ void __yktrace_drop_irtrace(void *trace);
...
+// This is similar to the non-const variant borrowed from Google benchmark:
 // https://github.com/google/benchmark/blob/e451e50e9b8af453f076dec10bd6890847f1624e/include/benchmark/benchmark.h#L350
 //
-// Note that Google benchmark also defines a variant of this for constant data.
-// At the time of writing, the below NOOPT seems to suffice (even for
-// constants), but we may need to borrow the const version later.
-static inline void NOOPT(void *x) {
+// Our version works on a value, rather than a pointer.
+//
+// Note that Google benchmark also defines a variant for constant data. At the
```
ltratt, line 31: s/note that Google/note that the Google/

ykcapi/yk_testing.h
```diff
@@ -23,16 +23,16 @@ void __yktrace_drop_irtrace(void *trace);
...
+// time of writing, the below NOOPT_VAL seems to suffice (even for constants),
```
ltratt, line 32: s/the below//

ykcapi/yk_testing.h
```diff
@@ -23,16 +23,16 @@ void __yktrace_drop_irtrace(void *trace);
...
+// but we may need to borrow the const version later.
```
ltratt, line 33: s/borrow/consider using/

c_tests/tests/noopts.c
vext01: It's just convention to guard the inputs and the outputs of the computation you want to shield from optimisations:

> The key to these techniques is to focus on the data. You make the input to the computation opaque to the optimizer and the result of the computation opaque to the optimizer.

https://stackoverflow.com/questions/11748362/how-to-disable-an-optimization-in-llvm

I don't mind either way, it's not strictly required for this particular test on Linux today.

ykcapi/yk_testing.h
vext01: For better or worse, "Google benchmark" is the name of the project.

https://github.com/google/benchmark

c_tests/tests/noopts.c
ltratt: I'm not sure that SO link is the right one?

vext01: Oops! https://stackoverflow.com/questions/37786547/enforcing-statement-order-in-c/38025837#38025837

vext01:
That last commit should address all of the comments?

ykcapi/yk_testing.h
ltratt: Then it should presumably be "Google Benchmark" (capital 'b')?

vext01: 72060fb

c_tests/tests/noopts.c
ltratt: You're right and I'm wrong: it's not that we need the second `NOOPT_VAL` *as such* but putting it before `stop_tracing` ensures that `res` won't float past `stop_tracing`.

[FWIW, `res` seems a less than ideal variable name in many of these tests. It would probably be better just to use obviously meaningless names like `y`. Maybe we can think about that in a future PR?]

ltratt:
Please squash.

vext01:
splat.

PR #380, opened 2021-07-20
Handle non-const (mutable) global variables.

ptersilie:


c_tests/run.rs
```diff
@@ -37,6 +37,7 @@ fn mk_compiler(exe: &Path, src: &Path, opt: &str) -> Command {
         "-flto",
         "-Wl,--mllvm=--embed-bitcode-final",
         "-Wl,--lto-basic-block-sections=labels",
+        "-Wl,--export-dynamic",
```
ltratt, line 40: Hang on, are we compiling *every* test as `export-dynamic`?

ltratt: [I think we should probably only compile certain tests -- presumably a test specifically designed to test this feature -- with this flag. We could probably use lang_testers `env-var` thing to set `LDFLAGS` for such a test?]

ptersilie: Yes we are. I can hide it behind an env-var, but as @vext01 pointed out, we'll likely have to run everything with `export-dynamic` until we can find trace inputs at runtime.

vext01: It's not even clear to me that using trace inputs is beneficial to this solution. It seems to me like we trade off binary size for performance when assembling the JIT module.

vext01: Whichever solution we choose, it should be globally enabled, since any arbitrary program that we may wish to trace can access an otherwise unexported global variable.

ltratt: I must admit that I don't see what trace inputs have to do with global symbols.

vext01: It's about finding the global variable at runtime.

If there is an exported symbol in the binary, then we don't have to do anything special, as the loader "just finds" it via standard techniques. This does mean exporting symbols that wouldn't otherwise be exposed (hence --export-dynamic). This has the advantage of being a static technique with no runtime overhead, but it will bloat the binary a little (symbols are strings).

If you don't want to export symbols, then you have to tell LLVM how to find the address of the gobal another way. One solution would be to pass a pointer to the global as a trace input. Then reading/writing from/to it is a matter of using a LLVM `load`/`store`. But this would require something like stackmaps in order to be able to find where the global. This technique avoids bloating the binary, but has a runtime cost at the time of invoking JITted trace code: you'd have to dynamically locate the global variables, probably via stackmaps (but the result could be cached).

vext01:
Seems I need this change for a test I've just written.

At `-O2` and above, a call to `putchar()` will give:
```
call_ext: Non-const global variable stdout
```

ptersilie:
Sounds like it. Let's get this merged then. But I'm not sure what we are going to do about the `export symbols` bit.

c_tests/run.rs
ltratt: > It's about finding the global variable at runtime.

I understand that: I still don't understand why it relates to trace inputs.

ptersilie: What @vext01 is saying is: We currently have a problem accessing global variables inside a trace. The current workaround is using `export-dynamic` so the linker can find all globals (even those that are not usually exported). In the future a better solution might be to pass the globals to the trace via trace inputs.

ltratt: > We currently have a problem accessing global variables inside a trace.

Just to check: we're not sure why we have such a problem?

vext01: We know why we have the problem: Programs read and write global variables.

Therefore we need a way to discover the addresses of global variables.

Exporting all symbols is one such way to achieve this as it gives the binary a `name -> offset` mapping in the form of a symbol table.

ltratt:
OK, so after talking to @vext01 offline, really this PR is *entirely* about using "export symbols" to enable us to find global variables by name. This is a bit icky, but it's certainly acceptable for the time being. However, I think we need to:

  1. Document somewhere that you *have* to use `export-symbols` in order to use yk.
  2. Document in the test runner both *that* we've turned that on and *why* we've turned it on.

Shall we push a commit that does at least the second of those, and see how we feel?

vext01:
I agree with Laurie's actionables.

> find global variables by name. This is a bit icky

Although icky, there is precedent. When you run a normal binary (e.g. `/bin/ls`) the loader resolves all symbols by name. This is why name mangling is needed for languages like C++.

ltratt:
I'm not saying that finding global variables by name is icky: I'm saying using `export-symbols` is icky.

ltratt:
Two additional suggestions: we should add "FIXME" to the test runner; and raise an issue.

ptersilie:
Try this. Could probably do with another iteration.

c_tests/run.rs
```diff
@@ -37,6 +37,10 @@ fn mk_compiler(exe: &Path, src: &Path, opt: &str) -> Command {
         "-flto",
         "-Wl,--mllvm=--embed-bitcode-final",
         "-Wl,--lto-basic-block-sections=labels",
+        // A trace may reference globals that weren't exported by the linker. This can make it
+        // impossible to find the global's address at runtime and update the JITModule's mapping.
```
ltratt, line 41: s/and update the JITModule's mapping//

c_tests/run.rs
```diff
@@ -37,6 +37,10 @@ fn mk_compiler(exe: &Path, src: &Path, opt: &str) -> Command {
...
+        // Using `--export-dynamic` enforces that all globals are exported, allowing the compiled
```
ltratt, line 42: s/enforces that all globals are/forces all globals to be/

Also we should probably use the "AOT IR" / "JIT IR" terminology to avoid confusing our future selves :)

ptersilie: Fixed in 1684bec

ltratt:
Please squash.

ptersilie:
Squashed.

ltratt:
@ptersilie I think we should add something to the commit description that explains *why* the code looks as it does. At least something like "export-dynamic is a bit of a hack but we don't know how to do better" perhaps?

ltratt:
I forgot to say: please force push an update.

ptersilie:
Pushed. It's basically the description of the issue, but I believe that sums is up.

ltratt:
Perfect!

bors r+

PR #384, opened 2021-07-22
More ykcbf fixes

vext01:
This is a mixed bag of things that came up when getting ykcbf to JIT.

See individual commits.


vext01:
Gah, github hasn't added Lukas as a co-author of the first commit. I guess `Co-authored-by:` has to come last.

I'll fix this at squash time.

yktrace/src/hwt/mapper.rs
```diff
@@ -127,8 +128,25 @@ impl HWTMapper {
                     }
                 }
             } else {
-                ret_irblocks.extend(irblocks);
+                for irblock in irblocks.into_iter() {
+                    // A basic block may map to >1 *machine* basic block. If we see the same block
+                    // repeated, then we need to check if we re-executing the same basic block or
+                    // if we are executing next machine basic block for that same block. In the
+                    // latter case, we mustn't record the block in the trace again.
+                    let mut append = true;
```
ltratt, line 136: The logic in lines 136-147 is a bit hard to follow. I *think* it's equivalent to something like :

```
if ret_irblocks.last() == Some(irblock) && prev_block.map(|x| x.first_instr()) != Some(block.first_instr()) {
    ret_irblocks.push(irblock);
}
```

c_tests/tests/switch.c
```diff
@@ -0,0 +1,41 @@
...
+    break;
+  case 2:
+    res = 12;
+    break;
+  case 3:
+    res = 4;
+    break;
+  default:
+    res += 1;
+  }
```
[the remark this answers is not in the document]
vext01: I thought that we were not going to guard the resulting values? Only the inputs?

https://github.com/ykjit/yk/pull/376#discussion_r672232482

(To be honest, I'm more comfortable with guarding both inputs and outputs)

yktrace/src/hwt/mapper.rs
```diff
@@ -183,7 +183,12 @@ impl HWTMapper {
     fn map_block(&mut self, block: &hwtracer::Block) -> Vec<IRBlock> {
         let block_vaddr = block.first_instr();
         let (obj_name, block_off) = code_vaddr_to_off(block_vaddr as usize).unwrap();
-        let block_len = block.last_instr() - block_vaddr;
+
+        // The `+ 1` here is because hwtracer gives us the address of the last instruction in the
```
ltratt, line 187: What happens if the last instruction is more than 1 byte long?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -115,11 +115,16 @@ class JITModBuilder {
...
     CallInst *noinline_func = nullptr;
 
+    bool ExpectUnmappable = false;
+
     // Iterate over the trace and stitch together all traced blocks.
     for (size_t Idx = 0; Idx < Len; Idx++) {
       auto FuncName = FuncNames[Idx];
 
-      // FIXME: Deal with holes in the trace, e.g. calls to libc.
+      if ((ExpectUnmappable) && (FuncName == nullptr)) {
```
ltratt, line 124: I think the brackets around `ExpectUnmappable` are pointless?

yktrace/src/hwt/mapper.rs
vext01:
```
if ret_irblocks.last() == Some(irblock)
```

`if let`?

You can't combine `if let` with `&&` hence this logic being a bit messy.

I'll have a go at simplifying it a bit later.

ltratt:
In the commit message for https://github.com/ykjit/yk/pull/384/commits/b5b030750093ddb71b531532113c1fcf8a27dda4, could we define what an "unmappable hole" is? I think I know, but I'm not sure.

yktrace/src/hwt/mapper.rs
ltratt: I deliberately didn't use `if let` because `if let` doesn't compose syntactically.

c_tests/tests/switch.c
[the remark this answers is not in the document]
ltratt: You're just making me repeat what I said just below that comment aren't you?

> You're right and I'm wrong

This should mean that I lose face, but since I'm wrong so often, repeating one of my wrongnesses one more time won't hurt me :p

yktrace/src/hwt/mapper.rs
vext01: It doesn't matter. As long as at least one byte of an instruction is captured, the correct thing will happen.

This ickyness is all born out of a quirk in libipt. It can only report the address of the last instruction in a PT block and not the last byte of the block.

c_tests/tests/switch.c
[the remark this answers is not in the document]
vext01: Ah, actually I screwed up. I didn't see your "You're right and I'm wrong" comment and *removed* the second guard from the `NOOPT_VAL` test! So I should go back and correct that.

😆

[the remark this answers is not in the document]
ltratt: Oops, yep!

yktrace/src/hwt/mapper.rs
ltratt: I think this is crucial "As long as at least one byte of an instruction is captured, the correct thing will happen." I think the comment at the moment is a bit wooly, but that sentence captures something important. One thing I'm not sure about is what "the correct thing" is, but maybe that's me being an idiot. Can we rephrase the comment capture the crucial bit?

vext01: Yes we can. Let me try to explain it here first, as I think what I said above isn't quite correct. Hopefully then we can boil the comment down to something concise and accurate.

It's to do with the way we store the [block map](https://llvm.org/docs/Extensions.html#sht-llvm-bb-addr-map-section-basic-block-address-map). We use an interval tree which stores a collection of machine basic block ranges. When we get a PT block, we want to know which machine block(s) that block maps to. An interval tree is perfect for this, as you can query it with a range and any and all overlapping ranges are returned.

Let's run some examples. Suppose we assign each instruction in a block with a letter, and that each instance of the letter represents a byte of that instruction. Let's use a pipe to delimit machine blocks.

Suppose that the 4 machine blocks of a function are:
```
abbcc|ddee|f|gghh
0     5    9 10
```

[the numbers underneath are the start address of the machine block]

If libipt gives us a `(block-addr, last-instr-addr)` pair. So if we got (0, 7) from libipt, then we expect to get the first two blocks. The code in-tree would get this right even though it would query that range with an exclusive upper bound.

Where the in-tree code breaks down is when the block is one-byte long. So if we got (9, 9), then we'd query the range `9..9` which is an empty range. By definition, querying an interval tree with an empty range always yields the empty set, which means we erroneously fail to map the block.

What the `+1` does is it ensures that the upper bound is exclusive. Then the 1-byte block is correctly mapped. We could have said "if the block length is 1, then do a query with an exclusive upper bound", but there's no need. Since every instruction is at least one byte long, adding one will not break the cases where the length of the block is longer than a single byte. Therefore adding one is always correct (you can never spill into the next block and report one too many blocks).

Therefore:
> As long as at least one byte of an instruction is captured, the correct thing will happen.

Is wrong. It should read:
> As long as at least one byte of the machine block is captured, the correct thing will happen.

In other words, in order for the correct blocks to appear in the trace, we must query the interval tree with a range that overlaps with at least one byte of each corresponding machine basic block.

Does that make sense? @ptersilie how about you?

ltratt: OK, I think I understand. How about something like "The HW mapper gives us the start address of the last instruction in a block, which works OK except for blocks with a single 1 byte instruction for which we generate an interval `(x,x)` i.e. an empty interval. Fortunately we don't need to calculate the precise end address of the last instruction in a block (on variable width instruction sets that's impractical!): instead we just need to ensure we're at least 1 byte into that last instruction (hence the `+1` below). Since every instruction is at least 1 byte long, this is always safe, and ensures that we always generate a valid interval."

vext01: Almost.

> we just need to ensure we're at least 1 byte into that last instruction

We just need to make sure we are one byte into the last machine block.

Do you agree?

ltratt: > We just need to make sure we are one byte into the last machine block.

I don't think "last machine block" makes sense (last machine block relative to ... ?). I *think* "last instruction" is correct (at least if I've understood what the HW mapper and interval tree do).

vext01: I think you are right actually.

ltratt: OK. Shall we use my suggestion above (but with perhaps "s/single 1 byte instruction/that consist of 1 instruction and where that instruction is 1 byte long/")?

vext01: Reading this again, I think this could be more objective:

How about:
> We need to check which LLVM machine block(s) the hardware block coincides with (whose intervals overlap). Ideally we'd query the entire range of the hardware block, but libipt can only tell us the range up until the beginning of the last CPU instruction in the block. This is actually fine, as for a machine block to match, at least one-byte must overlap with the queried range. Therefore we can safely query the range libipt gives us (just ensuring that the upper bound is inclusive so that 1-byte blocks are captured).

ltratt: I again fear that I am once again slipping into arrogance, but this seems both less precise and less easy to read than my suggestion?

vext01: Let me try to win you over to my way of thinking:

Our query interval must overlap each of the corresponding machine blocks by at least one byte. This is the weakest sufficient condition for correctly mapping blocks.

Suppose that we have a query that should map to the first 3 blocks of the following blockmap layout:

```
blockmap: |-----|---|-|----|
expected: |-----|---|-|       This is the correct mapping result.
query1  : |-----------|       OK and exact.
query2  :     |-------|       OK, overlaps all expected blocks too.
query3  :       |-----|       Bad, no byte in the query overlaps with the first block.  
query4  : |---------|         Bad, no byte in the query overlaps with the third block.
```

I find this way of thinking much more intuitive.

Note also that:

> which works OK except for blocks with a single 1 byte instruction for which we generate an interval (x,x) i.e. an empty interval

Isn't quite right. The scenario we are guarding against is when the last block in a query is a 1-byte block. Query 4 serves as a counter-example. The query is not empty, but the 3rd block is missed because the interval bound lacked an exclusive upper bound.

Makes sense?

vext01: All of this to say, we just need to ensure the upper bound is an inclusive bound :P

vext01:
Just one outstanding conversation (the one on the inclusive upper bounds).

yktrace/src/hwt/mapper.rs
ltratt: How about merging the two into something like:

"The HW mapper gives us the start address of the last instruction in a block, which gives us an exclusive upper bound when the interval tree expects an inclusive upper bound. This is a problem when the last block in our sequence has a single 1 byte instruction: the exclusive upper bound will not include any of that block's bytes. Fortunately, we don't need to calculate the precise end address of the last instruction in a block (on variable width instruction sets that's impractical!): instead we just need to ensure we're at least 1 byte into that last instruction (hence the +1 below). Since every instruction (and thus every block) is at least 1 byte long, this is always safe, and ensures that we generate an inclusive upper bound."

vext01: Jolly good. Fixed in 708e8e4d4609.

vext01:
Squash time?

ltratt:
Please squash.

vext01:
splat.

ltratt:
bors

vext01:
oops. hold on.

ltratt:
It's OK, I fat-fingered the bors command anyway!

vext01:
Fixed the co-authoring. Ready.

Issue #385, opened 2021-07-23
Refactor `createModule()`


vext01:
This function is responsible for assembling a trace by inlining the block that the hardware trace executed through.

The function is essentially nested loops:
```
for BB in Blocks {
  for Inst in BB {
    ...
  }
}
```

However, as we've added more and more JIT support, the function has become unmanageable.

The most icky aspect of this function is due to the fact that an LLVM `call` does not terminate a block. This means that when there is an inlined call, we have to stop copying instructions mid-block, start copying instructions from the callee's blocks, and when the callee returns, resume the original block **from where we broke off earlier**.

This has been implemented with a complicated spaghetti of `break`s and `continue`s, that ensure that each loop only runs proper when appropriate.

For example, a block like:
```
main:
  ...
  bb5:
    ...
    call f
    ...
```

Might give a trace like `<..., main:bb5, f:bb0, main:bb5, ...>`. Note that `main:bb5` appears twice because PT *does* consider a call to terminate a block. When building the trace, the second time we see `main:bb5` the loop attempts to `continue` enough times until we reach the instruction after call.

The problem is that this skipping logic is tangled up with the non-skipping logic and it's a real headache.

I suggest that:

 - We don't use nested loops to implement stepping over the trace. Instead we can have one loop and an "instruction pointer" which is a `(func, bb)` pair. The first thing the loop does is look up the block and instruction to inspect next. Then jumping to arbitrary places in a block is easy by modifying the instruction pointer, and we don't need a maze of `break`s and `return`s.

 - We use a separate function for containing the skipping logic so as to disentangle it from the main copying logic.

ltratt:
We should assign this to someone... who?

vext01:
I didn't assign anyone just yet, as now isn't the time for this refactoring.

vext01:
(But I'm happy to work on this when the time comes)

ltratt:
I think you should assign it to yourself then: it makes it less likely that this will be forgotten in the future.

vext01:
#399 and #401 have addressed most of this.

Remaining tasks:

 - Review the fields of the `JITModBuilder` class. Should some of these be local scope in `createModule()`?
 - Group the incoming trace information into structs.

PR #386, opened 2021-07-23
Correctly resume from after external calls.

vext01:
**This is a draft**

Before we would run the block containing the call twice, which is
incorrect. This is the reason ykcbf printed everything twice.

Only a draft for now because I think I can do some minor refactoring to make the code neater. I wanted to share the code in case it helps some of Lukas' Ruby issues.

[We will also need to do some [quite major refactoring](https://github.com/ykjit/yk/issues/385) in the medium-term]

### DIFF
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -69,6 +69,8 @@ public:
     }
   }
 
+  // FIXME: this function needs to be refactored.
+  // https://github.com/ykjit/yk/issues/385
   Module *createModule(char *FuncNames[], size_t BBs[], size_t Len,
                        Module *AOTMod, char *FAddrKeys[], uint64_t FAddrVals[],
                        size_t FAddrLen) {
@@ -116,6 +118,7 @@ public:
     CallInst *noinline_func = nullptr;
 
     bool ExpectUnmappable = false;
+    Instruction *ExtCallResume = nullptr;
 
     // Iterate over the trace and stitch together all traced blocks.
     for (size_t Idx = 0; Idx < Len; Idx++) {
@@ -139,21 +142,33 @@ public:
 
       // Iterate over all instructions within this block and copy them over
       // to our new module.
+      // FIXME: It would be nice to merge the two skipping mechanisms
+      // (ExtCallResume/last_call).
       for (auto I = BB->begin(); I != BB->end(); I++) {
-        // Skip calls to debug intrinsics (e.g. @llvm.dbg.value). We don't
-        // currently handle debug info and these "pseudo-calls" cause our blocks
-        // to be prematurely terminated.
-        if (isa<DbgInfoIntrinsic>(I))
-          continue;
+        // If we've returned from an external call, skip ahead to the
+        // instruction where we left off.
+        if (ExtCallResume != nullptr) {
+          if (&*I != ExtCallResume) {
+            continue;
+          } else {
+            ExtCallResume = nullptr;
+          }
+        }
 
-        // If we've returned from a call, skip ahead to the instruction where
-        // we left off.
+        // If we've returned from an internal call, skip ahead to the
+        // instruction where we left off.
         if (last_call != nullptr) {
           if (&*I == last_call) {
             last_call = nullptr;
           }
           continue;
         }
+
+        // Skip calls to debug intrinsics (e.g. @llvm.dbg.value). We don't
+        // currently handle debug info and these "pseudo-calls" cause our blocks
+        // to be prematurely terminated.
+        if (isa<DbgInfoIntrinsic>(I))
+          continue;
 
         if (isa<CallInst>(I)) {
           CallInst *CI = cast<CallInst>(&*I);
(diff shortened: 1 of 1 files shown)


ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -69,6 +69,8 @@ class JITModBuilder {
     }
   }
 
+  // FIXME: this function need to be refactored.
```
vext01, line 72: Gah, need**s*

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -139,22 +142,34 @@ class JITModBuilder {
 
       // Iterate over all instructions within this block and copy them over
       // to our new module.
+      // FIXME It would be nice to merge the two skipping mechanisms
```
ltratt, line 145: `FIXME` or `FIXME:` (I'm indifferent, but we should be consistent).

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -139,22 +142,34 @@ class JITModBuilder {
...
+          }
+        }
 
-        // If we've returned from a call, skip ahead to the instruction where
-        // we left off.
+        // If we've returned from an internal call, skip ahead to the
+        // instruction where we left off.
         if (last_call != nullptr) {
           if (&*I == last_call) {
             last_call = nullptr;
```
ltratt, line 162: Silly question perhaps: does this deal with arbitrary levels of "if we've returned from an internal call" due to inlining?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -176,10 +191,16 @@ class JITModBuilder {
                   CF->getName(), JITMod);
               VMap[CF] = DeclFunc;
             }
+            copyInstruction(&Builder, (Instruction *)&*I);
             // We should expect an "unmappable hole" in the trace. This is
             // where the trace followed a call into external code for which be
             // have no IR, and thus we cannot map blocks for.
             ExpectUnmappable = true;
+            // Peek at the instruction we will need to resume at.
+            auto PeekIter = I;
```
ltratt, line 200: Aren't these three lines the same as `ExcCallInst = &*(I++);` ?

ykllvmwrap/src/jitmodbuilder.cc
vext01: Yes. Lukas implemented that. See the fib test.

ltratt: The thing that worries me is that `last_call` suggests we can only handle this to one level. Perhaps that's all we need to handle, but it's not obvious to me.

ykllvmwrap/src/jitmodbuilder.cc
vext01: `PeekIter = I` makes a copy of the iterator, leaving the original iterator untouched. Apparently this is how you peek an iterator in C++: copy the iterator.

ltratt: My eyes hurt! But I have no idea about this stuff, so I shall not rant further :)

ykllvmwrap/src/jitmodbuilder.cc
vext01: `last_call` is populated by popping from a stack of inlined calls.

But @ptersilie is your guy for this logic. I'm still a little over-whelmed by it!

ykllvmwrap/src/jitmodbuilder.cc
vext01: It's a C++ thing. When you have a value on the stack, and assign something from it, you call a "copy constructor" to make a fresh instance.

It tripped me too when I first started with C++.

vext01: Just to be 100% sure:

```
#include <iostream>
#include <vector>

using namespace std;

int main(void) {
    vector<int> MyVec = {1, 2, 3};
    auto it1 = MyVec.begin();
    auto it2 = it1;
    it2 += 2;
    cout << "it1: " << *it1 << "\n";
    cout << "it2: " << *it2 << "\n";
}
```

Gives:
```
it1: 1
it2: 3
```

vext01:
That should be all comments addressed.

ykllvmwrap/src/jitmodbuilder.cc
ltratt: Can you file an issue / talk to @ptersilie about this? I agree that it's not something for this PR, but it does look fishy, and I don't want us to forget to look at it.

ltratt:
Please squash.

vext01:
Sorry, one more change. A variable name was misleading. OK to squash?

ltratt:
Please squash.

vext01:
splat.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: `last_call` is populated here:
```
if (isa<ReturnInst>(I)) {
  last_call = inlined_calls.back();
}
```
where `inlined_calls` is a vector of calls we've followed (and the tracer has inlined).

PR #387, opened 2021-07-27
Mapper fixes

vext01:
This makes Hello World in BF JIT!

I need to come clean about one aspect of this PR:
> // Part of a PT block mapped to a machine block in the LLVM block address
map, but the machine block has no corresponding IR blocks. This happens because LLVM can introduce extra `MachineBasicBlock`s to help with laying out machine code. For our purposes these extra blocks can be ignored.

The sentence starting "this happens because" is pure speculation. Despite some grepping, asking in discord, and asking a [stackoverflow question](https://stackoverflow.com/questions/68532562/llvm-when-does-machinebasicblockgetbasicblock-return-nullptr), I'm still not 100% sure why `MachineBasicBlock::getParent()` may return `nullptr`. The [LLVM docs](https://llvm.org/doxygen/classllvm_1_1MachineBasicBlock.html#a856266240556bbae90bc5ba75c156eef) are vague...

If we want a conclusive answer, I'll need to spend more time on this. What do you think?

### DIFF
--- a/yktrace/src/hwt/mapper.rs
+++ b/yktrace/src/hwt/mapper.rs
@@ -195,25 +195,16 @@ impl HWTMapper {
             .query(block_off, block_off + block_len)
             .collect::<Vec<_>>();
 
-        // If a PT block maps to multiple IR blocks, then the IR blocks should be at consecutive
-        // addresses (they should be related only by "fall-thru", without control flow dispatch, as
-        // depicted in the above doc string). For debug builds, we check this.
-        #[cfg(debug_assertions)]
-        let mut prev_ent: Option<&intervaltree::Element<_, _>> = None;
-
+        // In the case that a PT block maps to multiple machine blocks, it may be tempting to check
+        // that they are at consecutive address ranges. Unfortunately we can't do this because LLVM
+        // sometimes appends `nop` sleds (e.g. `nop word cs:[rax + rax]; nop`) to the ends of
+        // blocks for alignment. This padding is not reflected in the LLVM block address map, so
+        // blocks may not appear consecutive.
         ents.sort_by(|x, y| x.range.start.partial_cmp(&y.range.start).unwrap());
         for ent in ents {
             // Check that the MachineBasicBlock observed in the trace has a corresponding BasicBlock.
             // PERF: can we guarantee this won't happen and downgrade to a debug assertion?
             assert!(!ent.value.corr_bbs.is_empty());
-
-            #[cfg(debug_assertions)]
-            {
-                if let Some(prev) = prev_ent {
-                    debug_assert!(ent.range.start == prev.range.end);
-                }
-                prev_ent = Some(ent);
-            }
 
             let func_name = self.symb.find_code_sym(&obj_name, ent.value.f_off).unwrap();
             self.faddrs.insert(func_name.clone(), ent.value.f_off);
--- a/yktrace/src/hwt/mapper.rs
+++ b/yktrace/src/hwt/mapper.rs
@@ -129,15 +129,27 @@ impl HWTMapper {
                 }
             } else {
                 for irblock in irblocks.into_iter() {
-                    // A basic block may map to >1 *machine* basic block. If we see the same block
-                    // repeated, then we need to check if we re-executing the same basic block or
-                    // if we are executing next machine basic block for that same block. In the
-                    // latter case, we mustn't record the block in the trace again.
-                    if !(ret_irblocks.last() == Some(&irblock)
-                        && prev_block.as_ref().map(|x| x.first_instr())
-                            != Some(block.first_instr()))
-                    {
-                        ret_irblocks.push(irblock);
+                    if let Some(irblock) = irblock {
+                        // An IR basic block may map to >1 *machine* basic block. If we see the
+                        // same IR block repeated, then we need to check if we re-executing the
+                        // same basic block or if we are executing the next machine basic block for
+                        // that same block. In the latter case, we mustn't record the block in the
+                        // trace again.
+                        if !(ret_irblocks.last() == Some(&irblock)
+                            && prev_block.as_ref().map(|x| x.first_instr())
+                                != Some(block.first_instr()))
+                        {
+                            ret_irblocks.push(irblock);
+                        }
+                    } else {
+                        // Part of a PT block mapped to a machine block in the LLVM block address
+                        // map, but the machine block has no corresponding IR blocks.
+                        //
+                        // FIXME: https://github.com/ykjit/yk/issues/388
+                        // We *think* this happens because LLVM can introduce extra
+                        // `MachineBasicBlock`s to help with laying out machine code. If that's the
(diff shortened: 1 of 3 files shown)


ptersilie:
We have at least one example, where such a machine block is quite clearly padding (i.e. `nop`s). So it's more of an educated guess than pure speculation.

vext01:
> We have at least one example where such a machine block is quite clearly padding (i.e. nops). So it's more of an educated guess than pure speculation.

That I'm sure about, but I'm talking about the second commit, where we get blocks with no corresponding IR blocks.

yktrace/src/hwt/mapper.rs
```diff
@@ -195,26 +195,17 @@ impl HWTMapper {
...
         ents.sort_by(|x, y| x.range.start.partial_cmp(&y.range.start).unwrap());
         for ent in ents {
             // Check that the MachineBasicBlock observed in the trace has a corresponding BasicBlock.
             // PERF: can we guarantee this won't happen and downgrade to a debug assertion?
             assert!(!ent.value.corr_bbs.is_empty());
 
-            #[cfg(debug_assertions)]
-            {
-                if let Some(prev) = prev_ent {
-                    debug_assert!(ent.range.start == prev.range.end);
```
ltratt, line 213: Can't we adjust this so that it's `ent.rang.start > prev.range.end`?

yktrace/src/hwt/mapper.rs
```diff
@@ -195,26 +195,17 @@ impl HWTMapper {
...
-        // If a PT block maps to multiple IR blocks, then the IR blocks should be at consecutive
-        // addresses (they should be related only by "fall-thru", without control flow dispatch, as
-        // depicted in the above doc string). For debug builds, we check this.
-        #[cfg(debug_assertions)]
-        let mut prev_ent: Option<&intervaltree::Element<_, _>> = None;
-
+        // In the case that a PT block maps to multiple machine blocks, it may be tempting to check
+        // that they are at consecutive address ranges. Unfortunately we can't do this because LLVM
+        // sometimes appends `nop` sleds (e.g. `nop word cs:[rax + rax]; nop`) to the ends of
+        // blocks for alignment. This padding is not reflected the LLVM block address map, so
```
ltratt, line 231: s/the LLVM/in the LLVM/

yktrace/src/hwt/mapper.rs
```diff
@@ -129,15 +129,24 @@ impl HWTMapper {
...
-                    // latter case, we mustn't record the block in the trace again.
-                    if !(ret_irblocks.last() == Some(&irblock)
-                        && prev_block.as_ref().map(|x| x.first_instr())
-                            != Some(block.first_instr()))
-                    {
-                        ret_irblocks.push(irblock);
+                    if let Some(irblock) = irblock {
+                        // A basic block may map to >1 *machine* basic block. If we see the same
+                        // block repeated, then we need to check if we re-executing the same basic
+                        // block or if we are executing next machine basic block for that same
```
ltratt, line 135: s/next machine/the next machine/

yktrace/src/hwt/mapper.rs
```diff
@@ -129,15 +129,24 @@ impl HWTMapper {
...
-                    // if we are executing next machine basic block for that same block. In the
-                    // latter case, we mustn't record the block in the trace again.
-                    if !(ret_irblocks.last() == Some(&irblock)
-                        && prev_block.as_ref().map(|x| x.first_instr())
-                            != Some(block.first_instr()))
-                    {
-                        ret_irblocks.push(irblock);
+                    if let Some(irblock) = irblock {
+                        // A basic block may map to >1 *machine* basic block. If we see the same
+                        // block repeated, then we need to check if we re-executing the same basic
```
ltratt, line 134: I think you need to specify which block type is which explicitly: I'm not sure what "the same basic block" refers to.

yktrace/src/hwt/mapper.rs
```diff
@@ -129,15 +129,24 @@ impl HWTMapper {
...
+                                != Some(block.first_instr()))
+                        {
+                            ret_irblocks.push(irblock);
+                        }
+                    } else {
+                        // Part of a PT block mapped to a machine block in the LLVM block address
+                        // map, but the machine block has no corresponding IR blocks. This happens
+                        // because LLVM can introduce extra `MachineBasicBlock`s to help with
+                        // laying out machine code. For our purposes these extra blocks can be
+                        // ignored.
```
ltratt, line 149: The PR description is a bit vague: do we actually see this happen? If we haven't yet seen it happen, but *think* it can happen, I suggest putting `errx` or similar here so that we're notified of it and can check that we do the right thing if/when we see it happen.

yktrace/src/hwt/mapper.rs
vext01: Above we have:
```
ents.sort_by(|x, y| x.range.start.partial_cmp(&y.range.start).unwrap());
```

So that assertion wouldn't be very valuable.

ltratt: Point taken.

yktrace/src/hwt/mapper.rs
vext01: It does indeed happen when JITing ykcbf.

ltratt: How do we know that we can ignore these? Are they always fall-through or ... ?

vext01: I can add a not in the commit message at squash time if you like.

ltratt: How does that relate to my questions?

vext01:
All comment addressed.

@ltratt what did you think about:

> The sentence starting "this happens because" is pure speculation. Despite some grepping, asking in discord, and asking a stackoverflow question, I'm still not 100% sure why MachineBasicBlock::getParent() may return nullptr. The LLVM docs are vague...

ltratt:
I think we should be clear what is speculation and what isn't. We probably need a FIXME and an issue so that we know to look at this, but I don't think we want to block this PR from being merged, as it's still better than what we had before.

yktrace/src/hwt/mapper.rs
vext01: (we raced, will answer after lunch)

vext01: > How do we know that we can ignore these?

My hypothesis is that if the LLVM IR captures all required program semantics, then extra blocks are either semantically ineffectual, or contain semantics transferred from other IR blocks. In both cases, the LLVM IR would remain stand alone.

> Are they always fall-through or ... ?

I think the block in question contained an `add`, which would suggest that the block is semantically meaningful, but I can't say if that was due to code moving out of a normal IR block and into one of these "extra" blocks or not. I'd have to dig deeper.

e83176b should make it clear that more work is required. It references an issue I've just created.

ltratt:
Please squash.

vext01:
Thanks. Splat.

PR #401, opened 2021-08-10
Refactor round 2

vext01:
Various small refactorings to aid readability.

### DIFF
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -187,27 +187,25 @@ public:
     this->FAddrKeys = FAddrKeys;
     this->FAddrVals = FAddrVals;
     this->FAddrLen = FAddrLen;
+
+    JITMod = new Module("", AOTMod->getContext());
   }
 
   // FIXME: this function needs to be refactored.
   // https://github.com/ykjit/yk/issues/385
   Module *createModule() {
-    LLVMContext &JITContext = AOTMod->getContext();
-    JITMod = new Module("", JITContext);
-    uint64_t TraceIdx = NextTraceIdx.fetch_add(1);
-    if (TraceIdx == numeric_limits<uint64_t>::max())
-      errx(EXIT_FAILURE, "trace index counter overflowed");
-
+    LLVMContext &JITContext = JITMod->getContext();
     // Get var args from start_tracing call.
     auto Inputs = getTraceInputs(AOTMod->getFunction(FuncNames[0]), BBs[0]);
 
-    std::vector<Type *> InputTypes;
-    for (auto Val : Inputs) {
-      InputTypes.push_back(Val->getType());
-    }
-
     // Create function to store compiled trace.
+    uint64_t TraceIdx = NextTraceIdx.fetch_add(1);
+    if (TraceIdx == numeric_limits<uint64_t>::max())
+      errx(EXIT_FAILURE, "trace index counter overflowed");
     TraceName = string(TRACE_FUNC_PREFIX) + to_string(TraceIdx);
+    std::vector<Type *> InputTypes;
+    for (auto Val : Inputs)
+      InputTypes.push_back(Val->getType());
     llvm::FunctionType *FType =
         llvm::FunctionType::get(Type::getVoidTy(JITContext), InputTypes, false);
     llvm::Function *JITFunc = llvm::Function::Create(
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -165,6 +165,26 @@ class JITModBuilder {
     }
   }
 
+  void handleReturnInst(Instruction *I) {
+    ResumeAfter = InlinedCalls.back();
+    InlinedCalls.pop_back();
+    if (call_stack > 0) {
+      call_stack -= 1;
+      if (call_stack == 0) {
+        ResumeAfter = NoInlineFunc;
+      }
+      return;
+    }
+    // Replace the return variable of the call with its return value.
+    // Since the return value will have already been copied over to the
+    // JITModule, make sure we look up the copy.
+    auto OldRetVal = ((ReturnInst *)&*I)->getReturnValue();
+    if (OldRetVal != nullptr) {
+      assert(ResumeAfter.hasValue());
+      VMap[get<1>(ResumeAfter.getValue())] = getMappedValue(OldRetVal);
+    }
+  }
+
   Function *createJITFunc(vector<Value *> *TraceInputs) {
     // Compute a name for the trace.
     uint64_t TraceIdx = NextTraceIdx.fetch_add(1);
@@ -309,23 +329,7 @@ public:
         }
 
         if (isa<ReturnInst>(I)) {
-          ResumeAfter = InlinedCalls.back();
-          InlinedCalls.pop_back();
-          if (call_stack > 0) {
-            call_stack -= 1;
-            if (call_stack == 0) {
-              ResumeAfter = NoInlineFunc;
-            }
-            continue;
-          }
-          // Replace the return variable of the call with its return value.
-          // Since the return value will have already been copied over to the
-          // JITModule, make sure we look up the copy.
-          auto OldRetVal = ((ReturnInst *)&*I)->getReturnValue();
-          if (OldRetVal != nullptr) {
-            assert(ResumeAfter.hasValue());
-            VMap[get<1>(ResumeAfter.getValue())] = getMappedValue(OldRetVal);
-          }
+          handleReturnInst(&*I);
           break;
         }
 
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -185,6 +185,15 @@ class JITModBuilder {
     }
   }
 
+  void handlePHINode(Instruction *I, Function *F, size_t Idx) {
+    assert(Idx > 0);
+    auto LBIt = F->begin();
+    std::advance(LBIt, BBs[Idx - 1]);
+    BasicBlock *LastBlock = &*LBIt;
+    Value *V = ((PHINode *)&*I)->getIncomingValueForBlock(LastBlock);
+    VMap[&*I] = getMappedValue(V);
+  }
+
   Function *createJITFunc(vector<Value *> *TraceInputs) {
     // Compute a name for the trace.
     uint64_t TraceIdx = NextTraceIdx.fetch_add(1);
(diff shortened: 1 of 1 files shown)


vext01:
Just realised that the first commit message needs to be revised. We didn't move the code that makes the trace name.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -187,27 +187,25 @@ class JITModBuilder {
     this->FAddrKeys = FAddrKeys;
     this->FAddrVals = FAddrVals;
     this->FAddrLen = FAddrLen;
+
+    JITMod = new Module("", AOTMod->getContext());
```
ltratt, line 253: Is it safe to have an empty module name?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -187,27 +187,25 @@ class JITModBuilder {
...
     auto Inputs = getTraceInputs(AOTMod->getFunction(FuncNames[0]), BBs[0]);
 
-    std::vector<Type *> InputTypes;
-    for (auto Val : Inputs) {
-      InputTypes.push_back(Val->getType());
-    }
-
     // Create function to store compiled trace.
+    uint64_t TraceIdx = NextTraceIdx.fetch_add(1);
+    if (TraceIdx == numeric_limits<uint64_t>::max())
```
ltratt: *Technically* this check is, I think, wrong: the trace index will overflow on the next `fetch_add`. But... is the explicit failure worth it? In other words, just before line 202, I'd add `assert(NextTraceIdx != uin64_t::max())` (or something to that effect).

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -165,6 +165,26 @@ class JITModBuilder {
     }
   }
 
+  Function *createJITFunc(vector<Value *> *TraceInputs) {
+    // Compute a name for the trace.
+    uint64_t TraceIdx = NextTraceIdx.fetch_add(1);
+    if (TraceIdx == numeric_limits<uint64_t>::max())
```
ltratt, line 200: Same comment as before (this is probably better as a debug `assert` before the `fetch_add`).

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -185,6 +185,19 @@ class JITModBuilder {
     return JITFunc;
   }
 
+  // Variables that are used (but not defined) inbetween starting and stopping
+  // tracing need to be replaced with function arguments which the user passes
+  // into the compiled trace. This loop creates a mapping from those original
+  // variables to the function arguments of the compiled trace function.
+  void mapTraceInputs(vector<Value *> &TraceInputs, Function *JITFunc) {
+    for (size_t Idx = 0; Idx != TraceInputs.size(); Idx++) {
```
ltratt, line 222: Would s/!=/</ not be clearer?

ykllvmwrap/src/jitmodbuilder.cc
vext01: Yep.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -165,6 +165,26 @@ class JITModBuilder {
     }
   }
 
+  void handleReturnInst(Instruction *I) {
+    ResumeAfter = InlinedCalls.back();
+    InlinedCalls.pop_back();
+    if (call_stack > 0) {
+      call_stack -= 1;
+      if (call_stack == 0) {
+        ResumeAfter = NoInlineFunc;
```
ltratt, line 174: Is it just me, or does this assign to a variable, then the function returns, so the variable assignment can't have any effect?

ykllvmwrap/src/jitmodbuilder.cc
vext01: This is existing code that has been moved. Shall we fix this in a separate commit?

ltratt: OK. There are a couple of these, so doing it in a new commit (but as part of this PR) is fine with me.

ykllvmwrap/src/jitmodbuilder.cc
vext01: It's a field.

[One of the tasks on my list is to minimise the number of fields in the class, but that will be a separate PR]

ltratt: Can we use `this` or `self` or whatever for these? IMHO, the "implicit field thing" is a horrible language feature!

vext01: You can use the `this` pointer, but it's only really idiomatic to use it when disambiguating a local variable of the same name.

ltratt: Java has the same expectation... and it seems stupid to me. It reduces one's ability to reason about code locally (as I've just shown), particularly in diffs. We don't have to change all the existing code, but IMHO we should consider using `this` for new stuff. Anyway, not an issue here and now.

ykllvmwrap/src/jitmodbuilder.cc
vext01: Wait. The fix you are proposing wouldn't be thread-safe, would it? The check itself needs to be atomic, which is why I think I tried to do it with a single `fetch_add`.

You are right that we lose one trace ID though...

ltratt: Is this `fetch_add` an atomic instruction? It doesn't have an ordering, so I assumed it's just a bog-standard non-thread-safe version.

ykllvmwrap/src/jitmodbuilder.cc
vext01: I agree that it's not nice, but since we've adopted LLVM's coding conventions for our C++ code, I think the right thing to do is follow the precedent.

[It's not *that* bad. If you see a variable undefined locally, it's in another scope]

ltratt: > [It's not that bad. If you see a variable undefined locally, it's in another scope]

It really is bad, because now I have to *remember* the entire set of local variables to know if something's a variable or a field -- two things that have very different semantics. For diffs this means I have to look at the entire function, otherwise (as in this case) I will get completely the wrong impression. Anyway, I won't insist upon this, and it's not a concern for this PR.

ykllvmwrap/src/jitmodbuilder.cc
vext01: It's using the default ordering, which is actually too strong here I think:
```
T fetch_add( T arg,
             std::memory_order order = std::memory_order_seq_cst ) noexcept;
```

Since the counter only depends on itself, and since there's a total ordering on atomic ops for the same value, I think it can be weakened to `memory_order_relaxed`.

Do you agree?

ykllvmwrap/src/jitmodbuilder.cc
vext01: To be clear, I totally agree. But it's no worse than global variables. They too force you to look around more :P

ykllvmwrap/src/jitmodbuilder.cc
ltratt: Yes, relaxed ordering is fine. I suggest this should actually be in a self contained function called something like `next_trace_idx` that can encapsulate this check. It should probably be something like (in Rust, because I'm lazy):

```rust
let x = self.current_idx.fetch_add(Ordering::Relaxed, 1);
debug_assert!(x < u64::max());
x
```

[We could squeeze out the last value in the u64 range... but it doesn't really seem worth it.]
```

vext01: That's pretty much what we have. `x < u64::max()` is equivalent to checking that we don't have `x == u64::max`.

I will move it into a function though. Good idea.

ykllvmwrap/src/jitmodbuilder.cc
vext01: Shouldn't we do this check always? An assert would only do the check in debug builds.

ltratt: It takes something like 100 years of continuous incrementing for a 64 bit int to overflow on a single core machine, so having a permanent assert for this seems a bit OTT: it's, ahem, not very likely to trigger. That said, I *like* debug asserts because they can catch other weirder errors (e.g. someone accidentally sets this to u64::max() or similar), but without a runtime cost.

vext01: So on bencher13, we could overflow that integer in just under 3-years (unreasonable assuming that a) there's 0-cost threading overhead, and b) a program could spawn so much JIT compilation. But you get my point).

When this integer overflows the program is likely to explode anyway (probably with a  symbol linkage error), so the question then becomes: when do you want it to explode? When the invalid ID is forged, or later?

ltratt: There is not zero threading overhead and it couldn't spawn that much JIT compilation ;) If you managed to spill this in our case in under 100,000 years, I'd be flabbergasted. It's not going to explode in anyone's lifetimes *unless* someone does something idiotic, at which point a debug assert should help catch that.

vext01: > at which point a debug assert should help catch that

Why would a user running a JIT for years be using a debug build though?

I have a better idea (for a later PR).

The trace's name is only useful for debugging (for setting breakpoints in gdb). So we can only assign a name to the trace in debug builds. In release builds we can let LLVM do whatever it does to de-duplicate JIT symbol names. Sounds good?

(end extreme bikeshedding)

ltratt: > Why would a user running a JIT for years be using a debug build though?

We're literally talking tens of thousands years, minimum, for this to happen. No-one will have a single yk process running for that long ;)

> The trace's name is only useful for debugging (for setting breakpoints in gdb). So we can only assign a name to the trace in debug builds. In release builds we can let LLVM do whatever it does to de-duplicate JIT symbol names. Sounds good?

Works for me!

vext01: https://github.com/ykjit/yk/issues/402

ltratt: OK. Let's at least put that stuff in a function is this PR, though, so addressing the issue later becomes easier?

ykllvmwrap/src/jitmodbuilder.cc
vext01: Fixed in 4ccb5ac0c15

ykllvmwrap/src/jitmodbuilder.cc
vext01: Other stuff fixed in 15e019e7c7679

vext01:
That's everything. OK to force push a fix to the first commit message?

ltratt:
Please squash.

vext01:
Splat. Commit msg revised.

I left the concurrency bits in a separate commit.

PR #403, opened 2021-08-10
Refactor handleCallInst function.

ptersilie:
More commits simplifying and improving readability of the `jitmodbuilder` code, in particular `handleCallInst`.

### DIFF
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -107,15 +107,21 @@ class JITModBuilder {
     return false;
   }
 
+  // Add an external declaration for the given function to JITMod.
+  void declareFunction(Function *F) {
+    assert(JITMod->getFunction(F->getName()) == nullptr);
+    auto DeclFunc = llvm::Function::Create(F->getFunctionType(),
+                                           GlobalValue::ExternalLinkage,
+                                           F->getName(), JITMod);
+    VMap[F] = DeclFunc;
+  }
+
   void handleCallInst(CallInst *CI, Function *CF, size_t &CurInstrIdx) {
     if (CF->isDeclaration()) {
       // The definition of the callee is external to AOTMod. We still
       // need to declare it locally if we have not done so yet.
       if (VMap[CF] == nullptr) {
-        auto DeclFunc = llvm::Function::Create(CF->getFunctionType(),
-                                               GlobalValue::ExternalLinkage,
-                                               CF->getName(), JITMod);
-        VMap[CF] = DeclFunc;
+        declareFunction(CF);
       }
       if (RecCallDepth == 0) {
         copyInstruction(&Builder, (Instruction *)&*CI);
@@ -138,11 +144,7 @@ class JITModBuilder {
       if (isRecursiveCall(CF)) {
         if (VMap[CF] == nullptr) {
           StringRef CFName = CF->getName();
-          // Declare function.
-          auto DeclFunc = llvm::Function::Create(CF->getFunctionType(),
-                                                 GlobalValue::ExternalLinkage,
-                                                 CFName, JITMod);
-          VMap[CF] = DeclFunc;
+          declareFunction(CF);
           for (size_t i = 0; i < FAddrLen; i++) {
             char *FName = FAddrKeys[i];
             uint64_t FAddr = FAddrVals[i];
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -96,7 +96,7 @@ class JITModBuilder {
   }
 
   void handleCallInst(CallInst *CI, Function *CF, size_t &CurInstrIdx) {
-    if ((StartTracingInstr != nullptr) && (CF->isDeclaration())) {
+    if (CF->isDeclaration()) {
       // The definition of the callee is external to AOTMod. We still
       // need to declare it locally if we have not done so yet.
       if (VMap[CF] == nullptr) {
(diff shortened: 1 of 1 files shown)


ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -95,6 +95,17 @@ class JITModBuilder {
     }
   }
 
+  bool isRecursiveCall(Function *F) {
```
ltratt, line 94: Can we add a doc string for this?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -123,31 +134,26 @@ class JITModBuilder {
...
+        if (VMap[CF] == nullptr) {
+          StringRef CFName = CF->getName();
+          // Declare function.
+          auto DeclFunc = llvm::Function::Create(CF->getFunctionType(),
+                                                 GlobalValue::ExternalLinkage,
+                                                 CFName, JITMod);
+          VMap[CF] = DeclFunc;
+          for (size_t i = 0; i < FAddrLen; i++) {
+            char *FName = FAddrKeys[i];
+            uint64_t FAddr = FAddrVals[i];
```
ltratt: Is it worth breaking this out as a separate variable? It only seems to be used once (on line 149)?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -137,11 +141,7 @@ class JITModBuilder {
       if (isRecursiveCall(CF)) {
         if (VMap[CF] == nullptr) {
           StringRef CFName = CF->getName();
-          // Declare function.
-          auto DeclFunc = llvm::Function::Create(CF->getFunctionType(),
-                                                 GlobalValue::ExternalLinkage,
-                                                 CFName, JITMod);
-          VMap[CF] = DeclFunc;
+          declareFunction(CF);
```
ltratt, line 151: Possibly silly question: can the same function be declared on both lines 121 and 144?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -113,6 +113,18 @@ class JITModBuilder {
     VMap[F] = DeclFunc;
   }
 
+  void addGlobalMappingForFunction(Function *CF) {
```
ltratt, line 112: Let's add a docstring.

ykllvmwrap/src/jitmodbuilder.cc
vext01: If you declare a duplicate function name, LLVM will add a de-duplications suffix to the symbol name.

That said, I think `declareFunction()` should check so as to not bloat up the IR. It should probably also `assert` that the signature is the same?

ltratt: > That said, I think declareFunction() should check so as to not bloat up the IR. It should probably also assert that the signature is the same?

Agreed on both points. [Although -- if it's possible -- better than checking if the function has been declared would be to work out how not to double declare it in the first place?]

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -67,11 +67,7 @@ class JITModBuilder {
   // Instruction at which to continue after an a call.
   Optional<tuple<size_t, CallInst *>> ResumeAfter;
   // Depth of nested calls when outlining a recursive function.
-  // FIXME: can we kill this?
   size_t RecCallDepth = 0;
-  // Function currently being outlined.
-  // FIXME: this should be an Optional?
-  tuple<size_t, CallInst *> NoInlineFunc;
```
vext01, line 74: Brilliant!

vext01:
Not much to add on top of what laurie has already said.

ykllvmwrap/src/jitmodbuilder.cc
vext01: It's inevitable when the trace calls the same symbol a second time.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: I find it improves readability. But I don't feel strongly about it. `FName` is also only used once.

ykllvmwrap/src/jitmodbuilder.cc
ltratt: Ah!

ykllvmwrap/src/jitmodbuilder.cc
ltratt: I don't feel strongly either, so do as you see fit.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: We already guard against that with `if (VMap[CF] == nullptr) {`. We never call `declareFunction` if it's already in the `VMap`.

ltratt: Even better. Still, an `assert` to make this clear in `declareFunction` might be a good idea.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Fixed in 27db37f.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Fixed in f71b25b.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: I think d730754 should do the trick. See if you agree at @vext01.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: I think I prefer it as is.

ykllvmwrap/src/jitmodbuilder.cc
vext01: LGTM

ptersilie:
> Based on another discussion, I think this function's docstring might better be something like "Find the machine code corresponding to the AOT IR `CF` and ensure there's an entry from `CF`'s name to that machine code".

Try ed59f3e.

ltratt:
Please squash.

ptersilie:
Squashed.

ptersilie:
Forgot to push a formatting change.

ptersilie:
There it is.

ltratt:
Please squash.

ptersilie:
Squashed.

Issue #397, opened 2021-08-09
Checking if there is a VMap entry with `VMap[x] == nullptr` is incorrect.


vext01:
If the key doesn't exist, this mutates the mapping! And that later confuses LLVM's mapping code in some cases.

The correct way is to use `find()` and check if the resulting iterator is `end()`.

PR #404, opened 2021-08-11
Fix VMap element existence check.

ptersilie:
Previously we used `VMap[Var] != nullptr` to check if an element exists
in the VMap. Unfortunately, C++ thinks its a good idea to insert a
default value (in this case `nullptr`) whenever an element is not found.
This lead to errors in LLVM which checks existence in the VMap via
`VMap.find(Var) != VMap.end()`, which now returns true after our above
check. In other words, checking if an element exists in the VMap in our
code, pushed it into existence in LLVM's world, even if it didn't
exist in the VMap before.

The solution, is to use the appropriate existence check
`VMap.find(Var) != VMap.end()` in our code. This means we also no longer
need to id-map constants to work around the `Unexpected null mapping`
error during instruction cloning.


ltratt:
This looks good to me, but @vext01 will be a better reviewer than I.

ptersilie:
Addresses https://github.com/ykjit/yk/issues/397.

vext01:
I fully approve.

PR #405, opened 2021-08-11
Make modules more debuggable by annotating debug (DWARF) information.

### LINKED ISSUE #400: Re-think the printing modes.
During our sprint, I think we learned that:
 - The second column of "SBS" mode isn't very useful.
 - The line info of "SBS" mode is *very* useful.
 - Inlining thresholds annotated in SBS mode are somewhat useful.

I propose that we:
 - Remove SBS mode.
 - Make all print modes show debug line info (and inlining thresholds?).

This means:
 - Making the non-SBS modes print their modules line-by-line and porting the line info code over to them.
 - Finding out how to print the module-level bits like globals.

We should probably finish #392 first.

vext01:
We do this by:
 - Copying dbg! metadata over into the JIT module.
 - Displaying function, file, line and column annotations when printing
   the trace.
 - Retiring the "SBS" print mode.

Debug info, when available, is printed on the line preceeding the
instruction to which it applies using an LLVM comment (`;`). By doing it
this way (in a comment) the syntax of the IR remains valid.

E.g.:
```
; main() tests/simple.c:29:10
store void (i32*)* %24, void (i32*)** %10, align 8, !dbg !49
; main() tests/simple.c:30:7
call void @llvm.dbg.declare(metadata i32* %11, metadata !52, metadata !DIExpression()), !dbg !53
store i32 0, i32* %11, align 4, !dbg !53
; main() tests/simple.c:31:3
%25 = load void (i32*)*, void (i32*)** %10, align 8, !dbg !54
call void %25(i32* %11), !dbg !54
```

This works for all print modes.

Fixes #400


vext01:
This makes #392 redundant. Closing that now.

ltratt:
Shouldn't we have some tests that include file name and line information in them?

ltratt:
Ah, you're a "guilty until proven innocent" chap I see ;) I don't think we should exclude tests for this until and unless they prove to be fragile.

ltratt:
Don't forget that lang_tester can set environment variables, so I don't think there should be an issue getting the debug info out?

vext01:
Try 8dd8052.

ltratt:
Perfect! Please squash.

vext01:
Mind if i also sync the branch with master at the same time?

ltratt:
Squashing implies rebasing so go ahead!

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -469,5 +456,20 @@ class JITModBuilder {
       if (G->hasInitializer())
         NewGV->setInitializer(MapValue(G->getInitializer(), VMap));
     }
+
+    // Ensure that the JITModule has a `!llvm.dbg.cu`.
+    // This code is borrowed from LLVM's `cloneFunction()` implementation.
+    // OPT: Is there a faster way than scanning the whole module?
```
ptersilie, line 462: I assume we only ever run this in debug mode, so it's probably not worth worrying about. Otherwise, we might be able to inline it into the processing of operands.

vext01:
splat.

PR #406, opened 2021-08-11
Fix constant expressions.

ptersilie:
See first commit message for details.


vext01:
>  it does not remap the global variable, since we haven't copied it into JITMod yet.

If you copy the global first, can we avoid the dance then?

c_tests/tests/constexpr.c
```diff
@@ -0,0 +1,31 @@
+// Compiler:
+// Run-time:
+
+#include <assert.h>
```
vext01, line 4: Can we add a comment saying what the test checks?

c_tests/tests/constexpr.c
```diff
@@ -0,0 +1,31 @@
+// Compiler:
+// Run-time:
```
vext01, line 2: Did we want to check for the presence of a constant expr somewhere in the module?

ptersilie: Fixed in 5ec3b32.

c_tests/tests/constexpr.c
ptersilie: Fixed in e4ccdcf.

vext01:
LGTM. Please squash.

ptersilie:
Squashed.

vext01:
Needs cfmt.

ptersilie:
Fixed.

Issue #407, opened 2021-08-12
Calls to functions with internal linkage.


vext01:
A C function qualified `static` causes the function to be marked with internal linkage. This means that the function will never appear in the dynamic symbol table, even if you build your target with `-Wl,-export-dynamic`.

This means that if a trace needs to call one of these symbols, linkage will fail when LLVM loads the JITted code.

We can find the addresses of internally linked symbols by inspecting the static symbol table of the binary (using something like `nlist(3)` from libbsd). This could be an expensive operation as it requires filesystem access (the symbol table is not loaded into memory). Caching will probably be necessary.

vext01:
> This means that if a trace needs to call one of these symbols, linkage will fail when LLVM loads the JITted code.

But here's the twist. If the hardware tracer executed through such a symbol, then `yk` discovers the symbol address from the PT trace.

So this may not be a concern after all.

vext01:
I'm going to close this for now.

PR #408, opened 2021-08-12
More linkage tests

vext01:
I had intended to write a failing test to form the basis of a fix for #407, however, I was unable to make a test where a trace called a static function from another object and the function has no IR. It doesn't look like this situation can happen...

Nonetheless I think it would still be good to add these two tests, as they check scenarios that were not previously tested.

Note that I had to extend the test runner. See individual commits.

### DIFF
--- a/c_tests/Cargo.toml
+++ b/c_tests/Cargo.toml
@@ -12,5 +12,6 @@ harness = false
 
 [dev-dependencies]
 lang_tester = "0.7.0"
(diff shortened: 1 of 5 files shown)


ltratt:
Looks OK to me!

ptersilie:
Does it need a rebase?

vext01:
rebased.

PR #410, opened 2021-08-12
Re-enable higher AOT optimisation levels in the C test suite.

vext01:
I need this so that I can test a problematic PHI node case, but I can
only get the IR I want at -O1 (LLVM tends to emit allocas instead at
-O0).

I had to fix some tests, mostly due to the recent addition of debug
metadata (and comments) in traces.

Some tests had other problems. If I couldn't easily fix them, then I
skipped those for now.


c_tests/tests/debug_debuginfo.c
```diff
@@ -1,19 +1,4 @@
...
-//       ; main() tests/debug_debuginfo.c:28:9
-//       store i8* null, i8** %2, align 8, !dbg !3
-//       ; main() tests/debug_debuginfo.c:29:7
-//       store i32 2, i32* %0, align 4, !dbg !12
-//       ; main() tests/debug_debuginfo.c:30:37
-//       %3 = load i8*, i8** %2, align 8, !dbg !13
-//       ret void
-//     }
-//     ...
+// ignore: impractical to test all optimisation levels.
```
ltratt, line 1: Couldn't we have separate test files for the different opt levels?

vext01: We could, but I didn't see the benefit in checking exact position of the comments and the line and column numbers in the comments for every single opt level. Especially since debuginfo is approximate.

vext01: I could make one more generic test which just checks if there is at least on debug comment in the IR. How about that?

ltratt: Let's say the current test is the O0 test, why not put it in `debug_debuginfo.c.O0` as-is?

vext01: I don't know what you mean.

Do you mean paste the verbatim IR for each optimisation level into a different file?

What would be the value of such a specific test?

vext01: Note that this PR already introduces a`debug_debuginfo.c.O0` which does test the verbatim IR.

I just don't see the worth in testing the verbatim IR of approximate debuginfo for all opt levels. It will be a pain to update when things change.

ltratt: Mea culpa! You've already done what I wanted but I was too stupid to realise.

vext01: Ah, OK!

vext01:
The reason this was failing is that a PR introduced new tests in the meantime, and they were failing. Merged master in here (with merge commit) and fixed tests.

OK to squash?

ltratt:
Please squash.

vext01:
splat.

vext01:
Bear with me, we have a seg fault when testing with `--release`.

vext01:
Seg fault fixed. OK to re-squash?

ltratt:
Please squash.

vext01:
splat.

PR #412, opened 2021-08-18
Mapper: Ensure we are looking up an offset from the right object.

vext01:
The mapper uses `code_vaddr_to_offset()` to find the file offset of a
block. The resulting offset is then used to query the block map.

The code is incorrect because `code_vaddr_to_offset()` also succeeds on
externally linked code such as shared objects, yet the blockmap only
describes blocks in the main binary, thus sometimes we look up an offset
for the wrong file.

Most of the time this is harmless, as block offsets in different objects
rarely collide, and thus looking up a foreign offset in the blockmap
gives no result anyway. However, when experimenting with tracing the
Ruby interpreter, we did observe a collision, leading to
`find_code_sym()` to fail and in turn the JIT to crash.

The correct thing to do is to check that the offset in question comes
from the main binary and failing that report that there's external code
in the trace (return an empty vector).

Unfortunately this is very hard to devise a test for (we'd need objects
with colliding block addresses).

### DIFF
--- a/yktrace/src/hwt/mapper.rs
+++ b/yktrace/src/hwt/mapper.rs
@@ -212,6 +212,17 @@ impl HWTMapper {
         let block_vaddr = block.first_instr();
         let (obj_name, block_off) = code_vaddr_to_off(block_vaddr as usize).unwrap();
 
+        // Currently we only read in a block map and IR for the currently running binary (and not
+        // for dynamically linked shared objects). Thus, if we see code from another object, we
+        // can't map it.
+        //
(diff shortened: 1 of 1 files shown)


vext01:
I've just been thinking about this more. In the future it may be possible to examine the blockmaps of shared objects if they too were compiled with LTO and have IR embedded.

We could add a `FIXME` to that effect?

vext01:
> We could add a FIXME to that effect?

Like in 01d221bc18.

ltratt:
Please squash.

vext01:
splat.

PR #414, opened 2021-08-20
Handle indirect branches

vext01:
This comes in two stages. See individual commits.


ltratt:
s/brances/branches/ :)

vext01:
OK, will fix after review.

vext01:
> This doesn't parse (or make sense) to me: "the code which generates ... are then dead". Is this talking about the generated code or ... ?

So if we have an operand `%13`, then the "code which generates it" is instruction which defines the SSA variable `%13`, e.g. `%13  = blah`.

Can you think of a better wording?

ltratt:
> So if we have an operand %13, then the "code which generates it" is instruction which defines the SSA variable %13, e.g. %13 = blah.

Now I can't parse that :) What does "is instruction which defines" mean? I *think* what you might be saying is that we generate IR that we don't need?

vext01:
> The method overloading here seems a bit unnecessary? I assume that this function is called by external code, and then via overloading we call the method at line 245? Can we collapse them into one function or are both functions needed?

The methods are private.

I did it this was because:
 - It's not inconceivable that in the future we may want to do the same for multiple values at once, hence the variant accepting a vector.
 - Then, if you only want to delete one `Value` then it would be annoying to have to create a vector at each call site.

ltratt:
I think splitting things into two functions without knowing that we'll need either isn't a great idea: if we need to split it later, let's do so, but if we only need one function now, that will be easier to read.

vext01:
> Now I can't parse that :) What does "is instruction which defines" mean? I think what you might be saying is that we generate IR that we don't need?

I didn't expect this to be controversial :P

```
%1 = call f();
%2 = add(%1, 1);
```

`%1` and `%2` are both `Value`s and `Instruction`s at the same time.

The first instruction is the instruction that *defines* `%1`. The second instruction uses `%1` as an operand to `add`.

ltratt:
Right, so I think the original comment *might* be better written as something like:

> It isn't necessary to copy the indirect branch into the JITted code, as the successor block is known from the trace. Omitting the branch then means that the IR instruction which assigns to the `address` operand of the `indirectbr` is no longer needed and can be removed.

Is that still saying what you expected?

vext01:
Well it does, but it sounds clunky.

In the LLVM world, "definition" is the accpeted term, e.g. [here](https://llvm.org/docs/LangRef.html#well-formedness).

ltratt:
OK, so s/assigns/defines/ in my comment then?

vext01:
That should be all comments addressed, but I'd like Lukas to look at this too..

ptersilie:
Code looks good to me. I'll leave @ltratt to deal with the comments, since he's already started. :slightly_smiling_face:

ltratt:
"omitting it does mean that the code defines the `address` operand" doesn't parse. I also am not a fan of the "however" in that sentence, because it makes it sound like a *bad* thing, when it's actually a *good* thing!

vext01:
Ready to squash then?

ltratt:
I *think* we've agreed to change:

> However, omitting it does mean that the code defines the address operand of the indirectbr (and its dependencies) are then dead and can be removed.

to something like:

>  However, omitting it does mean that the code which defines the address operand of the indirectbr (and anything which recursively is related only to that operand) is dead and can be removed.

Sound like a plan?

vext01:
Yep. Try c3f395d.

vext01:
Those two commits should be good? I hope :P

[sorry, i messed up the first push]

ltratt:
Please squash.

vext01:
Splat.

PR #415, opened 2021-08-20
Deal with indirect function calls.

ptersilie:
This implements ways to deal with function calls, where the callee isn't
statically known, e.g.

%1 = @foo
call %1

Since at the time we trace the call instruction we can't tell what %1
is, we need to look ahead in the trace which function the tracer saw
next.

There's two scenarios we have to deal with: if we have IR for the passed
function, we can inline it as usual (note that in the future we don't
want to do this unless the function was promoted); if we don't have IR,
we can't inline and leave the call as is (we do however need to make
sure that the called function has been redeclared in the JITModule).


ptersilie:
Just realised that my description didn't make sense. Fixed in the PR comment, but will have to reword the first commit during rebase later.

ltratt:
Can we be consistent with terminology? We use (at least) "function pointers", "indirect calls", and "dynamic calls" in different places I think. FWIW, I think "indirect calls" is probably the best term.

ptersilie:
> Can we be consistent with terminology? We use (at least) "function pointers", "indirect calls", and "dynamic calls" in different places I think. FWIW, I think "indirect calls" is probably the best term.

Okay, will fix, some of it during rebasing.

ltratt:
> Okay, will fix, some of it during rebasing.

What would be fixed during rebasing?

ltratt:
> Function pointer is correct here.

But every call is to a function pointer. So there must surely be something different about this PR (although now you've got me wondering!).

vext01:
> But every call is to a function pointer.

A static call doesn't involve a function pointer.

[At the asm level, `call myfunc` (direct) vs. `call [rax]` (indirect via func ptr stored in a register)]

Or did I get the wrong end of the stick?

ltratt:
OK, so offline I think we've agreed that at least the first commit needs to be clear this is about "indirect calls" i.e. function calls to a dynamically determined address.

ptersilie:
The latest commits try to clean up the terminology a bit. Will fix the commit messages during rebasing.

ptersilie:
@ltratt Try that last commit.

vext01:
Some of my comments are unaddressed.

ptersilie:
Which ones? I think I've addressed all. Try refreshing a few times. ;-)

ptersilie:
Ready to squash?

vext01:
Go ahead.

ptersilie:
Squashed, rebased, and commit messages updated.

vext01:
Needs fmt.

ptersilie:
Fixed.

PR #417, opened 2021-08-23
Handle (non-debug) intrinsics.

vext01:
If a call to an intrinsic generates code, it is inlined. This means that
there are never IR blocks for a callee frame. In this case we can simply
continue with the current block.

Co-authored-by: Lukas Diekmann <lukas.diekmann@gmail.com>


ltratt:
Are we sure this is true for all intrinsics?

I also found the description hard to make sense of. I think it might better be something like "Intrinsics generate code *within* the current function rather than being turned into a call. If we see a 'call' to an intrinsic, we assume it will not lead to a call in the eventual machine code, and that it will simply be inlined within the current basic block. Note that this means a 'call' to an intrinsic does not necessarily or start or end a basic block." ?

vext01:
> Are we sure this is true for all intrinsics?

Lukas may have found a counter-example. Watch this space.

ltratt:
I think we might want to close this PR?

vext01:
I think so too. @ptersilie is that your understanding too?

ptersilie:
No why? This is still a fix we'll need even with all the codegen optimisations disabled.

ltratt:
Fair enough. But do we then know that all intrinsics always inline code?

ptersilie:
> Fair enough. But do we then know that all intrinsics always inline code?

Unfortunately, to that I don't currently have an answer.

ltratt:
What we *could* do if we're not entirely sure about *all* intrinsics is have a white-list where we only let through intrinsics that we know are inlined. But, yes, having a look at this would be great, thanks!

ltratt:
@vext01 I think this PR is probably dead?

vext01:
Probably, but it depends on if what @ptersilie is working on now works or not.

@ptersilie shall I close this?

ptersilie:
Yeah, let's close it. We can still look at it later for reference if needed.

PR #418, opened 2021-08-24
Refactor: group some fields into an abstracted type

vext01:
This is a refactoring that @ptersilie and I have superficially discussed in the past.

It adds a little more code than I expected, but I hope that the added safety aspects outweigh that cost?

See the first commit message for details.

PR #420, opened 2021-08-25
Group function address fields into a proper type.

vext01:
Similar to #418, this groups several related fields into a proper type.

Meat of the work in 6700865.

acac261 is optional. What do you think?


ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -75,6 +75,27 @@ class InputTrace {
...
+  map<string, uintptr_t> Map;
+
+public:
+  FuncAddrs(char **FuncNames, size_t *VAddrs, size_t Len) {
+    for (size_t I = 0; I < Len; I++) {
+      Map.insert(pair<char *, size_t>(FuncNames[I], VAddrs[I]));
+    }
+  }
+
+  // Lookup the address of the specified function name or return 0 on failure.
```
ltratt, line 90: If this is returning an address should it return `void */NULL` ?

vext01: Good question.

Although they are broadly equivalent, it may be better as a `void *`, as we eventually pass this do an LLVM API function which accepts `void *`.

PR #422, opened 2021-09-14
Hide thread tracer

vext01:
To aid the reviewer, I've put the main bulk of the work into the first commit. The commit message should explain the intent of the change.

The second commit is just adjustments to make tests pass again (and isn't all that interesting).

We can squash them together later.

We don't have (and can't easily make) tests for `ykrt` yet, so we will just have to eyeball those changes.


yktrace/src/lib.rs
```diff
@@ -12,6 +13,15 @@ mod hwt;
 pub use errors::InvalidTraceError;
 pub use hwt::mapper::BlockMap;
 
+thread_local! {
+    // When `Some` contains the `ThreadTracer` for the current thread. When `None`, the current
+    // thread is not being traced.
+    //
+    // We hide the `ThreadTracer` in a thread local (rather than returning it to the consumer of
+    // yk). This ensures that the `ThreadTracer` itself cannot appear in traces.
+    pub static THREAD_TRACER: RefCell<Option<ThreadTracer>> = RefCell::new(None);
```
bjorn3, line 22: There is a nightly feature that allows writing

```suggestion
    pub static THREAD_TRACER: RefCell<Option<ThreadTracer>> = const { RefCell::new(None) };
```

This makes TLS accesses cheaper as it doesn't have to check if it needs to lazily initialize it every time, instead evaluating the initializer at compile time and storing it's result in the `.tdata` section which libc copies to form the TLS storage for a thread.

ltratt:
A couple of doc comments, but otherwise this looks OK!

yktrace/src/lib.rs
vext01: Thanks! I've implemented that in eac6196.

vext01:
@ltratt your comments are addressed in 4800062.

ltratt:
Please squash.

vext01:
splat.

ptersilie:
Looks good.

vext01:
Ready to go @ltratt ?

PR #425, opened 2021-10-21
Implement support for the new control point design. 

ptersilie:
Depends on: https://github.com/ykjit/ykllvm/pull/7

5198c13 just disables tests so can mostly be ignored.

The new control point design replaces a call to a dummy control point
with a dynamically generated control point.

This new control point accepts and returns an "IO struct", whose members
are derived from the live state of the interpreter just before the
control point.

Generation of the control point is implemented in ykllvm. This change
implements the remainder of the logic (namely JIT runtime support).

Currently only two tests run. In order to get those working we had to
fix a bug relating to PHI nodes.

Fixes ykjit#394


c_tests/tests/simple_interp_loop.c
```diff
@@ -0,0 +1,79 @@
...
+#define NOP 0
+#define DEC 1
+#define RESTART_IF_NOT_ZERO 2
+#define EXIT 3
+
+int
+main(int argc, char **argv)
+{
+  // A hard-coded program to execute.
+  int prog[] = {0, 0, 1, 2, 0, 3};
```
ltratt, line 45: Do we need such a long "program"? I know that sounds silly, but my guess is that when we start doing optimisations and so on, weird things will break, and then we'll be grateful to have a test called `simple_interp_loop` be the simplest possible interpreter loop! Maybe, in fact, we should have a sequence of "gradually less simple" interpreter loops? e.g. one loop that just has a `DEC` instruction, one that has `RESTART_IF_NOT_ZERO`, and so on?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -160,6 +144,18 @@ class JITModBuilder {
   // Function virtual addresses discovered from the input trace.
   FuncAddrs FAddrs;
 
+  // Stack mapping the last executed block index for each frame. This is
```
ltratt, line 147: First sentence doesn't parse?

PR #433, opened 2021-11-04
Use intrinsic annotations to correctly assemble traces.

ptersilie:
Requires: https://github.com/ykjit/ykllvm/pull/10

Since LLVM may decide to entirely remove intrinsic calls, its impossible
to reliably map a PT trace back to LLVM IR. In other words, if an
intrinsic stays a call we see the same block twice, if it was inlined we
don't.

With our recent change that annotates intrinsic calls with inlining
decisions, we can make the right decision during trace construction:
e.g. if the intrinsic was inlined we don't need to expect a dpulicated
block and can just move on.


c_tests/tests/intrinsics.c
```diff
@@ -0,0 +1,46 @@
...
+int main(int argc, char **argv) {
+  int res = 0;
+  int loc = 0;
+  int i = 3;
+  NOOPT_VAL(res);
+  NOOPT_VAL(i);
+  NOOPT_VAL(loc);
+  while (i > 0) {
+    yk_control_point(loc);
+    memcpy(&res, &i, 4);
```
vext01, line 39: So this test fails before, correct?

ptersilie: It does yes. But only on `O0`, since the other opt levels remove `memcpy` at the LLVM IR level.

c_tests/tests/intrinsics.c
```diff
@@ -0,0 +1,46 @@
+// Compiler:
+//   env-var: YKD_PRINT_JITSTATE=1
+// Run-time:
+//   env-var: YKD_PRINT_IR=jit-pre-opt
+//   stderr:
```
ltratt, line 5: Would I be an idiot if I said I couldn't see anything about `memcpy` in the test?

ptersilie: You wouldn't. But since `memcpy` is being inlined and we don't have negative matches yet to check that it doesn't appear, there's not much we can test here, unfortunately.

ltratt: Aha, makes sense.

ptersilie:
Ready for re-review.

ltratt:
If @vext01 is OK with this, then I think we're ready for squashing.

c_tests/tests/intrinsics.c
vext01: I was about to say: why don't we check for the presence of inlining metadata, but that wouldn't be portable... so, I agree, we can't do much more.

vext01:
LGTM please squash.

ptersilie:
Squashed. We'll have to wait for https://github.com/ykjit/ykllvm/pull/10 though before merging.

vext01:
```
/home/buildbot-worker3/worker3/yk-buildbot-build-script-hw/build/c_tests/tests/intrinsics.c:38:22: error: incompatible integer to pointer conversion passing 'int' to parameter of type 'YkLocation *' [-Werror,-Wint-conversion]
    yk_control_point(loc);
                     ^~~
/home/buildbot-worker3/worker3/yk-buildbot-build-script-hw/build/c_tests/../ykcapi/yk.h:41:35: note: passing argument to parameter here
void yk_control_point(YkLocation *);
```

This is one of my recent changes. Use `yk_location_new()` to get a new location and pass that to the control point instead.

ptersilie:
Fixed in 79e7150. Okay to squash?

ltratt:
Please squash.

ptersilie:
Squashed.

PR #436, opened 2021-11-18
Implement aborting guards for simple branch instructions.

vext01:
When the trace compiler encounters an LLVM `br` instruction in a trace,
it now emits a guard with the appropriate check.

For now a failing guard simply branches to a block which aborts the
program (eventually a guard failure should invoke a stop-gap
interpreter, but we are not ready for that yet).

This of course now means that the `JITFunc` is no longer one linear
block, but that may change in the future, depending on how we decide to
transfer to the stop-gap interpreter.

Without implementing trace stitching, there is currently no way for a
test to exit with success, so for now all of the (unskipped) tests now
expect to crash.

I've strengthened the existing tests, including checking for emission of
the guard blocks.

I also had to change the lang test name matchers so that we could match
a basicblock *value* (prefixed with `%`) to its *label* (without the
prefix).

Soon we should implement guards for LLVM `switch` and `indirectbr`
instructions too, but those should come in future PRs.


c_tests/tests/simple.c
```diff
@@ -36,16 +53,17 @@ void _yk_test(int i, int res) {
...
+  int i = 5;
   NOOPT_VAL(res);
   NOOPT_VAL(i);
   while (i > 0) {
     yk_control_point(loc);
+    fprintf(stderr, "i=%d\n", i);
     res += 2;
     i--;
   }
+  // FIXME: unreachable due to aborting guard failure earlier.
```
ltratt, line 65: Instead of `FIXME` should we put some sort of abort / `err` / whatever in to enforce that?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -332,6 +373,52 @@ class JITModBuilder {
     }
   }
 
+  // Given an `IRBlock`, find and return the LLVM data structures for the basic
+  // block and its parent function.
+  std::pair<Function *, BasicBlock *> getLLVMAOTFuncAndBlock(IRBlock *IB) {
+    Function *F = AOTMod->getFunction(IB->FuncName);
+    if (!F)
```
ltratt, line 380: This looks more like an `assert` in disguise to me?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -332,6 +373,52 @@ class JITModBuilder {
...
+          LibcErrxFuncTy, GlobalValue::ExternalLinkage, "errx", JITMod);
+
+      // Create the block.
+      GuardFailBB = BasicBlock::Create(Context, "guardfail", JITFunc);
+      IRBuilder<> FailBuilder(GuardFailBB);
+      Value *ExitCode = ConstantInt::get(IntTy, EXIT_SUCCESS);
+      Value *Fmt =
+          FailBuilder.CreateGlobalStringPtr(StringRef("guard-failure"));
+      FailBuilder.CreateCall(LibcErrxFuncTy, LibcErrxFunc, {ExitCode, Fmt});
+      FailBuilder.CreateUnreachable(); // `errx(3)` in non-convergent.
```
ltratt, line 417: I can't parse the comment.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -461,20 +550,14 @@ class JITModBuilder {
       assert(MaybeIB.hasValue());
       IRBlock IB = MaybeIB.getValue();
 
-      // Get a traced function so we can extract blocks from it.
-      Function *F = AOTMod->getFunction(IB.FuncName);
-      if (!F)
-        errx(EXIT_FAILURE, "can't find function %s", IB.FuncName);
+      Function *F;
+      BasicBlock *BB;
+      std::tie(F, BB) = getLLVMAOTFuncAndBlock(&IB);
```
ltratt, line 555: What does `std::tie` do here?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -559,11 +642,29 @@ class JITModBuilder {
           continue;
         }
 
-        if ((isa<BranchInst>(I)) || isa<SwitchInst>(I)) {
-          // FIXME Replace all potential CFG divergence with guards.
-          continue;
+        if (isa<BranchInst>(I)) {
+          // Should not be an unmappable hole in the trace after a branch.
```
ltratt, line 646: Do we know that `Idx + 1` is always within bounds?

ykllvmwrap/src/jitmodbuilder.cc
vext01: it unpacks a tuple.

(and no, you can't put the type inside the tie, I tried).

This was made more concise in c++17, but we don't target it, I don't think.

ltratt: Got it. [I have no preferences on C++ versions. I'm fine with whatever.]

ykllvmwrap/src/jitmodbuilder.cc
vext01: The bounds are checked with an assertion [here](https://github.com/vext01/yk_pv/blob/8b1144172b3123dff939624b1cbfa92700fe0b32/ykllvmwrap/src/jitmodbuilder.cc#L68).

ykllvmwrap/src/jitmodbuilder.cc
vext01: Alas:
```
  cargo:warning=src/jitmodbuilder.cc:552:12: error: decomposition declarations are a C++17 extension [-Werror,-Wc++17-extensions]
  cargo:warning=      auto [F, BB] = getLLVMAOTFuncAndBlock(&IB);
  cargo:warning=           ^~~~~~~
  cargo:warning=1 error generated.
  exit status: 1
```

I don't fancy changing the C++ standard, so let's just roll with it.

c_tests/tests/simple.c
vext01: I've added an abort, but kept the `FIXME` as justification.

In 1a67a96.

ykllvmwrap/src/jitmodbuilder.cc
vext01: Fixed in 1a67a96.

ykllvmwrap/src/jitmodbuilder.cc
vext01: Fixed annoying typo in 1a67a96.

vext01:
I think that's all comments addressed.

c_tests/tests/simple.c
ltratt: Good idea!

ykllvmwrap/src/jitmodbuilder.cc
ltratt: OK.

ykllvmwrap/src/jitmodbuilder.cc
ltratt: OK, so that `assert` rules out memory unsafety, but I think my question stands as an issue of correctness (as it would in e.g. Python): do we know for sure that there's always a `Idx + 1`?

vext01: There must be another block in the trace following a branch, yes.

ltratt: OK.

ltratt:
Please squash.

vext01:
splat.

PR #437, opened 2021-11-19
Implement guards for LLVM `switch` statements.

vext01:
This is similar to the way be do guards for regular `br` instructions and I did a fair amount of refactoring to allow for code-sharing.

### DIFF
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -258,14 +258,16 @@ class JITModBuilder {
     }
   }
 
-  void handleBranchInst(Function *JITFunc, BasicBlock *NextBlock,
-                        Instruction *I) {
+  // Emits a guard for a LLVM `br` instruction, returning a pointer to the
+  // guard success block, or null if no guard was required.
+  BasicBlock *handleBranchInst(Function *JITFunc, BasicBlock *NextBlock,
+                               Instruction *I) {
     assert(isa<BranchInst>(I));
     BranchInst *BI = cast<BranchInst>(I);
     LLVMContext &Context = JITMod->getContext();
 
     if (BI->isUnconditional())
-      return; // Control-flow can not diverge. No guard required.
+      return nullptr; // Control-flow can not diverge. No guard required.
 
     // A conditional branch should have two successors and one of them should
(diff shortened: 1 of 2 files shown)


ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -701,6 +699,49 @@ class JITModBuilder {
     return JITMod;
   }
 
+  void handleBranchingControlFlow(Instruction *I, size_t TraceIdx,
+                                  Function *JITFunc) {
+    // Fisrt, peek ahead in the trace and retrieve the next block. We need this
+    // so that we can insert an appropriate guard into the trace.
+    assert(InpTrace[TraceIdx + 1].hasValue()); // Should be a mappable block.
```
ltratt, line 706: I think this needs to be preceded by an assert that either: checks there is a `TraceIdx + 1`; checks that `TraceIdx` is an element which, by definition, has a successor.

vext01: `InpTrace` has an overloaded `[]` operator that does that assertion.

We [discussed this in the last PR](https://github.com/ykjit/yk/pull/436#pullrequestreview-810106747). Same code, just moved.

ltratt: > InpTrace has an overloaded [] operator that does that assertion.

I don't think that's good enough: if this was Python I'd be asking for the same assertion, even though `[]` in Python is also memory safe. The point is that as a reader of this code I can't tell that `Idx + 1` exists from just looking at this, so I waste my limited mental capacity (very limited mental capacity :p) wondering about this -- I'm not going to know that the overloaded `[]` ensures it.

vext01: Another block has to exist after a branch, and tracing cannot be stopped by a branch.

We can add a comment to this effect if you like.

ltratt: A brief comment would be fine!

vext01: Try 458fae6 (which also fixes a typo).

ltratt:
Please squash.

vext01:
splat.

PR #445, opened 2021-11-29
Update tests so that the control point takes a `Location`.

vext01:
This also exposed a bug where we failed to map a value if it was defined
by a Phi node from a prior block. Fixed in this change.

Companion to https://github.com/ykjit/ykllvm/pull/13 -- Companion PR merges first.

### DIFF
--- a/c_tests/tests/simple_interp_loop2.c
+++ b/c_tests/tests/simple_interp_loop2.c
@@ -62,17 +62,24 @@ int main(int argc, char **argv) {
   int prog[] = {0, 0, 1, 2, 0, 3};
   size_t prog_len = sizeof(prog) / sizeof(prog[0]);
 
-  // The program counter (FIXME: also serving as a location ID for now).
+  // Create one location for each potential PC value.
+  YkLocation locs[prog_len];
+  for (int i = 0; i < prog_len; i++)
+    locs[i] = yk_location_new();
+
+  // The program counter.
   int pc = 0;
 
   NOOPT_VAL(pc);
   NOOPT_VAL(prog);
   NOOPT_VAL(prog_len);
   NOOPT_VAL(mem);
+  NOOPT_VAL(locs);
 
   // interpreter loop.
   while (true) {
-    yk_control_point(pc);
+    YkLocation *loc = &locs[pc];
+    yk_control_point(loc);
     assert(pc < prog_len);
     int bc = prog[pc];
     fprintf(stderr, "pc=%d, mem=%d\n", pc, mem);
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -366,7 +366,7 @@ class JITModBuilder {
   // Delete the dead value `V` from its parent, also deleting any dependencies
   // of `V` (i.e. operands) which then become dead.
   void deleteDeadTransitive(Value *V) {
-    assert(V->user_empty()); // The value should be dead.
+    assert(V->user_empty()); // The value must be dead.
     vector<Value *> Work;
     Work.push_back(V);
     while (!Work.empty()) {
@@ -623,6 +623,15 @@ class JITModBuilder {
     }
   }
 
+  // Determines if the LLVM values `V1` and `V2` are instructions defined
+  // within the same LLVM `BasicBlock`. `V1` and `V2` must both be an instance
+  // of `Instruction`.
+  bool areInstrsDefinedInSameBlock(Value *V1, Value *V2) {
+    assert(isa<Instruction>(V1) && isa<Instruction>(V2));
+    return cast<Instruction>(V1)->getParent() ==
+           cast<Instruction>(V2)->getParent();
+  }
+
 public:
   // Store virtual addresses for called functions.
   std::map<GlobalValue *, void *> GlobalMappings;
@@ -660,8 +669,9 @@ public:
     // control point and finish above it. This means that alloca'd variables
     // become undefined (as they are defined outside of the trace) and thus
     // need to be remapped to the input of the compiled trace. SSA values
-    // remain correct as phi nodes at the beginning of the trace automatically
-    // select the appropriate input value.
+    // (from the same block as the control point) remain correct as phi nodes
+    // at the beginning of the trace automatically select the appropriate input
+    // value.
     //
     // For example, once patched, a typical interpreter loop will look like
     // this:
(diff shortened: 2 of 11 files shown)


ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -623,6 +623,16 @@ class JITModBuilder {
     }
   }
 
+  // Determines if the LLVM values `V1` and `V2` are instructions defined
+  // within the same LLVM `BasicBlock`.
+  bool isDefinedInSameBlock(Value *V1, Value *V2) {
+    if (!isa<Instruction>(V1) || !isa<Instruction>(V2))
```
ltratt, line 629: I wonder if we should say that this function only accepts `Instruction`s and this can then be an `assert`?

vext01: It's designed this way so that the caller can determine if two values are instructions in the same block without having to determine if either value is a non-instruction type (e.g. a global or a constant) at each call site.

ltratt: I understand what it *can* do: I'm suggesting it might be doing too much and that it might not be useful to allow the user to check if non-instructions are in the same module (on the basis that we won't use that use case and thus if it ever happens it's a bug that won't be caught by `return false`). I'm happy to be wrong and for you to tell me that it does need to support non-instructions though!

vext01: After experimentation, what I don't like about pulling the instruction check logic outside of the function is that you end up duplicating the map insertion and that junks up the call site.

```diff
diff --git a/ykllvmwrap/src/jitmodbuilder.cc b/ykllvmwrap/src/jitmodbuilder.cc
index 3fd3661..d9114ed 100644
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -625,11 +625,7 @@ class JITModBuilder {
 
   // Determines if the LLVM values `V1` and `V2` are instructions defined
   // within the same LLVM `BasicBlock`.
-  bool isDefinedInSameBlock(Value *V1, Value *V2) {
-    if (!isa<Instruction>(V1) || !isa<Instruction>(V2))
-      return false;
-    Instruction *I1 = cast<Instruction>(V1);
-    Instruction *I2 = cast<Instruction>(V2);
+  bool isDefinedInSameBlock(Instruction *I1, Instruction *I2) {
     return I1->getParent() == I2->getParent();
   }
 
@@ -737,10 +733,13 @@ public:
       //  We need an entry in this map for any live variable that isn't defined
       //  by a PHI node at the top of the block cotaining the call to the
       //  control point.
-      if (!(isa<PHINode>(InsertedOperand) &&
-            isDefinedInSameBlock(InsertedOperand, IVI))) {
-        InsertValueMap[*IVI->idx_begin()] = IVI->getInsertedValueOperand();
-      }
+                       if (isa<Instruction>(InsertedOperand)) {
+                               Instruction *InsertedOperandInst = cast<Instruction>(InsertedOperand);
+                               if (!(isDefinedInSameBlock(InsertedOperandInst, IVI) && isa<PHINode>(InsertedOperandInst)))
+                                       InsertValueMap[*IVI->idx_begin()] = IVI->getInsertedValueOperand();
+                       } else
+                               InsertValueMap[*IVI->idx_begin()] = IVI->getInsertedValueOperand();
+
       OutS = IVI->getAggregateOperand();
     }
```

ltratt: I don't think I've got my point across very well -- sorry!

What I'm asking is: do we *ever* ask whether a non-`Instruction` is part of a module? My guess based on my understanding of the code is "no".

vext01: Currently we don't (based on current test coverage), but I don't see why we couldn't.

`InsertedOperandInst` could easily be a global or a constant. Neither of which are an `Instruction` that can be parented by a `BasicBlock` according to the LLVM type system.

ltratt: I understand it *can* be those, but based on the current code I think `return false` will lead to bugs which `assert(isInstruction(value))`.

vext01: I don't follow. How could it lead to bugs? Can you give an example?

c_tests/tests/simple_interp_loop2.c
```diff
@@ -1,7 +1,7 @@
 // Compiler:
 //   env-var: YKD_PRINT_JITSTATE=1
 // Run-time:
-//   env-var: YKD_PRINT_IR=jit-pre-opt
+//   env-var: YKD_PRINT_IR=jit-pre-opt,aot
```
ptersilie, line 4: aot left in?

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Isn't a `PHINode` always an `Instruction`?

vext01: Yes, but it's the other operand that I'm not sure about.

vext01: @ltratt try 0b8e03d4d3bc69885366423b519e15b9ba38f244.

I was hoping to prove you wrong with tests, but failed. Thus I've changed the name of the function and changed the `return false` case to an assertion.

Nothing is set in stone, and we can change this if a test later proves us wrong.

ptersilie: What I meant was that we only ever call `isDefinedInSameBlock` on `PHINode`s, so an assert would be fine there.

ltratt: Works for me!

vext01: > What I meant was that we only ever call isDefinedInSameBlock on PHINodes, so an assert would be fine there.

Unless I'm mistaken, the operands we pass to `areInstsDefinedInSameBlock()` are: `IVI->getInsertedOperand()` which might be a PHI; and `IVI` which is never a PHI.

So we call the function on "maybe PHI", but (we think) "always instruction" values.

vext01:
OK to squash?

vext01:
@ptersilie you ok with this?

ltratt:
Let's move on with this: please squash.

vext01:
splat.

PR #447, opened 2021-12-06
Move most of the control point logic into to Rust code.

vext01:
The second commit is the meat of this PR.

This is a companion to https://github.com/ykjit/ykllvm/pull/15. That PR will be merged first.


vext01:
I'd like to emphasize this passage from the second commit message:

> Some of the design introduced in this change is pretty questionable, e.g. namely storing pointers in Rust globals using `usize`. I plan for this to be eliminated when we properly hook in `transition_location()`, but I feel it would be foolish to not use this interim PR as a stepping stone.

Please don't spend too much time on reviewing those parts of the PR. They will change soon.

ykrt/src/mt.rs
```diff
@@ -139,9 +169,78 @@ impl MT {
         }
     }
 
-    /// `Location`s represent a statemachine: this function transitions to the next state (which
-    /// may be the same as the previous state!). If this results in a compiled trace, it returns
-    /// `Some(pointer_to_trace_function)`.
+    pub fn set_loc_code_ptr(loc: *mut Location, code: *const c_void) {
```
ptersilie, line 172: What does this do? Might benefit from a doc comment.

ykrt/src/mt.rs
```diff
@@ -139,9 +169,78 @@ impl MT {
...
+        let js = &mut mt.inner.jit_state.lock().unwrap();
+        js.trace_code = code as usize;
+    }
+
+    /// FIXME: This is an interim version of `transition_location()` with (over-)simplified logic.
+    /// The state machines of `Location`s isn't properly hooked in, but this is the proposed
+    /// interface.
+    ///
+    /// This function is called once upon each execution of the control point. The internal state
+    /// of the supplied `Location` (if any) may be mutated, maybe moving the `Location` to a
```
ptersilie, line 183: this doesn't really parse after `maybe moving`. Is it: "for example by moving the Location to a different state than its state machine"?

ykrt/src/mt.rs
```diff
@@ -139,9 +169,78 @@ impl MT {
...
+    /// different state its state machine.
+    ///
+    /// The returned `usize` signals what action the control point should take next (if any). The
+    /// potential values for this value are as follow:
```
ptersilie, line 187: `the potential values are:`

ptersilie:
Typo in commit: "e.g. ~~namely~~ storing pointers".

ykrt/src/mt.rs
vext01: I'll add a comment.

It's needed because when the control point calls `__yktrace_irtrace_compile()` a pointer to JIT code is returned, and we need some way to store that (eventually into the location, but for now) into the global variable that stores the trace pointer.

This isn't very optimal, but I wanted to get something simple working first.

vext01:
> Typo in commit: "e.g. namely storing pointers".

Will fix during squash.

OK to squash?

ltratt:
> One of the tests was kind-of-failing with:
[...]
> It printed this on stderr if you ran with `--nocapture` but didn't
> actually fail the test (a quirk of langtester?).

Huh? Does that mean the command returned a non-zero exit code? If so, that sounds like a lang_tester bug!

ykrt/src/mt.rs
```diff
@@ -139,9 +169,78 @@ impl MT {
...
+    /// This encoding scheme assumes that the allocator used to obtain heap memory for JITted code
+    /// will align allocations with at least two zero-valued least-significant bits. In other
+    /// words, we assume (and assert) that the `JITACTION_*` values are invalid aligned pointer
+    /// values.
+    ///
+    /// If a null `Location` pointer is passed, then this signals to the JIT that a loop cannot
+    /// start at this point in the interpreter. In this case the function always returns
+    /// `JITACTION_NOP`.
+    pub fn transition_location_simple(loc: *mut Location) -> usize {
+        if loc != ptr::null_mut() {
```
ltratt, line 213: It might be clearer (and use less indentation!) to start this function with:

```rust
if loc == ptr::null_mut() {
  return JITACTION_NOP;
}
```

yktrace/src/hwt/mapper.rs
```diff
@@ -108,6 +108,61 @@ impl HWTMapper {
...
+    ///
+    /// (where `bbN` is a mappable block, `ctrlp` is a block in the control point, and `?` is
+    /// unmappable blocks), then this function would canonicalise the trace to:
+    ///
+    /// ```ignore
+    /// [bb1, bb2, bb3, ctrlp, bb4]
+    /// ```
+    fn canonicalise_control_point(irblocks: &mut Vec<IRBlock>) {
+        let ctrlp_sym_c = CString::new(CONTROL_POINT_SYM).unwrap();
+        assert!(irblocks[0].func_name == ctrlp_sym_c);
```
ltratt, line 129: Should these be `debug_assert`s?

yktrace/src/hwt/mapper.rs
```diff
@@ -108,6 +108,61 @@ impl HWTMapper {
...
+        let mut in_ctrlp = true;
+        let mut ctrlp_start_idx = Some(0);
+        let mut delete_ranges = Vec::new();
+        for (ib_idx, ib) in &mut irblocks.iter().enumerate() {
+            if in_ctrlp {
+                // If we encounter an unmappable hole, or another block from CONTROL_POINT_SYM,
```
ltratt, line 135: I find this sentence hard to match up with the following `if` condition (which triggers only if we haven't encountered an unmappable hole).

yktrace/src/hwt/mapper.rs
```diff
@@ -108,6 +108,61 @@ impl HWTMapper {
...
+                // then we remain inside the control point. Otherwise we have just exited the
+                // control point.
+                if !(ib.is_unmappable() || ib.func_name == ctrlp_sym_c) {
```
ltratt, line 138: Might this be clearer as:

```rust
if !ib.is_unmappable() || ib.func_name != ctrlp_sym_c { ... }
```

yktrace/src/hwt/mapper.rs
```diff
@@ -108,6 +108,61 @@ impl HWTMapper {
...
+    /// unmappable blocks), then this function would canonicalise the trace to:
+    ///
+    /// ```ignore
+    /// [bb1, bb2, bb3, ctrlp, bb4]
+    /// ```
+    fn canonicalise_control_point(irblocks: &mut Vec<IRBlock>) {
+        let ctrlp_sym_c = CString::new(CONTROL_POINT_SYM).unwrap();
+        assert!(irblocks[0].func_name == ctrlp_sym_c);
+        let mut in_ctrlp = true;
+        let mut ctrlp_start_idx = Some(0);
```
ltratt, line 131: This doesn't need to be an `Option` AFAICS: we always assign `Some(...)` to it?

yktrace/src/hwt/mapper.rs
```diff
@@ -108,6 +108,61 @@ impl HWTMapper {
...
+        let mut delete_ranges = Vec::new();
```
ltratt, line 132: Reading the code below it feels like it might be simpler to copy the "keep" things rather than later deleting the "don't keep" things? I might be wrong!

vext01:
> Huh? Does that mean the command returned a non-zero exit code? If so, that sounds like a lang_tester bug!

The lang_tester stderr specification was invalid, so I don't think we even got as far as compiling / running the test.

I must admit, I didn't look deeply, but I can do after this PR, if you like.

ltratt:
> The lang_tester stderr specification was invalid, so I don't think we even got as far as compiling / running the test.

OK. I think we'll need to update the commit message to be less confusing, because as written, it sounds like a lang_tester bug!

yktrace/src/hwt/mapper.rs
vext01: I was trying to do it copy-free, but if you think it's not worthwhile, I could probably simplify the code a bit.

vext01:
> it sounds like a lang_tester bug

it may well be. I don't know without looking deeper.

ltratt:
>>  it sounds like a lang_tester bug
> it may well be. I don't know without looking deeper.

I think you should quickly investigate before we go further with this PR. If there's a lang_tester bug, we've got many problems! If it's not a lang_tester bug, then we can fix the commit message in this PR to be more accurate.

vext01:
About the lang_tester thing, if I make one of the examples invalid, with the following diff:

```diff
diff --git a/examples/fm_options/lang_tests/simple.py b/examples/fm_options/lang_tests/simple.py
index fce8fde..995f27f 100644
--- a/examples/fm_options/lang_tests/simple.py
+++ b/examples/fm_options/lang_tests/simple.py
@@ -3,7 +3,7 @@
 #   stdout:
 #     $1
 #     b
-#     $1
+#     ...$1
```

And run the example:
```
$ cargo run --example fm_options
...
running 3 tests
test lang_tests::nondeterministic.py ... ok
test lang_tests::echo.py ... okthread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: "Can't mix name matching with wildcards at start of line 3."', src/tester.rs:718:47


test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

So it appears that an invalid test spec, means that the test passes, and this is a lang_tester bug?

ykrt/src/mt.rs
vext01: Fixed in f9c355c.

yktrace/src/hwt/mapper.rs
vext01: Yep. Fixed in f9c355c.

yktrace/src/hwt/mapper.rs
vext01: Try the change in f9c355c.

yktrace/src/hwt/mapper.rs
vext01: Yes, but you meant `&&` I think?

[One of the few things I remember from college education: [De Morgan's Laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) :laughing:]

vext01: Fixed in f9c355c.

yktrace/src/hwt/mapper.rs
vext01: Fixed in f9c355c.

vext01:
I believe all that is outstanding now is the question of whether to re-write `canonicalise_control_point()`.

ltratt:
> So it appears that an invalid test spec, means that the test passes, and this is a lang_tester bug?

Yes, this is a lang_tester flaw.

yktrace/src/hwt/mapper.rs
ltratt: Deleting a vec's elements does do copying. However, I don't feel strongly: you should write this whichever you way you think is clearer.

vext01: Ah, yes, shuffling the elements down...

Let's try a rewrite to see if it looks cleaner.

ltratt:
@vext01 https://github.com/softdevteam/lang_tester/pull/98 should fix the lang_tester issue you saw. It works for me locally, but it would be great if you can replicate it at your end too!

yktrace/src/hwt/mapper.rs
vext01: 8470f3b7 (and 79b6ef67) is the most concise I managed to get it. What do you think?

[It seems tempting to chain the first two `if`s but you end up in a pickle it turns out]

ltratt: Works for me!

ykrt/src/mt.rs
ltratt: @vext01 Has this been addressed?

ykrt/src/mt.rs
vext01: I interpret Lukas' silence as, yes, maybe?

ykrt/src/mt.rs
vext01: I believe so.

vext01:
I mustn't forget to update commit messages at squash time.

ykrt/src/mt.rs
ltratt: Can you point out which commit added a comment?

ptersilie: It has been addressed!

vext01: Appears I lied! Sorry. Try 63a9f67.

ykrt/src/mt.rs
vext01: This one is addressed in here: https://github.com/ykjit/yk/pull/447/commits/c2415f256c08431217aed5f7effc83cfc69d10eb#diff-9fa85f3ab0ab08c2bb4337dd17d47605f520c2780158e50e66eff7cd9dc1b29bR183

ykrt/src/mt.rs
vext01: Sorry, looks like I commented in the wrong place before.

This one is fixed here: https://github.com/ykjit/yk/pull/447/commits/c2415f256c08431217aed5f7effc83cfc69d10eb#diff-9fa85f3ab0ab08c2bb4337dd17d47605f520c2780158e50e66eff7cd9dc1b29bR187

vext01:
I *think* that's everything?

ltratt:
Please squash.

vext01:
splat and updated commit messages.

ltratt:
The commit message for https://github.com/ykjit/yk/pull/447/commits/af00e8724f6e8334839574b9eb211740b95e539e looks like it's merged "throw away" commit messages in?

ltratt:
Please force push an update.

vext01:
Fixed. I suppose that commit should stay separate?

ltratt:
Agreed!

bors r+

PR #451, opened 2021-12-15
Add parser for LLVM stackmaps.

ptersilie:
This commit implements a parser for LLVM stackmaps which we will later
use to initialise a stopgap interpreter during a guard failure. Unlike
other standard stackmap parsers, this one doesn't create a 1:1
representation of the stackmap as it contains more data than we care
about (for example things like function/const/record counts whose only
purpose is to aid parsing in the stackmap and serve no other purpose
otherwise). Instead, the chosen representation makes it easier to query
the data during a guard failure, e.g. values referred by `ConstIndex`
are folded into `Locations` and records can be retrieved via the return
address of the `__llvm_deoptimize` call, which is calculated during
parsing.

I'm cautiously marking this as a draft as I'm aware of a few things that might be contentious which I'm hoping to get some opinions on (e.g. my use of allow dead code which I only did because I wasn't sure if this data would be needed later on). But the more I think about it, there's probably no need to store `Record`s at all, and instead have `SMQuery.map` directly store a vector of `Location`s, as we don't really care about the other data. But let's see what you guys think first.

Once this PR is merged there'll be a follow up PR which hooks this stuff into our guards.

### DIFF
--- a/yksmp/.gitignore
+++ b/yksmp/.gitignore
@@ -1,0 +1,2 @@
+Cargo.lock
+target/
--- a/yksmp/Cargo.toml
+++ b/yksmp/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "yksmp"
+version = "0.1.0"
+edition = "2018"
+authors = ["The Yorick Developers"]
+license = "Apache-2.0 OR MIT"
+
+[dependencies]
+object = "0.28.1"
--- a/yksmp/src/lib.rs
+++ b/yksmp/src/lib.rs
@@ -1,0 +1,342 @@
+//! Note that LLVM currently only supports stackmaps for 64 bit architectures. Once they support
+//! others we will need to either make this parser more dynamic or create a new one for each
+//! architecture.
+#[cfg(not(target_arch = "x86_64"))]
+compile_error!("The stackmap parser currently only supports x86_64.");
+
+use std::collections::HashMap;
+use std::convert::TryFrom;
+use std::convert::TryInto;
+use std::error;
+use std::ffi::c_void;
+use std::{ptr, slice};
+
+struct Function {
+    addr: u64,
+    record_count: u64,
+}
+
+struct Record {
+    offset: u32,
+    locs: Vec<Location>,
+}
+
+#[derive(Debug)]
+enum Location {
+    Register(u16),
+    Direct(u16, i32),
+    Indirect(u16, i32),
+    Constant(u32),
+    LargeConstant(u64),
+}
+
+/// Parses LLVM stackmaps version 3 from a given address. Provides a way to query relevant
+/// locations given the return address of a `__llvm_deoptimize` function.
+struct StackMapParser<'a> {
+    data: &'a [u8],
+    offset: usize,
+}
+
+impl StackMapParser<'_> {
+    pub fn parse(data: &[u8]) -> Result<HashMap<u64, Vec<Location>>, Box<dyn error::Error>> {
+        let mut smp = StackMapParser { data, offset: 0 };
+        smp.read()
+    }
+
+    fn read(&mut self) -> Result<HashMap<u64, Vec<Location>>, Box<dyn error::Error>> {
+        // Read version number.
+        if self.read_u8() != 3 {
+            return Err("Only stackmap format version 3 is supported.".into());
+        }
+
+        // Reserved
+        assert_eq!(self.read_u8(), 0);
+        assert_eq!(self.read_u16(), 0);
+
+        let num_funcs = self.read_u32();
+        let num_consts = self.read_u32();
+        let num_recs = self.read_u32();
+
+        let funcs = self.read_functions(num_funcs);
+        let consts = self.read_consts(num_consts);
+
+        let mut map = HashMap::new();
+
+        // Check that the records match the sum of the expected records per function.
+        assert_eq!(
+            funcs.iter().map(|f| f.record_count).sum::<u64>(),
+            u64::from(num_recs)
+        );
+
+        // Parse records.
+        for f in funcs {
+            let records = self.read_records(f.record_count, &consts);
+            for r in records {
+                let key = f.addr + u64::from(r.offset);
+                map.insert(key, r.locs);
+            }
+        }
+
+        Ok(map)
+    }
+
+    fn read_functions(&mut self, num: u32) -> Vec<Function> {
+        let mut v = Vec::new();
+        for _ in 0..num {
+            let addr = self.read_u64();
+            let _stack_size = self.read_u64();
+            let record_count = self.read_u64();
+            v.push(Function { addr, record_count });
+        }
+        v
+    }
+
+    fn read_consts(&mut self, num: u32) -> Vec<u64> {
+        let mut v = Vec::new();
+        for _ in 0..num {
+            v.push(self.read_u64());
+        }
+        v
+    }
+
+    fn read_records(&mut self, num: u64, consts: &Vec<u64>) -> Vec<Record> {
+        let mut v = Vec::new();
+        for _ in 0..num {
+            let _id = self.read_u64();
+            let offset = self.read_u32();
+            self.read_u16();
+            let num_locs = self.read_u16();
+            let locs = self.read_locations(num_locs, consts);
+            // Padding
+            self.align_8();
+            self.read_u16();
+            let num_liveouts = self.read_u16();
+            self.read_liveouts(num_liveouts);
+            self.align_8();
+            v.push(Record { offset, locs });
+        }
+        v
+    }
+
+    fn read_locations(&mut self, num: u16, consts: &Vec<u64>) -> Vec<Location> {
+        let mut v = Vec::new();
+        for _ in 0..num {
+            let kind = self.read_u8();
+            self.read_u8();
+            let _size = self.read_u16();
+            let dwreg = self.read_u16();
+            self.read_u16();
+
+            let location = match kind {
+                0x01 => {
+                    self.read_i32();
+                    Location::Register(dwreg)
+                }
+                0x02 => {
+                    let offset = self.read_i32();
+                    Location::Direct(dwreg, offset)
+                }
+                0x03 => {
+                    let offset = self.read_i32();
+                    Location::Indirect(dwreg, offset)
+                }
+                0x04 => {
+                    let offset = self.read_u32();
+                    Location::Constant(offset)
+                }
+                0x05 => {
+                    let offset = self.read_i32();
+                    Location::LargeConstant(consts[usize::try_from(offset).unwrap()])
+                }
+                _ => unreachable!(),
+            };
+
+            v.push(location)
+        }
+        v
+    }
+
+    fn read_liveouts(&mut self, num: u16) {
+        for _ in 0..num {
+            let _dwreg = self.read_u16();
+            let _size = self.read_u8();
+        }
+    }
+
+    fn align_8(&mut self) {
+        self.offset += (8 - (self.offset % 8)) % 8;
+    }
+
+    fn read_u8(&mut self) -> u8 {
+        let d = u8::from_ne_bytes(self.data[self.offset..self.offset + 1].try_into().unwrap());
+        self.offset += 1;
+        d
+    }
+
+    fn read_u16(&mut self) -> u16 {
+        let d = u16::from_ne_bytes(self.data[self.offset..self.offset + 2].try_into().unwrap());
+        self.offset += 2;
+        d
+    }
+
+    fn read_u32(&mut self) -> u32 {
+        let d = u32::from_ne_bytes(self.data[self.offset..self.offset + 4].try_into().unwrap());
+        self.offset += 4;
+        d
+    }
+
+    fn read_i32(&mut self) -> i32 {
+        let d = i32::from_ne_bytes(self.data[self.offset..self.offset + 4].try_into().unwrap());
+        self.offset += 4;
+        d
+    }
+
+    fn read_u64(&mut self) -> u64 {
+        let d = u64::from_ne_bytes(self.data[self.offset..self.offset + 8].try_into().unwrap());
+        self.offset += 8;
+        d
+    }
+}
+
+struct Registers {
+    rax: usize,
+    rdx: usize,
+    rcx: usize,
+    rbx: usize,
+    rsi: usize,
+    rdi: usize,
+    rsp: usize,
+}
+
+impl Registers {
+    fn from_ptr(ptr: *const c_void) -> Registers {
+        Registers {
+            rax: Registers::read_from_stack(ptr, 0),
+            rdx: Registers::read_from_stack(ptr, 1),
+            rcx: Registers::read_from_stack(ptr, 2),
+            rbx: Registers::read_from_stack(ptr, 3),
+            rsi: Registers::read_from_stack(ptr, 4),
+            rdi: Registers::read_from_stack(ptr, 5),
+            rsp: Registers::read_from_stack(ptr, 6),
+        }
+    }
+
+    fn read_from_stack(rsp: *const c_void, off: isize) -> usize {
+        unsafe {
+            let ptr = rsp as *const usize;
+            ptr::read::<usize>(ptr.offset(off))
+        }
+    }
+
+    fn get(&self, id: u16) -> usize {
+        match id {
+            0 => self.rax,
+            1 => self.rdx,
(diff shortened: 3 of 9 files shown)


ptersilie:
Whoops, forgot to mark this as "Draft", and looks like you can't change it afterwards. So, please don't merge just yet.

ltratt:
You can change PRs to/from draft somewhere in the UI (unless it's changed).

yksmp/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "yksmp"
+version = "0.1.0"
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib"]
```
vext01, line 7: This will cause linkage problems later down the line. From memory, if you have `>1` cdylib crate linked, then you get multiple copies of the Rust std lib and they symbols clash.

Can it be a rlib instead?

yksmp/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+#[derive(Debug)]
+enum Location {
+    Register(u16),
+    Direct(u16, i32),
+    Indirect(u16, i32),
+    Constant(u32),
+    LargeConstant(u64),
+}
+
+#[allow(dead_code)]
```
vext01, line 37: I'm OK with leaving these in for now, as long as we add a `FIXME: is this needed` or something.

yksmp/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+struct LiveOut {
+    dwreg: u16,
+    size: u8,
+}
+
+impl StackMapParser<'_> {
```
vext01, line 43: Would be good to move the struct def next to its impl block.

yksmp/Cargo.toml
ptersilie: Actually might not be needed at all anymore now that I've moved it into yk. This used to be a standalone repo.

ptersilie: Fixed in 0f86fdd.

yksmp/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+        }
+        v
+    }
+
+    fn align_8(&mut self) {
+        self.offset += (8 - (self.offset % 8)) % 8;
+    }
+
+    fn read_u8(&mut self) -> u8 {
+        let d = u8::from_ne_bytes(self.data[self.offset..self.offset + 1].try_into().unwrap());
```
vext01, line 181: Eventually we should make the errors bubble up and have a proper error type, but I think that will just slow us down for now.

How about add a `FIXME` for now?

yksmp/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    fn read_u64(&mut self) -> u64 {
+        let d = u64::from_ne_bytes(self.data[self.offset..self.offset + 8].try_into().unwrap());
+        self.offset += 8;
+        d
+    }
+}
+
+/// Maps stackmap records to the addresses where they are relevant. Allows users to retrieve the
```
vext01, line 211: Are they addresses or file offsets?

yksmp/src/lib.rs
ptersilie: I think I completely changed my mind about it in the last hour. It's unlikely we'll need any of this stuff, and if we absolutely do, we can always easily put it back in. This then allows us to simplify `SMQuery`, or more accurately get rid of it entirely and just return a HashMap instead. But let's see what @ltratt thinks first.

yksmp/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+        self.map.get(&addr).map(|r| &r.liveouts)
+    }
+
+    /// Get all locations given the return address of a __llvm_deoptimize function.
+    fn get_locations(&self, addr: u64) -> Option<&Vec<Location>> {
+        self.map.get(&addr).map(|r| &r.locs)
+    }
+}
+
+struct Registers {
```
vext01, line 234: These registers are x86_64 specific. We discussed potentially guarding this with a `cfg`? Is it worth it?

yksmp/src/lib.rs
ptersilie: I was wondering about that. If this errors, something went seriously wrong and the program needs to crash, as I don't think we can recover from that. A guess an error would be nice, but would unlikely be useful to any user of yk.

yksmp/src/lib.rs
ptersilie: Addresses. We'll later pass `__builtin_return_address(0)` into this to retrieve the correct stackmap record.

yksmp/Cargo.toml
vext01: Do the tests still pass? I see that you are exposing ` __yk_stopgap()` to C.

This symbol should probably go in `ykcapi`. You can still write it in Rust by having ` __yk_stopgap()` as a stub which calls into `yksmap`.

yksmp/src/lib.rs
ptersilie: I actually think these should just be `u64` instead. There's a comment at the top saying this whole parser is x64 specific and we'll need to either make it dynamic or write new ones, once LLVM adds support for other architectures.

yksmp/Cargo.toml
vext01: If we do this, the symbol name for the wrapper should probably be `__yksmap_stopgap()`. That's the convention we've used so far.

yksmp/tests/deopt.c
```diff
@@ -0,0 +1,20 @@
...
+}
+
+void foo() {
+    int x = 11;
+    int y = 13;
+    int z;
+    if (x == 11) {
+        z = 3;
+    } else {
+    // Insert deopt here.
```
vext01, line 12: Odd indent?

yksmp/src/lib.rs
vext01: The interpreter could theoretically continue I suppose.

yksmp/Cargo.toml
ptersilie: This isn't hooked properly into yk yet, so this function isn't really used anywhere atm. I think these changes should probably go into the new PR that hooks everything in.

yksmp/src/lib.rs
vext01: ok.

yksmp/Cargo.toml
vext01: > Do the tests still pass?

I see, no test yet calls `__yk_stopgap`.

yksmp/tests/deopt.c
ptersilie: Fixed in 9ddfdb0.

yksmp/src/lib.rs
vext01: OK, so should the map be keyed by `usize` then?

yksmp/src/lib.rs
```diff
@@ -0,0 +1,375 @@
+use std::collections::HashMap;
+use std::convert::TryFrom;
+use std::convert::TryInto;
+use std::ffi::c_void;
+use std::{ptr, slice};
+
+/// Parses LLVM stackmaps version 3 from a given address. Provides a way to query relevant
+/// locations given the return address of a `__llvm_deoptimize` function.
```
ltratt, line 8: Is this supposed to be a paragraph break (with a blank line)? Or should "note" appear on this line?

yksmp/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+    offset: usize,
+}
+
+struct Function {
+    addr: u64,
+    record_count: u64,
+}
+
+struct Record {
+    offset: u32,
```
ltratt, line 23: Are the `offset`s 32 bits on all platforms?

yksmp/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+/// Note that LLVM currently only supports stackmaps for 64 bit architectures. Once they support
+/// others we will need to either make this parser more dynamic or create a new one for each
+/// architecture.
+struct StackMapParser<'a> {
+    data: &'a [u8],
+    offset: usize,
+}
+
+struct Function {
+    addr: u64,
```
ltratt, line 18: Are function addresses always 64 bits on all platforms?

yksmp/src/lib.rs
ltratt: If we don't need 'em, let's nuke 'em.

yksmp/Cargo.toml
ptersilie: I could possibly test this by calling `__yk_stopgap` from `__llvm_deoptimize` in the `deopt.c` test, but then we would have to add langtester to this. Since I was going to use langtester once this is hooked into yk, I felt there is no need to do it here as well.

yksmp/src/lib.rs
ptersilie: How? If we guard fail we need the current values from the trace to continue interpretation. If we can't read the stackmap we can't get those values. And I don't want to say: `Sorry, we couldn't read the stackmap, but we'll continue interpretation with the old bogus values anyway, so if your interpreter does something completely different be warned!`.

yksmp/src/lib.rs
```diff
@@ -0,0 +1,375 @@
+use std::collections::HashMap;
+use std::convert::TryFrom;
+use std::convert::TryInto;
+use std::ffi::c_void;
+use std::{ptr, slice};
+
+/// Parses LLVM stackmaps version 3 from a given address. Provides a way to query relevant
+/// locations given the return address of a `__llvm_deoptimize` function.
+/// Note that LLVM currently only supports stackmaps for 64 bit architectures. Once they support
```
ltratt, line 9: This is a very important thing to note in a minor comment! I think this comment should be a `//!` comment at the top of the module and we *must* add a compile-time guard that prevents this compiling on architectures we don't know for sure that it works on. In practise I think we must say "this only works on x86" as I don't think it's even true that it will work on all 64-bit platforms.

yksmp/src/lib.rs
ptersilie: Since this is all super x64 specific it could also be `u64`.

yksmp/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+}
+
+#[allow(dead_code)]
+struct LiveOut {
+    dwreg: u16,
+    size: u8,
+}
+
+impl StackMapParser<'_> {
+    pub fn parse(data: &[u8]) -> Result<SMQuery, &'static str> {
```
ltratt, line 44: It's generally more idiomatic to do `Result<SMQuery, Box<dyn Error>>`. This allows you to have dynamically generated strings and other error thingies rather than being limited to only `'static str`s.

yksmp/src/lib.rs
ptersilie: LLVM stackmaps currently only support x64 architectures. So yes.

yksmp/src/lib.rs
ptersilie: LLVM stackmaps support X86_64, PowerPC, Aarch64 and SystemZ. But since I have no idea if this code runs on those, a guard sounds like a good idea.

ltratt: In an ideal world, as and when we test a new platform we can relax the compile-time guard.

ptersilie: Can I just put `#[cfg(target_arch = "x86_64")]` at the top of the file or would that just guard the first import?

ltratt: Dunno.

yksmp/src/lib.rs
vext01: you are right. sorry :)

yksmp/src/lib.rs
vext01: true. just `usize` has a pointer implication, sort of. Up to you.

yksmp/src/lib.rs
vext01: I love `Box<dyn Error>`. It has a small performance cost, but it probably doesn't matter here.

yksmp/src/lib.rs
ptersilie: Fixed in bcef01a.

yksmp/src/lib.rs
ptersilie: Fixed in 0f3997f.

yksmp/src/lib.rs
ptersilie: Fixed in 0f3997f.

yksmp/src/lib.rs
ptersilie: Fixed in 8be5b0d.

yksmp/src/lib.rs
ptersilie: Same as above: stackmap limitation.

yksmp/src/lib.rs
ptersilie: Fixed in c5f3000.

ptersilie:
Think this is ready for re-review.

vext01:
LGTM.

@ltratt, how about you?

yksmp/Cargo.toml
ltratt: Perhaps we should add an `unreachable!()` to `__yk_stopgap`? It feels weird to have code that's not being tested at all.

yksmp/src/lib.rs
```diff
@@ -0,0 +1,342 @@
...
+        match l {
+            Location::Direct(reg, off) => {
+                // When using `llvm.experimental.deoptimize` then direct locations should always be
+                // in relation to RSP.
+                assert_eq!(*reg, 7);
+                // We need to add 2 bytes to the value of RSP to factor in the return address and
+                // the pushed RBP value of the previous frame.
+                let addr = registers.get(*reg) + (*off as usize) + 16;
+                let v = unsafe { ptr::read::<u8>(addr as *mut u8) };
+                println!("Direct: {}", v);
```
ltratt, line 273: I'm entirely allergic to this sort of `println`: let's make it a `todo()`.

ptersilie: The reason I made this `print` instead of `todo` is so that we can test this using langtester until we have a stopgap interpreter. If I use todo I can only test the first value.

ltratt: But we're not testing this function at all in this PR are we?

ptersilie: But we will in the very next PR tomorrow. I can remove it temporarily though I hardly see the point of doing so.

ltratt: I understand, but if we're not careful at some point we'll leave in a `print` that should be a `todo` and regret it. It's a good discipline for us to adhere to.

ptersilie: Under normal circumstances I'd agree with you, but this is a bit of a special case. If we use `todo`s we won't be able to test this until we've implemented the stopgap interpreter which will be a while.

ltratt: In the next PR we can reconsider: but, and especially since this function isn't tested in this PR, these prints should be todos.

ptersilie: Making @ltratt happy, one commit at a time: 659b4c1.

ltratt: It's so easy to make me happy :)

ltratt:
Please squash.

yksmp/src/lib.rs
ltratt: > It has a small performance cost, but it probably doesn't matter here.

FWIW, in this case it has no additional cost in the `Ok` case (which is all we care about): `Box` has the same width as `&'static str`.

ptersilie:
Squashed.

ptersilie:
Sorry, forgot to format the c tests. Fixed now.

ltratt:
Please squash.

ptersilie:
Squashed.

vext01:
The failure is because my "by pointer" work is "half-merged". We need the yk side to be merged:
https://github.com/ykjit/yk/pull/452

yksmp/.gitignore
```diff
@@ -0,0 +1,2 @@
+Cargo.lock
+target/
```
bjorn3, line 2: This file is not necessary anymore.

ltratt: I didn't know that! @ptersilie could be in another/the next PR?

ptersilie: Will do!

PR #452, opened 2021-12-15
The control point now passes the live vars struct by pointer.

vext01:
The ykllvm pass that generates the control point now passes the live variables struct by pointer. This has implications for the trace compiler, which makes assumptions about the shape of the IR around the control point.

This is companion to https://github.com/ykjit/ykllvm/pull/16

The ykllvm PR must be merged first.


ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -160,7 +160,7 @@ class JITModBuilder {
   // before the control point) to the extracted values (appearing after the
   // control point). This map helps to match inserted values to their
   // corresponding extracted values using their index in the struct.
-  std::map<uint64_t, Value *> InsertValueMap;
+  std::map<uint64_t, Value *> UpsideDownMap;
```
ptersilie, line 163: Sure this is better than `CPIndexMap` or just `IndexMap`? `UpsideDownMap` sounds like something from a Netflix show.

ltratt: I must admit the whole "upside down" stuff doesn't make sense to me. I suspect I'm missing something obvious, but I'm not quite sure why. I wonder if "reverse order" or similar would be clearer?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -663,35 +663,46 @@ class JITModBuilder {
...
+
+    // When executing the interpreter loop AOT code, execution starts before
+    // the control point and flows below it, but when assembling a trace
+    // control flow is upside-down: we start collecting instructions below the
+    // control point and finish above it.
+    //
+    // This has implications: without extra logic, alloca'd variables become
+    // undefined (as they are defined outside of the trace) and thus need to be
+    // remapped to the input of the compiled trace. SSA values (from the same
+    // block as the control point) remain correct as phi nodes at the beginning
```
ptersilie, line 677: Since everything is a pointer into a struct now, can there even be phi nodes now?

ykllvmwrap/src/jitmodbuilder.cc
vext01: How about `LiveIndexMap`?

ptersilie: I can live with that.

ltratt: @vext01 In terms of the general "upside down", would "reverse order" be correct? I still don't quite understand what "upside down" means here.

ykllvmwrap/src/jitmodbuilder.cc
vext01: > Since everything is a pointer into a struct now, can there even be phi nodes now?

Not everything is a pointer. For example, an `i32` is copied into and out of the struct by value. It's the struct that is passed by pointer.

Also, the data in the struct is only used by the insides of the control point. The rest of the time, the "raw" values are used, so yes, we do still see PHI nodes for those.

e.g. in the `simple_interp_loop1` test:

```
25:
  %26 = phi i32 [ %19, %13 ], [ 0, %100 ], [ %103, %101 ], [ %96, %92 ], !dbg !82
  ...
  %43 = getelementptr %YkCtrlPointVars, %YkCtrlPointVars* %3, i32 0, i32 14, !dbg !93
  store i32 %26, i32* %43, align 4, !dbg !93
  ...
  call void @yk_new_control_point(%struct.YkLocation* %28, %YkCtrlPointVars* %3), !dbg !93
  ...
  %100 = getelementptr %YkCtrlPointVars, %YkCtrlPointVars* %0, i32 0, i32 14, !dbg !8
  %101 = load i32, i32* %100, align 4, !dbg !8
  ...
```

Hopefully that makes sense?

ykllvmwrap/src/jitmodbuilder.cc
vext01: It's to do with the disparity between the order of execution between AOT code and code collected by tracing.

Suppose the AOT code is:
```
bb1:
  <stuff1>
  control_point()
  <stuff2>
  goto bb1
```

since traces start after the control point, and end before it, a trace of one iteration of the above loop will be like:

```
<stuff2>
<stuff1>
```

So the code is "upside-down". There is an attempt to explain this in [this comment](https://github.com/ykjit/yk/pull/452/files#diff-b76ed8a29ed650e7c2a44a5809bc05a22cca958d0b9f5532dab74cce8503d112R669).

This map we are talking about here is needed (without going into mind-boggling detail) to deal with the case where variables are undefined in `<stuff2>` if you don't have the context of `<stuff1>` to tell you where they came from.

So.. propose a name :)

ltratt: Ah! I think it would be clearer to say that "a traced loop begins at the control point; if, as is likely, the control point is not at the beginning of the surrounding `while` loop, code between the start of that `while` loop and the control point will be recorded at the end of the loop." I think I'd drop the "upside down" thing entirely (even though I sort-of now see where you're coming from).

vext01: But what is the name of the map to be?

ltratt: `LiveIndexMap` is fine. I also wonder if I can be cheeky and ask if you'd also consider rephrasing the comments to remove "upside down" so that thickies like me aren't confused in the future?

vext01: Yep, that's fine.

ltratt: Thanks!

vext01: @ptersilie I still think this logic could be simplified now that the struct is passed by pointer. One for another PR though.

vext01: @ltratt try 2df8c8a

ltratt: Modulo the typo, that's fine!

vext01: 760691d fixes typo

ltratt:
I'm now happy, but @ptersilie still needs to comment on one or two things.

vext01:
@ptersilie are you OK with this?

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: This looks good.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Yes, that makes sense.

ptersilie:
Looks good.

vext01:
OK to squash?

ptersilie:
Yes.

vext01:
splat.

PR #454, opened 2021-12-17
Hook in transition_location properly.

vext01:
This commit is a rough and ready attempt at hooking in the full-blown `transition_location()` logic.

I've only tested it with one test "simple.c". This works with small adjustments to the test.

Lots of code goes away, which is good.

Don't review this deeply, but is this roughly what we were talking about?

If so, @ltratt before I continue, there are some questions annotated `LAURIE: `.

~I'll have a draft companion PR in a moment too.~ Here is the companion: https://github.com/ykjit/ykllvm/pull/17.

### DIFF
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -22,7 +22,7 @@ uint64_t getNewTraceIdx() {
 }
 
 #define TRACE_FUNC_PREFIX "__yk_compiled_trace_"
-#define YK_NEW_CONTROL_POINT "yk_new_control_point"
+#define YK_NEW_CONTROL_POINT "__ykrt_control_point"
 #define YK_CONTROL_POINT_ARG_IDX 1
 
 // Dump an error message and an LLVM value to stderr and exit with failure.
@@ -826,10 +826,6 @@ public:
       BasicBlock *BB;
       std::tie(F, BB) = getLLVMAOTFuncAndBlock(&IB);
 
-      if (F->getName() == YK_NEW_CONTROL_POINT) {
-        continue;
-      }
-
       assert(LastCompletedBlocks.size() >= 1);
       LastCompletedBlocks.back() = NextCompletedBlock;
       NextCompletedBlock = BB;
@@ -889,6 +885,7 @@ public:
               break;
             }
           } else if (CF->getName() == YK_NEW_CONTROL_POINT) {
+            ExpectUnmappable = true; // control point is always opaque.
             if (NewControlPointCall == nullptr) {
               NewControlPointCall = &*CI;
             } else {
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -1,21 +1,15 @@
 //! The main end-user interface to the meta-tracing system.
 
-// FIXME these were added during the big code deletion event that occurred after we decided to use
-// LLVM IR instead of SIR/TIR. Large chunks of the system suddenly became todo!() meaning that code
-// that we do still want to keep (some of which is scary concurrency stuff) became dead in from
-// Rust's POV.
-#![allow(unreachable_code, unused_variables, dead_code)]
-
-use core::ffi::c_void;
 use std::{
     cell::Cell,
     cmp,
     collections::VecDeque,
+    ffi::c_void,
     marker::PhantomData,
     ptr,
     sync::{
         atomic::{AtomicU32, AtomicUsize, Ordering},
-        Arc, Mutex as RustMutex,
+        Arc,
     },
     thread,
 };
@@ -36,33 +30,12 @@ use yktrace::{start_tracing, stop_tracing, CompiledTrace, IRTrace, TracingKind};
 pub type HotThreshold = u32;
 #[cfg(target_pointer_width = "64")]
 type AtomicHotThreshold = AtomicU32;
-const DEFAULT_HOT_THRESHOLD: HotThreshold = 50;
+
+// FIXME: just for parity with existing tests for now.
+const DEFAULT_HOT_THRESHOLD: HotThreshold = 0;
 
 static GLOBAL_MT: Lazy<MT> = Lazy::new(|| MT::new());
 thread_local! {static THREAD_MTTHREAD: MTThread = MTThread::new();}
-
-/// Actions that can be issued to the control point when transitioning a `Location`. These values
-/// are tags used in a `usize`. See the comment next to `transition_location_simple()` for detailed
-/// information.
-const JITACTION_NOP: usize = 1;
-const JITACTION_START_TRACING: usize = 2;
-const JITACTION_STOP_TRACING: usize = 3;
-
-// The mask for the above tagging scheme.
-const JITACTION_MASK: usize = 0b11;
-
-// FIXME: This struct should not exist. It is part of an over-simplified interim design. Not thread
-// safe, only allows one binary trace etc. To fix this we need to properly use the information
-// encoded in a `Location` by hooking in full-blown `transition_location()`.
-//
-// Also the pointers in this struct have been stored as `usize`s, as Rust gets upset that pointers
-// can't be shared across threads safely. Rust doesn't know that the design is (for now)
-// single-threaded.
-struct JITState {
-    start_loc: usize,  // really a `*const Location`.
-    trace_code: usize, // really a `*const c_void`.
-    are_we_tracing: bool,
-}
 
 #[derive(Clone)]
 /// A meta-tracer. Note that this is conceptually a "front-end" to the actual meta-tracer akin to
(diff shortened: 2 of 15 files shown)


ykrt/src/mt.rs
```diff
@@ -318,6 +210,9 @@ impl MT {
                             // We've initialised this Location and obtained the lock, so we can now
                             // start tracing for real.
                             let tid = Arc::clone(&mtt.tid);
+                            // FIXME: this and other jit-state prints should be behind an opt-in
+                            // `jit-state-print` feature.
+                            eprintln!("jit-state: start-tracing");
```
ltratt, line 215: Oh, god, my poor code polluted by this monstrosity! 📦:p

[In even the medium term, this is actually really hard to do properly: what does it mean in a multi-threaded context? I doubt that `print`s are the right way of doing it. I'll fight that battle another day!]

ykrt/src/mt.rs
```diff
@@ -357,10 +252,19 @@ impl MT {
             let hl = unsafe { ls.hot_location() };
             let hl_ptr = hl as *mut _ as *mut ();
             match hl {
-                HotLocation::Compiled(tr) => {
+                HotLocation::Compiled(compiled_trace) => {
```
ltratt, line 255: I know you don't like short variable names, but that's the convention in this file, so let's not mix two coding styles.

ykrt/src/mt.rs
```diff
@@ -357,10 +252,19 @@ impl MT {
...
+
+                    // FIXME: this loop shouldn't exist. Trace stitching should be implemented in
+                    // the trace itself.
+                    // https://github.com/ykjit/yk/issues/442
+                    loop {
+                        eprintln!("jit-state: enter-jit-code");
+                        compiled_trace.exec(ctrlp_vars);
+                        eprintln!("jit-state: exit-jit-code");
+                    }
+                    // LAURIE: should the lock be held while the trace is being executed?
```
ltratt, line 267: No: in fact, the lock *mustn't* be held, otherwise the system will fail to make progress in some situations. For the time being it's fine to put the `unlock` before the `loop`.

ykrt/src/mt.rs
```diff
@@ -381,6 +285,8 @@ impl MT {
                     };
                     *hl = HotLocation::Compiled(tr);
                     loc.unlock();
+                    // LAURIE: Can we execute the JITted code now? Otherwise we have to interpret
```
ltratt, line 288: We *could* but that's a bit of extra complexity that I think we should avoid in this PR. Leaving a FIXME in to remind us to do it would be a good idea though.

ykrt/src/mt.rs
```diff
@@ -434,16 +341,21 @@ impl MT {
...
-            // FIXME compile IR trace to executable code here.
-            let compiled = todo!();
-            *mtx.lock() = Some(Box::new(compiled));
-            // FIXME: although we've now put the compiled trace into the mutex, there's no
-            // guarantee that the Location for which we're compiling will ever be executed
-            // again. In such a case, the memory has, in essence, leaked.
-        });
-
-        self.queue_job(f);
+        // FIXME: Until we can find a way of testing non-deterministic and time-sensitive JIT
```
ltratt, line 344: I'm decidedly unkeen on this change because you're commenting out very critical code that will now go untested.

I'd actually prefer to cheat and let compilation carry in a thread but wait for it after the `let f = ...` stuff with something like:

```
    loop {
            match loc.try_lock() {
                Ok(x) => ls = x,
                Err(_) => todo!()
            }
            let hl = unsafe { ls.hot_location() };
            let hl_ptr = hl as *mut _ as *mut ();
            match hl {
                  HotLocation::Compiled(_) = break;
                  _ => ()
            }
    }
```

Warning: untested!

ltratt:
Broadly speaking this looks good: my only real comment is that I'd prefer to avoid commenting out `queue_job` if possible, because that means we don't exercise quite a bit of code. I've suggested a hack that sort-of gets around this (but I haven't tested it!).

vext01:
> I'd prefer to avoid commenting out queue_job if possible, because that means we don't exercise quite a bit of code.

Initial testing suggests that that code does work somewhat, but I refrained from using it because it introduces non-determinism that would mean adjusting the tests quite a bit. 

On Monday, I'll put it back and see how we go.

ykrt/src/mt.rs
vext01: OK.

ykrt/src/mt.rs
vext01: Good. That's what I thought.

ykrt/src/mt.rs
vext01: the reason I did this is because there are so many different kinds of traces involved. In this file I think it's important to differentiate from an IR trace and a compiled trace.

ykrt/src/mt.rs
vext01: To be clear, these will be guarded so that they are only present for testing.

This is the same semantics as what we do currently (albeit it used to be in generated IR in ykllvm). I had to introduce them to have the current tests pass.

(We should probably prefix the events with some kind of thread identifier though).

ltratt: I've realised that this sort of testing really isn't going to work well in multi-threading. I understand that we need to do *something* but it's not going to be simple print statements. Maybe global counters will be the way to go -- I need to think a bit more.

ykrt/src/mt.rs
ltratt: `ctr` if you wish. But let's not mash styles like this.

ltratt:
> Just saw your suggested code, which (I think) keeps determinism for now.

It should do. It's basically the world's longest spin lock!

ykrt/src/mt.rs
vext01: Did you mean `mtx.lock()`?

vext01: (instead of `loc.try_lock()`)

I'm a bit confused, as `loc` doesn't exist.

vext01: `ls` is also undefined.

ltratt: Erp, true, that stuff is defined in the main lock.

I think the easiest thing is to get this PR in and for me to do something in a subsequent PR.

ltratt:
@vext01 Let's go through the minor comments, but I think the "let's get queue_job going" thing is too much for this PR.

vext01:
Sounds good to me.

It may not be a quick job, as one of the tests is failing in an odd way.

Bear with me.

ykrt/src/mt.rs
vext01: I went with `tr`, as `ctr` invokes "counter" in my mind. See 0df447df6.

ykrt/src/mt.rs
vext01: Fixed in 0df447df6

ykrt/src/mt.rs
vext01: Fixed added in 0df447df6.

You will also notice in 3bccbb593 that I had to make some of the tests run longer, as the JIT now takes one more iteration than is strictly necessary to enter JIT code.

ykrt/src/mt.rs
vext01: Left as-is then.

ykrt/src/mt.rs
vext01: I've feature-gated those for now. See 7a4e8799dc9.

vext01:
Note that none of the multi-threading code is tested with more than a single thread. I'll leave that for another PR (for you perhaps?), as this PR is already huge.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -854,7 +850,6 @@ class JITModBuilder {
           continue;
 
         if (isa<CallInst>(I)) {
-
```
vext01, line 857: In hindsight, probably shouldn't have killed this line.

ykrt/src/mt.rs
```diff
@@ -4,18 +4,18 @@
 // LLVM IR instead of SIR/TIR. Large chunks of the system suddenly became todo!() meaning that code
 // that we do still want to keep (some of which is scary concurrency stuff) became dead in from
 // Rust's POV.
-#![allow(unreachable_code, unused_variables, dead_code)]
+//#![allow(unreachable_code, unused_variables, dead_code)]
```
vext01, line 7: Gah. I always find more stuff when looking at the diff in GH.

ykrt/src/mt.rs
ltratt: Good idea!

vext01:
I hope this is close now @ltratt

ltratt:
Please squash.

vext01:
squashed and wrote a proper commit message (please review).

vext01:
Don't forget that this depends on https://github.com/ykjit/ykllvm/pull/17

ltratt:
For clarity in the commit: s/Before the JIT/Previously, the JIT/ ?

vext01:
Fixed.

vext01:
This is ready to go now.

PR #460, opened 2022-01-05
Integrate deoptimisation into guard failures.

ptersilie:
Until now we've simply crashed when we encountered a guard failure. This
PR integrates the previously added stackmap parser into our guard
failures so we can read out the live variables at that point. This is
another step towards proper stopgapping, though until we have a stopgap
interpreter we only print out the live variables we've found.


ykcapi/src/lib.rs
```diff
@@ -48,6 +52,116 @@ pub extern "C" fn __ykutil_get_llvmbc_section(res_addr: *mut *const c_void, res_
     }
 }
 
+struct Registers {
```
ptersilie, line 55: The following changes are mostly moved over from yksmp/src/lib.rs.

c_tests/run.rs
```diff
@@ -103,6 +103,8 @@ fn mk_compiler(exe: &Path, src: &Path, opt: &str, extra_objs: &[PathBuf]) -> Com
         "-Wall",
         // Enable LTO with lld.
         "-fuse-ld=lld",
+        // Don't remove frame pointers as we'll need them for guard failures.
+        "-fno-omit-frame-pointer",
```
vext01, line 107: So we couldn't address from the stack pointer? I guess it complicates things a lot?

c_tests/tests/const_global.c
```diff
@@ -8,7 +8,7 @@
 //     jit-state: stop-tracing
 //     --- Begin jit-pre-opt ---
 //     ...
-//     define internal void @__yk_compiled_trace_0(%YkCtrlPointVars* %0) {
+//     define internal void @__yk_compiled_trace_0(%YkCtrlPointVars* %0, i64* %1, i64 %2) {
```
vext01, line 11: remind me what the extra trace args are again.

c_tests/run.rs
ptersilie: If I remember correctly this is needed so we can get the return address with `__builtin_return_address(0)`.

c_tests/tests/ptr_global.c
```diff
@@ -5,7 +5,8 @@
 //   stderr:
 //     ...
 //     i=25
-//     ptr_global: guard-failure
+//     Stopgap
```
vext01, line 8: Should this be something like `jit-state: stopgap` to be in-line with our other jit state prints?

c_tests/tests/const_global.c
ptersilie: It's explained further down in `jitmodbuilder`. These are the address and size of the stackmap.

c_tests/tests/simple.c
```diff
@@ -39,7 +38,11 @@
 //     jit-state: exit-jit-code
 //     jit-state: enter-jit-code
 //     i=1
-//     simple: guard-failure
+//     Stopgap
+//     ...
+//     Indirect: 10006 ...
```
vext01, line 43: What are the numbers being matched here again?

c_tests/tests/ptr_global.c
ptersilie: Could do. It doesn't matter too much I guess, since this is going to go as soon as we have the stopgap interpreter.

ykcapi/src/lib.rs
```diff
@@ -48,6 +52,116 @@ pub extern "C" fn __ykutil_get_llvmbc_section(res_addr: *mut *const c_void, res_
...
+/// FIXME: Until we have the stopgap interpreter we simply print out the values we find.
+#[cfg(target_arch = "x86_64")]
+#[no_mangle]
+pub extern "C" fn yk_stopgap(
+    addr: *const c_void,
+    size: usize,
+    retaddr: usize,
+    rsp: *const c_void,
+) {
+    eprintln!("Stopgap");
```
vext01, line 111: If this really is a jit-state print, I think it needs to be compile-time guarded like the others.

ykcapi/src/lib.rs
```diff
@@ -48,6 +52,116 @@ pub extern "C" fn __ykutil_get_llvmbc_section(res_addr: *mut *const c_void, res_
...
+    rcx: usize,
+    rbx: usize,
+    rsi: usize,
+    rdi: usize,
+    rsp: usize,
+}
+
+impl Registers {
+    fn from_ptr(ptr: *const c_void) -> Registers {
+        Registers {
```
vext01, line 67: Are these the caller-save registers defined by the ABI?

c_tests/tests/simple.c
ptersilie: These are some live values we expect. When the guard fails we have just done `res += 2` within a loop, so we are expecting at least two live variables, before the addition (`10006`) and after the addition (`10008`). There are other live variables but these are an easy way to check that stackmaps are working.

ykcapi/src/lib.rs
ptersilie: Elaborate.

ykcapi/src/lib.rs
```diff
@@ -48,6 +52,116 @@ pub extern "C" fn __ykutil_get_llvmbc_section(res_addr: *mut *const c_void, res_
...
+    }
+}
+
+/// Parses the stackmap and saved registers from the given address (i.e. the return address of the
+/// deoptimisation call).
+/// FIXME: Until we have the stopgap interpreter we simply print out the values we find.
+#[cfg(target_arch = "x86_64")]
+#[no_mangle]
+pub extern "C" fn yk_stopgap(
+    addr: *const c_void,
```
vext01, line 106: It would be helpful to prefix `addr` and `size` with `sm_` so that a reader knows it is the stackmap we are talking about.

ykcapi/src/lib.rs
vext01: The other jit-state prints are [guarded like this](https://github.com/ykjit/yk/blob/cef36509b8a31636b804e314ffa8c5732199ffeb/ykrt/src/mt.rs#L207-L208).

If this is a jit-state print too, it should be guarded in the same way.

This may not be a jit-state print of course. I'm not sure.

ykcapi/src/lib.rs
ptersilie: Actually, this is supposed to be all possible registers, but I just realised we are missing a few. Whoops.

vext01: OK, but what's not clear to me:

Later we do:
```
```

and `from_ptr()` loads the registers from the stack. Why would *all* registers have been spilled at this point? Does the deoptimise intrinsic implicitly do that for us?

ykcapi/src/lib.rs
```diff
@@ -48,6 +52,116 @@ pub extern "C" fn __ykutil_get_llvmbc_section(res_addr: *mut *const c_void, res_
...
+            SMLocation::Constant(v) => {
+                eprintln!("Constant: {}", v);
+            }
+            SMLocation::LargeConstant(v) => {
+                eprintln!("Large constant: {}", v);
+            }
+        }
+    }
+    // FIXME: Initialise stopgap interpreter here.
+    process::exit(0);
```
vext01, line 162: Exit code zero is indicating success. I wonder if it would make more sense to return non-zero, but does that get in the way of lang_tester?

ykcapi/yk.h
```diff
@@ -48,4 +48,42 @@ YkLocation yk_location_new(void);
...
+    // the value of RSP in a variable so we know where to look for the register
+    // values. The order in which we push the registers is equivalent to the
+    // Sys-V x86_64 ABI, which the stackmap format uses as well. However, since
+    // we are interested in the value of RSP in the previous stack frame, which
+    // has changed when we entered the __llvm_deoptimize function, we push RBP
+    // instead, which is equivalent. Note, that because of this there is no
+    // (easy) way to access RBP from the previous frame, so we just assume it's
+    // never referenced by the stackmap.
+    void *rsp;
+    __asm__ volatile (
```
vext01, line 65: I think this block of code answers my question about what registers are spilled.

Should we move this block into a method in `Registers` so that they are right next to each other?

ykcapi/src/lib.rs
ptersilie: Probably makes sense to guard it as it's only needed for the tests.

ykcapi/src/lib.rs
ptersilie: I spilled them manually at the beginning of the `__llvm_deoptimize` call: https://github.com/ykjit/yk/pull/460/files#diff-65ddb41c4a2e6cf6d0de3ba8be066c3ac59faa885f4f52b065b5c35b8352697eR65

ykcapi/src/lib.rs
ptersilie: Yes, it gets into the way of lang_tester.

ykcapi/yk.h
ptersilie: Not sure we can, to be honest. Don't know if calling another function here would mess with the register values. The very first thing we need to do in `llvm_deoptimize` is save the registers so that they don't get overwritten.

ykcapi/src/lib.rs
ptersilie: I think it's time to refactor this struct a bit so we don't explicitly name all registers.

ykcapi/yk.h
```diff
@@ -48,4 +48,42 @@ YkLocation yk_location_new(void);
...
+            "push %%rsi\n"
+            "push %%rbx\n"
+            "push %%rcx\n"
+            "push %%rdx\n"
+            "push %%rax\n"
+            "movq %%rsp, %0\n": "=rm" (rsp));
+
+    // We can use this function's return address to lookup the correct record
+    // in the stackmap.
+    void *ret = __builtin_return_address(0);
```
vext01, line 78: New line here?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -115,6 +116,26 @@ Value *getYkCtrlPointVarsStruct(Module *AOTMod, InputTrace &InpTrace) {
   return CI->getArgOperand(YK_CONTROL_POINT_ARG_IDX);
 }
 
+
+/// Extract all live variables that need to be passed into the control point.
+/// This is currently an overapproximation and will return some variables that
```
vext01, line 121: shall we put `FIXME: ` before the second line?

ykllvmwrap/src/memman.cc
```diff
@@ -5,35 +5,47 @@
...
-MemMan::~MemMan() {}
-
-uint8_t *alloc_mem(uintptr_t Size, unsigned Alignment,
+uint8_t *allocateSection(uintptr_t Size, unsigned Alignment,
                    std::vector<AllocMem> *Vec) {
   uintptr_t RequiredSize = Alignment * ((Size + Alignment - 1) / Alignment + 1);
-  auto Ptr = (unsigned char *)mmap(0, RequiredSize, PROT_WRITE,
-                                   MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+  auto Ptr = (uint8_t *) mmap(0, RequiredSize,
+                PROT_READ | PROT_WRITE,
```
vext01, line 12: We added the READ permission here. Why's that again?

ptersilie: I think on some systems WRITE doesn't imply READ, but I imagine stackmaps won't work on those systems anyway, so it doesn't really matter too much. Just though it's more accurate this way.

ykcapi/yk.h
vext01: I see. At least a comment saying that this must stay in parity with `Registers` then?

ykllvmwrap/src/memman.cc
vext01: FWIW, it's fairly normal to have pages RW, so I'm happy with the change anyway.

ykcapi/src/lib.rs
ptersilie: Partly fixed in efcb294.

I've decided against adding more registers since we currently don't have tests that use them. Once we do we'll hit an `unreachable!` so we'll know to add them.

ykcapi/src/lib.rs
ptersilie: Reworded in e42675f. We decided offline that adding the guards isn't worth it.

ykcapi/src/lib.rs
ptersilie: 1d282d4

ykcapi/yk.h
ptersilie: I think this is already addressed in efcb294 via the added `unreachable!`.

ykcapi/yk.h
ptersilie: Fixed in 59787fc

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Fixed in dd3af2f

ptersilie:
Think that's all comments addressed. Should we squash and then wait for @ltratt to have a look?

vext01:
@ltratt is bound to have queries and comments. I think we should let him have at least a quick look.

ptersilie:
Yes, I just thought it's less distracting for him if we squash away the already addressed stuff first.

c_tests/run.rs
ltratt: Before we turn frame pointers off (which, AIUI, can have a big effect on optimisations), we should see if there's another way of doing this.

c_tests/tests/ptr_global.c
ltratt: I tend to agree with Edd: it's a good discipline to keep to the convention, even if we only need it temporarily.

ykllvmwrap/src/memman.cc
ltratt: > I think on some systems WRITE doesn't imply READ

That's correct: even more, some systems (e.g. OpenBSD) cause a segfault if a page is `WRITE` but not `READ`.

ykcapi/src/lib.rs
```diff
@@ -48,6 +52,99 @@ pub extern "C" fn __ykutil_get_llvmbc_section(res_addr: *mut *const c_void, res_
...
+        }
+    }
+
+    fn read_from_stack(&self, off: isize) -> usize {
+        unsafe {
+            ptr::read::<usize>(self.addr.offset(off))
+        }
+    }
+
+    fn get(&self, id: u16) -> usize {
```
ltratt, line 72: Presumably this function (and maybe others?) should be `#[cfg(target_arch = "x86_64")]`.

ykcapi/src/lib.rs
```diff
@@ -48,6 +52,99 @@ pub extern "C" fn __ykutil_get_llvmbc_section(res_addr: *mut *const c_void, res_
...
+            SMLocation::Constant(v) => {
+                eprintln!("Constant: {}", v);
+            }
+            SMLocation::LargeConstant(v) => {
+                eprintln!("Large constant: {}", v);
+            }
+        }
+    }
+    // FIXME: Initialise stopgap interpreter here.
+    process::exit(0);
```
ltratt, line 145: Wouldn't a `todo` be better? Or is this to ensure that tests can run?

ykcapi/yk.h
```diff
@@ -48,4 +48,43 @@ YkLocation yk_location_new(void);
 // will occur.
 void yk_location_drop(YkLocation);
 
+void yk_stopgap(void *addr, uintptr_t size, uintptr_t retaddr, void *rsp);
+
+void __llvm_deoptimize(void* addr, uintptr_t size) {
```
ltratt, line 53: Should this be `#[cfg(target_arch = "x86_64")]`?

ykcapi/yk.h
```diff
@@ -48,4 +48,43 @@ YkLocation yk_location_new(void);
...
+    // We can use this function's return address to lookup the correct record
+    // in the stackmap.
+    void *ret = __builtin_return_address(0);
+
+    // Forward stackmap address and size, and the return and rsp addresses to
+    // the function that parses the stackmap and initiates the stopgap
+    // interpreter.
+    yk_stopgap(addr, size, (uintptr_t) ret, rsp);
+
+    // Reverse the registers we've pushed onto the stack at the beginning of
```
ltratt, line 85: This doesn't look like it reverses anything to me, but perhaps I've misunderstood.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -115,6 +116,26 @@ Value *getYkCtrlPointVarsStruct(Module *AOTMod, InputTrace &InpTrace) {
   return CI->getArgOperand(YK_CONTROL_POINT_ARG_IDX);
 }
 
+
```
ltratt, line 119: Double blank line?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -351,6 +372,10 @@ class JITModBuilder {
     // Create the function.
     std::vector<Type *> InputTypes;
     InputTypes.push_back(TraceInputs->getType());
+    // Add arguments for stackmap pointer and size.
+    // FIXME: x64 only!
```
ltratt, line 376: Let's add a `#if <x86>` here then.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -235,7 +236,15 @@ extern "C" void *compileModule(string TraceName, Module *M,
     errx(EXIT_FAILURE, "Couldn't compile trace: %s",
          EE->getErrorMessage().c_str());
 
-  return (void *)EE->getFunctionAddress(TraceName);
+  // Allocate space for compiled trace address, stackmap address, and stackmap
+  // size (plus padding).
+  // FIXME This is a temporary hack until the redesigned hot location is up.
+  uintptr_t *ptr = (uintptr_t *) malloc(32);
```
ltratt, line 242: Let's at least make this `malloc(sizeof(uintptr_t) * 3)`.

c_tests/run.rs
ptersilie: Looks like there's a `RA` register on x64 which supposedly contains the return address. Let me see if that works.

c_tests/tests/ptr_global.c
ptersilie: Forgot the "Fixed in e42675f"

ykcapi/src/lib.rs
ltratt: OK.

ykcapi/src/lib.rs
ptersilie: The latter.

ykcapi/src/lib.rs
```diff
@@ -48,6 +52,99 @@ pub extern "C" fn __ykutil_get_llvmbc_section(res_addr: *mut *const c_void, res_
     }
 }
 
+struct Registers {
+    addr: *const usize,
+}
+
+impl Registers {
```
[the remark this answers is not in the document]
ptersilie: Fixed in a8383f5

ykcapi/src/lib.rs
ptersilie: Fixed in a8383f5

c_tests/run.rs
vext01: That's the first I've heard of this. Got a link?

ykcapi/yk.h
vext01: I agree that it needs rewording.

ykllvmwrap/src/ykllvmwrap.cc
vext01: I assume that the other 8 bytes are the padding the comment mentions?

ykcapi/yk.h
ptersilie: Fixed in 0a14653.

ykcapi/yk.h
ptersilie: Fixed in 3e7cd91

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Fixed in 39cee38.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Fixed in 39cee38.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: What about the padding?

ptersilie: Oh, seems like padding isn't needed for malloc.

ptersilie:
> > It's the ID of the register according the the ABI for the platform.
> 
> Ah, let's reference that PDF then! Also it's probably worth clearing up why we have both `off` and `id`. I _think_ I can guess why, but it's not 100% clear to me.

The doc string for `Registers` already mentions the DWARF numbering, but I can try and make it more clear.

ykllvmwrap/src/ykllvmwrap.cc
ptersilie: Fixed in f5ca58d.

yktrace/src/lib.rs
```diff
@@ -156,19 +156,30 @@ pub struct CompiledTrace {
...
+    /// Pointer to the stackmap, required to parse the stackmap during a guard failure.
+    smptr: *const c_void,
+    /// The stackmaps size.
+    smsize: usize,
 }
 
 use std::mem;
+use std::slice;
 impl CompiledTrace {
     pub fn new(code_ptr: *const c_void) -> Self {
```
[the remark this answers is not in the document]
ptersilie: Fixed in eda4548.

ptersilie:
Try e793efb and 8bb62ce

c_tests/run.rs
ptersilie: Ah, so `fno-omit-frame-pointer` stores the return address in the `RA` register. With `fomit-frame-pointer` this register can be used as general purpose register, which explains why this may have a performance implication. So we may have to read the return address from the stack manually.

ptersilie: Fixed in 4cd89ad

ptersilie:
With @vext01's help found a way to keep `omit-frame-pointer` and get the return address and previous frame pointer via a different route.

Ready for re-review I think.

ykcapi/yk.h
```diff
@@ -51,40 +51,37 @@ void yk_location_drop(YkLocation);
...
-            "push %%rbx\n"
-            "push %%rcx\n"
-            "push %%rdx\n"
-            "push %%rax\n"
-            "movq %%rsp, %0\n": "=rm" (rsp));
-
-    // We can use this function's return address to lookup the correct record
-    // in the stackmap.
-    void *ret = __builtin_return_address(0);
+    // Sys-V x86_64 ABI, which the stackmap format uses as well. This function
```
vext01, line 59: > This function...

I think this bit can be made shorter.

All we really need to say is that we use a naked function so that the C compiler cannot emit a function prolog that would fiddle with the stack pointer.

You don't need to say anything about asm, as naked functions in C are by definition, purely asm.

ykcapi/yk.h
```diff
@@ -51,40 +51,37 @@ void yk_location_drop(YkLocation);
...
-    // the function that parses the stackmap and initiates the stopgap
-    // interpreter.
-    yk_stopgap(addr, size, (uintptr_t) ret, rsp);
-
-    // Leave the stack as we found it.
-    __asm__ volatile ("add $0x40, %%rsp\n" :);
+    __asm__ volatile (
+        ".intel_syntax;\n"
+        // FIXME: Add other registers that may be referenced by the stackmap.
+        "push rsp;\n"
```
vext01, line 68: I think we can add more comments to this bit, like:

 - push register contents to the stack.
 - set up the <blah> argument for the call to `yk_stopgap`.
 - free stack space.

vext01: We might also say that rdi and rsi are passed right through.

ykcapi/yk.h
ltratt: I hadn't heard of naked functions before and had to look them up. Fortunately I chose my search terms carefully!

ykcapi/yk.h
ptersilie: Is the first point not already included in the top comment? I don't think we need to explicitly say we freeing up stack space, the code makes that clear. And there's also no argument setup for `yk_stopgap` since we are just forwarding the arguments from the current function.

ltratt:
I have no comments beyond @vext01's comments.

ykcapi/yk.h
vext01: > there's also no argument setup for yk_stopgap since we are just forwarding the arguments from the current function.

Lies!

```
        "mov rdx, [rsp+64];\n"
        "mov rcx, rsp;\n"
```

> Is the first point not already included in the top comment?
> I don't think we need to explicitly say we freeing up stack space, the code makes that clear.

I just wonder if we will understand what this code does a few months down the line. It's fiddly stuff, so documenting our intention may help us out a great deal later.

ykcapi/yk.h
ptersilie: Try 268e8ce

ltratt:
As and when @vext01 is happy, please squash.

ykcapi/yk.h
vext01: nitpick: "stackframe" or "stack frame"? Sorry to be a pain.

ptersilie:
Just 3 more commits for you to review @ltratt and we are ready.

ykcapi/yk.h
```diff
@@ -71,9 +72,12 @@ __attribute__((naked)) void __llvm_deoptimize(void* addr, uintptr_t size) {
         "push rcx;\n"
         "push rdx;\n"
         "push rax;\n"
+        // 1st arg to yk_stopgap: stackmap addr (already in rdi).
+        // 2nd arg to yk_stopgap: stackmap size (already in rsi).
+        // 3rd arg to yk_stopgap: return address
         "mov rdx, [rsp+64];\n"
```
ltratt: Silly question: why `+64` if the 3 things mentioned above only add up to 24?

ptersilie: The offset corresponds with the push instructions not the arguments. The 64 looks up the element on the stack when we enter the `llvm_deoptimize` call which is the return address.

ltratt: OK, then I'm not sure I understand why the comments are placed where they are? Can we reorg this a bit, and maybe explain the `64` too?

ptersilie: The `1st` and `2nd` don't need to be there, but I don't know where else to put them. The `3rd` comment is there to tell us that `rdx` contains the return address which is an argument into `yk_stopgap`.

ltratt: I have no idea, other than to say that the current situation seems confusing :) @vext01 might have ideas?

ptersilie: Try d8487c8

ltratt:
Please squash.

ptersilie:
Squashed.

ptersilie:
Sorry, forgot to format.

ptersilie:
Okay to squash?

ykcapi/src/lib.rs
```diff
@@ -48,6 +52,122 @@ pub extern "C" fn __ykutil_get_llvmbc_section(res_addr: *mut *const c_void, res_
...
+    /// Creates a Registers struct from a given a pointer on the stack containing spilled
+    /// registers.
+    fn from_ptr(ptr: *const c_void) -> Registers {
+        Registers {
+            addr: ptr as *const usize,
+        }
+    }
+
+    /// Read the spilled register value at the offset `off` from the previous stack frame.
+    fn read_from_stack(&self, off: isize) -> usize {
```
bjorn3, line 72: This should be unsafe. This can be used to read arbitrary (including uninitialized and unmapped) memory.

ykcapi/src/lib.rs
```diff
@@ -48,6 +52,122 @@ pub extern "C" fn __ykutil_get_llvmbc_section(res_addr: *mut *const c_void, res_
...
+        unsafe { ptr::read::<usize>(self.addr.offset(off)) }
```
bjorn3, line 73: Is it intended that the offset uses pointer arithmetic instead of integer arithmetic?

ykcapi/src/lib.rs
```diff
@@ -48,6 +52,122 @@ pub extern "C" fn __ykutil_get_llvmbc_section(res_addr: *mut *const c_void, res_
...
+    }
+
+    /// Retrieve the previous frame's register value given by its DWARF register number `id`. This
+    /// number additionally functions as an offset into the the spilled stack to find that
+    /// register's value.
+    #[cfg(target_arch = "x86_64")]
+    fn get(&self, id: u16) -> usize {
```
bjorn3, line 80: Either this function or `from_ptr` should be unsafe as the address field could be a dangling or otherwise invalid pointer.

ykcapi/src/lib.rs
ptersilie: I assume you mean that I should mark the function as unsafe rather than wrapping the body in an unsafe block?

ykcapi/src/lib.rs
ptersilie: Yes, `addr` is a pointer into the stack where we read from different offsets. Why is it important to make clear that we are using pointer arithmetic here?

ykcapi/src/lib.rs
bjorn3: Yes

ykcapi/src/lib.rs
bjorn3: I initially got confused. If you don't want to change the docs that is fine too.

ykcapi/src/lib.rs
ptersilie: Thanks! Fixed in 04406d1

ykcapi/src/lib.rs
ptersilie: Thanks! Fixed in 04406d1

ptersilie:
Ready for re-review.

vext01:
LGTM.

vext01:
You need to resolve merge conflicts.

ptersilie:
Squashed and rebased.

ltratt:
bors r=

ptersilie:
Huh, this is odd. The tests succeed if I run them manually. Investigating.

ptersilie:
The failing tests are release tests. Could it be that we don't set the debug printing flag in release mode?

ptersilie:
This should fix it.

ptersilie:
Squashed!

ptersilie:
Of course I only checked the rust formatting, not the C formatting. Sorry.

ltratt:
Please squash.

ptersilie:
Squashed.

PR #461, opened 2022-01-05
Make compilation happen in parallel.

### LINKED ISSUE #458: Tracing into the control point?
I *think* we might be accidentally including parts of the internal control point mechanism in the trace. If I change `__yk_control_point` to:

```c
#[no_mangle]
pub extern "C" fn __ykrt_control_point(loc: *mut Location, ctrlp_vars: *mut c_void) {
    debug_assert!(!ctrlp_vars.is_null());
    if !loc.is_null() {
        eprintln!("before {:?}", unsafe {&*loc}.load(std::sync::atomic::Ordering::Relaxed));
        MT::transition_location(unsafe { &*loc }, ctrlp_vars);
        eprintln!("after {:?}", unsafe { &*loc}.load(std::sync::atomic::Ordering::Relaxed));
    }
}
```

then for `simple_interp_loop.c` I see this at the end:

```
before LocationInternalState { x: 4 }
counting
after LocationInternalState { x: 4 }
pc=1, mem=3
before LocationInternalState { x: 4 }
counting
after LocationInternalState { x: 4 }
pc=2, mem=3
before LocationInternalState { x: 4 }
counting
after LocationInternalState { x: 4 }
pc=3, mem=2
before LocationInternalState { x: 27176560 }
Compiling(Mutex { data: Some(CompiledTrace { entry: 0x7f7b80b00000 }) })
after LocationInternalState { x: 27176560 }
state 3
state 4
```

where (long story!) `state 3` == `Compiled` and `state 4` == `DontTrace`. There is only one place in the current code (in `mt.rs`) that sets `DontTrace` -- and that line of code definitely isn't being called. It looks very much like some bit of the code is "manually" incrementing a `Location` by 1 (hence the move from 3 to 4).

There's nothing obvious which does this increment, so I'm wondering if the most likely cause is that we're inlining parts of the control point into the trace. For example, when we trace multiple opcodes in ykcbf, I wonder if we include the control points for each of the opcodes *as we're tracing it and that also ends up in the compiled machine code*?

Help is very much appreciated, as I am not sure where to start hunting this one down!

vext01:
This change makes compilation occur on a thread and provides infrastructure for tests to serialise compilation via the `YKD_SERIALISE_COMPILATION` environment variable (which is only available when the `c_testing` feature is used).

This is a little more complicated that it perhaps needs to be, as the compilation thread doesn't have the ability to update a `Location`'s state itself when compilation has completed. Perhaps we should look into this shortly, but I think that should be a separate PR.

I've yet to devise a test which doesn't use `YKD_SERIALISE_COMPILATION`. It's not clear if it's possible to test anything meaninful with the full determinism of threads.

All tests pass, with zero changes to them other than adding the env var.

Additional to commit message: this approach adds no extra code to the interpreter loop that would have to be potentially filtered out from the trace later.

Fixes #458

### DIFF
--- a/ykrt/src/location.rs
+++ b/ykrt/src/location.rs
@@ -273,6 +273,59 @@ impl Location {
             }
         }
     }
+
+    /// If the location is in the `Compiling` state, then wait for compilation to complete.
+    ///
+    /// This is used in testing, where we need some way to tame non-determinism in order to write
+    /// meaningful tests.
+    #[cfg(feature = "c_testing")]
+    pub fn block_if_compiling(&self) {
+        let ls = self.load(Ordering::Relaxed);
+
+        // If the location is in the counting state, then compilation cannot be occurring.
+        if ls.is_counting() {
+            return;
+        }
+
+        // A simple spin-lock which exits if either the location has no compilation pending, or if
+        // compilation has just finished for the location.
+        loop {
+            let again = match self.try_lock() {
+                Err(()) => true, // Failed to acquire lock, spin and retry.
+                Ok(ls) => {
+                    if let HotLocation::Compiling(mtx) = unsafe { ls.hot_location() } {
+                        // Compilation is either pending, ongoing, or just completed.
+                        if let Some(ct) = mtx.try_lock() {
+                            // We acquired the lock on the compiled trace storage. This means that
+                            // either:
+                            //
+                            // a) compilation is pending and the compilation thread hasn't yet
+                            // acquired its initial lock on the compiled trace storage. In this
+                            // case, the compiled trace storage remains `None` and we should spin
+                            // more until it becomes `Some`.
+                            //
+                            // b) Compilation has finished and the compilation thread has stored a
+                            // compiled trace and release the lock on its storage. In this case the
+                            // trace storage is `Some`, and we can stop spinning.
+                            ct.is_none()
+                        } else {
+                            // We failed to acquire the lock on the compiled trace storage, so the
+                            // compilation thread must be holding the lock. In this case,
+                            // compilation has started and is ongoing.
+                            true
+                        }
+                    } else {
+                        // If the location is not being compiled, then we don't need to spin.
+                        false
+                    }
+                }
+            };
+            self.unlock();
+            if !again {
+                break;
+            }
+        }
+    }
 }
 
 impl Drop for Location {
--- a/c_tests/Cargo.toml
+++ b/c_tests/Cargo.toml
@@ -18,4 +18,4 @@ lang_tester = "0.7.0"
 once_cell = "1.8.0"
 tempfile = "3.2.0"
 ykcapi = { path = "../ykcapi", features = ["c_testing"] }
-ykrt = { path = "../ykrt", features = ["jit_state_debug"] }
+ykrt = { path = "../ykrt", features = ["jit_state_debug", "c_testing"] }
--- a/c_tests/tests/const_global.c
+++ b/c_tests/tests/const_global.c
@@ -2,6 +2,7 @@
 //   env-var: YKD_PRINT_JITSTATE=1
 // Run-time:
 //   env-var: YKD_PRINT_IR=jit-pre-opt
+//   env-var: YKD_SERIALISE_COMPILATION=1
 //   stderr:
 //     jit-state: start-tracing
 //     i=5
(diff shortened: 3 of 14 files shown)


ykrt/src/location.rs
```diff
@@ -273,6 +273,64 @@ impl Location {
...
+    ///
+    /// This is used in testing, where we need some way to tame non-determinism in order to write
+    /// meaningful tests.
+    ///
+    /// FIXME: Note that we can't simply wait for the location to move into a non-`Compiling`
+    /// state, as the compilation thread is (currently) unable to update the `HotLocation` on it's
+    /// own accord: currently the system relies on an interpreter thread to notice that compilation
+    /// is done and move the location into the next state.
+    #[cfg(feature = "c_testing")]
+    pub fn block_on_pending_compilation(&self) {
```
ltratt, line 287: I think this might be better called `block_if_compiling`?

ykrt/src/location.rs
```diff
@@ -273,6 +273,64 @@ impl Location {
             }
         }
     }
+
+    /// If the location is in the `Compiling` state, then wait for compilation to complete.
+    ///
+    /// This is used in testing, where we need some way to tame non-determinism in order to write
+    /// meaningful tests.
+    ///
+    /// FIXME: Note that we can't simply wait for the location to move into a non-`Compiling`
```
ltratt, line 282: This isn't correct and probably is best simply removed. When a `Location` moves to the non-counting state, we `malloc` memory: that malloc'd chunk is written to (not the `Location`).

ykrt/src/location.rs
vext01: huh, that's odd. I renamed that to `block_on_compilation()` at some point. I must have accidentally undone it.

`block_if_compiling()` is also fine by me, so let's use that.

ykrt/src/location.rs
vext01: We are talking about slightly different things.

Currently, when the compilation thread finishes compiling a trace, it simply stores the trace inside the existing `HotLocation::Compiling(...)`, changing the inner `Option` from `None` to `Some`. Then later an interpreter thread (when inside the control point) notices that the location is `HotLocation::Compiling(Some(ct))` and flips it to `HotLocation::Compiled(ct)`.

What this comment is trying to say is that the waiting logic could be much simpler if the compilation thread itself flipped the `HotLocation` to `Compiled` the moment compilation is completed. The wait logic could then simply spin until the `HotLocation`is no longer `HotLocation::Compiling` any more. Further it can do this without any knowledge of how compiled traces are stored, or of the mutex that guards them.

(This is assuming I can get such a scheme through the borrow checker. I have not yet tried).

ltratt: > What this comment is trying to say is that the waiting logic could be much simpler if the compilation thread itself flipped the HotLocation to Compiled the moment compilation is completed.

I see what you're saying. It might be possible to simplify the state machine, but I don't think this is the right place to document that: I think the FIXME is still best removed.

vext01: I'll remove it if I can look at this simplification next. Deal?

vext01: [I wouldn't have thought it was that difficult, and it simplifies some of that devlish concurrency code]

Anyway, do you find afde2f621 OK? It's not quite what you asked for, but I'm hoping we can compromise.

ykrt/src/location.rs
vext01: Fixed in cbb5403a34

ykrt/src/location.rs
ltratt: > I wouldn't have thought it was that difficult

At a minimum it requires introducing ref counting somewhere.

I'm mostly OK with https://github.com/ykjit/yk/commit/afde2f6216aae88043fa06f147060bffa8db7fe4 but it sounds like a correctness fix when it's a "nice to have". Can we make that clearer?

vext01: Try f1f56c1

vext01: > At a minimum it requires introducing ref counting somewhere.

Yes, my gut said the borrow checker may not like this. I don't know without trying, but I'll leave that to you now.

ltratt:
Please squash.

vext01:
splat.

ltratt:
Can we change the commit title (and where appropriate the body) to say something like "Reenable parallel computation"?

vext01:
Done.

ltratt:
s/meaninful/meaningful/ in the commit. Please force push an update.

vext01:
Done.

Issue #462, opened 2022-01-05
Compilation thread should transition a location to Compiled the moment compilation is complete.


vext01:
Currently, when the compilation thread finishes compiling a trace, it simply stores the trace inside the existing `HotLocation::Compiling(...)`, changing the inner `Option` from `None` to `Some`. Then later an interpreter thread (when inside the control point) notices that the location is `HotLocation::Compiling(Some(ct))` and flips it to `HotLocation::Compiled(ct)`.

It would make more sense if the compilation thread itself flipped the `HotLocation` to `Compiled` the moment compilation is completed.

Also the wait logic in `Location::block_if_compiling()` could then simply spin until the `HotLocation` is no longer `HotLocation::Compiling` any more, and it can do this without any knowledge of how compiled traces are stored, or of the mutex that guards them.

ltratt:
I'll take this one on (it's not critical).

PR #465, opened 2022-01-07
Remove intermediate compiled state

ltratt:
After tidying up the "big" comment mhe meat of this PR is in https://github.com/ykjit/yk/commit/5c36d6b8d8d802d518aef3b2ebe07531adb1961f which fixes https://github.com/ykjit/yk/issues/462.

### DIFF
--- a/ykrt/src/location.rs
+++ b/ykrt/src/location.rs
@@ -42,37 +42,37 @@ pub struct Location {
     //  │             │ start tracing
     //  │             ▼
     //  │           ┌──────────────┐
-    //  |           |              | incomplete  ┌─────────────┐
-    //  │           │   Tracing    │────────────▶|  DontTrace  |
-    //  |           |              |             └─────────────┘
+    //  │           │              │ incomplete  ┌─────────────┐
+    //  │           │   Tracing    │────────────▶│  DontTrace  │
+    //  │           │              │             └─────────────┘
     //  │           └──────────────┘
     //  │             │ start compiling trace
     //  │             │ in thread
     //  │             ▼
     //  │           ┌──────────────┐
-    //  |           |  Compiling   |
+    //  │           │  Compiling   │
     //  │           └──────────────┘
     //  │             │
     //  │             │ trace compiled
     //  │             ▼
     //  │           ┌──────────────┐
-    //  └───────────│   Compiled   |
+    //  └───────────│   Compiled   │
     //              └──────────────┘
     //
-    // We hope that a Location soon reaches the Compiled state (aka "the happy state") and stays
+    // We hope that a Location soon reaches the `Compiled` state (aka "the happy state") and stays
     // there.
     //
-    // The state machine is encoded in a usize in a not-entirely-simple way, as we don't want to
+    // The state machine is encoded in a `usize` in a not-entirely-simple way, as we don't want to
     // allocate any memory for Locations that do not become hot. The layout is as follows (on a 64
     // bit machine):
     //
     //   bit(s) | 63..3   | 2           | 1         | 0
     //          | payload | IS_COUNTING | IS_PARKED | IS_LOCKED
     //
-    // In the Counting state, IS_COUNTING is set to 1, and IS_PARKED and IS_LOCKED are unused and
-    // must remain set at 0. The payload representing the count is incremented locklessly. All
-    // other states have IS_COUNTING set to 0 and the payload is the address of a boxed
-    // HotLocation, access to which is controlled by IS_LOCKED.
+    // In the `Counting` state, `IS_COUNTING` is set to 1, and `IS_PARKED` and `IS_LOCKED` are
+    // unused and must remain set at 0. The payload representing the count is incremented
+    // locklessly. All other states have `IS_COUNTING` set to 0 and the payload is the address of a
+    // `Box<HotLocation>`, access to which is controlled by `IS_LOCKED`.
     //
     // The possible combinations of the counting and mutex bits are thus as follows:
     //
--- a/ykrt/src/location.rs
+++ b/ykrt/src/location.rs
@@ -76,18 +76,19 @@ pub struct Location {
     //
     // The possible combinations of the counting and mutex bits are thus as follows:
     //
-    //   payload       | IS_COUNTING | IS_PARKED | IS_LOCKED | Notes
-    //   --------------+-----------+---------+---------+-----------------------------------
-    //   <count>       | 1           | 0         | 0         | Start state
-    //                 | 1           | 0         | 1         | Illegal state
-    //                 | 1           | 1         | 0         | Illegal state
-    //                 | 1           | 1         | 1         | Illegal state
-    //   <HotLocation> | 0           | 0         | 0         | Not locked, no-one waiting
-    //   <HotLocation> | 0           | 0         | 1         | Locked, no thread(s) waiting
-    //   <HotLocation> | 0           | 1         | 0         | Not locked, thread(s) waiting
-    //   <HotLocation> | 0           | 1         | 1         | Locked, thread(s) waiting
+    //   payload          | IS_COUNTING | IS_PARKED | IS_LOCKED | Notes
+    //   -----------------+-----------+---------+---------+-----------------------------------
+    //   count            | 1           | 0         | 0         | Start state
+    //                    | 1           | 0         | 1         | Illegal state
+    //                    | 1           | 1         | 0         | Illegal state
+    //                    | 1           | 1         | 1         | Illegal state
+    //   Box<HotLocation> | 0           | 0         | 0         | Not locked, no-one waiting
+    //   Box<HotLocation> | 0           | 0         | 1         | Locked, no thread(s) waiting
+    //   Box<HotLocation> | 0           | 1         | 0         | Not locked, thread(s) waiting
+    //   Box<HotLocation> | 0           | 1         | 1         | Locked, thread(s) waiting
     //
-    // where `<count>` is an integer and `<HotLocation>` is a boxed `HotLocation` enum.
+    // where `count` is an integer and `Box<HotLocation>` is a pointer to a `malloc`ed block of
+    // memory in the heap containing a `HotLocation` enum.
     //
     // The precise semantics of locking and, in particular, parking are subtle: interested readers
     // are directed to https://github.com/Amanieu/parking_lot/blob/master/src/raw_mutex.rs#L33 for
--- a/ykrt/src/location.rs
+++ b/ykrt/src/location.rs
@@ -9,7 +9,6 @@ use std::{
 };
 
 use crate::mt::HotThreshold;
-use parking_lot::Mutex;
 use parking_lot_core::{
     park, unpark_one, ParkResult, SpinWait, UnparkResult, UnparkToken, DEFAULT_PARK_TOKEN,
 };
@@ -49,9 +48,9 @@ pub struct Location {
     //  │             │ start compiling trace
     //  │             │ in thread
     //  │             ▼
-    //  │           ┌──────────────┐
-    //  │           │  Compiling   │
-    //  │           └──────────────┘
+    //  │           ┌──────────────┐             ┌───────────┐
+    //  │           │  Compiling   │────────────▶│  Dropped  │
+    //  │           └──────────────┘             └───────────┘
     //  │             │
     //  │             │ trace compiled
     //  │             ▼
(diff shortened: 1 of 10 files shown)


ykrt/src/location.rs
```diff
@@ -104,6 +104,12 @@ impl Location {
         }
     }
 
+    pub(super) unsafe fn from_locationinner(ls: LocationInner) -> Self {
```
bjorn3, line 107: `from_location_inner` looks a bit nicer to me.

ykrt/src/location.rs
```diff
@@ -49,9 +48,9 @@ pub struct Location {
     //  │             │ start compiling trace
     //  │             │ in thread
     //  │             ▼
-    //  │           ┌──────────────┐
-    //  │           │  Compiling   │
-    //  │           └──────────────┘
+    //  │           ┌──────────────┐             ┌───────────┐
+    //  │           │  Compiling   │────────────▶│  Dropped  │
+    //  │           └──────────────┘             └───────────┘
```
vext01, line 53: Dropped means that the `HotLocation` was dropped, not the `Location` itself, right?

Should there be edges from other states into `Dropped` too? E.g. `Compiled -> Dropped`?

ykrt/src/location.rs
vext01: It's because `locationinner` is corresponding to the `LocationInner` type. I don't feel strongly about this one.

ykrt/src/location.rs
```diff
@@ -263,6 +268,7 @@ impl Location {
     /// Try obtaining the lock, returning the new `State` if successful.
     pub(super) fn try_lock(&self) -> Result<LocationInner, ()> {
         let mut ls = self.load(Ordering::Relaxed);
+        // FIXME: this could be in the counting state
```
vext01, line 271: What is the consequence of this though? Can we add some more detail?

ykrt/src/location.rs
```diff
@@ -294,27 +300,8 @@ impl Location {
...
-                            // compiled trace and release the lock on its storage. In this case the
-                            // trace storage is `Some`, and we can stop spinning.
-                            ct.is_none()
-                        } else {
-                            // We failed to acquire the lock on the compiled trace storage, so the
-                            // compilation thread must be holding the lock. In this case,
-                            // compilation has started and is ongoing.
-                            true
-                        }
+                    if let HotLocation::Compiling = unsafe { ls.hot_location() } {
```
vext01, line 303: Fantastic!

ykrt/src/location.rs
```diff
@@ -334,9 +321,25 @@ impl Drop for Location {
         let ls = self.load(Ordering::Relaxed);
         if !ls.is_counting() {
             debug_assert!(!ls.is_locked());
-            let hr = unsafe { ls.hot_location() };
-            unsafe {
-                Box::from_raw(hr);
+            self.lock().unwrap();
```
vext01, line 324: swap lines 324 and 325 to reduce the critical section slightly?

vext01: or do you require an absolutely up-to-date `hl` for correctness?

ykrt/src/location.rs
```diff
@@ -334,9 +321,25 @@ impl Drop for Location {
...
+            let hl = unsafe { ls.hot_location() };
+            if let HotLocation::Compiled(_) = hl {
+                // FIXME: we can't drop this memory as another thread may still be executing the
```
vext01, line 327: Currently the trace code would not be freed if you dropped the `CompiledTrace`. Not sure if that changes anything?

ykrt/src/location.rs
```diff
@@ -334,9 +321,25 @@ impl Drop for Location {
...
+                // trace that's pointed to. There should be a ref count that we decrement and free
+                // memory if it reaches zero.
+                self.unlock();
+            } else if let HotLocation::Compiling = hl {
+                // Another thread has a pointer to the underlying `HotLocation`, so we have to
+                // signal to that thread that they need to free the `HotLocation`.
+                *hl = HotLocation::Dropped;
```
vext01, line 334: I take it this scheme is devised so as to avoid using a `RefCell` (which would be a simpler, but less-performant way of doing things).

When would the user drop a `Location`, besides when the interpreter is ending? Is it worth the added complexity?

ykrt/src/location.rs
```diff
@@ -334,9 +321,25 @@ impl Drop for Location {
...
+                self.unlock();
+            } else if let HotLocation::DontTrace | HotLocation::Tracing(_) = hl {
+                self.unlock();
```
vext01, line 337: Are the repeated unlocks necessary? I assume you are trying to make the critical section small?

ykrt/src/location.rs
```diff
@@ -457,8 +460,16 @@ pub(super) struct ThreadIdInner;
 /// A `Location`'s non-counting states.
 #[derive(EnumDiscriminants)]
 pub(super) enum HotLocation {
+    /// Points to executable machine code that can be executed instead of the interpreter for this
+    /// HotLocation.
     Compiled(Box<CompiledTrace>),
-    Compiling(Arc<Mutex<Option<Box<CompiledTrace>>>>),
+    /// This HotLocation is being compiled in another thread.
+    Compiling,
+    /// Whilst this HotLocation was being compiled in another thread, the [Location] was dropped.
```
vext01, line 468: why `[Location]`?

bjorn3: That turns it into a link to the docs for the `Location` type.

ykrt/src/mt.rs
```diff
@@ -342,13 +322,28 @@ impl MT {
...
+    fn queue_compile_job(&self, trace: IRTrace, hl_ptr: *const HotLocation) {
+        let hl_ptr = hl_ptr as usize;
         let f = Box::new(move || {
             let code_ptr = trace.compile();
-            *mtx.lock() = Some(Box::new(CompiledTrace::new(code_ptr)));
-            // FIXME: although we've now put the compiled trace into the mutex, there's no
-            // guarantee that the Location for which we're compiling will ever be executed
-            // again. In such a case, the memory has, in essence, leaked.
+            let ct = Box::new(CompiledTrace::new(code_ptr));
+            let tmp_ls = LocationInner::new().with_hotlocation(hl_ptr as *mut HotLocation);
```
vext01, line 330: This seems subtle. Can we add a comment explaining what is happening?

I *think* we use a temporary location with a copy of the heap pointer (whose contents may be changed between compilation starting and completing), so that the original location can still be dropped without the compilation thread being hosed. Right?

ykrt/src/location.rs
vext01: Ah. I did not know that. So should `HotLocation` (and all other type names throughout the repo) also have the square brackets?

ykrt/src/location.rs
ltratt: > Dropped means that the HotLocation was dropped, not the Location itself, right?

They're a bit interlinked, but in this case it means the `Location` was dropped and that the `HotLocation` now has the status `Dropped`.

ykrt/src/location.rs
ltratt: This is a reminder for another PR that I just happened to spot while here.

ykrt/src/location.rs
ltratt: That would be incorrect: you can't deref the `hot_location` without locking.

ykrt/src/location.rs
ltratt: That's a secondary issue :)

ykrt/src/location.rs
ltratt: `RefCell`s aren't `Sync` so you'd need to add a `Mutex` (or similar) around them.

ykrt/src/location.rs
ltratt: > I assume you are trying to make the critical section small?

Exactly! Since we know that `Location`s will be highly contended, reducing the critical sections is very important.

ykrt/src/location.rs
ltratt: If you generate `cargo doc`s from this, that'll turn into a clickable link to `Location`.

ykrt/src/location.rs
vext01: Uh. Did I mean to say `Arc<RefCell>`?

Basically, why add this complexity if the only time a location would be dropped is a time where you no longer care about performance?

ykrt/src/mt.rs
ltratt: Good point. Fixed in a77b95f.

ykrt/src/location.rs
vext01: And the latter query?

> Should there be edges from other states into Dropped too? E.g. Compiled -> Dropped?

ykrt/src/location.rs
ltratt: > Did I mean to say Arc<RefCell>?

Only you know what you meant to say :p

> Basically, why add this complexity if the only time a location would be dropped is a time where you no longer care about performance?

If I was to put back an `Arc` or `Mutex`, we'd end up with the intermediate state where a `Location` has compiled but isn't yet in the `Compiling` state. In a sense this whole PR is about getting rid of that.

ykrt/src/location.rs
ltratt: > Should there be edges from other states into Dropped too? E.g. Compiled -> Dropped?

No (at least... not yet... that might change in the future).

ykrt/src/location.rs
vext01: Can the deref and locking operations be combined to prevent programming errors? i.e. doing a deref does an implicit lock? I guess that become annoying when you have different modes of locking (lock and try_lock).

Maybe I talked myself out of it.

ykrt/src/location.rs
ltratt: The comment is complete IMHO. I don't think we should obsess about this: it's a bug, there's a reminder, it's not related to this PR, best to move on.

ykrt/src/location.rs
ltratt: > Can the deref and locking operations be combined to prevent programming errors?

Not unless you know something that I don't ;)

ykrt/src/location.rs
vext01: OK. So when you drop a location that is already `Compiled` it just ceases to be. I think I understand.

ykrt/src/location.rs
vext01: Yes, but why is `Location` special? Do the other types not get to have a clickable link too?

ltratt: There's not much advantage to having a reference to `Self` be clickable.

vext01:
Please squash.

ltratt:
Squashed.

vext01:
Looks like the incorrect live values were extracted from the stack. May need @ptersilie to move this forward?

ltratt:
I think I need to rebase this against master to pick up the fix that (IIRC) @ptersilie needed in https://github.com/ykjit/yk/commit/e76de0d01316cd02a46981e96b37f967a15c9ad0. OK for me to force push an update?

vext01:
Go ahead.

ltratt:
Rebased.

ltratt:
I can replicate the failure with `cargo test --release`. Investigating.

ltratt:
@vext01 I think we can try again while @ptersilie looks at https://github.com/ykjit/yk/pull/465/commits/90eee70f5c823a94faec12dd1b4d0c04e93ead2c as a separate activity. I think that fixing that should be done in another PR.

ltratt:
I'm officially confused: the `Indirect` values change, sometimes increasing, sometimes decreasing. @ptersilie I think is going to look into this?

ltratt:
@vext01 I've looked a bit and, while we don't know yet, but a plausible guess is that this PR accidentally highlights incorrect assumptions in the guard failure tests about what the optimiser can do (given that the failures only happen with `cargo test --release`). I think you should re:r+ this PR and we can fix those problems in a parallel PR: I don't think we should hold this PR up in the interim.

ptersilie:
It looks like we just had to adjust the expected values since the number of iterations in some tests were changed. All seems to be working now. I couldn't see any differences between `cargo test` and `cargo test --release`.

That last commit needs merging though.

ltratt:
I now have the same commit locally :)

@vext01 OK for me to squash?

ltratt:
Squashed -- and thanks @ptersilie !

PR #468, opened 2022-01-10
Remove the SERIALISE_COMPILATION spinlock.

ltratt:
This commit moves the "serialise computation" concept from ykcapi to ykrt, where we change `queue_compile_job` to block compilation until it's complete. Thus when `queue_compile_job` returns, its not only compiled a `Location`, but it's updated it to the `Compiled` state too.

As well as simplifying the code, removing the spinlock has a noticeable performance impact: roughly speaking, the test suite's wall-clock time decreases by about 40% (from ~14s to ~8.5s) and the total CPU time (across all cores) decreases by about 50%.

### DIFF
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -11,20 +11,12 @@
 #![feature(once_cell)]
 
 use std::convert::TryInto;
-#[cfg(feature = "c_testing")]
-use std::env;
 use std::ffi::c_void;
-#[cfg(feature = "c_testing")]
-use std::lazy::SyncLazy;
 use std::process;
 use std::{ptr, slice};
 use ykrt::{Location, MT};
 use yksmp::{Location as SMLocation, StackMapParser};
 use ykutil;
-
-#[cfg(feature = "c_testing")]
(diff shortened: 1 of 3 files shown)


vext01:
`s/SERIALISE_COMPUTATION/SERIALISE_COMPILATION/g` in PR title, description and commit message.

ltratt:
OK for me to force push an update?

vext01:
Go ahead.

ltratt:
Force pushed.

vext01:
So this is a "no functional change" type of affair? It doesn't alter at what iteration we first enter jit code or anything like that?

ltratt:
Correct.

Note that I have spotted a big problem over the weekend that I'm amazed I hadn't appreciated before: the entire test suite for the location state machine was removed and not put back in any form. So none of this stuff is testable, which is highly embarrassing. My next PR will get us to the point that we can actually test this stuff again. In the meantime, we should all dunk ourselves in cold water for not adding a test suite back!

vext01:
Sometimes you have to delete stuff to move forward. I think the assumption was that the old tests were not going to be usable. If they are then great, we can pull them out of git history and no harm done.

bors r+

PR #469, opened 2022-01-10
Don't execute code in the location transition function

ltratt:
This PR is a vital step in allowing us to test the location state machine.

### DIFF
--- a/c_tests/Cargo.toml
+++ b/c_tests/Cargo.toml
@@ -16,5 +16,5 @@ regex = "1.5.4"
 [dev-dependencies]
 lang_tester = "0.7.0"
 tempfile = "3.2.0"
-ykcapi = { path = "../ykcapi", features = ["c_testing"] }
-ykrt = { path = "../ykrt", features = ["jit_state_debug", "c_testing"] }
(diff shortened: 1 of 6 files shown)


ltratt:
> Does starting/stopping traces, and spawning compilation threads not need to be moved outside too?

Possibly, but I'm not sure yet -- and I know the change in this PR *has* to be done. If we're happy with this PR, I think doing the other things will become much easier once we've established a precedent.

vext01:
I'm happy with it, but I don't see how you can test the state machine in isolation if those events are baked in to state transitions.

ltratt:
You're probably right that tracing needs to be moved out, though compiling in a thread probably doesn't need to be. But I think we can put off thinking about those until I actually start adding a test suite in. Clearly, this PR is a necessary (but perhaps not sufficient) step towards doing that.

vext01:
> though compiling in a thread probably doesn't need to be.

But if you leave the compilation in, then you have the requirement to have something to trace and compile. That's going to be annoying if all you want to test is the state machine, isn't it?

> Clearly, this PR is a necessary (but perhaps not sufficient) step towards doing that.

Of course, we are in agreement there.

Any squashing to do?

ltratt:
No squashing needed.

PR #471, opened 2022-01-11
Resolve linkage issue.

### LINKED ISSUE #470: Tests don't build
Any non-trivial test in `ykrt/src/mt.rs` such as:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn hot_threshold_passed() {
        let hot_thrsh: HotThreshold = 1500;
        let mt = MT::global();
        let loc = Location::new();
        for i in 0..hot_thrsh {
            MT::transition_location(&loc);
        }
    }
}
```

causes a linker error:

```
error: linking with `cc` failed: exit status: 1
  |
  = note: "cc" "-m64" "/home/ltratt/yk/target/debug/deps/ykrt-e549a82218bf7b47.13mjq23so3byahxg.rcgu.o"
<vast swathes of stuff cut>
          /home/ltratt/yk/ykllvmwrap/src/ykllvmwrap.cc:186: undefined reference to `__ykutil_get_llvmbc_section'
          collect2: error: ld returned 1 exit status
          
  = help: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
  = note: use the `-l` flag to specify native libraries to link
  = note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-link-libkindname)

warning: `ykrt` (lib test) generated 2 warnings
error: could not compile `ykrt` due to previous error; 2 warnings emitted
```

I wonder if this is something to do with the `thread_local!` we use?

vext01:
The first commit fixes #470. The second is a small tidy up. No functional change.

[Your test still doesn't work, it fails with `SIGILL`, but at least you don't get the linker issue. My guess would be that because you are driving `transition_location()` manually, when a trace eventually gets compiled, the trace compiler is going to look for a control point, and not find one, thus probably crashing something. I still think you need to move all of the tracing-related activities out of `transition_location` if you want to test the state machine, and only the state machine]

### DIFF
--- a/ykcapi/Cargo.toml
+++ b/ykcapi/Cargo.toml
@@ -13,7 +13,6 @@ libc = "0.2.97"
 ykllvmwrap = { path = "../ykllvmwrap" }
 ykrt = { path = "../ykrt" }
 yktrace = { path = "../yktrace" }
-ykutil = { path = "../ykutil" }
 yksmp = { path = "../yksmp" }
 
 [features]
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -16,7 +16,6 @@ use std::process;
 use std::{ptr, slice};
 use ykrt::{Location, MT};
 use yksmp::{Location as SMLocation, StackMapParser};
-use ykutil;
 
 // The "dummy control point" that is replaced in an LLVM pass.
 #[no_mangle]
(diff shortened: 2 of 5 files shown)


ltratt:
Thanks!

bors r+

PR #473, opened 2022-01-18
Add test for missing control point.

vext01:
Low hanging fruit in "operation improve test suite".

Second commit is just removing unrelated dead code.

Can't merge yet, as CI broken (PT not working in docker).

### DIFF
--- a/c_tests/tests/missing_ctrl_pt.c
+++ b/c_tests/tests/missing_ctrl_pt.c
@@ -1,0 +1,16 @@
+// Compiler:
+//   status: error
+//   stderr:
+//     ...
+//     ...error: ykllvm couldn't find the call to `yk_control_point()`
+//     ...
+
+// Check that the system crashes if the control point is missing.
+
+#include <stdlib.h>
+#include <stdbool.h>
+
+int main(int argc, char **argv) {
+  // Control point is missing!
+  return (EXIT_SUCCESS);
+}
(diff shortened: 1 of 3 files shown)


c_tests/tests/missing_ctrl_pt.c
```diff
@@ -0,0 +1,18 @@
...
+//     ld.lld: error: ykllvm couldn't find the call to `yk_control_point()`
+//     ...
+
+// Check that the system crashes if the control point is missing.
+
+#include <stdlib.h>
+#include <stdbool.h>
+
+int main(int argc, char **argv) {
+  while (true) {
```
ltratt, line 14: Do we even need the `while` loop to produce the "no control point" error?

c_tests/tests/missing_ctrl_pt.c
```diff
@@ -0,0 +1,18 @@
+// Compiler:
+//   status: error
+//   stderr:
+//     ...
+//     ld.lld: error: ykllvm couldn't find the call to `yk_control_point()`
```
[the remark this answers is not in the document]
vext01: Only `lld` understands LLVM LTO, but I guess it could go by a different executable name, so fixed in ec142ee1e6b9.

c_tests/tests/missing_ctrl_pt.c
ltratt: I think it would be clearer not to have the loop in there: it suggests we test for the existence of a loop with a control point when, I think, we test only for the existence of a control point.

[Maybe a good test for the next PR is "control point not in a loop"?!]

vext01:
squashed and signed

PR #474, opened 2022-01-19
Start work on the stopgap interpreter.

ptersilie:
**This is a draft PR to discuss the design of the stopgap interpreter. It is not ready
for merging yet as there's still quite a few IR instructions to be implemented.**

This PR sets up the skeleton of the stopgap interpreter. The interpreter
interacts with the LLVM module via llvm-sys rather than using a more abstracted
(and user-friendly) interface like llvm-ir. The reason is that llvm-ir parses
modules in its entirety, even parts of the module the stopgap interpreter won't
touch. Using llvm-sys we only parse what we actually need to get back to the
control point.

In order to initialise the stopgap interpreter we need to create a mapping from
stackmap values to variables (i.e. instructions) in the AOT module. This is
being setup in `jitmodbuilder.cc` during guard creation and then passed through
to the stopgap interpreter via the deoptimisation call. We also need to pass
the position in the AOT module from where interpretation needs to start, which
we previously didn't do either.

### DIFF
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -17,6 +17,11 @@ use std::{ptr, slice};
 use ykrt::{HotThreshold, Location, MT};
 use yksmp::{Location as SMLocation, StackMapParser};
 
+mod sginterp;
+use sginterp::{SGInterp, SGValue};
+
+mod llvmapihelper;
+
 // The "dummy control point" that is replaced in an LLVM pass.
 #[no_mangle]
 pub extern "C" fn yk_control_point(_loc: *mut Location) {
@@ -95,21 +100,57 @@ impl Registers {
     }
 }
 
+#[derive(Debug)]
+#[repr(C)]
+struct AOTVar {
+    bbidx: u32,
+    instridx: u32,
+    fname: *const i8,
+}
+
+#[derive(Debug)]
+#[repr(C)]
+pub struct AOTMap {
+    addr: *const c_void,
+    size: usize,
+}
+
+#[derive(Debug)]
+#[repr(C)]
+pub struct CurPos {
+    bbidx: u32,
+    instridx: u32,
+    fname: *const i8,
+}
+
 /// Parses the stackmap and saved registers from the given address (i.e. the return address of the
 /// deoptimisation call).
-/// FIXME: Until we have the stopgap interpreter we simply print out the values we find.
 #[cfg(target_arch = "x86_64")]
 #[no_mangle]
 pub extern "C" fn yk_stopgap(
     sm_addr: *const c_void,
     sm_size: usize,
+    aotmap: &AOTMap,
+    curpos: &CurPos,
     retaddr: usize,
     rsp: *const c_void,
 ) {
     // FIXME: remove once we have a stopgap interpreter.
     eprintln!("jit-state: stopgap");
+
+    // Parse AOTMap.
+    let aotmap = unsafe { slice::from_raw_parts(aotmap.addr as *const AOTVar, aotmap.size) };
+
     // Restore saved registers from the stack.
     let registers = Registers::from_ptr(rsp);
+
+    let mut sginterp = unsafe {
+        SGInterp::new(
+            curpos.bbidx,
+            curpos.instridx,
+            std::ffi::CStr::from_ptr(curpos.fname),
+        )
+    };
 
     // Parse the stackmap.
     let slice = unsafe { slice::from_raw_parts(sm_addr as *mut u8, sm_size) };
@@ -117,7 +158,8 @@ pub extern "C" fn yk_stopgap(
     let locs = map.get(&retaddr.try_into().unwrap()).unwrap();
 
     // Extract live values from the stackmap.
-    for l in locs {
+    // Skip first 3 locations as they don't relate to any of our live variables.
+    for (i, l) in locs.iter().skip(3).enumerate() {
         match l {
             SMLocation::Register(reg, size) => {
                 // FIXME: remove once we have a stopgap interpreter.
@@ -148,12 +190,30 @@ pub extern "C" fn yk_stopgap(
                     8 => unsafe { ptr::read::<u64>(addr as *mut u64) as u64 },
                     _ => unreachable!(),
                 };
+                let aot = &aotmap[i];
+                unsafe {
+                    sginterp.init_live(
+                        aot.bbidx,
+                        aot.instridx,
+                        std::ffi::CStr::from_ptr(aot.fname),
+                        SGValue::U64(v),
+                    );
+                }
                 // FIXME: remove once we have a stopgap interpreter.
-                eprintln!("Indirect: {} ({} {})", v, reg, off);
+                eprintln!("Indirect: {} ({} {} {})", v, reg, off, size);
             }
             SMLocation::Constant(v) => {
                 // FIXME: remove once we have a stopgap interpreter.
                 eprintln!("Constant: {}", v);
+                let aot = &aotmap[i];
+                unsafe {
+                    sginterp.init_live(
+                        aot.bbidx,
+                        aot.instridx,
+                        std::ffi::CStr::from_ptr(aot.fname),
+                        SGValue::U32(*v),
+                    );
+                }
             }
             SMLocation::LargeConstant(v) => {
                 // FIXME: remove once we have a stopgap interpreter.
@@ -161,7 +221,7 @@ pub extern "C" fn yk_stopgap(
             }
         }
     }
-    // FIXME: Initialise stopgap interpreter here.
+    unsafe { sginterp.interpret() };
     process::exit(0);
 }
 
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -25,6 +25,7 @@ use sginterp::{SGInterp, SGValue};
 const SM_REC_HEADER: usize = 3;
 
 mod llvmapihelper;
+mod llvmwrap;
 
 // The "dummy control point" that is replaced in an LLVM pass.
 #[no_mangle]
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -20,6 +20,10 @@ use yksmp::{Location as SMLocation, StackMapParser};
 mod sginterp;
 use sginterp::{SGInterp, SGValue};
 
+/// The first three locations of an LLVM stackmap record, according to the source, are CC, Flags,
+/// Num Deopts, which need to be skipped when mapping the stackmap values back to AOT variables.
+const SM_REC_HEADER: usize = 3;
+
 mod llvmapihelper;
 
 // The "dummy control point" that is replaced in an LLVM pass.
@@ -159,7 +163,7 @@ pub extern "C" fn yk_stopgap(
 
     // Extract live values from the stackmap.
     // Skip first 3 locations as they don't relate to any of our live variables.
-    for (i, l) in locs.iter().skip(3).enumerate() {
+    for (i, l) in locs.iter().skip(SM_REC_HEADER).enumerate() {
         match l {
             SMLocation::Register(reg, size) => {
                 // FIXME: remove once we have a stopgap interpreter.
--- a/ykcapi/src/llvmapihelper.rs
+++ b/ykcapi/src/llvmapihelper.rs
@@ -19,7 +19,7 @@ pub unsafe fn get_instruction(bb: LLVMBasicBlockRef, instridx: u32) -> LLVMValue
     instr
 }
 
-pub unsafe fn parse_const(c: LLVMValueRef) -> SGValue {
+pub unsafe fn llvm_const_to_sgvalue(c: LLVMValueRef) -> SGValue {
     let ty = LLVMTypeOf(c);
     let kind = LLVMGetTypeKind(ty);
     match kind {
--- a/ykcapi/src/llvmapihelper.rs
+++ b/ykcapi/src/llvmapihelper.rs
@@ -1,0 +1,37 @@
+use crate::sginterp::SGValue;
+use llvm_sys::core::*;
+use llvm_sys::prelude::{LLVMBasicBlockRef, LLVMValueRef};
+use llvm_sys::{LLVMTypeKind};
+
+pub unsafe fn get_basic_block(func: LLVMValueRef, bbidx: u32) -> LLVMBasicBlockRef {
+    let mut bb = LLVMGetFirstBasicBlock(func);
+    for _ in 0..bbidx {
+        bb = LLVMGetNextBasicBlock(bb);
+    }
+    bb
+}
+
+pub unsafe fn get_instruction(bb: LLVMBasicBlockRef, instridx: u32) -> LLVMValueRef {
+    let mut instr = LLVMGetFirstInstruction(bb);
+    for _ in 0..instridx {
+        instr = LLVMGetNextInstruction(instr);
+    }
+    instr
+}
+
+pub unsafe fn parse_const(c: LLVMValueRef) -> SGValue {
+    let ty = LLVMTypeOf(c);
+    let kind = LLVMGetTypeKind(ty);
+    match kind {
+        LLVMTypeKind::LLVMIntegerTypeKind => {
+            let width = LLVMGetIntTypeWidth(ty);
+            let val = LLVMConstIntGetZExtValue(c) as u64;
+            match width {
+                32 => SGValue::U32(val as u32),
+                64 => SGValue::U64(val),
+                _ => todo!(),
+            }
+        }
+        _ => todo!(),
+    }
+}
--- a/ykcapi/src/sginterp.rs
+++ b/ykcapi/src/sginterp.rs
@@ -1,0 +1,142 @@
+use llvm_sys::bit_reader::LLVMParseBitcodeInContext2;
+use llvm_sys::core::*;
+use llvm_sys::prelude::{LLVMModuleRef, LLVMValueRef};
+use llvm_sys::{LLVMOpcode};
+use std::collections::HashMap;
+use std::ffi::CStr;
+use std::mem::MaybeUninit;
+
+use crate::llvmapihelper::{self};
+
+/// Stopgap interpreter values.
+#[derive(Debug)]
+pub enum SGValue {
+    U32(u32),
+    U64(u64),
+}
+
+/// A frame holding live variables.
+struct Frame {
+    vars: HashMap<LLVMValueRef, SGValue>,
+}
+
+impl Frame {
+    fn new() -> Frame {
+        Frame {
+            vars: HashMap::new(),
+        }
+    }
+
+    /// Get the value of the variable `key` in this frame.
+    fn get(&self, key: &LLVMValueRef) -> Option<&SGValue> {
+        self.vars.get(key)
+    }
+
+    /// Insert new variable into this frame.
+    fn add(&mut self, key: LLVMValueRef, val: SGValue) {
+        self.vars.insert(key, val);
+    }
+}
+
+/// The stopgap interpreter. Used during guard failures to get back to the control point by
+/// interpreting LLVM IR.
+pub struct SGInterp {
+    /// LLVM IR module we are interpreting.
+    module: LLVMModuleRef,
+    /// Current frames.
+    frames: Vec<Frame>,
+    /// Current instruction being interpreted.
+    pc: LLVMValueRef,
+}
+
+impl SGInterp {
+    /// Create a new stopgap interpreter and initialise it to start interpretation at the location
+    /// given by a basic block index, instruction index, and function name.
+    /// FIXME: Support initalisation of multiple frames.
+    pub unsafe fn new(bbidx: u32, instridx: u32, fname: &CStr) -> SGInterp {
+        // Get AOT module IR and parse it.
+        let (addr, size) = ykutil::obj::llvmbc_section();
+        let membuf = LLVMCreateMemoryBufferWithMemoryRange(
+            addr as *const i8,
+            size,
+            "".as_ptr() as *const i8,
+            0,
+        );
+        let context = LLVMContextCreate();
+        let mut module: MaybeUninit<LLVMModuleRef> = MaybeUninit::uninit();
+        let module = {
+            LLVMParseBitcodeInContext2(context, membuf, module.as_mut_ptr());
+            module.assume_init()
+        };
+        // Create and initialise stop gap interpreter.
+        let func = LLVMGetNamedFunction(module, fname.as_ptr());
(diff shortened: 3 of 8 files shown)


ykcapi/src/lib.rs
```diff
@@ -95,29 +100,66 @@ impl Registers {
     }
 }
 
+#[derive(Debug)]
+#[repr(C)]
+struct AOTVar {
+    bbidx: u32,
+    instridx: u32,
+    fname: *const i8,
```
ltratt, line 108: Should this be `*const os::raw::c_char`?

ykcapi/src/lib.rs
```diff
@@ -95,29 +100,66 @@ impl Registers {
     }
 }
 
+#[derive(Debug)]
+#[repr(C)]
+struct AOTVar {
+    bbidx: u32,
```
ltratt, line 106: Are these types always `u32` no matter the platform?

ykcapi/src/lib.rs
```diff
@@ -95,29 +100,66 @@ impl Registers {
...
+    };
+
     // Parse the stackmap.
     let slice = unsafe { slice::from_raw_parts(sm_addr as *mut u8, sm_size) };
     let map = StackMapParser::parse(slice).unwrap();
     let locs = map.get(&retaddr.try_into().unwrap()).unwrap();
 
     // Extract live values from the stackmap.
-    for l in locs {
+    // Skip first 3 locations as they don't relate to any of our live variables.
```
ltratt, line 161: This is too dangerous to be an in-code constant: it must be a named constant so that we can search for it later if/when it needs to be changed.

ykcapi/src/lib.rs
```diff
@@ -148,20 +190,38 @@ pub extern "C" fn yk_stopgap(
...
+                    sginterp.init_live(
+                        aot.bbidx,
+                        aot.instridx,
+                        std::ffi::CStr::from_ptr(aot.fname),
+                        SGValue::U64(v),
+                    );
+                }
                 // FIXME: remove once we have a stopgap interpreter.
-                eprintln!("Indirect: {} ({} {})", v, reg, off);
+                eprintln!("Indirect: {} ({} {} {})", v, reg, off, size);
```
ltratt, line 203: Since we now have a stopgap interpreter, shouldn't this disappear?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -185,6 +185,10 @@ class JITModBuilder {
   // The LLVM type for a C `int` on the current machine.
   Type *IntTy;
 
+  // Map JIT instruction to basic block index and instruction index of the
+  // corresponding AOT instruction.
+  std::map<Value *, std::tuple<size_t, size_t, Instruction *>> AOTMap;
```
vext01, line 190: What's the third element in the tuple? The comment mentions only two elements.

vext01: Looks like the AOT instruction?

ykcapi/src/llvmapihelper.rs
```diff
@@ -0,0 +1,37 @@
...
+
+pub unsafe fn get_instruction(bb: LLVMBasicBlockRef, instridx: u32) -> LLVMValueRef {
+    let mut instr = LLVMGetFirstInstruction(bb);
+    for _ in 0..instridx {
+        instr = LLVMGetNextInstruction(instr);
+    }
+    instr
+}
+
+pub unsafe fn parse_const(c: LLVMValueRef) -> SGValue {
```
ltratt, line 22: This function doesn't look like it does parsing?

ykcapi/src/sginterp.rs
```diff
@@ -0,0 +1,142 @@
...
+use crate::llvmapihelper::{self};
+
+/// Stopgap interpreter values.
+#[derive(Debug)]
+pub enum SGValue {
+    U32(u32),
+    U64(u64),
+}
+
+/// A frame holding live variables.
```
ltratt, line 18: This is going to be very slow: can't we convert variables to indexes in a `Vec` rather than IDs in a `HashMap`?

ykcapi/src/sginterp.rs
```diff
@@ -0,0 +1,142 @@
...
+use llvm_sys::{LLVMOpcode};
+use std::collections::HashMap;
+use std::ffi::CStr;
+use std::mem::MaybeUninit;
+
+use crate::llvmapihelper::{self};
+
+/// Stopgap interpreter values.
+#[derive(Debug)]
+pub enum SGValue {
```
ltratt, line 13: Are the only two types of possible value `U32` and `U64`?

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Oh, that's the original AOT instruction. Will fix.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -608,6 +704,7 @@ class JITModBuilder {
     // module, we must remap them to point to new values in the JIT module.
     llvm::RemapInstruction(NewInst, VMap, RF_NoModuleLevelChanges);
     VMap[&*I] = NewInst;
+    insertAOTMap(I, NewInst, CurBBIdx, CurInstrIdx);
```
vext01, line 707: We talked about using a method to update the VMap and the AOTMap at once in the same place. Is it still possible?

ykcapi/src/lib.rs
vext01, line 161: Did we get to the bottom of this?

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: At the moment there's only a single place where we need to update the AOTMap. I'm not sure there'll be many other places (probably only constants), as we don't need to map things like function definitions.

ykcapi/src/sginterp.rs
```diff
@@ -0,0 +1,142 @@
...
+        let (addr, size) = ykutil::obj::llvmbc_section();
+        let membuf = LLVMCreateMemoryBufferWithMemoryRange(
+            addr as *const i8,
+            size,
+            "".as_ptr() as *const i8,
+            0,
+        );
+        let context = LLVMContextCreate();
+        let mut module: MaybeUninit<LLVMModuleRef> = MaybeUninit::uninit();
+        let module = {
```
ltratt, line 67: Might this be clearer as something like:

```rust
let mut module: MaybeUninit<...> = ...;
LLVMParseBitcodeInContext2(...);
let module = module.assume_init();
```

?

ykcapi/src/sginterp.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub unsafe fn new(bbidx: u32, instridx: u32, fname: &CStr) -> SGInterp {
+        // Get AOT module IR and parse it.
+        let (addr, size) = ykutil::obj::llvmbc_section();
+        let membuf = LLVMCreateMemoryBufferWithMemoryRange(
+            addr as *const i8,
+            size,
+            "".as_ptr() as *const i8,
+            0,
+        );
+        let context = LLVMContextCreate();
```
vext01, line 65: Should we be using the pre-existing context from the AOT module?

ykcapi/src/lib.rs
ptersilie: Yes. LLVM uses the first 3 elements of the stackmap to store (according to the source code) `CC, Flags, Num Deopts`. The last one is just the number of variables in the deopt call (here equal to `LiveVals`). The other two I'm not sure what the purpose is, and most importantly the documentation doesn't mention this at all, which is why I initially thought the stackmap has the same element size as the live variables we put in the deopt call.

vext01:
A preliminary skim suggests this looks good. No show-stoppers to report.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -464,18 +476,100 @@ class JITModBuilder {
       Instruction *CurrentInst = &CurrentBB->back();
       DominatorTree DT(*JITFunc);
       std::vector<Value *> LiveVals = getLiveVars(DT, CurrentInst);
+      // Naturally the current instruction is live too but wasn't included due
+      // to the way DominatorTree works.
+      LiveVals.push_back(CurrentInst);
+
+      // FIXME use more flexible type than int32
```
ltratt, line 483: `FIXME:` with a colon I think (will help us `grep` one day!).

ykcapi/src/sginterp.rs
ptersilie: Dunno, then we have to pass that through via the deopt call, but then we are running out of register arguments, making the setup to the deopt call even more complicated than it already is. Is there a problem in creating a new context?

ykcapi/src/lib.rs
ptersilie: Yes, once this stops being a draft I'll remove those.

ltratt:
I have lots of minor comments, but only one major one: tests. We need both blackbox and whitebox tests here.

ykcapi/src/lib.rs
vext01: OK. Add this to your list of places to improve LLVM docs.

ykcapi/src/lib.rs
vext01: @ptersilie Is it because there is no `char` type in LLVM?

ykcapi/src/llvmapihelper.rs
ptersilie: Depends on your definition of parsing I guess. ;-) It turns something like `Constant(Type(width: 32), 1` into `u32 1`. I'm open to suggestions.

ltratt: Let's keep "parsing" for "text to something" conversions :p

I guess this function might be called something like `llvm_const_to_sgconst`?

ykcapi/src/sginterp.rs
vext01: I don't understand how we could do that.

Given a variable `%myvar`, which is identified by a `Value *`, what would it's index be?

ltratt: Every function has a finite set of variables that we know in advance, so it's clearly doable. [It might be awkward, but that's another matter.]

ykcapi/src/sginterp.rs
vext01: Are you able to call `getContext()` on the module?

ykcapi/src/lib.rs
ptersilie: It's a `*i8` in LLVM. Is there a difference?

ykcapi/src/sginterp.rs
```diff
@@ -0,0 +1,142 @@
...
+        let instr = llvmapihelper::get_instruction(bb, instridx);
+        self.frames.last_mut().unwrap().add(instr, value);
+    }
+
+    /// Lookup the value of variable `var` in the current frame.
+    fn lookup(&self, var: &LLVMValueRef) -> Option<&SGValue> {
+        self.frames.last().unwrap().get(var)
+    }
+
+    /// Start interpretation of the initialised interpreter.
```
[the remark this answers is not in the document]
vext01: It depends how you define "start" i guess. FWIW, I knew what Lukas meant: It starts interpretation from the point of the guard failure.

ykcapi/src/sginterp.rs
```diff
@@ -0,0 +1,142 @@
...
+            match LLVMGetInstructionOpcode(self.pc) {
+                LLVMOpcode::LLVMBr => self.branch(self.pc),
+                LLVMOpcode::LLVMRet => self.ret(self.pc),
+                _ => todo!("{:?}", CStr::from_ptr(LLVMPrintValueToString(self.pc))),
+            }
+        }
+    }
+
+    /// Interpret branch instruction `instr`.
+    pub unsafe fn branch(&mut self, instr: LLVMValueRef) {
```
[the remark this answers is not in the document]
vext01: Is there a `LLVMBranchInstRef` Lukas?

ykcapi/src/lib.rs
ltratt: Ah, if it's definitely an `i8` in LLVM, we can do the same here.

[IIRC `char` in C99 is a synonym for `i8`, which is why I raised the query, but I might be wrong on that detail!]

vext01: Only a difference of intention, I'd say.

ltratt: "a difference of intention" That sounds deep but I don't know what it means!

ykcapi/src/sginterp.rs
vext01: I'm not sure if variable order is stable over serialisation or not.

Since this is essentially a question of optimisation, I suggest doing as Lukas has, and adding an `OPT:` referencing an issue.

ykcapi/src/sginterp.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// Add a live variable and its value to the current frame.
+    pub unsafe fn init_live(&mut self, bbidx: u32, instridx: u32, fname: &CStr, value: SGValue) {
+        let func = LLVMGetNamedFunction(self.module, fname.as_ptr());
+        let bb = llvmapihelper::get_basic_block(func, bbidx);
+        let instr = llvmapihelper::get_instruction(bb, instridx);
+        self.frames.last_mut().unwrap().add(instr, value);
+    }
+
+    /// Lookup the value of variable `var` in the current frame.
+    fn lookup(&self, var: &LLVMValueRef) -> Option<&SGValue> {
```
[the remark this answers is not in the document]
vext01: I think we should crash if we try to lookup a non-existent variable.

ykcapi/src/sginterp.rs
ltratt: `OPT` (or is it `YKOPT`?) works for me here.

ykcapi/src/lib.rs
ptersilie: Good point. We are actually using `size_t` in `jitmodbuilder.cc` but since LLVM doesn't have a `size_t` type, we probably need to add platform specific code when passing these values into the deoptimisation call.

ykcapi/src/sginterp.rs
ptersilie: Most values will be instructions, which do have an order (bbidx:instridx), which should be stable over serialisation. I'm not sure about constants though, they could appear in any order, even after serialisation.

ykcapi/src/sginterp.rs
ptersilie: No, there'll be loads more.

ykcapi/src/sginterp.rs
ltratt: This is about variables though, not values?

ykcapi/src/sginterp.rs
ltratt: OK.

ykcapi/src/sginterp.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// Create a new stopgap interpreter and initialise it to start interpretation at the location
+    /// given by a basic block index, instruction index, and function name.
+    /// FIXME: Support initalisation of multiple frames.
+    pub unsafe fn new(bbidx: u32, instridx: u32, fname: &CStr) -> SGInterp {
+        // Get AOT module IR and parse it.
+        let (addr, size) = ykutil::obj::llvmbc_section();
+        let membuf = LLVMCreateMemoryBufferWithMemoryRange(
+            addr as *const i8,
+            size,
+            "".as_ptr() as *const i8,
```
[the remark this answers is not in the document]
ptersilie: I will check. This is just to give the buffer a name which I don't care about. Maybe I can pass nullptr or something.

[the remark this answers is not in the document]
ltratt: I expect that, at least, you'll need to pass a `CString`.

ykcapi/src/sginterp.rs
ptersilie: Yeah, I guess we could put constants in a separate map and then put variables in a vector. Seems doable, but will probably remain a `YKOPT` for now.

ltratt: I'm confused: surely this `HashMap` is from variable names to values? Constants are just a value, right?

ykcapi/src/lib.rs
vext01: It makes clear that our intention is to store characters, even if there is no difference whatsoever to the compiler.

ykcapi/src/sginterp.rs
ptersilie: Probably. Dunno why I did this way tbh.

ykcapi/src/llvmapihelper.rs
ptersilie: Like it.

ykcapi/src/sginterp.rs
ptersilie: This map is from `* Value`s to interpreter value. Constants are also just values, but since I am not sure if their order remains stable over serialisation they probably will need to stay in a hashmap, but we certainly can move non constant variables (i.e. instructions) into a vector to increase performance.

ltratt: OK, then I think this is a design issue that I don't understand: LLVM defines variables (`%x` etc) and surely that's what we should be storing as keys, rather than any possible value? After all `2 = 3` (where `2` and `3` are constants) hardly makes sense.

vext01: > I guess we could put constants in a separate map

Then there's global variables. It's going to get complicated. I suggest we stick to a unified way of identifying variables.

> surely this HashMap is from variable names to values? 

LLVM doesn't identify variables using a name. It uses an opaque pointer value (which obviously isn't stable across serialisation).

Whilst you can assign variables names, there's no guarantee that the will carry into serialised IR. And if variables have no names and you print the IR, then the printer itself assigns them numeric names on the fly.

ykcapi/src/sginterp.rs
```diff
@@ -0,0 +1,142 @@
...
+        let instr = llvmapihelper::get_instruction(bb, instridx);
+        SGInterp {
+            module,
+            frames: vec![Frame::new()],
+            pc: instr,
+        }
+    }
+
+    /// Add a live variable and its value to the current frame.
+    pub unsafe fn init_live(&mut self, bbidx: u32, instridx: u32, fname: &CStr, value: SGValue) {
```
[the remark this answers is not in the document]
ptersilie: I used to call it `add_var`, but then I was worried the function will be confusing later, since this one is only to add variables to he frame during initialisation of the stopgap interpreter. Maybe `add_live` is a good compromise.

ykcapi/src/sginterp.rs
ltratt: Global variables shouldn't be stored in the frame, by definition, right?

My claim is that (whether they're in a generic `Value` struct or not), the only valid keys are variable "name"s, not constants, nor globals.

ykcapi/src/sginterp.rs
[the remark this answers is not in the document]
ptersilie: Yeah, I agree it should probably be `unreachable!`.

ykcapi/src/sginterp.rs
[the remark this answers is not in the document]
ptersilie: What @vext01 said. But I'm happy to change the comment to make it more clear.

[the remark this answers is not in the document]
ltratt: Yes please :)

ykcapi/src/sginterp.rs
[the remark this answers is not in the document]
ptersilie: There is no `LLVMBranchInstRef`, but can easily add a `debug_assert`.

ykcapi/src/lib.rs
```diff
@@ -95,29 +100,66 @@ impl Registers {
     }
 }
 
+#[derive(Debug)]
+#[repr(C)]
```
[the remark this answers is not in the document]
ptersilie: Yes, though we are generating the corresponding struct in LLVM, which now that you mention it I should probably double check that they are indeed also creating C structs.

ykcapi/src/sginterp.rs
vext01: > Global variables shouldn't be stored in the frame, by definition, right?

Ah, that's correct, yes.

> My claim is that (whether they're in a generic Value struct or not), the only valid keys are variable "name"s, not constants, nor globals.

The only valid variables in a live frame are local variables, but I don't understand this "name" thing.

ykcapi/src/lib.rs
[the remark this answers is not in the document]
ltratt: It would be good to add a comment here along the lines of "these mirror the structs X, Y, Z, in file ABC.cc in repo blah".

ykcapi/src/sginterp.rs
ptersilie: Oh, yes, you're right, we don't need to store globals and consts in the frame at all. That being said, we still can't use variable "names" as keys, since (as @vext01 pointer out above) there's no guarantee they have one. So it'll have to be pointers for now. But we'll later change the variable map to a vector, then key goes away anyway.

ltratt: It's standard CS theory: variables have names. Names don't have to be strings: they can be numbers, or pointers, or whatever.

vext01: I've never heard of that, but if we define an LLVM variable name as its opaque pointer, then yes, this map is keyed by "names".

Is the confusion then, that an LLVM "variable name" has a specific meaning in LLVM, that is different from CS theory?

ltratt: I think the confusion is because we assume "name" must be "string", but that needn't be the case: it just needs to be something that uniquely identifies that variable.

The challenge we have here -- and why I would suggest we might want to use a key that *isn't* an LLVM `Value` directly -- is that (in a normal language) we've got the equivalent of `HashMap<Object, Object>` which is why we end up thinking you can have constants or global variables as keys (when you can't). I would suggest having a struct something like `SGVarName` (or maybe even change the serialisation to make this clear!) so that we end up with `HashMap<VarName, Value>` to make clear that only a strict subset of possible `Value`s can be keys.

ptersilie: Just to double check: d79ec17 does what you meant?

ptersilie:
This was a very useful discussion. There's quite a lot to change (including parts of the design), so I think it's best to close this draft and open a new PR when everything is ready.

PR #475, opened 2022-01-20
yk_control_point() must be called in a loop.

vext01:
Add a test that checks the compiler will reject programs which do not
call the control point in an outer loop.

Also fix a test which didn't call the control point in a loop.

Requires: https://github.com/ykjit/ykllvm/pull/20


vext01:
A few extra commits to review there. Changes in response to changes on the compiler side, and some tidying.

ltratt:
I'm not a fan of `ctrl_pt` -- can we use `control_point` for the name? I've found descriptive test names (which means descriptive file names... I know!) are useful in the long term because there's no way I'll remember to grep for `ctrl_pt`!

vext01:
Those two commits do the renaming. Separate, because one of the tests was introduced in the previous PR and shouldn't be squashed in.

vext01:
OK to squash this one?

ltratt:
Please squash.

vext01:
splat

PR #477, opened 2022-01-27
Start moving away from a singleton `MT` object

ltratt:
At the moment our API forces us to use a singleton `MT` instance which we fish out whenever we need it. This is not only slow (it's in a `SyncLazy`), but it will cause us problems if/when we want to allow more than one `MT` in a single process.

This PR starts to move us away from that. First, we expose a `YkMt` type (https://github.com/ykjit/yk/commit/6ec9342dc6f948e771b293c7423667f7a8b002da). Then we add this new type as a first parameter to control points (https://github.com/ykjit/yk/commit/50505751f20c1c50d0164d4b4d165694af4f0350). This second commit needs the companion PR https://github.com/ykjit/ykllvm/pull/21.

This PR *does not go all the way*: there's still a fair bit more work to do to get rid of the singleton `MT`. However, because of the annoyance of the companion PR, and because I doubt I'll get to the rest of this stuff for a few days, I think it's better to get this reviewed and, maybe if I've not gone too far wrong, merged (it might have a small effect on some of @ptersilie's current work?).

Note that in https://github.com/ykjit/yk/commit/50505751f20c1c50d0164d4b4d165694af4f0350 I have had to replace several *truly evil* magic constants with named constants: I can't promise that I've got these right, because they're not explained in any way. Please check those parts particularly carefully!

### DIFF
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -24,7 +24,8 @@ uint64_t getNewTraceIdx() {
 
 #define TRACE_FUNC_PREFIX "__yk_compiled_trace_"
 #define YK_NEW_CONTROL_POINT "__ykrt_control_point"
-#define YK_CONTROL_POINT_ARG_IDX 1
+#define YK_CONTROL_POINT_ARG_VARS_IDX 2
+#define YK_CONTROL_POINT_NUM_ARGS 3
 
 // Dump an error message and an LLVM value to stderr and exit with failure.
 void dumpValueAndExit(const char *Msg, Value *V) {
@@ -113,7 +114,8 @@ Value *getYkCtrlPointVarsStruct(Module *AOTMod, InputTrace &InpTrace) {
   assert(F != nullptr);
   User *CallSite = F->user_back();
   CallInst *CI = cast<CallInst>(CallSite);
-  return CI->getArgOperand(YK_CONTROL_POINT_ARG_IDX);
+  assert(CI->arg_size() == YK_CONTROL_POINT_NUM_ARGS);
+  return CI->getArgOperand(YK_CONTROL_POINT_ARG_VARS_IDX);
 }
 
 /// Extract all live variables that need to be passed into the control point.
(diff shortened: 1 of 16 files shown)


vext01:
Is the end-goal to have `mt` user visible, or will it eventually be hidden?

My concern is that by making `mt` appear in the AOT module, operations on it could be traced, and we might have to try and strip them out after the fact.

This is reminiscent of why we had opted to [hide the thread tracer](https://github.com/ykjit/yk/pull/422/commits/9f5eb6fa1e58e8262d90bdb16919797056e182ee) (from #422).

Thoughts?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -24,7 +24,8 @@ uint64_t getNewTraceIdx() {
 
 #define TRACE_FUNC_PREFIX "__yk_compiled_trace_"
 #define YK_NEW_CONTROL_POINT "__ykrt_control_point"
-#define YK_CONTROL_POINT_ARG_IDX 1
+#define YK_CONTROL_POINT_ARG_VARS_IDX 2
```
vext01, line 27: Might be worth a comment against `YK_CONTROL_POINT_ARG_VARS_IDX`. Something like:

> The argument index of the live variables struct in `__ykrt_control_point`

Or maybe it's obvious? Not sure.

vext01:
> I don't think that's a concern here: the change to the API will have little or no effect on what's traced

I agree that currently the `mt` instance is only every passed into the control point for now. In this case, since the control point is omitted from the trace, the `mt` would become dead in the JIT module.

But if the user did `yk_set_hot_threshold(mt, 0);` in the interpreter loop. then the jitted code would in fact change the JIT hot threshold. I just want to ensure that we've thought that through.

ykllvmwrap/src/jitmodbuilder.cc
ltratt: I think it's probably obvious enough?

ltratt:
> But if the user did yk_set_hot_threshold(mt, 0); in the interpreter loop. then the jitted code would in fact change the JIT hot threshold. I just want to ensure that we've thought that through.

They can do that, and it's fine! `set_hot_threshold` is best thought of as a "suggestion" rather than a guarantee (because another thread might change it at the same time as you do). If you set it while a thread is tracing, it will have no impact on the current trace (since the hot threshold has already been exceeded).

vext01:
@ptersilie OK with this change?

ptersilie:
All good for me.

vext01:
I take it there's no squashing to do?

ltratt:
No squashing needed.

PR #478, opened 2022-02-01
Move __llvm_deoeptimize() into Rust code in ykcapi.

vext01:
Previously this assembler routine was in a C header file.

By moving it here, it means that non-C targets (e.g. upcoming test suites written in Rust) can see the symbol too.

Note that asm! defaults to Intel syntax on x86_64.


vext01:
A couple of notes:
 - The asm routine is a literal translation from C, `FIXME`s and all.
 - I managed to get the new test suite working with this (without having the compiler generate the symbol). I'm not sure what I had been doing wrong before.

ykcapi/src/lib.rs
```diff
@@ -170,6 +172,48 @@ pub extern "C" fn yk_stopgap(
     process::exit(0);
 }
 
+/// The `__llvm__deoptimize()` function required by `llvm.experimental.deoptimize` intrinsic, that
+/// we use for exiting to the stop-gap interpreter on guard failure.
+#[cfg(target_arch = "x86_64")]
```
vext01, line 177: It's just occurred to me that we should have a `compile_error!` for other arches. Agree?

vext01: Fixed in eb0bc4e

ltratt: Good idea, as without that people will get a "symbol not found" (which is better than nothing, but not always the most helpful message!).

ltratt:
I'm fine with this. If @ptersilie is too, then please squash.

ptersilie:
Looks good!

vext01:
splat.

PR #480, opened 2022-02-01
Initialisation tidying.

vext01:
This commit moves as much of the initialisation code from the top of
`JITModBuilder::createModule()` into the constructor, and in doing so,
eliminates some duplication of code.

The motivation for this is that we'd like to be able to test smaller
units of this struct in isolation, so it really helps if the
initialisation is captured (where possible) by the constructor.

There is more to come, but this is a start.

No functional changes.


vext01:
splat.

PR #482, opened 2022-02-02
Simplify code that passes the stackmap pointer and size into the trace.

vext01:
I think this is correct, but I would like @ptersilie to check.

### DIFF
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -1,5 +1,6 @@
 // Classes and functions for constructing a new LLVM module from a trace.
 
+#include "llvm/IR/DataLayout.h"
 #include "llvm/IR/DebugInfo.h"
 #include "llvm/IR/Dominators.h"
 #include "llvm/IR/IRBuilder.h"
@@ -179,6 +180,9 @@ class JITModBuilder {
 
   // The LLVM type for a C `int` on the current machine.
   Type *IntTy;
+
+  // A pointer-sized integer (i.e. a Rust `usize`) for the current machine.
+  IntegerType *PointerSizedIntTy;
 
   // A pointer to the Value representing the trace input struct.
   Value *TraceInputs;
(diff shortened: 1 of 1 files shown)


ltratt:
An obvious question: does the receiving side of this change (i.e. the compiled trace) expect a "64 bit" (specifically) input or `uintptr_t` equivalent?

vext01:
My change was purely mechanical, but looking deeper....

The code I changed is the code which generates the parent function of the JITted code.

This is called from Rust here:
https://github.com/ykjit/yk/blob/2e0ad668350f526144169e5cda17f552474ef350/yktrace/src/lib.rs#L192-L194

Where:
https://github.com/ykjit/yk/blob/2e0ad668350f526144169e5cda17f552474ef350/yktrace/src/lib.rs#L169-L172

So it appears that these two parameters need to be pointer sized, hence the variability per-platform.

ltratt:
> So it appears that these two parameters need to be pointer sized, hence the variability per-platform.

I don't understand this sentence.

vext01:
The two arguments in question are passed at the call site as `*const c_void` (pointer-sized) and `usize` (also pointer-sized).

This is why we can't use fixed-size argument types for all platforms when generating the JITted code function.

ltratt:
I understand your statements in isolation but I can't relate them to the PR: `uintptr_t` and equivalent aren't fixed size.

Put another way: is your comment in https://github.com/ykjit/yk/pull/482#issuecomment-1027826061 saying "this means the PR is wrong" or ... ?

vext01:
I believe the change to remain correct.

ltratt:
OK, so I can just ignore https://github.com/ykjit/yk/pull/482#issuecomment-1027826061 then?

ltratt:
After offline discussion, I now understand that https://github.com/ykjit/yk/pull/482#issuecomment-1027826061 was answering my question in https://github.com/ykjit/yk/pull/482#issuecomment-1027815141 in a good way :)

bors r+

ptersilie:
Looks good to me.

vext01:
My bad. OK to squash?

ltratt:
Please squash.

vext01:
splat.

PR #483, opened 2022-02-02
Push THREAD_MTTHREAD down off the fast path

ltratt:
This PR mostly aims to get `THREAD_MTTHREAD` off of the control point fast path (https://github.com/ykjit/yk/commit/e20071a8861940ce7410ada8603c05d124b432a2). However this requires an API change such that we rethink the relationship between `MT` and `MTThread` (https://github.com/ykjit/yk/commit/5457f8be379b948e3b1fb4f8b008c74ad4e2b886). As a sort-of bonus, I've also started better understanding how to test the location statemachine, so that's now slightly better tested (https://github.com/ykjit/yk/commit/823f7400e73e19321c20b71289dd3ed4042d0ac7), giving us a bit of confidence that we're not breaking too many things.

For full disclosure, this PR contains a genuine horror which you can see at https://github.com/ykjit/yk/commit/e20071a8861940ce7410ada8603c05d124b432a2#diff-9fa85f3ab0ab08c2bb4337dd17d47605f520c2780158e50e66eff7cd9dc1b29bR291: this is clearly terrible, but until there's a better test suite, I'm not overly confident about fixing this until the test suite is reextended.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -36,7 +36,6 @@ type AtomicHotThreshold = AtomicU32;
 
 const DEFAULT_HOT_THRESHOLD: HotThreshold = 50;
 
-static GLOBAL_MT: SyncLazy<MT> = SyncLazy::new(|| MT::new());
 thread_local! {static THREAD_MTTHREAD: MTThread = MTThread::new();}
 
 #[cfg(feature = "c_testing")]
@@ -51,13 +50,9 @@ pub struct MT {
 }
 
 impl MT {
-    /// Return a reference to the global [`MT`] instance: at any point, there is at most one of
-    /// these per process and an instance will be created if it does not already exist.
-    pub fn global() -> &'static Self {
-        &*GLOBAL_MT
-    }
-
-    fn new() -> Self {
+    // Create a new meta-tracer instance. Arbitrarily many of these can be created, though there
+    // are no guarantees as to whether they will share resources effectively or fairly.
+    pub fn new() -> Self {
         let inner = MTInner {
             hot_threshold: AtomicHotThreshold::new(DEFAULT_HOT_THRESHOLD),
             job_queue: (Condvar::new(), Mutex::new(VecDeque::new())),
@@ -438,7 +433,7 @@ mod tests {
 
     #[test]
     fn hot_threshold_passed() {
-        let mt = MT::global();
+        let mt = MT::new();
         let loc = Location::new();
         THREAD_MTTHREAD.with(|mtt| {
             for i in 0..mt.hot_threshold() {
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -129,36 +129,34 @@ impl MT {
     }
 
     pub fn control_point(&self, loc: &Location, ctrlp_vars: *mut c_void) {
-        THREAD_MTTHREAD.with(|mtt| {
-            match self.transition_location(mtt, loc) {
-                TransitionLocation::NoAction => (),
-                TransitionLocation::Execute(ctr) => {
-                    // FIXME: If we want to free compiled traces, we'll need to refcount (or use
-                    // a GC) to know if anyone's executing that trace at the moment.
-                    //
-                    // FIXME: this loop shouldn't exist. Trace stitching should be implemented in
-                    // the trace itself.
-                    // https://github.com/ykjit/yk/issues/442
-                    loop {
-                        #[cfg(feature = "jit_state_debug")]
-                        eprintln!("jit-state: enter-jit-code");
-                        unsafe { &*ctr }.exec(ctrlp_vars);
-                        #[cfg(feature = "jit_state_debug")]
-                        eprintln!("jit-state: exit-jit-code");
-                    }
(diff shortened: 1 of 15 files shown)


ykrt/src/mt.rs
```diff
@@ -286,16 +288,28 @@ impl MT {
                     unreachable!();
                 }
                 HotLocation::Tracing(opt) => {
-                    match mtt.tracing.get() {
-                        Some(other_hl_ptr) => {
+                    // FIXME: This is the sort of hack that keeps me awake at night: we really want
+                    // to return from the outer function, and to modify `hl`, but can't because
+                    // we're in a closure. The integer return value allows us to perform the
+                    // control flow we want.
+                    let r = THREAD_MTTHREAD.with(|mtt| {
```
vext01, line 295: I've been up against this sort of thing myself before. It's a pain.

Does the code touching the thread local have to be a closure, or can it be a regular function? I wonder if it would be better pulled-out of here into a `fn is_current_thread_tracing()` function?

Also, I understand that the scope of those magic numbers is small, but it's really hard to read. Can we define enum so that we get the opportunity to give the cases more easily understandable names?

Something like:

```
enum OngoingThreadTracing{
   NotTracing,
   AnotherLocation,
   ThisLocation,
}
```

What do you think?

vext01:
Other than that one comment, I think we need tests that check that the right thing happens when:

 - Multiple MTs are used in disjoint threads.
 - Multiple MTs are used in the same threads.

vext01:
OK, one more question.

When a user is going to use multiple MTs in an interpreter, should the MTs all be born and die at the same time?

If not, then would/should it would be possible for one MT to trace the creation and destruction of another MT. This hurts my head in the way that that film [Primer](https://en.wikipedia.org/wiki/Primer_(film)) does, where those guys take a time machine into a time machine and it all gets a bit confusing.

If all MTs should be born and die at the same time, should there be an all-encompassing data structure that creates them all together, ties their lifetimes together and automatically calls the destructors on them when it goes out of scope?

ykrt/src/mt.rs
ltratt: This is a temporary hack and I'm not sure an `enum` will make it much less gross to be honest. Turning the closure into a function won't help, unfortunately, because you still can't return from an outer function.

ltratt:
> I think we need tests that check that the right thing happen

Agreed on those. I will continue resuscitating the test suite over the coming days with these and, hopefully, more!

> When a user is going to use multiple MTs in an interpreter, should the MTs all be born and die at the same time?

If they're born and die at the same time, there's no point having more than one.

Basically at one point I had the idea that `MT` might need to take over the entire process address space, hence it would only make sense to have one `MT. I don't think it actually needs to do so, and thus loosening the restriction makes sense. If we're proven wrong at some point, rerestricting things won't be super hard to do.

ykrt/src/mt.rs
vext01: Can we at the very least put a comment next to where you consume each magic number saying what case it covers?

vext01:
> Agreed on those.

OK, so can we put something in the codebase or issue tracker somewhere to ensure that this won't be forgotten?

> If they're born and die at the same time, there's no point having more than one.

How comes? If you wanted one for the interpreter and one for your regex engine, you could pre-allocate exactly two ahead of time, couldn't you?

So can an an unbounded number of MTs be created?

Should this kind of thing be possible:

```
mt1 = new_mt();
for i in atoi(argv[1]) {
  control_point();
  realloc(more_mts, ...);
  more_mts[idx] = mt_new();
  loop {
     mt_i = mts[blah];
     control_point(mt_i);
  }
}
```

Here `mt1` witnesses the creation of the other mts.

ltratt:
> Should this kind of thing be possible:

Yes.

Though there is a variation on your example that would go wrong in weird ways: each `MT` must operate on a different set of `Location`s. We can't enforce that in the static type system, but it should be documented somewhere at some point I suppose.

ykrt/src/mt.rs
ltratt: Fixed in 4fd2f67.

vext01:
> Yes.

Mad! But so be it.

I'm still curious about this part:

> >  If they're born and die at the same time, there's no point having more than one.

> How comes? If you wanted one for the interpreter and one for your regex engine, you could pre-allocate exactly two ahead of time, couldn't you?

I'd have thought the common case would be where you have a statically known number of MTs.

ltratt:
Common -- maybe. But we can't enforce that in the type system either. I don't think we need to get *too* obsessed about this, because honestly we probably can't guess all the ways you could (or shouldn't...) use this in the future!

vext01:
That's fine. But it's good to have at least superficially discussed the scenarios that a design will allow, intentionally or otherwise!

Please squash.

ltratt:
Squashed.

ltratt:
Oops! cfmt fixed in https://github.com/ykjit/yk/pull/483/commits/462458f5ecc3864611286c4c7789c2fe99a0c45c. OK to squash?

ltratt:
Squashed.

PR #484, opened 2022-02-06
Test threshold incrementing and threading.

ltratt:
This is a partial port of an old test but, in so doing I realised that, in our current setup, where we don't retry incrementing the hot count endlessly, we can make only weak guarantees about what the final hot count will be.

That made me realise that the loop we had in the transition function isn't really *that* useful. It was originally designed to enforce semi-monotonicity, but I think that complexity is neither useful in practise or worth the extra lines of code.

In a bit more detail to justify the removal of the loop: what we really want the transition function to do is run very fast. In the case where one thread is running a loop, the hot count is likely to increment as expected. If there are multiple threads all hammering away at the same location, it's inevitable that some increments will be missed -- but in such a case, we're still going to move towards the threshold quickly enough, so why bother spinning a thread at all?

[The original loop also tried to handle the situation of the count going backwards. This can't be observed in the "normal, expected" sequence of transitions, but could be observed if a location goes back to the Counting state. That would require a thread that observed the location
in the Counting state being starved for a very long period of time. Given that this isn't likely to happen very often, it hardly seems worth having code in to try and handle that possibility.]

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -19,7 +19,6 @@ use std::{
 
 use num_cpus;
 use parking_lot::{Condvar, Mutex, MutexGuard};
-use parking_lot_core::SpinWait;
 use std::lazy::SyncLazy;
 
 use crate::location::{HotLocation, Location, LocationInner, ThreadIdInner};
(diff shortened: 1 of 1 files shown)


vext01:
> That would require a thread that observed the location
in the Counting state being starved for a very long period of time.

By which time the mutation that the thread wants to do is probably irrelevant anyway, right?

ltratt:
> By which time the mutation that the thread wants to do is probably irrelevant anyway, right?

Exactly.

> The only downside is that we have to deal with the non-determinism introduced in our tests, as demonstrated by your new test.

The non-determinism was always there but now we try slightly less hard to cover it up.

PR #485, opened 2022-02-08
A rough attempt at a "Trace Driver" test suite.

vext01:
This is a draft. Don't bother to review the code -- it needs tidying. I'm raising this to see if we are headed in the right direction.

The way this works:
 - The user provides a lang test, containing a LLVM module.
 - Unlike before, the provided module can be really bare-bones, devoid of lengthy control_point machinery. Just blocks containing instructions they want to simulate a trace through.
 - The user can liberally use `unreachable` terminators, for example in blocks that are intended to be at the end of a trace, or for blocks which will never be traced.
 - In the lang test, the user sets `env-var: TRACE_DRIVER_BBS=...` to "drive" the test down one given path of the CFG of the provided module. The value is a comma separated list of `func:bb` pairs, and `bb` is a block index (not it's name in the module yet. we could make it so tha t the user could use the name of the block, but this would require more work, as we'd have to parse the module and build a map of name->idx).
 - The trace compiler compiles the specified trace, faking control point machinery as necessary.
 - The user's test does the `env-var: YKD_PRINT_IR=...` trick as usual and matches the expected outcome at the desired stage in the pipeline (e.g. `jit-pre-opt`).

A good place to start is [this simple two-block test](https://github.com/ykjit/yk/compare/master...vext01:trace-driver-tests-v2#diff-89bfaa8f17998eedc6def69d274e139c435e845e664e7a0712883cfe65053f9e).

Once you've digested that, take a look at [this more involved test](https://github.com/ykjit/yk/compare/master...vext01:trace-driver-tests-v2#diff-d63cae56313cd4fdaa8939bc91feeaef1ae2c431f064e7d5d70e20003316ca3f), which checks that a conditional branch instruction gets an appropriate guard inserted.

Things for a later PR:
 - rename the `c_tests` directory, as this now contains more than C tests.
 - rename the `c_testing` feature as well.

Any comments on the design?

### DIFF
--- a/c_tests/run_trace_compiler.rs
+++ b/c_tests/run_trace_compiler.rs
@@ -1,0 +1,57 @@
+//! Lang tester harness for trace compiler tests.
+
+#![feature(once_cell)]
+
+use lang_tester::LangTester;
+use regex::Regex;
+use std::{env, fs::read_to_string, path::PathBuf, process::Command};
+
+const COMMENT: &str = ";";
+
+fn main() {
+    println!("Running trace compiler tests...");
+
+    // Find the `run_trace_driver_test` binary in the target dir.
+    let md = env::var("CARGO_MANIFEST_DIR").unwrap();
+    let mut run_tc_test = PathBuf::from(md);
+    run_tc_test.push("..");
+    run_tc_test.push("target");
+    #[cfg(cargo_profile = "release")]
+    run_tc_test.push("release");
+    #[cfg(cargo_profile = "debug")]
+    run_tc_test.push("debug");
+    run_tc_test.push("run_trace_compiler_test");
+
+    LangTester::new()
+        .test_dir("trace_compiler")
+        .test_file_filter(|p| {
+            if let Some(ext) = p.extension() {
+                ext == "ll"
+            } else {
+                false
+            }
+        })
+        .test_extract(move |p| {
+            read_to_string(p)
+                .unwrap()
+                .lines()
+                .skip_while(|l| !l.starts_with(COMMENT))
+                .take_while(|l| l.starts_with(COMMENT))
+                .map(|l| &l[COMMENT.len()..])
+                .collect::<Vec<_>>()
+                .join("\n")
+        })
+        .test_cmds(move |p| {
+            let mut run_tc_test = Command::new(run_tc_test.clone());
+            run_tc_test.arg(p);
+            vec![("Run-time", run_tc_test)]
+        })
+        .fm_options(|_, _, fmb| {
+            // Use `{{}}` to match non-literal strings in tests.
+            // E.g. use `%{{var}}` to capture the name of a variable.
+            let ptn_re = Regex::new(r"\{\{.+?\}\}").unwrap();
+            let text_re = Regex::new(r".+?\b").unwrap();
+            fmb.name_matcher(ptn_re, text_re)
+        })
+        .run();
+}
--- a/c_tests/trace_compiler/guard_eq.ll
+++ b/c_tests/trace_compiler/guard_eq.ll
@@ -1,0 +1,37 @@
+; Run-time:
+;   env-var: YKD_PRINT_IR=jit-pre-opt
+;   env-var: TRACE_DRIVER_BBS=main:0,main:1
+;   stderr:
+;      --- Begin jit-pre-opt ---
+;      ...
+;      define internal void @__yk_compiled_trace_0(...
+;        %{{0}} = alloca i32, align 4
+;        %{{1}} = icmp eq i32 1, 1
+;        br i1 %{{1}}, label %{{true}}, label %guardfail
+;
+;      guardfail:...
+;        call void (...) @llvm.experimental.deoptimize.isVoid(...
+;        ret void
+;
+;      {{true}}:...
+;        store i32 1, i32* %{{0}}, align 4
+;        ret void
+;      }
+;
+;      declare void @llvm.experimental.deoptimize.isVoid(...)
+;      --- End jit-pre-opt ---
+
+define void @main() {
+entry:
+    %0 = alloca i32
+    %1 = icmp eq i32 1, 1
+    br i1 %1, label %true, label %false
+
+true:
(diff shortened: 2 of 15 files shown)


vext01:
I'd also like to find a better name than "Trace Driver", as that has meaning in PyPy nomenclature.

ltratt:
Question: could/should we make these Rust-y `#[test]` esque things that can run in a single address space rather than us spinning up lots of processes with `llvm-as` and `trace_driver`?

Rather than `trace_driver` you could call this something like `trace_tester`?

vext01:
> Question: could/should we make these Rust-y #[test] esque things that can run in a single address space rather than us spinning up lots of processes with llvm-as and trace_driver?

Can you show a rough mock-up of the kind of test you have in mind?

> trace_tester

It's a bit generic. Arguably the existing test suite is testing traces too.

ltratt:
> Can you show a rough mock-up of the kind of test you have in mind?

RPython does all these tests as "intra process" normal Python tests. There are some disadvantages to that (in RPython the tests are run single-threaded...) but some advantages (you don't have to spin up lots of new processes).

So it would just be something like:

```rust
#[test]
fn test_blah {
  test_driver("<llvm>", "<fm test>);
}
```

I'm not saying this is definitely a good idea (maybe we can't sensibly run multiple tests of this type in the same address space), but I want to make sure we've at least thought about it.

vext01:
I'd have to do some more kludging to get each test to return a different bitcode module (currently the bitcode is an immutable lazy static for the entire process).

Do you think it will be OK to have large(ish) llvm modules and fm tests in embedded rust strings? Could be a bit clumsy?

ltratt:
My job is to make sure you've thought about the possibilities: your job is to pick the right one :)

vext01:
@ptersilie do you have any comments?

ptersilie:
This looks reasonable to me. The tests also seem intuitive enough to write by hand.

vext01:
> Could be a bit clumsy?

I'm still leaning towards keeping the tests in separate files, because:
 - Both llvm bitcode and the expected outcome are formats where we care about indentation, and experience suggests that embedding those into language-level strings is awkward.
 - By having the `.ll` files in a file, we can more easily directly experiment with them using standard LLVM tools like `llvm-as`, `opt` etc.

I suggest I try to remove the shell-out to `llvm-as`, but keep the tests in separate files.

Any objections?

ptersilie:
> I suggest I try to remove the shell-out to `llvm-as`, but keep the tests in separate files.

Sounds reasonable to me.

ltratt:
> I suggest I try to remove the shell-out to llvm-as, but keep the tests in separate files.

I have no objections!

vext01:
> I suggest I try to remove the shell-out to llvm-as, but keep the tests in separate files.

I've now implemented that. It actually simplified some sections of code quite a bit.

Before I carry on polishing, instead of "trace driver tests", how about just "trace compiler tests". Keep it simple :)

ltratt:
Pre-review squashing works for me -- please go ahead!

vext01:
splat.

vext01:
Note that a follow-up PR will rename the `c_tests` directory and some other stuff.

c_tests/run_trace_compiler.rs
```diff
@@ -0,0 +1,57 @@
...
+#![feature(once_cell)]
+
+use lang_tester::LangTester;
+use regex::Regex;
+use std::{env, fs::read_to_string, path::PathBuf, process::Command};
+
+const COMMENT: &str = ";";
+
+fn main() {
+    println!("Running trace compiler tests...");
```
ltratt, line 12: I don't think this is useful -- it'll just get in the way of `cargo test` style output I think?

c_tests/trace_compiler/guard_eq.ll
```diff
@@ -0,0 +1,38 @@
...
+    br i1 %1, label %true, label %false
+
+true:
+    store i32 1, i32 * %0
+    unreachable
+
+false:
+    store i32 0, i32 * %0
+    unreachable
+
```
ltratt, line 37: Blank line?

ltratt:
Modulo those two minor comments this looks good to me.

c_tests/run_trace_compiler.rs
vext01: I think it's useful to see what suite a failure is coming from. The problem is that lang test tests always identify as `lang_tests::` regardless of which suite they came from:

```
Running C tests with -Oz...
...
test lang_tests::simple_interp_loop1.c ... ok                                                              
test lang_tests::simple_interp_loop2.c ... ok
                                                                                                           
test result: ok. 44 passed; 0 failed; 31 ignored; 0 measured; 0 filtered out                               
                                                     
     Running run_trace_compiler.rs (target/debug/deps/trace_compiler_tests-2ecdc66af5638631)               
Running trace compiler tests...          
                                                     
running 4 tests                                                                                            
test lang_tests::smallest.ll ... ok           
test lang_tests::two_blocks.ll ... ok                 
```

As you can see already have similar prints for the C tests (where otherwise you wouldn't know what optimisation level a failure came from), and I don't think the prints look out of place.

Maybe this is something that could be addressed in `lang_tester`. If the suite name could be customised, we could see:

```
test c_prog_O0::my_test... ok
test c_prog_Oz::my_test... ok
test trace_compiler::other_test... fail
```

Then we wouldn't need extra prints.

c_tests/trace_compiler/guard_eq.ll
vext01: I think that's the implicit blank line vim inserts?

I don't see it in vim:
image

vext01: oh wait :P

vext01: Fixed in fb0a4a4

ltratt: "I don't see it in vim" :p

ltratt:
Please squash.

vext01:
splat.

PR #486, opened 2022-02-08
More tests

ltratt:


ykrt/src/mt.rs
```diff
@@ -525,4 +525,43 @@ mod tests {
...
+        }
+
+        assert_eq!(
+            hotlocation_discriminant(&loc),
+            Some(HotLocationDiscriminants::Tracing)
+        );
+        assert_eq!(mt.transition_location(&loc), TransitionLocation::NoAction);
+        assert_eq!(
+            hotlocation_discriminant(&loc),
+            Some(HotLocationDiscriminants::DontTrace)
```
vext01, line 566: Do I read it right that if a thread does while tracing, no other thread will ever trace the location? Surely not?

ykrt/src/mt.rs
```diff
@@ -567,6 +567,34 @@ mod tests {
...
+            mt.transition_location(&loc1),
+            TransitionLocation::StartTracing(_)
+        ));
+        assert_eq!(mt.transition_location(&loc2), TransitionLocation::NoAction);
+        assert_eq!(
+            hotlocation_discriminant(&loc1),
+            Some(HotLocationDiscriminants::Tracing)
+        );
+        assert!(loc2.load(Ordering::Relaxed).is_counting());
+        assert_eq!(loc2.load(Ordering::Relaxed).count(), THRESHOLD);
```
vext01, line 595: Is it worth checking that `loc2` can be subsequently traced, after loc1 has moved to "compiled"?

ykrt/src/mt.rs
ltratt: This is currently a simplification: we probably want something more sophisticated in the long term. Currently the test just aims to make sure the current code does *something* other than leave the location dangling.

ykrt/src/mt.rs
ltratt: Not only is that an excellent plan but it highlighted a bug in the transition code. 84f817f minimally fixes it. In essence we had at some point dropped the line of code (in 86b3fbc4d5e8951d9f7cae938168ab079989c193?) that said "this thread is done tracing and can trace another location" so a thread could only trace a single location in its lifetime!

An interesting question is whether:

1. The user should make sure that they don't trace "silly" locations (like the `loop[1|2]` tests do).

or:

2. Before we execute a compiled trace we check to see whether the trace is currently executing something.

Both approaches have pros and cons and, I must admit, I'm not immediately sure which I prefer. Fixing that might be beyond this PR, especially as I have another branch fixing another even nastier bug in the transition code (undefined behaviour) that I think I'd rather focus on first.

vext01: I'm OK with the fix.

> An interesting question is whether:

I didn't understand this bit, but it sounds like a later PR might explain better?

If so, please squash.

ltratt: > I didn't understand this bit, but it sounds like a later PR might explain better?

I hope so... I need to chew on it!

ltratt:
Squashed -- notice the updated commit message in https://github.com/ykjit/yk/pull/486/commits/e4e9bb096f7af882abafe786edc4d37b38bc7c57.

vext01:
```
---- lang_tests::simple_interp_loop2.c stderr ----
Pattern (error at line 33):
   ...
   |...
   |--- End jit-pre-opt ---
   |pc=0, mem=3
>> |pc=1, mem=3
   |pc=2, mem=3
   |pc=3, mem=2
   |jit-state: enter-jit-code
   ...
Text (error at line 863):
   ...
   |!39 = !DILocation(line: 111, column: 12, scope: !37)
   |--- End jit-pre-opt ---
   |pc=0, mem=3
>> |jit-state: start-tracing
   |pc=1, mem=3
   |pc=2, mem=3
   |pc=3, mem=2
   ...
---- lang_tests::simple_interp_loop2.c stdout ----
```

ltratt:
Oops, I forgot to commit f562bea. OK to squash?

vext01:
please squash

ltratt:
Squashed.

PR #493, opened 2022-02-11
Add trace compiler tests for mutable and immutable globals.

vext01:
For mutable globals, which can't be optimised away, I had to add a small
hack to get the trace through the compiler without getting symbol
errors.

### DIFF
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -1109,6 +1109,25 @@ createModuleForTraceCompilerTests(Module *AOTMod, char *FuncNames[],
   JITModBuilder JB = JITModBuilder::CreateMocked(
       AOTMod, FuncNames, BBs, TraceLen, FAddrKeys, FAddrVals, FAddrLen);
   auto JITMod = JB.createModule();
+
+  // When the trace compiler encounters a non-const global in a trace, it
+  // inserts an LLVM `global external` variable referencing the variable in the
+  // interpreter's address space. In the trace compiler tests, such global
+  // variables don't actually exist so we will get symbol resolution errors
+  // when generating code for a trace.
+  //
+  // To avoid this, we insert a dummy address for all global variables (but not
+  // functions) in the JITMod which are external references (have no
+  // initialiser), and which don't already have a known address in the
+  // `GlobalMappings` map. Since we don't actually plan to execute the trace,
+  // their address is inconsequential. We just need it to compile.
+  for (GlobalVariable &G : JITMod->globals()) {
+    if ((!isa<Function>(&G)) && (!G.hasInitializer()) &&
+        (JB.GlobalMappings.find(&G) == JB.GlobalMappings.end())) {
+      JB.GlobalMappings.insert({&G, (void *)0xdeadbeef});
+    }
+  }
+
   return make_tuple(JITMod, std::move(JB.TraceName),
                     std::move(JB.GlobalMappings));
 }
--- a/tests/trace_compiler/global_immut.ll
+++ b/tests/trace_compiler/global_immut.ll
@@ -1,0 +1,23 @@
+; Run-time:
+;   env-var: YKD_PRINT_IR=jit-pre-opt
+;   env-var: YKT_TRACE_BBS=main:0
+;   stderr:
+;      --- Begin jit-pre-opt ---
+;      ...
+;      define internal void @__yk_compiled_trace_0(...
+;        %{{0}} = load i32, i32* @g, align 4
+;        %{{1}} = add i32 %{{0}}, 1
+;        ret void
+;      }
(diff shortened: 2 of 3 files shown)


ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -1109,6 +1109,25 @@ createModuleForTraceCompilerTests(Module *AOTMod, char *FuncNames[],
   JITModBuilder JB = JITModBuilder::CreateMocked(
       AOTMod, FuncNames, BBs, TraceLen, FAddrKeys, FAddrVals, FAddrLen);
   auto JITMod = JB.createModule();
+
+  // When the trace compiler encounters a non-const global in a trace, it
+  // inserts an LLVM `global external` variable referencing the variable in the
+  // interpreter's address space. In the trace compiler tests, such global
```
ltratt, line 1115: Is this only a problem in tests? Should this be feature gated if so?

vext01: It's only an issue for tests, and the function that I've changed is [already gated](https://github.com/ykjit/yk/pull/493/files/18beb6706c3e0e1ba56be8fa3d0025c7110f7d58#diff-b76ed8a29ed650e7c2a44a5809bc05a22cca958d0b9f5532dab74cce8503d112L1103), so we should be grand.

vext01: The link didn't work as expected, but if you expand the diff upwards, you will see the function is guarded with `#ifdef YK_TESTING`.

vext01: Well, it's something that can arise in real interpreters, and I wanted to write tests that demonstrate that they are correctly externed (and not copied), otherwise writes to them might not be observed by non-jit-compiled code.

I think we should allow non-const globals in tests.

vext01: > [I'm probably missing something obvious!]

It's not obvious -- hurts my head too!

In a real interpreter you might have a C program like:
```
int my_global_int = 4;

int main() {
  // interpreter loop
  ...
  my_global_int++;
}
```

The global becomes an llvm global variable in the AOT module.

If we collect a trace that writes to that global, then when we execute that trace, mutations need to be to be observed by non-jit-compiled-code, should we ever exit the trace. We insert an extern variable reference to `my_global_int` in the JIT module and when we compile the trace, the extern matches a symbol in the process initiating the JIT compilation (i.e. the interpreter), and all is dandy.

[If the global is constant, we just inline it -- I'll be testing that at some point too]

Here's the crux of it though: in trace compiler tests, we provide a hypothetical AOT module, which may talk about global variables and whatnot, **but those variables don't really exist**. The hypothetical AOT module hasn't actually ever been code-genned and the process initiating the JIT compilation **is the test harness**. Obviously the test harness doesn't define symbols found in the hypothetical AOT module, and stuff explodes.

To demonstrate, the tests in this PR use a global variable `g` in the hypothetical AOT module. When we compile a trace which manipulates `g`, LLVM looks for a symbol called `g` in the current address space (that of the test harness) and doesn't find one, and borks.

This change looks for these "otherwise problematic" globals and tells LLVM (only when testing):

> mate, trust me: that variable is at address `0xdeadbeef`

This is obviously semantically incorrect, should we execute the trace, but for these kinds of tests, we only check that they compile.

Does it make sense?

ltratt: OK, I get it (I think).

> This is obviously semantically incorrect, should we execute the trace, but for these kinds of tests, we only check that they compile.

OK, so I guess my concern is that `jitmodbuilder.cc` isn't just about "tests that we compile but don't run" -- doesn't this PR mean that if at some point in the future we use it for tests that could run, things will go wrong in weird ways that we'll struggle to track down?

vext01: The way we fake the control point already means that we can never run the traces generated by this kind of test.

If we want to run traces, we should either use a C program test, or we can make the kind of test suite we discussed where you paste a whole post-our-passes LLVM AOT module into a file and make traces from that.

ltratt: I'm still nervous that we have this weird hack in that's only for testing but it's for *all* testing. If, say, the hack only kicked in if you specifiy `YK_TESTING_GLOBAL_HACK` (or whatever) it would make the hacks existence a bit clearer, and scoped in nature?

vext01: I hear ya, but this code path is only executed for the trace compiler test suite. See, the hack is inside `createModuleForTraceCompilerTests()`.

ltratt: OK, fair point.

PR #494, opened 2022-02-13
Rethink how tracing is started when a location reaches the hot threshold

ltratt:
This PR introduces a new test which showed up more than one race condition in the location transition function.

The first problem was that I was, even by my standards, incredibly stupid: I created a dummy Location and locked that, forgetting that did not lock the original Location. Not only did we have a race condition, but because we didn't lock things properly, we were in undefined behaviour. The fix is to resuscitate an old idea: the Compiling state is now:

```rust
  Compiling(Arc<Mutex<Option<Box<CompiledTrace>>>>)
```

Thus the compiling thread and any "normal" thread can communicate whether compilation has finished safely. This does mean that while a location is in the Compiling state, one has to perform two locks. Since we use parking_lot's `Mutex` type, in the (common) uncontended case this is fairly cheap.

As a happy coincidence, moving (back) to this scheme also got rid of some code that turned pointers into usize's: it might well have been correct, but it was very fragile.

The second problem is that there was a subtle race condition when deciding whether it was safe to start tracing or not. This race condition did not lead to undefined behaviour, but meant that two
threads could start trying to start tracing on the same location. The  fix for this is be less clever when starting tracing: we don't go around in a loop when starting tracing. If we fail to start tracing, we give up, on the basis that if tracing is the right thing to do, there'll be another opportunity soon.

Notice that the `Compiled::non_null` function, needed by `ykrt` for testing purposes requires the cunning "only enable a feature when testing" Cargo hack from https://github.com/rust-lang/cargo/issues/2911#issuecomment-749580481.

### DIFF
--- a/ykrt/Cargo.toml
+++ b/ykrt/Cargo.toml
@@ -10,10 +10,13 @@ num_cpus = "1.13.1"
 parking_lot = "0.12.0"
 parking_lot_core = "0.9.1"
 strum = { version = "0.23.0", features = ["derive"] }
-yktrace = { path = "../yktrace" }
+yktrace = { path = "../yktrace", default-features = false, features = ["yk_testing"] }
 
 [build-dependencies]
 regex = "1.5.4"
+
+[dev-dependencies]
+ykrt = { path = ".", features = ["yk_testing"] }
 
 [features]
 jit_state_debug = []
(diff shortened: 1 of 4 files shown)


ltratt:
I'm closing this in favour of a larger PR which does this and more.

PR #496, opened 2022-02-15
More tests and fixes

ltratt:
This PR completes (I think...) the basic testing suite of the `Location` transition function in the sense that every statement is exercised by at least one test. There is probably more that we could do, but this puts us in a much better place to consider refactoring and extending this core feature.

The first commit (https://github.com/ykjit/yk/commit/b1f8e4ef5946567cb905aa20d3f8c2fd8282ced0) contains both a new test and two fixes. The remaining two commits (https://github.com/ykjit/yk/commit/23c1f9c202f095b80f91854ad1f5cd6fa815c936 and https://github.com/ykjit/yk/commit/b21e3823320584faf148c3842fd47702f6d8f0c3) add tests but didn't highlight any new bugs (whether because the tests are unimaginative or the code is correct is an exercise for the reader).

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -8,7 +8,6 @@ use std::{
     collections::VecDeque,
     ffi::c_void,
     marker::PhantomData,
-    mem::forget,
     ptr,
     sync::{
         atomic::{AtomicU32, AtomicUsize, Ordering},
@@ -147,8 +146,8 @@ impl MT {
                 }
             }
             TransitionLocation::StartTracing(kind) => start_tracing(kind),
-            TransitionLocation::StopTracing(hl) => match stop_tracing() {
-                Ok(ir_trace) => self.queue_compile_job(ir_trace, hl),
+            TransitionLocation::StopTracing(x) => match stop_tracing() {
+                Ok(ir_trace) => self.queue_compile_job(ir_trace, x),
                 Err(_) => todo!(),
             },
         }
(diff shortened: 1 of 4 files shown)


ltratt:
The good news is that this test suite has proven its worth: I've got another PR in the wings (which I can't easily open until this one is merged; I wish we could stage PRs!) which simplifies the transition stuff significantly. Importantly, the test suite found a really subtle bug I introduced while doing so (I inverted a conditional...), that otherwise would have gone unnoticed. It means, for example, that in that upcoming PR we can do things like https://github.com/ykjit/yk/commit/cc72f06b17c69f85e0b76ec99bbb6f917b427c6b.

vext01:
I've only reviewed the first commit for now. Will resume once the proposed comments have been added: i want to make sure I've understood what's going on before proceeding.

ykrt/src/mt.rs
```diff
@@ -715,6 +715,44 @@ mod tests {
...
+            .join()
+            .unwrap();
+        }
+
+        assert_eq!(
+            hotlocation_discriminant(&loc),
+            Some(HotLocationDiscriminants::Tracing)
+        );
+        assert_eq!(mt.transition_location(&loc), TransitionLocation::NoAction);
+        assert_eq!(
```
vext01, line 754: I'd be tempted to repeat that last assertion once more.

ltratt: That won't really test anything meaningful.

vext01: It tests (to a limited degree) that further calls to `transition_location` don't exit the "do not trace state".

I'm assuming that once a location is forsaken, it's forsaken for good.

ltratt: It is -- but this is something I'd like to change soon. [An additional problem is that testing it once more doesn't really prove much -- in a sense you have to test it infinitely often, which will make CI unhappy :)]

ltratt:
@vext01 If you're OK with me doing so, I think this might be ready for squashing?

vext01:
Please go ahead.

ltratt:
Squashed.

vext01:
Scary. Please squash.

ltratt:
Squashed.

PR #498, opened 2022-02-21
Simplify location transition

ltratt:
This PR simplifies the location transition code bit by bit, using the test suite from #496 to give us reasonable confidence that we aren't introducing too many new bugs. You might notice that this PR deletes more code than it adds!

### DIFF
--- a/ykrt/src/location.rs
+++ b/ykrt/src/location.rs
@@ -3,7 +3,7 @@
 use std::{
     convert::TryFrom,
     sync::{
-        atomic::{AtomicUsize, Ordering},
+        atomic::{AtomicPtr, AtomicUsize, Ordering},
         Arc,
     },
 };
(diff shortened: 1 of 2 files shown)


vext01:
I think I get the rough idea behind dbfbacd, but it's a bit scary.

I'm going to have to trust you to some degree.

If that is OK by you, we can merge this.

ltratt:
Does the comment at https://github.com/ykjit/yk/commit/dbfbacd65ce5ba747e5b381e040734bf49704004#diff-9fa85f3ab0ab08c2bb4337dd17d47605f520c2780158e50e66eff7cd9dc1b29bR363 help, is it not good enough, or is it as-good-as-it-can-probably-be-without-you-spending-2-days-understanding-the-PR? I know myself that sometimes there's a limit to how much I can reasonably understand!

vext01:
It really helped! Without it I wouldn't have had a clue what was going on.

ltratt:
Then if you're as happy as you think you can reasonably be, you should feel free to merge.

vext01:
Sounds good.

bors r+

PR #500, opened 2022-02-23
Start work on the stopgap interpreter.

### LINKED ISSUE #443: Guards currently abort.
They should run a stop-gap interpreter.

ptersilie:
Companion PR: https://github.com/ykjit/ykllvm/pull/22

The stopgap interpreter is called during a guard failure and interprets
AOT IR until it reaches a point where the compiled interpreter can take
over (i.e. when reaching the control point again) or until it interprets
the return of the main interpreter loop.

The interpreter interacts with the LLVM module via llvm-sys rather than
using a more abstracted (and user-friendly) interface like llvm-ir. The
reason is that llvm-ir parses modules in its entirety, including parts
of the module the stopgap interpreter won't touch. Using llvm-sys we
only parse what we actually need to get back to the control point.

In order to initialise the stopgap interpreter we need to create a
mapping from stackmap values to variables (i.e. instructions) in the AOT
module. This is being setup in `jitmodbuilder.cc` during guard creation
and then passed through to the stopgap interpreter via the
deoptimisation call. We also need to pass the position in the AOT module
from where interpretation needs to start, which we previously didn't do
either.

In order to exit properly after interpreting a return from the main
interpreter loop, the stopgap interpreter returns a boolean to tell the
compiled interpreter that it has reached a return and needs to act
accordingly (see ykllvm changes).

Note, that the interpreter is still missing a lot of functionality and
takes some liberties to get the tests working (e.g. skipping function
calls).

Other things that need work in future PRs:
- The stopgap interpreter and control point need to return the value that was interpreted (currently always returns 0)
- The `__llvm_deoptimize` code currently depends on the calling frame to push `rbp` to the stack. This may not always be the case as LLVM may optimize this to free up `rbp` for general use. Instead we need to get the stack size from the stackmap and return this to the `__llvm_deoptimize` call in order to recover the calling frame and return from it. (possibly separate PR)

Fixes #443


c_tests/tests/simple.c
```diff
@@ -69,10 +73,9 @@ int main(int argc, char **argv) {
     res += 2;
     i--;
   }
-  abort(); // FIXME: unreachable due to aborting guard failure earlier.
+  printf("exit");
```
ltratt, line 76: Is this print testing anything useful?

ykcapi/src/lib.rs
```diff
@@ -107,29 +109,73 @@ impl Registers {
...
+    fname: *const i8,
+}
+
+/// Address and size of a vector containing AOTVars. Mirrors the struct defined in
+/// ykllvmwrap/jitmodbuilder.cc.
+#[derive(Debug)]
+#[repr(C)]
+pub struct AOTMap {
+    addr: *const c_void,
+    size: usize,
```
ltratt, line 128: Is `size` the size in bytes or the length in terms of ... ?

ykcapi/src/lib.rs
```diff
@@ -107,29 +109,73 @@ impl Registers {
...
     sm_size: usize,
+    aotmap: &AOTMap,
+    curpos: &CurPos,
     retaddr: usize,
     rsp: *const c_void,
-) {
+) -> bool {
     // FIXME: remove once we have a stopgap interpreter.
     eprintln!("jit-state: stopgap");
+    eprintln!("{:?} {} {} {:?}", sm_addr, sm_size, retaddr, rsp);
```
ltratt, line 155: Debug code I think.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -285,25 +295,33 @@ class JITModBuilder {
...
+                               size_t CurInstrIdx) {
     assert(isa<BranchInst>(I));
     BranchInst *BI = cast<BranchInst>(I);
     LLVMContext &Context = JITMod->getContext();
 
     if (BI->isUnconditional())
       return nullptr; // Control-flow can not diverge. No guard required.
 
+    if (BI->getCondition() == NewControlPointCall) {
+      // We don't need to guard against the control point return check, as this check can
```
ltratt, line 308: Not sure I understand this comment.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -470,29 +488,117 @@ class JITModBuilder {
       Instruction *CurrentInst = &CurrentBB->back();
       DominatorTree DT(*JITFunc);
       std::vector<Value *> LiveVals = getLiveVars(DT, CurrentInst);
+      // Naturally the current instruction is live too but wasn't included due
+      // to the way DominatorTree works.
+      LiveVals.push_back(CurrentInst);
+
+// Add arguments for stackmap pointer and size.
+#if defined(__x86_64)
```
vext01, line 496: I think there's a cluster of attributes that cache platform specific types. Should we move this up next to them?

c_tests/tests/simple.c
ptersilie: Yes, it's so that we know if we are not executing it twice, once in the stopgap interpreter, and then again in the compiled interpreter. In other words, it checks that the new control point return design does what we think it does and exists properly if the stopgap interpreter has interpreted a return.

ykcapi/src/lib.rs
ptersilie: Bytes.

yksgi/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+            vars: HashMap::new(),
+        }
+    }
+
+    /// Get the value of the variable `key` in this frame.
+    fn get(&self, key: &LLVMValue) -> Option<&SGValue> {
+        self.vars.get(key)
+    }
+
+    /// Insert new variable into this frame.
```
ltratt, line 48: The comment ("insert *new variable") doesn't match the code (which adds/updates a variable).

yksgi/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "yksgi"
+version = "0.1.0"
+authors = ["The Yk Developers"]
+edition = "2018"
```
vext01, line 5: may as well use rust 2021

yksgi/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+/// interpreting LLVM IR.
+pub struct SGInterp {
+    /// LLVM IR module we are interpreting.
+    module: LLVMModule,
+    /// Current frames.
+    frames: Vec<Frame>,
+    /// Current instruction being interpreted.
+    pc: LLVMValue,
+    /// Last basic block that was interpreted.
+    lastbb: Option<LLVMBasicBlock>,
```
ltratt, line 68: This looks like it owns the basic block? Should it be a reference, or a basic block index?

yksgi/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    pub fn as_str(&self) -> String {
+        format!("SGValue ( {}, {:?} )", self.val, unsafe { self.ty.as_str() })
+    }
+}
+
+/// A frame holding live variables.
+struct Frame {
+    vars: HashMap<LLVMValue, SGValue>,
```
vext01, line 33: Later we could consider using a FxHashMap (assuming we value performance over security).

ykcapi/src/lib.rs
ltratt: Ok, let's document that.

c_tests/tests/simple.c
ltratt: Aha!

yksgi/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+        } else if var.is_constant() {
+            llvmbridge::llvm_const_to_sgvalue(var.0)
+        } else {
+            // GlobalVariable, Function, etc.
+            todo!()
+        }
+    }
+
+    /// Lookup the value of variable `var` in the current frame.
+    unsafe fn setvar(&mut self, var: LLVMValue, val: SGValue) {
```
ltratt, line 117: Can we use consistent names for lookup/setvar? `var_set` and `var_lookup` are often good bets in my experience!

yksgi/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+            self.pc = LLVMValue(LLVMGetNextInstruction(self.pc.0));
+        }
+    }
+
+    unsafe fn add(&mut self) {
+        let op1 = self.pc.get_operand(0);
+        let op2 = self.pc.get_operand(1);
+        let val1 = self.lookup(&op1);
+        let val2 = self.lookup(&op2);
+        // FIXME: Can overflows happen? What do we do?
```
vext01, line 167: Good question!

It depends if the `add` was qualified with `nuw` or `nsw` (no unsigned overflow, no signed overflow).

https://llvm.org/docs/LangRef.html#add-instruction

ltratt, line 167: Overflow can happen, but what we should do I don't know! I would use `let res = val1.val.checked_add(val2).unwrap()` as we'll get a deterministic `panic` when we eventually encounter overflow in a real program.

yksgi/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    /// Interpret branch instruction `instr`.
+    pub unsafe fn branch(&mut self) {
+        debug_assert!(self.pc.is_br());
+        let succ = if LLVMIsConditional(self.pc.0) == 0 {
+            LLVMBasicBlock::new(LLVMGetSuccessor(self.pc.0, 0))
+        } else {
+            let cond = LLVMValue::new(LLVMGetCondition(self.pc.0));
+            let val = self.lookup(&cond);
+            if val.val == 1 {
```
ltratt, line 190: Can `val.val` only be 0 or 1?

yksgi/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+        self.lastbb = Some(LLVMBasicBlock::new(LLVMGetInstructionParent(self.pc.0)));
+        self.pc = succ.first();
+    }
+
+    pub unsafe fn call(&mut self) {
+        // FIXME: Properly implement calls.
+        let func = LLVMGetCalledValue(self.pc.0);
+        let mut size = 0;
+        let name = CStr::from_ptr(LLVMGetValueName2(func, &mut size)).to_str().unwrap();
+        if name.starts_with("llvm.lifetime") {
```
ltratt, line 205: Why no behaviour for this?

yksgi/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+            return;
+        } else if name.starts_with("puts") || name.starts_with("printf") {
+            let op = self.pc.get_operand(0);
+            let op2 = op.get_operand(0);
+            let op3 = LLVMGetInitializer(op2.0);
+            let mut l = 0 as usize;
+            let s = LLVMGetAsString(op3, &mut l);
+            //libc::write(2, s as *const libc::c_void, l as libc::size_t - 1);
```
ltratt, line 213: Debug code.

yksgi/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+        let op1 = self.pc.get_operand(0);
+        let op2 = self.pc.get_operand(1);
+        let val1 = self.lookup(&op1);
+        let val2 = self.lookup(&op2);
+        // FIXME: Can overflows happen? What do we do?
+        let res = val1.val + val2.val;
+        self.setvar(self.pc, SGValue::new(res, self.pc.get_type()));
+    }
+
+    /// Convert value to new type of same bit size (without changing any bits).
```
vext01, line 172: I'm confused by this part.

LLVM is SSA, so there should be no in-place mutations of variable types, right?

yksgi/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+            let op3 = LLVMGetInitializer(op2.0);
+            let mut l = 0 as usize;
+            let s = LLVMGetAsString(op3, &mut l);
+            //libc::write(2, s as *const libc::c_void, l as libc::size_t - 1);
+            libc::puts(s);
+        } else if name.starts_with("__ykrt_control_point"){
+            todo!()
+        }
+        else {
+            // FIXME: For now just skip all other calls.
```
ltratt, line 219: Can we `todo` this branch?

yksgi/src/llvmbridge.rs
```diff
@@ -0,0 +1,210 @@
...
+        LLVMTypeKind::LLVMIntegerTypeKind => {
+            let _width = LLVMGetIntTypeWidth(ty);
+            let val = LLVMConstIntGetZExtValue(c) as u64;
+            SGValue::new(val, LLVMType(ty))
+        }
+        _ => todo!(),
+    }
+}
+
+/// Get the original instructions of a live variable before it is being
```
vext01, line 161: nitpick: I'd re-word this comment, as it reads a little clumsily.

yksgi/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    unsafe fn switch(&mut self) {
+        debug_assert!(self.pc.is_switch());
+        let cond = self.pc.get_operand(0);
+        let val = self.lookup(&cond);
+        let num_dests = LLVMGetNumSuccessors(self.pc.0);
+        let mut succ = None;
+        for i in 1..num_dests { // Skip the default case.
+            let case = llvmbridge::llvm_const_to_sgvalue(LLVMGetOperand(self.pc.0, i * 2));
+            // Check the condition against each case.
```
ltratt, line 332: Not sure I understand the comment.

ltratt:
I have a number of minor comments, but nothing major!

yksgi/src/lib.rs
vext01: I was curious. llvm-sys defines the type as an empty enum: https://docs.rs/llvm-sys/latest/llvm_sys/enum.LLVMBasicBlock.html

Is this a way of tagging an opaque pointer or something?

yksgi/src/lib.rs
vext01: Ah, we both commented on this aspect. But in different places.

See [here](https://github.com/ykjit/yk/pull/500#discussion_r812843754)

yksgi/src/lib.rs
vext01: I think it's because it doesn't really do anything in the context of our interpreter:
https://llvm.org/docs/LangRef.html#int-lifestart

vext01:
Same here. No major issues.

yksgi/src/lib.rs
ptersilie: This is a bit confusing, since llvm-sys has types of the same name. But these are my own types defined in `llvmbridge.rs`. So the `LLVMBasicBlock` type above is actuallly just a wrapper around an `LLVMBasicBlockRef`: `pub struct LLVMBasicBlock(pub LLVMBasicBlockRef);`

yksgi/src/lib.rs
ltratt: OK, a quick comment explaining why would be helpful!

yksgi/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+        SGInterp {
+            module,
+            frames: vec![Frame::new()],
+            pc: instr,
+            lastbb: None,
+        }
+    }
+
+    /// Add a live variable and its value to the current frame.
+    pub unsafe fn init_live(&mut self, bbidx: usize, instridx: usize, fname: &CStr, val: u64) {
```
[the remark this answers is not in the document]
ptersilie: This function copies over the live variables we get from the deopt step. So it needs to do slightly more work than assigning a variable during interpretation. But fundamentally they are the same.

yksgi/src/lib.rs
ptersilie: Sounds good.

yksgi/src/lib.rs
ltratt: I'm going to be rude and say that `LLVMBasicBlock` is then not a great name :) Why bother even wrapping it? Can't we just use `LLVMBasicBlockRef` as is? As a bonus, the name is better!

vext01: Which is just a raw pointer that can be freely copied, I think?

If so, the block isn't owned.

yksgi/src/lib.rs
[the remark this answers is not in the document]
ltratt: I think `var_init` (based on my comment below about consistent naming) would be good. Then you don't need to use the confusing "live variable" terminology.

yksgi/src/lib.rs
ptersilie: > as we'll get a deterministic panic

Good idea. Will do that for now. But Edd seems to have found the right answer. If it's `nuw` we're good; if it's `nsw` we need to return a poisoned value, which marks it as UB. Though I'm not entirely sure what that means for the interpreter and how this information back to the compiled interpreter (if that's even possible).

yksgi/src/lib.rs
ltratt: Good idea.

yksgi/src/lib.rs
ptersilie: I'm wrapping it so that I can implement some functions on it.

ltratt: OK, so let's at least call it `LLVMBasicBlockRef` or similar.

yksgi/src/lib.rs
ptersilie: Nothing is mutated in place here. The old key is `src` (which is the operand of `pc`) and the new key is `pc`.

ltratt: I think "convert" is misleading: this is the equivalent, I think, of Rust's `transmute`?

yksgi/src/lib.rs
ltratt: I would much rather whitelist the functions we want to skip, even if there are a lot of them, so that we don't find ourselves skipping things unexpectedly in the future. I'd just stick a `print` in here, run the tests, capture all of the function names, then say `if function not in <big array> { panic!(...) }`.

yksgi/src/lib.rs
ptersilie: When I made the wrapper I didn't know `llvm-sys` has a type of the same name. If I call it `LLVMBasicBlockRef` though then it's again the same name as in `llvm-sys`. Maybe `LLVMBasicBlockRefWrapper` or `WBasicBlockRef`?

ltratt: Do we use both names in this namespace?

yksgi/src/lib.rs
ptersilie: The LLVM docs called it "convert" so I just did the same. I can just do s/Convert/Cast.

ltratt: "Cast" sounds good to me!

ykcapi/src/lib.rs
ptersilie: I was talking nonsense. This is actually the length. Fixed in fb19d6c

ykcapi/src/lib.rs
ptersilie: Fixed in da99e57

yksgi/src/lib.rs
ptersilie: Fixed in da99e57

ykllvmwrap/src/jitmodbuilder.cc
ltratt: Works for me!

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Not in `jitmodbuilder.cc` it seems.

yksgi/src/lib.rs
ptersilie: Fixed in fcae4c6

yksgi/Cargo.toml
ptersilie: We don't in any of the other cargo files though.

yksgi/src/lib.rs
ltratt: On reflection, I suggest moving the "add" part of this to the `init_live` function (which I *think* is the only time this functionality is needed?) and calling this function `update_var` so that it can only update an existing slot.

yksgi/src/lib.rs
ptersilie: Fixed in 710be13

yksgi/src/lib.rs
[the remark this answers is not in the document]
ptersilie: Fixed in 710be13

yksgi/src/lib.rs
ptersilie: We might in the case of `LLVMBasicBlockRef`, but probably not in the case of `LLVMBasicBlock`.

yksgi/src/lib.rs
ptersilie: It's also used in `set_var`. But I just realised that due to LLVM being SSA, we should never update a variable anyway, so this should just be insert. Let me double check this is actually the case and there wasn't a reason I added the update case.

yksgi/src/lib.rs
ltratt: If we want to use both names, then my suggestion is: don't import them and always qualify them. In other words, if two modules define `X` then don't `use A::X` or `use B::X`, but always spell out `A::X` or `B::X` in types.

ptersilie: I'm currently not importing them. As I said, I didn't know `llvm-sys` even defined a `LLVMBasicBlock` type, which is why it used this name for my wrapper.

ltratt: Then why don't we just use `LLVMBasicBlockRef` for this struct?

yksgi/src/lib.rs
ptersilie: Fixed in 10e488a

yksgi/src/lib.rs
ptersilie: Fixed in 532ac4d

yksgi/src/lib.rs
ptersilie: Fixed in b100fc9

yksgi/src/lib.rs
ptersilie: Fixed in https://github.com/ykjit/yk/commit/e8a388b9053bac0fa3a8977bd59ba6d75b960457

yksgi/src/llvmbridge.rs
ptersilie: Fixed in ffd4e75

yksgi/src/lib.rs
ptersilie: So just do `struct LLVMBasicBlockRef(llvm-sys::LLVMBasicBlockRef)`? I prefer the current solution over this tbh.

ltratt: > So just do struct LLVMBasicBlockRef(llvm-sys::LLVMBasicBlockRef)?

The current name is confusing (both Edd and I misunderstood it). I don't mind too much what we change it too -- I made a suggestion but there may be better alternatives!

ptersilie: Fixed in [ad4e087](https://github.com/ykjit/yk/pull/500/commits/ad4e0879b40340cc9b7d404138216344872d9efd)

ptersilie:
I fixed the `unsafe`s as discussed in 8baf399 , and then did some more refactoring in [f68320b](https://github.com/ykjit/yk/pull/500/commits/f68320b3323f187cbc441af226c6bbe64ff98839) which allowed me to remove some more `unsafe`s in [9683f34](https://github.com/ykjit/yk/pull/500/commits/9683f34bf3bb978c7789ceabd912756883e867b5).

There's still some `unsafe`s left which aren't really necessary, but that would require implementing more `llvm-sys` functions on the structs, so I've left those for now.

yksgi/src/lib.rs
ltratt: That wasn't the way I was expecting... but it does work for me!

ltratt:
Why does https://github.com/ykjit/yk/pull/500/commits/8baf3994fc4ef31ba4fe28893fedaf9f8eea0aa9 introduce a number of `unwrap`s?

ltratt:
Ah, some disappear in https://github.com/ykjit/yk/pull/500/commits/f68320b3323f187cbc441af226c6bbe64ff98839.

ptersilie:
Ready for re-review. Once there's no more comments I would like to squash and then continue working on the tests so we can hopefully merge this soon-ish.

yksgi/src/lib.rs
```diff
@@ -0,0 +1,357 @@
...
+
+/// The stopgap interpreter. Used during guard failures to get back to the control point by
+/// interpreting LLVM IR.
+pub struct SGInterp {
+    /// LLVM IR module we are interpreting.
+    module: Module,
+    /// Current frames.
+    frames: Vec<Frame>,
+    /// Current instruction being interpreted.
+    pc: Value,
```
ltratt, line 66: We seem to call `unwrap` on this `Value` quite a lot, but I can't work out why. Any hints?

ptersilie:
> Why does [8baf399](https://github.com/ykjit/yk/commit/8baf3994fc4ef31ba4fe28893fedaf9f8eea0aa9) introduce a number of `unwrap`s?

Maybe `unwrap` is a bad name for this. This is to unwrap my wrapper structs, e.g. which previously was `self.pc.0`. I could maybe name it `self.pc.val()` or `self.pc.get()` to differentiate it from Rust's `unwrap`.

ltratt:
In Rust, `unwrap` strongly implies "this could fail and will `panic` if it does". Is that the case here?

ptersilie:
No, the implementation is literally `fn unwrap(&self) -> LLVMValueRef { self.0 }`. Yeah, `unwrap` was a poor name choice.

ltratt:
OK, let's rename it :)

ptersilie:
Done. Gotta love `gR`. :-)

ltratt:
It is useful!

@vext01 any comments?

yksgi/src/llvmbridge.rs
```diff
@@ -40,116 +42,140 @@ impl LLVMFunction {
...
+    pub fn first(&self) -> LLVMValue {
         self.instruction(0)
     }
 
-    pub unsafe fn instruction(&self, instridx: usize) -> LLVMValue {
-        let mut instr = LLVMGetFirstInstruction(self.0);
-        for _ in 0..instridx {
-            instr = LLVMGetNextInstruction(instr);
+    pub fn instruction(&self, instridx: usize) -> LLVMValue {
+        unsafe {
```
vext01, line 72: Maybe this is a little anal, but i always try to make unsafe blocks as small as possible.

Here the `for` line needn't be inside unsafe.

yksgi/src/llvmbridge.rs
```diff
@@ -40,116 +42,140 @@ impl LLVMFunction {
         Self(func)
     }
 
-    pub unsafe fn bb(&self, bbidx: usize) -> LLVMBasicBlock {
-        let mut bb = LLVMGetFirstBasicBlock(self.0);
-        for _ in 0..bbidx {
-            bb = LLVMGetNextBasicBlock(bb);
+    pub fn bb(&self, bbidx: usize) -> LLVMBasicBlock {
+        unsafe {
```
vext01, line 46: Similarly here, the scope of `unsafe` can be constrained.

yksgi/src/llvmbridge.rs
```diff
@@ -25,10 +25,12 @@ impl LLVMModule {
...
     }
 
-    pub unsafe fn function(&self, name: *const i8) -> LLVMFunction {
-        let func = LLVMGetNamedFunction(self.0, name);
-        debug_assert!(!func.is_null());
-        LLVMFunction::new(func)
+    pub fn function(&self, name: *const i8) -> LLVMFunction {
+        unsafe {
+            let func = LLVMGetNamedFunction(self.0, name);
+            debug_assert!(!func.is_null());
```
vext01, line 31: And is this line really unsafe?

yksgi/src/llvmbridge.rs
ltratt: Agreed, minimising `unsafe` blocks is important.

yksgi/src/llvmbridge.rs
ptersilie: Yes, all llvm-sys calls are unsafe.

yksgi/src/llvmbridge.rs
ptersilie: Can do, just means there'll be 3 `unsafe`s instead of one.

ptersilie: Given the unsafe was previously in the function, I thought this is equivalent without making the code worse to read.

ptersilie: I know, I just thought a for-loop can't possibly be unsafe. But I guess it'll prevent us from adding other unsafe stuff inside the block without noticing. So I'll change it.

yksgi/src/llvmbridge.rs
ptersilie: Oh I see what you meant now.

yksgi/src/llvmbridge.rs
ptersilie: Fixed in 9a253a4

yksgi/src/llvmbridge.rs
ptersilie: Fixed in 9a253a4

yksgi/src/llvmbridge.rs
ptersilie: Fixed in 9a253a4

yksgi/src/llvmbridge.rs
ltratt: > I just thought a for-loop can't possibly be unsafe.

You can put arbitrary code in the conditional, for example, so yes, it can be unsafe!

ptersilie:
Okay to squash?

ltratt:
Fine with me -- but please wait for @vext01 as he has unresolved comments above. [I suspect they're resolved and he just hasn't clicked "resolved" but I'm not 100% sure.]

vext01:
all resolved, but there are merge conflicts to deal with.

ltratt:
Yes, squashing + rebasing seems the right way to go. The conflicts don't look to be too many!

ptersilie:
Yeah was going to rebase after squash.

vext01:
This fixes #443. You might like to put that in the commit message to have it auto-closed.

ltratt:
Could you tidy up the new commits and force push something that's a bit easier for my stupid old head to review?

ptersilie:
I forgot to squash away the "Fix up" commits, though the other commits stand on their own. But if you're okay with it I can rebase against master to get rid of those merge commits as well.

ltratt:
Please do ahead.

ltratt:
"only"

yktrace/src/lib.rs
```diff
@@ -231,7 +233,37 @@ impl CompiledTrace {
...
 
-    pub fn exec(&self, ctrlp_vars: *mut c_void) -> u8 {
+   #[naked]
+   #[no_mangle]
+    /// Taking the deoptimisation path during a guard failure causes the epilogue of the compiled
+    /// trace to be skipped. This means that used callee-saved registers (CSRs) are not restored.
+    /// Until we've figured out how to restore only the used registers we take the sledge hammer
+    /// approach and save and restore all CSRs here.
+    /// OPT: Find a way to only restore needed registers (ideally right within the deopt code).
+    pub extern "C" fn exec(&self, ctrlp_vars: *mut c_void) -> u8 {
```
ltratt, line 243: This should be guarded as x86_64 only.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -394,9 +394,18 @@ class JITModBuilder {
     }
   }
 
-  void handlePHINode(Instruction *I, BasicBlock *BB) {
+  void handlePHINode(Instruction *I, BasicBlock *BB, size_t CurBBIdx, size_t CurInstrIdx) {
     Value *V = ((PHINode *)&*I)->getIncomingValueForBlock(BB);
-    VMap[&*I] = getMappedValue(V);
+    // From the point of view of the compiled trace it's enough to just update the `Vmap` with the
```
ltratt, line 399: s/Vmap/VMap/ ?

ykcapi/src/lib.rs
```diff
@@ -102,11 +102,12 @@ impl Registers {
...
             )
         }
-        let mut val = self.read_from_stack(id.try_into().unwrap());
+        let val = self.read_from_stack(id.try_into().unwrap());
         // Due to the return address being pushed to the stack before we store RSP, its value is
         // off by 8 bytes.
         if id == 7 {
-            val += std::mem::size_of::<usize>();
+            todo!(); // Check this is still true now that llvm_deoptimize is a naked function.
+            // val += std::mem::size_of::<usize>();
```
ltratt, line 110: Commented code.

ltratt:
I only have some minor comments.

tests/tests/const_global.c
```diff
@@ -8,7 +8,7 @@
 //     jit-state: stop-tracing
 //     --- Begin jit-pre-opt ---
 //     ...
-//     define internal void @__yk_compiled_trace_0(%YkCtrlPointVars* %0, i64* %1, i64 %2) {
+//     define internal i8 @__yk_compiled_trace_0(%YkCtrlPointVars* %0, i64* %1, i64 %2) {
```
vext01, line 11: What's the `i8` return again?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -369,9 +394,18 @@ class JITModBuilder {
...
+    // incoming value `V`. However, this means there's no corresponding instruction in the JITMod
+    // and thus there's nothing mapping back to it during a guard failure. Creating a
+    // (non-sensical) select instruction here makes sure there's a mapping from a JITMod value back
```
vext01, line 402: I don't think `nonsensical` is the right word. Maybe `degenerate`. You might spin the comment in the direction of "an assignment would work, but LLVM doesn't have one, so we emulate one with a degenerate select statement"?

ykllvmwrap/src/jitmodbuilder.cc
vext01: Could you use [this](https://github.com/ykjit/yk/blob/41a38f74f60e68f28474e7e230ff1f52db6d13de/ykllvmwrap/src/jitmodbuilder.cc#L187) instead?

yksgi/src/lib.rs
```diff
@@ -0,0 +1,470 @@
...
+        self.vars.get(key)
+    }
+
+    /// Add new variable `key` with value `val`.
+    fn add(&mut self, key: Value, val: SGValue) {
+        self.vars.insert(key, val);
+    }
+}
+
+macro_rules! signext {
```
vext01, line 59: `FIXME: unit tests for this`?

yksgi/src/lib.rs
```diff
@@ -278,30 +279,40 @@ impl SGInterp {
...
-                // for now and always return the first element and hope the tests won't mind.
-                eprintln!("FIXME: GEP always returns first element!");
+
+            if curty.is_struct() {
+                let off = LLVMOffsetOfElement(layout, curty.get(), u32::try_from(idx).unwrap());
+                offset += off;
+                let nty = LLVMStructGetTypeAtIndex(curty.get(), u32::try_from(idx).unwrap());
+                curty = Type::new(nty);
+            } else {
+                // Pointer into an array
```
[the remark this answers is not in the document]
vext01: I think so.

https://llvm.org/docs/LangRef.html#getelementptr-instruction:
> The ‘getelementptr’ instruction is used to get the address of a subelement of an [aggregate](https://llvm.org/docs/LangRef.html#t-aggregate) data structure.

https://llvm.org/docs/LangRef.html#t-aggregate:
> Aggregate Types are a subset of derived types that can contain multiple member types. [Arrays](https://llvm.org/docs/LangRef.html#t-array) and [structs](https://llvm.org/docs/LangRef.html#t-struct) are aggregate types. [Vectors](https://llvm.org/docs/LangRef.html#t-vector) are not considered to be aggregate types.

ykllvmwrap/src/jitmodbuilder.cc
ltratt: I think perhaps the main concept we can want to get across is "this select instruction doesn't do anything bad". What the right term for that is I'm not immediately sure!

yksgi/src/lib.rs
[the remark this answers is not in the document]
ptersilie: Yes, I believe so.

vext01:
Small comments from me, otherwise good work Lukas!

yktrace/src/lib.rs
ptersilie: Fixed in 182a1fa

yksgi/src/lib.rs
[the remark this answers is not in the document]
ltratt: That document is a bit unclear about whether they're the *only* aggregate types, though. We should check (how, I dunno...).

ykcapi/src/lib.rs
ptersilie: That was on purpose as I'm not sure yet if this is still needed. Although you're probably gonna tell me to remove it nonetheless.

ltratt: `todo`s I like -- keep that! Commented code I don't like -- remove that!

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Fixed in 927836a

tests/tests/const_global.c
ptersilie: One of these: https://github.com/ykjit/yk/pull/500/files#diff-9fa85f3ab0ab08c2bb4337dd17d47605f520c2780158e50e66eff7cd9dc1b29bR42

```rust
const TRACE_RETURN_SUCCESS: u8 = 0;
const TRACE_GUARDFAIL_CONTINUE: u8 = 1;
const TRACE_GUARDFAIL_RETURN: u8 = 2;
```

Though this is likely subject to change at some later stage.

yksgi/src/lib.rs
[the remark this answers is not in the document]
ptersilie: As usual the source code gives a more conclusive answer: https://llvm.org/doxygen/Type_8h_source.html#l00256

ykcapi/src/lib.rs
ptersilie: Fixed in 08d7ccb

yksgi/src/lib.rs
ptersilie: Fixed in 2920f50

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Since this is used for `size_t`s we get from C, I think maybe `getIndexType` (https://llvm.org/doxygen/classllvm_1_1DataLayout.html#abbd9ed911d1d58b73f9571f300cc6a7b) is more appropriate, which is typically the same anyway.

ptersilie: Ah, that doesn't return an `IntegerType` so this won't work. According to [1] we should be able to just use `IntPtrTy` for `size_t` as well.

ptersilie: Fixed in 7799079

ptersilie:
I believe this addresses all comments.

ltratt:
Please squash ("squash" also means "rebase", so you never need further permission to do rebasing!).

ptersilie:
Hmm, the tests worked when I ran them manually, so will have to investigate why they fail in bors.

ptersilie:
Think this can be kicked again now.

PR #501, opened 2022-02-23
Add a concept of "trace failure threshold"

ltratt:
The aim of this PR (achieved in the final commit https://github.com/ykjit/yk/commit/14dd81bb7f883d18720f331a418f51c4762660d4) is to allow us to retry tracing a `Location` several times if it doesn't succeed, before giving up forever and marking it `DontTrace`.

In order to get to that point, we first change the `HotLocation` enum into a struct roughly from:

```rust
enum HotLocation { ... }
```

to:

```rust
enum HotLocationKind { ... }
struct HotLocation {
  kind: HotLocationKind
}
```

This allows us to store information such as "how often has this location failed tracing" in the `HotLocation` struct.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -18,7 +18,7 @@ use num_cpus;
 use parking_lot::{Condvar, Mutex, MutexGuard};
 use std::lazy::SyncLazy;
 
-use crate::location::{HotLocationKind, Location, LocationInner};
+use crate::location::{HotLocation, HotLocationKind, Location, LocationInner};
 use yktrace::{start_tracing, stop_tracing, CompiledTrace, IRTrace, TracingKind};
 
 // The HotThreshold must be less than a machine word wide for [`Location::Location`] to do its
@@ -194,9 +194,9 @@ impl MT {
                         // To avoid racing with another thread that may also try starting to trace this
                         // location at the same time, we need to initialise and lock the Location, which we
                         // perform in a single step.
-                        let hl_ptr = Box::into_raw(Box::new(HotLocationKind::Tracing(Arc::clone(
-                            &mtt.tracing,
-                        ))));
+                        let hl_ptr = Box::into_raw(Box::new(HotLocation {
+                            kind: HotLocationKind::Tracing(Arc::clone(&mtt.tracing)),
+                        }));
                         let new_ls = LocationInner::new().with_hotlocation(hl_ptr).with_lock();
                         debug_assert!(!ls.is_locked());
                         match loc.compare_exchange(ls, new_ls, Ordering::Acquire, Ordering::Relaxed)
@@ -252,7 +252,7 @@ impl MT {
                 }
             }
             let hl = unsafe { ls.hot_location() };
-            match hl {
+            match &hl.kind {
                 HotLocationKind::Compiled(ctr) => {
                     loc.unlock();
                     return TransitionLocation::Execute(*ctr);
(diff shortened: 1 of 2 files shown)


ykrt/src/mt.rs
```diff
@@ -75,6 +82,23 @@ impl MT {
...
+    /// Return this `MT` instance's current trace failure threshold. Notice that this value can be
+    /// changed by other threads and is thus potentially stale as soon as it is read.
+    pub fn trace_failure_threshold(&self) -> TraceFailureThreshold {
+        self.inner.trace_failure_threshold.load(Ordering::Relaxed)
+    }
+
+    /// Set the threshold at which a `Location` from which tracing has failed multiple times is
+    /// marked as "do not try tracing again".
+    pub fn set_trace_failure_threshold(&self, trace_failure_threshold: TraceFailureThreshold) {
+        if trace_failure_threshold < 1 {
```
vext01, line 94: Presumably a value of zero would effectively disable the JIT?

ltratt: Dunno -- I don't even want to think about what it does (or should...) mean!

vext01:
Looks good, but do i see it right that no test actually checks for the `DontTrace` state? Only the events leading up to it?

ltratt:
There are two tests that check for `DontTrace`, one of which is touched in this diff (https://github.com/ykjit/yk/pull/501/files#diff-9fa85f3ab0ab08c2bb4337dd17d47605f520c2780158e50e66eff7cd9dc1b29bR596).

vext01:
Not sure how I missed that.

bors r+

vext01:
Failures in release mode testing:
```
---- mt::tests::threaded_threshold stdout ----
thread '<unnamed>' panicked at 'assertion failed: c1.is_counting()', ykrt/src/mt.rs:512:21
thread '<unnamed>' panicked at 'assertion failed: c2.is_counting()', ykrt/src/mt.rs:515:21
thread '<unnamed>' panicked at 'assertion failed: c1.is_counting()', ykrt/src/mt.rs:512:21
thread '<unnamed>' panicked at 'assertion failed: c1.is_counting()', ykrt/src/mt.rs:512:21
thread '<unnamed>' panicked at 'assertion failed: c1.is_counting()', ykrt/src/mt.rs:512:21
thread '<unnamed>' panicked at 'assertion failed: c3.is_counting()', ykrt/src/mt.rs:518:21
thread '<unnamed>' panicked at 'assertion failed: c1.is_counting()', ykrt/src/mt.rs:512:21
thread '<unnamed>' panicked at 'assertion failed: c1.is_counting()', ykrt/src/mt.rs:512:21
thread '<unnamed>' panicked at 'assertion failed: c4.is_counting()', ykrt/src/mt.rs:521:21
thread '<unnamed>' panicked at 'assertion failed: c1.is_counting()', ykrt/src/mt.rs:512:21
...
```

ltratt:
Apparently I have not learnt from the last PR to do `test --release`!

ltratt:
Hmm, I cannot get this to fail with `test --release` at my end. This might need a bit of investigation...

ltratt:
OK, I can replicate this: there's something non-deterministic going on, and it triggers perhaps 1 time in 20. I'm unsure yet if it's related to this PR or not.

ltratt:
Of course, I just thought to do the easy test: the failure happens on master too. I suggest we try to merge this PR again, since the failure isn't related to this PR, while I work out what's going on.

PR #503, opened 2022-02-23
Make sure the test doesn't cause a transition to the Tracing state.

ltratt:
Although it was very hard to trigger (it took me quite a while to work out that, for example, I had to run the full test suite, and not over an SSH tunnel!), the `threaded_threshold` test was broken: it could, if the stars aligned, transition a `Location` often enough to move it to the `Tracing` state, at which point the test's assertions were then incorrect. This commit fixes that fault in the test, but makes sure that the test still runs for enough time to generate considerable contention.

This fixes the failure we observed in https://github.com/ykjit/yk/pull/501 -- but which had nothing to do with that PR!


vext01:
Nice one!

bors r+

PR #507, opened 2022-02-25
jit_state_env fix and a build system sanity check.

vext01:
The first commit ensures that the system checks `YKD_PRINT_JITSTATE` again.

The second is something that has been on my mind for a  while. There are two cargo features in our system which we need for testing and debugging, and which slow the system down:
 - `yk_testing`
 - `jit_state_debug`
 
It occurs to me that by building only what is necessary to use (but not test or debug) the JIT, we should never pull in those features, otherwise we incur unnecessary slowdowns.

["building only what is necessary to use (but not test or debug) the JIT" is defined as: building only the ykcapi shared object, and without enabling extra features on the command line]

The second commit ensures this, and adds a check to CI to make sure we always maintain this property.

I'm raising this a draft, as I have a question: should "yk_testing" and "jit_state_debug" be merged into one?

By default these features are only ever enabled together, but that's not to say that a user might not want to use jit state debugging without testing support...

Thoughts?

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -22,6 +22,14 @@ export PATH=${CARGO_HOME}/bin/:$PATH
 
 rustup toolchain install nightly --allow-downgrade --component rustfmt
 
+# There are some feature-gated testing/debugging switches which slow the JIT
+# down a bit. Check that if we build the system without tests, those features
+# are not enabled.
+for mode in "" "--release"; do \
+    cargo -Z unstable-options build ${mode} --build-plan -p ykcapi | \
+        awk '/yk_testing/ { ec=1 } /jit_state_debug/ { ec=1 } END {exit ec}'; \
+done
+
 cargo fmt --all -- --check
 
 # Build LLVM for the C tests.
--- a/ykrt/src/lib.rs
+++ b/ykrt/src/lib.rs
@@ -14,11 +14,11 @@ pub(crate) mod mt;
 pub use self::location::Location;
 pub use self::mt::{HotThreshold, MT};
 
-#[cfg(feature = "jit_state_debug")]
+#[cfg(feature = "yk_jitstate_debug")]
 static JITSTATE_DEBUG: SyncLazy<bool> = SyncLazy::new(|| env::var("YKD_PRINT_JITSTATE").is_ok());
 
 /// Print select JIT events to stderr for testing/debugging purposes.
-#[cfg(feature = "jit_state_debug")]
+#[cfg(feature = "yk_jitstate_debug")]
 pub fn print_jit_state(state: &str) {
     if *JITSTATE_DEBUG {
         eprintln!("jit-state: {}", state);
--- a/ykrt/src/lib.rs
+++ b/ykrt/src/lib.rs
@@ -6,8 +6,21 @@
 #![allow(clippy::type_complexity)]
 #![allow(clippy::new_without_default)]
 
+use std::{env, lazy::SyncLazy};
+
 mod location;
 pub(crate) mod mt;
 
 pub use self::location::Location;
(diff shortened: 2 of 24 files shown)


vext01:
First commit message should say "YKD_PRINT_JITSTATE" not "YK_PRINT_JITSTATE". Will fix later.

.buildbot.sh
```diff
@@ -22,6 +22,13 @@ export PATH=${CARGO_HOME}/bin/:$PATH
 
 rustup toolchain install nightly --allow-downgrade --component rustfmt
 
+# Check that if we build the system without support for testing and debugging
```
ltratt, line 25: I'm struggling to parse this sentence.

ykrt/src/lib.rs
```diff
@@ -6,8 +6,23 @@
...
 
 pub use self::location::Location;
 pub use self::mt::{HotThreshold, MT};
+
+#[cfg(feature = "jit_state_debug")]
+const JIT_STATE_ENV: &str = "YKD_PRINT_JITSTATE";
+
+/// Print select JIT events to stderr for testing/debugging purposes.
+#[cfg(feature = "jit_state_debug")]
+pub fn print_jit_state(state: &str) {
```
ltratt, line 22: This means we do an env var really frequently: I suggest we make this a global (like we do in `mt.rs` for the other thing). I also suspect this might better live in `mt.rs` too?

.buildbot.sh
vext01: I struggled to write it too. I'll see if I can come up with something better.

vext01:
And what about:

> I have a question: should "yk_testing" and "jit_state_debug" be merged into one?

.buildbot.sh
ltratt: Thanks!

ltratt:
Oops, I missed this. Sorry:

> I have a question: should "yk_testing" and "jit_state_debug" be merged into one?

I don't think so, at least not yet. We are very likely to want `yk_testing` without `jit_state_debug` IMHO.

.buildbot.sh
ltratt: Thanks!

vext01:
my next question then: should "jit_state_debug" have a "yk_" prefix for consistency?

ltratt:
> my next question then: should "jit_state_debug" have a "yk_" prefix for consistency?

Good idea -- yes!

ykrt/src/lib.rs
```diff
@@ -6,8 +6,23 @@
...
+    if env::var(JIT_STATE_ENV).is_ok() {
+        eprintln!("jit-state: {}", state);
+    } else {
+        panic!("ooo");
```
vext01, line 26: debug code left in!

vext01: Fixed in 08bce77

vext01:

.buildbot.sh
ptersilie: Ditto. It might help breaking this up into multiple sentences.

ptersilie:
Looks good to me (apart from one comment).

.buildbot.sh
vext01: Does d0293028143f4a95bdeb783cf67434486f9b9896 help?

ptersilie:
Works for me.

vext01:
You happy to take over in Laurie's absence? OK to squash?

ptersilie:
Yes, please squash.

vext01:
splat.

vext01:
Had to merge upstream to repro the error.

OK to squash?

ltratt:
Please squash and rebase.

vext01:
splat

PR #509, opened 2022-02-26
Rename yk_set_hot_threshold to yk_hot_threshold_set.

ltratt:
This reflects the ordering of other functions in the C API.

Perhaps we should also rename `set_hot_threshold` to `hot_threshold_set` in the Rust API too?

### DIFF
--- a/tests/tests/call_ext_in_obj.c
+++ b/tests/tests/call_ext_in_obj.c
@@ -27,7 +27,7 @@ extern int call_me_add(int);
 
 int main(int argc, char **argv) {
   YkMT *mt = yk_mt_new();
-  yk_set_hot_threshold(mt, 0);
+  yk_hot_threshold_set(mt, 0);
   YkLocation loc = yk_location_new();
 
   int i = 3;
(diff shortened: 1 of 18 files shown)


vext01:
I find `hot_threshold_set` less familiar. Is there a precedent elsewhere, or?

ltratt:
All the functions in https://github.com/ykjit/yk/blob/master/ykcapi/yk.h and https://github.com/ykjit/yk/blob/master/ykcapi/yk_testing.h follow the "bigger -> smaller" function naming pattern AFAICS.

vext01:
Ah. That's because in C there are no objects / structs in which to namespace things.

By that convention, your function would be something like `yk_mt_set_threshold()`.

ltratt:
Yes, I did wonder about the "mt" insertion, but that requires changing other functions too. Maybe in another PR?

vext01:
So we just close this, as `set_threshold` is the right order?

ltratt:
The precedence is `(set) (hot_threshold)` so I claim the method should be `(hot_threshold)_(set)` (if that makes sense).

vext01:
My apologies. You are right. I thought we had `drop_location` etc, but we actually have `location_drop`.

I don't have much of an opinion as long as it's consistent.

We can merge this one as long as we do the `mt` thing soon.

Plan?

ltratt:
Agreed!

PR #514, opened 2022-03-02
Fix silent failure of clang-format on the C test suite.

vext01:
The problem is twofold:

 - The clang-format config file for the C suite is invalid and causes
   clang-format to exit with error status.

 - The xtask program running clang-format doesn't check the exit status,
   so we have been failing silently.

This change fixes those, and fixes bad formatting that slipped through
as a result of the bugs.


ptersilie:
Looks good. Was the config always invalid or has it changed in an update of `clang`?

vext01:
I don't think it has ever worked. Probably my fault!

Issue #426, opened 2021-10-22
Investigate what happens if our C++ code unwinds over the Rust FFI.


vext01:
Probably undefined.

PR #516, opened 2022-03-03
Put back ignored tests, round 4.

vext01:
This puts back a multi-threading test and a struct test.

multi-threading requires a compiler change: https://github.com/ykjit/ykllvm/pull/26

Please check the multi-threading test carefully. I *think* I've put the pieces together correctly.


tests/tests/many_threads.c
```diff
@@ -1,51 +1,83 @@
...
-    __yktrace_start_tracing(HW_TRACING);
-    int res = 1 + 1;
-    void *tr = __yktrace_stop_tracing();
-    assert(res == 2);
-
-    void *ptr = __yktrace_irtrace_compile(tr);
-    __yktrace_drop_irtrace(tr);
-    void (*func)() = (void (*)())ptr;
-    func();
+#define ITERS 1000
```
ltratt, line 32: We probably need to set this much higher, otherwise it's quite possible that a thread will complete before any others have started. I'd probably set it to 100,000.

tests/tests/many_threads.c
```diff
@@ -1,51 +1,83 @@
...
+};
+
+static void *trace(void *arg) {
+  struct thread_data *td = (struct thread_data *)arg;
+
+  uintptr_t i = ITERS;
+  NOOPT_VAL(i);
+  while (i > 0) {
+    yk_mt_control_point(td->mt, td->loc);
+    fprintf(stderr, "thread #%d: %lu\n", td->tnum, i);
```
ltratt, line 47: We can't do this on every iteration as it will synchronise the threads. Ideally we wouldn't print out anything during the main loop at all: if we need to print something we should only print every *n* iterations (where *n* would be quite a high number).

ltratt: In fact, since we don't use this output in the test at all, can't we just get rid of this `printf` entirely?

tests/tests/many_threads.c
vext01: FWIW, in my testing I didn't see any interleaving with 1000 iters, but did at 100,000. I didn't up it because a) it makes the test slow, and b) we can't test interleaving anyway.

tests/tests/many_threads.c
vext01: Why would it synchronise threads? I saw interleaving in my testing.

vext01: e.g.:
```
jit-state: enter-jit-code
thread #24: 91591
jit-state: enter-jit-code
thread #9: 91338
jit-state: exit-jit-code
jit-state: enter-jit-code
jit-state: thread #9: 91337
exit-jit-code
jit-state: enter-jit-code
thread #15: 91362
jit-state: exit-jit-code
jit-state: enter-jit-code
jit-state: enter-jit-code
thread #16: 91499
jit-state: exit-jit-code
thread #26: 91629
jit-state: exit-jit-code
jit-state: exit-jit-code
jit-state: exit-jit-code
jit-state: exit-jit-code
jit-state: exit-jit-code
jit-state: exit-jit-code
jit-state: exit-jit-code
jit-state: exit-jit-code
jit-state: enter-jit-code
jit-state: thread #25: 91467
enter-jit-code
thread #13: 91326
jit-state: exit-jit-code
```

tests/tests/many_threads.c
vext01: seconds, yes.

tests/tests/many_threads.c
ltratt: It's hard to say anything with certainty, but `printf` and friends have to gain a lock; that means at most one thread will `print` at any point, and threads will probably put themselves to sleep on most iterations of the loop. This makes the test a lot less nondeterministic than we would like.

tests/tests/many_threads.c
ltratt: Seconds is fine -- and I bet you if you remove the `printf` you're down to tiny fractions of a second.

tests/tests/many_threads.c
vext01: Perhaps only one thread can printf at a time, but the log above clearly shows that the rest of the trace code is running with all kinds of interleaving.

ltratt: There's no "perhaps" about it ;) You're interpreting what I say as "the `printf` means there is no interleaving" but that is not what I'm saying: what I'm saying is that the `printf` reduces nondeterminism (note: "reduces" not "precludes"!). [Since we don't even use the `printf` output in the lang_tester test, the `printf` also makes the computer get hot without allowing us to test anything useful!]

It would be a better test for us if instead of a `printf` we did some sort of normal calculation (preferably one that the optimiser won't remove entirely) that doesn't do locking.

vext01: >  some sort of normal calculation

We are already checking that teach thread decrements down to zero, and a `NOOPT_VAL` ensures the loop can't disappear.

I'll just kill the printf.

I find this test unsatisfying, but I don't think it's possible to write a satisfying one.

ltratt: Here's an example of something we could do: we can pass a number to each thread based on how many we've created already, and do some sort of calculation on that in the loop. Then we can check that against the output at the end of the thread.

ltratt: Admittedly I'm not sure what we can actually execute in the stopgap interpreter at the moment. Maybe `i--`  is as far as we should push our luck?

vext01: True. We can still write the proper test now, and once Lukas merges his work, we will be testing the whole program as intended.

ltratt: Not sure I understand: are you saying we should to the test now or wait until the stopgap interpreter is merged?

vext01: Every test in the C suite currently can only test up to the point of stop-gapping, where upon the program exits (with success status for now, otherwise we couldn't test anything).

This threading test, like the others, should test up until we stop-gap, but we can still put assertions in for when we can stop-gap.

Agreed?

ltratt: I still don't know if you're suggesting adding more stuff in or not :)

vext01: In fact, it looks like some threads do manage to exit before the compiled trace is available, so we can actually check the result of some computation for those threads now without having stop-gapping.

vext01: > I still don't know if you're suggesting adding more stuff in or not :)

probably best to wait for the diff. Bear with me because something is crashing.

ltratt:
@vext01 Ah, I've just realised one thing. `many_threads` has *n* threads sharing 1 location; we should also, I think, try `*n* threads sharing *n* locations, to see what happens if we try starting PT in multiple threads simultaneously. I don't know if PT can only run in one thread at a time or not, so this may well bork, at which point we would `ignore` the test until we can fix things elsewhere I guess.

vext01:
We can get another test out of that scenario. Bear with me.

ltratt:
Agreed, this would be a second test.

vext01:
[8328c0e](https://github.com/ykjit/yk/pull/516/commits/8328c0ea1df725919a65a5f25d8b9b381eb16e6f):

 - removes the printf.
 - makes the checked computation for each thread unique to the thread
 - fixes a bogus cast that was causing a seg-fault.

If you are OK with that, I can devise the second test (based on the first).

ltratt:
Looks good to me!

vext01:
An amusing outcome (I'm hesitant to call it a bug) is that if we have `YKD_PRINT_IR` set and have a load of threads all compiling at once, then they all print their modules together and the output all interleaves, making matching IR impossible!

If we wanted to fix that, the the printing code would need synchronisation, but in turn that would (as you put it) "limit non-determinism" in the tests.

Then again, users will be using these IR dumps for optimising their intepreters (like an rpython users uses jitviewer to try an get a short trace). It wouldn't be user friendly to give them interleaved output.

Any thoughts on this?

vext01:
Additional context: I removed `YKD_SERIALISE_COMPILATION` for the second new test because otherwise it makes the test very slow indeed.

ltratt:
I think it would be fine for the printing routines to have a global lock (but only during printing!).

vext01:
OK, I think I'll do that in a separate PR and have a weaker test in the interim.

I also want to make it so that we print the AOT module at most once, since it doesn't change. Currently each trace compilation prints it!

ltratt:
OK.

vext01:
That last push implements the new test and adds a few fixes to the existing one.

But the new test is shaking out an illusive non-deterministic bug, which results in a crash like this:
```
many_threads_many_locs: /home/vext01/research/ykllvm/llvm/lib/ExecutionEngine/Orc/ThreadSafeModule.cpp:21: llvm::orc::ThreadSafeModule llvm::orc::cloneToNewContext(const llvm::orc::ThreadSafeModu
le&, llvm::orc::GVPredicate, llvm::orc::GVModifier): Assertion `TSM && "Can not clone null module"' failed.
```

I think this will take a while to investigate, and the compiler change that we merged earlier means that `master` is busted until we merge one of the (unrelated to threading) changes from this branch.

I'm going to extract that into a fresh PR so as to unbreak `master`, the debug the crash.

ltratt:
Please squash.

vext01:
We can't merge this until I figure out what is causing the crash.

ltratt:
Sorry, somehow I missed the bug report!

A working guess is that we might be causing LLVM problems by performing more than one compilation job in parallel?

vext01:
CC @ptersilie

This is complicated. I'll try my best to explain.

The `Can not clone null module` crash above is in fact *not* a concurrency bug, really.

Currently when a thread wants to stop-gap, well we don't have an interpreter, so at the moment it simply calls `process::exit()` to abort the running program (we exit with success so that we can still test something up until this point).

But `process::exit()` is wrong. When we do this Rust and C++ start running their destructors and tearing down the execution environment. Meanwhile, other interpreter threads are happily executing, oblivious to the world disintegrating around them. Eventually the `GlobalAOTMod` is destructed, and a null pointer is left in its wake, and boom.

The obvious solution would be to use `pthread_exit()`, allowing the test to `pthread_join()` the thread (which it already expects to do). This way the one thread exits, without "rug pulling" the other threads.

But this itself is problematic, as for some reason Rust tries to unwind. Unwinding through a mix of C, Rust, asm, and (especially) `__llvm_optimise()` (which expects us to do weird things to the stack) sounds at best, very very difficult.

I quickly experimented with the `C-unwind` ABI and using the `abort` strategy for only `libykcapi`, but no cigar.

On the one hand, given that in the not-so-long-term, we won't need the thread to exit at all (the thread will stop-gap properly), it may not be worth our time to allow the thread exit properly.

On the other hand, what happens if the user's program, that is being stop-gapped, calls `pthread_exit()` itself? We would probably have to intercept the call and do some magic, otherwise we'd be in the same sinking ship as before.

ltratt:
OK, so it sounds like one reasonable possibility is that we say "we've put back some of these tests a bit too early and we need to wait for a more feature complete stopgap interpreter before putting them back"?

vext01:
That was how I started my above response, but I got cold feet in light of:

> what happens if the user's program, that is being stop-gapped, calls pthread_exit() itself? We would probably have to intercept the call and do some magic, otherwise we'd be in the same sinking ship as before.

We could cross that bridge when we come to it.. I'm not sure what's best.

ltratt:
Somewhere we should start some sort of "gotchas" or "wedontknowwhatodointhesecases" file. Probably as part of ykdocs?

vext01:
I've been raising issues for this kind of thing, and I now actively review our issue list. Should be sufficient?

vext01:
Related #426

ltratt:
IMHO, issues are for our internal usage, but we might want to document for interpreter authors something like "`pthread_exit` might cause you problems".

vext01:
Ah I see. This list would be for consumption by interpreter authors. Happy to make one.

ltratt:
Thanks!

Do you also want to update (shrink) this PR to just the tests that we can get to work at the moment, and hold back the other tests for when the stopgap interpreter is in better shape?

vext01:
I was thinking of including the changes to the threading tests, as they currently are for the pre-big-rewrite API. We can at-least get them closer to what they will be once we enable them.

Would you be OK with that?

ltratt:
So they'd be updated but still `ignored`? Fine by me.

vext01:
> So they'd be updated but still ignored?

Correct. Will do this now.

vext01:
If you are happy with 7dfd6d2, then this needs rebasing and squashing into shape.

ltratt:
Please rebase + squash.

vext01:
splat.

ltratt:
"Reinstate the many_threads.c and derive one further test." doesn't quite parse as a commit title. Please force push an update.

vext01:
Try that

Issue #519, opened 2022-03-09
Deal with nested loops properly


vext01:
For the situation where there are nested loops in the interpreter. The inner loop will be JIT compiled first, then later the outer loop will eventually be compiled too. It would be best if the outer loop called the existing trace for the inner loop, rather than duplicating JITted code for the inner loop.

ltratt:
> The inner loop will be JIT compiled first, then later the outer loop will eventually be compiled too.

Not necessarily: either loop can be compiled in either order.

We should probably check in more detail what RPython does for nested loops, as I don't remember the whole story. For example, one thing we *could* do is say that if you have loops where the outer loop becomes hot before the inner loop, it's fine to unroll in the inner loop (this would be the current behaviour).

vext01:
> Not necessarily: either loop can be compiled in either order.

Due to non-atomicity of hot counts?

ltratt:
This is a totally normal property of loops: nothing to do with JITs. [One random example of an interesting case is if the inner loop is behind an `if`.]

vext01:
Good point! I have no idea why I didn't consider this case!

ltratt:
Closing this because it's not very specific about which of the possible problems with loops it might be talking about.

PR #520, opened 2022-03-09
Fix a bug where we may trace execution of a trace.

vext01:
In doing so, make the interp_loop tests more realistic (don't have
locations for bytecode positions where a loop can't possibly start).

Note that this gives us correctness, but is not really what we want in
the long run.

The situation we are talking about here is essentially where there are
nested loops in the interpreter. The inner loop will be JIT compiled
first, then later the outer loop will eventually be compiled too. It
would be best if the outer loop called the existing trace for the inner
loop, rather than duplicating JITted code for the inner loop.

We will attempt to implement the correct behaviour later.


vext01:
The commit message probably needs revision in light of discussion in #519

ltratt:
I should probably a co-author of this horror, so @ptersilie will have to be the reviewer.

vext01:
I'll add `co-authored-by` at the end.

ptersilie:
Looks good. I'll merge once you've done the co-authored bit.

vext01:
If I'm going to force push, i may as well sync the branch too.

vext01:
added co-author tag, and synced branch.

PR #521, opened 2022-03-09
Our JITted traces use the C ABI and are unsafe.

vext01:
I noticed that we are erroneously calling our traces with the Rust ABI.
It's working by chance.

[There's a bit of churn in the diff, but essentially I just added `unsafe extern "C"` to the place where we cast the pointer to the code to a callable function]

[See [here](https://github.com/ykjit/yk/blob/7d4e6d084ae5daa25ceeddee108634e7f9ea56d2/ykllvmwrap/src/jitmodbuilder.cc#L391) for proof that our trace code is C ABI]

### DIFF
--- a/yktrace/src/lib.rs
+++ b/yktrace/src/lib.rs
@@ -234,8 +234,12 @@ impl CompiledTrace {
     pub fn exec(&self, ctrlp_vars: *mut c_void) {
         #[cfg(feature = "yk_testing")]
         assert_ne!(self.entry as *const (), std::ptr::null());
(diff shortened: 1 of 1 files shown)


vext01:
gah `s/API/ABI/` in the commit message!

ltratt:
I'd probably suggest:

```
unsafe {
  let f = ...;
  f();
}
```

as being clearer. What do you think?

vext01:
Agreed. Fixed in [9715a83](https://github.com/ykjit/yk/pull/521/commits/9715a832431093ed15a8d6783dc73a30f976afd6)

ltratt:
Please squash and rebase.

vext01:
splat.

PR #528, opened 2022-03-15
Deal with lifetime intrinsics

vext01:
With a compiler change, this also allows us to reinstate the `recursion.c` test.

See commits for more info.

PR for compiler change ~coming soon~ [here](https://github.com/ykjit/ykllvm/pull/27)


ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -979,8 +979,15 @@ class JITModBuilder {
           continue;
 
         if (isa<CallInst>(I)) {
-
           if (isa<IntrinsicInst>(I)) {
+            // llvm.lifetime.start.* and llvm.lifetime.end.* have no meaninful
+            // program semantics for our purposes, and generate no code anyway.
```
ltratt, line 984: "For our purposes"?

vext01: Calls to these functions are only used by LLVM's stack slot colouring pass.

ltratt: I plead ignorance: are these annotations not needed at all, or not necessary by our stage in the pipeline? And how do these affect recursion?

vext01: Sorry, let me elaborate.

Calls to `llvm.lifetime.*` are annotations added by upstream LLVM at some stage in their AOT pipeline for a later optimisation pass (of theirs) to pick up. For whatever reason, they leave the annotations in, and thus we see them in our AOT module.

Currently, when the trace compiler sees a call to an intrinsic, it needs to know a) was the call inlined? and if not, b) what symbol does it need to call then.

For `llvm.lifetime.*` there is no symbol that it could call and there is no code that could have been inlined, as these calls are not really calls. They generate no code and have no program semantics.

So all I'm doing here is teaching the trace compiler to ignore them.

I hope that makes sense?

ltratt: Should we filter these out at the AOT stage?

ltratt:
Would I be correct in saying that this PR contains two different things? https://github.com/ykjit/yk/pull/528/commits/a99164a24c571ccd4d4c5422dbae6e942e79dba9 (and https://github.com/ykjit/yk/pull/528/commits/a0f032ae727e604b059581f9032184cc1d0605df?) are to do with intrinsics. The other commits are to do with recursion. I don't think those two sets of commits are related?

ykllvmwrap/src/jitmodbuilder.cc
vext01: We could but:

 - I don't know if there are runtime analyses in the wild that rely on these annotations in embedded IR.
 - I don't know if there are LLVM plugin passes (external to the LLVM codebase) that rely on these annotations.
 - It's harder for us to do, as we'd have to add a whole pass.

ltratt: Fair enough. Can we at least clarify more thoroughly in this PR *why* we're excluding these? [As an aside: presumably we expect LLVM to add these back in while compiling the trace if it needs to?]

vext01: > presumably we expect LLVM to add these back in while compiling the trace if it needs to?

That's my understanding.

vext01:
Yeah, it's a bit annoying. The recursion test didn't work because it caused lifetime annotations to be emitted and that crashed the trace compiler. But if you fix that, then you hit bugs to do with self recursion.

Sorry if it wasn't clear.

ykllvmwrap/src/jitmodbuilder.cc
ltratt: @bjorn3 Is this only used by rustc (and not any other front end)?

bjorn3: It is valid to omit them entirely, but doing so will increase stack usage.

ltratt: OK, if we update the comment, I think we'll be good to move on.

vext01: Try [0264dc7](https://github.com/ykjit/yk/pull/528/commits/0264dc7b9be43d738c798ffa8c1c86d4c99ae8d4)

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -980,10 +980,16 @@ class JITModBuilder {
...
-            // `llvm.lifetime.start.*` and `llvm.lifetime.end.*` are not really
-            // calls, but rather left-in annotations used in LLVM optimisation
-            // passes. Since these "calls" generate no machine code, nor have
-            // program semantics, we can simply ignore them.
+            // `llvm.lifetime.start.*` and `llvm.lifetime.end.*` are
+            // annotations added by some compiler front-ends to allow backends
+            // to perform stack slot optimisations.
+            //
+            // They are not really calls, and since they generate no machine
+            // code, nor have program semantics, we can simply ignore them.
```
ltratt, line 988: On reflection I'm not keen on "nor have program semantics" which is easily misread. How about something like "Removing these annotations makes generated code slightly less efficient, but does not affect correctness, so we remove them to make our lives easier." ?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -980,10 +980,16 @@ class JITModBuilder {
...
+            //
+            // OPT: Consider leaving the annotations in, or generating our own
```
ptersilie, line 990: There may just be a pass we can run over our trace to generate them. Could worth having a look at some point.

vext01: Yeah, I think generating our own could give more tight bounds. It's an optimisation, so low priority for now.

ltratt: Agreed.

ykllvmwrap/src/jitmodbuilder.cc
vext01: Fixed in [5e85507](https://github.com/ykjit/yk/pull/528/commits/5e8550799cdb350b7cf62adca9f0e89cac3a6012)

vext01:
Ready for squashing?

ltratt:
Please squash.

vext01:
Mind if i rename test `self_recursion.c` to `direct_recursion.c` at the same time?

Just realised the filename isn't accurate.

vext01:
splat

PR #529, opened 2022-03-17
Add more recursion tests

vext01:
This adds 3 more recursion tests.

### DIFF
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -928,7 +928,7 @@ public:
     std::vector<void *> NewFAddrVals;
     for (Function &F : AOTMod->functions()) {
       NewFAddrKeys.push_back(const_cast<char *>(F.getName().data()));
-      NewFAddrVals.push_back((void *)0x1);
+      NewFAddrVals.push_back((void *)YK_INVALID_FUNC_VADDR);
     }
 
     JITModBuilder JB(AOTMod, FuncNames, BBs, TraceLen, &NewFAddrKeys[0],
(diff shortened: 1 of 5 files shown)


ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -918,8 +918,21 @@ class JITModBuilder {
...
+    // functions in the AOT module, so that the trace compiler can outline
+    // calls to them if neccessary.
+    //
+    // The actual addresses inserted don't matter, as the trace compiler suite
+    // only compiles traces (without executing them).
+    std::vector<char *> NewFAddrKeys;
+    std::vector<void *> NewFAddrVals;
+    for (Function &F : AOTMod->functions()) {
+      NewFAddrKeys.push_back(const_cast<char *>(F.getName().data()));
+      NewFAddrVals.push_back((void *)0xdeadbeef);
```
ltratt, line 931: I would tend to prefer `0` here (since if anyone dereferences that, it's more likely to segfault than 0xdeadbeef which is a valid virtual address location).

vext01: Sadly we can't do this. The entries we are adding to this map eventually get passed to `ExecutionEngine::addGlobalMapping()` and (although this isn't documented)  if you pass `NULL` pointers to it, then this has special meaning to LLVM: that the symbol address is unknown.

If we try to point symbols to NULL, LLVM will halt compilation and return an error message like:
```
Symbol not found: f
```

So we have to use a non-null value. Would `0x1` be preferable perhaps?

ltratt: `0x1` sounds like a reasonable compromise.

ltratt:
Should we make the `0x1` a constant `IMPOSSIBLE_DOO_DAH` (maybe you can think of a better name :)) so we don't later wonder why `0x1`? That way we can also add a comment along the lines of "ideally this would be 0, but that has other semantics for thingy XYZ"?

vext01:
[5e3e1a8](https://github.com/ykjit/yk/pull/529/commits/5e3e1a806a6d30685aec34fc64fa3c6e14b94361) does that.

OK to squash?

ltratt:
Please squash.

vext01:
Splat.

It was only necessary to squash one commit. The rest stand alone nicely.

PR #530, opened 2022-03-18
Better errors for the trace compiler suite.

vext01:
Save our future selves some time by giving good error messages.

### DIFF
--- a/tests/src/bin/run_trace_compiler_test.rs
+++ b/tests/src/bin/run_trace_compiler_test.rs
@@ -7,14 +7,23 @@ use memmap2;
 use std::{collections::HashMap, env, ffi::CString, fs::File};
 use yktrace::{IRBlock, IRTrace};
 
-fn main() {
+const BBS_ENV: &str = "YKT_TRACE_BBS";
+
+fn main() -> Result<(), String> {
     // Build the trace that we are going to have compiled.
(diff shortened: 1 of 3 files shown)


tests/src/bin/run_trace_compiler_test.rs
```diff
@@ -7,14 +7,23 @@ use memmap2;
...
+
+fn main() -> Result<(), String> {
     // Build the trace that we are going to have compiled.
     let mut bbs = vec![IRBlock::unmappable()];
-    for bb in env::var("YKT_TRACE_BBS").unwrap().split(",") {
-        let mut elems = bb.split(":");
-        let func = elems.next().unwrap();
-        let bb_idx = elems.next().unwrap().parse::<usize>().unwrap();
-        bbs.push(IRBlock::new(CString::new(func).unwrap(), bb_idx));
+    if let Ok(tbbs) = env::var(BBS_ENV) {
```
ltratt, line 25: `if let Ok(...) = ... { } else { ... }` might look better / more idiomatic as a `match`?

vext01: I don't think it adds any value TBH.

ltratt: OK.

PR #533, opened 2022-03-29
Add an assertion and fix a failure of the assertion. 

vext01:
See commit messages for more info.

Companion to https://github.com/ykjit/ykllvm/pull/29

### DIFF
--- a/tests/langtest_c.rs
+++ b/tests/langtest_c.rs
@@ -119,6 +119,7 @@ fn mk_compiler(exe: &Path, src: &Path, opt: &str, extra_objs: &[PathBuf]) -> Com
         "-Wl,--mllvm=--disable-early-taildup", // Interferes with the BlockDisambiguate pass.
         "-Wl,--mllvm=--disable-tail-duplicate", // ^^^
         "-Wl,--mllvm=--yk-disable-tail-call-codegen", // Interferes with the JIT's inlining stack.
+        "-Wl,--mllvm=--yk-no-fallthrough",     // Fallthrough optimisations distort block mapping.
         // Ensure control point is patched.
         "-Wl,--mllvm=--yk-patch-control-point",
         // Ensure we can unambiguously map back to LLVM IR blocks.
--- a/tests/langtest_c.rs
+++ b/tests/langtest_c.rs
@@ -111,9 +111,12 @@ fn mk_compiler(exe: &Path, src: &Path, opt: &str, extra_objs: &[PathBuf]) -> Com
         // Disable machine passes that would interfere with block mapping.
         //
         // If you are trying to figure out which pass is breaking the mapping, you can add
-        // "-Wl,--mllvm=--print-after-all" to see the MIR after each pass. When you have foud the
+        // "-Wl,--mllvm=--print-before-all"/"-Wl,--mllvm=--print-after-all" to see the MIR
+        // before/after each pass. You can make the output smaller by filtering the output by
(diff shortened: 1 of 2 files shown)


ltratt:
Should there be a test for this?

vext01:
I should have mentioned. The new assertion on the yk side was failing with existing tests.

ltratt:
Aha, OK, let's get that into a commit message during squashing. Otherwise OK to me.

vext01:
Agreed. I'd also like @ptersilie to give this the once-over.

tests/langtest_c.rs
```diff
@@ -119,6 +119,7 @@ fn mk_compiler(exe: &Path, src: &Path, opt: &str, extra_objs: &[PathBuf]) -> Com
         "-Wl,--mllvm=--disable-early-taildup", // Interferes with the BlockDisambiguate pass.
         "-Wl,--mllvm=--disable-tail-duplicate", // ^^^
         "-Wl,--mllvm=--yk-disable-tail-call-codegen", // Interferes with the JIT's inlining stack.
+        "-Wl,--mllvm=--yk-no-fallthrough",     // Distorts the block mapping.
```
ptersilie, line 125: This flag doesn't distort the block mapping? Not using it distorts the mapping. So it un-distorts the block mapping.

ptersilie:
This looks like this would also be solved with the "make function calls terminate blocks" pass we did once but never merged. I wonder if it's worth revisiting that at some point.

Apart from a "the the" typo in the first commit, looks good.

ltratt:
> This looks like this would also be solved with the "make function calls terminate blocks" pass

Is that a more general / better solution?

ptersilie:
I'm not sure. That's something we'd have to work out. Might be worth discussing with @vext01 offline at some point.

vext01:
> This looks like this would also be solved with the "make function calls terminate blocks" pass we did once but never merged.

Do you mean make the `CallInstr` a terminator, or split each block after a call with an unconditional branch?

The former is a large change for the LLVM language. The latter doesn't address the underlying issue: that we may or may not resume from a call in the same block or another block.

ptersilie:
I meant the latter, but I think I see now that this doesn't solve the issue.

vext01:
> I meant the latter, but I think I see now that this doesn't solve the issue.

The problem case is as follows:
```
bbN:
  call f()
  br bbM

bbM:
  ...
```

And we may or may not resume in `BBM` after the call to `f`.

The call in `BBN` is already the last non-terminating instruction, so presumably the pass you are proposing would not change the block at all.

ptersilie:
Yeah, I can't argue with that.

vext01:
OK to squash?

ptersilie:
Please do.

ptersilie:
Oh wait, there's still one comment.

tests/langtest_c.rs
vext01: fixed in [d142b30](https://github.com/ykjit/yk/pull/533/commits/d142b30250c2aa94dc7b203adc200c20630a1557)

ptersilie:
Please squash.

vext01:
splat.

vext01:
bors r-

My bad. I was supposed to update the commit message.

vext01:
In the first commit, the paragraph starting "For now we first add an assertion..." was reworded as per @ltratt's request.

If it looks good, please re-run bors @ptersilie.

PR #539, opened 2022-04-11
Implement missing stuff required to run ykcbf.

vext01:
And put back the `bf.c` test derived from ykcbf.

This is just a collection of changes implementing missing stop-gap
interpreter features.

### DIFF
--- a/yksgi/src/lib.rs
+++ b/yksgi/src/lib.rs
@@ -185,6 +185,7 @@ impl SGInterp {
                 LLVMOpcode::LLVMICmp => self.icmp(),
                 LLVMOpcode::LLVMLoad => self.load(),
                 LLVMOpcode::LLVMPHI => self.phi(),
+                LLVMOpcode::LLVMPtrToInt => self.ptrtoint(),
                 LLVMOpcode::LLVMRet => {
                     if self.ret() {
                         // We've interpreted the return of the control point's caller, so we need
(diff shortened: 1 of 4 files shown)


yksgi/src/lib.rs
```diff
@@ -199,12 +200,36 @@ impl SGInterp {
...
+        let op_true = self.pc.get_operand(1);
+        let op_false = self.pc.get_operand(2);
+
+        if op_true.get_type().is_vector() || op_false.get_type().is_vector() {
+            todo!(
+                "vector switches are not implemented: {:?}",
+                self.pc.as_str()
+            );
+        }
+        assert!(cond.get_type().get_int_width() == 1); // Only `i1` is valid.
```
ptersilie, line 222: I think we should make this a `debug_assert` as this should typically have already been caught by the IR verifier.

yksgi/src/lib.rs
```diff
@@ -199,12 +200,36 @@ impl SGInterp {
...
         }
     }
 
+    fn select(&mut self) {
+        let cond = self.pc.get_operand(0);
+        let op_true = self.pc.get_operand(1);
+        let op_false = self.pc.get_operand(2);
+
+        if op_true.get_type().is_vector() || op_false.get_type().is_vector() {
+            todo!(
```
ptersilie, line 217: Is @ltratt okay with strings in todo's these days? I've been trying not to use them (even though one in `interpet()` would be useful) to avoid his wrath.

ltratt: You're making me out to be a monster -- just because I'm big, green, and hairy doesn't mean I'm a monster ;)

I've always been fine with strings in `todo`s because they still cause the program to terminate (they just give you more information about *why* the program terminated).

yksgi/src/lib.rs
```diff
@@ -395,6 +431,14 @@ impl SGInterp {
         }
     }
 
+    fn ptrtoint(&mut self) {
+        let src = self.pc.get_operand(0);
+        let srcty = src.get_type();
+        assert!(srcty.is_pointer());
+        let srcval = self.var_lookup(&src).val;
+        self.var_set(self.pc, SGValue::new(srcval, self.pc.get_type()));
```
ptersilie, line 439: You can probably use `SGValue::with_type` here, which creates a copy with a new type.
```
let srcval = self.var_lookup(&src);
self.var_set(self.pc, SGValue::with_type(self.pc.get_type()));
```

That being said, this will all go away when we get rid of the type within `SGValue`.

ptersilie:
Just some minor comments.

yksgi/src/lib.rs
vext01: Oops yes. That was my intention. I've been writing too much C++.

vext01: Fixed in 8ba0b9c

vext01:
OK to squash?

yksgi/src/lib.rs
ltratt: @ptersilie Binary size? You have an active imagination :p

ptersilie:
Oh, this still needs a rebase.

ltratt:
Ouch, yes, this has conflicts, so I can't even realistically hope that bors/buildbot will manage the merge automatically. @vext01 Please rebase and force push.

vext01:
rebased.

PR #540, opened 2022-04-13
Make the trace compiler actually optimise.

vext01:
Due to incorrect usage of the pass manager, our trace compiler has not
been optimising traces!

This commit does 3 things.

First it makes the trace compiler optimise at -O2, as was intented from
the outset.

After making this change it became immediately obvious that
optimisations were occurring: the whole JIT function was removed because
we had it marked for internal linkage. So the second thing this commit
does is mark the JIT function as `ExternalLinkage`.

And finally, now that we optimise, items in the global mappings could be
optimised out. For such cases we shouldn't register global mappings for
them, or LLVM assertions will fail.


ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -288,9 +291,10 @@ void *compileIRTrace(FN Func, char *FuncNames[], size_t BBs[], size_t TraceLen,
   PassManagerBuilder Builder;
   Builder.OptLevel = 2; // FIXME Make this user-tweakable.
   legacy::FunctionPassManager FPM(JITMod);
+  legacy::PassManager MPM;
   Builder.populateFunctionPassManager(FPM);
```
ptersilie, line 295: Do we still need the function passmanager or is it superseded by the module passmanager?

ptersilie:
Just one question. Otherwise looks good.

ykllvmwrap/src/ykllvmwrap.cc
vext01: We need both.

From [the docs](https://llvm.org/doxygen/classllvm_1_1PassManagerBuilder.html#details):
```
A simple example of using it would be:

PassManagerBuilder Builder;
Builder.OptLevel = 2;
Builder.populateFunctionPassManager(FPM);
Builder.populateModulePassManager(MPM);
```

PR #545, opened 2022-04-20
Fix stopgapping out of nested functions.

ptersilie:
Until now our stopgap interpreter was only able to recover from a guard
failure if it happened at the top level of the compiled interpreter.
This commit adds the neccessary extra information to the guards so that
the interpreter can recover even when the guard failure happens deep
inside a nested (and in the JIT inlined) function.

### DIFF
--- a/tests/c/interp_inside_call.c
+++ b/tests/c/interp_inside_call.c
@@ -1,49 +1,53 @@
 // Run-time:
-//   env-var: YKD_PRINT_IR=jit-pre-opt
+//   env-var: YKD_PRINT_IR=aot,jit-pre-opt
 //   env-var: YKD_SERIALISE_COMPILATION=1
 //   env-var: YKD_PRINT_JITSTATE=1
 //   stderr:
-//     ...
 //     jit-state: start-tracing
-//     3:1
-//     jit-state: stop-tracing
-//     --- Begin jit-pre-opt ---
 //     ...
-//     --- End jit-pre-opt ---
-//     2:1
+//     jit-state: stop-tracing
+//     ...
 //     jit-state: enter-jit-code
-//     1:1
+//     ...
 //     jit-state: enter-stopgap
 //     ...
+//     jit-state: exit-stopgap
+//  stdout:
+//     ...
+//     i: 5 ret: 9
+//     ...
+//     i: 2 ret: 6
+//     i: 1 ret: 105
 
-// Check that we can handle struct field accesses.
+// Check that we can stopgap outside of nested, inlined calls.
 
-#include <assert.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <yk.h>
 #include <yk_testing.h>
 
-struct s {
-  int x;
-};
+__attribute__((noinline)) int f(int a, int b) {
+  if (a > 1) {
+    return a + b;
+  } else {
+    return a + b + 100;
+  }
+}
 
 int main(int argc, char **argv) {
   YkMT *mt = yk_mt_new();
   yk_mt_hot_threshold_set(mt, 0);
   YkLocation loc = yk_location_new();
 
-  struct s s1 = {argc};
-  int y1 = 0, i = 3;
+  int i = 5;
+  int ret = 0;
   NOOPT_VAL(i);
   while (i > 0) {
     yk_mt_control_point(mt, &loc);
-    NOOPT_VAL(s1);
-    y1 = s1.x;
-    fprintf(stderr, "%d:%d\n", i, s1.x);
+    ret = f(i, argc + 3);
+    printf("i: %d ret: %d\n", i, ret);
     i--;
   }
-  assert(y1 == 1);
 
   yk_location_drop(loc);
   yk_mt_drop(mt);
--- a/tests/c/interp_inside_call2.c
+++ b/tests/c/interp_inside_call2.c
@@ -1,32 +1,47 @@
 // Run-time:
-//   env-var: YKD_PRINT_IR=jit-pre-opt
+//   env-var: YKD_PRINT_IR=aot,jit-pre-opt
 //   env-var: YKD_SERIALISE_COMPILATION=1
 //   env-var: YKD_PRINT_JITSTATE=1
 //   stderr:
 //     jit-state: start-tracing
-//     3: 5
+//     ...
 //     jit-state: stop-tracing
-//     --- Begin jit-pre-opt ---
 //     ...
-//     %{{res}} = add nsw i32...
-//     ...
-//     --- End jit-pre-opt ---
-//     2: 5
 //     jit-state: enter-jit-code
-//     1: 5
+//     ...
 //     jit-state: enter-stopgap
 //     ...
+//     jit-state: exit-stopgap
+//  stdout:
+//     ...
+//     i: 5 ret: 12
+//     ...
+//     i: 2 ret: 9
+//     i: 1 ret: 108
 
-// Check that function calls with arguments do the right thing
+// Check that we can stopgap outside of nested, inlined calls.
 
 #include <stdio.h>
 #include <stdlib.h>
 #include <yk.h>
 #include <yk_testing.h>
 
+__attribute__((noinline)) int h(int a, int b) {
+  if (a > 1) {
+    return a + b;
+  } else {
+    return a + b + 100;
+  }
+}
+
+__attribute__((noinline)) int g(int a, int b) {
+  int c = b + 2;
+  return h(a, c);
+}
+
 __attribute__((noinline)) int f(int a, int b) {
-  int c = a + b;
-  return c;
+  int c = b + 1;
+  return g(a, c);
 }
 
 int main(int argc, char **argv) {
@@ -34,13 +49,13 @@ int main(int argc, char **argv) {
   yk_mt_hot_threshold_set(mt, 0);
   YkLocation loc = yk_location_new();
 
-  int i = 3, two = 2, three = 3;
+  int i = 5;
+  int ret = 0;
   NOOPT_VAL(i);
   while (i > 0) {
     yk_mt_control_point(mt, &loc);
-    NOOPT_VAL(two);
-    NOOPT_VAL(three);
-    fprintf(stderr, "%d: %d\n", i, f(two, three));
+    ret = f(i, argc + 3);
+    printf("i: %d ret: %d\n", i, ret);
     i--;
   }
 
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -19,7 +19,7 @@ use std::convert::{TryFrom, TryInto};
 use std::ffi::c_void;
 use std::{ptr, slice};
 use ykrt::{print_jit_state, HotThreshold, Location, MT};
-use yksgi::SGInterp;
+use yksgi::{self, SGInterp};
 use yksmp::{Location as SMLocation, StackMapParser};
 
 /// The first three locations of an LLVM stackmap record, according to the source, are CC, Flags,
@@ -124,25 +124,16 @@ struct AOTVar {
     bbidx: usize,
     instridx: usize,
     fname: *const i8,
+    sfidx: usize,
 }
 
-/// Address and length of a vector containing AOTVars. Mirrors the struct defined in
+/// Address and length of a vector. Mirrors the struct defined in
 /// ykllvmwrap/jitmodbuilder.cc.
 #[derive(Debug)]
 #[repr(C)]
-pub struct AOTMap {
+pub struct CVec {
     addr: *const c_void,
     length: usize,
-}
-
-/// Location (basic block index, instruction index, function name) in the AOTModule where the guard
-/// failure occured. Mirrors the struct defined in ykllvmwrap/jitmodbuilder.cc.
-#[derive(Debug)]
-#[repr(C)]
-pub struct CurPos {
-    bbidx: usize,
-    instridx: usize,
-    fname: *const i8,
 }
 
 /// Reads the stackmap and saved registers from the given address (i.e. the return address of the
@@ -152,28 +143,27 @@ pub struct CurPos {
 pub extern "C" fn yk_stopgap(
     sm_addr: *const c_void,
     sm_size: usize,
-    aotmap: &AOTMap,
-    curpos: &CurPos,
+    aotmap: &CVec,
+    actframes: &CVec,
     retaddr: usize,
     rsp: *const c_void,
 ) -> u8 {
     // FIXME: remove once we have a stopgap interpreter.
     #[cfg(feature = "yk_jitstate_debug")]
-    print_jit_state("stopgap");
+    print_jit_state("enter-stopgap");
 
     // Parse AOTMap.
     let aotmap = unsafe { slice::from_raw_parts(aotmap.addr as *const AOTVar, aotmap.length) };
 
+    // Parse active frames vector.
+    let activeframes = unsafe {
+        slice::from_raw_parts(actframes.addr as *const yksgi::FrameInfo, actframes.length)
+    };
+
     // Restore saved registers from the stack.
     let registers = Registers::from_ptr(rsp);
 
-    let mut sginterp = unsafe {
-        SGInterp::new(
-            curpos.bbidx,
-            curpos.instridx,
-            std::ffi::CStr::from_ptr(curpos.fname),
-        )
-    };
+    let mut sginterp = unsafe { SGInterp::new(activeframes) };
 
     // Parse the stackmap.
     let slice = unsafe { slice::from_raw_parts(sm_addr as *mut u8, sm_size) };
@@ -184,26 +174,26 @@ pub extern "C" fn yk_stopgap(
     // Skip first 3 locations as they don't relate to any of our live variables.
     for (i, l) in locs.iter().skip(SM_REC_HEADER).enumerate() {
         match l {
-            SMLocation::Register(reg, size) => {
-                // FIXME: remove once we have a stopgap interpreter.
-                let val = unsafe { registers.get(*reg) };
-                eprintln!("Register: {} ({} {})", val, reg, size);
+            SMLocation::Register(reg, _size) => {
+                let _val = unsafe { registers.get(*reg) };
+                todo!();
             }
-            SMLocation::Direct(reg, off, size) => {
+            SMLocation::Direct(reg, off, _size) => {
                 // When using `llvm.experimental.deoptimize` then direct locations should always be
                 // in relation to RBP.
                 assert_eq!(*reg, 6);
                 let addr = unsafe { registers.get(*reg) as *mut u8 };
                 let addr = unsafe { addr.offset(isize::try_from(*off).unwrap()) };
-                let v = match *size {
-                    1 => unsafe { ptr::read::<u8>(addr as *mut u8) as u64 },
-                    2 => unsafe { ptr::read::<u16>(addr as *mut u16) as u64 },
-                    4 => unsafe { ptr::read::<u32>(addr as *mut u32) as u64 },
-                    8 => unsafe { ptr::read::<u64>(addr as *mut u64) as u64 },
-                    _ => unreachable!(),
-                };
-                // FIXME: remove once we have a stopgap interpreter.
-                eprintln!("Direct: {} ({} {})", v, reg, off);
+                let aot = &aotmap[i];
+                unsafe {
+                    sginterp.var_init(
+                        aot.bbidx,
+                        aot.instridx,
+                        std::ffi::CStr::from_ptr(aot.fname),
+                        aot.sfidx,
+                        addr as u64,
+                    );
+                }
             }
             SMLocation::Indirect(reg, off, size) => {
                 let addr = unsafe { registers.get(*reg) as *mut u8 };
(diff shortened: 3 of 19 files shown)


tests/c/interp_inside_call.c
```diff
@@ -0,0 +1,54 @@
...
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_PRINT_JITSTATE=1
+//   stderr:
+//     jit-state: start-tracing
+//     ...
+//     jit-state: stop-tracing
+//     ...
+//     jit-state: enter-jit-code
+//     ...
+//     jit-state: stopgap
```
vext01, line 12: can your next PR make this `enter-stopgap`?

tests/c/interp_inside_call.c
```diff
@@ -0,0 +1,54 @@
...
+//     ...
+//  stdout:
+//     ...
+//     i: 5 ret: 9
+//     ...
+//     i: 2 ret: 6
+//     i: 1 ret: 105
+
+// Check that function calls with arguments do the right thing
```
vext01, line 21: This description needs to be updated.

tests/c/interp_inside_call2.c
```diff
@@ -0,0 +1,64 @@
...
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_PRINT_JITSTATE=1
+//   stderr:
+//     jit-state: start-tracing
+//     ...
+//     jit-state: stop-tracing
+//     ...
+//     jit-state: enter-jit-code
+//     ...
+//     jit-state: stopgap
```
vext01, line 12: I wonder if we could expose more of the stopgap interpreter's workings for testing purposes?
Something like a `YKD_PRINT_SG_*` to print stuff like: the initial frames, maybe the instructions as they are interpreted? Overkill?

I just wonder if we are really testing much here.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -199,9 +218,9 @@ class JITModBuilder {
   // The function inside which we build the IR for the trace.
   Function *JITFunc;
 
-  // Map JIT instruction to basic block index and instruction index of the
-  // corresponding AOT instruction.
-  std::map<Value *, std::tuple<size_t, size_t, Instruction *>> AOTMap;
+  // Map JIT instruction to basic block index, instruction index, function
+  // name, and stackframe of the corresponding AOT instruction.
```
vext01, line 222: s/stackframe/stack frame index/ ?

Is a lower index an older frame?

tests/c/interp_inside_call.c
ptersilie: Sure.

ptersilie: And add a `exit-stopgap` while we're there.

yksgi/src/lib.rs
```diff
@@ -109,32 +121,53 @@ impl SGInterp {
...
+    ) {
         let func = self.module.function(fname.as_ptr());
         let bb = func.bb(bbidx);
         let instr = bb.instruction(instridx);
-        let orgaot = unsafe { get_aot_original(&instr) };
+        let orgaot = if sfidx == 0 {
+            unsafe { get_aot_original(&instr) }
+        } else {
+            // Only the root stackframe has a control point call so there's no need to find
+            // "original" aot variables for the other frames.
```
vext01, line 163: I'm not sure I understand this bit.

vext01:
Nice work Lukas. I only have minor comments/queries.

tests/c/interp_inside_call2.c
ptersilie: We could, but then we'd also have to add different lang tests for different O-levels. In this tests for example, for anything >= O1, the conditional branch is moved to the root frame, so there's only one initial frame.

But no matter what the optimiser does, this test will check that we can guard fail out of a nested call, so I think that's fine.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Yes.

ykcapi/src/lib.rs
```diff
@@ -124,36 +124,27 @@ struct AOTVar {
...
+    sfidx: usize,
 }
 
-/// Address and length of a vector containing AOTVars. Mirrors the struct defined in
+/// Address and length of a vector. Mirrors the struct defined in
 /// ykllvmwrap/jitmodbuilder.cc.
 #[derive(Debug)]
 #[repr(C)]
-pub struct AOTMap {
+pub struct VecLen {
```
ltratt, line 134: This isn't an ideal struct name -- it's not the length of a vector, but an actual vector! Why not call it `AOTVars` or `AOTVarsVec` or similar?

ykcapi/src/lib.rs
```diff
@@ -124,36 +124,27 @@ struct AOTVar {
...
 /// deoptimisation call).
 #[cfg(target_arch = "x86_64")]
 #[no_mangle]
 pub extern "C" fn yk_stopgap(
     sm_addr: *const c_void,
     sm_size: usize,
-    aotmap: &AOTMap,
-    curpos: &CurPos,
+    aotmap: &VecLen,
+    actframes: &VecLen,
```
ltratt, line 147: What does `act` mean?

ltratt: Ah, "active". I get it.

ykcapi/src/lib.rs
```diff
@@ -195,15 +185,18 @@ pub extern "C" fn yk_stopgap(
...
+                        aot.bbidx,
+                        aot.instridx,
+                        std::ffi::CStr::from_ptr(aot.fname),
+                        aot.sfidx,
+                        addr as u64,
+                    );
+                }
                 // FIXME: remove once we have a stopgap interpreter.
-                eprintln!("Direct: {} ({} {})", v, reg, off);
+                eprintln!("Direct: {} ({} {}) {}", addr as u64, reg, off, size);
```
ltratt, line 199: Since we do have a stopgap interpreter, why do we still have this `eprintln`?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -121,14 +121,28 @@ class FuncAddrs {
...
   Function *Func = Before->getFunction();
   for (auto &BB : *Func) {
     if (!DT.dominates(cast<Instruction>(Before), &BB)) {
       for (auto &I : BB) {
         if ((!I.getType()->isVoidTy()) &&
             (DT.dominates(&I, cast<Instruction>(Before)))) {
+          std::tuple<size_t, size_t, Instruction *, size_t> Entry = AOTMap[&I];
+          size_t StackFrameIdx = std::get<3>(Entry);
+          if (StackFrameIdx > CallDepth) {
+            // Variables from a previously inlined function call are no longer
```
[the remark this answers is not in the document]
ltratt: If possible I'd also like to see a test which tickles this code and is currently marked "ignored" so that we know to come back and fix it.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -269,33 +288,35 @@ class JITModBuilder {
         // function calls so we know when we left the recusion.
         OutlineCallDepth += 1;
         InlinedCalls.push_back(make_tuple(CurInstrIdx, CI));
-        return;
       }
       // If this is a recursive call that has been inlined, or if the callee
       // has the "yk_outline" annotation, remove the inlined code and turn it
       // into a normal (outlined) call.
-      if ((isRecursiveCall(CF) || CF->hasFnAttribute(YK_OUTLINE_FNATTR))) {
+      else if ((isRecursiveCall(CF) || CF->hasFnAttribute(YK_OUTLINE_FNATTR))) {
```
ltratt, line 295: The inner brackets are pointless as we have `((... || ...))` which is the same as `(... || ...)`.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -520,49 +548,83 @@ class JITModBuilder {
...
+      for (size_t I = 0; I < ActiveFrames.size(); I++) {
+        std::tuple<size_t, size_t, StringRef> FrameInfo = ActiveFrames[I];
+        size_t BBIdx = std::get<0>(FrameInfo);
+        size_t InstrIdx = std::get<1>(FrameInfo);
+        StringRef Fn = std::get<2>(FrameInfo);
+
+        auto GEP =
+            FailBuilder.CreateGEP(ActiveFrameSTy, ActiveFrameVec,
+                                  {ConstantInt::get(PointerSizedIntTy, I),
+                                   ConstantInt::get(Int32Ty, 0)});
```
ltratt, line 584: Why `Int32Ty`?

yksgi/src/lib.rs
```diff
@@ -109,32 +121,53 @@ impl SGInterp {
     /// Create a new stopgap interpreter and initialise it to start interpretation at the location
     /// given by a basic block index, instruction index, and function name.
     /// FIXME: Support initialisation of multiple frames.
-    pub unsafe fn new(bbidx: usize, instridx: usize, fname: &CStr) -> SGInterp {
+    pub unsafe fn new(activeframes: &[FrameInfo]) -> SGInterp {
         // Get AOT module IR and parse it.
         let module = Module::from_bc();
+
+        // Initialise frames.
+        let mut frames = Vec::new();
```
ltratt, line 129: Maybe `Vec::with_capacity(activeframes.len())` to avoid `realloc`s?

yksgi/src/lib.rs
ltratt: I'm not sure what "original" means in this context. [Also s/aot/AOT/.]

tests/c/interp_inside_call.c
ptersilie: Fixed in d3e1bd1ea

ykcapi/src/lib.rs
ptersilie: Because I am now using it for stackframes too. So basically anytime I need to pass a vector and its length from C into Rust I need to use this struct. I'm open for suggestions. Maybe `VecPtrAndLen`?

ykcapi/src/lib.rs
ptersilie: Active. Dunno why I named it this. Can probably just rename it to `frames`.

ykcapi/src/lib.rs
ptersilie: I actually think they are still useful. I was planning to put them behind an ENV in a separate PR.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: You are correct! Must be one of @vext01 hidden tests that you missed in a previous PR. ;-)

vext01: `((...) || (...))` is idiomatic C++.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -520,49 +548,83 @@ class JITModBuilder {
...
+      StructType *VecLenStructTy =
+          StructType::get(Context, {VecPtrTy, PointerSizedIntTy});
+
+      // Create a vector of active stackframes (i.e. basic block index,
+      // instruction index, function name). This will be needed later to point
+      // the stopgap interpeter at the correct location from where to start
+      // interpretation and to setup its stackframes.
       // FIXME: Use function index instead of string name.
-      StructType *CurPosSTy = StructType::get(
+      ActiveFrames.push_back(
```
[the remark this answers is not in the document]
ptersilie: We do when we pass this to Rust. In `jitmodbuilder.cc` we could put `CurrBBIdx` and `InstrIdx` in a struct, as `FuncName` is retrieved lazily.

ykllvmwrap/src/jitmodbuilder.cc
ltratt: @vext01 That isn't what's happening here though.

ykcapi/src/lib.rs
ltratt: I think this sort of uber-low-level-never-useful-to-yk-users debugging should, if not killed entirely (my preference!), put behind a compile-time flag. There's no point us paying a run-time cost for checking for this IMHO.

ykllvmwrap/src/jitmodbuilder.cc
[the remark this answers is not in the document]
ltratt: I think a struct in C might also be good idea -- I find it really easy to mix up the types in tuples.

ykllvmwrap/src/jitmodbuilder.cc
vext01: Indeed.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Good question. This can also be `PointerSizedIntTy` I guess. In this case I know that my struct doesn't have more than i32 fields so it could even be something much smaller (e.g. i8).

ltratt: I'm not sure what "my struct ... fields" means here. What is this integer actually indexing?

tests/c/interp_inside_call.c
ptersilie: Fixed in 6eaa3e09e62

ykllvmwrap/src/jitmodbuilder.cc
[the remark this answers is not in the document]
ptersilie: Fixed in 78bf9b44f.

There already are tests that tickle this. I had to add this `if` to fix them as they would fail otherwise. There's nothing to fix here really. Once we have implemented proper liveness we can remove this whole function.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Fixed in 6d7cdbe

ykllvmwrap/src/jitmodbuilder.cc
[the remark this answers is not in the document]
ltratt: The comment makes it sound like a correctness fix though?

yksgi/src/lib.rs
ptersilie: Try 3eff2493. You might need to re-read the description of `get_aot_var` to get some context.

ykcapi/src/lib.rs
ptersilie: I guess we can kill it. Do you want it in this PR or a separate one?

@vext01 Any thoughts?

ykllvmwrap/src/jitmodbuilder.cc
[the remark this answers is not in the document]
ptersilie: I'm not sure. The current "liveness analysis" is picking up more live variables than it needs to at each guard failure. That is not necessarily incorrect and just mean we are initialising more variables in the stopgap interpreter than we have to. With this PR, however, this means that the extra variables the liveness picks up, may come from the frame of a function call that is already finished (at the time of a guard failure). Trying to add those to the stopgap interpreter thus results in an error (as the stackframe index of such variables is higher than there are active frames in the stopgap interpreter). The above code fixes this by omitting live variables whose stackframe is already finished.

In other words, the current code is not incorrect, but contains a fix that will likely not be needed any more once we have proper liveness analysis in `jitmodbuilder`.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Fixed in 614f01f4d

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: This GEP instruction makes an index into a vector of `ActiveFrameVec` structs, where the struct looks like this:
```
struct ActiveFrameVec {
    bbidx: usize,
    instridx: usize,
    fname: *const i8,
}
```
So the first `ConstantInt(PointerSizedIntTy, I)` indexes into the vector, while the second `ConstantInt::get(Int32Ty, 0)` indexes into the first field of that struct. Since this struct only has 3 fields I could change the `Int32Ty` to an `Int8Ty`.

yksgi/src/lib.rs
ltratt: I think there's something implicit in the comment (but it's worth thinking explicitly about!): when we deoptimise in a trace, we know that we only have to recreate stackframes going back to the most recent control point in the trace (note: this might not be the control point that started the trace!). I think that might explain some of the confusion @vext01 and I have?

ykcapi/src/lib.rs
vext01: I don't think there's any determinism guarantees of what we are printing, so testing them is probably not ideal.

I don't feel super-strongly.

yksgi/src/lib.rs
ptersilie: Fixed in b57d92c1d9.

ykllvmwrap/src/jitmodbuilder.cc
vext01: Lukas has gone with `(... || ...)`. I'm OK with it if you are Laurie.

ykllvmwrap/src/jitmodbuilder.cc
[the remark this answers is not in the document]
ltratt: I think I understand: can you maybe try rephrasing the comment to bring the above out more clearly?

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: I don't follow. I thought this was just about there being an extra pair of outer brackets?

ykllvmwrap/src/jitmodbuilder.cc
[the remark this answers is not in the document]
vext01: Something along the lines of:

> Our liveness analysis over-approximates the live set and includes values which cannot be mapped back... Until we have precise liveness, we filter out the extra values in an ad-hoc fashion.

?

ykcapi/src/lib.rs
ltratt: Why not call it `CVec` or similar?

ykllvmwrap/src/jitmodbuilder.cc
ltratt: Ah! Maybe this bit of code just needs a short comment to make this clear (it never hurts to comment why code is correct in this sort of case!).

ykllvmwrap/src/jitmodbuilder.cc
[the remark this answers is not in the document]
ltratt: >In this case CurrBBIdx and InstrIdx have the same type, so I'm not sure we win much putting them into a struct.

On the contrary, this is *exactly* why a struct is useful: at the moment one can easily get the two values of the same type in the wrong order, which will statically type check, but lead to weird errors at run-time. With a struct one has to explicitly name them, which makes it a bit harder to get them wrong.

I don't understand the "both variables are not always required at the same time" bit though.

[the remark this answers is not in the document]
ptersilie: `CurBBIdx` and `CurInstrIdx` are used throughout this code at different locations. They only happen to be both needed when we put things into this `ActiveFrames` vector. So what I was saying is, that I don't want to replace all their occurrences with a struct. But I can do so just for the `ActiveFrames` vector. I will do a commit and then we can see if we like it.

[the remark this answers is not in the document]
ltratt: Unless I'm missing something, whether they're stored in a struct or tuple is irrelevant. You've extracted from them the tuple and passed them around individually at some points, right? You can still do that with a struct.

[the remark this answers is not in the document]
vext01: Since there are two `size_t` members, I'd prefer a struct, so that we can't accidentally mix up field indices.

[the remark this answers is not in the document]
ptersilie: Actions speak more precisely than words, so let's see if you like this commit: c1308174

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Let's try this: 8d593502

yksgi/src/lib.rs
ptersilie: While what you said is correct, it's not what this code is about. This is about initialising the correct variables in the stopgap interpreter.

Let's say we have this (simplified) AOT code:
```
%1 = alloca
store 1, %1
...
store %1 in YKCPStruct
control_point(YKCPStruct)
%2 = load YKCPStruct // this is the same as %1
store 2, %2
```

Now in the trace we only see:
```
%2 = load YKCPStruct
store 2, %2
```

When we guard fail, the stackmap will tell us that there's a variable %2 which has a value (e.g. `0xFF123`). So when we initialise the stopgap interpreter we do `Vars[%2] = 0xFF123`. However, the stopgap interpreter may now encounter the original variable `%1`, which wasn't initialise by the stackmap (since the trace has no knowledge of it). The workaround thus is, to check if `%2` is something we passed in via the `YKCPStruct`, and if so, also assign its value to the passed in value, in this case `%1`.

ykcapi/src/lib.rs
ptersilie: Fixed in b83cf7c2

yksgi/src/lib.rs
ltratt: What happens if `%1` and `%2` don't have the same value?

ltratt:
I am now mostly happy, but there are several comments for @vext01.

yksgi/src/lib.rs
ptersilie: They have to. Conceptually they are the same variable. It's just that the trace/stackmap doesn't know that as all it sees is the values passed in via the `control_point` call.

vext01: Yes, it's the value of the same high-level variable before/after the CP.

vext01:
Just the FIXME to do then.

vext01:
> There is one already

Ah, I see the confusion.

I'm not (immediately) worried that we used string IDs for functions.

What I am worried about is storing a `StringRef`, as by doing so we have to think about the lifetimes of things.

That said, what you we really do?

vext01:
If you think that there's no avoiding storing the `StringRef`, please squash.

ptersilie:
There are two issues:
1) Using `StringRef` might be problematic
2) Storing strings in the AOTMap, is wasteful

Both will be solved by giving function names an index and use that to reference them in the AOTMap. That index just needs to be stable within a module, so I'm not yet sure what to use. But I believe that's for a separate PR.

vext01:
please squash

ptersilie:
Squashed!

PR #546, opened 2022-04-20
Teach the stop-gap interpreter how to call AOT functions.

vext01:
When encountering a call instruction, we don't interpret the callee, but
instead call the AOT version of the function. This should be faster.

We do this via libffi bindings to Rust, as we need the ability to call
addresses whose interfaces not known until runtime.

### DIFF
--- a/yksgi/src/lib.rs
+++ b/yksgi/src/lib.rs
@@ -1,10 +1,14 @@
+use libc::dlsym;
+use libffi::middle::{arg as ffi_arg, Arg as FFIArg, Builder as FFIBuilder, CodePtr as FFICodePtr};
 use llvm_sys::core::*;
 use llvm_sys::target::{LLVMABISizeOfType, LLVMOffsetOfElement};
 use llvm_sys::{LLVMOpcode, LLVMTypeKind};
+use std::alloc::{alloc, dealloc, Layout};
 use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::ffi::{c_void, CStr};
 use std::ptr;
+use std::slice;
 
 mod llvmbridge;
 use llvmbridge::{get_aot_original, BasicBlock, Module, Type, Value};
@@ -317,16 +321,13 @@ impl SGInterp {
     /// Implement call instructions. Returns `true` if the control point has been reached,
     /// `false` otherwise.
     unsafe fn call(&mut self) -> bool {
-        let func = LLVMGetCalledValue(self.pc.get());
-        if !LLVMIsAInlineAsm(func).is_null() {
+        let func = self.pc.get_called_value();
+        if func.is_inline_asm() {
             // FIXME: Implement calls to inline asm. Just skip them for now, as our tests won't run
             // otherwise.
             return false;
         }
-        let mut size = 0;
-        let name = CStr::from_ptr(LLVMGetValueName2(func, &mut size))
-            .to_str()
-            .unwrap();
+        let name = func.get_name();
         if ALWAYS_SKIP_FUNCS.contains(&name) {
             // There's no point calling these functions inside the stopgap interpreter so just skip
             // them.
@@ -364,8 +365,61 @@ impl SGInterp {
             // FIXME: When we see the control point we are done and can just return.
             return true;
         } else {
-            // FIXME: Properly implement calls.
-            todo!("{:?}", self.pc.as_str())
+            // We are going to do a native call via libffi, starting with looking up the address of
+            // the callee in the virtual address space.
+            //
+            // OPT: This could benefit from caching, and we may already know the address of the
+            // callee if we inlined it when preparing traces:
+            // https://github.com/ykjit/yk/issues/544
+            debug_assert!(func.is_function());
+            let fptr = dlsym(ptr::null_mut(), name.as_ptr() as *const i8);
+            if fptr == ptr::null_mut() {
+                todo!("couldn't find symbol: {}", name);
+            }
+
+            if func.is_vararg_function() {
+                todo!("calling a varargs function");
+            }
+
+            // Now build the calling interface and collect the values of the callee's arguments.
+            let mut builder = FFIBuilder::new();
+            let num_args = usize::try_from(self.pc.get_num_arg_operands()).unwrap();
+            let arg_lay = Layout::array::<u64>(num_args).unwrap();
+            let arg_mem = alloc(Layout::array::<u64>(num_args).unwrap());
+            let mut arg_mem_p = arg_mem.cast::<u64>();
+            for i in 0..num_args {
+                // The first N operands are the first N operands of the function being called.
+                let arg = self.pc.get_operand(u32::try_from(i).unwrap());
+                builder = builder.arg(arg.get_type().ffi_type());
+                *arg_mem_p = self.var_lookup(&arg).val;
+                arg_mem_p = arg_mem_p.add(1);
+            }
+            builder = builder.res(self.pc.get_type().ffi_type());
+            let cif = builder.into_cif(); // OPT: cache CIFs for repeated calls to same func sig.
+
+            // Actually do the call.
+            let ret_ty = self.pc.get_type();
+            let arg_slice = slice::from_raw_parts_mut::<u64>(arg_mem as *mut u64, num_args);
+            let ffi_arg_vals: Vec<FFIArg> = arg_slice.iter().map(|a| ffi_arg(a)).collect();
+            if ret_ty.is_integer() {
+                // FIXME: https://github.com/ykjit/yk/issues/536
+                match ret_ty.get_int_width() {
+                    32 => {
+                        let rv = cif.call::<u32>(FFICodePtr(fptr), &ffi_arg_vals) as u64;
+                        self.var_set(self.pc, SGValue::new(rv, ret_ty));
+                    }
+                    _ => todo!(),
+                }
+            } else if ret_ty.is_void() {
+                cif.call::<()>(FFICodePtr(fptr), &ffi_arg_vals);
+            } else {
+                todo!("{:?}", ret_ty.as_str());
+            };
+            // `ffi_arg_vals` contains raw pointers to `arg_mem`, so we had better drop those
+            // before we deallocate `arg_mem`.
+            drop(ffi_arg_vals);
+            dealloc(arg_mem, arg_lay);
+            return false;
         }
     }
 
--- a/yksgi/src/llvmbridge.rs
+++ b/yksgi/src/llvmbridge.rs
@@ -1,4 +1,5 @@
 use crate::SGValue;
+use libffi::middle::Type as FFIType;
 use llvm_sys::bit_reader::LLVMParseBitcodeInContext2;
 use llvm_sys::core::*;
 use llvm_sys::prelude::{LLVMBasicBlockRef, LLVMModuleRef, LLVMTypeRef, LLVMValueRef};
@@ -112,6 +113,10 @@ impl Type {
         matches!(self.kind(), LLVMTypeKind::LLVMIntegerTypeKind)
     }
 
+    pub fn is_void(&self) -> bool {
+        matches!(self.kind(), LLVMTypeKind::LLVMVoidTypeKind)
+    }
+
     pub fn get_int_width(&self) -> u32 {
         debug_assert!(self.is_integer());
         unsafe { LLVMGetIntTypeWidth(self.0) }
@@ -119,6 +124,24 @@ impl Type {
 
     pub fn as_str(&self) -> &CStr {
         unsafe { CStr::from_ptr(LLVMPrintTypeToString(self.0)) }
+    }
+
+    /// Converts an LLVM type into a libffi type.
+    pub fn ffi_type(&self) -> FFIType {
+        match self.kind() {
+            LLVMTypeKind::LLVMIntegerTypeKind => {
+                // FIXME: https://github.com/ykjit/yk/issues/536
+                match unsafe { LLVMGetIntTypeWidth(self.0) } {
+                    8 => FFIType::u8(),
+                    16 => FFIType::u16(),
+                    32 => FFIType::u32(),
+                    64 => FFIType::u64(),
+                    _ => todo!(),
+                }
+            }
+            LLVMTypeKind::LLVMVoidTypeKind => FFIType::void(),
+            _ => todo!("{:?}", self.as_str()),
+        }
     }
 }
 
@@ -165,6 +188,19 @@ impl Value {
         unsafe { !LLVMIsASwitchInst(self.0).is_null() }
     }
 
+    pub fn is_inline_asm(&self) -> bool {
+        !unsafe { LLVMIsAInlineAsm(self.0).is_null() }
+    }
+
+    pub fn is_function(&self) -> bool {
+        unsafe { !LLVMIsAFunction(self.0).is_null() }
+    }
+
+    pub fn is_vararg_function(&self) -> bool {
+        let el_ty = self.get_type().get_element_type().get();
+        (unsafe { LLVMIsFunctionVarArg(el_ty) }) != 0
+    }
+
     pub fn is_call(&self) -> bool {
         unsafe { !LLVMIsACallInst(self.0).is_null() }
     }
@@ -180,6 +216,14 @@ impl Value {
         }
     }
 
+    pub fn get_num_arg_operands(&self) -> u32 {
+        unsafe { LLVMGetNumArgOperands(self.0) }
+    }
+
+    pub fn get_called_value(&self) -> Value {
+        unsafe { Self::new(LLVMGetCalledValue(self.0)) }
+    }
+
     pub fn opcode(&self) -> LLVMOpcode {
         unsafe {
             debug_assert!(!LLVMIsAInstruction(self.0).is_null());
@@ -193,6 +237,12 @@ impl Value {
 
     pub fn kind(&self) -> LLVMValueKind {
         unsafe { LLVMGetValueKind(self.0) }
+    }
+
+    pub fn get_name(&self) -> &str {
+        let mut size = MaybeUninit::<usize>::uninit();
+        let name = unsafe { CStr::from_ptr(LLVMGetValueName2(self.0, size.as_mut_ptr())) };
+        name.to_str().unwrap()
     }
 }
 
@@ -228,7 +278,7 @@ pub unsafe fn get_aot_original(instr: &Value) -> Option<Value> {
             // If this isn't a pointer it can't be YKCtrlPointVars.
             return None;
         }
-        let ty = Type(LLVMGetElementType(ty.0));
+        let ty = ty.get_element_type();
         if !ty.is_struct() {
             // If this isn't a struct it can't be YKCtrlPointVars.
             return None;
--- a/tests/c/stopgap_call_int_ret.c
+++ b/tests/c/stopgap_call_int_ret.c
@@ -1,0 +1,48 @@
+// Run-time:
+//   env-var: YKD_PRINT_IR=jit-pre-opt
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_PRINT_JITSTATE=1
+//   stderr:
+//     ...
+//     f: 3
+//     jit-state: enter-jit-code
+//     ...
+//     jit-state: enter-stopgap
+//     ...
+//     f: 2
+//     jit-state: exit-stopgap
+//     ...
+
+// Check the stop-gap interpreter can call out to AOT-compiled functions.
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk.h>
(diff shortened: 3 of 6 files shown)


ptersilie:
> whose interfaces *are* not known until runtime

yksgi/src/lib.rs
```diff
@@ -313,8 +316,68 @@ impl SGInterp {
...
+
+            // Now build the calling interface and collect the values of the callee's arguments.
+            //
+            // It may be tempting to use `LLVMGetArgOperand()` to extract callee operands from our
+            // `CallInst`, but at the time of writing this wraps the C++ method
+            // `FuncletPadInst::getArgOperand()` which is not what we want.
+            //
+            // `CallBase::getArgOperand()` isn't currently exposed by the LLVM C API, so we instead
+            // use the implicit knowledge that for a `CallInst` the first `N` operands are the
+            // operands of the function being called (the remaining operand is the callee).
```
ptersilie, line 343: This is fine and there are various other places where we similarly need to use implicit knowledge about the order of arguments of a specific instruction. I don't see those special LLVM functions being exposed anytime soon as I imagine there's others.

yksgi/src/lib.rs
```diff
@@ -313,8 +316,68 @@ impl SGInterp {
...
+            //
+            // Despite this, `LLVMGetNumArgOperands()` *is* generic and does the right thing for
+            // any callable function-like thing, so we can still use that for the loop bounds.
+            let mut builder = FFIBuilder::new();
+            let mut arg_vals = Vec::new();
+            for i in 0..LLVMGetNumArgOperands(call_inst) {
+                let arg = Value::new(LLVMGetOperand(call_inst, i.try_into().unwrap()));
```
ptersilie, line 350: if you wrap call_inst inside `Value`, e.g. `Value::new(call_inst)`, then you can use `get_operand(idx)`. I would probably do that in line 288. Oh, now that I look at it, it is already wrapped `self.pc`. So you could just leave it in `self.pc`.

yksgi/src/lib.rs
```diff
@@ -283,7 +285,8 @@ impl SGInterp {
     /// Implement call instructions. Returns `true` if the control point has been reached,
     /// `false` otherwise.
     unsafe fn call(&mut self) -> bool {
-        let func = LLVMGetCalledValue(self.pc.get());
+        let call_inst = self.pc.get();
+        let func = LLVMGetCalledValue(call_inst);
```
ptersilie, line 289: Regarding my comment below I would probably change this to:
```
        let func = LLVMGetCalledValue(self.pc.get());
```
and leave `self.pc` as is, and replace `call_inst` with `self.pc/self.pc.get()` below.

yksgi/src/llvmbridge.rs
```diff
@@ -120,6 +125,24 @@ impl Type {
     pub fn as_str(&self) -> &CStr {
         unsafe { CStr::from_ptr(LLVMPrintTypeToString(self.0)) }
     }
+
+    /// Converts an LLVM type into a libffi type.
+    pub fn ffi_type(&self) -> FFIType {
+        match unsafe { LLVMGetTypeKind(self.0) } {
```
ptersilie, line 131: `match unsafe { self.kind() } {`

yksgi/src/llvmbridge.rs
```diff
@@ -169,6 +192,10 @@ impl Value {
         unsafe { Type(LLVMTypeOf(self.0)) }
     }
 
+    pub fn get_ffi_type(&self) -> FFIType {
```
ptersilie, line 195: Is it cleaner if we do `ret_val.get_type().ffi_type()` instead of having this extra function here?

ptersilie:
You were right. This is nasty. But apart from some minor comments looks good to me.

yksgi/src/lib.rs
ltratt: I tend to agree: the unusual name of the variable isn't helpful for readers.

yksgi/src/lib.rs
```diff
@@ -313,8 +316,68 @@ impl SGInterp {
...
+                // once the loop body ends.
+                arg_vals.push(self.var_lookup(&arg).val);
+            }
+            let ret_val = Value::new(call_inst);
+            builder = builder.res(ret_val.get_ffi_type());
+            let cif = builder.into_cif(); // OPT: cache CIFs for repeated calls to same func sig.
+
+            // Actually do the call.
+            let ret_ty = ret_val.get_type();
+            let ffi_arg_vals: Vec<FFIArg> = arg_vals.iter().map(|a| ffi_arg(a)).collect();
```
ltratt, line 364: An `into_iter` here *might* help the compiler reuse the underlying Vec? I might be too hopeful though!

vext01: Wouldn't we risk destructing the old vector, leaving not one dangling pointer, but a whole vector of them?

yksgi/src/lib.rs
vext01: In light of that, is this comment too long?

ptersilie: Yeah, I guess we could shorten it a bit. If we write long paragraphs for each weirdness/missing feature in the LLVM C API we'll be here for a long time.

yksgi/src/llvmbridge.rs
vext01: I prefer to have the helper to make the consumers more concise.

yksgi/src/lib.rs
vext01: I reviewed this part of code and overhauled some of the wrapper usage. See d0327ba5d.

yksgi/src/lib.rs
vext01: Should be addressed by d0327ba5d, albeit by a slightly different means.

yksgi/src/lib.rs
```diff
@@ -313,8 +316,68 @@ impl SGInterp {
...
+                // FIXME: https://github.com/ykjit/yk/issues/536
+                match ret_ty.get_int_width() {
+                    32 => {
+                        let rv = cif.call::<u32>(FFICodePtr(fptr), &ffi_arg_vals) as u64;
+                        self.var_set(self.pc, SGValue::new(rv, ret_ty));
+                    }
+                    _ => todo!(),
+                }
+            } else if ret_ty.is_void() {
+                dbg!(&arg_vals);
```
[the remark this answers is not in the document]
vext01: Fixed in 90263fe0beb

yksgi/src/llvmbridge.rs
vext01: Fixed in b0811a97

yksgi/src/lib.rs
vext01: Aggressively shortened in 130766753 :)

yksgi/src/llvmbridge.rs
ptersilie: I just feel like `get_ffi_type` shouldn't be a function on a `Value` as it solely relates to the type. Adding 3 lines to save 7 characters (`_type()`) looks like a net loss to me. ;-)

vext01:
Something went wrong with git here. Attempting to fix.

yksgi/src/llvmbridge.rs
vext01: Fixed in ce3da0363

vext01:
OK fixed. I had to do a force push, so I've gone back and updated all the references to git hashes in my prior comments.

yksgi/src/lib.rs
ptersilie: I like how you ignored what @ltratt and I both agreed on (replacing `call_inst` with `self.pc`, and instead did something completely different. :D I guess it's fine now, though I didn't really have an issue with using the `LLVM*` calls. I tried to only make wrappers for those that I thought will be used more than once. I didn't think we needed one for `LLVMIsFunctionVarArg` for example.

Anyway, let's see what @ltratt thinks.

vext01: *ignored* is a bit strong. Truth be told, once I had added and used wrappers for various things, it made more sense to do it like this.

yksgi/src/llvmbridge.rs
```diff
@@ -188,6 +188,18 @@ impl Value {
...
+    pub fn is_inline_asm(&self) -> bool {
+        !unsafe { LLVMIsAInlineAsm(self.0).is_null() }
+    }
+
+    pub fn is_function(&self) -> bool {
+        unsafe { !LLVMIsAFunction(self.0).is_null() }
+    }
+
+    pub fn is_vararg_function(&self) -> bool {
+        unsafe { LLVMIsFunctionVarArg(LLVMGetElementType(LLVMTypeOf(self.0))) != 0 }
```
ptersilie, line 200: s/LLVMTypeOf(self.0)/self.get_type()

yksgi/src/llvmbridge.rs
```diff
@@ -217,6 +237,12 @@ impl Value {
     pub fn kind(&self) -> LLVMValueKind {
         unsafe { LLVMGetValueKind(self.0) }
     }
+
+    pub fn get_name(&self) -> &str {
+        let mut size = MaybeUninit::<usize>::uninit();
+        let name = unsafe { CStr::from_ptr(LLVMGetValueName2(self.0, size.as_mut_ptr())) };
```
ptersilie, line 239: Somewhere else in this code I just did:
```
        let mut size = 0;
        let name = CStr::from_ptr(LLVMGetValueName2(func, &mut size))
            .to_str()
            .unwrap();
```
Do we need to use `MaybeUninit` here? If so, we should change the other one too.

yksgi/src/lib.rs
vext01: Wait, having re-read the conversation, I'm confused.

> unusual name of the variable 

What is unusual about `call_inst`? `self.pc` is a pointer to a `CallInst` instruction.

(I introduced the variable to avoid repeatedly doing `self.pc` everywhere)

yksgi/src/llvmbridge.rs
vext01: Your suggestion works (just for now?), but the [Rust docs](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#out-pointers) recommend using `MaybeInit` for out pointers, so I've gotten into the habit of using it.

yksgi/src/llvmbridge.rs
vext01: I considered this. It would have to be `self.get_type().get()`, at which point I thought "meh".

yksgi/src/lib.rs
ptersilie: It's not too unusual for me, but I still think it's unneccesary (we don't do this anywhere else in the code either).

You added `let call_inst = self.pc` so you don't have to repeatedly type `self.pc` but instead have to repeatedly type `const_inst` which is even 3 characters longer? ;-)

vext01: It's not about making it short. It's about avoiding (potentially) extra field lookups, and it also makes clear that `self.pc` doesn't change in the interim.

ptersilie: Fair enough. Though the compiler is going to remove the field lookup anyway. It's just that the rest of the code never does this, so I just wanted to keep it in style. If you feel strongly about it I'm happy for you to leave it, if @ltratt is happy too.

vext01: Let me put it differently.

If I gave you code like:
```
f(a + b)
g(a + b)
h(a + b)
```

Wouldn't you change it to:
``
let sum = a + b;
f(sum)
g(sum)
h(sum)
```

?

Of course you would :P Even if the compiler can do common sub-expression elimination, it just makes for nicer code.

Comment (vext01) [reply yksgi/src/lib.rs]: > Though the compiler is going to remove the field lookup anyway.

We raced, but I preempted this comment :P

File: yksgi/src/llvmbridge.rs (hunk 8 above)
Comment (ptersilie) [reply yksgi/src/llvmbridge.rs:200]: Maybe a good candidate to implement `LLVMGetElementType`. I think we are using it in other places too. So we get maybe `self.get_type().get_element()`?

File: yksgi/src/lib.rs (hunk 3 above)
Comment (ptersilie) [reply yksgi/src/lib.rs]: I would argue that `self.pc` isn't the same as `a + b`. But I think this isn't worth to argue about, so I'll leave the decision to @ltratt.

**Comment by vext01** [comment]
That push resolves the conflicts.

[I was unable to find a way to retrospectively sign a merge commit during a rebase. Does anyone know how?]

OK to squash?

**Comment by vext01** [comment]
For future reference, this works:
https://stackoverflow.com/posts/9958215/revisions

I'll do this for future PRs.

File: yksgi/src/llvmbridge.rs (hunk 8 above)
Comment (vext01) [reply yksgi/src/llvmbridge.rs]: Actually that wrapper already exists, just we don't use it.

Fixed in af5b4fc5df (to be squashed in) and cce4e3a3c (to be left separate).

**Comment by vext01** [comment]
OK to squash?

File: yksgi/src/lib.rs (hunk 6 above)
Comment (ltratt) [reply yksgi/src/lib.rs:401]: If it's not safe, Rust will tell you.

File: yksgi/src/lib.rs (hunk 3 above)
Comment (ltratt) [reply yksgi/src/lib.rs]: Of course @vext01 is right in general, but in this instance, I *really* want to know if you're using the pc or not. IMHO either call the variable `pc` or use `self.pc` (the two are probably optimised to the same, so it's a matter of style).

Comment (vext01) [reply yksgi/src/lib.rs]: Is the confusion stemming from the fact that `self.pc` is a program counter in the sense that it is a pointer to the current instruction?

When we are in the `call()` method of the interpreter we know that `self.pc` can only point to an instruction that does some kind of a call, hence `self.pc` is a pointer to an instruction that makes some kind of call, hence `call_inst`.

File: yksgi/src/lib.rs (hunk 6 above)
Comment (vext01) [reply yksgi/src/lib.rs:401]: Wrong. Creating a dangling pointer is safe in Rust. If we then pass that dangling pointer into an unsafe interface like `CIF::call()` (as we do here) then boom.

File: yksgi/src/lib.rs (hunk 3 above)
Comment (ltratt) [reply yksgi/src/lib.rs]: I understand that reasoning in one way -- but what's the point of that degree of indirection, when it confuses the code reader, and there's a clearer equivalent which requires no more characters to type?

File: yksgi/src/lib.rs (hunk 6 above)
Comment (ltratt) [reply yksgi/src/lib.rs:401]: Hang on, what will dangle? `Vec.into_iter` is safe Rust: nothing is left dangling.

File: yksgi/src/lib.rs (hunk 3 above)
Comment (vext01) [reply yksgi/src/lib.rs]: I find it much clearer to narrow the general `self.pc` into a more descriptive name, given the invariant of the  context.

In fact, I find the name `self.pc` confusing itself. I think it should be `self.cur_inst`.

Comment (ltratt) [reply yksgi/src/lib.rs]: If you want to change `pc` to `cur_inst` that's a debate / PR we can have -- but this isn't the right place to do it. Consistency has a virtue in and of itself, and `cur_inst` just doesn't sit with the rest of the code base. Sorry to be a dick about it!

File: yksgi/src/lib.rs (hunk 6 above)
Comment (vext01) [reply yksgi/src/lib.rs:401]: `ffi_arg` caches the reference argument you pass it as a raw pointer, so if you consume what is referenced, then the pointer dangles.

See:
 - [here](https://docs.rs/libffi/latest/src/libffi/middle/mod.rs.html#49-51)
 - [and here](https://docs.rs/libffi/latest/src/libffi/middle/mod.rs.html#40-42)

Comment (vext01) [reply yksgi/src/lib.rs:401]: (Note that I imported `arg()` as `ffi_arg()` for clarity)

Comment (ltratt) [reply yksgi/src/lib.rs:401]: > ffi_arg() caches the reference argument you pass it as a raw pointer

Then this code is undefined behaviour: there is no guarantee that a `Vec`s backing memory stays at a stable address.

File: yksgi/src/lib.rs (hunk 3 above)
Comment (vext01) [reply yksgi/src/lib.rs]: It's not a hill I'm willing to die on, so I'll back down and quietly fume in a corner :P

Comment (ltratt) [reply yksgi/src/lib.rs]: You have my permission to print my ugly mug onto a dartboard :)

File: yksgi/src/lib.rs (hunk 6 above)
Comment (vext01) [reply yksgi/src/lib.rs:401]: I don't see how `libffi` crate could work if that were the case.

Is it not the case that a `Vec`'s backing storage stays put if you don't mutate the vector?

Comment (ltratt) [reply yksgi/src/lib.rs:401]: > Is it not the case that a Vec's backing storage stays put if you don't mutate the vector?

It kind of depends how you read https://doc.rust-lang.org/std/vec/struct.Vec.html -- I would be pessimistic and say "no" as it only says that *some* operations won't change the backing store. That's a weaker guarantee than I suspect was intended, but it is what it is.

Personally, I think that if you want to refer to the *address* of something, you must be in control of the allocation (e.g. a `Box` is fixed in memory), otherwise you're at the mercy of other code.

Comment (vext01) [reply yksgi/src/lib.rs:401]: That would be horribly inefficient. There must be a better way. This is the whole point of libffi -- you don't know the shape of the function until runtime.

The [example in the `libffi` docs](https://docs.rs/libffi/latest/libffi/middle/struct.Cif.html#examples) uses an array. What if we make a slice from the vec so that the compiler can't possibly move the backing storage?

Comment (ltratt) [reply yksgi/src/lib.rs:401]: Let's try and split two things apart.

First, `ffi::arg` is this horror https://docs.rs/libffi/latest/src/libffi/middle/mod.rs.html#49 -- IMHO that should be an `unsafe` function because converting `&T` to `*const T` does not communicate to the user that they can't move `&T` afterwards. [It would be better, probably, if `Arg::new` took in a `*const T` in the first place.]

Second, the example in https://docs.rs/libffi/latest/libffi/middle/struct.Cif.html#examples a) does not create a `Vec`, it creates a fixed-size array b) does not reference addresses in that fixed-size array.

The problem is that we're effectively doing this:

```rust
let v = vec![1,2,3];
let addr = &v[0] as *const _;
```

If the `Vec` moves its backing store -- and the docs do not say that it won't -- then `addr` can point to an invalid address and we're in undefined behaviour. This is the situation we're in.

Just to reiterate: the interace to `libffi::Arg::new` seems designed to let seemingly safe Rust code descend into actually unsafe Rust code without making clear that's happened. Eek! Also, allocating a chunk of memory for the args will be *exactly as efficient as creating a Vec to store them* (we know how big the chunk will be at allocation time), so there's no real reason not to do it.

vext01: After a lot of offline back and forth, we've decided to use a manually allocated chunk of memory (which can't move).

Hopefully dc39229b does the trick. Tests pass, but I'm unsure if getting a mutable reference to the slice and writing into it is correct. Please @ltratt can you have a look?

yksgi/src/lib.rs
vext01: Fixed in fc190961fe2d

ltratt:
> Is it undefined if we only read from the slice once it has been fully initialised?

I believe that creating a Rust object (inc. a slice) whose backing memory is (even partly) uninitialised is undefined behaviour.

jacob-hughes:
Yes a reference to uninitialized data is considered an _invalid reference_. A Rust program is UB as soon as any invalid value is created, regardless of if it's used. See https://www.ralfj.de/blog/2020/07/15/unused-data.html

vext01:
Those two commits address those comments.

But I have a question: can the memory for the frames move while the stop-gap interpreter is being run? They are stored as a vector of hashmaps in `self`. If they don't move, we can reference them directly and avoid the `alloc()` dance once and for all.

ltratt:
> can the memory for the existing frames move while the stop-gap interpreter is being run?

Not sure what you mean. If you're asking "can something I haven't allocated, and that isn't `Pin`, move when I don't expect it" the answer is broadly "yes". But you might not mean that?

vext01:
To clarify: currently we are copying the argument values out of the SGI's stack frame and referencing those copies when making the call. Can we directly reference the SGI's frames without the copy step? Sounds like the answer is no.

ltratt:
I don't understand the context well enough. Sorry.

vext01:
@ptersilie what do you think?

vext01:
I've convinced myself that we can't reference the frames directly unless they too are allocated in a way that they cannot move.

OK to squash?

ptersilie:
I'm okay with squashing.

yksgi/src/lib.rs
```diff
@@ -414,6 +415,8 @@ impl SGInterp {
             } else {
                 todo!("{:?}", ret_ty.as_str());
             };
+            drop(ffi_arg_vals);
```
ltratt, line 418: I think a comment explaining why `drop` is needed would be useful (I had to stop and think!).

vext01: Is 4d5b199 OK?

ltratt:
Please squash.

vext01:
splat.

vext01:
gah, hold on, typo in the commit message

vext01:
OK, ready.

PR #549, opened 2022-04-25
Tidy ups

ltratt:
Some random tidy-ups, partly found by Clippy.

The only tricky one is https://github.com/ykjit/yk/commit/e0aceedc0b78eb97fc08bedf52462ecbf463a2ba which @vext01 might want to take a quick peek at.

### DIFF
--- a/tests/src/bin/run_trace_compiler_test.rs
+++ b/tests/src/bin/run_trace_compiler_test.rs
@@ -10,7 +10,7 @@ use yktrace::{IRBlock, IRTrace};
 const BBS_ENV: &str = "YKT_TRACE_BBS";
 
 fn parse_bb(bb: &str) -> Result<(CString, usize), Box<dyn Error>> {
-    let mut elems = bb.split(":");
+    let mut elems = bb.split(':');
     let func = elems.next().ok_or("malformed function name")?;
     let bb_idx = elems
         .next()
@@ -23,7 +23,7 @@ fn main() -> Result<(), String> {
     // Build the trace that we are going to have compiled.
     let mut bbs = vec![IRBlock::unmappable()];
     if let Ok(tbbs) = env::var(BBS_ENV) {
-        for bb in tbbs.split(",") {
+        for bb in tbbs.split(',') {
             if let Ok((func, bb_idx)) = parse_bb(bb) {
                 bbs.push(IRBlock::new(func, bb_idx));
             } else {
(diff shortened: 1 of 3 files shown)


ltratt:
bors try

yktrace/src/lib.rs
```diff
@@ -25,7 +25,7 @@ thread_local! {
     //
     // We hide the `ThreadTracer` in a thread local (rather than returning it to the consumer of
     // yk). This ensures that the `ThreadTracer` itself cannot appear in traces.
-    pub static THREAD_TRACER: RefCell<Option<ThreadTracer>> = const { RefCell::new(None) };
+    pub static THREAD_TRACER: RefCell<Option<ThreadTracer>> = RefCell::new(None);
```
vext01, line 28: Creating a `RefCell` (calling [`RefCell::new()`](https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.new)) is `const`, so this may be right? IIRC, the `const` isn't talking about whether the data is mutable, but whether the call can be evaluated at compile time(?).

(I didn't know you had to explcitiely mark it `const` though)

ltratt: Clippy complained obliquely about this `const`: I'm not sure what it means for a `RefCell` to be const if it's instantiated multiple times in multiple threads. Unless we know why it's here, I think that at most `const` is an optimisation (though possibly not even that in this case), and possible it's incorrect.

vext01: > I'm not sure what it means for a RefCell to be const if it's instantiated multiple times in multiple threads.

Should be fine. It's a thread local.

ltratt: What happens if `const` here means "share one RefCell across multiple threads"? I don't think we know precisely what `const` means, and there's a small chance it interacts incorrectly with thread locals.

vext01: That'd be bonkers if that's what happened, but I'll find the answer.

vext01: I can't find it spelled out anywhere, but some part of tokio has applied the [same optimisation recently](https://github.com/tokio-rs/tracing/pull/2003/commits/2020bbaae9541136f2bd97de8f15ad955bf1921d) (it's still under PR, but I think the intent is clear).

FWIW, [tracking issue for the feature](https://github.com/rust-lang/rust/issues/84223) (no mention of surprising semantic changes).

I just don't see why `thread_local! {const ...}` would make something *not* a thread local :P

ltratt: Maybe try running clippy locally and see what you think of its complaint here?

vext01: I'm now questioning clippy. Is it is conflating const data with a const block?!

https://rust-lang.github.io/rust-clippy/master/index.html#declare_interior_mutable_const

Look at the examples. This is different to what is going on here!

And [there it is](https://github.com/rust-lang/rust-clippy/issues/8493). This is a clippy bug!

ltratt: Good find. So I should delete the commit that makes this change, I think?

vext01: I think so, yes. May as well force push that now.

ltratt:
Force pushed.

PR #550, opened 2022-04-25
Drop mutexes in the `Compiling` state properly.

ltratt:
Previously we were calling `drop(&Arc<Mutex<...>>)` which is a no-op. The fix is relatively easy: we treat the `Compiling` state as the other states which have allocated a `Box`, turning it back into a  `Box` which is then implicitly dropped.

### DIFF
--- a/ykrt/src/location.rs
+++ b/ykrt/src/location.rs
@@ -306,10 +306,10 @@ impl Drop for Location {
                 // trace that's pointed to. There should be a ref count that we decrement and free
                 // memory if it reaches zero.
                 self.unlock();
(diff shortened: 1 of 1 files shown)


ptersilie:
Where is the code that turns the `Compiling` state into a `Box`? Or is that happening implicitly?

ltratt:
It's the `Box::from_raw(hl)` bit. Basically this PR just makes sure that line is called.

ptersilie:
Oh I see. This just updates the condition leading to the boxing. Gotcha!

bors r+

PR #551, opened 2022-04-26
Allow arbitrary return types in control points caller.

### LINKED ISSUE #537: Stopgap interpreter assumes that the function enclosing the intepreter loop returns i32.

ptersilie:
We previously required (as a shortcut) that the function containing the control point always returns an `int`. This PR fixes this by allowing arbitrary types in the control points caller. We first fix being able to return `void` in aa0e2b829 and then enable other types in b6ebc9b91.

Companion PR: https://github.com/ykjit/ykllvm/pull/34
Fixes #537


ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -684,6 +689,21 @@ class JITModBuilder {
       FailBuilder.CreateStore(
           ConstantInt::get(PointerSizedIntTy, LiveVals.size()), GEP);
 
+      // Store the stackmap address and length in a separate struct to save
+      // arguments.
+      AllocaInst *StackMapStruct = FailBuilder.CreateAlloca(
+          VecLenStructTy, ConstantInt::get(PointerSizedIntTy, 1));
+      GEP = FailBuilder.CreateGEP(VecLenStructTy, StackMapStruct,
+                                  {ConstantInt::get(PointerSizedIntTy, 0),
+                                   ConstantInt::get(Int32Ty, 0)});
```
ltratt, line 698: Why `Int32Ty`?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -1081,7 +1100,16 @@ class JITModBuilder {
     PointerSizedIntTy = DL.getIntPtrType(Context);
 
     // Create a function inside of which we construct the IR for our trace.
-    JITFunc = createJITFunc(TraceInputs);
+    Type *ReturnTy;
+    if (CPCI->getNumArgOperands() != YK_CONTROL_POINT_NUM_ARGS) {
+      // This means we are running the trace_compiler tests which only uses a
```
ltratt, line 1105: Are we sure this is the only way this can happen?

ykllvmwrap/src/jitmodbuilder.cc
vext01: I was curious.

According to the [docs](https://llvm.org/docs/LangRef.html#i-getelementptr):
> When indexing into an array, pointer or vector, integers of any width are allowed, and they are not required to be constant.

So may as well use the smallest type that can express the constant (here `i1`)?

vext01: No, no, no :P

This is a struct, so i read the wrong bit:

> When indexing into a (optionally packed) structure, only i32 integer constants are allowed (when using a vector of indices they must all be the same i32 integer constant).

ltratt: Aha!

ptersilie:
If there are no more comments then I believe this is also ready for merging once https://github.com/ykjit/ykllvm/pull/34 is in.

ltratt:
Please squash (and don't forget to rebase at the same time!).

ltratt:
Gah, wrong PR. But please force push a rebase anyway!

ptersilie:
Done.

Issue #541, opened 2022-04-14
Don't reuse guard failure blocks.


ptersilie:
We currently reuse the same guard failure block for all guard failures. This is incorrect for obvious reasons (i.e. different guards can have different live variables).

PR #552, opened 2022-04-27
Fix bugs to make ykcbf run.

### LINKED ISSUE #547: Segfault
Using ykcbf I can reliably cause a segfault by:

```
$ echo 123456 | ./bf_simple_yk lang_tests/numwarp.bf
```

Here's a `gdb` backtrace:

```
core::ptr::read<u8> (src=0x1002390ab)
    at /rustc/879aff385a5fe0af78f3d45fd2f0b8762934e41e/library/core/src/ptr/mod.rs:1096
1096    /rustc/879aff385a5fe0af78f3d45fd2f0b8762934e41e/library/core/src/ptr/mod.rs: No such file or directory.
(gdb) bt
#0  core::ptr::read<u8> (src=0x1002390ab)
    at /rustc/879aff385a5fe0af78f3d45fd2f0b8762934e41e/library/core/src/ptr/mod.rs:1096
#1  0x00007ffff52dad8f in yksgi::SGInterp::load (self=0x7fffffffd918)
    at yksgi/src/lib.rs:445
#2  0x00007ffff52d9107 in yksgi::SGInterp::interpret (self=0x7fffffffd918)
    at yksgi/src/lib.rs:223
#3  0x00007ffff52c1fd7 in ykcapi::yk_stopgap (sm_addr=0x7ffff4ac2000, 
    sm_size=488, aotmap=0x7fffffffdc50, actframes=0x7fffffffe060, 
    retaddr=140737298282887, rsp=0x7fffffffdc08) at ykcapi/src/lib.rs:236
#4  0x00007ffff52c27e9 in ykcapi::__llvm_deoptimize (addr=<optimized out>, 
    size=<optimized out>, aotmap=<optimized out>, curpos=<optimized out>)
    at ykcapi/src/lib.rs:259
#5  0x0000000000001000 in ?? ()
#6  0x00007ffff52c27d0 in ?? ()
   from /home/ltratt/yk/target/debug/deps/libykcapi.so
#7  0x00007fffffffdc50 in ?? ()
#8  0x00007fffffffe060 in ?? ()
#9  0x00007fffffffe5f8 in ?? ()
#10 0x00000000000001e8 in ?? ()
#11 0x00007ffff4ac2000 in ?? ()
#12 0x00007fffffffe260 in ?? ()
#13 0x00007fffffffdc48 in ?? ()
#14 0x00007ffff4abb987 in __yk_compiled_trace_4 ()
#15 0x00007ffff57f707d in yktrace::CompiledTrace::real_exec (self=0x7fffe8006740, ctrlp_vars=0x7fffffffe558) at yktrace/src/lib.rs:277
#16 0x00007ffff57f6f90 in yktrace::CompiledTrace::exec (self=<optimized out>, ctrlp_vars=<optimized out>) at yktrace/src/lib.rs:246
#17 0x0000000000000000 in ?? ()
```

ptersilie:
This fixes #547, which happened because LLVM produced a GEP instruction with a negative index, which is allowed for vectors (struct indexes however are always positive).

Disclaimer: I'm not 100% sure this implementation is correct, mainly because it requires vector indexes to be i64, which to me suggest that the size of a vector is limited to half of u64. I spent a good amount of time digging through LLVM to see how they implement GEP but didn't quite learn much more, apart from that they also use `GetSExtValue` for constants, suggesting that (at least) constant values can only be i64s.


yksgi/src/lib.rs
```diff
@@ -425,38 +425,48 @@ impl SGInterp {
...
+                // Struct indexes are always positive.
+                let idx = if op.is_constant() {
+                    LLVMConstIntGetZExtValue(op.get())
+                } else {
+                    self.var_lookup(&op).val
+                };
                 let off = LLVMOffsetOfElement(layout, curty.get(), u32::try_from(idx).unwrap());
-                offset += off;
+                // Since vector indexes expect an i64, and struct indexes are limited to 32bit, we
+                // can savely cast to i64 here.
```
ltratt, line 451: s/savely/safely/

yksgi/src/lib.rs
```diff
@@ -425,38 +425,48 @@ impl SGInterp {
...
+                offset += off as i64;
```
ltratt, line 452: I think we can use `i64::from(off)` here (which is safer than `as`).

ptersilie: It would have to be a `try_from` since `off` is ` u64`. Fixed in 837ac91.

yksgi/src/lib.rs
ptersilie: Fixed in 837ac91.

yksgi/src/lib.rs
ltratt: Ah, then `try_from` is the right thing, but the comment makes no sense to me as it mention "32 bit" -- but there doesn't seem to be 32 bits anywhere in this calculation?

ptersilie: I've just revised the comment in ee2a5c8, but now realise, that this was referring to the index `idx`, which will always be i32 when indexing into a struct. But I am forcing the offset to be i64, which is not the same. For example if `idx` is `u32::max`, then `LLVMOffsetOfElement` will certainly be bigger than `u32::max` and thus not fit into `i64`, right? Which is the same worry I mentioned in my disclaimer above.

ltratt: Que?

ptersilie: Forget what I said earlier. Try with [53d58d7](https://github.com/ykjit/yk/pull/552/commits/53d58d79bd9684e29eb6aa9ade86cdb4262941b8)

ltratt:
Please squash.

ltratt:
I missed something! We should add the "test" from https://github.com/ykjit/yk/issues/547 to `.buildbot.sh` so this bug never reoccurs.

ltratt:
@ptersilie This one is still open.

ptersilie:
Yes, I was working on adding the test, when I realised ykbf is not working properly due to https://github.com/ykjit/yk/issues/541. So I need to fix that one first. Might just do it in this PR.

ptersilie:
The last commit fixes #541. The change looks bigger than it is due to a change in indentation.

ptersilie:
One thing I'm unsure about now is how to test this. I don't think I can use `ykcbf` as a test anymore, since there's no guarantee that it'll run long enough to stopgap, and thus hit the bugs this PR fixes. Especially with https://github.com/ykjit/yk/pull/553 it will become even less likely that we'll ever hit the stopgap interpreter.

ptersilie:
I initially made a test for `cargo test`. So are you suggesting only testing this inside `.buildbot.sh`? The problem is that with `YKD_SERIALISE_COMPILATION=1` I have not had the patience to see if it ever finishes. It might well take _literal ages_.

ptersilie:
I haven't pushed it yet, since I wasn't sure it's useful.

ptersilie:
Here you go.

tests/c/bfbug.c
```diff
@@ -0,0 +1,165 @@
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=0
```
ltratt, line 2: This is the default so doesn't need to be set.

[It means the test won't test much useful yet, admittedly, as we already know.]

ltratt:
Maybe what we should do is mark the new test as `ignored: too slow with current decoding speeds to be run` or similar?

tests/c/bfbug.c
ptersilie: Indeed. And setting it to `1` means it won't terminate in a reasonable amount of time (I'm pretty sure I had it running for at least 20 minutes), and thus might not be something we want bors to run each time.

ptersilie:
> Maybe what we should do is mark the new test as `ignored: too slow with current decoding speeds to be run` or similar?

I can live with that.

ptersilie:
> LLVM should optimise away repeated guards if we're lucky. Might be worth checking first.

Not for unrolled loops I think. I checked `jit-post-opt` and could still see multiple guards, one for each unrolled iteration. I don't think you can remove those as the guard is needed.

ptersilie:
I've added some ignored tests, but probably will have to change the ignore message, since with your PR the test now isn't too slow to run, but in contrary too fast to even hit the tracer.

ltratt:
Yes, maybe it's more like "ignore: runs too fast to check what should be checked. reenable when the PT decoder is faster"?!

ptersilie:
Fixed. I neeed to squash and rebase too.

ltratt:
Please squash.

tests/c/bfbug.c
```diff
@@ -0,0 +1,165 @@
...
+//     ...
+//     jit-state: enter-jit-code
+//     jit-state: exit-jit-code
+//     ...
+//     jit-state: enter-stopgap
+//     ...
+//   stdin:
+//     1234
+
+// This is bf_base.c from https://github.com/ykjit/ykcbf modified to hard-code the input to the
```
ltratt, line 15: Just realised: this is probably based on `yk_simple_bf.c` not `bf_base.c`?

tests/c/bfbug.c
```diff
@@ -0,0 +1,165 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+#define CELLS_LEN 30000
+#define INPUT_PROG                                                             \
```
ltratt, line 30: We can't remove the copyright on this: we need to credit Daniel B Cristofani.

tests/c/bfbug.c
ptersilie: Fixed in [c4ec947](https://github.com/ykjit/yk/pull/552/commits/c4ec947b6c1b9d4c9cb7ee9f795f92c1492a2a71)

tests/c/bfbug.c
ptersilie: Fixed in [830647f](https://github.com/ykjit/yk/pull/552/commits/830647fbefcdbf3989cdfb2bed7b5311acdd18ba)

tests/c/bfbug.c
ltratt: Works for me, though the commented line is now very long?

vext01:
What's the status of this PR?

ptersilie:
Needs a re-review from @ltratt and then squashing and rebasing.

tests/c/bfbug.c
ptersilie: Shortened in https://github.com/ykjit/yk/pull/552/commits/01ddf7364f7133b9134b2b6d978a6a076d815a73

And while I was there I reformatted too which I hadn't done yet.

ltratt:
Please squash.

ptersilie:
Okay to squash/rebase?

ptersilie:
Squashed.

PR #553, opened 2022-05-06
Move raw trace -> IR trace mapping into a compilation thread.

ltratt:
**DRAFT PR: DO NOT MERGE!**

Previously the HW tracer did the mapping immediately after stopping PT, stalling the rest of the system while the (potentially very slow) mapping was computed.

This commit returns the unmapped trace, allowing the compilation thread to do the mapping itself.

@ptersilie can you see if this solves the stalling problem you were seeing? If so, I will go through and tidy this up before making the final PR.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -22,7 +22,7 @@ use crate::{
     location::{HotLocation, HotLocationKind, Location, LocationInner},
     print_jit_state,
 };
-use yktrace::{start_tracing, stop_tracing, CompiledTrace, IRTrace, TracingKind};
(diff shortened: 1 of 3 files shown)


ptersilie:
Yes, this works and allows my ykcbf test to finish running now. Though I'm not sure how useful that test is now, since we are likely never seeing a compiled trace. At least not until we filter out obviously large traces before they are being decoded.

vext01:
This looks pretty inoffensive to me :)

ltratt:
Squashed.

ptersilie:
Hah, all because a missing (and irrelevant) `,`. :D

ltratt:
Sausages. OK to squash?

ptersilie:
Please squash.

PR #557, opened 2022-07-25
Fixes due to ykllvm upgrading to LLVM 14

ptersilie:
Compantion PR to: https://github.com/ykjit/ykllvm/pull/35


ltratt:
Looks good to me. Will wait for @vext01 to confirm.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -579,10 +579,6 @@ class JITModBuilder {
     IntegerType *Int32Ty = Type::getInt32Ty(Context);
     PointerType *Int8PtrTy = Type::getInt8PtrTy(Context);
 
-    // Create struct type for storing a vector and its length.
```
vext01, line 581: I don't understand this commit.

> According to the LLVM docs "opaque pointer mode is automatically disabled for IR and bitcode files that explicitly mention i8* style typed pointers".

This is talking about old-style LLVM pointer types, e.g. `i8*` and new-style ones `ptr`. The latter being opaque.

AFAIK, the only reason we don't use opaque pointers for the JIT, is that many of the intrinsics we use still have an old-style pointer in their signature.

But I don't see any opaque pointers being removed here. What have I missed?

vext01:
Just one comment from me.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Previous we've used `VecPtrTy` for all those `Vec/Len` structs. The opaque pointer type of which is created just underneath your comment ;-) See https://github.com/ykjit/yk/pull/557#diff-b76ed8a29ed650e7c2a44a5809bc05a22cca958d0b9f5532dab74cce8503d112L582. `PointerType::get(Context, 0);` creates an opaque pointer.

My understanding as to why opaque pointer mode is turned off is because we are using `i8*` pointers throughout this code. But I may have misunderstood the LLVM docs. If we can we should probably just replace everything with opaque pointers as LLVM recommends. But that would be in a different PR anyway, if it's possible.

vext01: I suggest we don't move to opaque pointers just yet, as patchpoints, which we theorise we will need, has this signature:

```
declare void
  @llvm.experimental.patchpoint.void(i64 <id>, i32 <numBytes>,
                                     i8* <target>, i32 <numArgs>, ...)
```

Note the `i8*`. This would stop us using opaques.

vext01: So really, this commit just needs an updated commit message I think?

ptersilie: What do you want to update about the commit message? Does it say something wrong?

vext01: Ah sorry. So it does indeed remove opaque pointers.

vext01:
Any squashing required, or ready to go?

ptersilie:
Think this is good to go.

ptersilie:
Oh, we also have to change `.buildbot.sh` which still points to 13.x. I've made 14.x the default branch in ykllvm, so I think we can just omit the checkout line here: https://github.com/ykjit/yk/blob/master/.buildbot.sh#L39

ptersilie:
Let's try this.

ptersilie:
I don't understand why this fails. Investigating.

ptersilie:
Using latest rust-nightly (which buildbot does) required some additional fixes.

Let's give this another go.

ltratt:
Please squash

vext01:
I'm ok with those changes also. Squash as you see fit.

ptersilie:
Not sure these last two commits require squashing. They stand on their own as each fixes an issue related to upgrading to lastest rust nightly. Though they probably could have been in their own PR. But not sure it's worth it as this point to change it.

vext01:
OK, so let's go!

bors r+

ptersilie:
bors try

ptersilie:
bors try

ptersilie:
bors try

ptersilie:
bors try

ptersilie:
bors try

ltratt:
Please squash.

ptersilie:
Got into a right mess with all the different branches and repos. All fixed now. Last 4 commits need squashing.

ltratt:
Please squash 'em :)

ptersilie:
Squashed.

PR #560, opened 2022-08-15
Don't use the LLVM symbolizer to look up symbols.

vext01:
Part of the mapping phase of JIT requires the ability to map a block's
virtual address to the symbol from which it came.

Before we were using the `llvm::LLVMSymbolizer` API for this, but
recently when debugging an issue, we found that this API favours debug
info over the symbol table. This means that each time we map a block, we
look for DWARF debug info, and that seems a little wasteful and
questionably accurate.

This change therefore switches to using `dladdr(3)` instead.

Using the test suite as a kind of benchmark, running the tests with a
single thread like this:
```
multitime -n 10 env RUST_TEST_THREADS=1 cargo test --release
```

Before this commit:
```
1: env RUST_TEST_THREADS=1 cargo test --release
real        38.131      0.636       36.739      38.260      39.202
user        260.719     0.979       259.012     260.509     262.621
sys         561.254     1.084       559.982     561.146     563.089
```

After this commit:
```
1: env RUST_TEST_THREADS=1 cargo test --release
real        36.826      0.775       35.561      37.182      38.165
user        252.629     0.911       250.878     252.606     254.124
sys         541.665     2.248       538.241     542.094     544.800
```

(Note that 3 new tests were added, so this isn't quite apples to apples)

Although the speedup isn't dramatic, the code is simplified quite a bit,
so I think this is worthwhile.

I also think that this could be sped up further with caching, but that's
one for another day.


ykutil/src/addr.rs
```diff
@@ -81,11 +82,36 @@ pub fn off_to_vaddr(containing_obj: &Path, off: u64) -> Option<usize> {
     None // Not found.
 }
 
+// Given a virtual address in the current address space, (if possible) determine the name of the
```
ltratt, line 85: These should be `///` doc comments.

ykutil/src/addr.rs
```diff
@@ -81,11 +82,36 @@ pub fn off_to_vaddr(containing_obj: &Path, off: u64) -> Option<usize> {
...
+// symbol this belongs to, and the path to the object from which it came.
+//
+// On success returns `Ok((symbol_name, object_path))`, or on failure `Err(())`.
+//
+// This function uses `dladdr()` internally, and thus inherits the same symbol visibility rules
+// used there. For example, this function will not find unexported symbols.
+//
+// Note that the returned object path may not be canonical.
```
ltratt, line 93: What does this last sentence mean?

ltratt:
Out of curiosity, if you do profile an instance (with `perf` or `valgrind` or whatever) how much time is spent in `vaddr_to_sym_and_obj` (and its descendents)? If it is substantial, maybe we should just do the interval tree thing now?

ykutil/src/addr.rs
vext01: As in, a [canonical path](https://doc.rust-lang.org/std/fs/fn.canonicalize.html).

We could clarify that this is just the absolute path.

ltratt: I know what a canonical path is, but it doesn't make sense here, as we may (are likely to be!) running on a different box where the paths no longer make sense other than (hopefully) unique IDs.

vext01:
Both perf and valgrind seemed upset with the dwarf info in my binary (e.g. perf spews `addr2line: DWARF error: invalid or unhandled FORM value: 0x25`), but I did manage to tease this out of perf after blindly ignoring the cascading errors for about 30 seconds (is it trustworthy?).

I adjusted the `simple_interp_loop2` test to run more iterations (a few seconds of runtime) and did:

```
$ perf record -g --call-graph dwarf /tmp/edd
$ perf report -G
```

And got:

```
Samples: 2K of event 'cycles', Event count (approx.): 1810423079                                                                                                                                                                              
  Children      Self  Com  Shared Object      Symbol                                                                                                                                                                                          
-   27.12%     0.00%  edd  libykcapi.so       [.] yktrace::hwt::mapper::HWTMapper::map_trace                                                                                                                                                 ◆
   - yktrace::hwt::mapper::HWTMapper::map_trace                                                                                                                                                                                              ▒
      + 25.34% <alloc::boxed::Box<I,A> as core::iter::traits::iterator::Iterator>::next                                                                                                                                                      ▒
      - 1.78% yktrace::hwt::mapper::HWTMapper::map_block                                                                                                                                                                                     ▒
         + 1.72% ykutil::addr::code_vaddr_to_off                                                                                                                                                                                             ▒
```

Overall context:
```
Samples: 2K of event 'cycles', Event count (approx.): 1810423079                                                                                                                                                                                Children      Self  Command  Shared Object        Symbol                                                                                                                                                                                    
+   46.22%     0.00%  edd      [unknown]            [k] 0xffffffffffffffff                                                                                                                                                                    
+   45.10%     0.00%  edd      libc-2.31.so         [.] __libc_start_main                                                                                                                                                                     
+   45.10%     3.48%  edd      edd                  [.] main                                                                                                                                                                                  
+   43.46%    19.41%  edd      libc-2.31.so         [.] __memset_avx2_erms                                                                                                                                                                    +   42.93%     0.00%  edd      [kernel.kallsyms]    [k] asm_exc_page_fault                                                                                                                                                                    
+   42.01%     3.59%  edd      libykcapi.so         [.] __ykrt_control_point                                                                                                                                                                  
+   32.97%     0.32%  edd      libykcapi.so         [.] ykrt::mt::MT::control_point                                                                                                                                                           
+   32.93%     2.27%  edd      libykcapi.so         [.] ykrt::mt::MT::transition_location                                                                                                                                                     
+   27.12%     0.00%  edd      libykcapi.so         [.] yktrace::hwt::mapper::HWTMapper::map_trace                                                                                                                                            +   27.03%     0.00%  edd      libykcapi.so         [.] ykrt::mt::MT::queue_job::{{closure}}                                                                                                                                                  
+   27.03%     0.00%  edd      libykcapi.so         [.] lock_api::mutex::MutexGuard<R,T>::unlocked                                                                                                                                            
+   27.03%     0.00%  edd      libykcapi.so         [.] <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once                                                                                                              
+   27.03%     0.00%  edd      libykcapi.so         [.] core::ops::function::FnOnce::call_once{{vtable.shim}}                                                                                                                                 
+   27.03%     0.00%  edd      libykcapi.so         [.] ykrt::mt::MT::queue_compile_job::{{closure}}                                                                                                                                          
+   27.03%     0.00%  edd      libykcapi.so         [.] <yktrace::hwt::PTTrace as yktrace::UnmappedTrace>::map                                                                                                                                
+   26.98%     0.00%  edd      libykcapi.so         [.] __rust_try                                                                                                                                                                            
+   26.98%     0.00%  edd      libykcapi.so         [.] std::panicking::try::do_call                                                                                                                                                          
+   26.98%     0.00%  edd      libykcapi.so         [.] <core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once                                                                                         +   26.98%     0.00%  edd      libykcapi.so         [.] std::thread::Builder::spawn_unchecked_::{{closure}}::{{closure}}                                                                                                                      
+   26.98%     0.00%  edd      libykcapi.so         [.] std::sys_common::backtrace::__rust_begin_short_backtrace                                                                                                                              
+   26.94%     0.00%  edd      libc-2.31.so         [.] __GI___clone (inlined)                                                                                                                                                                +   26.94%     0.00%  edd      libpthread-2.31.so   [.] start_thread                                                                                                                                                                          +   26.93%     0.00%  edd      libykcapi.so         [.] std::sys::unix::thread::Thread::new::thread_start                                                                                                                                     +   26.93%     0.00%  edd      libykcapi.so         [.] core::ops::function::FnOnce::call_once{{vtable.shim}}                                                                                                                                 +   26.93%     0.00%  edd      libykcapi.so         [.] std::thread::Builder::spawn_unchecked_::{{closure}}                                                                                                                                   
+   26.93%     0.00%  edd      libykcapi.so         [.] std::panic::catch_unwind                                                                                                                                                              +   26.93%     0.00%  edd      libykcapi.so         [.] std::panicking::try                                                                                                                                                                   +   25.34%     0.00%  edd      libykcapi.so         [.] <alloc::boxed::Box<I,A> as core::iter::traits::iterator::Iterator>::next                                                                                                              +   25.34%     0.00%  edd      libykcapi.so         [.] <hwtracer::backends::perf_pt::PerfPTBlockIterator as core::iter::traits::iterator::Iterator>::next                                                                                    +   25.32%     0.00%  edd      libykcapi.so         [.] perf_pt_next_block                                                                                                                                                                    
+   25.32%     0.00%  edd      libykcapi.so         [.] pt_blk_next                                                                                                                                                                           +   25.32%     0.00%  edd      libykcapi.so         [.] pt_blk_proceed                                                                                                                                                                        
+   25.32%     0.00%  edd      libykcapi.so         [.] pt_blk_proceed_no_event                                                                                                                                                               +   25.12%     0.00%  edd      libykcapi.so         [.] pt_blk_msec_lookup                                                                                                                                                                    
+   25.12%     0.00%  edd      libykcapi.so         [.] pt_blk_msec_fill                                                                                                                                                                      
+   24.39%     0.05%  edd      libykcapi.so         [.] pt_bcache_alloc                                                                                                                                                                       
+   24.39%     0.00%  edd      libykcapi.so         [.] pt_section_request_bcache                                                                                                                                                             
+   24.39%     0.00%  edd      libykcapi.so         [.] pt_section_alloc_bcache                                                                                                                                                               +   24.08%     0.00%  edd      [kernel.kallsyms]    [k] exc_page_fault                                                                                                                                                                        
+   24.08%     0.00%  edd      [kernel.kallsyms]    [k] do_user_addr_fault                                                                                                                                                                    
+   24.04%     0.13%  edd      [kernel.kallsyms]    [k] handle_mm_fault                                                                                                                                                                       
+   23.00%     0.10%  edd      [kernel.kallsyms]    [k] do_huge_pmd_anonymous_page                                                                                                                                                            
+   20.33%    20.28%  edd      [kernel.kallsyms]    [k] clear_page_erms                                                                                                                                                                       +   11.61%     0.27%  edd      [kernel.kallsyms]    [k] clear_huge_page
```
This was on a release mode rust build with debug info. C code built at `-g -O2`.

I believe the `0xffff...` entry it the jitted code. Maybe this is what upsets the tools?

If we trust this info, mapping vaddrs to symbols is a bit under 2% of runtime.

ltratt:
Ok, that's good enough: it's not worth us worrying about right now. This PR is still good enough for its simplification aspects. Merge as-is?

ykutil/src/addr.rs
vext01: I'm not sure i follow.

This function queries addresses from the same virtual address space as the currently running executable. Not sure how another machine factors into it. What am I missing?

ltratt: "Canonical paths" only make sense for actual files. We don't have the original files here (at least if I understand things correctly!): what we have is a binary which contains files. The paths we're getting here are the "full paths" of those files when they were compiled.

vext01: Another detail, I added this comment because [one of my assertions](https://github.com/ykjit/yk/pull/560/files#diff-59272cdbf7ded73ed95ce99cc146e32b5e6ff445c7e18e8b12efed0fb0a90c1dR255-R261) failed when i tried to compare two equivalent, but non-canonical paths. The linker doesn't report canonical paths.

ltratt: Right, so we're using these paths as IDs, right? And they're not necessarily reliable IDs is what I think you're saying.

vext01: Well, yes, but they are the only ID the linker (and LLVM IR) really speaks when it comes to identifying functions...

ltratt: Right, so I would suggest that we really need to be talking in terms of IDs, not canonical paths (but I'm willing to be proven wrong!). Does it mean that the comment should be "these IDs are not reliable" or ... ?

vext01: I know what you are saying, but I dislike the term ID in this context.

Yes, a "canonical path" is a "unique ID", but a "canonical path" is a more accurate term here: all canonical paths are a unique ID, but not all unique IDs are a canonical path. Why use the more general term and leave room for misinterpretation?

Can we compromise on something like:

> Note that the returned object path may not uniquely identify an object file, since `dladdr(3)` doesn't necessarily return canonical paths.

vext01: Or, maybe better, kill the sentence. We've already said this uses `dladdr(3)` under the hood, so a reader can know what to expect :)

ltratt: Killing the sentence works for me!

vext01:
OK to squash?

ltratt:
There's also https://github.com/ykjit/yk/pull/560#discussion_r945686875.

vext01:
My apologies. Fixed in a865cfc.

ltratt:
Please squash.

vext01:
splat.

sorry about the double push. forgot to sign.

PR #561, opened 2022-08-17
Update the stackmap parser for live variable groupings.

vext01:
This change reflects a ykllvm change where each stackmap live variable
can now map to many locations, instead of just one.

Companion to https://github.com/ykjit/ykllvm/pull/40

Compiler side merges first.


vext01:
This is now ready.

PR #565, opened 2022-09-13
Store live values in memory rather than the stack.

ptersilie:
This significantly reduces the number of instructions of large traces and as such takes pressure of the JIT compiler. As an example: before, a large Lua trace consisting of 80k instructions would have 60k of those come from guard failure code. Now, those 60k instructions are reduced to a few hundred, which significantly speeds up both trace construction and compilation.

The downside of this is that we now store more information in memory which needs to stay alive as long as the trace does.


ptersilie:
I had a quick look at combining all malloced regions together using `realloc`. But this is a bit more involved than I currently think is worth the time. For future reference, I believe there's two approaches:

1) After the trace is fully constructed go back to each guard failure and patch the argument in the deopt call with the latest realloced address + offset.
2) During guard failure construction pass only an offset into the deoptcall. Then when the trace is executed pass in the address of the reallocated address which we store alongside the trace.

vext01:
> The downside of this is that we now store more information in memory which needs to stay alive as long as the trace does.

The information just moved from llvm instructions to our own store. It's all memory at the end of the day.

vext01:
> I had a quick look at combining all malloced regions together using realloc. But this is a bit more involved than I currently think is worth the time.

agreed.

ltratt:
I might be missing out on something (I don't see multiple malloc calls in the commit), but what I was expecting was something like:

```c
size_t current_size = ...;
size_t  = 0;
void *block = malloc(...);
for guard in ... {
   if (off + sizeof(guard) > current_size) {
    block = realloc(block, current_size + ...);
    current_size += ...
   }
   ...
}
```

i.e. all the guards stored contiguously. Maybe they are anyway?

ykcapi/src/lib.rs
```diff
@@ -153,9 +153,13 @@ pub extern "C" fn yk_stopgap(
     print_jit_state("enter-stopgap");
 
     // Parse AOTMap.
+    // FIXME: The memory behind this slice was malloced during compilation of the trace and thus
+    // must only be freed if the trace is too.
```
vext01, line 157: You can implement freeing now in the `drop` of the trace, right? It just won't be called (since the trace isn't dropped).

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -671,53 +678,28 @@ class JITModBuilder {
 
     // Create a vector in which to store the locations of the corresponding
     // AOT variables.
-    StructType *AOTLocTy =
-        StructType::get(Context, {PointerSizedIntTy, PointerSizedIntTy,
-                                  Int8PtrTy, PointerSizedIntTy});
-    AllocaInst *AOTLocVec = FailBuilder.CreateAlloca(
-        AOTLocTy, ConstantInt::get(PointerSizedIntTy, LiveValues.size()));
-    std::map<std::string, Value *> FuncPtrMap;
+    AOTInfo *AOTMappingVec = (AOTInfo *) malloc(sizeof(AOTInfo) * LiveValues.size());
```
vext01, line 681: C-style cast in C++ code here. Is the cast even required in C++?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -671,53 +678,28 @@ class JITModBuilder {
...
-                                   ConstantInt::get(Int32Ty, 2)});
-      FailBuilder.CreateStore(FPtr, GEP);
-      GEP = FailBuilder.CreateGEP(AOTLocTy, AOTLocVec,
-                                  {ConstantInt::get(PointerSizedIntTy, I),
-                                   ConstantInt::get(Int32Ty, 3)});
-      FailBuilder.CreateStore(
-          ConstantInt::get(PointerSizedIntTy, StackFrameIdx), GEP);
+      const char* FName = AOTVar->getFunction()->getName().data();
+
+      AOTInfo AI;
```
vext01, line 691: Would it be neater to use an initializer instead of lots of assignments?

```
AOTInfo AI  = { BBIdx, InstrIdx, ...}
```

vext01:
Is that the whole change? Somehow I expected it to be more work than this.

ptersilie:
> > The downside of this is that we now store more information in memory which needs to stay alive as long as the trace does.
> 
> The information just moved from llvm instructions to our own store. It's all memory at the end of the day.

But before it was allocated on the stack so we didn't have to worry about freeing it.

vext01:
> But before it was allocated on the stack so we didn't have to worry about freeing it.

The IR that builds the stack storage also takes heap space and is (for now) immortal, just like the malloc'd data.

Don't worry about it :)

ptersilie:
> Is that the whole change? Somehow I expected it to be more work than this.

Yeah, it took some experimenting but as an example I was able to reuse the `AOTVar` struct in `ykcapi` so this did not require any changes on that end. We basically went from allocating stuff on the stack to allocating it on the heap. The format remained exactly the same.

ltratt:
> If realloc moves the buffer, then we will have to go back and patch old traces.

Yes, so we'll need to store offsets (which are stable) in the generated code, not addresses.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: I tried that first, but the compiler didn't like it. I can try and have another go.

ptersilie:
> Yes, so we'll need to store offsets (which are stable) in the generated code, not addresses.

Yes, that's the second solution I proposed above. It's slightly annoying as we need to funnel the actual address back to the trace when we execute it, which requires me to change a lot of files. It's not hard, just annoying.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -671,53 +678,28 @@ class JITModBuilder {
...
+      AI.FName = FName;
+      AI.FrameIdx = StackFrameIdx;
+      AOTMappingVec[I] = AI;
     }
 
     // Store the live variable vector and its length in a separate struct to
     // save arguments.
     AllocaInst *AOTLocs = storeVecWithLength(
-        FailBuilder, AOTLocVec,
+        FailBuilder, ConstantInt::get(PointerSizedIntTy, (uintptr_t)AOTMappingVec),
```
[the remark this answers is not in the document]
ptersilie: What should I use instead?

vext01:
I think you should be able to do it in such a way that the trace calls a routine to do the lookup, so that we don't have to patch each trace, or pass around the info.

What bothers me a bit is that lookups into the guard failure table are not going to be atomic and will require syncronisation (one thread reallocs the buffer, while another is using it). 

Would it not be better to have failing out of one trace totally independent of other traces?

What is the advantage of storing the memory contiguously?

I don't think we can say which approach is better without measurements.

ltratt:
>  What bothers me a bit is that lookups into the guard failure table are not going to be atomic and will require syncronisation (one thread reallocs the buffer, while another is using it).

The block is created in a single thread. There are no race issues here.

ptersilie:
We are talking about a), which I haven't done in this PR.

ltratt:
I'm talking about "a) all guard failures for a single trace are stored contiguously." [(b) would be bonkers!]

vext01:
OK, so to implement a), (much easier) can we not pre-scan the trace to determine the overall size required for guard failure info, then allocate the block?

In terms of finding the buffer, I think you can either:
 - pass a pointer to it into the trace, or
 - give the buffer a symbol like `__guard_failure_table_0`, where `0` is the trace number.

ltratt:
I'm happy to leave *how* it's implemented up to @ptersilie and yourself :)

vext01:
I'd still be happy for this to be done incrementally in two PRs.

ptersilie:
> OK, so to implement a), (much easier) can we not pre-scan the trace to determine the overall size required for guard failure info, then allocate the block?

Not really as the size is determined but the number of live values which we only know once we hit the branch that creates the guard failure.

> In terms of finding the buffer, I think you can either:
> 
>     * pass a pointer to it into the trace, or
> 
>     * give the buffer a symbol like `__guard_failure_table_0`, where `0` is the trace number.

I've actually decided to patch the deopt calls. It's the simplest solution and only requires me to temporarily store the deopt instructions in a vector and then before we finalise the trace we just patch one of the arguments. It should only be a simple loop.

ptersilie:
I've overhauled the code. All live AOT values are now collected in a one contiguous memory block and then passed into the guard failure. For this the `CompiledTrace` now stores a pointer to the allocated memory, which allows us to easily free it when the trace is dropped. ([4482223](https://github.com/ykjit/yk/pull/565/commits/44822239d234a103bb343e04140afba93c6b406d))

I've also cleaned up some other stuff, moved code around, fixed comments, and renamed some variables.

Ready for re-review.

ltratt:
Works for me! @vext01 ?

ykcapi/src/lib.rs
```diff
@@ -136,13 +136,23 @@ pub struct CVec {
...
+}
+
 /// Reads the stackmap and saved registers from the given address (i.e. the return address of the
 /// deoptimisation call).
 #[cfg(target_arch = "x86_64")]
 #[no_mangle]
 pub extern "C" fn yk_stopgap(
     stackmap: &CVec,
-    aotmap: &CVec,
+    aotvals: &LiveAOTVals,
```
vext01, line 155: Small nit: would it be better to receive this argument as a raw pointer? We are only going to cast it in a moment anyway.

ykcapi/src/lib.rs
```diff
@@ -152,10 +162,14 @@ pub extern "C" fn yk_stopgap(
...
-    // Parse AOTMap.
-    let aotmap = unsafe { slice::from_raw_parts(aotmap.addr as *const AOTVar, aotmap.length) };
+    // Parse the live AOT values.
+    let aotvalsptr =
+        unsafe { (aotvals.addr as *const u8).offset(isize::try_from(aotvals.offset).unwrap()) };
+    let aotvals = unsafe { slice::from_raw_parts(aotvalsptr as *const AOTVar, aotvals.length) };
 
     // Parse active frames vector.
+    // Note that the memory behind this slice is allocated on the stack (of the compiled trace) and
+    // thus doesn't need to be freed.
```
vext01, line 172: We replace the frame of the trace in the event of guard failure, is that correct?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -296,23 +307,24 @@ class JITModBuilder {
     GlobalMappings.insert({CF, FAddr});
   }
 
-  // Store a vector and its length as a struct. This allows us to save
-  // arguments when passing this information via the deoptimisation call.
-  AllocaInst *storeVecWithLength(IRBuilder<> &Builder, Value *Vec, Value *Len) {
+  // Generate LLVM IR to create a struct and store the given values.
```
vext01, line 310: Does this store into the heap allocation?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -665,66 +680,41 @@ class JITModBuilder {
...
-        ConstantInt::get(PointerSizedIntTy, ActiveFrames.size()));
+    AllocaInst *ActiveFramesStruct = createAndFillStruct(
+        FailBuilder, {ActiveFrameVec, ConstantInt::get(PointerSizedIntTy,
+                                                       ActiveFrames.size())});
+
+    // Make more space to store the locations of the corresponding live AOT
+    // values for this guard failure.
+    size_t Offset = LiveAOTSize;
+    LiveAOTSize += sizeof(AOTInfo) * LiveValues.size();
+    LiveAOTAddr = (AOTInfo *)realloc(LiveAOTAddr, LiveAOTSize);
```
vext01, line 691: We should use `reallocarray(3)` here to avoid the possibility of a heap overflow.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -665,66 +680,41 @@ class JITModBuilder {
...
+    // Get a pointer to this guard failure's region in the memory block.
+    AOTInfo *CurrentRegion = (AOTInfo *)((char *)LiveAOTAddr + Offset);
```
vext01, line 693: We should probably use C++ casts throughout.

vext01: Can `LiveAOTAddr` be typed `AOTInfo *` from the outset to avoid the repeated casts?

vext01: Also, since the size of `struct AOTInfo` is fixed, would it be better to keep track of not an offset, but a "next free element" index. Then you can do:

```
AOTInfo *CurrentRegion = LiveAOTAddr[Idx];
```

thus killing the cast to `char *`.

This would also go hand-in-hand with the `reallocarray` change I proposed.

yktrace/src/lib.rs
```diff
@@ -210,14 +212,19 @@ impl CompiledTrace {
...
     pub fn new(code_ptr: *const c_void) -> Self {
-        let slice = unsafe { slice::from_raw_parts(code_ptr as *const usize, 3) };
+        let slice = unsafe { slice::from_raw_parts(code_ptr as *const usize, 4) };
         let funcptr = slice[0] as *const c_void;
         let smptr = slice[1] as *const c_void;
         let smsize = slice[2] as usize;
+        let aotvals = slice[3] as *mut c_void;
+        // We heap allocated this array in ykllvmwrap to pass the data here. Now that we've
+        // extracted it we no longer need to keep the array around.
+        unsafe { libc::free(code_ptr as *mut c_void) };
```
vext01, line 222: I'm not clear on this part.

Is the intention to free the AOTLiveMap array?

I think some of the confusion stems from the fact that `code_ptr` appears to be more than a pointer to code.

vext01:
A few minor comments from me.

ykcapi/src/lib.rs
ptersilie: Yes, we are rolling back to the frame that called the trace, so everything on the stack gets nuked.

ykcapi/src/lib.rs
ptersilie: I don't follow. It's a struct and I want to read out its fields. Passing this as a raw pointer would make reading out those fields much less nice.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: It's an alloca, which is stack allocated.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: I could,  but then I would have to keep around another variable "Idx", because I still need "Offset" to pass it into the deopt call.

yktrace/src/lib.rs
ptersilie: Yeah, that variable name is awful. I dunno why it's called that tbh. This doesn't free the AOTLiveMap. It just frees a struct we made in ykllvmwrap to reduce the amount of arguments we need to pass to deopt. The AOTLiveMap can only be freed when the trace dies.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Also, I find thinking about a "next free element" index more difficult than thinking about an offset. But that might just be me.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: What's the advantage of this?

vext01: It checks that `sizeof(AOTInfo) * LiveValues.size();` doesn't overflow, which can lead to memory errors.

ykllvmwrap/src/jitmodbuilder.cc
vext01: OK, let's leave it for now.

yktrace/src/lib.rs
vext01: Does that struct, the one used for argument reduction, really need to be heap allocated?

Shall we rename this variable then?

ykllvmwrap/src/jitmodbuilder.cc
vext01: My comment about casts still stands.

ykcapi/src/lib.rs
vext01: Sorry, I misread.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: > Can `LiveAOTAddr` be typed `AOTInfo *` from the outset to avoid the repeated casts?

It is typed `AOTInfo *` from the outset. But `realloc` returns a `void *` so I need to cast it back to `AOTInfo *`.

yktrace/src/lib.rs
ptersilie: Unfortunately, it does. Though what I said above wasn't fully correct. It's not used to reduce arguments. It's used to return various things from the trace compiler back to Rust (https://github.com/ykjit/yk/blob/master/ykllvmwrap/src/ykllvmwrap.cc#L224), so we can construct a `CompiledTrace` from it. Once we've constructed the `CompiledTrace` the struct is no longer needed. At least it only lives a short time and once we got the redesigned `HotLocation` up, I imagine we would instead pass in pointers which we then write to.

ykllvmwrap/src/jitmodbuilder.cc
vext01: Sure, but why not use a C++ cast instead of a C cast? We are writing C++ code after all :)

yktrace/src/lib.rs
vext01: Maybe later we should pass the struct by value, but let's leave it for now. Possibly add a `FIXME`?

ptersilie: There's already a fixme here: https://github.com/ykjit/yk/blob/master/ykllvmwrap/src/ykllvmwrap.cc#L224

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Fixed in [9f64524](https://github.com/ykjit/yk/pull/565/commits/9f64524a8c6ae37403c0a32a379e26fa9abd2270) and [3bab316](https://github.com/ykjit/yk/pull/565/commits/3bab31655c95d2572784bae2201be15289a2404d).

To be fair, this whole realloc business wouldn't have been necessary had I stuck to using a simple heap allocated vector. But that would have made @ltratt unhappy. I'm doing such splits here to please the both of you, my balls are touching the floor. :coconut: :laughing:

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Fixed in [9f64524](https://github.com/ykjit/yk/pull/565/commits/9f64524a8c6ae37403c0a32a379e26fa9abd2270)

ptersilie:
Think that addresses all. Ready for re-review.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -686,11 +686,14 @@ class JITModBuilder {
 
     // Make more space to store the locations of the corresponding live AOT
     // values for this guard failure.
-    size_t Offset = LiveAOTSize;
-    LiveAOTSize += sizeof(AOTInfo) * LiveValues.size();
-    LiveAOTAddr = (AOTInfo *)realloc(LiveAOTAddr, LiveAOTSize);
+    size_t CurPos = LiveAOTNum;
+    LiveAOTNum += LiveValues.size();
+    LiveAOTArray = (AOTInfo *)reallocarray(LiveAOTArray, LiveAOTNum, sizeof(AOTInfo));
+    if (LiveAOTArray == NULL) {
```
ltratt, line 692: This can just be an `assert` IMHO since we don't appear to be able to recover from an error.

ptersilie: Fixed in [a52ef2a](https://github.com/ykjit/yk/pull/565/commits/a52ef2afdae235c5ef07bfb6ba44e2631607ec16)

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -221,10 +222,11 @@ extern "C" void *compileModule(string TraceName, Module *M,
   // Allocate space for compiled trace address, stackmap address, and stackmap
   // size.
   // FIXME This is a temporary hack until the redesigned hot location is up.
-  uintptr_t *ptr = (uintptr_t *)malloc(sizeof(uintptr_t) * 3);
+  uintptr_t *ptr = (uintptr_t *)malloc(sizeof(uintptr_t) * 4);
   ptr[0] = EE->getFunctionAddress(TraceName);
   ptr[1] = (uintptr_t)SMR.Ptr;
   ptr[2] = SMR.Size;
+  ptr[3] = (uintptr_t)AOTMappingVec;
```
vext01, line 229: You are gonna hate me, but the PR still introduces a lot of C-style casts.

vext01: thanks.

so `static_cast` didn't work, hence `reinterpret_cast`?

ptersilie: Yeah, `static_cast` isn't allowed cause we're casting a pointer to an int.

ptersilie:
Ok to squash now?

vext01:
You may :)

ptersilie:
Squashed! :tada: 

I had to squash everything into one commit, as after all those changes I couldn't salvage any other commits. At least not without more effort than it's worth.

ptersilie:
Hold on, I need to format again.

ptersilie:
Done. Also forgot to rename another variable. Ok to squash?

vext01:
Go for it

ptersilie:
Done.

PR #566, opened 2022-09-14
Avoid current_exe() and path canonicalisation when mapping.

vext01:
This basically works by `replacing std::env::current_exe()` by querying the linker for the object containing `main()` and using that path exclusively.

This has the knock-on effect that we no longer have one mechanism giving us a non-canonical path (`dladdr(3)`, when asking where a virtual address came from), and one mechanism giving us (possibly) an absolute path (`std::env::current_exe()`). Therefore we no longer need to canonicalise paths, which is a win in itself because canonicalisation hits the filesystem, slowing us down.

Running numwarp.bf on the input `1234`:

BEFORE:
```
1: sh -c "echo \"1234\" |  ./bf_simple_yk lang_tests/numwarp.bf"
real        1.638       0.011       1.623       1.642       1.654
user        1.241       0.014       1.218       1.244       1.256
sys         0.348       0.014       0.331       0.344       0.369
```

AFTER:
```
1: sh -c "echo \"1234\" |  ./bf_simple_yk lang_tests/numwarp.bf"
real        1.479       0.007       1.471       1.474       1.489
user        1.194       0.013       1.172       1.197       1.208
sys         0.235       0.016       0.210       0.233       0.259
```

So about 10% quicker, which is close enough to the ~7% cost of canonicalisation that I measured using perf.

(this is with Lukas' experimental PT cache and guard compression diffs applied, as without those, the cost of canonicalisation is swamped by other inefficiencies).

### DIFF
--- a/yktrace/src/hwt/mapper.rs
+++ b/yktrace/src/hwt/mapper.rs
@@ -17,6 +17,7 @@ use std::{
     io::{prelude::*, Cursor, SeekFrom},
 };
 use ykutil::addr::{code_vaddr_to_off, off_to_vaddr_main_obj, vaddr_to_sym_and_obj};
+use ykutil::obj::SELF_BIN_PATH;
 
 const BLOCK_MAP_SEC: &str = ".llvm_bb_addr_map";
 static BLOCK_MAP: LazyLock<BlockMap> = LazyLock::new(BlockMap::new);
@@ -215,8 +216,7 @@ impl HWTMapper {
         // FIXME: https://github.com/ykjit/yk/issues/413
         // In the future we could inline code from shared objects if they were built for use with
         // yk (i.e. they have a blockmap and IR embedded).
-        let cur_exe = env::current_exe().unwrap();
-        if obj_name != cur_exe {
+        if obj_name != *SELF_BIN_PATH {
             return Vec::new();
         }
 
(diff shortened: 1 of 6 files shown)


vext01:
I haven't spent much time tidying this. I suspect I can probably refactor some of the surrounding code given more time.

vext01:
All tests pass, by the way.

ltratt:
This looks like a definite improvement to me! Any reason this is a draft PR? It looks mergeable to me!

vext01:
> Any reason this is a draft PR? It looks mergeable to me!

Well! I thought you would object to the hackiness of it all.

Also:

> I haven't spent much time tidying this. I suspect I can probably refactor some of the surrounding code given more time.

We can merge it if you like, and I can look to do more tidying tomorrow in another PR.

vext01:
Can do.

ptersilie:
:+1:

vext01:
I could also rebase this at squash time if you like.

ltratt:
Please squash.

vext01:
splat.

vext01:
> error: entry symbol `main` declared multiple times

well, that's a new one on me.

vext01:
I was only able to reproduce this after upgrading Rust!

Then I boiled it down to this (in a lib crate):

```
use libc::{c_char, c_int};

extern "C" {
    fn main(argc: c_int, argv: *const *const c_char) -> c_int;
}

pub fn f() {
    dbg!(main as *const u8);
}

#[cfg(test)]
mod tests {
    use super::f;

    #[test]
    fn it_works() {
        f();
    }
}
```

Running `cargo test` will make it explode.

```
...
   Compiling ppp v0.1.0 (/tmp/ppp)
error: entry symbol `main` declared multiple times
  |
  = help: did you use `#[no_mangle]` on `fn main`? Use `#[start]` instead

```

My guess: some kind of clash with the test binary main, or something like that?

ltratt:
No idea!

vext01:
This tuned out to be a Rust bug.

https://github.com/rust-lang/rust/issues/101906

I've worked around it with C code for now.

OK to squash?

ltratt:
Please squash.

vext01:
splat.

PR #567, opened 2022-09-22
Fish the LLVM bitcode out of the address space.

vext01:
ykllvm now puts the LLVM bitcode into a loaded section, so we can take it out of our address space instead of loading it from disk.

This doesn't speed anything up, but it does eliminate a use of `current_exe()` and also a couple of dependency crates.

### DIFF
--- a/ykutil/src/obj.rs
+++ b/ykutil/src/obj.rs
@@ -1,11 +1,7 @@
 //! Utilities for dealing with object files.
 
 use libc::{c_void, dladdr, Dl_info};
-use memmap2::Mmap;
-use object::{self, Object, ObjectSection, Section};
-use std::{
-    convert::TryFrom, env, ffi::CStr, fs, mem::MaybeUninit, path::PathBuf, ptr, sync::LazyLock,
-};
+use std::{ffi::CStr, mem::MaybeUninit, path::PathBuf, ptr, sync::LazyLock};
 
 extern "C" {
     fn find_main() -> *const c_void;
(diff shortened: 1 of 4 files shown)


ykutil/src/obj.rs
```diff
@@ -27,27 +23,24 @@ pub static SELF_BIN_PATH: LazyLock<PathBuf> = LazyLock::new(|| {
...
-/// The current executable parsed to an `object::File`.
-static EXE_OBJ: LazyLock<object::File> =
-    LazyLock::new(|| object::File::parse(&**EXE_MMAP).unwrap());
+/// The `llvm.embedded.module` symbol in the `.llvmbc` section.
+#[repr(C)]
+struct EmbeddedModule {
+    /// The length of the bitcode.
+    len: usize,
+    /// The start of the bitcode itself.
+    first_byte_of_bitcode: u8,
```
ptersilie, line 32: This may be a silly question, but why is this a `u8` and why can we just use a reference to this `u8` later on to load the bitcode from? I assume some magic happens in `#[link_name = "llvm.embedded.module"]` but I'd be curious to know what this magic is.

vext01: The `.llvmbc` section is now magically loaded as the interpreter starts, and it contains a symbol `llvm.embedded.module` whose binary format we know exactly at runtime:

 - a `usize` `length` field.
 - a bitcode payload of length `length` (the previous field).

(and here's no padding between these)

Thanks to the recent `ykllvm` change, the `extern C` block anchors a Rust-level struct with identical ABI to this symbol (hence the `link_name` bit). This means that when we read this struct in Rust, we actually read the symbol generated by ykllvm.

We don't know the length of the payload statically, but if we did, I'd have used an array instead of a `u8`. Since we don't, the `first_byte_of_bitcode` field just lets us get a reference to the beginning of the bitcode payload.

The LLVM API to load bitcode requires a pointer to the bitcode and the length of it. The length is `len` and the pointer to the bitcode is `&first_byte_of_bitcode`. The memory for the section is static, and thus eternal, so it's safe to pass pointers to it around.

Hope that makes sense.

ltratt: Since there's no padding, I would probably suggest removing `first_byte_of_bitcode` and making the calculation of that address `std::mem::sizeof::<usize>()`, but I don't feel strongly about it.

vext01: That's what I started with, but changed to just using the ABI of the struct, since it requires fewer imports and unsafe blocks. I prefer it this way.

ltratt: OK.

vext01:
bors try

vext01:
bors try-

vext01:
bors try

ltratt:
bors try-

PR #575, opened 2022-10-07
Make JITted trace code debuggable.

vext01:
This change adds an environment variable `YKD_TRACE_DEBUGINFO`, which when set to "1" causes the JIT to add debug locations to the JITMod.

This allows debuggers conforming to the gdb JIT interface to display higher-level representations of the JITted machine code as it is being debugged.

The "source code" displayed is a temporary file generated on the fly. It contains jit-post-opt IR annotated with AOT source code locations.

The temporary file is managed by a `tempfile::NamedTempFile` which is stored in `CompiledTrace`. This means that when we finally get a mechanism to evict old traces, the temp file will be deleted as the compiled trace is dropped.

In order to show good quality output, I've also had to give the blocks in the JITMod meaningful names. This caused some test flux.

Some of the code here is UNIX specific, but I've guarded the sections that would have to be adapted for other OSs.

C++17 is required for `std::filesystem`.

image


ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -17,18 +18,27 @@
...
+#include <unistd.h>
 
 #include "jitmodbuilder.h"
 #include "memman.h"
 
+// When we create a compilation unit for our JIT debug info, LLVM forces us to
+// choose a language from one of those "recognised" by the DWARF spec (see
+// `Dwarf.def` in the LLVM sources). If you choose a value out of range, then an
+// LLVM assertion will fail, so we have to pretend to be another language.
+#define YKJIT_DWARF_LANG dwarf::DW_LANG_Cobol74
```
ltratt, line 40: Is this comment saying "we don't just pick one at random, but we pick one that's highly unlikely to be used in any system we're part of"?

vext01: Yeah, kind of. That and `Cobol74` amuses me.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -40,24 +50,52 @@ struct BitcodeSection {
   size_t len;
 };
 
-// An annotator for `Module::print()` which adds debug location lines.
+// If possible, return a string describing the location of an instruction in
+// the AOT-compiled interpreter source code.
+//
+// Since it is undesirable to show duplicated debug locations, the callee
+// should pass in the last annotation that it saw (in `LastAnnot`), and this
+// function will return `None` if the debug location string was the same as
```
ltratt, line 58: Is `None` the right C++ name for this concept?

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -232,6 +270,97 @@ extern "C" void *compileModule(string TraceName, Module *M,
   return ptr;
 }
 
+/// Write the string `S` in its entirety to the file descriptor `FD`.
+void writeString(int FD, string S) {
+  const char *Buf = S.c_str();
+
+  size_t Remain = strlen(Buf);
+  while (Remain) {
+    assert(Remain >= 0);
```
ltratt, line 279: IMHO this `assert` would better come before the `Remain -= ` line which could (if `write` is silly) underflow. `assert(Wrote <= Remain)` might be one idea?

ykllvmwrap/src/ykllvmwrap.cc
vext01: I believe so: https://llvm.org/doxygen/classllvm_1_1Optional.html

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -232,6 +270,97 @@ extern "C" void *compileModule(string TraceName, Module *M,
...
+  DISubroutineType *ST = DIB.createSubroutineType({});
+  DISubprogram *DS =
+      DIB.createFunction(DF, TraceName, TraceName, DF, 1, ST, 1,
+                         DINode::FlagZero, DISubprogram::SPFlagDefinition);
+  JITFunc->setSubprogram(DS);
+
+  // For each instruction in the trace IR, emit a human-readable version of the
+  // instruction into the temporary file and update the instruction's debug
+  // location to point to this line.
+  size_t LineNo = 1;
```
ltratt, line 330: Does `DILocation::get` start at 1?

ykllvmwrap/src/ykllvmwrap.cc
ltratt: Is the "we shouldn't clash with something else" property important? It's probably worth clarifying in the comment either way.

[I have no attachment to `Cobol74` either way ;)]

ykllvmwrap/src/ykllvmwrap.cc
ltratt: I will admit to being a bit suspicious of this. Does `DILocation` define itself to be 1-indexed?

ykllvmwrap/src/ykllvmwrap.cc
vext01: It doesn't really matter.

 - It's just a field that identifies what language compiler generated the code. I've never seen it displayed outside of `dwarfdump`.
 - There can be no clash. Just as a linker can link objects compiled by the same compiler, so can `ExecutionEngine` load code generated by the same "language" as the AOT code was compiled with.

ltratt: OK.

ykllvmwrap/src/ykllvmwrap.cc
vext01: It's not documented in LLVM. It's just an `unsigned`.

If we suppose that they took precedence from DWARF, then:

> line --  An unsigned integer indicating a source line number. Lines are numbered
beginning at 1. The compiler may emit the value 0 in cases where an
instruction cannot be attributed to any source line.

(from https://dwarfstd.org/doc/DWARF4.pdf)

ltratt: I'm still a bit suspicious, but less so than before!

ltratt:
@ptersilie I'm happy, but this PR mostly affects you!

ykllvmwrap/src/ykllvmwrap.cc
vext01: Please carefully check [fc913f5](https://github.com/ykjit/yk/pull/575/commits/fc913f52c738a6f558cd5330dc891a491542e314).

I realised that `write()` returns `ssize_t`.

ykllvmwrap/src/ykllvmwrap.cc
```diff
@@ -275,16 +275,21 @@ void writeString(int FD, string S) {
...
-    int Wrote = write(FD, Buf, Remain);
+    ssize_t Wrote = write(FD, Buf, Remain);
     if (Wrote == -1) {
       if (errno == EINTR)
         continue;
       else
         err(EXIT_FAILURE, "write");
     }
 
+    assert(static_cast<size_t>(Wrote) <= Remain);
```
ltratt, line 292: Note that if a silly implementation of `write` returns something < -1 then the `static_cast` will (I assume!) throw the signedness away. You could do `debug_assert(Wrote >= 0 && static_cast<size_t>(Wrote) <= Remain)`?

[Dunno if C++ has `debug_assert` but, if so, I'd be fine with using that. I don't think we need to `assert` this in release mode all the time.]

vext01: if `write` returned `-1` we don't get to this portion of code.

In C++ `assert()` is a `debug_assert!()`. I opted to use a hard error with `err()` as we probably want to know if this happens in release mode.

vext01: Oh wait. `< -1`, let me think.

ltratt: Note: I'm suggesting a stronger check that `write` doesn't return a value lower than `-1`. If that happened, the `static_cast<size_t>` would just remove that information, which would be confusing.

vext01: I'll attempt a fix, but if a platform doesn't do POSIX, then you probably have other issues :P

ltratt: I think `assert(Wrote >= 0 && static_cast<size_t>(Wrote) <= Remain)` is easy to understand and catches the < -1 horror?

vext01: Fixed in fc913f5..4533234

ltratt: Genius code :p

yktrace/src/lib.rs
```diff
@@ -137,7 +141,32 @@ impl IRTrace {
...
 
-    pub fn compile(&self) -> Result<*const c_void, Box<dyn Error>> {
+    // If necessary, create a temporary file for us to write the trace's debugging "source code"
+    // into. Elsewhere, the JIT module will have `DebugLoc`s inserted into it which will point to
+    // lines in this temporary file.
+    //
+    // If the `YKD_TRACE_DEBUGINFO` environment variable is set to "1", then this function returns a
+    // `NamedTempFile`, a non-negative file descriptor, and a path to the file.
+    //
+    // If the `YKD_TRACE_DEBUGINFO` environment variable is *not set to "1", then no file is
```
ptersilie, line 151: s/*not/*not*?

ptersilie:
I can't say I understood all of the details in here, but the results speak for themselves. Looking forward to using this!

yktrace/src/lib.rs
vext01: Fixed in e6a912f

ltratt:
Please squash.

vext01:
splat.

PR #577, opened 2022-10-10
Fix (and check) building without `yk_jitstate_debug` works.

vext01:
See individual commits.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -59,6 +59,13 @@ clang-format --version
 cargo xtask cfmt
 git diff --exit-code
 
+# Check that building `ykcapi` in isolation works. This is what we'd be doing
+# if we were building release binaries, as it would mean we get a system
+# without the (slower) `yk_testing` and `yk_jitstate_debug` features enabled.
+for mode in "" "--release"; do
+    cargo build ${mode} -p ykcapi;
+done
+
 cargo test
 cargo test --release
 cargo bench
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -27,7 +27,7 @@ rustup toolchain install nightly --allow-downgrade --component rustfmt
 # are not enabled.
 for mode in "" "--release"; do \
     cargo -Z unstable-options build ${mode} --build-plan -p ykcapi | \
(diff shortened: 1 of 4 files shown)


ltratt:
The commit message for https://github.com/ykjit/yk/pull/577/commits/9372f54b8efb9c89dfbec30dd1893674b621e08c makes my head hurt. Can we rephrase it? [Obviously only at squashing time!]

vext01:
How about just:

```
This fixes two related things:

 - If you were to do `cargo build -p ykcapi`, then you'd get build error
   because one import of (and call to) `print_jit_state()` wasn't
   properly guarded.

 - After fixing the above, some imports become redundant when the
   feature is off, and thus also need to be guarded.
```

ltratt:
Does "If you were to do" mean "Before this commit this was the behaviour"?

vext01:
That's right.

ltratt:
Can we be clearer about that? Even just to say "Previously, `cargo build -p ykcapi` caused a build error since a `print_jit_state()` was accidentally compiled in." or similar?

ltratt:
"get A build error" -- otherwise, works for me!

Annoyingly, one comment I made elsewhere has got stuck in github's interface (it says "pending" but...) so I'll paste it here: line 36 of `.buildbot.sh` has, I think, pointless `\` chars?

vext01:
Fine by me.

OK to squash and fix that then?

ltratt:
Let's push a commit for the fix (it's a habit I like :)) and then I'm sure we can soon squash!

vext01:
You mean push an empty commit with the corrected commit message?

vext01:
Ah, i see what you mean, sorry. Try [4e61a0b](https://github.com/ykjit/yk/pull/577/commits/4e61a0be96334baf9c9c7f0ccfcf69bc91ab7ee5)

ltratt:
Please squash.

vext01:
splat.

ltratt:
Please squash (also you probably want to rebase against master at the same time).

vext01:
done.

PR #578, opened 2022-10-10
Stack reconstruction after guard failure.

ptersilie:
It's finally here. The long awaited stack reconstruction PR. It's not perfect, there's likely still bugs, and I'm sure some of the comments could be improved. But I've been dragging this branch around for way to long, so it's time to get it in and iterate on top of that. Naturally, this passes all the tests, and we can even run a simple Lua program.

There's a companion PR coming too which adds prologue information (e.g. callee saved registers) to stackmaps and is needed by this PR. But I haven't finished cleaning that up yet, so we can start working through the ~800 changes in this PR while we wait. I'm sure there'll be loads to discuss and unpack here. Enjoy!

Companion PR: https://github.com/ykjit/ykllvm/pull/48


ltratt:
Is there any use in reviewing the 4 commits separately?

ptersilie:
Probably not. Want me to squash first?

ptersilie:
Squashed.

tests/c/debug_debuginfo.c
```diff
@@ -3,7 +3,7 @@
 //   env-var: YKD_SERIALISE_COMPILATION=1
 //   stderr:
 //     ...
-//     define i8 @__yk_compiled_trace_0(...
+//     define ptr @__yk_compiled_trace_0(...
```
vext01, line 6: Why did the trace return type change again?

tests/c/interp_inside_call2.c
```diff
@@ -12,6 +12,7 @@
 //     jit-state: enter-stopgap
 //     ...
 //     jit-state: exit-stopgap
+//     jit-state: exit-jit-code
```
vext01, line 15: Should the `jit-state: enter-stopgap` and `jit-state: exit-stopgap` states be replaced by a single `jit-state: deoptimise`?

tests/c/debug_debuginfo.c
vext01: A later change explains, I think:

>  Returns the address of a reconstructed stack or null if there wasn't a guard failure.

ykcapi/src/lib.rs
```diff
@@ -41,22 +41,22 @@ pub extern "C" fn yk_mt_control_point(_mt: *mut MT, _loc: *mut Location) {
 }
 
 // The "real" control point, that is called once the interpreter has been patched by ykllvm.
+// Returns the address of a reconstructed stack or null if there wasn#t a guard failure.
```
vext01, line 44: wasn#t

tests/c/interp_inside_call2.c
ptersilie: Not sure what advantage this would have. We know we are deoptimising when we see `enter-stopgap/exit-stopgap`.

ykcapi/src/lib.rs
vext01: How does the reconstructed stack get used? It seems that a pointer to it is returned directly to the caller. What does the caller then do with it?

ykcapi/src/lib.rs
```diff
@@ -147,20 +147,70 @@ pub struct LiveAOTVals {
...
+pub extern "C" fn __ykrt_reconstruct_frames(
+    newframesptr: *const c_void,
+) {
+    unsafe {
+        asm!(
+            // The first 8 bytes of the new frames is the size of the map, needed for copying it
+            // over. Move it into rdx.
+            "mov rdx, [rdi]",
+            // Then adjust the address to where the new stack actually starts.
+            "add rdi, 8",
```
vext01, line 164: Not sure I understand this bit. rdi containst the size of the map, which we then add 8 to, but the second instruction is talking about an address. What am i missing?

ykcapi/src/lib.rs
```diff
@@ -147,20 +147,70 @@ pub struct LiveAOTVals {
...
+            "pop r10",
+            "pop r9",
+            "pop r8",
+            "pop rbp",
+            "pop rdi",
+            "pop rsi",
+            "pop rbx",
+            "pop rcx",
+            "pop rdx",
+            "pop rax",
```
vext01, line 193: Can we put a FIXME in here to remind us to restore floating point registers at some stage?

ykcapi/src/lib.rs
```diff
@@ -178,9 +228,9 @@ pub extern "C" fn yk_stopgap(
     // Restore saved registers from the stack.
     let registers = Registers::from_ptr(rsp);
 
-    let mut sginterp = unsafe { SGInterp::new(activeframes, retvalptr) };
+    let mut sginterp = unsafe { SGInterp::new(activeframes, frameaddr) };
```
vext01, line 231: For some reason I was expecting yksgi to disappear. Are we keeping it?

ykcapi/src/lib.rs
```diff
@@ -218,6 +268,7 @@ pub extern "C" fn yk_stopgap(
             SMLocation::Indirect(reg, off, size) => {
                 let addr = unsafe { registers.get(*reg) as *mut u8 };
                 let addr = unsafe { addr.offset(isize::try_from(*off).unwrap()) };
+                //println!("direct addr: {:?}", addr);
```
vext01, line 271: debugging left in here.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -1448,8 +1441,9 @@ class JITModBuilder {
     // the trace.
     assert(!isOutlining());
 
-    Builder.CreateRet(ConstantInt::get(Type::getInt8Ty(JITMod->getContext()),
-                                       TRACE_RETURN_SUCCESS));
+    // If the trace succeeded return a nullpointer instead of a reconstructed
```
vext01, line 1444: nitpicking: `nullpointer` -> `null pointer`. Sorry :P

yksgi/Cargo.toml
```diff
@@ -9,7 +9,15 @@ license = "Apache-2.0 OR MIT"
 libc = "0.2.97"
 libffi = "3.0.0"
 ykutil = { path = "../ykutil" }
+yksmp = { path = "../yksmp" }
 llvm-sys = "140"
+memmap2 = "0.5.2"
+capstone = "0.11.0"
+
+[dependencies.object]
```
vext01, line 17: What is `object` needed for?

yksgi/src/lib.rs
```diff
@@ -8,19 +8,25 @@ use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::ffi::{c_void, CStr};
 use std::ptr;
+
```
vext01, line 11: Any comments on this line @ltratt? :)

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
         }
     }
 
+    // Generate frames from stackmap information after a guard failure. The new frames are stored
+    // inside some allocated memory whose pointer this function returns. The frames are then later
+    // `memcpy`ed to the actual stack by ykcapi::__ykrt_reconstruct_frames.
```
vext01, line 173: another nitpick: backtick `ykcapi::__ykrt_reconstruct_frames`

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
...
+    pub fn reconstruct_frames(&self, btmframeaddr: *mut c_void) -> *const c_void {
+        // Load the stackmap and text sections from the binary. The former we need to parse in the
+        // stackmaps. The latter is required to disassemble the binary in order to distinguish
+        // branches from calls.
+        // FIXME: Don't use current_exe.
```
vext01, line 178: OK, this is where object is used. We shouldn't need to look into the binary file for either `.text` or the stackmap section. They are both loaded now. This could be a separate PR if @ltratt is OK with that.

We might also be able to avoid instantiating capstone using some magic. What do you need? the size of the instruction in the AOT code which could become a guard failure in the JIT code?

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
...
+        let exemmap = unsafe { memmap2::Mmap::map(&file).unwrap() };
+        let object = object::File::parse(&*exemmap).unwrap();
+        let sec = object.section_by_name(".llvm_stackmaps").unwrap();
+        let text = object.section_by_name(".text").unwrap();
+
+        // Create a capstone dissassembler.
+        // so we can figure out the size of the instruction we
+        // return too in order to adjust the return address in our reconstructed stackmap.
+        let cs = Capstone::new()
+        .x86()
```
vext01, line 190: weird indent.

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
...
+        let pathb = env::current_exe().unwrap();
+        let file = fs::File::open(&pathb.as_path()).unwrap();
+        let exemmap = unsafe { memmap2::Mmap::map(&file).unwrap() };
+        let object = object::File::parse(&*exemmap).unwrap();
+        let sec = object.section_by_name(".llvm_stackmaps").unwrap();
+        let text = object.section_by_name(".text").unwrap();
+
+        // Create a capstone dissassembler.
+        // so we can figure out the size of the instruction we
+        // return too in order to adjust the return address in our reconstructed stackmap.
```
vext01, line 188: too -> to

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
...
+        let mut rsp = rbp;
+        // Keep track of the addresses of the current and previous frame so we can update the RBP
+        // register as needed.
+        let mut currframe = btmframeaddr;
+        let mut nextframe = btmframeaddr;
+
+        for (i, frame, rec, pinfo) in smaps {
+            debug_assert!(rec.size != u64::MAX);
+
+            // WRITE RBP
```
vext01, line 271: WRITE -> Write

ykcapi/src/lib.rs
```diff
@@ -278,7 +329,8 @@ pub extern "C" fn __llvm_deoptimize(
     // stackmap).
     unsafe {
         asm!(
-            // Save registers to the stack.
+            // Save registers the may be referenced by the stackmap to the stack before they get
```
ltratt, line 332: Sentence doesn't parse. I'm not sure if it's supposed to be "THAT may be referenced..."?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -1312,6 +1303,9 @@ class JITModBuilder {
                 (IID == Intrinsic::lifetime_end))
               continue;
 
+            if (IID == Intrinsic::frameaddress)
```
ltratt, line 1306: Why do we skip this one?

yksgi/src/lib.rs
ltratt: I think... let me check... yes, this is extraneous space isn't it?

ltratt: Being serious, we can collapse these into `use std::{convert.., ffi::.., }` and so on rather than having Java-esque multi-line `use`s.

yksgi/src/lib.rs
```diff
@@ -98,6 +104,19 @@ signext!(sext_i8, i8);
 signext!(sext_i16, i16);
 signext!(sext_i32, i32);
 
+unsafe fn get_stackmap_call(pc: Value) -> Value {
```
ltratt, line 107: Why is the whole function `unsafe`?

vext01:
Good work @ptersilie. I'm willing to bet this is the scariest low-level code you've ever written to date!

yksgi/Cargo.toml
vext01: I answered this one myself in a later comment.

yksgi/src/lib.rs
ltratt: These should also be `///` doc string quotes. And then `[ykcapi::__ykrt_reconstruct_frames]` (i.e. put it in backticks) and rustdoc will  make it a clickable link! [I don't know, but it wouldn't surprise me if this also allows editors to make this a followable link.]

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
...
+        // Vec holding currently active register values.
+        let mut registers = vec![0; 16];
+
+        // Parse the stackmap.
+        let slice = unsafe { slice::from_raw_parts(sec.address() as *mut u8, usize::try_from(sec.size()).unwrap()) };
+        let smentries = StackMapParser::get_entries(slice);
+
+        // The final size of the memory we need to allocate. Inititialised with space to store
+        // registers for register recovery.
+        let mut memsize: usize = 15 * 8;
```
ltratt, line 206: 8 -> `sizeof::<*const ()>` or similar.

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
         }
     }
 
+    // Generate frames from stackmap information after a guard failure. The new frames are stored
+    // inside some allocated memory whose pointer this function returns. The frames are then later
+    // `memcpy`ed to the actual stack by ykcapi::__ykrt_reconstruct_frames.
+    pub fn reconstruct_frames(&self, btmframeaddr: *mut c_void) -> *const c_void {
```
ltratt, line 174: This function currently seems very x86 only. Should it be guarded as such?

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
...
+                        pinfo = Some(&entry.pinfo);
+                        rec = Some(r);
+                        break;
+                    }
+                }
+            };
+            let rec = rec.unwrap();
+            let pinfo = pinfo.unwrap();
+            if i > 0 {
+                // We don't need to include the bottom-most frame, which is the frame containing
```
ltratt, line 232: I don't understand how the comment relates to the `memsize +=`.

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
...
+                unsafe { ptr::write(rsp as *mut u64, currframe as u64) };
+            }
+
+            // Now that we've (potentially) written the last frame's address, update currframe to
+            // the actual current frame.
+            currframe = nextframe;
+
+            // Update RBP to represent this frame's address.
+            if pinfo.hasfp {
+                registers[6] = currframe as u64;
```
ltratt, line 286: Is this cast always safe?

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
...
+            // AOT stackmap call.
+            let smcall = unsafe { get_stackmap_call(frame.pc) };
+            for (j, lv) in rec.live_vars.iter().enumerate() {
+                // Adjust the operand index by 2 to skip stackmap ID and shadow bytes.
+                let op = smcall.get_operand(u32::try_from(j+2).unwrap());
+                let val = if let Some(val) = frame.get(&op) {
+                    val.val
+                } else {
+                    todo!()
+                    //println!("probably return pointer which will disappear");
```
ltratt, line 324: Commented code (you can put text in `todo` if you want e.g. `todo!("probably return a pointer which will disappear")`).

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
...
+                        registers[usize::from(*reg)] = val;
+                    }
+                    SMLocation::Direct(reg, off, _) => {
+                        if i == 0 {
+                            // skip first frame
+                            continue;
+                        }
+                        debug_assert!(op.is_alloca());
+
+                        // The sizes reported by the stackmap aren't always correct. But we can get
```
ltratt, line 348: Shouldn't we fix stackmaps then? [Not in this PR, but in general.]

vext01: I'm not sure it's worth the effort right now to be honest.

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
...
+                        // the correct size from the IR.
+                        let eltype = unsafe { LLVMGetAllocatedType(op.get()) };
+                        let size = i64::try_from(unsafe { LLVMABISizeOfType(layout, eltype)}).unwrap();
```
ltratt, line 351: Do we need a signed size here?

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
...
+                rsp = unsafe { rbp.offset(-isize::try_from(rec.size).unwrap())};
+                if pinfo.hasfp {
+                    // The stack size recorded by the stackmap includes a pushed RBP. However, we
+                    // will have already adjusted the "virtual RSP" at the beginning if `hasfp` is
+                    // true. If that's the case, re-adjust the "virtual RSP" again to account for
+                    // this.
+                    rsp = unsafe { rsp.offset(8)};
+                }
+            }
+            // WRITE RETURN ADDRESS.
```
ltratt, line 389: Not sure we need this shouty line :)

yksgi/src/lib.rs
```diff
@@ -547,6 +831,7 @@ impl SGInterp {
             None
         } else {
             let op = self.pc.get_operand(0);
+            println!("ret op: {:?}", op.as_str());
```
ltratt, line 834: Dugging print.

yksmp/src/lib.rs
```diff
@@ -14,11 +14,15 @@ use std::error;
 struct Function {
     addr: u64,
     record_count: u64,
+    stack_size: u64,
```
ltratt, line 17: Can we not make this `usize` and de-x86 it?

yksmp/src/lib.rs
```diff
@@ -14,11 +14,15 @@ use std::error;
...
 }
 
-struct Record {
-    offset: u32,
-    live_vars: Vec<LiveVar>,
+pub struct Record {
```
ltratt, line 20: Similarly these are all very x86 centric, probably unnecessarily so?

ykcapi/src/lib.rs
ptersilie: We first copy the size of the map into `rdx`: `mov rdx, [rdi]`.
Then we add 8 to move the pointer of `rdi` to the actual beginning of the new stack. We then use that adjusted `rdi` as an argument for `memcpy` in order to copy over the new stack frames.

ykcapi/src/lib.rs
ptersilie: All the initialisation code from SGI is still needed for stack reconstruction as well. I'm also a bit hesitant to remove SGI entirely as it might be useful in the future for debugging purposes. I imagined that at some point we have a switch to toggle stack reconstruction/stopgap interpreter.

ykcapi/src/lib.rs
vext01: ah, i see what's gone wrong. I thought you were adding to rdx. My bad.

ykcapi/src/lib.rs
ptersilie: Sorry, yes. It's supposed to be `Save registers that`. Will fix.

ykcapi/src/lib.rs
ptersilie: This is also explain better in the commit message. The caller checks the value of the return and if it's not null, then it jumps to `__ykrt_reconstruct_frames` which copies over the new stack.

yksgi/src/lib.rs
ptersilie: I haven't formatted yet. So hopefully the formatter will pick that up. But will double check it does.

yksgi/src/lib.rs
ptersilie: Will fix.

yksgi/src/lib.rs
ptersilie: Yes I probably need @vext01 help to get rid of the `current_exe`. Will do a follow up PR immediately once this is in.

Regarding the capstone thing. I need to size of the instruction before stackmap. Which may be a `call` or an `icmp` or `test`. If it's a call, I need its size in order to adjust the record offset.

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
...
+
+        // Now that we've calculated the stack's size, allocate it using mmap.
+        let mmap = unsafe {
+            libc::mmap(std::ptr::null_mut(), memsize, libc::PROT_READ | libc::PROT_WRITE, libc::MAP_PRIVATE | libc::MAP_ANONYMOUS, -1, 0)
+        };
+
+        // Get the modules layout which we'll need to extract type sizes of LLVM IR.
+        let layout = self.module.datalayout();
+
+        // Generate and write frames to the mmap. Since the stack grows downwards and we need to
```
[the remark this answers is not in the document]
ptersilie: If it isn't already I will guard it.

yksgi/src/lib.rs
```diff
@@ -149,6 +168,271 @@ impl SGInterp {
...
+                    let mut tmp = unsafe { rbp.offset(isize::try_from(idx * 8).unwrap()) };
+                    if pinfo.hasfp {
+                        tmp = unsafe { tmp.offset(8) };
+                    }
+                    let val = registers[usize::from(*reg)];
+                    unsafe { ptr::write(tmp as *mut u64, val) };
+                }
+            }
+
+            // WRITE STACKMAP LOCATIONS.
```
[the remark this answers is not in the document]
ptersilie: I did the caps on purpose to divide this code into sections and make them easier to find. You don't seem to be a fan.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: This isn't needed in the compiled trace as it's only passed into the control point which goes away as well in a compiled trace.

yksgi/src/lib.rs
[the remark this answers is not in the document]
vext01: I don't mind as long as it is consistent.

yksgi/src/lib.rs
ptersilie: Tbh I'm still slightly confused about `unsafe` and where to use it. I think my reasoning was that depending on which value of pc you pass in this might not work. So it's on the caller to make sure the argument is safe.

yksgi/src/lib.rs
ptersilie: Would it help if it was above the condition? The comment explains why we skip `i=0`.

yksgi/src/lib.rs
ptersilie: On x86_64 yes. Let me see if I can turn them into usizes.

yksgi/src/lib.rs
ptersilie: Actually, I need to reinvestigate this code here. I think this can be removed entirely now.

yksgi/src/lib.rs
ptersilie: Can allocas have negative sizes?

yksgi/src/lib.rs
ptersilie: I did the caps on purpose to divide this code into sections and make them easier to find. Each caps title writes a different section to the new stack.

yksgi/src/lib.rs
ptersilie: Oh, I see. You are saying we don't need an `i64` here. And I agree.

yksgi/src/lib.rs
ptersilie: [bc9be2b](https://github.com/ykjit/yk/pull/578/commits/bc9be2bb1a42334c625104ce5309e9a7b5f0cb8b)

ykcapi/src/lib.rs
ptersilie: Fixed in [c45e5e6](https://github.com/ykjit/yk/pull/578/commits/c45e5e644b2bc6b43c9d456a8192a5cc3ddbaff8)

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Fixed in [c45e5e6](https://github.com/ykjit/yk/pull/578/commits/c45e5e644b2bc6b43c9d456a8192a5cc3ddbaff8)

yksgi/src/lib.rs
ptersilie: Fixed in [c45e5e6](https://github.com/ykjit/yk/pull/578/commits/c45e5e644b2bc6b43c9d456a8192a5cc3ddbaff8)

yksgi/src/lib.rs
ptersilie: Fixed in [c45e5e6](https://github.com/ykjit/yk/pull/578/commits/c45e5e644b2bc6b43c9d456a8192a5cc3ddbaff8)

ykcapi/src/lib.rs
ptersilie: Fixed in [c45e5e6](https://github.com/ykjit/yk/pull/578/commits/c45e5e644b2bc6b43c9d456a8192a5cc3ddbaff8)

yksgi/src/lib.rs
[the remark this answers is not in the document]
ptersilie: Fixed in [c45e5e6](https://github.com/ykjit/yk/pull/578/commits/c45e5e644b2bc6b43c9d456a8192a5cc3ddbaff8)

yksgi/src/lib.rs
ptersilie: Fixed in [c45e5e6](https://github.com/ykjit/yk/pull/578/commits/c45e5e644b2bc6b43c9d456a8192a5cc3ddbaff8)

yksmp/src/lib.rs
ptersilie: Just checked, this is already u64 in LLVM, so we inherit the type here. https://llvm.org/docs/StackMaps.html#stackmap-format

yksmp/src/lib.rs
ptersilie: `size` we later populate with the functions stack size which is also u64 in LLVM's stackmap format.

yksgi/src/lib.rs
ptersilie: I could probably make `registers` store `usize`, but we are storing `SGValue.val` in them, which is `u64`. We had a discussion some time ago that `SGValue` probably shouldn't store `u64`s. But I don't think that's a change for this PR, and I'm also unsure how high on the list this should go.

yksgi/src/lib.rs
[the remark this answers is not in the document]
ptersilie: Fixed in [fb53ebe](https://github.com/ykjit/yk/pull/578/commits/fb53ebedc5a38dfb9dc4e330a6a906fdf4f4c073)

yksgi/src/lib.rs
ptersilie: Fixed in [3e270bc](https://github.com/ykjit/yk/pull/578/commits/3e270bceda6db7073f58c4fd90cea45ea6bcadab)

ptersilie:
Think this is ready for re-review.

tests/c/interp_inside_call2.c
vext01: But we don't stopgap, do we?

ykcapi/src/lib.rs
vext01: I find it confusing that we instantiate a stop gap interpreter, but then don't *really* use it.

yksgi/src/lib.rs
vext01: We may be able to insert something at compile-time, but let's roll with this for now.

Perhaps add a FIXME?

ykcapi/src/lib.rs
ltratt: I tend to agree with Edd. If we do ever need a stopgap interpreter, we'll probably have to rewrite more-or-less from scratch anyway. I'd be inclined to remove it entirely.

yksgi/src/lib.rs
ltratt: The formatter probably won't merge them, but I might be wrong. [Let's also run the formatter before raising the PR :)]

vext01:
The main outstanding thing with this is the confusion around a "stop gap interpreter"

Is it worth pulling the frame initialisation code out of yksgi?

yksgi/src/lib.rs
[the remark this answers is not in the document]
ltratt: I don't like the shouting either :)

ykllvmwrap/src/jitmodbuilder.cc
ltratt: Can we add a quick comment for our future selves?

yksgi/src/lib.rs
ltratt: In some sense *any* function won't work if you call it incorrectly! In our context `unsafe` should mostly mean "if you call this with the wrong arguments the program may exhibit undefined behaviour (inc. but not limited to segfaulting)". Is this one of those functions?

yksgi/src/lib.rs
```diff
@@ -30,6 +30,9 @@ static ALWAYS_SKIP_FUNCS: [&str; 8] = [
 static SKIP_FOR_NOW_FUNCS: [&str; 2] = ["yk_location_drop", "yk_mt_drop"];
 static YK_CONTROL_POINT_FUNC: &str = "__ykrt_control_point";
 
+static POINTER_USIZE: usize = std::mem::size_of::<*const ()>();
```
ltratt, line 33: Slightly confusing variable name. `SIZEOF_POINTER` would be better. [Personally I'd just `use std::mem::size_of` and then always use `size_of::<*const ()>` in the code. I don't care too much though.]

yksgi/src/lib.rs
ltratt: I was thinking the other way around. I understand that on x86_64 the registers can be treated as `u64`: I just wasn't entirely sure if `currframe` could be case to a `u64` in all cases or not. [I'm not sure what type `currframe` has. Knowing that might make it obvious.]

yksgi/src/lib.rs
ltratt: Maybe we should raise an issue? At the moment it looks like we're doing work at run-time that we *could* semi-easily shift to AOT.

yksgi/src/lib.rs
ltratt: What type does `LLVMABISizeOfType` return? I wonder if the `try_from` is even needed now.

yksmp/src/lib.rs
ltratt: OK.

ltratt:
I've done a round of commenting / resolving, including comments that @vext01 raised that I was trivially able to confirm are resolved.

yksgi/src/lib.rs
ptersilie: I would still need to have a constant as an `isize` tho unless I want to do `isize::try_from(size_of::<*const ()>).unwrap()` everywhere, which IMO distracts from the code quite a bit.

ptersilie: Can I do two constants: `ISIZEOF_POINTER` and `USIZEOF_POINTER`?

ptersilie:
> The main outstanding thing with this is the confusion around a "stop gap interpreter"
> 
> Is it worth pulling the frame initialisation code out of yksgi?

Not sure this is easily done. All the initialisation stuff on yksgi (e.g. `var_init`) is needed for stack reconstruction. So unless `yksgi` provides an API so lookup variables I don't think this is easy to pull out.

yksgi/src/lib.rs
ptersilie: Good spot. It's `c_ulonglong` which means we don't need the `try_from`.

Fixed in [cfe5de7](https://github.com/ykjit/yk/pull/578/commits/cfe5de756bc0bd05c3f44dc7a0298e3a5ad1233a)

yksgi/src/lib.rs
ltratt: I prefer size_of, but it's something upon which reasonable people can differ. So long as we change the variable name to something less confusing, I'm OK with a variable.

yksgi/src/lib.rs
ptersilie: Got you. So `rec.size` is the size of each stack frame. In order to calculate how much memory we need to allocate for out new stack, we add the frame sizes of all frames together. However, the stackmap (which we iterate over) includes the bottom frame (the caller of the control point). But since that frame already exists, we don't need to allocate space for it. Does that clear things up?

yksgi/src/lib.rs
ltratt: > Can I do two constants: ISIZEOF_POINTER and USIZEOF_POINTER?

If you need them, of course!

yksgi/src/lib.rs
ptersilie: See if [1f2afe8](https://github.com/ykjit/yk/pull/578/commits/1f2afe8397349af56993f7de0c1e1152aa212eac) works better for you.

ltratt:
@ptersilie I think we should keep this PR as-is, but maybe an upcoming PR can rip out the bits of the stopgap interpreter we're never likely to need again.

yksgi/src/lib.rs
ptersilie: Not sure about segfaulting. I'm inheriting the unsafe from the llvm-sys calls (e.g. `LLVMGetPreviousInstruction`). Not entirely certain how you can get them to segfault but I imagine handing them a nullptr would do the trick.

So as long as we don't do that we should mark the LLVM-calls unsafe instead of the whole function?

tests/c/interp_inside_call2.c
ptersilie: I misunderstood you. You want to rename it. I guess we can. Is what we are doing now not stopgapping anymore? Just looking at the definition of stopgap I would still say it fits.

yksgi/src/lib.rs
ptersilie: Yeah, I may have rushed this PR out a bit to quickly, but it was late last night and I wanted to wrap it up.

yksgi/src/lib.rs
vext01: I'd say roll with capstone and deal with a better way in a follow up PR.

tests/c/interp_inside_call2.c
vext01: Regardless of what you call it, whereas before stopgapping was an interval of time (in which we happened to be running an interpreter), now it's an instantaneous event. We flick from JITted code back to the interpreter without any intermediate "user program execution". So to start/stop stopgapping makes little sense in my head.

And I do still think it should be called "deoptimise" and not "stopgap". The name "stopgap" implies a temporary compromise, which no longer conveys what we are really doing any more.

ykcapi/src/lib.rs
vext01: There's always the git history.

yksgi/src/lib.rs
vext01: https://github.com/ykjit/yk/issues/579

tests/c/interp_inside_call2.c
ltratt: Edd's summed up my position well.

ykcapi/src/lib.rs
ltratt: Agreed. I'm going to resolve this particular comment but let's assume that we rip out most of the SGI in an upcoming PR.

yksgi/src/lib.rs
ltratt: I seem to be saying this a lot, but I agree with Edd.

yksgi/src/lib.rs
ltratt: Got it!

yksgi/src/lib.rs
ltratt: Thanks!

ltratt:
I've done another round of resolving -- we're getting there!

tests/c/interp_inside_call2.c
ptersilie: I'm not attached to stopgapping and get your point. So I'm happy to rename it.

ltratt: Correct: "guard failure" / "deoptimisation" (they're the same thing for us) is effectively a discreet event.

ptersilie: Well this wasn't fun. [303a233](https://github.com/ykjit/yk/pull/578/commits/303a23327920839ec2c4f03f89b9c7c455029e3d)

ltratt: But now it's done :)

ykcapi/src/lib.rs
ptersilie: Fixed in [7660a3c](https://github.com/ykjit/yk/pull/578/commits/7660a3c9813795998bfb6a958058c2d30b60fc9f)

yksgi/src/lib.rs
ptersilie: I concur. This will likely require a ykllvm change too. So better for another time.

ptersilie: Added a fixme: [9294cdb](https://github.com/ykjit/yk/pull/578/commits/9294cdb1fea1440e857e172f981b9a5886501ac2)

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: Fixed in [f689da1](https://github.com/ykjit/yk/pull/578/commits/f689da1d385791e1c7a821174b1712ddb0dd18b9)

yksgi/src/lib.rs
ptersilie: The `get_stackmap_call` isn't checking any preconditions. I don't know what might result in a segfault or UB in `LLVMGetPreviousInstruction`, so I wouldn't know what to check for. So maybe marking the function as `unsafe` is fine here?

ptersilie: Okay, I had a poke around in the LLVM API sources to see where there is potential for UB. It looks like as long as you pass the API the correct type its expecting things should be fine. So I've added some extra asserts to make sure we are not passing nonsense and removed the unsafe from the function. Does that sound ok? See [5993877](https://github.com/ykjit/yk/pull/578/commits/599387756b3ae04da9bccab4d02c586b72e52a7a)

yksgi/src/lib.rs
ptersilie: I think I'd prefer two as having them as constants makes the main code much more readable. Renamed them in [5993877](https://github.com/ykjit/yk/pull/578/commits/599387756b3ae04da9bccab4d02c586b72e52a7a)

yksgi/src/lib.rs
ptersilie: Does this explanation resolve this for you?

ptersilie:
Think this might be ready for a (hopefully final) re-review.

yksgi/src/lib.rs
ltratt: Perfect!

yksgi/src/lib.rs
ltratt: Wrong commit?

ltratt: Aha it should be this one https://github.com/ykjit/yk/pull/578/commits/90ce72e5904bc878733751e61e90ac82f839d302

yksgi/src/lib.rs
```diff
@@ -278,7 +278,7 @@ impl SGInterp {
             // If the current frame has pushed RBP we need to do the same (unless we are processing
             // the bottom-most frame).
             if pinfo.hasfp && i > 0 {
-                rsp = unsafe { rsp.offset(-POINTER_ISIZE) };
+                rsp = unsafe { rsp.offset(-ISIZEOF_POINTER) };
```
ltratt, line 281: Isn't this `rsp.sub(USIZEOF_POINTER)` in disguise? [I tend to prefer avoiding negative `isize`s because we inevitably end up having usize <-> isize conversions and those are surprisingly easy to get wrong.]

yksgi/src/lib.rs
ltratt: Let's manually merge the `use`s now and be done with it?

yksgi/src/lib.rs
ptersilie: I didn't know there was a `sub` call. :man_facepalming: Fixed in cf3e4be3b4da5c85522c35649d24fd7e155dd76a.

Sadly this (in combination with add) doesn't allow us to get rid of the ISIZE_POINTER constant, since we still have some places where we get passed an isize offset.

yksgi/src/lib.rs
ptersilie: Fixed in 3724ca6647067bc052cb351907bdff5dd114d2e1.

yksgi/src/lib.rs
ptersilie: Fixed in 0c4cc76c6098d90ce398f5ea2e4463c5b8536a8b

yksgi/src/lib.rs
ptersilie: Oh sorry.

ptersilie:
Also found a bug which I introduced in [cf3e4be](https://github.com/ykjit/yk/pull/578/commits/cf3e4be3b4da5c85522c35649d24fd7e155dd76a). Fixed in [0c4cc76](https://github.com/ykjit/yk/pull/578/commits/0c4cc76c6098d90ce398f5ea2e4463c5b8536a8b)

ptersilie:
@ltratt back to you good Sir!

yksgi/src/lib.rs
ltratt: OK.

ltratt:
Please squash.

ptersilie:
Sorry, another cfmt before I can squash: [4160934](https://github.com/ykjit/yk/pull/578/commits/4160934153311668fe55f1ec40b479678037a5d2)

ltratt:
Please squash.

ptersilie:
Squashed. But we can't merge until https://github.com/ykjit/ykllvm/pull/48 is in. But it's getting late now. So maybe we should continue tomorrow.

vext01:
Is this ready to go now?

ptersilie:
Yes. Let's merge!

ptersilie:
Sorry. This needs another squash.

ltratt:
Please squash (and rebase against master).

ptersilie:
Squashed and rebased.

PR #585, opened 2022-11-24
Ykpt block binding

vext01:
See individual commit messages. The first commit is the main bulk of work.

Sorry, this isn't as reviewable as I would have liked. A lot of moving code and churn.

Compiler change soon.

### DIFF
--- a/hwtracer/src/decode/ykpt/mod.rs
+++ b/hwtracer/src/decode/ykpt/mod.rs
@@ -1,9 +1,28 @@
 //! The Yk PT trace decoder.
 
-use crate::{decode::TraceDecoder, errors::HWTracerError, Block, Trace};
+use crate::{
+    decode::TraceDecoder,
+    errors::HWTracerError,
+    llvm_blockmap::{BlockMapEntry, SuccessorKind, LLVM_BLOCK_MAP},
+    Block, Trace,
+};
+use intervaltree;
+use std::{
+    collections::VecDeque,
+    convert::TryFrom,
+    fmt::{self, Debug},
+    path::PathBuf,
+};
+use ykutil::{
+    addr::{code_vaddr_to_off, off_to_vaddr},
+    obj::SELF_BIN_PATH,
+};
 
 mod packet_parser;
-use packet_parser::PacketParser;
+use packet_parser::{
+    packets::{Packet, PacketKind},
+    PacketParser,
+};
 
 pub(crate) struct YkPTTraceDecoder {}
 
@@ -19,8 +38,29 @@ impl TraceDecoder for YkPTTraceDecoder {
         let itr = YkPTBlockIterator {
             errored: false,
             parser: PacketParser::new(trace.bytes()),
+            cur_loc: ObjLoc::OtherObjOrUnknown,
+            tnts: VecDeque::new(),
+            pge: false,
         };
         Box::new(itr)
+    }
+}
+
+/// Represents a location in the instruction stream of the traced binary.
+#[derive(Eq, PartialEq)]
+enum ObjLoc {
+    /// A known location in the "main binary object" of the program.
+    MainObj(u64),
+    /// Either a location not in the "main binary object", or an unknown location.
+    OtherObjOrUnknown,
+}
+
+impl Debug for ObjLoc {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::MainObj(v) => write!(f, "MainObj[0x{:x}]", v),
+            Self::OtherObjOrUnknown => write!(f, "OtherObjOrUnknown"),
+        }
     }
 }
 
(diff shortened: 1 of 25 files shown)


ltratt:
In the first commit message:

> PT would never tell you about an unconditional jump.

s/would/will/

> For a conditional jumps

s/jumps/jump/

> For a conditional jumps, you can easily infer from the instruction stream the two potential addresses at which execution may continue

I don't think this is true since I can have conditional indirect jumps. If it said "For a conditional direct jump" I would be fine with it.

They can easily be fixed at squashing time.

vext01:
> I don't think this is true since I can have conditional indirect jumps.

I don't think there's such a thing as a conditional indirect jump in X86 is there? (scours the manual...)

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -19,29 +38,232 @@ impl TraceDecoder for YkPTTraceDecoder {
...
+            Block::unknown()
+        }
+    }
+
+    fn do_next(&mut self) -> Result<Option<Block>, HWTracerError> {
+        // Read as far ahead as we can using static successor info encoded into the blockmap.
+        if let ObjLoc::MainObj(b_off) = self.cur_loc {
+            // We know where we are in the main object binary, so there's a chance that there's a
+            // blockmap entry for this location.
+            let ent = Self::lookup_blockmap_entry(b_off.to_owned());
```
ltratt, line 115: Why `Self::` and not `self.`?

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -19,29 +38,232 @@ impl TraceDecoder for YkPTTraceDecoder {
...
+            if ent.is_none() {
+                // In the absence of blockmap info, we cannot know the successor block.
+                self.cur_loc = ObjLoc::OtherObjOrUnknown;
+                return Ok(Some(Block::unknown()));
+            }
+            let ent = ent.unwrap();
```
ltratt, line 121: Even though this `unwrap` is safe now, we should avoid these whenever possible. In this case, I think we can not only do that, but make the code clearer with:

```rust
let ent = match self.lookup_blockmap_entry(...) {
  Some(x) => x,
  None => {
    // In the absence of blah blah
    ...
  }
}
```

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -19,29 +38,232 @@ impl TraceDecoder for YkPTTraceDecoder {
...
+                    }
+                    // Find where to go next based on whether the trace took this branch or not.
+                    let target_off = if self.tnts.pop_front().unwrap() {
+                        *taken_target
+                    } else {
+                        if let Some(ntt) = not_taken_target {
+                            *ntt
+                        } else {
+                            // Divergent control flow.
+                            panic!();
```
ltratt, line 173: `panic` or `todo`?

hwtracer/src/decode/ykpt/mod.rs
vext01: it doesn't access any of the instance's state, so can be static.

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -19,29 +38,232 @@ impl TraceDecoder for YkPTTraceDecoder {
...
+            }
+        }
+        Ok(true)
+    }
+
+    /// Keep decoding packets until we receive a TIP update.
+    fn seek_tip(&mut self) -> Result<Option<Block>, HWTracerError> {
+        debug_assert_eq!(self.cur_loc, ObjLoc::OtherObjOrUnknown);
+        loop {
+            if let Some(_) = self.packet()? {
```
ltratt, line 207: This should either be `if self.packet()?.is_some()` or a `match` IMHO.

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -19,29 +38,232 @@ impl TraceDecoder for YkPTTraceDecoder {
...
+                return Ok(None); // No more packets.
+            }
+        }
+    }
+
+    /// Fetch the next packet and update iterator state.
+    fn packet(&mut self) -> Result<Option<Packet>, HWTracerError> {
+        let ret = if let Some(pkt_or_err) = self.parser.next() {
+            let pkt = pkt_or_err?;
+            if pkt.kind() == PacketKind::TIPPGE {
```
ltratt, line 221: `match pkt.kind()` instead of two `if`s?

hwtracer/src/decode/ykpt/mod.rs
vext01: In this case I disagree. I dislike introducing long indented scopes. I find it really gets in the way of comprehension.

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -19,29 +38,232 @@ impl TraceDecoder for YkPTTraceDecoder {
...
+                debug_assert!(!self.pge);
+                self.pge = true;
+            }
+            if pkt.kind() == PacketKind::TIPPGD {
+                debug_assert!(self.pge);
+                self.pge = false;
+            }
+            if let Some(vaddr) = pkt.target_ip() {
```
ltratt, line 229: Can this be true *and* the packet be TIPPGE or TIPPGD?

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -19,29 +38,232 @@ impl TraceDecoder for YkPTTraceDecoder {
...
+                if self.pge {
+                    match code_vaddr_to_off(vaddr) {
+                        Some((obj, off)) if obj == *SELF_BIN_PATH => {
```
ltratt, line 232: Is the path comparison safe (i.e. canonicalised in some sense)?

hwtracer/src/decode/ykpt/mod.rs
vext01: That should be a `todo!()`. My bad.

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -19,29 +38,232 @@ impl TraceDecoder for YkPTTraceDecoder {
...
+                            self.cur_loc = ObjLoc::MainObj(off)
+                        }
+                        _ => self.cur_loc = ObjLoc::OtherObjOrUnknown,
+                    }
+                    self.tnts.clear();
```
ltratt, line 237: Why do we clear?

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -19,29 +38,232 @@ impl TraceDecoder for YkPTTraceDecoder {
...
+                }
+            } else {
+                return Ok(None); // No more packets.
+            }
+        }
+    }
+
+    /// Fetch the next packet and update iterator state.
+    fn packet(&mut self) -> Result<Option<Packet>, HWTracerError> {
+        let ret = if let Some(pkt_or_err) = self.parser.next() {
```
ltratt, line 219: Te `let ret = ` is unnecessary. Remove that and you can also remove the `ret` at the end of the function.

hwtracer/src/decode/ykpt/mod.rs
vext01: Ah, yes, it's a hangover from hacking.

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -19,29 +38,232 @@ impl TraceDecoder for YkPTTraceDecoder {
...
 impl<'t> Iterator for YkPTBlockIterator<'t> {
     type Item = Result<Block, HWTracerError>;
 
     fn next(&mut self) -> Option<Self::Item> {
         if !self.errored {
-            // FIXME: For now this is dummy code to prevent dead-code warnings. Later block binding
-            // logic will go here.
-            self.parser.next().unwrap().unwrap();
+            match self.do_next() {
+                Ok(None) => None,
```
ltratt, line 257: In general I suggest order `match` arms by the expected frequency, as the optimiser is somewhat likely to respect that. So `Ok(Some(blk))` should probably come first.

hwtracer/src/decode/ykpt/mod.rs
vext01: Yes. Both of those packets deliver a TIP update.

image

image

yktrace/src/hwt/mapper.rs
```diff
@@ -271,4 +126,60 @@ impl HWTMapper {
         }
         ret
     }
+
+    /// Obtain the next mapped block.
+    ///
+    /// `last_yielded` must be a reference to the last successfully mapped block returned by the
+    /// iterator (or `None` if this is the first time you are calling `next`).
```
ltratt, line 133: Why don't we cache a reference to `last_yielded` rather than forcing it on the user?

yktrace/src/hwt/mapper.rs
```diff
@@ -271,4 +126,60 @@ impl HWTMapper {
...
+    pub fn next(
+        &mut self,
+        last_yielded: Option<&IRBlock>,
+    ) -> Option<Result<IRBlock, HWTracerError>> {
+        while self.buffered_irblocks.is_empty() {
+            let block = self.trace_iter.next();
+            if block.is_none() {
+                return None;
+            }
+            let block = block.unwrap();
```
ltratt, line 143: `unwrap` is unnecessary if `match` is used.

hwtracer/src/decode/ykpt/mod.rs
vext01: Because at this point any remaining TNTs would be ones for past control flow decisions that we were unable to make use of due to imperfect knowledge of the control flow.

There are two situations I know of where this happens:

 - execution passed through code that wasn't compiled by ykllvm, so we get TNT packets, but no static successor info to use them against.
 -  At the start of a trace, often a TNT packet comes in prior to the `TIP.PGE` packet indicating the start of tracing. Don't ask we why :P

ltratt: Can we add a comment (or, even better, an assertion) for this?

hwtracer/src/decode/ykpt/mod.rs
ltratt: Got it!

vext01:
> I don't think there's such a thing as a conditional indirect jump in X86 is there? (scours the manual...)

There isn't: https://stackoverflow.com/questions/16763460/conditional-jump-to-register

hwtracer/src/decode/ykpt/mod.rs
vext01: For `SELF_BIN_PATH`, yes: https://github.com/ykjit/yk/blob/61715402137f784e8ba8d7bb9505f597e0ebb733/ykutil/src/addr.rs#L29

hwtracer/src/decode/ykpt/mod.rs
vext01: Yes, and I will try :)

hwtracer/src/decode/ykpt/mod.rs
vext01: Debugging left-over :)

yktrace/src/hwt/mapper.rs
vext01: I tried, but failed to get it through the borrow checker. I can try again if you like.

An earlier version used a cloned block, which is awful since it contains a string.

yktrace/src/hwt/mapper.rs
vext01: Same as before: I don't want to indent the remainder of the (already heavily indented) function.

ltratt: Getting rid of `unwrap` (which is dangerous and bloats the generated code) 100% beats indentation worries.

yktrace/src/hwt/mapper.rs
ltratt: Since we're handing out references, I can't see an obvious lifetime issue with this, but I might be wrong. It does feel like it's worth quickly (re)trying.

yktrace/src/hwt/mapper.rs
vext01: Two questions:
 - Why is it dangerous? `block` cannot be `None` at line 143.
 - Why does it bloat the code?

ltratt: `unwrap`s get detached from their surrounding code over time. IMHO every `unwrap` must come with a comment explaining why it's correct, because of this danger.

`unwrap` has to put a branch and carry around the necessary error debugging stuff -- even though *we* know the error can't happen. [Sometimes the compiler might be able to get rid of that, but we can't guarantee that.]

yktrace/src/hwt/mapper.rs
vext01: The problem is, you end up trying to store a reference to something that is immediately about to move.

In the `next()` of the mapper:
```
          ...
          let ret = self.buffered_irblocks.pop().unwrap();                                                                                                                                         
          self.last_yielded = Some(&ret);                                                                                                                                                          
          Some(Ok(ret))
```

This obviously doesn't float with the borrow checker, but did you have a different idiom in mind?

ltratt: Ah, I see that this iterator consumes itself (so to speak). In which case, yes, there's not much we can do here (unless we return something like `Result<&IRBlock, ...>` but that might have other consequences).

vext01: Indeed. The question then becomes: where do the returned blocks then live? The whole purpose of this change is to make the iterator lazy, but to return a reference, we'd be forcing them to be buffered up and collected somewhere anyway...

We may have to suck it up :(

ltratt: Fair enough.

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in 3f6535d0c5dd64637ac2e9f40c0c968a64e05170.

yktrace/src/hwt/mapper.rs
vext01: Fixed in 3f6535d0c5dd64637ac2e9f40c0c968a64e05170

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in e1c20f15a7781308efdffb90052da762bc121a8e

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in 390e15a85418d17488860b63e19252b4b05b6e69

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in a973d4f121354307871e3cdd6c2b2cfbe22d7e18

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in 099ae515c5487c4fda7b7d3f8d5c1ed658fc9b88.

(It's not easy to make an assertion I'm afraid)

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in f115368476ec9f8a99db8e4d4467efe9af7c235d

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in dd3bdd3dbec19c97e64116206b00b5c8ac4789a5

vext01:
I think that's everything

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -19,29 +38,232 @@ impl TraceDecoder for YkPTTraceDecoder {
...
 /// Iterate over the blocks of an Intel PT trace using the fast Yk PT decoder.
 struct YkPTBlockIterator<'t> {
     /// Set to true when an error has occured.
     errored: bool,
-    /// PT packet iterator.
+    /// The packet iterator used to drive the decoding process.
     parser: PacketParser<'t>,
+    /// Keeps track of where we are in the traced binary.
+    cur_loc: ObjLoc,
+    /// A vector of "taken/not-taken" (TNT) decisions. These come in in batches and get buffered
```
ptersilie, line 75: s/in in/in

vext01: "in in" is actually correct, but I will propose a re-wording.

vext01: [1a32fbc](https://github.com/ykjit/yk/pull/585/commits/1a32fbc45255b2d8cb31a96b9e971c480d4789e8)

ltratt: I don't think "in in" is correct, but the rephrasing is fine :)

hwtracer/src/llvm_blockmap.rs
```diff
@@ -0,0 +1,216 @@
...
+    io::{prelude::*, Cursor, SeekFrom},
+    slice,
+    sync::LazyLock,
+};
+
+pub static LLVM_BLOCK_MAP: LazyLock<BlockMap> = LazyLock::new(BlockMap::new);
+
+#[derive(Debug)]
+pub enum SuccessorKind {
+    // FIXME notes on divergence.
```
ptersilie, line 18: Did you mean to leave this FIXME here?

hwtracer/src/llvm_blockmap.rs
```diff
@@ -0,0 +1,216 @@
...
+/// The information for one LLVM `MachineBasicBlock`.
+#[derive(Debug)]
+pub struct BlockMapEntry {
+    /// Function offset.
+    f_off: u64,
+    /// Indices of corresponding BasicBlocks.
+    corr_bbs: Vec<u64>,
+    /// Successor information.
+    succ: SuccessorKind,
+    /// FIXME: doc
```
ptersilie, line 60: And another.

hwtracer/src/llvm_blockmap.rs
vext01: Oops. I was supposed to comment up this struct!

hwtracer/src/decode/ykpt/mod.rs
ltratt: I understand that... but is that differentiation important? When I see `Self::` I wonder if there's an important reason for that.

vext01: Um, not really. It's just good form to not hand out more capabilities than necessary, no?

ptersilie: I must admit I always find it a bit confusing, like there's more to it than you would think.

vext01: It's a bit like not adding `pub` to a method that doesn't need to be exposed.

ptersilie:
I feel like I'm missing some background to fully understand what's going on here, so only a shallow review from me.

hwtracer/src/llvm_blockmap.rs
vext01: Fixed in e137494e3af8f39d2dcb03f2fcfb77e31af7fb8d

hwtracer/src/llvm_blockmap.rs
vext01: Fixed in 1bb4a8c279fe9d7ad4295c13dac7ef71935fe20c

hwtracer/src/decode/ykpt/mod.rs
vext01: What are we doing with this then gents? Do you still want `s/Self/self/`?

ptersilie: Arguably, if `lookup_blockmap_entry` doesn't need access to the instance, then maybe the function should live inside it's `impl`?

vext01: We could move it out, yeah, but it makes sense to logically group it with the iterator, as it won't be used elsewhere.

ltratt: Let's just s/Self/self/ and be done with it.

vext01: Fixed in 1607f68c13a0cf64dd5bad4079cc98beab26eddf

ltratt:
We're now more-or-less good, with the minor `unwrap` issue remaining.

ltratt:
Please squash.

vext01:
splat.

(don't forget the compiler PR needs to go in first)

vext01:
Ah, we need to fix those bits in the commit message... ok for me to do that?

ltratt:
Please squash.

vext01:
splat.

vext01:
bors r-

> (don't forget the compiler PR needs to go in first)

https://github.com/ykjit/ykllvm/pull/50

vext01:
My apologies. That last commit fixes formatting. OK to squash?

ltratt:
Please squash.

vext01:
splat.

PR #587, opened 2022-11-28
Fix a couple of low-hanging ykpt-related bugs.

vext01:
See individual commits.

### DIFF
--- a/tests/hwtracer_ykpt/void_ret.c
+++ b/tests/hwtracer_ykpt/void_ret.c
@@ -1,0 +1,21 @@
+// Run-time:
+
+#include <stdio.h>
+#include <yk_testing.h>
+
+void __attribute__((noinline)) f() {
+  fputs("inside f\n", stderr);
+  return;
+}
+
+int main(void) {
+  int i = 4;
+  NOOPT_VAL(i);
+
+  void *tc = __hwykpt_start_collector();
+  while (i-- > 0)
+    f();
+  void *trace = __hwykpt_stop_collector(tc);
+
+  __hwykpt_libipt_vs_ykpt(trace);
+}
--- a/tests/src/hwtracer_ykpt.rs
+++ b/tests/src/hwtracer_ykpt.rs
@@ -15,7 +15,6 @@ use hwtracer::{
     collect::{TraceCollector, TraceCollectorBuilder, TraceCollectorKind},
     Trace,
 };
-use itertools::{EitherOrBoth::*, Itertools};
 use yktrace::hwt::HWTMapper;
 
 #[no_mangle]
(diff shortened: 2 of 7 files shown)


tests/hwtracer_ykpt/void_ret.c
```diff
@@ -0,0 +1,21 @@
...
+  fputs("inside f\n", stderr);
+  return;
+}
+
+int main(void) {
+  int i = 4;
+  NOOPT_VAL(i);
+
+  void *tc = __hwykpt_start_collector();
+  while (i-- > 0)
```
ltratt, line 16: Isn't this a `for` loop in disguise?

tests/src/hwtracer_ykpt.rs
```diff
@@ -56,22 +56,13 @@ pub extern "C" fn __hwykpt_assert_basic(trace: *mut Box<dyn Trace>) {
...
-
-    for pair in ipt_blks.iter().zip_longest(ykpt_blks) {
-        match pair {
-            Both(expect, got) => assert_eq!(expect, &got),
-            _ => panic!("iterator lengths differ"),
-        }
+    while let Some(got) = ipt_mapper.next(last.as_ref()) {
+        let got = got.unwrap();
+        let expect = ykpt_mapper.next(last.as_ref()).unwrap().unwrap();
+        dbg!(&expect, &got);
```
ltratt, line 62: I don't think the debugging output was supposed to be left in?

vext01: oops :)

tests/hwtracer_ykpt/void_ret.c
vext01: Kind of.

It can't be the for loop we all want it to be:
```
/home/vext01/research/yk/tests/hwtracer_ykpt/void_ret.c:13:16: error: expected expression
  for (int i = NOOPT_VAL(4); i > 0; i--)
               ^
```

`NOOPT_VAL` isn't an expression (it's inline asm), so it hoses that idea.

So, it would have to be this cursed loop instead:
```
int i = 4;
NOOPT_VAL(4);
...
for (; i > 0; i--) {
    ...
```

At which point I find what we started with more concise and clearer:

```
int i = 4;
NOOPT_VAL(4);
...
while (i-- > 0)
    ...
```

tests/src/hwtracer_ykpt.rs
vext01: Fixed in 00787dc1b9d795980e5ec6e28dc01676910da7ff

vext01:
I think that's everything?

ltratt:
Please squash.

vext01:
splat.

PR #588, opened 2022-12-01
Add a (failing) test for "foreign reentrant" code.

vext01:
As requested yesterday :)

### DIFF
--- a/tests/c/reentrant.c
+++ b/tests/c/reentrant.c
@@ -1,0 +1,58 @@
+// ignore: can't trace reentrant code yet. see below.
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_PRINT_IR=jit-pre-opt
+//   env-var: YKD_PRINT_JITSTATE=1
+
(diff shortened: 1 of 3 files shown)


vext01:
@ptersilie just remembered @ltratt is away today. Do you wanna look at this?

ptersilie:
Looks good to me.

vext01:
I did wonder this. We would need to think about if it's possible for foreign code to jump (or call) to the return address of the call to the foreign code. This is the only source of ambiguity I can think of.

In general foreign code can do anything it wants of course...

ltratt:
I think the "right" check is that a) we're back to the same source location and b) the same stack address. [The latter condition solves the "got back to the same source location because of e.g. a recursive call".]

bors r+

ltratt:
> So we'd have to keep some kind of abstract stack pointer, and update it based upon the instruction stream.

Yes.

>Even then, if the stack pointer is updated using some dynamic value (e.g. mov rsp, [r11]) we would be in trouble.

Anyone doing that gets what they deserve!

PR #590, opened 2022-12-21
Implement the ability to skip over foreign code.

vext01:
Long story short: this requires disassembling the instruction stream. Read the module comment at the top of `hwtracer/src/decode/ykpt/mod.rs` for the whole story.

Companion PR: https://github.com/ykjit/ykllvm/pull/54

### DIFF
--- a/hwtracer/src/block.rs
+++ b/hwtracer/src/block.rs
@@ -16,7 +16,10 @@ pub enum Block {
         /// Virtual address of *any* byte of the last instruction in this block.
         last_instr: BlockAddr,
     },
-    /// An unkonwn address range.
+    /// An unknown virtual address range.
+    ///
+    /// This is required because decoders don't have perfect knowledge about every block
+    /// in the virtual address space.
     Unknown,
 }
 
@@ -46,8 +49,12 @@ impl Block {
         }
     }
 
-    pub fn unknown() -> Self {
-        Self::Unknown
+    /// Returns `true` if `self` represents an unknown virtual address range.
+    pub fn is_unknown(&self) -> bool {
+        match self {
+            Self::Unknown => true,
+            _ => false,
+        }
     }
 
     /// If `self` represents a known address range, returns the address range, otherwise `None`.
--- a/hwtracer/src/decode/ykpt/mod.rs
+++ b/hwtracer/src/decode/ykpt/mod.rs
@@ -1,4 +1,41 @@
 //! The Yk PT trace decoder.
+//!
+//! The decoder works in two modes:
+//!
+//!  - compiler-assisted decoding
+//!  - disassembly-based decoding
+//!
+//! The former mode is for decoding portions of the trace that are for "native code" (code compiled
+//! by ykllvm). Code built with ykllvm gets static control flow information embedded in the end
+//! binary in a special section. Along with dynamic information provided by the PT packet stream,
+//! we have everything we need to decode these parts of the trace without disassembling
+//! instructions. This is the preferred mode of decoding.
+//!
+//! The latter mode is a fallback for decoding portions of the trace that are for "foreign code"
+//! (code not built with ykllvm). For foreign code there is no static control flow edge information
+//! available, so to decode these parts of the trace we have to disassemble the instruction stream
+//! (like libipt does).
+//!
+//! You may now be asking: why not just skip the parts of the trace that are for foreign code?
+//! After all, if a portion of code wasn't built with ykllvm, then we won't have IR for it anyway,
+//! meaning that the JIT is unable to inline it and we'd have to emit a call into the JIT trace.
+//! Why bother decoding the bit of the trace we don't actually care about?
+//!
+//! The problem is, it's not always easy to identify which parts of a PT trace are for native or
+//! foreign code: it's easy if the CPU that doesn't implement the deferred TIP optimisation (see
+//! the Intel 64 and IA32 Architectures Software Developer's Manual, Vol 3, Section  32.4.2.3).
+//!
+//! Deferred TIPs mean that TNT decisions can come in "out of order" with TIP updates. For example,
+//! a packet stream `[TNT(0,1), TIP(addr), TNT(1,0)]` may be optimised to `[TNT(0, 1, 1, 0),
+//! TIP(addr)]`. If the TIP update to `addr` is the return from foreign code, then when we resume
+//! compiler-assisted decoding then we need only the last two TNT decisions in the buffer
+//! (discarding the first two as we skip over foreign code). The problem is that without successor
+//! block information about the foreign code we can't know how how many TNT decisions correspond
+//! with the foreign code, and thus how many decisions to discard.
+//!
+//! We therefore have to disassemble foreign code, popping TNT decisions as we encounter
+//! conditional branch instructions. We can still use compiler-assisted decoding for portions of
+//! code that are compiled with ykllvm.
 
 use crate::{
     decode::TraceDecoder,
@@ -6,16 +43,21 @@ use crate::{
     llvm_blockmap::{BlockMapEntry, SuccessorKind, LLVM_BLOCK_MAP},
     Block, Trace,
 };
-use intervaltree;
+use iced_x86;
+use intervaltree::IntervalTree;
+use phdrs;
 use std::{
     collections::VecDeque,
     convert::TryFrom,
     fmt::{self, Debug},
-    path::PathBuf,
+    ops::Range,
+    path::{Path, PathBuf},
+    slice,
+    sync::LazyLock,
 };
 use ykutil::{
-    addr::{code_vaddr_to_off, off_to_vaddr},
-    obj::SELF_BIN_PATH,
+    self,
+    obj::{PHDR_MAIN_OBJ, SELF_BIN_PATH},
 };
 
 mod packet_parser;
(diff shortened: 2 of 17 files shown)


vext01:
A couple of sneaky extra commits. The first will be squashed, the second is separate.

hwtracer/src/block.rs
```diff
@@ -50,6 +50,13 @@ impl Block {
         Self::Unknown
     }
 
+    pub fn is_unknown(&self) -> bool {
```
ltratt, line 53: What does `unknown` mean here? [A doc string might be in order?]

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -1,21 +1,63 @@
...
+//!
+//! Deferred TIPs mean that TNT decisions can come in "out of order" with TIP updates. For example,
+//! a packet stream `[TNT(0,1), TIP(addr), TNT(1,0)]` may be optimised to `[TNT(0, 1, 1, 0),
+//! TIP(addr)]`. If the TIP update to `addr` is the return from foreign code, then when we resume
+//! compiler-assisted decoding then we need only the last two TNT decisions in the buffer
+//! (discarding the first two as we skip over foreign code). The problem is that without successor
+//! block information about the foreign code we can't know how how many TNT decisions correspond
+//! with the foreign code, and thus how many decisions to discard.
+//!
+//! We therefore surrender and disassemble foreign code, popping TNT decisions as we encounter
```
ltratt, line 36: s/surrender and/have to/ ?

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -38,32 +80,151 @@ impl TraceDecoder for YkPTTraceDecoder {
...
+                let key = vaddr..(vaddr + memsz);
+                segs.push((key.clone(), ()));
+            }
+        }
+    }
+    let tree = segs.into_iter().collect::<IntervalTree<usize, ()>>();
+    CodeSegs { tree }
+});
+
+const PT_MAX_COMPRETS: usize = 64;
```
ltratt, line 110: Doc string?

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -38,32 +80,151 @@ impl TraceDecoder for YkPTTraceDecoder {
...
+/// FIXME: For now this assumes that no dlopen()/dlclose() is happening.
+struct CodeSegs {
+    tree: IntervalTree<usize, ()>,
+}
+
+impl CodeSegs {
+    /// Obtain the virtual address range and a slice of memory for the segment containing the
+    /// specified virtual address.
+    fn seg<'s: 'a, 'a>(&'s self, vaddr: usize) -> Segment<'a> {
+        let mut hits = self.tree.query(vaddr..(vaddr + 1));
```
ltratt, line 124: Is this a `for` loop in disguise? i.e. can I do `for x in self.tree.query(vaddr..(vaddr + 1))`?

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -38,32 +80,151 @@ impl TraceDecoder for YkPTTraceDecoder {
...
+                    write!(f, "ObjLoc::OtherObjOrUnknown(0x{:x})", e)
+                } else {
+                    write!(f, "ObjLoc::OtherObjOrUnknown(???)")
+                }
+            }
         }
     }
 }
 
+/// The returns addresses that can appear on the compressed return stack.
```
ltratt, line 175: s/returns/return/

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -97,126 +280,355 @@ impl<'t> YkPTBlockIterator<'t> {
...
-                        return Ok(Some(self.lookup_block_from_main_bin_offset(target_off)));
-                    } else {
-                        // Call target isn't known statically.
-                        self.cur_loc = ObjLoc::OtherObjOrUnknown;
-                        return Ok(Some(Block::unknown()));
-                    }
-                }
+        match self.cur_loc {
+            ObjLoc::MainObj(b_off) => {
+                // We know where we are in the main object binary, so there's a chance that there's a
```
ltratt, line 298: Not sure I understand this comment (is it saying "this allows to handle parts of the main executable that don't have block information"?).

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -97,126 +280,355 @@ impl<'t> YkPTBlockIterator<'t> {
...
+                                };
+
+                                self.cur_loc = ObjLoc::MainObj(off + 1);
+                                return self.lookup_block_from_main_bin_offset(off + 1);
+                            } else {
+                                // A regular uncompressed return that relies on a TIP update.
+                                //
+                                // Note that `is_return_compressed()` has already updated
+                                // `self.cur_loc()`.
+                                let ret = match self.cur_loc {
```
ltratt, line 384: Maybe:

```rust
match ... {
  ObjLoc::MainObj(off) => return ...,
  _ => return Block::Unknown
}
```

?

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -97,126 +280,355 @@ impl<'t> YkPTBlockIterator<'t> {
...
+                    let vaddr = match self.cur_loc {
+                        ObjLoc::MainObj(off) => self.off_to_vaddr(&PHDR_MAIN_OBJ, off)?,
+                        ObjLoc::OtherObjOrUnknown(opt_vaddr) => match opt_vaddr {
+                            Some(vaddr) => vaddr,
+                            None => unreachable!(),
+                        },
+                    };
+
+                    if inst.flow_control() == iced_x86::FlowControl::IndirectCall {
+                        if inst.next_ip() as usize == vaddr {
```
ltratt, line 528: Is the `as` always safe? If so, we should briefly say why. [I dislike `as` in Rust!]

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -97,126 +280,355 @@ impl<'t> YkPTBlockIterator<'t> {
...
+                    dis.set_ip(self.branch_target_vaddr(&inst));
+                    reposition = true;
                 }
-            } else {
-                // In the absence of blockmap info, we cannot know the successor block.
-                self.cur_loc = ObjLoc::OtherObjOrUnknown;
-                return Ok(Some(Block::unknown()));
+                iced_x86::FlowControl::Call => {
+                    if inst.code() == iced_x86::Code::Syscall {
+                        // Do nothing. We assume the kernel isn't traced.
```
ltratt, line 556: Does a syscall (seem to) immediately return?

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -97,126 +280,355 @@ impl<'t> YkPTBlockIterator<'t> {
...
+        self.disassemble(start_vaddr)
     }
 
-    /// Keep decoding packets until we have some TNT decisions buffered.
-    fn seek_tnt(&mut self) -> Result<bool, HWTracerError> {
-        while self.tnts.is_empty() {
+    /// Keep decoding packets until we encounter a TNT packet.
+    fn seek_tnt(&mut self) -> Result<(), HWTracerError> {
+        loop {
             let pkt = self.packet()?; // Potentially populates `self.tnts`.
```
ltratt, line 599: Does this return `Err` if empty? If not, could we get an infinite loop?

hwtracer/src/decode/ykpt/mod.rs
vext01: It's like this because generally speaking an interval tree query can return more than one result, but in our scenario that is impossible because our query is a point interval (a range of one element, which cannot span multiple segments). Due to this there should be exactly zero or one result.

hwtracer/src/decode/ykpt/mod.rs
vext01: Indeed. Like the PLT code we discovered last week. I'll try a rewording at some point.

hwtracer/src/decode/ykpt/mod.rs
vext01: It is safe for x86_64 (it's `u64` to `usize`), but this is a leftover from hacking. In principle i aim to never use `as`, but during hacking I find it too tedious to do the `usize::try_from(...).unwrap()` dance. Will fix later.

hwtracer/src/decode/ykpt/mod.rs
vext01: It makes a `TIP.PGD` packet (tracing disabled) upon entry to the kernel, and a `TIP.PGE` (tracing enabled) packet upon reentry to userspace. We handle these already.

There is the option to trace the kernel, but as we've discussed, that'd be nonsense.

hwtracer/src/decode/ykpt/mod.rs
vext01: If the trace is exhausted it returns the new `HWTracer::NoMorePackets` error.

hwtracer/src/decode/ykpt/mod.rs
ltratt: But after this line is a `while` loop that pumps `hits`?

hwtracer/src/decode/ykpt/mod.rs
ltratt: Can we clarify the comment to something like "We've turned off kernel tracing elsewhere, so syscalls are TIP.PGD immediately followed by TIP.PGE"?

hwtracer/src/decode/ykpt/mod.rs
ltratt: Got it!

hwtracer/src/decode/ykpt/mod.rs
vext01: I don't think so.

We just want to make sure that we consume a sole result from the query. In debug mode we call `next()` twice to ensure there isn't another element. I'm not sure how a while loop would factor into this.

hwtracer/src/block.rs
vext01: Fixed in c4d926c7500dda90f238e7c3dd42107d49d4ed5a

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in d4bdb709ef7e773e42409937f0f7d1e6452ff43e

(Sorry, got mixed into another change, but should be easy to see)

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in d4bdb709ef7e773e42409937f0f7d1e6452ff43e

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in cc4209f889d54c8e30a1c48953453821052436c2

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in 53778d2bd9c41d6d5b0486c8640d65160ae55b74

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in a466b1e0bdc4a1769572b886c98264e8f206661d

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in 98d683e662382b103e3ad05e1d5dcacd5cacfdc9

hwtracer/src/decode/ykpt/mod.rs
vext01: Fixed in 0fa48b4f824eb454261fe5378557e0d6581a4149

hwtracer/src/block.rs
ltratt: I'm still a bit sure what "unknown" means. I guess it means "not mapped in this executable" or something like that?

ltratt:
Only a couple of minor things left.

hwtracer/src/block.rs
vext01: It means "an unkown virtual address range". The notion was introduced with the new ipt decoder.

It's explained in [this commit message](https://github.com/ykjit/yk/pull/585/commits/aafa6f83fce1471c8940dfad503b8a38a9baa20c) from an earlier PR (the passage starting "The shape of `Block` had to change...").

We can expand the doc string for the variant if you want.

ltratt: I think we need to explain somewhere in the code (maybe a doc string at the top of the block or wherever) what "unknown" means -- but I'm happy with wherever you might think the best place is.

vext01: c72dec74b5adb84cf772b451fab3aa8943576628 expands the docstring and unifies usage of the variant a bit.

ltratt:
Please squash.

vext01:
bear with me, I rebased and that required me to rebuild llvm...

vext01:
squashed and rebased.

PR #592, opened 2023-01-11
Make the ykpt trace decoder the default.

vext01:
Fixes the last remaining failing test and sets our decoder as the default.

(The comment about a necessary compiler change was incorrect. Removed)

### DIFF
--- a/hwtracer/src/decode/ykpt/mod.rs
+++ b/hwtracer/src/decode/ykpt/mod.rs
@@ -394,12 +394,13 @@ impl<'t> YkPTBlockIterator<'t> {
             }
             SuccessorKind::Dynamic => {
                 // We can only know the successor via a TIP update in a packet.
-                //
-                // FIXME: can't test without fixing indirect branch support for the
-                // block disambiguator pass in ykllvm.
-                //
-                // Test already written: tests/hwtracer_ykpt/indirect_jump.c
-                todo!();
+                self.seek_tip()?;
+                match self.cur_loc {
+                    ObjLoc::MainObj(off) => {
+                        return Ok(self.lookup_block_from_main_bin_offset(off)?)
+                    }
+                    _ => return Ok(Block::Unknown),
+                };
             }
         }
     }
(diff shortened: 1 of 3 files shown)


hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -394,12 +394,13 @@ impl<'t> YkPTBlockIterator<'t> {
...
                 // We can only know the successor via a TIP update in a packet.
-                //
-                // FIXME: can't test without fixing indirect branch support for the
-                // block disambiguator pass in ykllvm.
-                //
-                // Test already written: tests/hwtracer_ykpt/indirect_jump.c
-                todo!();
+                self.seek_tip()?;
+                match self.cur_loc {
+                    ObjLoc::MainObj(off) => {
```
ltratt, line 399: I think the `return`s are superfluous (I think the block falls through to the function end)?

vext01: I was following the precedent of the other cases.

It's a left-over from when this function was inlined into `do_next()`. I'll push a follow up commit that updates them all.

vext01: Done in [eaea9da](https://github.com/ykjit/yk/pull/592/commits/eaea9da2c1fe1703b81f031fe8b0c4ef12a21e82)

ltratt:
Please squash.

vext01:
None required.

vext01:
Can't repro.

Doing a try to see if it is non-deterministic...

bors try

vext01:
A completely different error!

vext01:
That last push fixes the second failure. The first appears to be non-deterministic and potentially rare...

bors try

vext01:
@ltratt any thoughts on what to do with this. Seems there is a rare segfault, possibly (probably?) unrelated to this change.

ltratt:
Please squash.

Then, separately, I suggest you make a loop that just keeps running `.buildbot.sh` (or, at least, the tests) until you see a crash again, perhaps attaching gdb, and we might be able to at least see *where* it crashes (why it crashes might be a lot harder!).

vext01:
> we might be able to at least see where it crashes

Agreed.

Squashed

vext01:
So this crash isn't that rare.

I've already started debugging. Shall we retry this just to get it in?

PR #595, opened 2023-01-18
Handle FUP/TIP.* packets that pair with MODE.* packets.

vext01:
The Intel manual explains that if MODE.* packets precede a TIP.* or a FUP packet, then the FUP packet is consumed.

These cases communicate the execution mode of the CPU (WRT 16/32/64-bit mode, and transaction state) and are emitted when the mode changes, but also in PSB+ sequences (a decoder may have missed a prior mode change due to being desynchronised)

This differs from the case (that we've already partially handled), where a FUP packet indicates an interruption in the regular control flow of the instruction stream.

This change ensures that we can handle MODE changes. Fixes one class of non-deterministic crashes.

While here, properly parse the fields of MODE.* packets and check that we only ever see 64-bit mode.

### DIFF
--- a/hwtracer/src/decode/ykpt/mod.rs
+++ b/hwtracer/src/decode/ykpt/mod.rs
@@ -62,6 +62,7 @@ use ykutil::{
 
 mod packet_parser;
 use packet_parser::{
+    packets::Bitness,
     packets::{Packet, PacketKind},
     PacketParser,
 };
@@ -84,6 +85,7 @@ impl TraceDecoder for YkPTTraceDecoder {
             tnts: VecDeque::new(),
             comprets: CompressedReturns::new(),
             pge: false,
+            unbound_modes: false,
         };
         Box::new(itr)
     }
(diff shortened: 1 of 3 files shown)


hwtracer/Cargo.toml
```diff
@@ -17,6 +17,7 @@ ykutil = { path = "../ykutil" }
 intervaltree = "0.2.7"
 byteorder = "1.4.3"
 leb128 = "0.2.5"
+valve = { path = "/home/vext01/research/valve" }
```
vext01, line 20: gah, just noticed that i left debugging aids in. Will push a commit.

vext01:
Sorry about that. Try now.

ltratt:
Does this fix all of the known non-deterministic errors? Should we try the `.buildbot.sh` thing again?

vext01:
There are a couple more to look at first.

ltratt:
OK. Please squash.

PR #598, opened 2023-01-31
Run tests for both PT decoders.

vext01:
[wrote this branch a week or two ago and forgot about it! may also help tracking down non-deterministic bugs]

We run the C suite twice, once for each PT decoder.

(It doesn't make sense to do so in the other suites as they either explicitly test one specific decoder, or don't use PT to get a trace)

In order to select which decoder to use we introduce the `YKD_FORCE_TRACE_DECODER` env var. This is strictly a debugging/testing facility. Eventually we will add an API that allows the interpreter authors to select a decoder if they wish (but not now).

### DIFF
--- a/hwtracer/Cargo.toml
+++ b/hwtracer/Cargo.toml
@@ -24,3 +24,6 @@ iced-x86 = { version = "1.18.0", features = ["decoder"]}
 [build-dependencies]
 cc = "1.0.62"
 rerun_except = "0.1.2"
+
+[features]
+yk_testing = []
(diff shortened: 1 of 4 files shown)


hwtracer/Cargo.toml
```diff
@@ -24,3 +24,6 @@ iced-x86 = { version = "1.18.0", features = ["decoder"]}
 [build-dependencies]
 cc = "1.0.62"
 rerun_except = "0.1.2"
+
+[features]
+yk_testing = []
```
ltratt, line 29: Should we add a comment here along the lines of "internal flag, used only for testing"? Otherwise external users might wonder if they should care about it.

vext01: We could do, but this idiom appears in many crates in the workspace.

I'm about to do a ykdocs update, so maybe it would be better said there just once?

ltratt: Works for me!

PR #600, opened 2023-02-02
Start optimizing ykpt decoder

vext01:
Please review individual commits.

The first three commits are tidy ups and refactorings that put the code surrounding dladdr into a shape that is easy to optimise in one place.

The last commit is the first optimisation I'm proposing.

## decoding a trace that does a lot of disassembling, before and after

violin

violin2

## decoding a trace that ykpt can mostly use compiler-assisted decoding for

violin3

violin4

## Observations

 - ykpt disassembly is improved by about 6x
 - ykpt compiler-assisted decode unaffected
 - we are still slower for longer traces

## Discussion

Undoubtedly we can do better with a more involved cache, but this is such a low-effort/high-gain change that I think this is a good start point. Especially since dladdr() is no-longer a bottleneck. The next bottleneck is in iterating the program headers. Perf suggests we spend 40% of decoding time doing this in some cases. This will be my next PR.

I also tried using a `dashmap` instead of `#[cached]` (which uses a `RwLock` instead of a `Mutex`) and found it to be slower.

What do you think?


ykutil/src/addr.rs
```diff
@@ -11,15 +11,20 @@ use std::{
     ptr::null,
 };
 
+pub fn dladdr(vaddr: usize) -> Result<Dl_info, ()> {
```
ltratt: A short doc string here might be useful (e.g. for the `Err` case).

ykutil/src/obj.rs
```diff
@@ -21,11 +22,7 @@ pub static SELF_BIN_PATH: LazyLock<PathBuf> = LazyLock::new(|| {
     if addr == ptr::null_mut() as *mut c_void {
         panic!("couldn't find address of main()");
     }
-    let mut info = MaybeUninit::<Dl_info>::uninit();
-    if unsafe { dladdr(addr, info.as_mut_ptr()) } == 0 {
-        panic!("couldn't find Dl_info for main()");
-    }
-    let info = unsafe { info.assume_init() };
+    let info = dladdr(addr as usize).unwrap(); // ptr to usize cast always safe.
```
ltratt, line 25: I think this can be `usize::from(addr)`?

Also, is the `unwrap` guaranteed not to fail?

yktrace/src/hwt/mapper.rs
```diff
@@ -109,16 +109,23 @@ impl<'a> HWTMapper<'a> {
...
-                for bb in ent.value.corr_bbs() {
-                    ret.push(Some(IRBlock::new(
-                        sio.sym_name().to_owned(),
-                        usize::try_from(*bb).unwrap(),
-                    )));
+                debug_assert_eq!(
+                    obj_name.to_str().unwrap(),
+                    sio.dli_fname().unwrap().to_str().unwrap()
+                );
+                if let Some(sym_name) = sio.dli_sname() {
```
ltratt, line 116: Does `faddrs` support the `entry` API? That can save the double lookup (e.g. with `or_insert` and friends).

ykutil/src/addr.rs
```diff
@@ -5,16 +5,66 @@ use libc::{self, c_void, Dl_info};
...
 };
 
-pub fn dladdr(vaddr: usize) -> Result<Dl_info, ()> {
+/// A Rust wrapper around `libc::Dl_info` using FFI types.
+///
+/// The strings inside are handed out by the loader and can (for now, since we don't support
+/// dlclose) be considered of static lifetime.
+#[derive(Debug)]
+pub struct DLInfo {
+    dli_fname: Option<&'static CStr>,
```
ltratt, line 21: I *think* `&'static` is probably OK here but it does require us to guarantee that anything we reference can never be unloaded. Is that true? Or, at least, should we make that requirement explicit to the user somewhere?

ykutil/src/addr.rs
```diff
@@ -5,16 +5,66 @@ use libc::{self, c_void, Dl_info};
...
+    dli_fbase: usize,
+    dli_sname: Option<&'static CStr>,
+    dli_saddr: usize,
+}
+
+impl From<Dl_info> for DLInfo {
+    fn from(dli: Dl_info) -> Self {
+        let dli_fname = if !dli.dli_fname.is_null() {
```
ltratt, line 29: Isn't this something like `dli.dli_fname.map(|x| CStr::from_ptr(x))` in disguise?

ykutil/src/addr.rs
vext01: It's discussed in the docstring for this struct. dlopen/dlclose would cause us some problems.

There are many places in the JIT that assume dlopen/dlclose don't happen right now. Later I think we can wrap the libc versions of these functions and call a cache invalidation routines. But I don't want to go there at this moment in time. Is that OK?

ykutil/src/addr.rs
vext01: I can't see any `ptr::map()` in the std lib... What am I missing?

`dli.dli_fname` is a [`*const c_char`](https://docs.rs/libc/0.2.139/libc/struct.Dl_info.html).

ykutil/src/obj.rs
ltratt: I generally don't find `expect` super helpful, so I'm OK with the `unwrap` here, but I like to document it with something like "// If this fails, there's nothing useful we can do other than exit".

ykutil/src/addr.rs
ltratt: OK.

ykutil/src/addr.rs
ltratt: Oh, oops!

yktrace/src/hwt/mapper.rs
vext01: That would look like this:
```diff
-                    if !self.faddrs.contains_key(sym_name) {
-                        self.faddrs
-                            .insert(sym_name.to_owned(), sio.dli_saddr() as *const c_void);
-                    }
+                    self.faddrs.entry(sym_name.into()).or_insert_with(|| sio.dli_saddr() as *const c_void);
```

Am I right that the (required) `.into()` is going to clone a string each time? Perhaps it is better to leave it as-is?

ltratt: Hmm, yes, maybe! OK, let's leave as-is.

ykutil/src/addr.rs
vext01: Fixed in dfda6853f96232217fd0927ce41039281e711e10

ykutil/src/obj.rs
vext01: Fixed in e5a024335eae9846fc47681829db0e5e0ed15587

vext01:
Ready for re-review.

ykutil/src/obj.rs
ltratt: What about the `usize::from` comment?

vext01: Woops, missed that.

Whats the advantage actually? We always cast from the same abi to the same abi.

I guess if some code changed around it, we would catch mistakes?

vext01: Well, we can't use `from` without an additional `as` here anyway:
```
the trait bound `usize: From<*const c_void>` is not satisfied
```

ltratt: If we can't use it, fair enough.

ltratt:
Please squash.

vext01:
splat.

PR #602, opened 2023-02-07
Fix a rare assertion failure.

vext01:
The FUP workaround code checks for a sequence of three PT packets that indicate that the interruption was an "easy workaround case".

In the case that the third packet happens to be a `TIPPGD`, the decoder's `self.pge` was not correctly updated. This meant that when the decoder sees the corresponding `TIPPGE` and it asserts `!self.pge`, the assertion fails.

This change reshuffles the code a bit so that the update to `self.pge` always happens correctly.

### DIFF
--- a/hwtracer/src/decode/ykpt/mod.rs
+++ b/hwtracer/src/decode/ykpt/mod.rs
@@ -656,42 +656,39 @@ impl<'t> YkPTBlockIterator<'t> {
         let ret = if let Some(pkt_or_err) = self.parser.next() {
             let mut pkt = pkt_or_err?;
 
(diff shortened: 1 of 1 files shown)


ltratt:
Possibly silly question: is there any way that we can include test data (i.e. from PT), even if just in buildbot? It's not exactly a question for this PR, but might be worth thinking about?

bors r+

vext01:
By test data, do you mean a captured PT trace?

ltratt:
Yes. I don't know if we can minimise them to be small enough to fit in a test (sounds hard!), or host bigger traces outside the repo such that buildbot.sh can download them. It might be too hard, but I thought it's worth raising as an idea.

vext01:
We would also have to store the binary executable somewhere...

PR #604, opened 2023-02-13
Ensure C++ stuff gets rebuilt when .cpp or .h files change.

vext01:
I'm sure you've seen it where you change something in ykllvmwrap, and re-run the tests and old code is run...

### DIFF
--- a/ykllvmwrap/build.rs
+++ b/ykllvmwrap/build.rs
@@ -1,6 +1,10 @@
+use rerun_except::rerun_except;
 use std::{env, process::Command};
 
 fn main() {
+    // Ensure changing C++ source files or headers retriggers a build.
+    rerun_except(&[]).unwrap();
+
     // Compile our wrappers with the right LLVM C++ flags.
     let cxxflags_out = Command::new("llvm-config")
         .arg("--cxxflags")
--- a/ykllvmwrap/Cargo.toml
+++ b/ykllvmwrap/Cargo.toml
@@ -12,6 +12,7 @@ ykutil = { path = "../ykutil" }
 
 [build-dependencies]
 cc = "1.0.72"
+rerun_except = "1.0.0"
 
 [features]
 yk_testing = []


ykllvmwrap/build.rs
```diff
@@ -1,6 +1,9 @@
+use rerun_except::rerun_except;
 use std::{env, process::Command};
 
 fn main() {
+    rerun_except(&[]).unwrap();
```
ptersilie, line 5: What exactly does this do?

vext01: build.rs provides a way to enumerate the files that, if changed, should retrigger a `build.rs` script.

`rerun_except` is laurie's crate that inverts the logic: what files should *not* retrigger a build.

with laurie's way, you can't forget to add a file.

https://crates.io/crates/rerun_except

ptersilie: Oh I see. Clever. Should we add a small comment?

vext01: I think it should be pretty obvious once the reader has seen the `rerun_except` crate

ptersilie: It's just I hadn't seen the crate at the time and thus had no idea what this does. So I thought what could happen to me, could happen to someone else. But I'm not gonna fight for that comment.

vext01: There ya go: cc24536115d4f1b5abcbb036cc4a7e6f439f05f3

vext01:
OK to squash?

ykllvmwrap/build.rs
ltratt: Isn't `rerun_except(&[]).unwrap();` the same as not using rerun_except?

ptersilie:
You know how to make me happy. Please squash.

ykllvmwrap/build.rs
ptersilie: Oh I see. So cargo reruns everything anyway and you use `rerun_except` to exclude files. If that's the case then I agree that this line seems redundant.

ltratt: I think this is a question for @vext01: does this PR change the rebuild behaviour for C/C++ files?

vext01: On `master`:
```
$ cargo build
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
$ touch ykllvmwrap/src/memman.cc 
$ cargo build
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
$ 
```

On this branch:
```
 cargo build
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
$ touch ykllvmwrap/src/memman.cc 
$ cargo build
   Compiling ykllvmwrap v0.1.0 (/home/vext01/research/yk2/ykllvmwrap)
   Compiling yktrace v0.1.0 (/home/vext01/research/yk2/yktrace)
   Compiling ykrt v0.1.0 (/home/vext01/research/yk2/ykrt)
   Compiling tests v0.1.0 (/home/vext01/research/yk2/tests)
   Compiling ykcapi v0.1.0 (/home/vext01/research/yk2/ykcapi)
    Finished dev [unoptimized + debuginfo] target(s) in 15.78s
$ 
```

The [Cargo docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection) are ambiguous IMHO:

> When rebuilding a package, Cargo does not necessarily know if the build script needs to be run again. By default, it takes a conservative approach of always re-running the build script if any file within the package is changed.

Is a `.cc` file "within the package" if it is in the crate's directory? It appears not, but I'd assume the opposite.

ltratt: Weird -- cargo doesn't work as I would have expected in this case. But you're right that the docs are unclear as to what should happen.

If `rerun_except([])` does what we want, I'm fine with it -- I'm just a bit surprised it's different than cargo!

ltratt:
@ptersilie I'm fine with this PR now.

ykllvmwrap/build.rs
vext01: Might be worth adding a test for this case to `rerun_except`?

vext01:
Squashed and ready to go.

PR #605, opened 2023-02-15
Simplify trace compiler "Have I seen the control point yet" and `ExpectUnmappable` logic.

vext01:
This fixes some oddities that I spotted while accommodating mappable callback from foreign code.

See individual commits.

It hurt my head, but the result is good I think.


ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -1236,12 +1261,18 @@ class JITModBuilder {
   Module *createModule() {
     // Initialise the memory block holding live AOT values for guard failures.
     BasicBlock *NextCompletedBlock = nullptr;
+
     for (size_t Idx = 0; Idx < InpTrace.Length(); Idx++) {
       optional<IRBlock> MaybeIB = InpTrace[Idx];
-      if (ExpectUnmappable && !MaybeIB.has_value()) {
+
+      // If we previously saw a call to foreign code, there will now be a
```
vext01, line 1268: gah, "an"

ptersilie:
This looks fine to me. Though I couldn't work out what this paragraph is saying:
```
In addition, becuase the compiler would start from the beginning of the
first block of the trace (not after the control point), it would think
that it was seeing the call to the control point, which is impossible,
as tracing couldn't have started. Some sloppy `ExpectUnmappable`
handling was letting this through and "working by chance" (tm).
```

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -1098,29 +1094,41 @@ class JITModBuilder {
   //
   // FIXME: this assumes that there is a single call-site of the control point.
   // https://github.com/ykjit/yk/issues/479
-  static tuple<CallInst *, Value *> GetControlPointInfo(Module *AOTMod) {
+  static tuple<CallInst *, size_t, Value *>
```
ltratt, line 1097: I think we need to update the doc string to talk about the `size_t` including what it means if the `for` loop goes over bounds (i.e. doesn't call `break`).

vext01:
> Though I couldn't work out what this paragraph is saying:

Basically, the trace compiler thought that the first block in the (mapped) PT trace corresponded with the call to the control point, when really it was the return *from* the control point. And some unfortunate sloppy logic was making this eventually sync-up by chance later.

ykllvmwrap/src/jitmodbuilder.cc
vext01: The for loop cannot exit by normal means by construction.

We are saying "Get the block that contains instruction I and then iterate over the block until you see I".

I experimented with debug only assertions, but it really cluttered up the code, so I removed it.

Will update the doc string though.

ltratt: > The for loop cannot exit by normal means by construction.

That's not obvious to me -- an `assert` would surely be a good thing here (or, at least, a comment).

vext01: I'll show you what the assert looks like and you can choose :)

vext01: Doc string: a45b81512da07e09d32b8dc791e3a3c86089c31e
Assertion: 033ab519d4f9ece83f82e44ca5c832372805be54

vext01: I must admit, I find the assertion a little redundant. It's like saying: check that a member of a set is in that set.

ltratt: Ironically, the assert helped me realise that! So I agree that this was me being silly and that 033ab51 should be reverted.

ptersilie:
> Basically, the trace compiler thought that the first block in the (mapped) PT trace corresponded with the call to the control point

I thought we had the whole `NewCtrlPointCall` logic to find the return from the control point. Are you saying it wasn't actually doing that?

vext01:
> I thought we had the whole NewCtrlPointCall logic to find the return from the control point. Are you saying it wasn't actually doing that?

It was, but in a very convoluted and subtle way...

This way is more explicit and requires less state.

ykllvmwrap/src/jitmodbuilder.cc
vext01: Reverted in e6b3b7907eff1d82ae86e6b522345d42425733b0

ltratt:
@ptersilie I'm OK with this PR so I'll let you do the remainder with it.

ptersilie:
Please squash.

vext01:
splat.

vext01:
The CI job failed because we encountered a new kind of packet!

image

I'll handle that in a new PR. If you re-run the CI this will likely pass.

ltratt:
Well I never!

bors r+

PR #606, opened 2023-02-16
Parse EXSTOP and fix weird error formatting.

vext01:
See commits.


ltratt:
Does this PR *just* parse EXSTOP or does it do something else too? [I was sort-of guessing it would cause a trace abort, but maybe that's a silly idea!]

vext01:
It just consumes the bytes of the packet so that the parser is aligned to parse the next packet (after the CPU is powered up again there will be more packets).

ltratt:
So no packets are lost because of EXSTOP?

vext01:
Not that I can infer from the docs.

See the [image from the last PR](https://user-images.githubusercontent.com/604955/219388131-40ac31ac-5c72-47d8-a535-0f5373fa0f70.png)

ltratt:
OK, thanks.

bors r+

PR #613, opened 2023-03-01
Make the mapper eager.

vext01:
Prior to this change, the mapper was a pseudo-iterator with a `next()` but using a non-standard interface requiring the consumer to pass in the last value the iterator yielded. This was required in order to collapse consecutive unmappable blocks, and to omit the initial unmappable prefix (both assumptions the trace compiler makes).

This interface is not only odd, but it is hindering my work on allowing foreign code to call back to mappable code.

This change makes the mapper eagerly collect the mapped trace, instead of working like an iterator. By doing so, we kill the weird interface and simplify the code quite a bit.

At first I was hesitant to make this change: I'm biased towards iterators, as they can often be made to use less memory than collecting. But upon inspection, both consumers of this interface (yktrace and one of the test suites) simply collect the trace anyway, so there's really no need for the mapper to be an iterator.

### DIFF
--- a/yktrace/src/hwt/mapper.rs
+++ b/yktrace/src/hwt/mapper.rs
@@ -11,18 +11,14 @@ use ykutil::{
 };
 
 /// Maps each entry of a hardware trace back to the IR block from which it was compiled.
-pub struct HWTMapper<'a> {
+pub struct HWTMapper {
     faddrs: HashMap<CString, *const c_void>,
-    trace_iter: &'a mut dyn Iterator<Item = Result<Block, HWTracerError>>,
-    buffered_irblocks: Vec<IRBlock>,
 }
 
-impl<'a> HWTMapper<'a> {
-    pub fn new(trace_iter: &'a mut dyn Iterator<Item = Result<Block, HWTracerError>>) -> Self {
+impl<'a> HWTMapper {
+    pub fn new() -> Self {
         Self {
             faddrs: HashMap::new(),
-            trace_iter,
-            buffered_irblocks: Vec::new(),
         }
     }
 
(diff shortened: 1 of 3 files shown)


yktrace/src/hwt/mapper.rs
```diff
@@ -134,59 +130,66 @@ impl<'a> HWTMapper<'a> {
...
 
-    /// Obtain the next mapped block.
+    /// Map the *machine* blocks of the specified trace into LLVM IR blocks.
     ///
-    /// `last_yielded` must be a reference to the last successfully mapped block returned by the
-    /// iterator (or `None` if this is the first time you are calling `next`).
-    pub fn next(
+    /// The returned trace is guaranteed to have the following properties:
+    ///
+    ///  - The returned trace will never start with an unmappable code block. It is a given that
```
ltratt, line 137: I eventually worked out what this meant, but can we try simplifying the explanation to make it clearer?

yktrace/src/hwt/mapper.rs
```diff
@@ -134,59 +130,66 @@ impl<'a> HWTMapper<'a> {
...
+    ///    every raw trace will start by returning out of the foreign code that turned on trace
+    ///    collection. We omit this prefix.
+    ///
+    ///  - To save memory, in the returned trace consecutive unmappable block entries will have
```
ltratt, line 141: s/in the returned trace//

s/will have been/are/

yktrace/src/hwt/mapper.rs
```diff
@@ -134,59 +130,66 @@ impl<'a> HWTMapper<'a> {
...
-                if let Some(blk) = last_yielded {
-                    if !blk.is_unmappable() {
-                        self.buffered_irblocks.push(IRBlock::unmappable());
-                    }
+                // trace isn't empty (we never report the leading unmappable code in a trace). We
+                // also take care to collapse consecutive unmappable blocks into one.
+                if let Some(last) = ret.last_mut() {
+                    if !last.is_unmappable() {
+                        ret.push(IRBlock::unmappable());
+                    } // otherwise don't push, thus collapsing repeated unmappable blocks.
```
ltratt, line 159: I tend to find this sort of comment easier to read with an `else` so:

```rust
if !last.is_unmappable() {
  ret.push(IRBlock::unmappable());
} else {
  // otherwise don't push, thus collapsing repeated unmappable blocks.
}
```

I don't feel strongly about it though.

yktrace/src/hwt/mapper.rs
```diff
@@ -134,59 +130,66 @@ impl<'a> HWTMapper<'a> {
...
                 for irblock in irblocks.into_iter() {
-                    if let Some(irblock) = irblock {
-                        // The `BlockDisambiguate` pass in ykllvm ensures that no high-level LLVM
-                        // IR block ever branches straight back to itself, so if we see the same
-                        // high-level block more than once consecutively in a trace, then we know
-                        // that the IR block has been lowered to multiple machine blocks during
-                        // code-gen, and that we should only push the IR block once.
-                        if last_yielded != Some(&irblock) {
-                            self.buffered_irblocks.push(irblock);
+                    if let Some(irb) = irblock {
```
ltratt, line 163: This feels like it's a `match` in disguise?

```rust
match irblock {
  Some(irblock) => ...
  None => ...
}

File: yktrace/src/hwt/mapper.rs [hunk 5]
```diff
@@ -134,59 +130,66 @@ impl<'a> HWTMapper<'a> {
...
+                        if let Some(last) = ret.last() {
```
ltratt, line 164: Maybe a `match` for this to bring out the control flow?

```rust
match ret.last() {
  Some(last) if &irb != last { ... },
  Some(_) => {// the big comment},
  None => {ret.push(irb)
}
```

vext01: And bring it out we shall: c4f521c4c72fbc69c2fd593776c254ffeff306cf

yktrace/src/hwt/mapper.rs
vext01: Try 42e670140381fcf31b31ac5cc91ed93f53697fc1

yktrace/src/hwt/mapper.rs
vext01: Superseded by 42e670140381fcf31b31ac5cc91ed93f53697fc1

yktrace/src/hwt/mapper.rs
vext01: Fixed in e3ac9f932e6cb56003498490264cf643b0683310

yktrace/src/hwt/mapper.rs
vext01: It's equivalent. I tend to prefer `if let Some` for binary choiced stuff, as match tends to indent more.

vext01:
Over to you.

ltratt:
Please squash.

vext01:
splat.

ltratt:
Please also rebase against master.

vext01:
My apologies, done.

vext01:
```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: HWBufferOverflow'
```

😆 

Try again?

vext01:
(we need to address that eventually)

vext01:
https://github.com/ykjit/yk/issues/614

PR #616, opened 2023-03-07
Fix some memory leaks.

ptersilie:
7851c981e0db7055882d7d76258c97553c1f04d3 fixes a large memory leak that lead to some lua programs running out of memory.

c8d861e9f874243c1b8c537d0597262e3bfe3f28 fixes a smaller leak. Interestingly, analysing program execution with heaptrack doesn't seem to find this leak. I wonder if this has to do with `The munmap() function shall remove any mappings for those entire  pages containing  any  part  of  the address space...`. Which I read as: if the allocation is smaller than a page, then it would be freed wholly regardless of the size argument being passed to `munmap`.

### DIFF
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -156,8 +156,11 @@ pub extern "C" fn __ykrt_reconstruct_frames(newframesptr: *const c_void) {
     unsafe {
         asm!(
             // The first 8 bytes of the new frames is the size of the map, needed for copying it
-            // over. Move it into rdx.
+            // over. Move it into RDX, but reduce it by 8 bytes, since we'll also adjust RDI next
+            // to make it point to the actual beginning of the new frames (jumping over the length
+            // value stored at the beginning of RDI).
             "mov rdx, [rdi]",
+            "sub rdx, 8",
             // Then adjust the address to where the new stack actually starts.
             "add rdi, 8",
             // Make space for the new stack, but use 8 bytes less in order to overwrite this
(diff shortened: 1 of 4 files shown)


ykfr/src/llvmbridge.rs
```diff
@@ -23,6 +24,7 @@ impl Module {
         let mut module: MaybeUninit<LLVMModuleRef> = MaybeUninit::uninit();
         LLVMParseBitcodeInContext2(context, membuf, module.as_mut_ptr());
```
vext01, line 25: Can you add and link to an issue (if one doesn't already exist) saying that this should use the loading logic from ykllvmwrap to take advantage of a thread's cache of the IR (instead of loading it every time).

ykcapi/src/lib.rs
```diff
@@ -156,8 +156,10 @@ pub extern "C" fn __ykrt_reconstruct_frames(newframesptr: *const c_void) {
     unsafe {
         asm!(
             // The first 8 bytes of the new frames is the size of the map, needed for copying it
-            // over. Move it into rdx.
+            // over. Move it into rdx, but reduce it by 8 bytes, since we'll move the pointer along
+            // next.
```
vext01, line 160: I'm not sure what you mean by "move the pointer along". Let's be more precise.

ykfr/Cargo.toml
```diff
@@ -12,7 +12,6 @@ ykutil = { path = "../ykutil" }
 yksmp = { path = "../yksmp" }
 llvm-sys = "140"
 memmap2 = "0.5.2"
-capstone = "0.11.0"
```
vext01, line 15: good!

ykcapi/src/lib.rs
ptersilie: [03eb360](https://github.com/ykjit/yk/pull/616/commits/03eb3608dc24d9ab3ca3909900a3f1e188c9aae6)

vext01:
Ready to squash?

ltratt:
This comment hasn't shown up in the PR (GitHub...) https://github.com/ykjit/yk/commit/7851c981e0db7055882d7d76258c97553c1f04d3#r103380122

ltratt:
Please squash.

ptersilie:
Squashed.

PR #618, opened 2023-03-13
Clone global variables recursively.

ptersilie:
If a global variable references other globals, we need to clone those into the new module too.

### DIFF
--- a/tests/c/strarray.c
+++ b/tests/c/strarray.c
@@ -1,0 +1,64 @@
+// Run-time:
+//   env-var: YKD_PRINT_IR=aot
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_PRINT_JITSTATE=1
+//   stderr:
+//     jit-state: start-tracing
+//     pepper
+//     jit-state: stop-tracing
+//     --- Begin aot ---
+//     ...
+//     @fruits = internal constant [5 x ptr] [ptr @.str.2, ptr @.str.3, ptr @.str.4,...
+//     ...
+//     @.str.2 = private unnamed_addr constant [6 x i8] c"apple\00", align 1...
+//     @.str.3 = private unnamed_addr constant [7 x i8] c"banana\00", align 1...
+//     @.str.4 = private unnamed_addr constant [7 x i8] c"tomato\00", align 1...
+//     ...
+//     --- End aot ---
+//     cucumber
+//     jit-state: enter-jit-code
+//     tomato
+//     jit-state: exit-jit-code
+//     jit-state: enter-jit-code
+//     banana
+//     jit-state: deoptimise
+//     ...
+//     jit-state: exit-jit-code
+//   stdout:
+//     exit
+
+// Check that when we clone a `GlobalVariable` into the JITMod, and that global
+// references other `GlobalVariables`, the other globals are cloned as well.
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+static const char *const fruits[] = {"apple", "banana", "tomato", "cucumber",
+                                     "pepper"};
+
+int main(int argc, char **argv) {
+  YkMT *mt = yk_mt_new();
+  yk_mt_hot_threshold_set(mt, 0);
+  YkLocation loc = yk_location_new();
+
+  int res = 9998;
+  int i = 4;
+  NOOPT_VAL(loc);
+  NOOPT_VAL(res);
+  NOOPT_VAL(i);
+  while (i > 0) {
+    yk_mt_control_point(mt, &loc);
+    fprintf(stderr, "%s\n", fruits[i]);
+    res += 2;
+    i--;
+  }
+  printf("exit");
+  NOOPT_VAL(res);
+  yk_location_drop(loc);
+  yk_mt_drop(mt);
+  return (EXIT_SUCCESS);
+}
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -875,6 +875,21 @@ class JITModBuilder {
     if (OldGV->isConstant()) {
       GV->copyAttributesFrom(&*OldGV);
       cloned_globals.push_back(OldGV);
+      // If this is a vector containing other constants, we need to clone
+      // those as well.
+      Type *Ty = OldGV->getValueType();
+      Constant *C = OldGV->getInitializer();
+      if (Ty->isArrayTy()) {
+        if ((cast<ArrayType>(Ty))->getElementType()->isPointerTy()) {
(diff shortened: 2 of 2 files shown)


tests/c/strarray.c
```diff
@@ -0,0 +1,54 @@
...
+//     jit-state: exit-jit-code
+//     jit-state: enter-jit-code
+//     banana
+//     jit-state: deoptimise
+//     ...
+//     jit-state: exit-jit-code
+//   stdout:
+//     exit
+
+// Check that basic trace compilation works.
```
ltratt, line 20: This comment isn't very informative. In fact, I'd suggest it's misleading! Isn't the point of this test the recursive global thingy? I wondered if it might sense to even do something like:

```
static char *x = "...";
static char *y = "...";
static char *[] = {x, y};
```

I'm not sure if that's valid C or not though!

ptersilie: Oh forgot to update that description. Will fix.

ptersilie: Fixed the comment in [ff92c79](https://github.com/ykjit/yk/pull/618/commits/ff92c79ca99ac236b53c21d2cf101b8ca17bb192). There's no need to actually create different globals. In LLVM IR a global array will automatically create a separate global for each element, and then put the pointers in the array instead.

ltratt: >  There's no need to actually create different globals.

I understand (or, at least, guessed!) that, but wouldn't it be clearer if we separated this out rather than relying on LLVM's (possibly fragile) internal behaviour?

ltratt: Also, the comment says "global arrays" but the PR is "recursive global variables". I assume the former is just a special case of the latter?

ptersilie: The PR says "clone global variables recursively", i.e. look at each element in the global array and clone them too if they are global variables.

I could add a check to see that the AOT IR does indeed contain a global variable array with pointers to other globals.

ltratt: > The PR says "clone global variables recursively", i.e. look at each element in the global array

But notice you said "clone global *variables* recursively" but then global *array*. It's confusing for thickies like me!

ptersilie: I'm talking about an LLVM `GlobalVariable` that is underneath a `ConstantArray`.

ptersilie: Let me try another fix.

ptersilie: Try if this works better for you: [26cd73c](https://github.com/ykjit/yk/pull/618/commits/26cd73c19abe075fc753c6b0ee448e24591c797a)

ltratt:
Works for me! If @vext01 is OK with this, you should feel free to squash.

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -875,6 +875,23 @@ class JITModBuilder {
     if (OldGV->isConstant()) {
       GV->copyAttributesFrom(&*OldGV);
       cloned_globals.push_back(OldGV);
+      // If this is a vector containing other constants, we need to clone
+      // those as well.
+      Type *Ty = OldGV->getValueType();
+      Constant *C = OldGV->getInitializer();
+      if (Ty->isArrayTy()) {
+        if ((cast<ArrayType>(Ty))->getElementType()->isPointerTy()) {
+          auto GVA = dyn_cast<ConstantArray>(C);
```
vext01, line 884: How do we feel about `auto`?

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -875,6 +875,23 @@ class JITModBuilder {
...
+          for (size_t I = 0; I < Ty->getArrayNumElements(); I++) {
+            auto Elem = GVA->getAggregateElement(I);
+            if (isa<GlobalVariable>(Elem)) {
```
vext01, line 887: We can fold these two `if`s together.

vext01:
Just a couple of comments.

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: [429140f](https://github.com/ykjit/yk/pull/618/commits/429140f792f0ed9365cf1404b56c8e492130c172)

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: [fd302af](https://github.com/ykjit/yk/pull/618/commits/fd302aff8d6e40b6e04dbf92b966ee67583b6ac7)

ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -881,7 +881,7 @@ class JITModBuilder {
       Constant *C = OldGV->getInitializer();
       if (Ty->isArrayTy()) {
         if ((cast<ArrayType>(Ty))->getElementType()->isPointerTy()) {
-          auto GVA = dyn_cast<ConstantArray>(C);
+          ConstantArray *GVA = dyn_cast<ConstantArray>(C);
           for (size_t I = 0; I < Ty->getArrayNumElements(); I++) {
             auto Elem = GVA->getAggregateElement(I);
```
vext01, line 886: Another auto here :)

ltratt:
Is there a particular reason to consider `auto` a problem? After all, we heavily use type inference in Rust, so I don't immediately see why it should be banned in C++ code?

ykllvmwrap/src/jitmodbuilder.cc
ptersilie: [08a1217](https://github.com/ykjit/yk/pull/618/commits/08a12177f45f2b8a623b0018c23670365af83b91)

vext01:
Please squash.

vext01:
We should review our rules on `auto` because it does have some very useful moments.

E.g. https://dominikberner.ch/structured-bindings/

(a.k.a. goodbye `std::tie`, which makes unpacking a tuple of `n` items a one-liner, instead of a `n+1`-liner)

ptersilie:
Squashed.

ptersilie:
> We should review our rules on `auto` because it does have some very useful moments.
> 
> E.g. https://dominikberner.ch/structured-bindings/
> 
> (a.k.a. goodbye `std::tie`, which makes unpacking a tuple of `n` items a one-liner, instead of a `n+1`-liner)

That's neat!

ptersilie:
Ok to squash?

vext01:
yup

ptersilie:
Squashed

vext01:
> smmultisrc2: src/jitmodbuilder.cc:1314: Module *JITModBuilder::createModule(): Assertion `std::get<1>(ResumeAfter.value())->getParent() == BB' failed.

ptersilie:
That's the nondeterministic bug. So we probably just have to run it again.

vext01:
OK, I think I'm going to have to see if i can repro this.

bors r+

PR #619, opened 2023-03-15
Add optimisations to make running Lua tests more tolerable.

ptersilie:
See commits for more info. Bottom line is, this significantly speeds up lua tests, to continue working on correctness.

### DIFF
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -238,6 +238,8 @@ pub extern "C" fn yk_stopgap(
     let mut framerec = unsafe { FrameReconstructor::new(activeframes) };
 
     // Parse the stackmap of the JIT module.
+    // OPT: Parsing the stackmap and initialising `framerec` is slow and could be heavily reduced
+    // by caching the result.
     let slice = unsafe { slice::from_raw_parts(stackmap.addr as *mut u8, stackmap.length) };
     let map = StackMapParser::parse(slice).unwrap();
     let live_vars = map.get(&retaddr.try_into().unwrap()).unwrap();
--- a/ykfr/src/lib.rs
+++ b/ykfr/src/lib.rs
@@ -1,3 +1,5 @@
+#![feature(once_cell)]
+
 use llvm_sys::{core::*, target::LLVMABISizeOfType};
 use object::{Object, ObjectSection};
 use std::{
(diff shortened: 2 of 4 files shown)


ykcapi/src/lib.rs
```diff
@@ -238,6 +238,8 @@ pub extern "C" fn yk_stopgap(
     let mut framerec = unsafe { FrameReconstructor::new(activeframes) };
 
     // Parse the stackmap of the JIT module.
+    // YKFIXME: Parsing the stackmap and initialising `framerec` has a 40ms overhead, which could
```
ltratt, line 241: Should this be `YKOPT`?

vext01: Also the 40s metric is specific to yklua. I think it's enough just to say "this is really slow" or something like this.

ykfr/src/lib.rs
```diff
@@ -6,12 +8,32 @@ use std::{
...
 };
-use yksmp::{Location as SMLocation, StackMapParser};
+use yksmp::{Location as SMLocation, SMEntry, StackMapParser};
 
 mod llvmbridge;
 use llvmbridge::{get_aot_original, Module, Type, Value};
 
+pub static AOT_STACKMAPS: LazyLock<Vec<SMEntry>> = LazyLock::new(|| {
+    // Load the stackmap from the binary to parse in the stackmaps.
+    // FIXME: Don't use current_exe.
```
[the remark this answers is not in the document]
vext01: The `YK` prefix is only used in ykllvm to distinguish our FIXME from theirs.

ykfr/src/lib.rs
```diff
@@ -6,12 +8,32 @@ use std::{
...
+    let pathb = env::current_exe().unwrap();
+    let file = fs::File::open(&pathb.as_path()).unwrap();
```
vext01, line 22: There are several places in the system where the stackmap section is being loaded from disk. At some point we should have ykllvm mark the section loadable and fish that data out of the virtual address space.

Can you add a FIXME for this please?

ykcapi/src/lib.rs
ptersilie: Both fixed in [c8af45b](https://github.com/ykjit/yk/pull/619/commits/c8af45b27186f79fa414ac5cffb345b1fd48077c)

ykfr/src/lib.rs
ptersilie: I'm not aware of other places where we parse the AOT stackmap. The only other place where `StackmapParser` is used is to get the stackmaps of a trace.

ptersilie:
Addressed all comments.

ykfr/src/lib.rs
vext01: Admittedly I only did a quick grep for the string `".llvm_stackmaps"`, which appears in many places.

vext01:
please squash

ptersilie:
Squashed.

PR #621, opened 2023-03-16
Add clangd support for our C++ code.

vext01:
Jump to definition now works reliably, even into LLVM headers.

### DIFF
--- a/ykllvmwrap/build.rs
+++ b/ykllvmwrap/build.rs
@@ -1,5 +1,48 @@
+use glob::glob;
 use rerun_except::rerun_except;
-use std::{env, process::Command};
+use std::{
+    env,
+    fs::File,
+    io::{Read, Write},
+    path::{Path, PathBuf},
+    process::Command,
+};
+use tempfile::TempDir;
+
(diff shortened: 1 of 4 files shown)


ykllvmwrap/build.rs
```diff
@@ -38,8 +80,19 @@ fn main() {
     if env::var("PROFILE").unwrap() == "release" {
         comp.flag("-DNDEBUG");
     }
+
+    // We use a compiler wrapper to capture compilation commands and generate a
+    // `compile_commands.json` database for clangd.
+    let td = TempDir::new().unwrap();
+    env::set_var("YKLLVMWRAP_JSON_TEMP", td.path());
```
ltratt, line 87: I guess there's no way to do this *just* for the ykllvmwrap subprocess?

vext01: Afraid not. I had hoped for a `compiler.env(key, val)`, but there isn't one.

ltratt:
OK with me. I'll leave this one to @ptersilie.

PR #626, opened 2023-03-21
Various suggestions from Clippy.

ltratt:
This doesn't fix everything, but it does reduce what remains to something more reasonable.

### DIFF
--- a/hwtracer/src/decode/ykpt/mod.rs
+++ b/hwtracer/src/decode/ykpt/mod.rs
@@ -356,13 +356,11 @@ impl<'t> YkPTBlockIterator<'t> {
                 // `seek_tnt()` has populated `self.tnts()`.
                 let target_off = if self.tnts.pop_front().unwrap() {
                     *taken_target
+                } else if let Some(ntt) = not_taken_target {
+                    *ntt
                 } else {
-                    if let Some(ntt) = not_taken_target {
-                        *ntt
-                    } else {
-                        // Divergent control flow.
-                        todo!();
-                    }
+                    // Divergent control flow.
+                    todo!();
                 };
                 self.cur_loc = ObjLoc::MainObj(target_off);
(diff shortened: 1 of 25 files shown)


hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -415,16 +413,16 @@ impl<'t> YkPTBlockIterator<'t> {
                     // If there are calls in the block that come *after* the current position in the
                     // block, then we will need to follow those before we look at the successor info.
                     if let Some(blk) = self.maybe_follow_blockmap_call(b_off, &ent.value)? {
-                        return Ok(blk);
+                        Ok(blk)
```
vext01, line 416: I don't prefer this, but if that's what clippy suggests.

ltratt: I have a marginal preference for the adjusted version, just because it makes clear that this isn't early termination. [And, as you say, Clippy does suggest it!]

hwtracer/src/decode/ykpt/packet_parser/packets.rs
```diff
@@ -226,7 +223,7 @@ pub(in crate::decode::ykpt) struct ShortTNTPacket {
     /// being interpreted as a short TNT with no stop bit.
     #[deku(bits = "7", assert = "*branches != 0x1 && *branches != 0x0")]
     branches: u8,
-    #[deku(bits = "1", assert = "*magic == false", temp)]
+    #[deku(bits = "1", assert = "!*magic", temp)]
```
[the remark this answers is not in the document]
ltratt: Clippy gave a slightly botched warning for these, but I could work out what it was getting at without too much difficulty.

ltratt:
Looks like we're good to r+ this one?

PR #629, opened 2023-03-22
Simplify control flow.

ltratt:
Replace `if/elseif/else{assert(false)}` with `if/else{assert(...)}`.

### DIFF
--- a/ykllvmwrap/src/jitmodbuilder.cc
+++ b/ykllvmwrap/src/jitmodbuilder.cc
@@ -1536,7 +1536,12 @@ public:
           }
           assert(CallStack.curMappableFrame());
           tryStopOutlining();
-        } else if (UR->StackAdjust > 0) {
+        } else {
+          // If the stack size hasn't changed, either we've got something wrong,
+          // or the unmappable code has transitioned back to mappable code in a
+          // way we don't currently support.
+          assert(UR->StackAdjust > 0);
+
           // The stack got bigger as a result of executing the foreign code. It
           // must have called deeper and eventually into mappable code.
           //
@@ -1552,11 +1557,6 @@ public:
           auto [NextFunc, BB] = getLLVMAOTFuncAndBlock(NextIB);
           CallStack.pushFrame(
               StackFrame::CreateMappableFrame(NextFunc, nullptr));
-        } else {
-          // The stack size didn't change. That could indicate that foreign code
-          // is "falling through" into mappable code (which we don't allow).
-          // Either way, something has gone wrong.
-          assert(false);
         }
         continue;
       }


ykllvmwrap/src/jitmodbuilder.cc
```diff
@@ -1536,7 +1536,12 @@ class JITModBuilder {
           }
           assert(CallStack.curMappableFrame());
           tryStopOutlining();
-        } else if (UR->StackAdjust > 0) {
+        } else {
+          // If the stack size hasn't changed, either we've got something wrong,
+          // or the unmappable code has jumped back to mappable code. We could
+          // support the latter, but currently don't.
```
vext01, line 1542: s/tt/t/

vext01:
Just a typo

ykllvmwrap/src/jitmodbuilder.cc
ltratt: I believe "latter" is correctly spelt.

vext01: Ah, I thought you were saying "we could support this later", sorry.

"jumped back" isn't quite right though. it could be a forward of backward jump or simply a fall-thru. It's not necessarily "going back" to mappable code that we've seen before, it may be new code.

ltratt: s/back to/into/ ?

vext01: I'd just say "entered", since there may be no jump involved.

e.g.:

```
foreign_bb:
  nop

mappable_bb:
  nop
```

vext01: No, "entered" doesn't sound right either.

ltratt: I think "transitioned" is clearer than "entered" because the latter implies call/branch?

vext01: Yes, I like that.

So something like:

>  If the stack size hasn't changed it means we've transitioned to mappable code via some non-standard control flow (e.g. a jump or a fallthru). We could support this later, but currently don't.

vext01: (where "could" is optimistic)

ltratt: "Non-standard" seems incorrectly judgmental!

ltratt: See if 8a0ddba works for you.

vext01:
works for me, please squash away.

ltratt:
Squashed.

ykllvmwrap/src/jitmodbuilder.cc
vext01: This is fine.

(By non-standard, I meant what you call "BCF": bonkers control flow :) )

PR #630, opened 2023-03-23
For now crash if we encounter setjmp or _setjmp in unmapped code.

vext01:
Unfortunately we can only check this in the ykpt decoder, and only if the call is direct. It's better than nothing.

### DIFF
--- a/hwtracer/src/decode/ykpt/mod.rs
+++ b/hwtracer/src/decode/ykpt/mod.rs
@@ -49,10 +49,11 @@ use std::{
     cell::Cell,
     collections::VecDeque,
     convert::TryFrom,
+    ffi::CString,
     fmt::{self, Debug},
     ops::Range,
     path::{Path, PathBuf},
-    slice,
+    ptr, slice,
     sync::LazyLock,
 };
 use ykutil::{
@@ -501,6 +502,20 @@ impl<'t> YkPTBlockIterator<'t> {
             .map_err(|_| HWTracerError::DisasmFail("failed to set position".to_owned()))?;
         let mut reposition: bool = false;
 
+        // `as usize` below are safe casts from raw pointer to pointer-sized integer.
+        let longjmp_vaddr = {
+            let func = CString::new("longjmp").unwrap();
+            u64::try_from(unsafe { libc::dlsym(ptr::null_mut(), func.as_ptr()) } as usize).unwrap()
+        };
+        let us_longjmp_vaddr = {
+            let func = CString::new("_longjmp").unwrap();
+            u64::try_from(unsafe { libc::dlsym(ptr::null_mut(), func.as_ptr()) } as usize).unwrap()
+        };
+        let siglongjmp_vaddr = {
+            let func = CString::new("siglongjmp").unwrap();
+            u64::try_from(unsafe { libc::dlsym(ptr::null_mut(), func.as_ptr()) } as usize).unwrap()
+        };
+
         loop {
             let vaddr = usize::try_from(dis.ip()).unwrap();
             let (obj, off) = self.vaddr_to_off(vaddr)?;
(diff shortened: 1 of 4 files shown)


ltratt:
IMHO `setjmp` isn't the problem: `longjmp` is. Example: a trace calls `setjmp`, passes a reference to the buffer to unmappable code which calls `longjmp`.

vext01:
Interesting.

The case i was trying to cover was where the interpreter long jumps out of the trace into the middle of unmapped (thus outlined) code.

Are these then two subtly different horror stories?

ltratt:
Yes and no: `longjmp` is the culprit in both cases.

vext01:
I'm wondering if we've thought this through adequately.

Calls to `longjmp` in unmapped code could be fine, as long as they jump back to a place that is contained in the unmapped code...

If we reject all calls to `longjmp` in unmapped code (that's what you are asking for, isn't it?), we may incur false positives. Is that ok?

ltratt:
`longjmp` is not *always* bad, but it is clearly *sometimes* bad. Working out which case is which will require thought.

But, either way, I think this PR is incorrect in pointing the finger at innocent ol' `setjmp`. I suggest we close this PR and do a bit of thinking separately?

vext01:
Well I can easily reject unmapped code calling longjmp, if we are ok with false positives.

Doing anything more clever will require more intricate static or dynamic analyses. I'm not sure I want to get into that right now.

ltratt:
Agreed. Let's try borking on `longjmp` and seeing how much pain that causes us.

vext01:
Try this.

Considered writing a rust wrapper around `dlsym()` in ykutil, but then I'd have to make everything use it, and I'm trying to cut down on distractions :P

hwtracer/src/decode/ykpt/mod.rs
```diff
@@ -620,9 +625,10 @@ impl<'t> YkPTBlockIterator<'t> {
                     } else {
                         let target_vaddr = self.branch_target_vaddr(&inst);
 
-                        // We can't (yet) handle setjmp in foreign code. Crash hard if we spot it.
-                        debug_assert!(setjmp_vaddr == 0 || target_vaddr != setjmp_vaddr);
-                        debug_assert!(us_setjmp_vaddr == 0 || target_vaddr != us_setjmp_vaddr);
+                        // We can't (yet) handle longjmp in unmapped code. Crash if we spot it.
```
ltratt, line 628: These should be `assert` IMHO as the system is immediately incorrect if we notice this.

vext01: If we want them to be hard asserts we should make some effort to cache them.

vext01: We call `dlsym()` three times per disassembly session.

ltratt: That's up to you.

ltratt: Can we say "Encountered longjmp" in the `panic`?

[I also note, with mild amusement, that `panic` does not address the caching issue you were worried with!]

vext01: > [I also note, with mild amusement, that panic does not address the caching issue you were worried with!]

Like I said, you gave me the opportunity to be lazy :)

vext01: Fixed in fbce6a1bd6db912e0080bd44ba607a3612694d8d

ltratt:
Please squash and rebase.

vext01:
splat

ltratt:
Can you rebase too? Please force push.

vext01:
done

ltratt:
Please squash.

vext01:
splat.

PR #633, opened 2023-03-27
Unignore some tests.

ltratt:
Some of these had incorrect reasons given for ignoring them.

Interestingly, `bfbug.c` seems to cause our "nondeterministic" bug to happen deterministically (well, it happened 5 times in a row, and that's more failures than I've seen in the other sometime-failing test!).

I also noticed that `smmultisrc.c` occasionally fails with:

```
src/jitmodbuilder.cc:1543: Module *JITModBuilder::createModule(): Assertion `UR->StackAdjust > 0' failed.
```


vext01:
Great stuff!

bors r+

PR #635, opened 2023-03-29
More Clippy fixes

ltratt:
The first of these required a manual change (https://github.com/ykjit/yk/commit/5bc2998e5317bc5f95f297290d427fa99603a171) and the rest are fairly mechanical (https://github.com/ykjit/yk/commit/f35941df4e883b1bc70f3a8e3aa45534b2034b77).

### DIFF
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -13,7 +13,6 @@
 #[cfg(feature = "yk_testing")]
 mod testing;
 
-use libc;
 use std::arch::asm;
 use std::convert::{TryFrom, TryInto};
 use std::ffi::{c_char, c_void, CString};
@@ -25,6 +24,7 @@ use ykrt::{HotThreshold, Location, MT};
 use yksmp::{Location as SMLocation, StackMapParser};
 
 #[no_mangle]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
 pub extern "C" fn yk_mt_new(err_msg: *mut *const c_char) -> *mut MT {
     match MT::new() {
         Ok(mt) => Box::into_raw(Box::new(mt)),
@@ -45,6 +45,7 @@ pub extern "C" fn yk_mt_new(err_msg: *mut *const c_char) -> *mut MT {
 }
 
 #[no_mangle]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
 pub extern "C" fn yk_mt_drop(mt: *mut MT) {
     unsafe { Box::from_raw(mt) };
 }
@@ -58,6 +59,7 @@ pub extern "C" fn yk_mt_control_point(_mt: *mut MT, _loc: *mut Location) {
 // The "real" control point, that is called once the interpreter has been patched by ykllvm.
 // Returns the address of a reconstructed stack or null if there wasn#t a guard failure.
 #[no_mangle]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
 pub extern "C" fn __ykrt_control_point(
     mt: *mut MT,
     loc: *mut Location,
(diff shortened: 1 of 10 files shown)


ykcapi/src/lib.rs
```diff
@@ -25,6 +24,7 @@ use ykrt::{HotThreshold, Location, MT};
 use yksmp::{Location as SMLocation, StackMapParser};
 
 #[no_mangle]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
```
vext01, line 27: Is this related to: https://github.com/rust-lang/rust-clippy/issues/3045 ?

yksmp/src/lib.rs
```diff
@@ -40,6 +40,10 @@ pub struct LiveVar {
 }
 
 impl LiveVar {
+    pub fn is_empty(&self) -> bool {
```
vext01, line 43: What's this all about?

ykcapi/src/lib.rs
ltratt: Yes.

yksmp/src/lib.rs
ltratt: Clippy complains if you have `len` without `is_empty`. Sometimes I disable that lint, but if it's easy to implement, it is a standard Rust convention so...

ykcapi/src/lib.rs
vext01: Should we just let it warn? I worry we will forget to remove these annotations.

This would of course work against the idea of adding regular automated clippy checks...

ltratt: Notice this warning is only disabled for this one function.

vext01: Yes, I know. Still...

Also we have many instances of this for other functions.

ltratt: The warning only triggers on this function. We definitely want to be warned of other uses, hence why I restricted to disabling it on this function alone.

vext01: Yes, I understand. We are talking past each other maybe :)

If this function were to change such that the warning is no longer a false positive, but a true positive, then would it not be dangerous to have this annotation around?

ltratt: ?

vext01: The issue I linked explains that the clippy warning often triggers erroneously, as a false positive, right? As I understand, you are silencing a false positive here, right?

So what if a future code edit means that the clippy warning is actually justified, and that we've silenced a real concern?

vext01: I'm asking: should we *not* suppress the false positive. We don't have to fix every warning.

ltratt: What's the point of a warning that we then ignore? It's literally worse than suppressing it!

yksmp/src/lib.rs
vext01: Did you add the config file?

vext01: ah, crate level.

Not what I was expecting, but OK.

If we end up needing that annotation in multiple crates later, let's use a config file.

ykcapi/src/lib.rs
vext01: I give up. Please squash.

yksmp/src/lib.rs
ltratt: I think per-crate is slightly better myself, because it forces us to think a bit more about it. I don't feel strongly about it though.

ltratt:
Squashed.

ltratt:
Our old friend again. Retry?

vext01:
We really need to fix that.

bors r+

vext01:
Uh oh... again...

Perhaps I should pause what I'm doing and priorities fixing this?

PR #638, opened 2023-04-05
Fix some non-deterministic bugs.

vext01:
This fixes our non-deterministic failures.

The first commit is the meat, the next commit raises a question I'd like to investigate next.

I have an idea why this is happening, but don't want to address it in this same PR.

(this was hard work :( )

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -70,8 +70,11 @@ for mode in "" "--release"; do
     cargo build ${mode} -p ykcapi;
 done
 
-cargo test
-cargo test --release
+for i in $(seq 10); do
+    cargo test
+    cargo test --release
+done
+
 cargo bench
 
 # Run examples.
--- a/hwtracer/src/decode/ykpt/mod.rs
+++ b/hwtracer/src/decode/ykpt/mod.rs
@@ -701,6 +701,26 @@ impl<'t> YkPTBlockIterator<'t> {
         }
     }
 
+    /// Skip packets up until and including the next `PSBEND` packet. The first packet after the
+    /// `PSBEND` is returned.
+    fn skip_psb_plus(&mut self) -> Result<Packet, HWTracerError> {
+        loop {
+            if let Some(pkt_or_err) = self.parser.next() {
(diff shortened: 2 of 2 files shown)


ltratt:
This looks fine to me, but let's put `cargo test` in a loop inside `.buildbot.sh` for, say, 10 iterations, to catch "obvious" non-determinism. [Our overnight cron version of this should probably run things 1000 times or similar.]

ltratt:
Works for me! Merge as-is?

vext01:
yup

vext01:
OK, so this is caused by new Rust.


The `lazy_cell` fix is easy:

```diff
-#![feature(once_cell)]
+#![feature(lazy_cell)]
```

But this is scary:

```
thread '<unnamed>' panicked at 'misaligned pointer dereference: address must be a multiple of 0x8 but is 0x200b2d', ykutil/src/obj.rs:154:23
```

Investigating...

vext01:
The loader has placed the embedded bitcode at an unaligned address!

ltratt:
Yikes!

vext01:
It has probably always been this way, but Rust is now checking for misaligned pointer derefs it seems...

It's not clear to me why there is an alignment requirement here. I thought that misaligned reads were valid on x86_64, but I may be wrong? I know sparc64 would outright crash...

You will probably tell me it is UB I imagine :)

ltratt:
I think "Rust the language" defines misaligned reads/writes as UB, no matter the underlying platform. [IMHO this is the correct definition: you don't want people to write code that will explode in weird ways on different platforms!]

vext01:
Here's a discussion I had on the Rust Zulip about this:

https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/misaligned.20pointer.20deref.20after.20updating.20Rust.2E

And here's an issue about the error message not being quite right:

https://github.com/rust-lang/rust/issues/109997

vext01:
The failures in this PR have been addressed separately in https://github.com/ykjit/ykllvm/pull/60 and https://github.com/ykjit/yk/pull/639

PR #672, opened 2023-05-05
Build ykllvm from submodule

jacob-hughes:
This builds ykllvm using the build.rs script in ykbuild. Build args can
be passed to cmake using the `YKLLVM_BUILD` environment variable. For
example:

    `YKLLVM_BUILD_ARGS="-DLLVM_CACHE_BUILD=ON,-DLLVM_CCACHE_DIR=/opt/ccache" cargo build`

A release build of ykllvm is always built, however ykllvm will be built
with assertions when using cargo's debug profile.

If for whatever reason you want to point to your own version of ykllvm,
you can use the `YKLLVM_INSTALL_DIR=/path/to/ykllvm` environment variable
and ykllvm will not be built and deps such as ykllvmwrap will use the
`YKLLVM_INSTALL_DIR` path instead.

### DIFF
--- a/ykbuild/Cargo.toml
+++ b/ykbuild/Cargo.toml
@@ -4,7 +4,15 @@ version = "0.1.0"
 authors = ["The Yk Developers"]
 edition = "2021"
 license = "Apache-2.0 OR MIT"
+# The following makes ykbuild's cargo metadata available to any crate which
+# depends on it, providing depedents with the location of the ykllvm install dir.
+links = "ykbuild"
 
 [dependencies]
 glob = "0.3.1"
 tempfile = "3.3.0"
+
+[build-dependencies]
+cmake = "0.1"
+num_cpus = "1.0"
+which = "4.4.0"
--- a/ykbuild/build.rs
+++ b/ykbuild/build.rs
@@ -1,0 +1,54 @@
+use std::env;
+use which::which;
+
+const YKLLVM: &str = "../ykllvm/llvm";
+const PROFILE: &str = "Release";
+
+fn main() {
+    if env::var("YKB_YKLLVM_INSTALL_DIR").is_ok() {
+        println!(
+            "cargo:ykllvm={}",
+            env::var("YKB_YKLLVM_INSTALL_DIR").unwrap()
+        );
+        return;
+    }
+
+    // Ninja builds are faster, so use this if it's available.
+    let use_ninja = which("ninja").is_ok();
+    let is_debug = env::var("PROFILE").unwrap() == "debug";
+    let nprocs = format!("-j {}", num_cpus::get());
+
+    let mut ykllvm = cmake::Config::new(YKLLVM);
+    ykllvm
+        .profile(PROFILE)
+        .generator(if use_ninja { "Ninja" } else { "Unix Makefiles" })
+        .define("LLVM_INSTALL_UTILS", "ON")
+        .define("BUILD_SHARED_LIBS", "ON")
+        .define("LLVM_ENABLE_PROJECTS", "lld;clang")
+        .define(
+            "LLVM_ENABLE_ASSERTIONS",
+            if is_debug { "ON" } else { "OFF" },
+        )
+        // Due to an LLVM bug, PIE breaks our mapper, and it's not enough to pass
+        // `-fno-pie` to clang for some reason:
+        // https://github.com/llvm/llvm-project/issues/57085
+        .define("CLANG_DEFAULT_PIE_ON_LINUX", "OFF")
+        .build_arg(nprocs);
+
+    if let Ok(args) = env::var("YKB_YKLLVM_BUILD_ARGS") {
+        for arg in args.split(",") {
+            if !arg.starts_with("-D") {
+                panic!("YKB_YKLLVM_BUILD_ARGS must only contain -D arguments");
+            }
+            let (k, v) = arg.strip_prefix("-D").unwrap().split_once("=").unwrap();
+            ykllvm.define(k, v);
+        }
+    }
+
+    let dsp = ykllvm.build();
+
+    // We need to be able to locate the ykllvm install llvm bins from other
+    // crates, so this sets a `DEP_YKBUILD_YKLLVM` env var which can be accessed
+    // from any other crate in the yk workspace.
+    println!("cargo:ykllvm={}/bin/", dsp.display())
+}
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -38,8 +38,6 @@ test -d book
 cd ..
 
 # Build LLVM for the C tests.
-mkdir -p target && cd target
-git clone https://github.com/ykjit/ykllvm
 cd ykllvm
 mkdir build
 cd build
@@ -59,13 +57,17 @@ cmake -DCMAKE_INSTALL_PREFIX=`pwd`/../inst \
 cmake --build .
 cmake --install .
 export PATH=`pwd`/../inst/bin:${PATH}
-cd ../../..
+cd ../../
 
 # Check that clang-format is installed.
 clang-format --version
 # Check C/C++ formatting using xtask.
 cargo xtask cfmt
-git diff --exit-code
+
+# This is used to check clang-tidy output, but the dirty submodule from building
+# ykllvm is also shown.
+# FIXME: Add build/ to .gitignore in ykllvm
+git diff --exit-code --ignore-submodules
 
 # Check that building `ykcapi` in isolation works. This is what we'd be doing
 # if we were building release binaries, as it would mean we get a system
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "ykllvm"]
+	path = ykllvm
+	url = https://github.com/ykjit/ykllvm
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -6,6 +6,7 @@ fn ignore_dir(entry: &DirEntry) -> bool {
     if entry.path().starts_with("./target")
         || entry.path().starts_with("./.cargo")
         || entry.path().starts_with("./.git")
+        || entry.path().starts_with("./ykllvm")
     {
         return false;
     }
--- a/ykbuild/wrap-clang++.sh
+++ b/ykbuild/wrap-clang++.sh
@@ -6,6 +6,8 @@
 
 set -e
 
+export PATH=${DEP_YKBUILD_YKLLVM}:${PATH}
+
 echo "clang++ $@" > $(mktemp -p ${YK_CC_TEMPDIR})
 
 clang++ $@
--- a/ykllvmwrap/Cargo.toml
+++ b/ykllvmwrap/Cargo.toml
@@ -9,6 +9,7 @@ license = "Apache-2.0 OR MIT"
 cc = "1.0.72"
 libc = "0.2.117"
 ykutil = { path = "../ykutil" }
+ykbuild = { path = "../ykbuild" }
 
 [build-dependencies]
 cc = "1.0.72"


ykbuild/Cargo.toml
```diff
@@ -4,7 +4,13 @@ version = "0.1.0"
 authors = ["The Yk Developers"]
 edition = "2021"
 license = "Apache-2.0 OR MIT"
+links = "ykbuild"
```
ltratt, line 7: Doesn't this mean ykbuild links with a library called ykbuild? Is that necessary?

ykbuild/build.rs
```diff
@@ -0,0 +1,52 @@
+use cmake;
+use std::env;
+use num_cpus;
+use which::which;
+
+const YKLLVM: &str = "../ykllvm/llvm";
+const PROFILE: &str = "Release";
+
+fn main() {
+
```
ltratt, line 10: Extraneous blank line.

ykbuild/build.rs
```diff
@@ -0,0 +1,52 @@
+use cmake;
```
ltratt, line 1: I *think* that `use <crate>;` on its own has no effect in Rust so the first three `use`s can go.

ykbuild/build.rs
```diff
@@ -0,0 +1,52 @@
...
+        .define("LLVM_INSTALL_UTILS", "ON")
+        .define("LLVM_ENABLE_PROJECTS", "lld;clang")
+        .define("LLVM_ENABLE_ASSERTIONS", if is_debug { "ON" } else { "OFF" })
+        // Due to an LLVM bug, PIE breaks our mapper, and it's not enough to pass
+        // `-fno-pie` to clang for some reason:
+        // https://github.com/llvm/llvm-project/issues/57085
+        .define("CLANG_DEFAULT_PIE_ON_LINUX", "OFF")
+        .build_arg(nprocs);
+
+    match env::var("YKLLVM_BUILD_ARGS") {
```
ltratt, line 33: I think this will be easier as `if let Ok(args) = env::var("...")`?

ykbuild/Cargo.toml
jacob-hughes: Not quite, this is a hack so that we can define an `DEP_YKBUILD_YKLLVM` env var in `ykbuild`'s build.rs which is available for any other crate in the yk workspace which depends on it. This then lets other crates know where the install location of the ykllvm bins are.

See https://doc.rust-lang.org/cargo/reference/build-script-examples.html#using-another-sys-crate

jacob-hughes:
This now uses the `YKB` prefix for environment variables: 48772d0dc8386e75259895b4c563e321f62b4297

ykbuild/build.rs
jacob-hughes: Fixed 14793f44003e305148f38689fbc55733e2eed6b2

ykbuild/Cargo.toml
vext01: Can you add a comment explaining this. It is quite surprising at a glance.

ykbuild/build.rs
```diff
@@ -0,0 +1,51 @@
...
+    let use_ninja = which("ninja").is_ok();
+    let is_debug = env::var("PROFILE").unwrap() == "debug";
+    let nprocs = format!("-j {}", num_cpus::get());
+
+    let mut ykllvm = cmake::Config::new(YKLLVM);
+    ykllvm.profile(PROFILE)
+        .generator(if use_ninja { "Ninja" } else { "Unix Makefiles" })
+        .define("LLVM_INSTALL_UTILS", "ON")
+        .define("LLVM_ENABLE_PROJECTS", "lld;clang")
+        .define("LLVM_ENABLE_ASSERTIONS", if is_debug { "ON" } else { "OFF" })
```
vext01, line 25: This needs `-DBUILD_SHARED_LIBS=ON`.

See https://github.com/ykjit/ykllvm/pull/67.

ykbuild/build.rs
jacob-hughes: Fixed here 4371375ffbc3d273b344e1df8f9f9bc9c85f2c24

ykbuild/build.rs
```diff
@@ -0,0 +1,51 @@
...
+    // Ninja builds are faster, so use this if it's available.
+    let use_ninja = which("ninja").is_ok();
+    let is_debug = env::var("PROFILE").unwrap() == "debug";
+    let nprocs = format!("-j {}", num_cpus::get());
+
+    let mut ykllvm = cmake::Config::new(YKLLVM);
+    ykllvm.profile(PROFILE)
+        .generator(if use_ninja { "Ninja" } else { "Unix Makefiles" })
+        .define("LLVM_INSTALL_UTILS", "ON")
+        .define("LLVM_ENABLE_PROJECTS", "lld;clang")
```
vext01, line 27: `clangd` is in `clang-tools-extra`. Should we add it? Without it the LSP won't work for our C/C++ code.

The user would also have to make sure this `clangd` is in `$PATH` when starting their editor (or provide their own `clangd`).

ykbuild/build.rs
jacob-hughes: Fixed 4dc0284131b98b7c2641863a68b73adde28d6c39

vext01:
How does this work for interpreters downstream of `yk`. Can their CI share the ykllvm that yk builds?

vext01:
> YKLLVM_BUILD_ARGS="-DLLVM_CACHE_BUILD=ON,-DLLVM_CCACHE_DIR=/opt/ccache"

Why are the args comma separated? I imagine this is why only -D stuff is supported? If they were space separated, wouldn't you be able to pass them directly on to cmake?

Maybe I've overlooked something?

vext01:
To build ykllvm with distcc effectively, I'd need the ability to pass a `-j` to ninja/make.

(those tools don't know that a distcc setup has more cores than the local machine has)

Is there a way to append a `-j`?

vext01:
Maybe that amounts to passing a `-j` to cmake? e.g. `cmake --build build -j40`?

jacob-hughes:
> How does this work for interpreters downstream of `yk`. Can their CI share the ykllvm that yk builds?

I'm not sure I understand this. Can you elaborate?

vext01:
By default cmake will run the build with as many cores as you have on your system.

(side note, I see you manually do this -- I don't think it's necessary `let nprocs = format!("-j {}", num_cpus::get());`)

When I'm using distcc, I can borrow cores from other machines. For example `DISTCC_HOSTS="localhost/8 remote1/8 beefy/36"` means that I have 52 cores available for compilation jobs. If we allow cmake to automatically choose a `-j` it will choose too few (8) and make the build slower than necessary.

ykbuild/Cargo.toml
jacob-hughes: Sure, how about this 62cc65dd562b90ff64dd544915e1b1554d6559c5?

jacob-hughes: Sure, how about this 62cc65dd562b90ff64dd544915e1b1554d6559c5?

vext01:
> I'm not sure I understand this. Can you elaborate?

Just realised that I elaborated on the wrong thing!

If we look at the CI script for an interpreter, we can see that they build their own ykllvm. e.g.:
https://github.com/ykjit/ykcbf/blob/main/.buildbot.sh#L22

My question is, can we re-use the one `yk` will build and avoid building twice (or 4 times if you factor in with and without `--release`)?

jacob-hughes:
> > I'm not sure I understand this. Can you elaborate?
> 
> Just realised that I elaborated on the wrong thing!
> 
> If we look at the CI script for an interpreter, we can see that they build their own ykllvm. e.g.: https://github.com/ykjit/ykcbf/blob/main/.buildbot.sh#L22
> 
> My question is, can we re-use the one `yk` will build and avoid building twice (or 4 times if you factor in with and without `--release`)?

It shouldn't be necessary to build ykllvm in buildbot because `cargo build` should "just work". I've removed this and if it doesn't break CI then we know it's working (39baf3adc3a553d21223c67316db2a08bff9abeb).

vext01:
> It shouldn't be necessary to build ykllvm in buildbot because cargo build should "just work". I've removed this and if it doesn't break CI then we know it's working (https://github.com/ykjit/yk/commit/39baf3adc3a553d21223c67316db2a08bff9abeb).

I understand, but we need to do something for:
  - [ykcbf](https://github.com/ykjit/ykcbf/blob/main/.buildbot.sh#L22)
  - [yklua](https://github.com/ykjit/yklua/blob/bcfc63282355011c8268acd9f545cbc6f39837df/.buildbot.sh#L21)

These are external projects that depend upon `yk`, but which at present build their own ykllvm. We don't want that in the long run.

jacob-hughes:
> > It shouldn't be necessary to build ykllvm in buildbot because cargo build should "just work". I've removed this and if it doesn't break CI then we know it's working ([39baf3a](https://github.com/ykjit/yk/commit/39baf3adc3a553d21223c67316db2a08bff9abeb)).
> 
> I understand, but we need to do something for:
> 
> * [ykcbf](https://github.com/ykjit/ykcbf/blob/main/.buildbot.sh#L22)
> * [yklua](https://github.com/ykjit/yklua/blob/bcfc63282355011c8268acd9f545cbc6f39837df/.buildbot.sh#L21)
> 
> These are external projects that depend upon `yk`, but which at present build their own ykllvm. We don't want that in the long run.

They will need PRs to reflect these changes I think

jacob-hughes:
bors try

jacob-hughes:
bors try

ltratt:
```
clang-format --version
clang-format: not found
```

Sounds like a PATH issue?

ltratt:
After a chat with @jacob-hughes this morning, we've decided to go ahead with this PR, even though we know there are various situations where it causes multiple versions of LLVM to be built. The reason for this is a) I still believe this is an improvement on the current situation where yk is prohibitively hard to build b) the environment variables we're defining will be backwards compatible if we use a different approach in the future c) we can see how bad the "multiple LLVMs" thing is in practise and refine this approach in the future.

jacob-hughes:
bors try

vext01:
@jacob-hughes does the build script running multiple times also mean that the install step is run multiple times? How big is an LLVM install?

ltratt:
I think @jacob-hughes understands when (which is not quite the same as "why") LLVM will be built multiple times. I think that's something we'll have to live with for a little bit. After all, you can solve this by using an environment variable to point to your own (single) copy of LLVM if you find it annoying.

vext01:
I do find it annoying indeed.

(We also discussed that the problem would disappear if we didn't use a workspace)

vext01:
> Define the env var and you won't have multiple copies :)

The problem is, you can have one out of date copy (you forget to rebuild after the submodule updates), which is what this PR aims to solve.

ltratt:
>The problem is, you can have one out of date copy (you forget to rebuild after the submodule updates), which is what this PR aims to solve.

What this PR really aims to do is to make yk buildable by mortals. If you are annoyed by multiple LLVM copies, then the env var allows you to restore the pre-PR behaviour -- which does, yes, put a burden on you, but it's no worse than it is before the PR.

vext01:
I mildly disagree that it is no worse (disk space, time), but get this in and I will investigate in conference downtime.

jacob-hughes:
bors try

ltratt:
```
thread 'main' panicked at 'clang-format failed on ./ykllvm/clang/lib/Format/MacroCallReconstructor.cpp', xtask/src/main.rs:33:25
```

jacob-hughes:
bors try

jacob-hughes:
bors try-

jacob-hughes:
bors try -

jacob-hughes:
bors try

jacob-hughes:
But I cancelled it

jacob-hughes:
bors try-

jacob-hughes:
bors try

jacob-hughes:
bors try

ltratt:
If the `try` works, please squash.

jacob-hughes:
bors try

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #678, opened 2023-05-11
Add the ability to pass arbitrary arguments to ykllvm build.

vext01:
Before every element in `YKB_YKLLVM_BUILD_ARGS` was assumed to be a `-D` argument passed to cmake configure step via `cmake::Config::define()` (from the `cmake` crate).

This change allows the user to pass other kinds of arguments and to different stages of the cmake dance.

Each element in `YKB_YKLLVM_BUILD_ARGS` is now of the form:

  `kind:arg[=val]`

NOTE: I've not written/updated any docs for this, as I think @ltratt is working on that separately(?)
Where `kind` specifies which method of `cmake::Config` the remainder of the element applies to.

For example:

```
YKB_YKLLVM_BUILD_ARGS="define:LLVM_CCACHE_BUILD=On,define:LLVM_CCACHE_DIR=/opt/ccache,build_arg:-j32"
```

Results in the build script calling:

```
ykllvm.define("LLVM_CCACHE_BUILD", "On");
ykllvm.define("LLVM_CCACHE_DIR", "/opt/ccache");
ykllvm.build_arg("-j32");
```

The former two being configure-time arguments, the latter being build-time.

Currently only the `define`, `build_arg` and `generator` kinds are supported, but it woulnd't be hard to add more.


ykbuild/build.rs
```diff
@@ -36,12 +36,25 @@ fn main() {
...
     if let Ok(args) = env::var("YKB_YKLLVM_BUILD_ARGS") {
+        // Caveat: this assumes no cmake argument contains a ',' or a ':'.
         for arg in args.split(",") {
-            if !arg.starts_with("-D") {
-                panic!("YKB_YKLLVM_BUILD_ARGS must only contain -D arguments");
+            let elems = arg.split(":").collect::<Vec<_>>();
+            if elems.len() != 2 {
+                panic!("Failed to parse YKB_YKLLVM_BUILD_ARGS");
+            }
+            match elems[0] {
```
ltratt, line 45: I prefer writing this sort of thing as:

```
match arg.split(":").collect::<Vec<_>>() {
  ["define", v] => ...,
  ["build_arg", v] => ...,
  ["generator", v] => ...
  [x, _] => panic!("Unknown kind {x}"),
  x => panic!("Incorrectly formatted option {x}")
}
```

because it keeps the splitting and whatnot more together. Up to you which you prefer!

vext01: I prefer that too! I didn't know you could match like that!

vext01: Fixed in [f3ebb76](https://github.com/ykjit/yk/pull/678/commits/f3ebb76b8b1e5baea268d2256295636da8eb4d55)

vext01: [87ea181](https://github.com/ykjit/yk/pull/678/commits/87ea1814ca55792b690c9d46f711f2a3583d6b22) completes the fix, although the build is still failing for me locally due to the `PATH` issue we discussed in #679 (it doesn't find `llvm-config`. I guess CI will work because it sets PATH.

vext01:
splat

ltratt:
Nice try :) You get to do the documentation. PR coming shortly.

vext01:
It was worth a shot ;)

PR #680, opened 2023-05-11
Rename `YKB_YKLLVM_INSTALL_DIR` to `YKB_YKLLVM_BIN_DIR`.

ltratt:
`YKB_YKLLVM_INSTALL_DIR` is ambiguous: is it the "common root" of the installation (e.g. `/usr/local`) or where the binaries are stored (e.g. `/usr/local/bin`). `YKB_YKLLVM_BIN_DIR` makes clear that it's the latter. As discussed on MM.

### DIFF
--- a/docs/src/dev/env.md
+++ b/docs/src/dev/env.md
@@ -13,6 +13,18 @@ testing only. It is enabled whenever the `tests` crate is being compiled, so a
 regular `cargo build` in the root of the workspace will enable the feature (to
 build *without* the feature enabled, do `cargo build -p ykcapi`).
 
+## Build-time Variables
+
+### `YKB_YKLLVM_BIN_DIR`
+
+Under normal circumstances, yk builds a copy of its LLVM fork "ykllvm" and uses
+it to build interpreters. You can use your own ykllvm build by specifying the
+directory where the executables (e.g. `clang`, `llvm-config`, and so on) are
+stored with `YKB_YKLLVM_BIN_DIR`. yk does not check your installation for
+compatibility: it is your responsibility to ensure that your ykllvm build
+matches that expected by yk.
+
+
 ## Run-time Variables
 
 ### `YKD_FORCE_TRACE_DECODER`
--- a/ykbuild/build.rs
+++ b/ykbuild/build.rs
@@ -5,11 +5,8 @@ const YKLLVM: &str = "../ykllvm/llvm";
 const PROFILE: &str = "Release";
 
 fn main() {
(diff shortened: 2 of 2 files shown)


docs/src/dev/env.md
```diff
@@ -13,6 +13,18 @@ testing only. It is enabled whenever the `tests` crate is being compiled, so a
 regular `cargo build` in the root of the workspace will enable the feature (to
 build *without* the feature enabled, do `cargo build -p ykcapi`).
 
+## Build-time Variables
+
+### `YKB_YKLLVM_BIN_DIR`
+
+Under normal circumstances, yk builds a copy of its LLVM fork "ykllvm" and uses
+it to build interpreters. You can use your own ykllvm build by specifying the
```
vext01, line 21: > and uses it to build interpreters

I'm not sure about this bit. Which interpreters do you mean here? yklua and ykcbf?

vext01:
I agree about the ambiguity.

One small query in the docs.

docs/src/dev/env.md
vext01: This repo doesn't build those interpreters, so I think we should remove that part.

ltratt: Er, actually, that's kind of a good point. At the moment it's a mess: we expect the user to use the *correct* `clang` for ykllvm. Arguable we should have `yk-config --cc` which gives you the path to our `clang` so you can do something like this in a Makefile:

```
CC=`yk-config --cc`
```

What do you think?

ltratt: I suspect (but am not 100% sure) I can do this in the PR I'm not working on.

I suggest we keep this documentation as-is, because it's sort-of true now, and we can make it better in the next PR.

jacob-hughes: +1. That works for me

jacob-hughes:
LGTM

jacob-hughes:
@vext01  happy for this to be merged?

vext01:
> I suspect (but am not 100% sure) I can do this in the PR I'm not working on.

Ah, I thought you meant in this PR, but now I realise you meant another PR.

Yes, I'm OK with this one as long as the `yk-config -cc` change comes soon.

ltratt:
It is coming.

vext01:
No squashing to do then?

vext01:
okie dokie then!

bors r+

ltratt:
Oops! OK to squash?

vext01:
Go ahead.

PR #681, opened 2023-05-11
Build ykllvm in `target/<debug|release>/ykllvm`.

ltratt:
Previously we tried building ykllvm in `OUT_DIR` but, unfortunately, that means when testing we can't access the Cargo variable we defined in https://github.com/rust-lang/cargo/issues/10927. Previously, things accidentally worked because we also required PATH to be set to `/path/to/ykllvm/bin`, but that means you could pick up unexpected versions of clang/clang++, for example.

This commit rethinks things. It builds a copy of ykllvm in `target/<debug|release>/ykllvm`. [Technically this breaks Cargo's rules, but since we're having to worm around a Cargo restriction, I think we can deal with this: if Cargo's rules change in the future, we can change the location easily enough.] `yk-config` gains `--cc` and `--cxx` flags which give the user the path to the C/C++ compiler.

I've tried to keep this commit as minimal as possible: it would certainly be possible to do various bits of cleaning-up at the same time, but I think that would obscure what we're most interested in.


ltratt:
bors try

docs/src/dev/env.md
```diff
@@ -17,12 +17,19 @@ build *without* the feature enabled, do `cargo build -p ykcapi`).
...
 ### `YKB_YKLLVM_BIN_DIR`
 
-Under normal circumstances, yk builds a copy of its LLVM fork "ykllvm" and uses
-it to build interpreters. You can use your own ykllvm build by specifying the
-directory where the executables (e.g. `clang`, `llvm-config`, and so on) are
-stored with `YKB_YKLLVM_BIN_DIR`. yk does not check your installation for
-compatibility: it is your responsibility to ensure that your ykllvm build
-matches that expected by yk.
+Under normal circumstances, yk builds a copy of its LLVM fork "ykllvm", which
+it also uses it to build interpreters (via the compilers `yk-config`). You can
```
jacob-hughes, line 21: s/compilers/compiler's?

ltratt: Oops, that bracket bit doesn't parse, full stop! Fixed in 6f96334.

docs/src/dev/env.md
```diff
@@ -17,12 +17,19 @@ build *without* the feature enabled, do `cargo build -p ykcapi`).
...
+(e.g. `clang`, `llvm-config`, and so on) are stored with
+`YKB_YKLLVM_BIN_DIR`.
+
+yk does not check your installation for compatibility: it is your
+responsibility to ensure that your ykllvm build matches that expected by yk.
+
+It is also undefined behaviour to move between defining this variable and not
+within a repository using `yk` (including the `yk` repository itself). If you
+want to set/unset `YKB_YKLLVM_BIN_DIR` then `cargo clean` any repositories
+using `yk` before rebuilding them.
```
jacob-hughes, line 32: I'm not sure I understand this paragraph. Is it saying you can't build one package using `YKB_YKLLVM_BIN_DIR` and then another without it?

ykbuild/wrap-clang++.sh
```diff
@@ -6,8 +6,5 @@
 
 set -e
 
-export PATH=${DEP_YKBUILD_YKLLVM}:${PATH}
-
-echo "clang++ $@" > $(mktemp -p ${YK_CC_TEMPDIR})
-
-clang++ $@
+echo "${YK_COMPILER_PATH} $@" > $(mktemp -p ${YK_COMPILER_TEMPDIR})
```
jacob-hughes, line 9: Isn't this going to grab the `clang` binary instead of the `clang++` one?

ltratt: I'm about 95% certain that we don't need *two* wrappers: we just need one and it can call `clang` or `clang++` depending on what `YK_COMPILER_PATH` is set to. Warning: I might be wrong. But I think we can tackle this in a follow-up clean-up PR.

ykllvmwrap/build.rs
```diff
@@ -49,6 +54,7 @@ fn main() {
     // Generate a `compile_commands.json` database for clangd.
     let ccg = CCGenerator::new("ykllvmwrap", &env::var("CARGO_MANIFEST_DIR").unwrap());
     env::set_var.call(ccg.build_env());
+    env::set_var("YK_COMPILER_PATH", ykllvm_bin("clang"));
```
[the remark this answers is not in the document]
ltratt: It should! Although... it worked anyway?! Maybe clang adapts based on flags or whatnot? Anyway, fixed in ea24a1f.

ykbuild/wrap-clang++.sh
jacob-hughes: I think you're right actually. However, I think it's calling the wrong binary from ykllvmwrap (https://github.com/ykjit/yk/pull/681#discussion_r1191575427)

jacob-hughes: Ah, I see you replied. Ignore me.

ykllvmwrap/build.rs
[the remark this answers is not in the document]
jacob-hughes: Ah, I didn't know that. Cool

jacob-hughes:
I'm happy for you to squash, but maybe you want to wait in case @vext01 had comments.

ltratt:
bors try

ltratt:
I think we're picking up more places where in CI we don't have the same set of commands as on a typical user system. This is good... but will be frustrating to fix as I can't replicate this easily on a normal machine. It might take a while for me to find all the right places.

ykbuild/wrap-clang++.sh
vext01: Perhaps this script should be named `wrap_compiler.sh`?

ltratt: Yes -- this is one of several clean-up opportunities I've identified. But I think that's for subsequent PR(s).

vext01: and delete `wrap-clang.sh`?

ltratt:
bors try

ykllvmwrap/build.rs
```diff
@@ -58,12 +64,22 @@ fn main() {
...
+        .arg("--libdir")
+        .output()
+        .unwrap()
+        .stdout;
     let lib_dir = std::str::from_utf8(&lib_dir).unwrap();
     println!("cargo:rustc-link-search={}", lib_dir);
     apply_llvm_ld_library_path();
 
-    let libs = llvm_config().arg("--libs").output().unwrap().stdout;
+    let libs = Command::new(&ykllvm_bin("llvm-config"))
```
vext01, line 77: How comes we removed `llvm_config()` as a function? I don't feel strongly, but it made sense to have one place to apply `--link-shared`.

ykbuild/wrap-clang++.sh
vext01: ok

vext01:
This looks good. I only had a couple of small comments.

When this is merged I think I can stop (if this is still the case) `cargo test` from re-running `cmake`.

ykllvmwrap/build.rs
ltratt: Partly a matter of taste, but it felt weird to have `llvm_config() -> Command` and then we'd still have to have `ykllvm_bin(...) -> Path` or similar.

docs/src/dev/env.md
ltratt: Basically you can't switch your ykllvm directory on yk without weird things happening (shared libraries not being in the right place etc). Happy to take suggestions for rephrasing!

ltratt:
> When this is merged I think I can stop (if this is still the case) cargo test (after cargo build) from re-running cmake.

Ah, cool, I was also going to look into that, but my solution has a definite sledgehammer feel to it, so you might have something better!

vext01:
> Ah, cool, I was also going to look into that, but my solution has a definite sledgehammer feel to it

Well, what's you solution? In case it's the same as mine!

ltratt:
I was going to store the hash of the ykllvm submodule (assuming `YKB_YKLLVM_BIN_DIR` isn't set) and only run `cmake` if that hash changes. Like I said -- a sledgehammer!

ltratt:
Um, maybe? Dunno! Can't hurt to try.

vext01:
OK, I'll try it when I can. I need to re-check my assumptions on the latest code.

ltratt:
bors try

ltratt:
bors try

ltratt:
bors try

ltratt:
bors try

ltratt:
bors try

ltratt:
bors try

jacob-hughes:
Sure, please squash.

ykbuild/build.rs
```diff
@@ -10,6 +17,17 @@ fn main() {
         return;
     }
 
+    // To avoid running cmake config/build/install, we only do anything with cmake if either a) no
+    // build of ykllvm exists b) the ykllvm submodule's hash has changed since we last built it.
+    //
+    // First we get the ykllvms submodule's hash.
+    let out = Command::new("git")
```
jacob-hughes, line 24: What will this do when you have a working tree inside the ykllvm submodule? I think that changes in the working tree would change the ykllvm commit hash to be `<hash>-dirty`. However, would this mean that further edits to an already dirty ykllvm tree would not trigger a rebuild? In such cases, do we need to add a rerun-if-changed command?

jacob-hughes: Sure, that makes sense.

jacob-hughes:
LGTM. I'll wait for @vext01 though

ltratt:
bors try

ltratt:
I was a bit fearful of this when I realised that yksmp's tests were merrily writing files in weird places: the way they're built really only ever worked by chance, and is now hugely incompatible with our new setup. I've spent an hour fixing them, and have simply ended up confusing myself. What I propose is deleting them for now and then, when at some point soon Lukas and I probably have to sit down and work out what they're really supposed to be doing, and then we can bring them back in.

bors try

ltratt:
bors try

ltratt:
@jacob-hughes @vext01 This is now ready for reviewing (and, maybe, squashing). The yksmp situation is something that we need to think about more carefully and I don't think this is the right PR to do it, as the tests previously only worked by accident for more than one reason (e.g. they always built in debug mode, even in release mode; they were placed in a directory that meant they were subject to race conditions; and the Makefile is hard to use for both "normal" testing and "cargo test" testing) -- I'm not even entirely sure what compiler they're supposed to be using. If you're willing to go with that, I suggest that I raise an issue so that we reinstate the tests, in one way or another, in a future PR.

ykbuild/build.rs
vext01: If the issue is still that `ykbuild` is both a build and normal dependency (cargo doesn't share deps between those), then [the solution I proposed earlier](https://github.com/ykjit/yk/pull/681#issuecomment-1544686556) may fix it naturally, only re-triggering a rebuild when a file changes.

I can try that tomorrrow, but I think this can go in in the meantime.

> at the moment this support is more aimed at "normal user of yk who doesn't know what to do with ykllvm"

I'd like to eventually get it so we can develop out of the submodule, as it would improve quality of life for developers too (they wouldn't have to worry about when and when not to rebuild ykllvm).

docs/src/dev/env.md
vext01: FWIW, I understood this part.

ykllvmwrap/build.rs
vext01: I guess I'm OK with it. We just have to remember to add `--link-shared` everywhere or we may statically link extra versions of llvm.

ykbuild/build.rs
```diff
@@ -58,6 +59,25 @@ fn main() {
...
+    // nightmare, especially if you try to do something clever. Fortunately, we're not trying to
+    // make something clever like a read-write lock: we just gain exclusive (which probably means
+    // "write" on most platforms) access. Unix and Windows both remove the lock if the process dies
+    // unexpectedly (though Windows says, in essence, that it doesn't guarantee to do the unlocking
+    // very quickly), so if we fail while building, this should not cause parallel runs of this
+    // build script to deadlock.
+    let mut lock_path = ykllvm_dir.clone();
+    lock_path.push("build_lock");
+    let lock_file = File::create(lock_path).unwrap();
+    lock_file.lock_exclusive().unwrap();
```
vext01, line 79: I think I may be able to kill this also. It's likely build-deps and run-deps building in parallel.

vext01:
LGTM as a base. Like I said, I think I have a plan to remove the hash caching and the lock file.

Why did some tests need to move?

ykbuild/build.rs
ltratt: > If the issue is still that ykbuild is both a build and normal dependency (cargo doesn't share deps between those), then https://github.com/ykjit/yk/pull/681#issuecomment-1544686556 may fix it naturally, only re-triggering a rebuild when a file changes.

Can't hurt to try!

ykbuild/build.rs
ltratt: Even if you can break those links (and I'm not sure it's possible), I would suggest keeping this: if anyone adds those links back in the future, this will stop themselves shooting them in the foot.

ltratt:
> Why did some tests need to move?

The commit message here might explain https://github.com/ykjit/yk/pull/681/commits/0b118f16cf10ee3e1d302f2268ea44ab59314b4a

ltratt:
Are we ready to squash or ... ?

jacob-hughes:
LGTM. If Edd’s happy please squash.

PR #686, opened 2023-05-15
We only need one compiler wrapper.

ltratt:
Since we set `YK_COMPILER_PATH` when calling the wrapper, there's no need to have different wrappers for different compilers. This allows us to remove a bit of complexity in ccgen too.

While I'm here, I have, self-indulgently, changed the name of the wrapper. I think `.sh` as a suffix is unhelpful, because a) I don't care what language an executable file is written in b) I might change that language later. I also initially called this "wrap-compiler" but that just felt a bit wrong: it should either be "wrap-compile" or, as I eventually settled upon as preferable, "compiler-wrapper". The latter is a bit longer, but feels like it's more likely to be understood.


vext01:
> I might change that language later. I also initially called this "wrap-compiler" but that just felt a bit wrong: it should either be "wrap-compile" or, as I eventually settled upon as preferable, "compiler-wrapper". The latter is a bit longer, but feels like it's more likely to be understood.

In order of preference for me:

 - `wrap-compiler` (says exactly what the script is)
 - `compiler-wrapper` (good, but longer)
 - `wrap-compile` (doesn't say how much of the compilation it wraps. All invocations? one invocation of the compiler?)

vext01:
LGTM.

Just have to agree on the script name.

ltratt:
`wrap-compiler` says what it does, but not what it is. I prefer `compiler-wrapper` myself.

vext01:
It's fine.

Resolve the merge conflict a force push.

ltratt:
I wonder why it says conflicts when `rebase` succeeds automatically? Anyway, force pushed!

PR #687, opened 2023-05-15
Shallow ykllvm clone

ltratt:


vext01:
The docs you are editing are the developer docs. Shouldn't we direct devs to `YKB_YKLLVM_BIN_DIR`?

ltratt:
Having looked at these things this afternoon, I think the `env` docs should probably be partly folded in to the "getting started" docs. But, honestly, I need to think slightly more carefully about this -- perhaps we need a slightly different structure to the whole docs full stop?

Either way, the current "getting started" instructions are completely out of date, so fixing them now to reflect reality feels like better than nothing.

vext01:
I'm OK with restructuring (in another PR later).

> Either way, the current "getting started" instructions are completely out of date, so fixing them now to reflect reality feels like better than nothing.

OK, but we should follow up soon. If I understood our meeting earlier, if a developer doesn't use `YKB_YKLLVM_BIN_DIR` and makes local edits to ykllvm, then it won't be rebuilt, and they should probably be aware of this.

ltratt:
> if a developer doesn't use YKB_YKLLVM_BIN_DIR and makes local edits to ykllvm, then it won't be rebuilt, and they should probably be aware of this.

I think that's a slightly different thing, but yes, documenting it in another PR can't hurt.

vext01:
As far as I know that's the only thing that causes a change of workflow between users and developers.

Yep, we can do that in another PR.

Anyway, this all looks OK.

@jacob-hughes OK with it?

docs/src/dev/getting_started.md
```diff
@@ -1,61 +1,28 @@
...
 
 The `yk` repo is a Rust workspace (i.e. a collection of crates). You can build
-and test in the usual ways using `cargo`.
-
-For example, to build and test the system, run:
+and test in the usual ways using `cargo`. For example, to build and test the
+system, run:
 
 ```
 cargo test
```
jacob-hughes, line 23: Minor nit: should this be `cargo test --release` for symmetry with `cargo build --release` above?

ltratt: In general, `cargo test` is a more thorough test than `cargo test --release` so I lean towards the former.

jacob-hughes: Fair enough, I don't feel that strongly about it.

PR #693, opened 2023-05-16
Talk about clangd and clang-format in the docs.

vext01:


ltratt:
> This assumes you have a `clang-format` in `$PATH`.

Hmm, I now realise that our earlier conversation only scratched the surface here. Rather than:

* burdening users with this warning
* the inevitable differences between external and ykllvm's `clang-format` s

why don't we build our own and automatically run that with `xtask fmt`? We'll win that back rather quickly, I suspect.

vext01:
> why don't we build our own and automatically run that with xtask fmt?

I think it should tie in with `YKB_YKLLVM_BIN_DIR` (if set) like the other places in the build system?

ltratt:
> I think it should tie in with YKB_YKLLVM_BIN_DIR (if set) like the other places in the build system?

s/should/must/ !

This should be fairly easy though, I hope. Make `xtask` depend on `ykbuild` then use `ykbuild::ykllvm_bin_dir("clang-format")` and it'll pick the right one up automatically.

xtask/src/main.rs
```diff
@@ -25,13 +26,19 @@ fn clang_format() {
         if let Some(ext) = entry.path().extension() {
             match ext.to_str().unwrap() {
                 "h" | "c" | "cpp" | "cc" => {
-                    let res = Command::new("clang-format")
+                    let mut clang_format = ykllvm_bin_dir();
```
ltratt, line 29: It's easier to do `llvm_bin("clang-format")`. [This will also `panic` if it can't find the binary.]

vext01: Ohhhh.

I was confused by:

> then use ykbuild::ykllvm_bin_dir("clang-format") and it'll pick the right one up automatically.

which didn't have the right args!

ltratt: Sorry!

vext01: Try [509747c](https://github.com/ykjit/yk/pull/693/commits/509747cbe7fd82c7093e423e214481775a035117)

ltratt:
Please rebase + squash.

vext01:
Just realised that we now need to update the earlier doc change! One moment!

vext01:
Does 39239e9366c3b8405d6fd2c87a5d73f84dc9e18d work for you?

ltratt:
Please squash and rebase.

vext01:
That was fiddly because the doc file I edited has gone in the meantime.

Arguably the `clangd` section isn't well-placed, but I'd like to move on.

PR #696, opened 2023-05-16
Change clangd database location

ltratt:
This PR (in several phases) moves the location of the clangd databases from `target/compile_commands/<crate-name>/compile_commands.json` to `target/<debug|release>|clangd/<crate-name>/compile_commands.json`. I have done some code tidying up at the same time, including renaming the module, because I once thought `ccgen` meant "compiler generate"!

Note: I don't use the clangd support, so please check things still look OK!


hwtracer/build.rs
```diff
@@ -7,7 +7,7 @@ use std::fs;
 use std::os::unix::fs as unix_fs;
 use std::path::{Path, PathBuf};
 use std::process::Command;
-use ykbuild::{ccgen::CCGenerator, ykllvm_bin};
+use ykbuild::{completion_wrapper::CompletionWrapper, ykllvm_bin};
```
vext01, line 10: `CompletionWrapper` isn't a good name IMHO.

`clangd` does other things than completion, e.g. diagnostics. It's basically rust-analyzer for C/C++.

Check the database specification and see if you can find a name that works for you:
https://clang.llvm.org/docs/JSONCompilationDatabase.html

ltratt: `LSPWrapper`?

vext01: Too generic IMHO.

It was the `CC` part of the name that was confusing right? So why not spell it out: `ClangdCompileCommandsGenerator`. It's long, but precise.

vext01: Or to be in parity with upstream terminology: `ClangdCompilationDBGenerator`?

ltratt: I still don't think we need to tie ourselves to clangd.

vext01: > `CompletionWrapper`

I don't love it, but won't fight.

I actually don't have an issue with `CCGen` especially since the type's docstring says everything you need to know:

> Generate a clangd compilation databases to give LSP support for Yk's C/C++ code.

Boom. Perfect!

Sometimes a type name can't be both short and precise. In such cases I think it's OK to lean on doc strings.

ltratt:
What semantic change?

vext01:
> What semantic change?

Maybe semantic is the wrong word, as there's actually no functional change.

Yes, move the database location. No, don't rename anything. Right?

ltratt:
> Yes, move the database location

I'm confused: that's what the PR does already.

vext01:
This PR does that *and* renames a bunch of stuff.

Unless I'm mistaken, we agreed to not rename anything?

vext01:
please squash

ltratt:
Merge as-is please (I think the individual commits are useful).

PR #697, opened 2023-05-17
Change `build_env` to return a `Vec`.

ltratt:
This allows us to put the necessary environment variables for the compiler wrapper in a single place.

### DIFF
--- a/hwtracer/build.rs
+++ b/hwtracer/build.rs
@@ -3,7 +3,7 @@
 use rerun_except::rerun_except;
 use std::env;
 use std::path::PathBuf;
-use ykbuild::{completion_wrapper::CompletionWrapper, ykllvm_bin};
+use ykbuild::completion_wrapper::CompletionWrapper;
 
 const FEATURE_CHECKS_PATH: &str = "feature_checks";
 
@@ -24,8 +24,9 @@ fn main() {
 
     // Generate a `compile_commands.json` database for clangd.
     let ccg = CompletionWrapper::new("hwtracer", &env::var("CARGO_MANIFEST_DIR").unwrap());
-    env::set_var.call(ccg.build_env());
-    env::set_var("YK_COMPILER_PATH", ykllvm_bin("clang"));
+    for (k, v) in ccg.build_env() {
+        env::set_var(k, v);
+    }
     c_build.compiler(ccg.wrapper_path());
 
     // Check if we should build the perf collector.
--- a/ykbuild/src/completion_wrapper.rs
+++ b/ykbuild/src/completion_wrapper.rs
@@ -1,6 +1,6 @@
 //! Generate compilation command databases for use with clangd.
 
-use crate::target_dir;
+use crate::{target_dir, ykllvm_bin};
 use glob::glob;
 use std::{
     env,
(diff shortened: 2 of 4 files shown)


hwtracer/build.rs
```diff
@@ -100,8 +100,9 @@ fn main() {
 
     // Generate a `compile_commands.json` database for clangd.
     let ccg = CompletionWrapper::new("hwtracer", &env::var("CARGO_MANIFEST_DIR").unwrap());
-    env::set_var.call(ccg.build_env());
-    env::set_var("YK_COMPILER_PATH", ykllvm_bin("clang"));
```
vext01, line 104: Just checking: `YK_COMPILER_PATH` isn't needed anymore? Is that because it's an arg to the wrapper now?

ltratt: It's now defined here https://github.com/ykjit/yk/pull/697/files#diff-974a79e2a0deb1dfad4350c727a981fb2ea2df6c09e874832df6d2048db5206aR79

ykbuild/src/completion_wrapper.rs
```diff
@@ -67,9 +67,19 @@ impl CompletionWrapper {
...
+    /// wrapper compiler's environment.
+    pub fn build_env(&self) -> Vec<(&str, String)> {
+        vec![
+            (
+                "YK_COMPILER_TEMPDIR",
+                self.tmpdir.path().to_str().unwrap().to_owned(),
+            ),
+            (
+                "YK_COMPILER_PATH",
+                ykllvm_bin("clang").as_os_str().to_str().unwrap().to_owned(),
```
vext01, line 80: is it correct to hardcode clang here?

I think it was OK when callsites chose their compiler, but I don't think it's right for this code to make that assumption, as this class is for both C and C++ code.

ltratt: They all chose clang (and we also found that `clang` will compile C++ code correctly, so maybe it doesn't matter?!). We can parameterise it if/when we know we need to?

vext01: > we also found that clang will compile C++ code correctly

That surprises me! And is what confused me.

I think at the very least we should add a comment saying why this works to avoid confusion by future readers.

(Ideally I think I'd prefer the explicit selection of `clang` or `clang++`, as there are advantages to compiling C code with a C++ compiler. Some people do it to get better type handling etc -- let's not do that now)

ltratt: Good idea. Fixed in 4e60a77.

vext01: Not quite what I meant. I was suggesting a comment that explains that `clang` will in fact compile C++ code.

ltratt: > I  was suggesting a comment that explains that clang will in fact compile C++ code

I'm not sure we entirely understand if/what the consequences of doing so might be, so I'd rather not explain things in that way.

vext01: That was the source of my confusion you see.

I think adding:

```
// Note that we hard-code `clang` because it is quite happy to compile both C *and* C++ code!
```

Would help our future selves and others.

ltratt: I don't want to say that, because I don't know exactly how true it is. I'd rather be conservative in what I claim.

vext01: > I don't know exactly how true it is

If we are unsure if clang correctly compiles C++ code, then we should not call clang, surely?

ltratt: I agree: I think that I should do that as the next PR.

vext01: Thanks. Sorry to be awkward.

vext01:
Please rebase and squash.

ltratt:
Squashed.

PR #698, opened 2023-05-18
 The `dir_field` parameter isn't necessary.

ltratt:
This is always going to be `CARGO_MANIFEST_DIR`, so we can reduce this
complexity from the API.

@vext01 We could also (in another PR) get rid of the `db_subdir` if we say it's always `CARGO_CRATE_NAME`. At the moment we wrap the `tests` crate -> `c_tests`: is that useful/necessary? I'm unsure.

### DIFF
--- a/hwtracer/build.rs
+++ b/hwtracer/build.rs
@@ -23,7 +23,7 @@ fn main() {
     let mut c_build = cc::Build::new();
 
     // Generate a `compile_commands.json` database for clangd.
-    let ccg = CompletionWrapper::new("hwtracer", &env::var("CARGO_MANIFEST_DIR").unwrap());
+    let ccg = CompletionWrapper::new("hwtracer");
     for (k, v) in ccg.build_env() {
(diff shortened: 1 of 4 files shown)


ltratt:
Hmm, why is `yklvvm` showing as changed? Investigating.

ltratt:
Hmm, not sure *how* I commited an old ykllvm submodule commit -- I need to keep an eye on that. Fixed now. @vext01 OK to squash that commit away?

vext01:
Tested?

ltratt:
?

vext01:
Have you tested LSP integration still works for the various codebases?

ltratt:
As I've said before: I don't use this.

ltratt:
I believe you can convince yourself easily enough (or not) from the diff.

vext01:
Please squash

ltratt:
Squashed.

PR #699, opened 2023-05-18
Allow users to choose the compiler they want wrapped

ltratt:
This restores some of the subtlety we lost with https://github.com/ykjit/yk/pull/686/files. As that PR showed, only yktracec wanted `clang++` wrapped: hwtracer and langtest_c want `clang` wrapped.

### DIFF
--- a/hwtracer/build.rs
+++ b/hwtracer/build.rs
@@ -3,7 +3,7 @@
 use rerun_except::rerun_except;
 use std::env;
 use std::path::PathBuf;
-use ykbuild::completion_wrapper::CompletionWrapper;
+use ykbuild::{completion_wrapper::CompletionWrapper, ykllvm_bin};
 
 const FEATURE_CHECKS_PATH: &str = "feature_checks";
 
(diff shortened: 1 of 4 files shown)


vext01:
Jolly good!

bors r+

vext01:
> /usr/include/stdc-predef.h:32:100: internal compiler error: Segmentation fault

let's retry.

bors r+

ltratt:
Hmm, that's the second time we've hit that -- but only since https://github.com/ykjit/ykllvm/pull/68 has been merged. It might be coincidence, but perhaps we've introduced something slightly wonky in ykllvm?

ltratt:
AOT. https://github.com/ykjit/yk/pull/699#issuecomment-1553059098 contains the specific message.

ptersilie:
That's building LLVM though. I don't think my recent change has any effect on that.

vext01:
Ah, could this be the age old "gcc sometimes crashes when you build llvm with it" thing @ptersilie?

(Locally I force the compiler to the system clang)

PR #700, opened 2023-05-19
Flatten some of the configuration layers in hwtracer.

ltratt:
Before this commit, there are sort-of 4 layers of configuration necessary to get a tracer: this more-or-less gets things down to 1 layer (with the configuration aspect slightly separate).

In some ways this commit is a bit brutal, and more work is needed, but this seems like a good first step towards simplifying things and making them more idiomatic Rust.

### DIFF
--- a/tests/src/hwtracer_ykpt.rs
+++ b/tests/src/hwtracer_ykpt.rs
@@ -12,17 +12,11 @@
 //! langtester suite) and then they call into this file to have assertions checked in Rust code.
 
 use hwtracer::decode::{TraceDecoderBuilder, TraceDecoderKind};
-use hwtracer::{
-    collect::{TraceCollector, TraceCollectorBuilder, TraceCollectorKind},
-    Trace,
-};
+use hwtracer::{collect::TraceCollector, Trace};
 
 #[no_mangle]
 pub extern "C" fn __hwykpt_start_collector() -> *mut TraceCollector {
-    let tc = TraceCollectorBuilder::new()
-        .kind(TraceCollectorKind::Perf)
-        .build()
-        .unwrap();
+    let tc = TraceCollector::default_for_platform().unwrap();
     tc.start_thread_collector().unwrap();
     Box::into_raw(Box::new(tc))
 }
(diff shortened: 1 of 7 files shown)


vext01:
I'm OK with this.

I've long known this API needed pruning. It's an artifact of a time where hwtracer stood alone.

If we've been too brutal, it will surely present sooner or later and we can change it.

bors r+

ltratt:
bors ping

ltratt:
@vext01 bors didn't kick the job off. Want to try again?

ltratt:
bors r-

Oops, I just realised that I forgot to remove the now-unused imports. 1 moment...

ltratt:
https://github.com/ykjit/yk/pull/700/commits/7c5222d596fa90fee0c44465802b04bd6d11de90 lops another 9 LoC off. OK to squash?

vext01:
Please squash

ltratt:
Squashed.

tests/src/hwtracer_ykpt.rs
```diff
@@ -11,23 +11,21 @@
...
 use hwtracer::decode::{TraceDecoderBuilder, TraceDecoderKind};
-use hwtracer::{
-    collect::{TraceCollector, TraceCollectorBuilder, TraceCollectorKind},
-    Trace,
-};
+#[cfg(feature = "yk_testing")]
+use hwtracer::{collect::TraceCollector, Trace};
 
 #[no_mangle]
+#[cfg(feature = "yk_testing")]
```
vext01, line 20: I suspect this guard is wrong.

ltratt: Agreed. Fixed in fe40adc.

ltratt:
Squashed.

vext01:
Our old friend:

> /usr/include/stdc-predef.h:32:100: internal compiler error: Segmentation fault

I wonder if we should switch the CI to building ykllvm with clang?

ltratt:
I thought there was a problem using clang? Either way, that's not something we should do in this PR.

vext01:
I think the problem is GCC, but agreed.

bors r+

PR #701, opened 2023-05-19
Move `test_helpers` out to the more descriptive module `work_loop`.

ltratt:
I tend to feel that `helper` as a name is too generic to be helpful, so I've made a specific `work_loop` module instead.

By shoving this code into its own module, we also make it less likely that external users who look at the source code will be burdened by `work_loop`.


vext01:
gcc crash again.

retrying.

bors r+

PR #702, opened 2023-05-22
Use clang and clang++ 13 for CI.

vext01:
This solves an issue we are having where GCC crashes when building ykllvm.

This was more of a rabbit hole than expected, as after switching to the Debian clang and clang++, we saw the compilers crash when building yk!

Therefore in this change we are forcing the system compiler to be clang and clang++ 13 (also from Debian packages).

Unfortunately `update-alternatives` only manages `cc` and `c++` (and not `clang` and `clang++`, which some parts of our build explicitly call), so we resort to symlinking `clang-13` and `clang++-13` into `/usr/bin`.

It's not ideal, but since this script is designed to run only inside a CI docker image, I can live with it.

If we get any more trouble from old compiler versions we can get newer packages from the official LLVM apt repo: https://apt.llvm.org/


vext01:
Similar changes will be needed for the following repos:
 - ykllvm
 - yklua
 - ykcbf

PR #703, opened 2023-05-22
Reduce traces by removing controlpoint excess.

ptersilie:
During tracing we often trace over the control point, copying accompanying stores and loads into the trace. These stores and loads serve no purpose since they store values in the heap only to immediately load them back out. These stores and loads make up a large portion of traces, especially where the context around the control point has a large number of live variables (which is often then case in interpreters).

Some example numbers:
| lua test | before | after |
| tpack.lua | 79s | 38s |
| db.lua | 61s | 53s |


yktracec/src/jitmodbuilder.cc
```diff
@@ -1761,6 +1761,61 @@ class JITModBuilder {
...
+            GetElementPtrInst *GEP = cast<GetElementPtrInst>(I);
+            if (GEP->getPointerOperand() == TraceInputs) {
+              // Collect stores
+              std::vector<Value *> Stores;
+              while (isa<GetElementPtrInst>(I)) {
+                assert(cast<GetElementPtrInst>(I)->getPointerOperand() == TraceInputs);
+                std::advance(I, 1);
+                assert(isa<StoreInst>(I));
+                Stores.insert(Stores.begin(), cast<StoreInst>(I)->getValueOperand());
+                std::advance(I, 1);
```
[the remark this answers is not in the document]
ptersilie: Probably. Would you prefer that?

yktracec/src/jitmodbuilder.cc
```diff
@@ -1761,6 +1761,61 @@ class JITModBuilder {
...
+              // We've seen the control point so do the same we used to do when
+              // encountering a control point.
+              VMap[CI] = ConstantPointerNull::get(
+                  Type::getInt8PtrTy(JITMod->getContext()));
+
+              // The next two trace blocks are the (unmappable) control point
+              // followed by the return to this block which only has a branch
+              // instruction left which we don't care about. So we can just skip
+              // over these and continue with the block after.
+              Idx += 2;
```
ltratt, line 1813: Can we add `assert`s just to check that the two instructions we skip are (at least) roughly of the form we expect?

ltratt:
I think we can easily add a test which prints out the new IR and so we can check that it doesn't have the extra gubbins in it?

ptersilie:
> I think we can easily add a test which prints out the new IR and so we can check that it doesn't have the extra gubbins in it?

I think that would require lang tester to be able to count, since the gubbins we are removing only can't exist in the middle of the trace, while being valid at the beginning and end of the trace.

yktracec/src/jitmodbuilder.cc
ptersilie: This instruction skips blocks, not instructions. Maybe @vext01 knows if there's something we can assert here.

vext01: You can assert that they are unmapped, as long as you don't mind taking on the extra computation to do the mapping. Probably ok for debug builds.

vext01: (I have a vague memory that mapping is currently eager anyway)

ltratt:
All we need to do, I believe, is to take a trace from this PR and convert it to something like:

```
...
;      guardfail:...
;        ...
;        %{{0}} = call {{type}} (...) blah blah
;        blah blah
...
```

where we make sure that particular sequence wouldn't have been generated before this PR. Our aim with this sort of testing isn't to have the most precise, perfect, tests imaginable -- but to make sure that we stop code going backwards at some point. Even almost-offensively-simple tests can do that.

yktracec/src/jitmodbuilder.cc
ptersilie: Only one of the blocks is unmapped. The other is mapped (it's the block after returning from the control point).

ltratt: > Probably ok for debug builds.

Sounds OK to me!

vext01: Perhaps the comment is wrong(?), but anyway, you can check that the mapability of the blocks is as expected, right?

ltratt:
> Some example numbers:

Are these for debug or release? [I'm not even sure if my question makes sense! I guess I mean "with assertions in" or "with assertions out"?]

ptersilie:
> > Some example numbers:
> 
> Are these for debug or release? [I'm not even sure if my question makes sense! I guess I mean "with assertions in" or "with assertions out"?]

Uhh, I think yk was debug mode when I got these numbers.

vext01:
Well, assuming you are manually building LLVM and using the `YKB_YKLLVM_BIN_DIR`  trick, you'd have to rebuild LLVM with `-DLLVM_ENABLE_ASSERTIONS=Off` and pass `--release` to `cargo` when you build the rust bit.

ptersilie:
I'm not sure what information this would give us? The bits I improved are in yk (making trace compiling faster) not lua itself. So if anything then I should be running yk in release mode to see how much difference my change made? Right? What am I missing?

ptersilie:
- Added a test to check for regressions.
- Updated a comment with something I only now realised (this change also removes unneeded guard failure blocks).

vext01:
> The bits I improved are in yk (making trace compiling faster) not lua itself

Ohhhh.

I thought that was runtime measurements :P

Maybe Laurie too?

ltratt:
I assumed this was measuring the time of yklua running a test -- but I also assume that it's measuring asserts (and whatnot) inside yk. I might be wrong!

yktracec/src/jitmodbuilder.cc
```diff
@@ -1806,10 +1806,12 @@ class JITModBuilder {
               VMap[CI] = ConstantPointerNull::get(
                   Type::getInt8PtrTy(JITMod->getContext()));
 
-              // The next two trace blocks are the (unmappable) control point
-              // followed by the return to this block which only has a branch
-              // instruction left which we don't care about. So we can just skip
-              // over these and continue with the block after.
+              // The next two trace blocks may be skpped as they only contain
```
ltratt, line 1809: s/skpped/skipped/

yktracec/src/jitmodbuilder.cc
```diff
@@ -1806,10 +1806,12 @@ class JITModBuilder {
...
+              // the (unmappable) control point followed by the return to the
+              // current block which only has a branch instruction left which we
+              // don't care about. Skipping the latter also means we don't
+              // create additional guard-failure blocks for branches that can
+              // never fail, reducing the trace even more.
```
ltratt, line 1814: Presumably the only branch that can never fail is the very final branch in a trace?

ltratt:
We don't need to do it for this PR, but going forward we should do the benchmarking with asserts off in yk and LLVM -- after all, we don't know in general if all we've actually improved is the number of `assert`s we're running! [In this PR's case, I think it's rather obvious that's not what's happening, which is why I'm not too concerned. But as performance differences become smaller, removing asserts will become more important.]

yktracec/src/jitmodbuilder.cc
ltratt: I see!

yktracec/src/jitmodbuilder.cc
ltratt: @ptersilie I think just this needs fixing?

ptersilie: [061d09c](https://github.com/ykjit/yk/pull/703/commits/061d09c6490da35edce50ae22725020fb094a9a5)

yktracec/src/jitmodbuilder.cc
ptersilie: [724e819](https://github.com/ykjit/yk/pull/703/commits/724e819718e73e4786be3b01e55ad785a4aed4d9)

yktracec/src/jitmodbuilder.cc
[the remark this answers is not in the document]
ptersilie: [de1a728](https://github.com/ykjit/yk/pull/703/commits/de1a7282012f672479b19aab2d529f62d8e8334e)

ptersilie:
Ready fir rereview.

ptersilie:
Squashed.

ptersilie:
Forgot about release tests. The IR is different in release mode. Can I write different tests for debug/release?

ltratt:
Please squash.

PR #704, opened 2023-05-24
Don't run the benchmarks for as long as Criterion wants.

ltratt:
In CI, we want the benchmarks to be built, and we want to check that they can run, but the results are meaningless. By setting `--profile-time 1` we lower the time for `cargo bench` from just over 4 minutes to just under 10 seconds. In other words, this commit reduces overall CI build time by a little under 10%!


vext01:
Good work!

bors r+

PR #705, opened 2023-05-25
Start simplifying the tracing API and making it more flexible

ltratt:
This PR aims to start simplifying the internal tracing APIs and make them more flexible. It is best viewed as a "useful checkpoint" rather than "this ties up every loose end".

Before this PR, we couldn't practically expect to dynamically change the tracer: they were effectively hard-coded at compile time, despite having various traits that made it look like they could be changed. This PR ultimately (I think/hope!) makes dynamic switching possible, although it doesn't expose a knob to the user to allow them to actually dynamically switch a tracer.

The three commits each compile and pass tests, though you can see from one to another an evolution in what I think possible and desirable. I think they are still usefully kept apart.

In terms of simplification, this PR is not the end. It is not even the beginning of the end. But it is, perhaps, the end of the beginning.

### DIFF
--- a/hwtracer/src/collect/mod.rs
+++ b/hwtracer/src/collect/mod.rs
@@ -3,7 +3,7 @@
 use crate::{errors::HWTracerError, Trace};
 use core::arch::x86_64::__cpuid_count;
 use libc::{size_t, sysconf, _SC_PAGESIZE};
-use std::{cell::RefCell, convert::TryFrom, sync::LazyLock};
+use std::{any::Any, convert::TryFrom, sync::LazyLock};
 
 #[cfg(collector_perf)]
 pub(crate) mod perf;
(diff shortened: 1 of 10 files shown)


hwtracer/src/collect/mod.rs
```diff
@@ -25,75 +25,54 @@ const PERF_DFLT_INITIAL_TRACE_BUFSIZE: size_t = 1024 * 1024; // 1MiB
 thread_local! {
     /// When `Some` holds the `ThreadTraceCollector` that is collecting a trace of the current
     /// thread.
-    static THREAD_TRACE_COLLECTOR: RefCell<Option<Box<dyn ThreadTraceCollector>>> = RefCell::new(None);
+    static THREAD_TRACE_COLLECTOR: RefCell<Option<Box<dyn ThreadTracer>>> = RefCell::new(None);
```
vext01: I prefer the specific terminology "collect" and "decode" rather than the more generic "tracer" which is kind of overloaded. Don't you agree?

ltratt: This bit ends up being changed in later commits. That said, even after this PR, the terminology we use is a bit mixed and, sometimes, confused. I'd like to address that in a future PR.

vext01: OK. I'll see how it plays out, but I avoid "tracer" because it's so ambiguous.

The system as a whole, we often call a tracer, yet here we are talking only about the bit that collects a trace.

ltratt: Agreed: this stuff needs a rethink across the whole repo. I don't have a very firm opinion on the right thing to do other than "what we do currently can surely be improved".

hwtracer/src/collect/mod.rs
```diff
@@ -3,7 +3,7 @@
 use crate::{errors::HWTracerError, Trace};
```
vext01, line 3: The commit message says something about "yktcare". Is that a typo?

ltratt: Oops, should be `yktrace`!

hwtracer/src/collect/mod.rs
```diff
@@ -22,43 +22,16 @@ static PERF_DFLT_AUX_BUFSIZE: LazyLock<size_t> = LazyLock::new(|| {
 
 const PERF_DFLT_INITIAL_TRACE_BUFSIZE: size_t = 1024 * 1024; // 1MiB
 
-thread_local! {
-    /// When `Some` holds the `ThreadTraceCollector` that is collecting a trace of the current
-    /// thread.
-    static THREAD_TRACE_COLLECTOR: RefCell<Option<Box<dyn ThreadTracer>>> = RefCell::new(None);
```
vext01: Aren't we going to end up bringing this back later to get back the "only one tracer per thread" property?

hwtracer/src/collect/mod.rs
vext01: OK, make sure we fix that at squash time.

hwtracer/src/collect/mod.rs
vext01: Surely undoing the rename here is what we need?

ltratt: In this case, this thread_local disappears later: but it's not just the things directly touched by this PR that have slightly confusing names. We really need to do a thorough consistency pass.

vext01:
Due to the scope of the changes, it's quite hard to take in, so I'm going to have to trust you to some degree here.

Let's push onward and see if anything comes up.

Do you want to force push that type out now?

vext01:
Yes, it's a bit messy as our understanding and requirements changed over time, and we've been doing the minimal changes to keep it somewhat working.

To be clear, I'm all in favour of simplifying it.

vext01:
> Do you want to force push that type out now?

^ Just in case this was missed. Go ahead and force push.

ltratt:
I'm going to wait until you and @ptersilie have had a look at the yklua fallout before force pushing (if I force push it'll mess your copies of this branch up).

hwtracer/src/collect/mod.rs
vext01: I don't see it.

If you search for `thread_local!` in the diff for this whole PR, you can see that it only appears in deleted lines.

@ptersilie and I have been debugging the branch and the problem appears to be a thread starts tracing when it is already tracing. This manifests as a "trace" which starts fine, but is later interrupted by the start of another trace. The trace compiler is then confused by the impossible control flow.

You need the thread local (or equivalent) to track whether the current thread is already tracing.

ltratt:
Force pushed the commit message update.

ltratt:
Might as well try a full build while I'm here.

bors try

vext01:
Do you want to merge this?

ltratt:
I think this is OK for merging -- our working assumption is that the PR isn't responsible for the problems we saw (and that are now a test in #709). If we do later find that there's a fundamental flaw in this PR, we can fix it in situ.

Issue #708, opened 2023-05-26
Unmappable code with unusual control flow triggers an assertion


ltratt:
As a result of #705, @vext01 and I *think* we might have found a case where unmappable code does things with the stack pointer that we don't handle properly, ultimately leading to:

```
src/jitmodbuilder.cc:1540: Module *JITModBuilder::createModule(): Assertion `!CallStack.curMappableFrame()' failed.
```

A simple example that triggers this is:

```c
int main(int argc, char **argv) {
  int i = 10;
  YkMT *mt;
  YkLocation loc = yk_location_new();
  while (i > 0) {
    mt = yk_mt_new(NULL);
    yk_mt_hot_threshold_set(mt, 0);
    yk_mt_control_point(mt, &loc);
    i--;
  }
  yk_location_drop(loc);
  yk_mt_drop(mt);
  return (EXIT_SUCCESS);
}
```

We think that `yk_mt_new` is calling some unmappable code which does something like `setjmp`/`longjmp`, although we haven't investigated the exact cause yet.

PR #709, opened 2023-05-26
Add a simple test which triggers the possibly-unmappable problem.

ltratt:
Gives us a concrete test for #708. `ignore`d at the moment since it (by definition) fails.

### DIFF
--- a/tests/c/awkward_unmappable.c
+++ b/tests/c/awkward_unmappable.c
@@ -1,0 +1,24 @@
+// ignore: Fails with: Assertion `!CallStack.curMappableFrame()' failed
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   status: success
+
+#include <assert.h>
+#include <stdlib.h>
+#include <yk.h>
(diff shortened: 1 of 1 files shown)


tests/c/awkward_unmappable.c
```diff
@@ -0,0 +1,26 @@
+// ignore: Fails with: Assertion `!CallStack.curMappableFrame()' failed
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   status: success
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
```
vext01, line 9: I think the headers could be minimised. string and stdio don't appear to be used?

vext01:
Just one comment from me.

tests/c/awkward_unmappable.c
ltratt: Good point. Fixed in 0143a1a.

vext01:
Please squash

ltratt:
Squashed.

PR #710, opened 2023-05-27
Push trace decoding selection to the tracer.

ltratt:
Since we may have different types of tracer (various types of hardware tracers; software tracers; and so on), `MT` can't sensibly offer the ability to pick different trace decoders -- different backends will have very different ideas of what decoders they wish to offer.

This commit thus removes the ability to specify a trace decoder within `MT`, but gives trace decoders a reference to a `Tracer`, so that users can configure a `Tracer` object and have that later control trace decoding (if they want to). At the moment, we don't make use of this facility, because we only have one decoder.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -1,6 +1,5 @@
 //! The main end-user interface to the meta-tracing system.
 
-use hwtracer::decode::TraceDecoderKind;
 #[cfg(feature = "yk_testing")]
 use std::env;
 use std::{
@@ -62,7 +61,6 @@ pub struct MT {
     /// How many worker threads are currently running. Note that this may temporarily be `>`
     /// [`max_worker_threads`].
     active_worker_threads: AtomicUsize,
-    trace_decoder_kind: TraceDecoderKind,
     tracer: Arc<dyn Tracer>,
 }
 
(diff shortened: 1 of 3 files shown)


vext01:
At the risk of looking silly:

> MT can't sensibly offer the ability to pick different trace decoders

Why not? Can you give an example where it couldn't be done?

I think my plan was to allow users to choose whatever they want, but tell them if things aren't compatible.

ltratt:
> At the risk of looking silly:
>>   MT can't sensibly offer the ability to pick different trace decoders
>Why not? Can you give an example where it couldn't be done?

Different tracing backends have different trace decoder possibilities. By having the decoder choice at the top-level, we allowed users to chose nonsensical combinations such as a libpt decoder with a software tracer.

vext01:
Right, but I'm arguing that one sensible solution would be to reject invalid combos.

Another thing that pops to mind: supposing we get a second PT decoder later, shouldn't the user have the ability to choose which one the use?

I'm OK with moving this forward for, but I disagree that we can't sensibly offer these options, and we may need to bring back some kind of configuration API later (maybe under a different guise).

Do you agree?

ltratt:
> I'm arguing that one sensible solution would be to reject invalid combos.

Why reject them dynamically when you can prevent them being expressed statically?

> supposing we get a second PT decoder later, shouldn't the user have the ability to choose which one the use?

This PR explicitly makes that sort of thing possible, but the user will have to do configure that choice with a `Tracer` object, not directly on the `MT` instance.

PR #711, opened 2023-05-30
Move yk-internal parts from `ykcapi` to `ykrt`

ltratt:
`ykcapi` is a good name for what we want that crate to do, but the actual contents of `ykcapi` have become "a C api and random internal yk code that needs to be part of a cdylib".

This PR moves nearly all of that internal yk code from `ykcapi` to `ykrt`. The only thing that's left is `__ykrt_control_point` -- and, arguably, there isn't a much better place for that than `ykcapi` (or, at least, I can't think of a much better place for it).


vext01:
Earlier on we had the rule "if possible, put all C functions in ykcapi, to avoid linkage issues".

I no longer remember what those issues were...

ykrt/src/deopt.rs
```diff
@@ -0,0 +1,59 @@
+//! Run-time deoptimisation support: when a guard fails, this module restores the state necessary
+//! to resume interpreter execution.
+
+use std::{arch::asm, ffi::c_void};
+
+/// The `__llvm__deoptimize()` function required by `llvm.experimental.deoptimize` intrinsic, that
+/// we use for exiting to the stop-gap interpreter on guard failure.
```
vext01, line 87: mention of stop-gap interpreter here.

vext01:
If it links, I'm OK with it.

How do we feel about the `__<crate>_<thing>` convention used? I often question it. I guess it doesn't hugely matter.

ltratt:
> Earlier on we had the rule "if possible, put all C functions in ykcapi, to avoid linkage issues".

I did encounter some linking issues when doing this PR, not all of which I understood. I *think* in retrospect, they were because I was splitting some things across two different crates when they needed to be in one crate.

ykrt/src/deopt.rs
ltratt: There are quite a lot of "stop-gap"s in the repo, including in test file names. If we like this PR, I will do another PR which goes through and fixes all those (I worried that doing so here would make the PR hard to read).

vext01: sounds good!

vext01:
Okie dokie!

bors r+

ltratt:
bors r-

I would like to check @ptersilie is OK with this too before merging.

vext01:
My apologies

ptersilie:
Looks good to me.

PR #716, opened 2023-05-31
Remove all remnants of the stopgap interpreter.

ptersilie:
The long removed stopgap interpreter was still being mentioned throughout the source code and some tests. This PR removes all mentions and some tests that were specific to the stopgap interpreter and aren't very useful anymore.


tests/c/return_char.c
```diff
@@ -1,50 +0,0 @@
-// Run-time:
```
vext01, line 1: Is it worth keeping some of these tests around anyway, even if they don't test an SGI?

vext01:
LGTM with one minor query.

tests/c/return_char.c
ltratt: In general I think "more tests" is better than "fewer tests" with one proviso: there's not much use having a load of tests that are really the same test in disguise (i.e. they all test the same thing) because it gives a misleading impression of coverage. I don't know if that's the case here: I'll be happy with whatever @ptersilie decides.

ptersilie: I've removed these because they don't really test anything that other tests don't already cover. Just look at the loop the interpreter is in. All it does is increment `i`. There's plenty of tests that already do that.

ptersilie:
bors try

ltratt:
Why does reinstating that file make a difference?

vext01:
When a test file has a `-O0` suffix, the expected outcome is read from that file instead of the normal place.

ltratt:
Oh, so it's the base case in a sense.

@ptersilie Please squash.

vext01:
I'm surprised we have any tests that are > -O0.

ptersilie:
I'm not entirely sure. The `.O0` extension means that this file is run (instead of the normal one) when we compile (the trace?) with `-O0`. But I actually don't see the difference between the files.

Now that I've looked closer at the previous error, I think it was langtester not being happy with the top level comment (expecting some langtester key there). So I now think that reinstating that file means that the normal version of that test is never ran. So maybe the right course of action is to remove the normal version and replace it with the `O0` version.

ptersilie:
> I'm surprised we have any tests that are > -O0.

Yeah, I'm now thinking the not `O0` version of the file should be deleted and the `O0` made the default. Since we don't have > `O0` the normal version was never ran, until I deleted the `O0` version.

ltratt:
> Yeah, I'm now thinking the not O0 version of the file should be deleted and the O0 made the default.

I'm fine with that. Let's do it in a PR immediately after this one?

ptersilie:
bors try

ltratt:
bors try-

ltratt:
@ptersilie This is OK for merging, right?

ltratt:
Please squash.

ltratt:
Hang on.

tests/c/internal_linkage_same_obj.c
```diff
@@ -16,7 +24,7 @@ static int call_me(int x) {
 }
 
 int main(int argc, char **argv) {
-  YkMT *mt = yk_mt_new(NULL);
+  YkMT *mt = yk_mt_new();
```
ltratt, line 27: This is wrong: `yk_mt_new` should take `NULL` here.

ltratt: This would have failed a compile-time check, so it didn't work.

tests/c/internal_linkage_same_obj.c
```diff
@@ -1,3 +1,11 @@
+// ignore: Requires stopgap support for nested functions.
```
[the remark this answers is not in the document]
ptersilie: Oh! Wait. I bollocked this up. Fixing.

ptersilie:
Okay, that should fix things.

ltratt:
That looks a bit more like it ;)

Please squash.

tests/c/internal_linkage_same_obj.c
ptersilie: It never ran cause it had the ignore tag. My bad :facepalm:

ltratt:
@ptersilie Have you squashed?

PR #717, opened 2023-05-31
Simplify `Tracer` & `ThreadTracer` in hwtracer.

ltratt:
This more-or-less makes these two traits similar to their namesakes in `yktrace`. Previously a `ThreadTracer` was passed to a `Tracer`, but that forces complexity onto the user of the API. By judicious use of `Arc` (which can be seen as very similar to `Box` in this setting), we simplify the API. Notably this lets us get rid of the horrible use of `Any`.

[Since I accidentally created two similar but distinct APIs last week, I've now had the opportunity to compare the two over time. `yktrace`'s API is clearly less worse than `hwtracer`'s, so it makes sense to make the latter look more like the former.]

### DIFF
--- a/hwtracer/src/collect/mod.rs
+++ b/hwtracer/src/collect/mod.rs
@@ -3,7 +3,10 @@
 use crate::{errors::HWTracerError, Trace};
 use core::arch::x86_64::__cpuid_count;
 use libc::{size_t, sysconf, _SC_PAGESIZE};
-use std::{any::Any, convert::TryFrom, sync::LazyLock};
+use std::{
+    convert::TryFrom,
+    sync::{Arc, LazyLock},
+};
 
 #[cfg(collector_perf)]
 pub(crate) mod perf;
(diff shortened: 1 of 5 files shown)


hwtracer/src/collect/mod.rs
```diff
@@ -22,21 +25,22 @@ static PERF_DFLT_AUX_BUFSIZE: LazyLock<size_t> = LazyLock::new(|| {
...
-    fn start_collector(&self) -> Result<Box<dyn ThreadTracer>, HWTracerError>;
-
-    /// Stop collecting a trace of the current thread.
-    fn stop_collector(&self, tt: Box<dyn ThreadTracer>) -> Result<Box<dyn Trace>, HWTracerError>;
+    fn start_collector(self: Arc<Self>) -> Result<Box<dyn ThreadTracer>, HWTracerError>;
 }
 
+/// Represents a thread which is currently tracing.
 pub trait ThreadTracer {
-    fn as_any(self: Box<Self>) -> Box<dyn Any>;
```
vext01, line 34: I don't even remember writing this...

ltratt: That's because you didn't.

vext01: That explains it.

vext01:
The only thing that sticks out here is the ambiguous (and unrelated to this PR) type name `Tracer`.

It's a trace collector, right?

ltratt:
The terminology is a mess: this PR doesn't make it worse (but it doesn't improve it either).

PR #719, opened 2023-05-31
Remove `bfbug.c`.

ltratt:
This test has never been enabled in our test suite, though the reason it's been `ignored` has changed over time. In a5d36f6f:

  // ignore: Consistently hits `Assertion `!CallStack.curMappableFrame()' failed`

In b2caa8c:

  // ignore: runs too fast to check what should be checked. reenable when the PT decoder is faster

I think that this test dates from the time when guard blocks could be merged together accidentally. We noticed that in ykcbf, from where this test derives. However, b2caa8c added more focussed tests that also check the problem doesn't reemerge.

The problem with `bfbug.c` is that it is *very* slow -- it takes over 10 minutes on bencher13 -- because it is from the slowest BF interpreter we wrote. So none of us are going to ever want to run this tests, because it takes several times longer than the rest of the test suite put together.

This commit thus deletes the test because a) it's too slow b) what it checks is also checked elsewhere.


ptersilie:
Sounds reasonable.

bors r+

PR #721, opened 2023-06-01
Remove unused method from trait.

ltratt:


vext01:
Ah.

Left over from the PT cull.

bors r+

PR #722, opened 2023-06-01
Fix clangd paths.

ltratt:
Note that these are all hard-coded to debug builds: yk does generate different completion sets for debug/release.


ltratt:
That directory no longer exists (if you have it, it means you have a partly out-of-date `target` directory): by definition there is no way of doing this that doesn't involve "picking" debug or release mode.

ltratt:
I'm closing this PR as I pushed to the wrong repo. Mea culpa!

PR #723, opened 2023-06-01
Fix clangd paths.

ltratt:
Note that these are all hard-coded to debug builds: yk does generate different completion sets for debug/release.

### DIFF
--- a/yktracec/.clangd
+++ b/yktracec/.clangd
@@ -1,2 +1,2 @@
 CompileFlags:
-    CompilationDatabase: ../target/debug/clangd/ykllvmwrap/
+    CompilationDatabase: ../target/debug/clangd/yktracec/
--- a/yktracec/.clangd
+++ b/yktracec/.clangd
@@ -1,2 +1,2 @@
 CompileFlags:
-    CompilationDatabase: ../target/compile_commands/ykllvmwrap/
+    CompilationDatabase: ../target/debug/clangd/ykllvmwrap/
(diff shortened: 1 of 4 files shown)


yktracec/.clangd
```diff
@@ -1,2 +1,2 @@
 CompileFlags:
-    CompilationDatabase: ../target/compile_commands/ykllvmwrap/
+    CompilationDatabase: ../target/debug/clangd/ykllvmwrap/
```
vext01, line 2: While we are here, may aswell update this last remaining `ykllvmwrap`.

It should be `yktracec`, but there is another place somewhere in the build system still referencing `ykllvmwrap` as the LSP is actually working for `yktracec` :P

vext01:
The LSP appears to be working for `yktracec` and `hwtracer` but not for the C tests.

```
$ ls target/debug/clangd
hwtracer/  ykllvmwrap/
```

Looks like the dir has gone MIA?

yktracec/.clangd
ltratt: Ooo, good spot! Fixed in 817d9e2.

ltratt:
You probably need to do `cargo test` before seeing `c_tests` in there (I can see it in my `target` directory, but I don't entirely remember what created it).

vext01:
I thought that too, but alas, running `cargo test` didn't make it appear.

vext01:
Can you try deleting your target dir and doing `cargo test`. Do you see the `clangd/c_tests` dir after?

ltratt:
> Can you try deleting your target dir and doing cargo test. Do you see the clangd/c_tests dir after?

Already done/doing just that! Will report back.

ltratt:
```
$ rm -rf target
$ cargo test
...
$ ls target/debug/clangd
c_tests  hwtracer  ykllvmwrap
```

So It Works For Me (TM).

ltratt:
OK to squash?

ltratt:
Ah, no, I'm wrong. This is a merge-as-is please.

PR #725, opened 2023-06-03
Report errors in `xtask cfmt` more clearly.

ltratt:
Clippy complained about an `expect` in this file: when I looked, I realised that when `clang-format` fails, we could report errors in two subtly different ways, neither of which told us what happened. This commit homogenises, and gives the user much more information about went wrong. This is probably overkill, but Clippy doesn't complain now!

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -27,15 +27,31 @@ fn clang_format() {
             match ext.to_str().unwrap() {
                 "h" | "c" | "cpp" | "cc" => {
                     let clang_format = ykllvm_bin("clang-format");
-                    let res = Command::new(&clang_format)
+                    let mut err_msg = None;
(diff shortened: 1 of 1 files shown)


xtask/src/main.rs
```diff
@@ -27,15 +27,31 @@ fn clang_format() {
...
+                                }
+                            }
+                        }
+                        Err(e) => err_msg = Some(e.to_string()),
+                    }
+                    if let Some(m) = err_msg {
                         panic!(
-                            "{:?} failed on {}",
-                            clang_format,
+                            "{} -i {}: {m}",
```
vext01, line 53: What's this `-i` here? Are we showing the arguments clang-format was run with? Can we print the `Command` so that this can't get out of sync?

PR #731, opened 2023-06-05
Merge `yktrace` into `ykrt`.

ltratt:
The contents of `yktrace` are inherently run-time only, so there's no good reason for this to be a crate of its own. As discussed at the in-person a couple of weeks back, merging this into `ykrt` helps clean up top-level crate proliferation.


vext01:
Yes! Agreed.

bors r+

PR #733, opened 2023-06-06
Provide `LocationInner::from_hotlocation`.

ltratt:
Previously we created a counting `LocationInner` then immediately converted it a non-counting state. This commit allows us to immediately create a non-locked non-counting state, avoiding that pointless intermediate stage.


vext01:
Seems fine to me.

bors r+

PR #735, opened 2023-06-09
Significantly simplify `Location`s and `Location` transitions.

ltratt:
Previously we did a lot of very clever, and very `unsafe`, stuff with `Location`s. Most obviously we used atomics extensively: we also relied on pointer-related stuff. While wanting to extend this code I found a case where we compared two pointers in a way that in C could be undefined behaviour, due to differing provenances -- I suspect the same is true of Rust, but pointer provenance rules are less developed in Rust than C. This has made me realise two things: firstly, we were being too clever for our own good; secondly, we're relying on parts of Rust whose semantics are unclear.

This commit thus radically simplifies all this code. It preserves the property we care about the most: `Location`s don't allocate memory unless they become hot. After that, most of the low-level details change. A `Location` is now either a count (i.e. a "normal"(ish) integer) or a pointer to an `Arc<Mutex<HotLocation>>`. We use `Arc::into_raw` and `Arc::from_raw`, relying on the neat property that even when the `Location` is dropped, the `Arc` will only be dropped if/when its strong count goes to 0 (which, if a thread is using the `Location` elsewhere, might not happen until later, if at all). Ultimately, this allows to rely on 4 simple `unsafe`s in `location.rs`: relative to before, it's easier to examine these and have confidence as to their correctness. Note: I'm not promising I've got everything right. But a) I hope there's less obviously wrong b) if there is something wrong it should be easier to spot than before!

I have deliberately chosen to keep the structure of the code, and the nature of the `Location` state machine as close as possible to what we had before. In particular, I have been able to maintain (for better or worse) the same performance properties: our benchmarks run in virtually the same time as before (one even runs slightly faster, but I think that's a happy accident, rather than a virtue of this commit).

### DIFF
--- a/ykrt/src/location.rs
+++ b/ykrt/src/location.rs
@@ -2,8 +2,9 @@
 
 use std::{
     convert::TryFrom,
+    mem,
     sync::{
-        atomic::{AtomicPtr, AtomicUsize, Ordering},
+        atomic::{AtomicUsize, Ordering},
         Arc,
     },
 };
@@ -13,10 +14,17 @@ use crate::{
     trace::CompiledTrace,
 };
 use parking_lot::Mutex;
-use parking_lot_core::{
-    park, unpark_one, ParkResult, SpinWait, UnparkResult, UnparkToken, DEFAULT_PARK_TOKEN,
-};
-use strum::EnumDiscriminants;
+
+#[cfg(target_pointer_width = "64")]
+const STATE_TAG: usize = 0b1; // All of the tag data must fit in this.
+#[cfg(target_pointer_width = "64")]
+const STATE_NUM_BITS: usize = 1;
+
+/// Because hot locations will be most common, we save ourselves the effort of ANDing bits away by
+/// having `STATE_HOT` be 0, expecting that `ptr & !0` will be optimised to just `ptr`.
+const STATE_HOT: usize = 0;
+/// In the not hot state, we have to do `(inner & !1) >> STATE_NUM_BITS` to derive the count.
+const STATE_NOT_HOT: usize = 0b1;
 
 /// A `Location` stores state that the meta-tracer needs to identify hot loops and run associated
 /// machine code.
@@ -32,8 +40,7 @@ use strum::EnumDiscriminants;
 #[repr(C)]
 #[derive(Debug)]
 pub struct Location {
-    // A Location is a state machine which operates as follows (where Counting is the start
-    // state):
+    // A Location is a state machine which operates as follows (where Counting is the start state):
     //
     //              ┌──────────────┐
     //              │              │─────────────┐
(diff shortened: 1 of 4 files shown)


ltratt:
bors try

ykrt/src/location.rs
```diff
@@ -62,39 +69,22 @@ pub struct Location {
...
-    //   Box<HotLocation> | 0           | 1         | 0         | Not locked, thread(s) waiting
-    //   Box<HotLocation> | 0           | 1         | 1         | Locked, thread(s) waiting
-    //
-    // where `count` is an integer and `Box<HotLocation>` is a pointer to a `malloc`ed block of
-    // memory in the heap containing a `HotLocation` enum.
-    //
-    // The precise semantics of locking and, in particular, parking are subtle: interested readers
-    // are directed to https://github.com/Amanieu/parking_lot/blob/master/src/raw_mutex.rs#L33 for
-    // a more precise definition.
+    // In the `STATE_NOT_HOT` state, the payload is an integer; in a `STATE_HOT` state, the payload
```
ptersilie, line 86: What's the content of the payload in `STATE_NOT_HOT`? I assumed from the description above, that a `Location` has two states: a counter or a pointer. But it seems there's a third that I don't see.

ykrt/src/location.rs
```diff
@@ -103,340 +93,118 @@ impl Location {
     pub fn new() -> Self {
         // Locations start in the counting state with a count of 0.
         Self {
-            inner: AtomicUsize::new(LocationInner::new().x),
```
ptersilie, line 106: So much red, so much joy! :-)

ykrt/src/location.rs
ptersilie: Looking at this more I might just be misinterpreting this table:
```
    //   bit(s) | 63..1   | 0
    //          | payload | <STATE_NOT_HOT|STATE_HOT>
```

What does the first row tell me?

ykrt/src/location.rs
```diff
@@ -103,340 +93,118 @@ impl Location {
...
-            x: self.x | STATE_IS_LOCKED,
+    /// If `self` has a [HotLocation] return a reference to the `Mutex` that directly wraps it, or
+    /// `None` otherwise.
+    pub(crate) fn hot_location(&self) -> Option<&Mutex<HotLocation>> {
+        let x = self.inner.load(Ordering::Relaxed);
+        if x & STATE_NOT_HOT == 0 {
+            // `Arc::into_raw::<Mutex<T>>` returns `*mut Mutex<T>` so the address we're wrapping is
+            // a pointer to the `Mutex` itself. By returning a `&` reference we ensure that the
+            // reference to the `Mutex` can't outlive this `Location`.
+            // caller can
```
ptersilie, line 169: random comment.

ykrt/src/mt.rs
```diff
@@ -912,14 +847,13 @@ mod tests {
         let loc1 = Location::new();
         let loc2 = Location::new();
 
-        // Get `loc1` to the point where there's a compiled trace for it.
+        // get `loc1` to the point where there's a compiled trace for it.
```
ptersilie, line 850: Why the lower case `get`?

ptersilie:
I won't pretend I fully understand every detail of this change, but I get the gist I think. `Location`s feel much easier to understand now due to the reduced state they can have (they amount of red in the diff is witness to that). And I'm glad to see that this simplified `mt.rs` as well. Just a few comments.

ykrt/src/location.rs
ltratt: Another way of reading this is "bit 0 == <STATE_NOT_HOT|STATE_HOT>, bits 1..63 = <payload>". Perhaps the table is not as useful as when there were multiple possibilities here. I can certainly get rid of it.

ykrt/src/location.rs
ltratt: Fixed in 3c78fc3.

ltratt:
@ptersilie Ready for rereview.

ykrt/src/location.rs
ptersilie: Oh! I see now.  I thought the two rows were different options. Maybe we can change this to
```
    //   bit(s) | 63..1   | 0
    // -------- | ------- | ---
    //          | payload | <STATE_NOT_HOT|STATE_HOT>
```
to make clearer the first column is the title header?

ltratt: I just inlined it into prose in https://github.com/ykjit/yk/commit/7c485783850902032c0532e5d30e11fd76b162e0 if that works for you?

ptersilie: Yeah, that works!

ptersilie:
Please squash.

PR #736, opened 2023-06-09
Implement trace looping.

ptersilie:
Trace looping PR

This PR does several things. First in bab39a1e787cc73f484e9396ce4ddccbb4604212 we simplify traces by removing an unnecessary guard failure around the control point, while we are here, we insert an additional branch `loopentry`, which we'll use soon.

Where before we had:

```llvm
%5 = getelementptr %YkCtrlPointVars, ptr %0, i32 0, i32 0
%6 = load ptr, ptr %5, align 8
%7 = icmp eq null, null
br i1 %7, label %guardsuccess, label %guardfail

guardfail:
...

guardsuccess:
store ptr %6, ptr @shadowstack_0, align 8
...
```

We now get:

```llvm
entry:
%5 = getelementptr %YkCtrlPointVars, ptr %0, i32 0, i32 0
%6 = load ptr, ptr %5, align 8
br label %2

loopentry:
store ptr %6, ptr @shadowstack_0, align 8
...
```

In 36a65c103cb4051b5d5dedcd23ef2f0f96440124 we implement trace looping, which is as simple as replacing the return at the end of the trace with a `br` instruction to the `loopentry` block. We don't need to worry about insert PHI nodes since all the important variables are on the stack/shadowstack.

First performance tests have resulted in a slight (~15%) slowdown, due to traces now taking longer to compile. Presumably, this is because having a loop allows LLVM to apply more optimisations.


vext01:
If you discount the trace compilation time, is there a speedup?

ptersilie:
> If you discount the trace compilation time, is there a speedup?

No, it looked pretty similar, but I didn't really look further into it.

tests/c/debug_simple_interp_loop2.c
```diff
@@ -42,7 +42,7 @@
 //       %{{fptr2}} = getelementptr %YkCtrlPointVars, ptr %0, i32 0, i32 0...
 //       store...
 //       ...
-//       ret ptr null
+//       br...
```
vext01, line 45: why don't we match the label name?

vext01:
LGTM. I expected it to be harder!

vext01:
So it's just "guard failure" -> "guard" where I pointed it out earlier, and a query about the tests.

tests/c/debug_simple_interp_loop2.c
ptersilie: We might change the label at some point? There is a test that checks this while being agnostic about the chosen block name: [tests/c/simple.c](https://github.com/ykjit/yk/pull/736/commits/36a65c103cb4051b5d5dedcd23ef2f0f96440124#diff-30045133b33c3db0b2bd22080a981b5cd9d1c5d79f8b1b3cf50c2bc2cbdec05f)

ptersilie:
> Do we have any tests which have post opt IR?

Not that I am aware of? @vext01?

ltratt:
>>    Do we have any tests which have post opt IR?
>Not that I am aware of! @vext01?

I think it would be good to have at least 1, maybe even 2 or 3. Preferably starting with something ludicrously small where pre-opt probably can't optimise it. Such an example is a bit harder to create than it first appears, though.

vext01:
5 tests use `jit-post-opt`, and I've added more in my promote branch (pre-opt has too much guff around which makes it hard to test).

ltratt:
> 5 tests use jit-post-opt, and I've added more in my promote branch (pre-opt has too much guff around which makes it hard to test).

OK, maybe we should wait for the promote PR then.

ptersilie:
I'm happy to wait, but at the same time, is any of this going to affect the changes here? We need trace looping one way or another, don't we?

tests/c/debug_simple_interp_loop2.c
ltratt: Just this one and then I think we'll be ready to go.

vext01:
Last commit LGTM. Now just "guard failure" -> "guard".

tests/c/debug_simple_interp_loop2.c
ltratt: I more meant: shouldn't we do it everywhere?

ptersilie:
> Last commit LGTM. Now just "guard failure" -> "guard".

I've commented on this already: https://github.com/ykjit/yk/commit/bab39a1e787cc73f484e9396ce4ddccbb4604212#r117269657

ltratt:
> I've commented on this already: https://github.com/ykjit/yk/commit/bab39a1e787cc73f484e9396ce4ddccbb4604212#r117269657

Agreed. So it's just the label issue (or not...) and we're done.

tests/c/debug_simple_interp_loop2.c
ptersilie: I just did the minimal change to get the test working and am leaning towards one place is enough to check this. But it's not a hill I'm going to die on, if you think it's useful.

ltratt: Fair enough.

ltratt:
Please squash and rebase.

ptersilie:
Oh, need to rebase too.

ltratt:
Yes, please force push.

ptersilie:
Rebased.

Once this change is in, we can tear out some of the deoptimisation logic (in ykllvm), since the control point will now always return a pointer (whereas before it could return null if the trace ran succesfully).

ptersilie:
Sorry, formatting failed. `cargo xtask fmt` doesn't work on my machine at the moment, and I thought I could get away with it. Ok to squash?

ltratt:
Please squash.

ptersilie:
Squashed.

PR #737, opened 2023-06-09
Spin for a bit while trying to gain a `HotLocation` lock.

ltratt:
Before this commit, if two threads tried to access the same `HotLocation` at the same time, at least one of them would immediately give up (depending on the underlying architecture, if weak CAS was used, both might give up!). This commit changes things so that threads spin for a bit to try and grab the lock in the expectation that contention will be brief. We use parking_lot_core's `SpinWait` as it gives us a good guide to "spinning doesn't look like it's worth it anymore: give up".

This makes one of our tests (only_one_thread_starts_tracing) a lot slower (0.4s vs 0.03s) but that is the absolute worse possible performance case: lots of threads hammer away at a `Location` without doing any other execution in between locking/unlocking. That could never happen in a real program, which must do at least a little bit of *something* between accessing a `Location`. Our more realistic benchmarks are unimpacted by this change, and I expect real VMs will be happier to spin briefly on the (probably rare) occasions they contend for a lock rather than immediately falling back to the interpreter.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -17,6 +17,7 @@ use std::{
 };
 
 use parking_lot::{Condvar, Mutex, MutexGuard};
+use parking_lot_core::SpinWait;
 #[cfg(feature = "yk_jitstate_debug")]
 use std::sync::LazyLock;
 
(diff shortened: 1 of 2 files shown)


ykrt/src/mt.rs
```diff
@@ -218,15 +219,30 @@ impl MT {
         let am_tracing = THREAD_MTTHREAD.with(|mtt| mtt.tracing.borrow().is_some());
         match loc.hot_location() {
             Some(hl) => {
-                // In general we don't want to contend with other threads, so if we can't grab the
-                // lock for the [HotLocation], we fall back to the interpreter. However, if this
-                // thread is tracing a loop, we absolutely have to grab the lock, because we need
-                // to see if this is the location we started at: if so, we have to stop tracing.
+                // If this thread is tracing something, we *must* grab the [HotLocation] lock,
+                // because we need to know for sure if `loc` is the point at which we should stop
+                // tracing. If this is thread is not tracing anything, however, it's not worth
```
ptersilie, line 224: one `is` too many

ptersilie:
Just one typo. LGTM.

ykrt/src/mt.rs
ltratt: Fixed in c8c0eb5.

ptersilie:
Please squash.

ltratt:
Squashed.

PR #738, opened 2023-06-10
This argument doesn't need to be mutable.

ltratt:


vext01:
Line 87 looks like mutation to me...

ltratt:
I see what you mean: depending on the config, we get `mut` warnings or not. FWIW, I think `YKD_FORCE_TRACE_DECODER` should be killed, as it suggests a global option when it's really PT specific. OK for me to remove it?

vext01:
> PT specific

PT specific in that we only have a PT decoder, if that's what you mean? It was intended to be general.

You can kill it if you want, but you may end up bringing it back when we have other kinds of decoder?

vext01:
Ah, just kill it -- we can add it back if need be.

ltratt:
I'll close this PR and open a new one with better info.

PR #742, opened 2023-06-12
Hoik the thread local access to the top of the function.

ltratt:
Since thread locals can (have to?) check for initialisation, and given that we always have to check parts of the thread local, it makes more sense to access it once at the top of the function, rather than multiple times inside. Apart from the indentation change (which makes this commit look large), the meaningful changes are small.

To my surprise, this gives a small, but statistically meaningful improvement in the `bench_single_threaded_control_point` benchmarks of about 4% -- I don't think this is a particularly worthwhile sellnig point for this commit though!


ptersilie:
Just a typo in the commit: `sellnig`. Please force push.

ltratt:
Force pushed commit typo update.

PR #743, opened 2023-06-12
The first incarnation of variable promotion.

vext01:
[Compiler-change ~~coming soon~~: [here](https://github.com/ykjit/ykllvm/pull/70)]

When a function is annotated with `__attribute__((yk_promote(...)))`, then the named arguments in `...` will be promoted to constants at trace -compilation time.

E.g.:

```
__attribute__((yk_promote(x,y))
void f(size_t x, size_t y, size_t z) {
  ...
}
```

A call to `f(1, 2, 3)` would have `x` and `y` baked in to the trace as constants `1` and `2` (and value guards would be inserted).

This change is just the runtime support. There is compiler support in ykllvm too.


tests/benches/promote.rs
```diff
@@ -0,0 +1,89 @@
...
+    let out = Command::new(bench_bin)
+        .arg(format!("{}", param))
+        .env("YKD_SERIALISE_COMPILATION", "1")
+        .env("YKD_PRINT_JITSTATE", "1")
+        .output()
+        .unwrap();
+    check_output(&out);
+}
+
+/// Benchmark a simple loop without variable promotion.
```
ptersilie, line 73: I think this comment needs to go a bit further down, since this function does indeed both with and without variable promotion.

yktracec/src/promote.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/// For the current thread, enable or disable constant recording.
+pub fn thread_record_enable(record: bool) {
+    VAL_REC.with_borrow_mut(|r| {
+        r.record_enable(record);
+    })
+}
+
+/// For the current thread, remove and return he oldest observation of the specified variable.
```
ptersilie, line 83: s/he/the

tests/benches/promote.c
```diff
@@ -0,0 +1,44 @@
+#include <assert.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+#ifdef DO_PROMOTE
```
ltratt, line 8: Why the `#ifdef`?

vext01: It's a program that's benchmarked with and without promotion, to see what difference it makes (if any!).

yktracec/src/promote.rs
```diff
@@ -0,0 +1,108 @@
...
+//!
+//! This is used for promoting variables to constants when compiling a trace.
+
+use std::{
+    cell::RefCell,
+    collections::{HashMap, VecDeque},
+    ffi::{c_char, CStr},
+};
+
+thread_local! {
```
ltratt, line 11: I wonder if this should, in the medium term, go into `MTThread`? That way we'd only have a single thread-local knocking around. Doesn't need to be done in this PR, though.

vext01: From memory, I failed to put this where I wanted due to circular crate dependencies.

yktracec/src/promote.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// This has to be specific to a thread to avoid cross-talk from other threads which may be
+    /// executing functions with promoted arguments at the same time.
+    pub static VAL_REC: RefCell<ValueRecorder> = RefCell::new(ValueRecorder::default());
+}
+
+/// The unique identifier for a promoted variable.
+#[derive(Debug, Eq, Hash, PartialEq)]
+pub struct PromoteKey {
+    /// The name of the function whose argument will be promoted.
+    func_name: String,
```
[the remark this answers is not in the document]
vext01: It must be unique, yes. Maybe we should rename it `func_sym_name` to reflect that?

PR #746, opened 2023-06-13
Remove YkCtrlPointVars stores from end of trace.

ptersilie:
At the end of each trace we previously wrote live variables back into the YkCtrlPointVars struct. Since we can only leave a trace via a guard failure this is no longer necessary (I'm not even sure it was even needed before that since all live variables are stored on the shadow stack anyway).

This further reduces the size of traces by a little bit, and should also result in a small performance boost as we no longer store all live variables on each loop iteration.


vext01:
LGTM

PR #750, opened 2023-06-15
A second stab at variable promotion.

vext01:
Compiler change: ~~coming soon~~ https://github.com/ykjit/ykllvm/pull/71

To hint to the JIT that a variable should be promoted, do:

  x = yk_promote(x);

During tracing, this function observes and records the concrete value of `x`. At trace compliation time, the value is fished out the value and replaces the variable with the observed constant.

As discussed with the team, some of the existing inefficiencies of our JITted code often prevent promotion from being effective. See the `promote_many_funcs.c` test for examples.

### DIFF
--- a/tests/benches/promote.c
+++ b/tests/benches/promote.c
@@ -1,0 +1,46 @@
+#include <assert.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+uint64_t inner(uint64_t a, uint64_t b, uint64_t x, uint64_t y, uint64_t z) {
+#ifdef DO_PROMOTE
+  a = yk_promote(a);
+  b = yk_promote(b);
+  x = yk_promote(x);
+  z = yk_promote(z);
+#endif
+  y += x * 3 - z;  // +1
+  y += a - 10 * b; // no-op
+  return y;
+}
+
+int main(int argc, char **argv) {
+  if (argc != 2) {
+    fprintf(stderr, "usage: promote <reps>\n");
+    exit(EXIT_FAILURE);
+  }
+  int reps = atoi(argv[1]);
+
+  YkMT *mt = yk_mt_new(NULL);
+  yk_mt_hot_threshold_set(mt, 0);
+  YkLocation loc = yk_location_new();
+
+  uint64_t a = 100, b = 10, x = 1, y = 0, z = 2;
+  NOOPT_VAL(a);
+  NOOPT_VAL(b);
+  NOOPT_VAL(x);
+  NOOPT_VAL(z);
+
+  for (int i = 0; i < reps; i++) {
+    yk_mt_control_point(mt, &loc);
+    y = inner(a, b, x, y, z);
+  }
+
+  assert(y == reps);
+  yk_location_drop(loc);
+  yk_mt_drop(mt);
+  return (EXIT_SUCCESS);
+}
--- a/tests/benches/promote.rs
+++ b/tests/benches/promote.rs
@@ -1,0 +1,91 @@
+//! Critereon benchmarks for measuring the impact of promotion.
+
+use criterion::{
+    criterion_group, criterion_main, measurement::Measurement, BenchmarkGroup, BenchmarkId,
+    Criterion, SamplingMode,
+};
+use std::{
+    env,
+    path::{Path, PathBuf},
+    process::{Command, Output},
+    time::Duration,
+};
+use tempfile::TempDir;
+use tests::mk_compiler;
+use ykbuild::ykllvm_bin;
+
+const SAMPLE_SIZE: usize = 30;
+const MEASUREMENT_TIME: Duration = Duration::from_secs(20);
+
+fn check_output(out: &Output) {
+    if !out.status.success() {
+        println!("{}", std::str::from_utf8(&out.stdout).unwrap());
+        eprintln!("{}", std::str::from_utf8(&out.stderr).unwrap());
+        panic!();
+    }
+}
+
+fn compile_bench(tempdir: &TempDir, promote: bool) -> PathBuf {
+    let md = env::var("CARGO_MANIFEST_DIR").unwrap();
+    let src = [&md, "benches", "promote.c"].iter().collect::<PathBuf>();
+
+    let mut exe = PathBuf::new();
+    exe.push(tempdir);
+    exe.push(src.file_stem().unwrap());
+
+    let mut compiler = mk_compiler(&ykllvm_bin("clang"), &exe, &src, "-O0", &[], true);
+    compiler.arg("-ltests");
+    if promote {
+        compiler.arg("-DDO_PROMOTE");
+    }
+    let out = compiler.output().unwrap();
+    check_output(&out);
+
+    exe
+}
+
+fn setup<'c, M: Measurement>(
+    c: &'c mut Criterion<M>,
+    group_name: &str,
+) -> (TempDir, PathBuf, PathBuf, BenchmarkGroup<'c, M>) {
+    let tempdir = TempDir::new().unwrap();
+    let bin_np = compile_bench(&tempdir, false);
+    let bin_p = compile_bench(&tempdir, true);
+
+    let mut group = c.benchmark_group(group_name);
+    group.sample_size(SAMPLE_SIZE);
+    group.measurement_time(MEASUREMENT_TIME);
+    group.sampling_mode(SamplingMode::Flat);
+
+    (tempdir, bin_np, bin_p, group)
+}
+
+fn run_bench(bench_bin: &Path, param: usize) {
+    assert!(bench_bin.exists());
+    let out = Command::new(bench_bin)
+        .arg(format!("{}", param))
+        .env("YKD_SERIALISE_COMPILATION", "1")
+        .env("YKD_PRINT_JITSTATE", "1")
+        .output()
+        .unwrap();
+    check_output(&out);
+}
+
+fn bench_promote(c: &mut Criterion) {
+    let (_tempdir, bin_np, bin_p, mut group) = setup(c, "promote");
+
+    let reps = 100000;
+    // Benchmark a simple loop *without* promotion.
+    group.bench_function(
+        BenchmarkId::new("without promotes", format!("{}", reps)),
+        |b| b.iter(|| run_bench(&bin_np, reps)),
+    );
+    // Benchmark a simple loop *with* promotion.
+    group.bench_function(
+        BenchmarkId::new("with promotes", format!("{}", reps)),
+        |b| b.iter(|| run_bench(&bin_p, reps)),
+    );
+}
+
+criterion_group!(promote_benchmarks, bench_promote);
+criterion_main!(promote_benchmarks);
--- a/tests/c/promote.c
+++ b/tests/c/promote.c
@@ -1,0 +1,72 @@
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_PRINT_JITSTATE=1
+//   env-var: YKD_PRINT_IR=jit-post-opt
+//   stderr:
+//     jit-state: start-tracing
+//     y=100
+//     jit-state: stop-tracing
+//     --- Begin jit-post-opt ---
+//     ...
+//     define ptr @__yk_compiled_trace_0(...
+//       ...
+//       %{{cond}} = icmp eq i64 {{x}}, 100
+//       br i1 %{{cond}}, label %{{succbb}}, label %{{failbb}}
+//
+//     {{failbb}}:...
+//       ...
+//       %{{deopt}} = call ptr (...) @llvm.experimental.deoptimize...
+//       ret ...
+//
+//     {{succbb}}:...
+//       ...
+//       %{{res}} = add {{size_t}} %{{arg1}}, 100...
+//       ...
+//       %{{cond2}} = icmp eq i64 {{x2}}, 100
+//       br i1 %{{cond2}}, label %{{succbb}}, label %{{failbb}}
+//     }
+//     ...
+//     --- End jit-post-opt ---
+//     y=200
+//     jit-state: enter-jit-code
+//     y=300
+//     y=400
+//     y=500
+//     jit-state: deoptimise
+//     jit-state: exit-jit-code
+
+// Check that promotion works in traces.
+
+#include <assert.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+size_t inner(size_t x, size_t y) {
+  size_t xp = yk_promote(x);
+  y += xp;
+  return y;
+}
+
+int main(int argc, char **argv) {
+  YkMT *mt = yk_mt_new(NULL);
+  yk_mt_hot_threshold_set(mt, 0);
+  YkLocation loc = yk_location_new();
+
+  size_t x = 100;
+  size_t y = 0;
+  NOOPT_VAL(x);
+
+  for (int i = 0; i < 5; i++) {
+    yk_mt_control_point(mt, &loc);
+    y = inner(x, y);
+    fprintf(stderr, "y=%" PRIu64 "\n", y);
+  }
+
+  NOOPT_VAL(y);
+  yk_location_drop(loc);
+  yk_mt_drop(mt);
+  return (EXIT_SUCCESS);
+}
--- a/tests/c/promote_guard.c
+++ b/tests/c/promote_guard.c
@@ -1,0 +1,87 @@
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_PRINT_JITSTATE=1
+//   env-var: YKD_PRINT_IR=jit-post-opt
+//   stderr:
+//     jit-state: start-tracing
+//     y=100
+//     jit-state: stop-tracing
+//     --- Begin jit-post-opt ---
+//     ...
+//     define ptr @__yk_compiled_trace_0(...
+//       ...
+//       %{{cond}} = icmp eq i64 {{x}}, 100
+//       br i1 %{{cond}}, label %{{succbb}}, label %{{failbb}}
+//
+//     {{failbb}}:...
+//       ...
+//       %{{deopt}} = call ptr (...) @llvm.experimental.deoptimize...
+//       ...
+//       ret ...
+//
+//     {{succbb}}:...
+//       ...
+//       %{{res}} = add {{size_t}} %{{arg1}}, 100...
+//       ...
+//       %{{cond2}} = icmp eq i64 %{{x2}}, 100
+//       br i1 %{{cond2}}, label %{{succbb}}, label %{{failbb}}
+//     }
+//     ...
+//     --- End jit-post-opt ---
+//     y=200
+//     jit-state: enter-jit-code
+//     y=300
+//     y=400
+//     y=500
+//     jit-state: deoptimise
+//     jit-state: exit-jit-code
+//     y=700
+//     jit-state: enter-jit-code
+//     y=800
+//     y=900
+//     y=1000
+//     jit-state: deoptimise
+//     jit-state: exit-jit-code
+//     y=1999
+
+// Check that promotions are guarded correctly.
+
+#include <assert.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+#define ELEMS 10
+
+size_t inner(size_t x, size_t y) {
+  size_t xp = yk_promote(x);
+  y += xp;
+  return y;
+}
+
+int main(int argc, char **argv) {
+  YkMT *mt = yk_mt_new(NULL);
+  yk_mt_hot_threshold_set(mt, 0);
+  YkLocation loc = yk_location_new();
+
+  // We will trace 100 baked into the trace, and every iteration where there's
+  // a 200 we should guard fail.
+  size_t xs[ELEMS] = {100, 100, 100, 100, 100, 200, 100, 100, 100, 999};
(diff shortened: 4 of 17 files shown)


tests/c/promote_many_funcs.c
```diff
@@ -0,0 +1,98 @@
...
+//     --- End jit-post-opt ---
+//     y=402
+//     jit-state: enter-jit-code
+//     y=603
+//     y=804
+//     y=1005
+//     jit-state: deoptimise
+//     jit-state: exit-jit-code
+
+// Check that many-variable, many-function  promotion works in traces.
```
ltratt, line 45: Double space after "many-function".

tests/c/promote_many_funcs.c
```diff
@@ -0,0 +1,98 @@
...
+//
+// FIXME: This test is showing that promotion is only on a best-effort basis.
```
ltratt, line 47: s/is only on a best-effort basis/is only a hint, rather than a guarantee/

yktracec/src/jitmodbuilder.cc
```diff
@@ -27,6 +29,8 @@ uint64_t getNewTraceIdx() {
   return TraceIdx;
 }
 
+// FIXME: A lot of these #defines should be C++ consts, then the compiler can
```
ltratt, line 32: This feels like an issue to be filed rather than a comment for this PR? [I agree with the comment FWIW!]

yktracec/src/jitmodbuilder.cc
```diff
@@ -1732,6 +1738,10 @@ class JITModBuilder {
             }
             CallStack.pushFrame(StackFrame::CreateForeignFrame());
             break;
+          } else if (CF->getName() == PromoteRecFnName) {
+            // A variable is being promoted to a constant.
```
ltratt, line 1742: I don't think this is quite correct: the result of an expression is being promoted to a constant. I can do `x = yk_promote(1)` (which is silly, I admit) or `x = yk_promote(1 + y)`, neither of which is a variable.

yktracec/src/jitmodbuilder.cc
```diff
@@ -1912,6 +1922,54 @@ class JITModBuilder {
...
+                     size_t CurBBIdx, size_t CurInstrIdx) {
+    // First lookup the constant value the trace is going to use.
+    uint64_t PConst = __yk_lookup_promote_usize();
+    Value *PConstLL =
+        Builder.getIntN(PointerSizedIntTy->getIntegerBitWidth(), PConst);
+    Value *PromoteVar = CI->getArgOperand(0);
+    Value *JITPromoteVar = getMappedValue(PromoteVar);
+
+    // Now emit a guard for the value, so that we deoptimise if the
+    // value ever changes.
```
ltratt, line 1937: I think this should be "so that we deoptimise if a later execution of the trace sees a different value for this `promote`" or similar.

yktracec/src/jitmodbuilder.cc
```diff
@@ -1912,6 +1922,54 @@ class JITModBuilder {
...
+    MPF->LastSMCall = cast<CallInst>(CI->getNextNonDebugInstruction());
+
+    // We need a way to independently reference both the pre-promoted variable
+    // and the newly promoted constant in our mappings. We acheive this is a
```
ltratt, line 1941: sp "achieve"

yktracec/src/promote.rs
```diff
@@ -0,0 +1,76 @@
...
+impl ValueRecorder {
+    /// Enable (`record=true`) or disable (`record=false`) observations.
+    pub fn record_enable(&mut self, record: bool) {
+        debug_assert_ne!(self.record_enable, record);
+        self.record_enable = record;
+    }
+
+    /// Record the concrete value of a variable to be promoted.
+    pub fn push(&mut self, val: usize) {
+        // OPT: a quicker way than a boolean check probably exists.
```
ltratt, line 38: I don't really understand this comment, and to the extent I do understand it, I'm not sure it's useful ;)

vext01: I think `self.record_enable` isn't a very good way to decide if to record or not. We discussed in the past some kind of indirect jump to skip, or not, recording.

ltratt: I'm not sure that is possible and, if it is, I don't think this comment will remind us, to be honest.

yktracec/src/jitmodbuilder.cc
vext01: Yes, I'd never thought of it like this.

So when we have `y = yk_promote(x)` we actually promote the (expression? lval?) `y` with the value of computing the *expression* `x`...

Do we want to allow things like:

```
return yk_promote(x + 1 / f());
```

?

We can reject that sort of thing if we don't want it, but at the same time, it will probably work...

If we want to allow it, I think we need to add some tests.

ltratt: I guess my question is: did you implement promoting of variables or values? I can see arguments for either of them, but I'm not sure which the PR actually does!

vext01: I intended to implement the former, but I think I've unwittingly implemented the latter. I'll experiment tomorrow.

ltratt: Is the difference important? I'm not sure it is, at least not at the moment.

vext01: I don't think it matters! The latter is more flexible for sure.

vext01: Yes, but let me add tests to ensure it does actually work :)

ptersilie: I assumed `yk_promote(x + y)` will just just promote whatever LLVM turns that expression into. E.g. in LLVM it would look like this:

```
$3 = add $1, $2
promote($3)
```

So it should just work™?

vext01: Spoiler: it does :)

yktracec/src/jitmodbuilder.cc
ptersilie: Maybe something like:
```
Even though the call `y = yk_promote(x)` is not copied over into the trace and `y` is being replaced with a constant,  we need keep hold of the original value of `y` so it can be accessed during deoptimisation. The simplest way is by inserting a dummy instruction that will be copied into the trace, creating a mapping back to its AOT value.
```

This is not great either (this is difficult to describe), but maybe you can merge this with your comment somehow?

yktracec/src/jitmodbuilder.cc
```diff
@@ -1912,6 +1922,54 @@ class JITModBuilder {
...
+    // slightly hacky way by making a dummy instruction that we can map to the
+    // pre-promoted variable.
+    Instruction *NewInst = SelectInst::Create(
```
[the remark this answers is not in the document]
ptersilie: Maybe something like:
```
Even though the call `y = yk_promote(x)` is not copied over into the trace and `y` is being replaced with a constant,  we need keep hold of the original value of `y` so it can be accessed during deoptimisation. The simplest way is by inserting a dummy instruction that will be copied into the trace, creating a mapping back to its AOT value.
```

This is not great either (this is difficult to describe), but maybe you can merge this with your comment somehow?

yktracec/src/jitmodbuilder.cc
vext01: Test showing it works in 27b51e994f1bec0c07211ec4e7f6edecb990f879

tests/c/promote_many_funcs.c
vext01: Fixed in Fixed in 96f196064a3dafea592a5e5a3da4c750e1a06e0a.

yktracec/src/jitmodbuilder.cc
vext01: Removed in Fixed in 96f196064a3dafea592a5e5a3da4c750e1a06e0a, in favour of #753

yktracec/src/jitmodbuilder.cc
vext01: I tried a rewording in 96f196064a3dafea592a5e5a3da4c750e1a06e0a. Any good?

yktracec/src/jitmodbuilder.cc
vext01: I really struggled with this comment, and I'm not 100% sure I understand.

Let's recap. We have:

```
    Instruction *NewInst = SelectInst::Create(
        ConstantInt::get(Type::getInt1Ty(JITMod->getContext()), 0),
        JITPromoteVar, JITPromoteVar);
    Builder.Insert(NewInst);

    VMap[CI] = NewInst;
    insertAOTMap(CI, NewInst, CurBBIdx, CurInstrIdx);

    BasicBlock *FailBB = getGuardFailureBlock(BB, CurBBIdx, CI, CurInstrIdx);
    ...

    VMap[CI] = PConstLL;
```

You'd expect to be able to kill the dummy instruction, and use `JITPromoteVar` directly, but:
```
    VMap[CI] = JITPromoteVar;
    insertAOTMap(CI, NewInst, CurBBIdx, CurInstrIdx);
    ...
```

But this gives:
```
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', ykrt/src/frame/mod.rs:271:42
```

https://github.com/ykjit/yk/blob/9f26837c47a8c28ea913cd0d15c50bc7c7b04b12/ykrt/src/frame/mod.rs#L271

And around here I get lost. We try to look something up in the frame that the stackmap knows about, but t he frame doesn't...

@ptersilie remind me how it works again?

ptersilie: Without the dummy instruction you map `JITPromoteVar` to two different AOT vars, which we can't handle. To fix this we could change the AOT mapping to allow multiple values, e.g. AOTMap[JITVar] = [AOT1, AOT2], but that's not a trivial change. Using a dummy instruction for now is much easier.

yktracec/src/jitmodbuilder.cc
vext01: At the llvm level, yes, you pass an SSA variable to `yk_promote` and receive back a distinct SSA variable which you should then use henceforth.

ltratt: OK, so imagine I have this C code:

```
x = 1 + 2 + 3;
y = promote(x);
f(y);
f(x);
```

When I call `f(x)` is that with a promoted value (in the sense that I've promoted `x` indirectly) or is only `f(y)` with a promoted value?

vext01: Only `f(y)` I reckon.

Do you agree @ptersilie?

[In your example `promote()` probably slows the code down, because `promote` is opaque and then
 the compiler can't assume `x==y` and that `y` is statically constant :)]

ltratt: If this is the case (and I suspect it is!) can we add a test that checks so?

yktracec/src/jitmodbuilder.cc
vext01: Fixed in 32e3f65d6ee9735f1324f5b0fc46449af2be23bb

yktracec/src/jitmodbuilder.cc
vext01: Tested in 04790bbf4de1566785ba3803c7049a075edf5872

ltratt: @vext01 Perfect -- thanks!

yktracec/src/jitmodbuilder.cc
ptersilie: New comment looks good!

vext01:
Is this ok to squash then?

ptersilie:
Fine by me, but should wait for @ltratt's ok too.

tests/benches/promote.c
```diff
@@ -0,0 +1,44 @@
+#include <assert.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+#ifdef DO_PROMOTE
+__attribute__((yk_promote("a", "b", "x", "z")))
```
ltratt, line 9: I thought we'd got rid of this attribute?

tests/benches/promote.rs
```diff
@@ -0,0 +1,89 @@
+//! Critereon benchmarks for measuring the impact of variable promotion.
```
ltratt, line 1: s/variable//

tests/benches/promote.rs
```diff
@@ -0,0 +1,89 @@
...
+    let out = Command::new(bench_bin)
+        .arg(format!("{}", param))
+        .env("YKD_SERIALISE_COMPILATION", "1")
+        .env("YKD_PRINT_JITSTATE", "1")
+        .output()
+        .unwrap();
+    check_output(&out);
+}
+
+/// Benchmark a simple loop without variable promotion.
```
ltratt, line 73: s/variable//

tests/c/promote.c
```diff
@@ -0,0 +1,72 @@
...
+//     --- End jit-post-opt ---
+//     y=200
+//     jit-state: enter-jit-code
+//     y=300
+//     y=400
+//     y=500
+//     jit-state: deoptimise
+//     jit-state: exit-jit-code
+
+// Check that variable promotion works in traces.
```
ltratt, line 38: s/variable//

tests/c/promote_guard.c
```diff
@@ -0,0 +1,87 @@
...
+//     y=700
+//     jit-state: enter-jit-code
+//     y=800
+//     y=900
+//     y=1000
+//     jit-state: deoptimise
+//     jit-state: exit-jit-code
+//     y=1999
+
+// Check that variable promotions are guarded correctly.
```
ltratt, line 47: s/variable//

tests/c/promote_many_funcs.c
```diff
@@ -0,0 +1,99 @@
...
+//     --- End jit-post-opt ---
+//     y=402
+//     jit-state: enter-jit-code
+//     y=603
+//     y=804
+//     y=1005
+//     jit-state: deoptimise
+//     jit-state: exit-jit-code
+
+// Check that many-variable, many-function promotion works in traces.
```
ltratt, line 45: I'm not really sure if this test is useful: multiple functions don't seem meaningfully different than single functions?

yktracec/src/jitmodbuilder.cc
```diff
@@ -1732,6 +1736,10 @@ class JITModBuilder {
             }
             CallStack.pushFrame(StackFrame::CreateForeignFrame());
             break;
+          } else if (CF->getName() == PromoteRecFnName) {
+            // A variable is being promoted to a constant.
```
ltratt, line 1740: s/variable/value/

yktracec/src/jitmodbuilder.cc
```diff
@@ -1912,6 +1920,63 @@ class JITModBuilder {
     finalise(AOTMod, &Builder);
     return JITMod;
   }
+
+  /// Handle promoted variables.
```
ltratt, line 1924: "Handle promotions"?

yktracec/src/jitmodbuilder.cc
```diff
@@ -1912,6 +1920,63 @@ class JITModBuilder {
...
+    BasicBlock *SuccBB = BasicBlock::Create(JITMod->getContext(),
+                                            GUARD_SUCCESS_BLOCK_NAME, JITFunc);
+    Value *Deopt = Builder.CreateICmp(CmpInst::Predicate::ICMP_NE,
+                                      JITPromoteVar, PConstLL);
+    Builder.CreateCondBr(Deopt, FailBB, SuccBB);
+
+    // Carry on constructing the trace module in the success block.
+    Builder.SetInsertPoint(SuccBB);
+
+    // Update the VMap so that henceforth, the promoted variable is in
```
ltratt, line 1969: Are we tracking variables here?

yktracec/src/promote.rs
```diff
@@ -0,0 +1,75 @@
+//! Software constant recorder.
+//!
+//! This is used for promoting variables to constants when compiling a trace.
```
ltratt, line 3: s/variables/values/

yktracec/src/promote.rs
```diff
@@ -0,0 +1,75 @@
...
+thread_local! {
+    /// The thread's value recorder.
+    ///
+    /// This has to be specific to a thread to avoid cross-talk from other threads which may be
+    /// executing functions with promoted arguments at the same time.
+    pub static VAL_REC: RefCell<ValueRecorder> = RefCell::new(ValueRecorder::default());
+}
+
+/// A value recorder observes and records constant values during tracing. The trace compiler
+/// queries this when building a trace to replace promoted variables with the observed constants.
```
ltratt, line 17: s/variables/values/

yktracec/src/promote.rs
```diff
@@ -0,0 +1,75 @@
...
+#[derive(Debug, Default)]
+pub struct ValueRecorder {
+    // `true` when recording promotions.
+    record_enable: bool,
+    // A vector of observed values of to-be-promoted variables, in the order their `yk_promote()`
```
ltratt, line 22: "A vector of promoted values in the order their `yk_promote()` calls were encountered in the trace"?

yktracec/src/promote.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /// Record the concrete value of a variable to be promoted.
+    pub fn push(&mut self, val: usize) {
+        if self.record_enable {
+            self.pbuf.push_back(val);
+        }
+    }
+
+    /// Remove and return he oldest recorded value.
```
ltratt, line 43: s/he oldest/the oldest/

yktracec/src/promote.rs
```diff
@@ -0,0 +1,75 @@
...
+    })
+}
+
+/// For the current thread, remove and return he oldest recorded value.
+#[no_mangle]
+pub extern "C" fn __yk_lookup_promote_usize() -> usize {
+    VAL_REC.with_borrow_mut(|r| r.pop())
+}
+
+/// For the current thread, record the concrete value of a variable to be promoted.
```
ltratt, line 65: "Promote a value: the returned value will be considered a constant in a trace. When not tracing this function is a no-op"?

ltratt:
Mostly I found a whole load of places where we say "promote variable" in ways that I believe are not quite right. Most of these are easily fixed: one (maybe two) might be subtler. I think the simple ones can all be done in a single commit: any that look a bit more complex might be best done as separate commits.

yktracec/src/jitmodbuilder.cc
ltratt: @ptersilie If you're happy with this please "resolve conversation".

yktracec/src/jitmodbuilder.cc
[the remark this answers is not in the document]
vext01: I believe this is resolved @ptersilie ?

tests/benches/promote.c
vext01: So did I! Investigating.

yktracec/src/jitmodbuilder.cc
vext01: This one is actually fine since we are working at the LLVM level.

In LLVM prose `Value` is the type of an "SSA variable". I'll update it anyway.

yktracec/src/jitmodbuilder.cc
vext01: Similarly to before, value and variable are (pretty much) the same at the llvm-level.

tests/c/promote_many_funcs.c
vext01: Killed in 3b9f810f45ca02020c1e7385dac60961deffb30c

tests/benches/promote.c
vext01: Fixed in db05458ef20e9e61ed2603a6e1a3032ef79707ae.

I wasn't running `cargo bench` locally, but even if I had, i might have missed this! Additional fix in 307c7ca6cc7191dba3cb79529b008b65b52f3fcd.

vext01:
All of the minor renamings are done in 321c28292e1b2e672898e6878bf820c789a7b1e2.

Is that everything? I think so.

yktracec/src/jitmodbuilder.cc
ltratt: Maybe say "LLVM Value" to disambiguate?

vext01: Fixed in 0d7133a24584c86f306688cdaa5fa1172ccc2306

ltratt: Is it worth (not in this PR!) homogenising that at some point?

vext01: I'm not against it.

ltratt: OK.

vext01:
I've just realised that the test we removed was demonstrating promotion not working as well as it should -- we spoke about having a test like that so we can notice when promotion works better.

Is that still the plan? bring it back, potentially renamed?

ltratt:
> I've just realised that the test we removed was demonstrating promotion not working as well as it should -- we spoke about having a test like that so we can notice when promotion works better.

Ah! Yeah, let's bring it back, but maybe rename and simplify it a bit if we can?

vext01:
Also, this probably needs to sync the submodule, otherwise the test rely on a feature the compiler won't have... Right?

ltratt:
> Also, this probably needs to sync the submodule

s/probably// :)

You can use the commit hash from your ykllvm PR for this (you don't have to wait for the merge).

vext01:
OK, I need to figure out how to do it.

yktracec/src/jitmodbuilder.cc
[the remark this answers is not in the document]
ptersilie: Works for me.

vext01:
splat.

vext01:
There we go. Ready.

PR #751, opened 2023-06-15
Rename `hwtracer` to `perftracer`

ltratt:
In the past (and there are still bits of those left), the design of hwtracer-as-was had abstractions intended to allow it to be generic over different hardware-tracing backends. However, that hasn't quite worked out as we expected: because the interface to a tracer (PT etc.) goes via an OS, hwtracer has implicitly ended up being heavily tied to Linux-perf.

From yk's perspective, this makes little difference: ykrt has its own abstractions over tracing backends, hopefully allowing it to handle any hardware or software backend we come across.

So rather than try and make hwtracer internally generic, this PR bites the bullet and acknowledges that hwtracer is always likely to be a Linux-perf-only tracer -- and renames it perftracer. https://github.com/ykjit/yk/commit/9871f9e2ae480171185bb856f39a0c4ef6d9a676 is the minimal renaming commit; https://github.com/ykjit/yk/commit/3346b990a29b7e77d71c0b2c1e6f749500a1bd37 removes the `perf_collector` feature.

There is quite a bit more simplification that could be done, but I'd rather check that this matches what we thought before going further.


ltratt:
ykrt already has the necessary abstractions https://github.com/ykjit/yk/blob/master/ykrt/src/trace/mod.rs#L370

vext01:
Something feels wrong. hwtracer isn't only about trace collectors (like `perf`). It also contains trace decoders (like `ykpt`). How you decode a trace really has nothing to do with how you collect it, so if we rename `hwtracer` to `perftracer`, then what business does a PT decoder have being in there?

Do you know what I mean?

ltratt:
I haven't yet worked out where the decoders should go: I *think* they should go into ykrt or maybe another crate. But since we only have one, it's a bit difficult to know for sure, because there is probably some coupling between "who did the tracing" and the decoder.

That does make me realise that another way of renaming/thinking about hwtracer is that it is `pttracer` and then you have Linux/BSD/etc support, as well as different PT decoders.

But, one way or another, I think that having hwtracer try to abstract over both the OS *and* the CPU feels like one abstraction too many. So I'd like to get rid of one of 'em, but I don't feel very strongly about which one: of the two, perhaps even the `pttracer` idea makes more sense to me!

vext01:
The more I think about it, the more I think hwtracer makes sense as a collection of trace collectors and decoders. That seems like quite a natural grouping, doesn't it?

> one abstraction too many

Maybe hwtracer should be the only point of abstraction?

ltratt:
> The more I think about it, the more I think hwtracer makes sense as a collection of trace collectors and decoders.

I suppose we could do that. Then `hwtracer` would become `tracer` and ykrt would be a bit simpler. Let me have a think (e.g. would that cope OK with software tracing?).

vext01:
I think we could have a software trace collector and a really insulting simple "decoder" which basically is a no-op :)

(Sorry if I've been awkward in this PR)

ltratt:
I've had another look into this.

First, it's not really possible to have a clean "collector/decoder" split, because there are (at least) three dimensions: OS API (e.g. Linux Perf), tracer (e.g. Intel PT), and tracer-specific decoder (e.g. ykpt). Anything we do is going to have some mess to it. Fortunately, ykrt has traits to abstract away whatever tracing backend we use, so that's not a real concern.

Second, hwtracer as it stands today hardcodes perf + PT all over the place: e.g. the perf config are in the should-be-agnostic collector module; the Perf collector hardcodes PT. Splitting this apart is of course possible, but it's going to be a lot of work: it could easily take me a couple of weeks as I have little knowledge of the many low-level APIs that are used. Until/unless we support another OS or tracer (e.g. Arm) I don't think that's useful to do now. 

Perhaps, ultimately, this should be a `tracer` crate with submodules for things like `perf` and `pt`. That's the least messy design I can come up with. If that works for you, I can perhaps do a very minimal job at that, though the full job is (as I mentioned above) more work than I want to take on now.

vext01:
Good to squash this and the `yk` companion?

ltratt:
> Which part of the ykpt is specific to perf?

My point exactly! ykpt (I think) is generic to a PT trace, but we might have multiple (OS) collectors that can generate a PT trace. But ykpt can't, for example, decode a CoreSight trace. Hence why I *think* there are at least three dimensions.

Anyway, I'm going to close this. I think refactoring hwtracer is a big job, and this PR isn't a step in the probable direction we want to take.

vext01:
> My point exactly! ykpt (I think) is generic to a PT trace, but we might have multiple (OS) collectors that can generate a PT trace. But ykpt can't, for example, decode a CoreSight trace.

Right, so that's why I tried to keep all encoders in `src/encode` and all decoders in `src/decode`. I imagined an API which would choose sensible defaults for the platform for users that don't care, or allow power users to choose their own collector/decoder combo, but reject invalid combos.

PR #755, opened 2023-06-19
Add llvm namespace to bit_cast call

Pavel-Durov:
bit_cast call can be ambiguous since the 
+same function is defined and used in the standard library. Adding llvm namespace call removes this ambiguity.

This change fixes the following build-time error:
```
cargo:warning=src/jitmodbuilder.cc:375:28: error: call to 'bit_cast' is ambiguous
  cargo:warning=          UnmappableRegion{bit_cast<ssize_t, size_t>(BBs[Idx])}});
  cargo:warning=                           ^~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/yktracec/src/jitmodbuilder.cc
+++ b/yktracec/src/jitmodbuilder.cc
@@ -372,7 +372,7 @@ public:
       // (a `size_t`) to encode the stack adjustment value (a `ssize_t`). The
       // cast below reverses that.
       return TraceLoc(variant<IRBlock, UnmappableRegion>{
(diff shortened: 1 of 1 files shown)


vext01:
Having spoken on the team sync meeting, I don't see any issue with using the `std` version of bitcast -- I'm not sure what the problem there was, but then again there's no issue with using the llvm bitcast either (assuming it has the same semantics -- I'd be surprised if it didn't).

[I notice the llvm version has a bunch of optional args that the std lib one doesn't...]

[also the llvm bit_cast defers to some compiler internal if it can: https://reviews.llvm.org/D142305]

bors r+

PR #757, opened 2023-06-19
Add _IO_fflush as an assertion option for dli_snam result 

Pavel-Durov:
Add _IO_fflush as an assertion option for the `dli_snam` result
On some machines `dli_sname` returns `_IO_fflush` and `fflush` on others for `libc::fflush`.


vext01:
LGTM

bors r+

PR #762, opened 2023-06-21
Deoptimise without returning via control point.

ptersilie:
After deoptimisation we used to return a pointer to the new stack frames back to the control point which, after a quick comparison, ran `__ykrt_reconstruct_frames` to copy over the new stack. This design would make it slightly annoying dealing with side-traces as we'd have to bubble up this pointer all the way back to the control point.

This change allows us to call `__ykrt_reconstruct_frames` immediately during deoptimisation, jumping to the correct location in the main interpreter without having to go via the control point.


ptersilie:
@ltratt Could you kindly double check that not dropping the `Arc` in https://github.com/ykjit/yk/pull/762/files#diff-9fa85f3ab0ab08c2bb4337dd17d47605f520c2780158e50e66eff7cd9dc1b29bR166 is ok here?

ltratt:
> @ltratt Could you kindly double check that not dropping the `Arc` in https://github.com/ykjit/yk/pull/762/files#diff-9fa85f3ab0ab08c2bb4337dd17d47605f520c2780158e50e66eff7cd9dc1b29bR166 is ok here?

I believe this is fine: the `Arc` is referenced, not dropped, so its ref count isn't being mutated.

ykrt/src/deopt.rs
```diff
@@ -255,15 +255,25 @@ unsafe extern "C" fn __ykrt_deopt(
         }
     }
 
-    unsafe { framerec.reconstruct_frames(frameaddr) }
+    // FIXME: This print doesn't really serve a purpose anymore, since `deoptimise` will always be
+    // followed by `exit-jit-code` as there's no other way to leave a trace.
+    #[cfg(feature = "yk_jitstate_debug")]
+    print_jit_state("exit-jit-code");
```
vext01, line 261: So this is never printed?

Surely we don't want that, and it must break lots of tests?

Perhaps the deopt code should print this somehow?

ykrt/src/deopt.rs
```diff
@@ -255,15 +255,25 @@ unsafe extern "C" fn __ykrt_deopt(
...
+    unreachable!();
 }
 
 /// After a guard failure, reconstructs the stack frames and registers and then jumps back to the
 /// AOT code from where to continue.
 #[cfg(target_arch = "x86_64")]
 #[naked]
 #[no_mangle]
-extern "C" fn __ykrt_reconstruct_frames(newframesptr: *const c_void) {
+extern "C" fn __ykrt_reconstruct_frames(newframesptr: *const c_void, newframedst: *const c_void) {
```
vext01, line 276: What was the outcome of the `-> !` discussion?

ykrt/src/deopt.rs
ltratt: This is the deopt code.

I think what Lukas is saying is that the only way for a trace to complete is for a guard to fail so "deoptimise" subsumes "exit-jit-code".

ykrt/src/frame/mod.rs
```diff
@@ -175,6 +177,17 @@ impl FrameReconstructor {
             // reconstructed.
             if i > 0 {
                 memsize += usize::try_from(rec.size).unwrap();
+            } else {
+                // Get the size of the frame containing the control point, which we'll later use to
+                // calculate the stack offset to write the new frames to. Note that we'll be adding
+                // this to a pointer retrieved via the `llvm.frameaddr` intrinsic, which doesn't
+                // include the pushed RBP at the beginning of the frame, whereas the size reported
```
vext01, line 184: RBP -> base pointer, since RBP is an x86_64 term and this file is (in theory) generic.

ykrt/src/deopt.rs
```diff
@@ -255,15 +255,25 @@ unsafe extern "C" fn __ykrt_deopt(
         }
     }
 
-    unsafe { framerec.reconstruct_frames(frameaddr) }
+    // FIXME: This print doesn't really serve a purpose anymore, since `deoptimise` will always be
```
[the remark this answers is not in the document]
ptersilie: Okay, it will require touching a lot of tests, so will do it in a separate commit.

ykrt/src/mt.rs
```diff
@@ -163,10 +163,8 @@ impl MT {
             TransitionLocation::Execute(ctr) => {
                 #[cfg(feature = "yk_jitstate_debug")]
                 print_jit_state("enter-jit-code");
-                let ptr = ctr.exec(ctrlp_vars, frameaddr);
-                #[cfg(feature = "yk_jitstate_debug")]
-                print_jit_state("exit-jit-code");
-                return ptr;
+                ctr.exec(ctrlp_vars, frameaddr);
```
vext01, line 166: Did you ask Laurie if the `Arc` remains consistent when you rug-pull Rusts's stack?

ykrt/src/deopt.rs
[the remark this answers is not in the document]
ltratt: Just check with @vext01 that he agrees with this too!

ykrt/src/trace/mod.rs
```diff
@@ -299,47 +299,11 @@ impl CompiledTrace {
...
-    /// Taking the deoptimisation path during a guard failure causes the epilogue of the compiled
-    /// trace to be skipped. This means that used callee-saved registers (CSRs) are not restored.
-    /// Until we've figured out how to restore only the used registers we take the sledge hammer
-    /// approach and save and restore all CSRs here.
-    /// OPT: Find a way to only restore needed registers (ideally right within the deopt code).
     pub extern "C" fn exec(
         &self,
         ctrlp_vars: *mut c_void,
         frameaddr: *mut c_void,
-    ) -> *const c_void {
```
vext01, line 314: Big deletions == good!

ykrt/src/deopt.rs
ptersilie: What @ltratt said. "deoptimise" will always be followed by "exit-jit-code" so there's no need to have both. I'll update the tests accordingly.

vext01: As long as we still get a jitstate print when we exit jitted code, I'm happy :)

vext01: Ah! I get it! Thanks

ykrt/src/deopt.rs
ptersilie: Not sure, @ltratt alluded in the call there's problems with it, so I wanted to get his opinion first.

ykrt/src/frame/mod.rs
ptersilie: The function is marked as `#[cfg(target_arch = "x86_64")]`, and I'm using the `RBP` term all over the place. I think it's fine.

ykrt/src/mt.rs
ptersilie: Yeah, see main conversation.

ykrt/src/frame/mod.rs
vext01: Ah, yes. OK!

ykrt/src/deopt.rs
ltratt: I think it *should* work here, so just try it. If it compiles, it works!

ptersilie: [88c2a64](https://github.com/ykjit/yk/pull/762/commits/88c2a64d7344ad43c881c1969b10c4f49b8a217b)

ykrt/src/deopt.rs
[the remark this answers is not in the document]
ptersilie: [878bfa3](https://github.com/ykjit/yk/pull/762/commits/878bfa3451e28df7d5942efa746e5d116e713e95)

ptersilie:
Ready for re-review.

vext01:
Last batch of commits LGTM.

ltratt:
I still need to go through the implications of not unwinding. Bear with me -- this might take me a while.

ptersilie:
Ah no problem. I can wait.

ptersilie:
I've just pushed a commit implementing the changes discussed earlier. We now pass the `Arc<CompiledTrace>` into the compiled trace, so we can properly drop the `Arc`. I took the opportunity to simplify the arguments to the compiled trace, which required a bit of rewriting in `jitmodbuilder.cc` and `deopt.rs`. But I think the code is a bit neater now. Can you imagine that after changing this is ran on the first try! :flushed:

@ltratt I also realised that we now have the `CompiledTrace` inside deopt, exactly like we wanted to for side traces. So we probably don't need your branch anymore which retrieves the `CompiledTrace` via an atomic. Would you agree?

ltratt:
> @ltratt I also realised that we now have the CompiledTrace inside deopt, exactly like we wanted to for side traces. So we probably don't need your branch anymore which retrieves the CompiledTrace via an atomic. Would you agree?

It wasn't done via atomics but I agree with your bigger point: we probably don't need the branch I created now. I'll keep it lurking around for a bit just in case we change our mind, and if we don't need it in say a week or so, I'll kill it.

The new commit looks good to me. @vext01 are you with it too?

vext01:
Looks OK to me. At least: this is what I expected. A lot of detail I can't vouch for.

Let's move ahead?

ltratt:
Please squash.

ptersilie:
Squashed. But had to do some formatting afterwards. So needs another squash.

ltratt:
Please squash.

ptersilie:
And squashed again!

ptersilie:
bors try

ptersilie:
Ok to squash?

ltratt:
Please squash.

PR #763, opened 2023-06-29
Add a wrapper around gdb_c_test to make it easier to use.

vext01:
Running the tool from cargo is a bit lengthy, and you have to remember where to put the `--` to start passing args to the underlying binary.

By using a shell wrapper:

```
cargo run --bin gdb_c_test -- -j -s -b10 simple.c
```

can become just:

```
bin/gdb_c_test -j -s -b10 simple.c
```


ptersilie:
Very good.

bors r+

PR #764, opened 2023-06-30
Adjust to control point simplification.

ptersilie:
These changes are a reaction to https://github.com/ykjit/ykllvm/pull/73

Due to the control point no longer returning its design has changed leading to some more simplfication on the Yk side too. See commits for details.


yktracec/src/jitmodbuilder.cc
```diff
@@ -1811,10 +1752,12 @@ class JITModBuilder {
                 assert(cast<GetElementPtrInst>(I)->getPointerOperand() ==
                        TraceInputs);
                 I++;
+                CurInstrIdx++;
```
ltratt, line 1696: Are `I` and `CurInstrIdx` now effectively the same value -- they seem to be always incremented together?

vext01, line 1696: Why all the `CurInstrIdx++`?

vext01:
So much code goes away :)

yktracec/src/jitmodbuilder.cc
vext01: I is an instruction iterator though.

ptersilie: `I` is an iterator pointing to an `Instruction`. `CurInstrIdx` is an integer referring to the index of that particular instruction in its parent basic block.

ltratt: Does the iterator / instruction know its index?

vext01: AFAIK, it does not :)

ptersilie: I need them for the `insertAOTMap` further below.

ptersilie:
> So much code goes away :)

It's great isn't it!

ptersilie:
Oh, so what do I have to do to link the ykllvm submodule here. I want to link the first commit to the first commit in the ykllvm PR, and the second commit here to the second commit in the ykllvm PR. Do I need to wait for the ykllvm PR to be merged first?

yktracec/src/jitmodbuilder.cc
vext01: I have a vague memory (maybe fabricated) that there is an iterator adaptor for enumerating iterators, but we'd have to be careful to always add an offset since we don't always start from instruction index 0 (due to resume points).

In fact, I think it would make comprehension harder, so let's not do that.

ltratt:
> Oh, so what do I have to do to link the ykllvm submodule here.

There's more than one way of doing it. That said, what I suggest for the initial PR is to change the submodule commit to your head commit in the ykllvm PR. If the ykllvm PR is squashed and the commit hash changes, it will then have to be changed here.

ltratt:
@ptersilie I suggest you add a new commit which uses the commit from the head of your ykllvm branch now (while ykllvm is going through CI).

ptersilie:
Added the submodule stuff. Will need squashing into the respective commits.

ltratt:
Please squash.

ptersilie:
bors r-

Sorry, some formatting I forgot.

ltratt:
Please squash.

ptersilie:
Done. Let's go!

PR #766, opened 2023-07-01
Explicitly drop values.

ltratt:
Newer rustc nightly warns that values like this are unused and should be explicitly dropped if that's what intended. That is what we intended, so this commit explicitly drops them.


vext01:
Do they not get dropped when they go out of scope?

ltratt:
> Do they not get dropped when they go out of scope?

Yes, but newer Rust now warns that you should explicitly drop them. [Which is a good idea IMHO.]

vext01:
Every value? Or just boxes?

ltratt:
The two instances in this PR cause new rustc nightly to generate warnings.

vext01:
Not sure i get it, but if rust warns, so be it.

bors r+

PR #768, opened 2023-07-02
Tidy up the control point a little bit

ltratt:
The main (though it's hardly big!) commit here is https://github.com/ykjit/yk/commit/8ae512e8b1df0baccb54326fee56df3c0783c3bc which removes the `*const c_void` return type from `MT::control_point`.

### DIFF
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -62,7 +62,7 @@ pub extern "C" fn __ykrt_control_point(
     if !loc.is_null() {
         let mt = unsafe { &*mt };
         let loc = unsafe { &*loc };
-        return mt.control_point(loc, ctrlp_vars, frameaddr);
+        mt.control_point(loc, ctrlp_vars, frameaddr);
     }
     std::ptr::null()
 }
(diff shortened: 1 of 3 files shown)


ykcapi/src/lib.rs
```diff
@@ -62,7 +62,7 @@ pub extern "C" fn __ykrt_control_point(
     if !loc.is_null() {
         let mt = unsafe { &*mt };
         let loc = unsafe { &*loc };
-        return mt.control_point(loc, ctrlp_vars, frameaddr);
+        mt.control_point(loc, ctrlp_vars, frameaddr);
     }
     std::ptr::null()
```
ptersilie, line 67: I think we can remove this too and update the signature above.

ptersilie:
Some comments.

ltratt:
I think we should wait for #764 to go in, then I'll come back to this PR.

yktracec/src/jitmodbuilder.cc
```diff
@@ -1332,6 +1332,7 @@ class JITModBuilder {
   GetControlPointInfo(Module *AOTMod) {
     Function *F = AOTMod->getFunction(YK_NEW_CONTROL_POINT);
     assert(F->arg_size() == YK_CONTROL_POINT_NUM_ARGS);
+    assert(F->getReturnType()->isPointerTy());
```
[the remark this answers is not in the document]
ltratt: Yep, this was changed by #764! So now fixed in b2573f4.

ykcapi/src/lib.rs
ptersilie: Huh, that's odd. Okay I will look into it independently.

ptersilie:
Please squash if needed.

ltratt:
Squashed.

PR #770, opened 2023-07-03
Store `MT`s in `Arc`s

ltratt:
This turned out to be easier than I expected (famous last words).

Note that `self: Arc<Self>` would probably be more common than `self: Arc<&Self>`, but for `MT` we can avoid incing/decing the `Arc` in many instances through the C API by using `self: Arc<&Self>` and `forget`. This seems like a reasonable compromise: if it turns out to be a problem we can probably change to the other representation easily enough.

### DIFF
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -11,14 +11,15 @@
 use std::{
     ffi::{c_char, c_void, CString},
     ptr,
+    sync::Arc,
 };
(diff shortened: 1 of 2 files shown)


ltratt:
bors try

ltratt:
@ptersilie Note that there was a merge conflict with master, so this will definitely need squashing before merging.

ptersilie:
This looks like what I need for side traces. Please squash.

PR #779, opened 2023-07-11
Start optimising the AOT module.

vext01:
**Companion PR: https://github.com/ykjit/ykllvm/pull/74**

Until now we've required the interpreter to be built with zero AOT optimisation. This is because we can't tolerate optimisations that change the block structure in the codegen backend: this would cause our mapper to break.

That's not to say we can't do *any* optimisation.

This change turns on one AOT optimisation: the instrcombine pass: https://llvm.org/doxygen/InstructionCombining_8cpp_source.html

(It does some simplifications and constant folding etc)

There are some fiddly details:

 - We still can't let the backend optimise things. A new flag, `-yk-optnone-after-ir-passes`, was added to prevent it from doing so. Even at -O0 the backend was optimising stuff.

 - The user is still required to build with O0. We add our pass on top of "no optimisations". Using higher optimisation levels right now will cause chaos.

 - This shook out a couple fo bugs in the trace compiler -- also fixed in this change.

I hope that in the future we can run with a full -O2 pipeline, just with backend optimisation turned off, but currently the trace compiler makes too many assumptions about the shape of the IR, and optimisations break those assumptions.

This way we can turn on individual optimisations and work our way up incrementally.


vext01:
Question: should we run tests with and without our curated set of optimisations enabled? Currently the optimisations are hard-coded on in `yk-config`.

ltratt:
> Question: should we run tests with and without our curated set of optimisations enabled?

Good idea -- ideally, yes! But we can do that in the next PR, rather than weigh this one down (unless it's easy to do).

vext01:
@ptersilie [103b3e6](https://github.com/ykjit/yk/pull/779/commits/103b3e6766ab4a6cf72b697b2017fca86bf830ad) adds `ctpop` to the whitelist, and i've reverted the llvm change from https://github.com/ykjit/ykllvm/pull/74.

OK to squash (both)?

ltratt:
If @ptersilie is OK with that, so am I.

ptersilie:
Looks good.

vext01:
FWIW, this didn't make a great deal of difference to performance according to ykbench:

# master branch

```
Benchmark 1: /home/vext01/research/yklua/src/lua ./yklua/fannkuchredux.lua 3
  Time (mean ± σ):      2.597 s ±  0.044 s    [User: 2.502 s, System: 0.066 s]
  Range (min … max):    2.532 s …  2.650 s    10 runs
 
Benchmark 1: /home/vext01/research/yklua/src/lua ./yklua/fasta.lua 10
  Time (mean ± σ):      2.874 s ±  0.034 s    [User: 2.763 s, System: 0.059 s]
  Range (min … max):    2.824 s …  2.948 s    10 runs
```
 
# this branch

```
Benchmark 1: /home/vext01/research/yklua/src/lua ./yklua/fannkuchredux.lua 3
  Time (mean ± σ):      2.563 s ±  0.034 s    [User: 2.465 s, System: 0.068 s]
  Range (min … max):    2.511 s …  2.632 s    10 runs
 
Benchmark 1: /home/vext01/research/yklua/src/lua ./yklua/fasta.lua 10
  Time (mean ± σ):      2.831 s ±  0.049 s    [User: 2.734 s, System: 0.057 s]
  Range (min … max):    2.776 s …  2.950 s    10 runs
```


This was using a release build of yk, but with the `yk_testing` feature enabled so we can do `YKD_SERIALISE_COMPILATION`.

Took me a while to realise that `YKD_SERIALISE_COMPILATION` is guarded like that. Maybe not desirable?

vext01:
Ah, we need to update the submodule, so let's merge ykllvm first?

ptersilie:
Please squash.

vext01:
splat.

(Left the second commit separate, since it's not needed for our test suite, only for yklua)

PR #780, opened 2023-07-13
Use a single shared global AOT module.

ptersilie:
Yk requires access to the AOT module at various times during execution, e.g. tracing, deoptimisation. Since accessing an LLVM Module isn't threadsafe, we used to clone the AOT module for each thread that needed it, which is slow. We now use a global AOT module that is only loaded once, and then shared among all threads. The idea is that if two threads need access to the AOT module at the same time, waiting for the lock to clear is probably faster than cloning the entire AOT module.

This change also has the advantage that pointers to LLVM values (e.g. blocks, instructions) stay the same, making it easier in the future to serialise information into guards for deoptimisation.

Unfortunately, the syntax to to access a `ThreadSafeModule` is a bit awkward, as it uses `LLVMOrcThreadSafeModuleWithModuleDo` which takes a function as input that is then called with the locked module. In the future we may want to consider wrapping a normal `Module` inside a Rust Mutex, which should allow us simplify this code back to something similar to what we had before.


ltratt:
@ptersilie What is the performance impact of this (rough figures are probably OK)?

ykrt/src/deopt.rs
```diff
@@ -143,37 +148,35 @@ extern "C" fn __llvm_deoptimize(
...
-    print_jit_state("deoptimise");
-
-    // Put the CompiledTrace back into an Arc, so it is dropped properly.
-    let ctr = Arc::from_raw(ctr);
+    nfi: Option<NewFramesInfo>,
+}
 
-    // FIXME: Check here if we have a side trace and execute it. Otherwise just increment the guard
-    // failure counter.
+// Function containing the actual code setting up the frame reconstructor. This function will be
```
ltratt, line 163: I think we can assume this is a function :) I guess "Reconstruct the stack frames for deopt" or similar is a clearer first sentence?

ptersilie:
Oh, completely forgot to add the performance results I got from ykbench. Appears to be a sweet 10% speedup.

Before

```
Benchmark 1: /home/lukasd/research/yklua/src/lua ./yklua/fannkuchredux.lua 3
  Time (mean ± σ):      4.667 s ±  0.017 s    [User: 4.555 s, System: 0.085 s]
  Range (min … max):    4.638 s …  4.688 s    10 runs
 
Benchmark 1: /home/lukasd/research/yklua/src/lua ./yklua/fasta.lua 10
  Time (mean ± σ):      5.492 s ±  0.029 s    [User: 5.377 s, System: 0.073 s]
  Range (min … max):    5.456 s …  5.541 s    10 runs
```

After

```
Benchmark 1: /home/lukasd/research/yklua/src/lua ./yklua/fannkuchredux.lua 3
  Time (mean ± σ):      4.238 s ±  0.012 s    [User: 4.141 s, System: 0.071 s]
  Range (min … max):    4.223 s …  4.259 s    10 runs
 
Benchmark 1: /home/lukasd/research/yklua/src/lua ./yklua/fasta.lua 10
  Time (mean ± σ):      5.061 s ±  0.020 s    [User: 4.957 s, System: 0.063 s]
  Range (min … max):    5.035 s …  5.087 s    10 runs
```

ptersilie:
I'm afraid these are with serialised compilation. Sorry, to get you excited. :wink:

ltratt:
Ah, then less than I expected, but still better than nothing.

ptersilie:
I think it's with yk compiled in debug mode though. Want me to take some measurements in release mode?

ltratt:
> Want me to take some measurements in release mode?

Good idea!

ptersilie:
Here are the release mode results: Looks like it went down to 5% speedup.

Before

```
Benchmark 1: /home/lukasd/research/yklua/src/lua ./yklua/fannkuchredux.lua 3
  Time (mean ± σ):      4.054 s ±  0.011 s    [User: 3.962 s, System: 0.072 s]
  Range (min … max):    4.039 s …  4.076 s    10 runs
 
Benchmark 1: /home/lukasd/research/yklua/src/lua ./yklua/fasta.lua 10
  Time (mean ± σ):      4.705 s ±  0.019 s    [User: 4.602 s, System: 0.066 s]
  Range (min … max):    4.676 s …  4.734 s    10 runs
```

After

```
Benchmark 1: /home/lukasd/research/yklua/src/lua ./yklua/fannkuchredux.lua 3
  Time (mean ± σ):      3.868 s ±  0.018 s    [User: 3.778 s, System: 0.068 s]
  Range (min … max):    3.847 s …  3.912 s    10 runs
 
Benchmark 1: /home/lukasd/research/yklua/src/lua ./yklua/fasta.lua 10
  Time (mean ± σ):      4.534 s ±  0.017 s    [User: 4.436 s, System: 0.062 s]
  Range (min … max):    4.510 s …  4.553 s    10 runs
```

ltratt:
Interesting -- thanks!

Just the one minor docstring comment.

ykrt/src/deopt.rs
ptersilie: Fixed in [cc9ce02](https://github.com/ykjit/yk/pull/780/commits/cc9ce02bb6df86ce89ffa838fd2996aef13b5029)

ptersilie:
Fixed the doc string. Any opinion on making our own `ThreadSafeModule` in Rust to avoid having to use the awkward `LLVMOrcThreadSafeModuleWithModuleDo`?

ykrt/src/deopt.rs
```diff
@@ -160,9 +160,8 @@ struct ReconstructInfo<'a> {
     nfi: Option<NewFramesInfo>,
 }
 
-// Function containing the actual code setting up the frame reconstructor. This function will be
-// called by `LLVMOrcThreadSafeModuleWithModuleDo` passing relevant info using the
-// `ReconstructInfo` struct.
+/// Actual code for reconstructing the stack. Called from `__ykrt_deopt` via
```
ltratt, line 163: "Actual code" feels redundant. Can we make the first sentence more action orientated? So something like "Reconstruct the stack." I think the "Called from" bit makes clear enough that there are intermediary functions.

ltratt:
> Any opinion on making our own ThreadSafeModule in Rust to avoid having to use the awkward LLVMOrcThreadSafeModuleWithModuleDo?

I don't know enough to have a very good opinion. Would changing help us much? I do wonder if LLVM's thread safe doo-dah does things we don't know about...

ptersilie:
> I don't know enough to have a very good opinion. Would changing help us much? I do wonder if LLVM's thread safe doo-dah does things we don't know about...

From what I can see, it doesn't do anything special. It holds a mutex and the module. When you call `WithModuleDo`, it locks the mutex, then passes the module into the function you give it, and when it finished running, unlocks the mutex again. I believe this is designed this way because the mutex doesn't contain the actual data (like in Rust).

We could easily replicated this with a Rust mutex, and then could do something like
```
let module = threadsafemodule.lock();
// do frame reconstruction
threadsafemodule.unlock();
```
without having to do this callback doodah and passing info via a mutable opaque pointer. I'm happy to discuss this with @vext01 when he's back and, if we decide to, do the change in a separate PR.

ptersilie:
Ok to squash?

ykrt/src/deopt.rs
ltratt: @ptersilie Ping.

ptersilie: Try this: [6329281](https://github.com/ykjit/yk/pull/780/commits/632928150a87d44d7f0c8d106e10b4d50a416d63)

ltratt:
Please squash.

ptersilie:
Squashed.

PR #781, opened 2023-07-13
Always build ykllvm when dirty.

ltratt:
If ykbuild detects changed/untracked files in the bundled ykllvm, it automatically triggers a rebuild. This makes quickly fiddling with the bundled ykllvm easier because you don't have to remember to manually force a ykllvm rebuild.

This commit does not affect those using "external" ykllvm's specified with `YKB_YKLLVM_BIN_DIR`.

### DIFF
--- a/ykbuild/build.rs
+++ b/ykbuild/build.rs
@@ -1,15 +1,18 @@
 #![feature(exit_status_error)]
 
 use fs4::FileExt;
+use rerun_except::rerun_except;
 use std::{
     env,
-    fs::{canonicalize, create_dir_all, read_to_string, write, File},
+    fs::{canonicalize, create_dir_all, File},
     path::Path,
     process::Command,
 };
 use which::which;
 
-/// Where is ykllvm's source code relative to this crate?
+/// The path we use to determine which ykllvm files will cause this build.rs file to be rerun.
+const YKLLVM_SRC_DEPEND_PATH: &str = "../ykllvm";
+/// The path we use to determine that the ykllvm submodule has been cloned.
 const YKLLVM_SUBMODULE_PATH: &str = "../ykllvm/llvm";
 
 fn main() {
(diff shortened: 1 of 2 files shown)


ykbuild/build.rs
```diff
@@ -50,15 +40,44 @@ fn main() {
     ykllvm_dir.push("ykllvm");
     create_dir_all(&ykllvm_dir).unwrap();
 
+    // To avoid running cmake config/build/install, we only do anything with cmake if any of the
+    // following are true:
+    //   * ykllvm isn't built.
+    //   * The ykllvm submodule has changed since we last built it.
+    //   * The ykllvm directory is dirty (i.e. there are new/changed files).
+    //   * The ykllvm directory was changed from dirty to clean since we last built it.
```
ptersilie, line 48: Why do we run `cmake` when the directory was changed from dirty to clean. Doesn't clean mean there's nothing to do?

ltratt: "Clean" here is the git concept of "clean" i.e. `git status` shows no changes. But if there were previously changes and now there aren't, it means that we need to rebuild to make sure that we don't have a build with the old (dirty) stuff in it.

ltratt:
OK, with https://github.com/ykjit/yk/pull/781/commits/792f67564146498984a14e69c607de7493489fb3 this finally works how I want/expect. It does mean that builds using the inbuilt ykllvm are about 1s slower, because we force cargo to scan the ykllvm directory (recursively) for changes. But it has the advantage that it's now pretty hard to get an incorrect yk build if you use the inbuilt ykllvm, even if you're making changes to ykllvm.

Note that there is no speed penalty if (as some power users like to do) you use a separate ykllvm build and specify `YKB_YKLLVM_BIN_DIR`.

@ptersilie @vext01 I think this is ready for (re)review.

ltratt:
bors try

ltratt:
The new commit (c91dc3bf8) doesn't rerun `cmake <configure>` unless it's a totally fresh build. That gives the following cost model for using the inbuilt ykllvm:


* you haven't changed ykllvm at all: 1s
* you've changed ykllvm but not cmake stuff: 15s
* you've changed ykllvm and cmake stuff: 30s

ltratt:
> cmake && cmake --build build && cmake --build build --target install

That is more-or-less what we run with a non-fresh build. [It's more like `cmake --build . && cmake --install .` but same difference.]

vext01:
Yes, but why not run it with a non-fresh build and let cmake track the changes? Then you can remove all this complexity about generating hashes and so on?

I assume something is too slow still?

vext01:
Gotcha. That's a shame.

ltratt:
OK to squash?

ltratt:
@vext01 You were right! https://github.com/ykjit/yk/pull/781/commits/2a8cd29079c5f4eb6b966676f752920875222202 is the result

vext01:
> You were right!

Doesn't happen often!

This looks much simpler :)

If @ptersilie is OK, I am.

ptersilie:
Looks good to me.

ltratt:
Squashed.

ltratt:
@ptersilie We think we raced with something: can we r+ again?

PR #785, opened 2023-07-17
Fix a memory error in __ykrt_reconstruct_frames().

vext01:
Before the code attempted to restore the pointer to the malloced block from a rdi, a register that has been trashed by a call to memcpy().

Fix by using a callee-save register (r12) instead.

This was shaken out by trying the JIT on Debian Bookworm.

No idea how it worked on the older Debian. Maybe a libc change.


ptersilie:
Please squash.

vext01:
splat.

PR #787, opened 2023-07-18
Rerun the build script, and possibly even configure, if env vars changes

ltratt:
`cargo` automatically rebuilds Rust projects if relevant environment variables change.

This PR brings that same idea to ykbuild: if you change `CC`/`CFLAGS`/`YKB_YKLLVM_BUILD_ARGS` (etc etc), ykbuild automatically reruns `cmake <configure>` as well as `cmake`. The aim is to make editing the inbuilt ykllvm easy to do without making mistakes. In order to do this, we have to cache "key=value" pairs in
`target/<debug|release>/yk_env_vars`.

AFAICS, this largely removes the need for `YKB_YKLLVM_BIN_DIR` while developing, though using that is still supported as before.

### DIFF
--- a/ykbuild/build.rs
+++ b/ykbuild/build.rs
@@ -3,19 +3,38 @@
 use fs4::FileExt;
 use rerun_except::rerun_except;
 use std::{
+    collections::HashMap,
     env,
-    fs::{canonicalize, create_dir_all, File},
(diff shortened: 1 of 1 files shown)


ltratt:
bors try

PR #790, opened 2023-07-19
Store LLVM values as pointers in guard failures.

ptersilie:
Now that the AOT module is shared between all threads, pointers to LLVM values remain the same in each in each thread. This allows us to store simple pointers within guard failures, instead of having to encode instructions via their basic block and instruction indexes. Mercifully, this change also gets rid of a lot of code related to encoding and decoding those instructions, most notably the awkward AOTMap.

### DIFF
--- a/ykrt/src/frame/llvmbridge.rs
+++ b/ykrt/src/frame/llvmbridge.rs
@@ -23,52 +23,8 @@ impl Module {
         Self(module)
     }
 
-    pub fn function(&self, name: *const i8) -> Function {
-        let func = unsafe { LLVMGetNamedFunction(self.0, name) };
-        debug_assert!(!func.is_null());
-        unsafe { Function::new(func) }
-    }
-
     pub fn datalayout(&self) -> LLVMTargetDataRef {
         unsafe { LLVMGetModuleDataLayout(self.0) }
-    }
-}
-
-pub struct Function(LLVMValueRef);
-
-impl Function {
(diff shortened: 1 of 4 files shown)


ykrt/src/frame/llvmbridge.rs
```diff
@@ -23,55 +23,11 @@ impl Module {
...
-        debug_assert!(!func.is_null());
-        unsafe { Function::new(func) }
-    }
-
     pub fn datalayout(&self) -> LLVMTargetDataRef {
         unsafe { LLVMGetModuleDataLayout(self.0) }
     }
 }
 
-pub struct Function(LLVMValueRef);
```
vext01, line 37: I didn't expect this to go. Just not used?

ptersilie:
I did a super rough benchmark, which gave me around 2% speedup. Could be coincidental but could be attributed to accessing instructions in `FrameReconstructor` now being slightly less work, and guardfailure blocks now being slightly smaller.

ykrt/src/frame/llvmbridge.rs
ptersilie: This was only used to get access to the live AOT instructions. Since we now pass pointers, this is no longer needed.

vext01:
This is a serious contender for "PR of the year". Needless to say. LGTM.

Not merging just yet, as I'm upgrading the CI server.

ltratt:
bors try

PR #792, opened 2023-07-20
Rework promote

vext01:
Simplify promote quite a bit. See commits.

@ltratt please scrutinise the second. I wasn't sure if we should even macro-mangle it at all now...

Requires a compiler change. Coming soon.


yktracec/src/jitmodbuilder.cc
```diff
@@ -1744,9 +1722,17 @@ class JITModBuilder {
...
-    PromoteMap[CI] = JITPromoteVar;
+    // You might think that we need to do: `VMap[CI] = PConstLL` so as to use
+    // the constant in place of the promoted value from here onward.
+    //
+    // As it happens, the guard we just inserted already gives LLVM everything
+    // it needs to promote the value for us.
+    //
+    // If we trace a `yk_promote(x)` when `x==5`, then the trace compiler
+    // makes a guard that says "if x!=5 then goto the guard failure block,
+    // otherwise goto the guard success block". LLVM can then easily infer
```
vext01, line 1733: In hindsight, I wonder if we should replace "easily" with "often (with sufficient AOT and JIT LLVM passes)", because we only noticed this behaviour when turning on the AOT `instcombine` pass.

ltratt: I would probably say "LLVM is then more likely to be able to infer that...".

vext01: how about "llvm is likely to be able to...".

Wasn't sure about "more likely" as without the guard it would be incorrect for LLVM to make that deduction at all.

ltratt: *Technically* nothing stops LLVM putting in a speculation (i.e. compiling two code paths). At some point, though, we're in angels-on-a-pin territory!

vext01: What I mean is that without the guard, nothing is known about the promoted value, so it would be miscompilation for llvm to assume that it is any constant value.  So "more likely" makes no sense, as without the guard it should be "impossible" for it to make any deduction of its value.

vext01: That's pretty convoluted and stating that doesn't help this explanation.

I could walk backwards everywhere I go, but I don't :P

ltratt: Yes, but there is a mildly interesting point: "miscompilation" is a more subtle concept than it first appears.

All of this is why I tend to prefer something like "more likely to be able to infer that" as it covers a multitude of possibilities.

vext01: I find "more likely" confusing -- is there something less confusing that we can agree on?

ltratt: I'm unsure what could be less confusing; or, at least, "less confusing" might end up being "less accurate"? Ultimately, this is a relative concept: LLVM can *already* infer this property if it so chooses. Promotion just makes it (much) more likely that it infers this property, but it's still not a guarantee that it does so.

ptersilie:
I feel vindicated by this PR as this is the design I originally envisioned.

ltratt:
@ptersilie Traditionally people just say "I told you so" :p

yktracec/src/jitmodbuilder.cc
```diff
@@ -1730,8 +1730,8 @@ class JITModBuilder {
     //
     // If we trace a `yk_promote(x)` when `x==5`, then the trace compiler
     // makes a guard that says "if x!=5 then goto the guard failure block,
-    // otherwise goto the guard success block". LLVM can then easily infer
-    // that `x==5` in the guard success block and it will propagate the
+    // otherwise goto the guard success block". It is then likely that LLVM
+    // will infer that `x==5` in the guard success block propagate the
```
vext01, line 1734: "and propagate"... sigh

yktracec/src/jitmodbuilder.cc
ltratt: OK.

yktracec/src/jitmodbuilder.cc
vext01: Fixed in 3c1b3a1

ltratt:
Please squash.

vext01:
anything to add @ptersilie ?

vext01:
(I will need to update the submodule once the ykllvm change is in)

ltratt:
You can push a commit referencing the last commit in your branch now if you want.

ptersilie:
> anything to add @ptersilie ?

I'm good.

vext01:
About the second commit. We introduced that macro mangling dance because the function was special in that ykllvm recognises it and does magic. We no longer do that. So should we still have the macro stuff? If so, why is this the only API function with that dance?

ltratt:
> Would that be correct? I'll be deleting that branch, won't that break the build?

That commit hash will end up being in the main branch, so it's OK (it's the parent commit of the bors merge commit).

Please squash.

ltratt:
> So should we still have the macro stuff? If so, why is this the only API function with that dance?

Sorry, I missed this! You're right, the macro rewriting is pointless, since nothing depends on the *name* of the function.

ltratt:
> You're right, the macro rewriting is pointless, since nothing depends on the name of the function.

I'm wrong. The name *is* important, so I still prefer the macro renaming, because it makes it harder for people to shoot themselves in the foot.

vext01:
OK, so that leads us to the second sub-question:

>  So should we still have the macro stuff? If so, why is this the only API function with that dance?

Why is `yk_promote()` special, and not (say) `yk_mt_new()`? Should that also do the macro dance?

ltratt:
> Why is yk_promote() special, and not (say) yk_mt_new()? 

`yk_promote` is *recognised and altered by the compiler*, making it completely different to most other functions. [I would personally prefer it to have a much more mangled name than just a `__` prefix, but I didn't want to be too much of a pain.]

vext01:
> yk_promote is recognised and altered by the compiler

`yk_promote` is not longer recognised by the compiler. At least not the AOT compiler.

That's what the last ykllvm PR was about: https://github.com/ykjit/ykllvm/pull/77

ltratt:
It is recognised by the JIT compiler (I should have been clearer) by name, which is why I strongly believe it should be mangled.

vext01:
Fair. That's all I wanted to confirm.

OK to squash then?

ltratt:
Please squash.

vext01:
splat.

PR #796, opened 2023-07-24
Turn on some more AOT optimisations.

vext01:
This is *roughly* a subset of LLVM's default LTO optimisation level 3, with some bits omitted.

In doing so I've identified a handful of optimisations that break yk. We should be able to chip away at these one-by-one in follow up PRs.

I don't have much to use for benchmarks. Of the 3 ykbench benchmarks, only one is in a working state right now (broken before my changes it seems):

### before
```
$ YKD_SERIALISE_COMPILATION=1 ./yklua/bench.sh /home/vext01/research/yklua/src/lua          
Benchmark 1: /home/vext01/research/yklua/src/lua ./yklua/fannkuchredux.lua 3                                                    
  Time (mean ± σ):      2.676 s ±  0.019 s    [User: 2.574 s, System: 0.078 s]                                                  
  Range (min … max):    2.654 s …  2.705 s    10 runs                                                                           
```

### after
```                                                                                                                                
$ YKD_SERIALISE_COMPILATION=1 ./yklua/bench.sh /home/vext01/research/yklua/src/lua           
Benchmark 1: /home/vext01/research/yklua/src/lua ./yklua/fannkuchredux.lua 3                                                    
  Time (mean ± σ):      2.497 s ±  0.019 s    [User: 2.401 s, System: 0.072 s]                                                  
  Range (min … max):    2.472 s …  2.528 s    10 runs                                                                           
```

This change also breaks the `tpack.lua` test and possibly more... None of these tests appear in CI or nightlies.

What do you think we should do? Merge this now, or wait until I can improve CI/nightlies?


ltratt:
> This change also breaks the tpack.lua test and possibly more

I think that's a problem: I don't think we should be breaking any tests with this PR that passed before it. Presumably that means we have to dial down the optimisations a bit?

vext01:
I can try finding the pass that broke that one test, but it's not clear to me which tests (of the whole suite) we should expect to run at all.

I think we could do with someone (me?) auditing all those tests and adding the ones that work to CI. Then we should aim to never break one that already works.

ltratt:
At the moment yk's test suite + yklua's test suite is our least worst oracle. I don't entirely know if all yklua tests pass, but we shouldn't turn on any optimisations that make yklua tests fail.

vext01:
> I don't entirely know if all yklua tests pass

Prior to my change, many fail.

ltratt:
Does this PR make more fail?

vext01:
Yep. At least one more.

Yep, that's bad. But what I'm saying is: to know what else I broken I need to audit them all, and while I'm there, I may as well add the passing tests to CI+nightlies. Then we can come back to this PR. Plan?

ltratt:
> But what I'm saying is: to know what else I broken I need to audit them all, and while I'm there, I may as well add the passing tests to CI+nightlies. Then we can come back to this PR. Plan?

Good idea!

vext01:
OK, so let's put this on pause.

ptersilie:
I also prefer we don't break `tpack.lua` as that is my goto test to check that things still work, especially when I fiddle around with deopt.

ltratt:
@vext01 A thought for this PR: if an interpreter uses `-fno-strict-aliasing` or similar, will we end up turning on optimisations that override that choice? If we do, that might explain some of the breakage.

ltratt:
@vext01 I've been chewing on this PR and have a couple of additional thoughts, which may or may not be good ones. See what you think!

1. In retrospect, I suspect we're lacking sufficient test cases in the sense that our only oracle for the correctness of the additional optimisations are a tiny handful of yklua tests. At a minimum we should probably try more Lua programs. Perhaps we should try getting ykcpython at least minimally running?
2. We should probably measure the effect on the pure interpreter (i.e. without it doing any JITing) to understand if it's having the effect we hope. In particular, I wonder if these optimisations are as effective with the shadow stack as we might like? [A cheaty way to turn the JIT off is to turn the hot threshold to a ludicrously high value.]

bin/yk-config
```diff
@@ -43,20 +43,30 @@ AOTPASSES="${AOTPASSES},rpo-function-attrs,globalsplit,wholeprogramdevirt"
...
-AOTPASSES="${AOTPASSES},argpromotion,deadargelim,instcombine,function-attrs"
-AOTPASSES="${AOTPASSES},memcpyopt,dse,mldst-motion,loop-flatten,indvars"
+#
+# FIXME: Also note that `inline` probably isn't doing anything since we require
+# the interpreter to be built -O0, menaning clang will slap `noinline`
+# everywhere.
+AOTPASSES="${AOTPASSES},aggressive-instcombine,inline"
+AOTPASSES="${AOTPASSES},globalopt,globaldce,argpromotion,deadargelim"
+#AOTPASSES="${AOTPASSES},instcombine" -- this second invocation breaks yklua
+#AOTPASSES="${AOTPASSES},jumpthreading,sroa,tailcallalim" -- break Yk test suite.
```
ltratt, line 60: I think `yk` is lower case?

vext01: it can be :) Will fix.

ltratt: I don't see two new commits?

vext01: refresh?

image

ltratt:
Please squash.

PR #799, opened 2023-07-27
Basic support for outputting JIT stats.

ltratt:
Setting the environment variable `YKD_JITSTATS` to a file path or `-` (which outputs to stderr) outputs JSON with stats about the yk compilation process. For example:

```
$ YKD_SERIALISE_COMPILATION=1 YKD_JITSTATS=- src/lua tests/db.lua
testing debug library and debug information
+
testing inspection of parameters/returned values
+
+
testing traceback sizes
testing debug functions on chunk without debug info
OK
{
    "traces_recorded_ok": 11,
    "traces_recorded_err": 0,
    "traces_compiled_ok": 10,
    "traces_compiled_err": 1
}
```

This commit is very much a "first stab to see if this looks like the sort of thing we want" rather than "here is a complete thing". That doesn't mean it's not useful already, but it clearly only captures a tiny amount of the information we might like to see captured.

### DIFF
--- a/docs/src/dev/debugging.md
+++ b/docs/src/dev/debugging.md
@@ -1,5 +1,16 @@
 # Debugging.
 
+## JIT statistics
+
+At the end of an interpreter run, yk can print out some simple statistics about
+what happened during execution. If the `YKD_JITSTATS=<path>` environment
+variable is defined, then JSON statistics will be written to the file at
+`<path>` once the interpreter "drops" the `YkMt` instance. `-` (i.e. a single
+dash) can be used in place of path, in which case the statistics will be
+written to `stderr`. Note that if the interpreter starts multiple yk instances,
+then statistics will be written (and, probably, overwritten) to `<file>`.
+
+
 ## Debugging JITted code.
 
 Often you will find the need to inspect JITted code with a debugger. If the
--- a/tests/c/jitstats1.c
+++ b/tests/c/jitstats1.c
@@ -1,0 +1,35 @@
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_JITSTATS=/dev/stderr
+//   stderr:
+//     {
+//       "traces_collected_ok": 1,
+//       "traces_collected_err": 0,
+//       "traces_compiled_ok": 1,
+//       "traces_compiled_err": 0
+//     }
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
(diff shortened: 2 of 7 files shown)


vext01:
This looks good!

I liked `-`: it's neat and acknowledges OSs where stderr isn't a file.

If you wanted to bring `-` back, I'd change `output_path: String` to a file (or file descriptor?), then you can write to either in a unified fashion.

And if I were to bikeshed: `s/recorded/collected/g` for parity with terminology elsewhere.

vext01:
Should it be `YKD_PRINT_JITSTATS` to be like the other `YKD_PRINT_*` vars?

ltratt:
Note that this PR *doesn't* print things! It writes to a file. [This is, IMHO, a more general model than "printing" to stdout/stderr.]

ptersilie:
Could name it `YKD_WRITE_JISTSTATS` to distringuish between printing and writing to file? But I'm not bothered either way.

ptersilie:
Looks good otherwise.

ltratt:
> And if I were to bikeshed: s/recorded/collected/g for parity with terminology elsewhere.

Good point. Fixed in 995620e0daf0704cefaac822738d522dae850426.

[I tend to think "recorded" is clearer than "collected", so maybe one day we could do a wholesale change. But not in this PR!]

ptersilie:
> Could name it `YKD_WRITE_JISTSTATS` to distringuish between printing and writing to file? But I'm not bothered either way.

Any comment on this suggestion?

ltratt:
`WRITE` feels superfluous to me, I must admit. [We put `PRINT` in because, from memory, we wanted to make the setting as obnoxious as possible so that normal users don't think it's something useful for them -- it's really an internal debugging aid for us. In contrast, `YKD_JITSTATS` is useful for normal interpreter authors.]

ptersilie:
Fine. Waiting for @vext01 to approve the other changes, then feel free to squash.

ltratt:
> acknowledges OSs where stderr isn't a file.

When isn't stderr a file?

vext01:
Does windows even have device nodes?

ltratt:
@vext01 Point taken. Reverted in c09f9943a5424fe0c7d3fabd35e4e1609d08c007. Also documented in 3891b3b4956a47d89745451f4ed82c6b8516ccfb.

vext01:
LGTM.

Please squash.

ltratt:
Squashed.

PR #800, opened 2023-07-30
Record basic timing information in `JITStats`.

ltratt:
This records the time spent in various VM states (compiling, deopting, interpreting, JIT executing) and shows it in `JITStats` output (in seconds). The "interpreting" state is an over-approximation, because we don't know what the interpreter is doing: if a thread puts itself to sleep, we will count that as execution time. These times are thus "best efforts" rather than "guaranteed to be perfect". They do, however, already tell a very interesting story:

```
$ YKD_SERIALISE_COMPILATION=1 YKD_JITSTATS=- src/lua tests/db.lua
testing debug library and debug information
+
testing inspection of parameters/returned values
+
+
testing traceback sizes
testing debug functions on chunk without debug info
OK
{
    "traces_collected_ok": 11,
    "traces_collected_err": 0,
    "traces_compiled_ok": 10,
    "traces_compiled_err": 1,
    "duration_compiling": 9.9087,
    "duration_deopting": 2.2660,
    "duration_interpreting": 0.144,
    "duration_jit_executing": 0.2
}
```

PR #801, opened 2023-07-31
Cache the stackmap of a compiled trace.

ptersilie:
Reading out the entire trace's stackmap each time a guard fails is slow, especially, since we only need a small part of that stackmap. We could either modify the stackmap parser to only read the part that is relevant for the current guard failure, or we read the entire stackmap immediately after compiling the trace and store it alongside it inside `CompiledTrace`. The latter has the advantage of moving the stackmap parsing from deopt to inside the compilation thread.

Storing the result of the parsed stackmap also means we no longer need to keep the original stackmap data around which can now be freed, squashing another memory leak that was on our todo list for a while.

### DIFF
--- a/ykrt/src/deopt.rs
+++ b/ykrt/src/deopt.rs
@@ -192,12 +192,8 @@ extern "C" fn ts_reconstruct(ctx: *mut c_void, module: LLVMModuleRef) -> LLVMErr
 
     let mut framerec = unsafe { FrameReconstructor::new(activeframes, module) };
 
-    // Parse the stackmap of the JIT module.
-    // OPT: Parsing the stackmap and initialising `framerec` is slow and could be heavily reduced
-    // by caching the result.
-    let slice = unsafe { slice::from_raw_parts(ctr.smptr as *mut u8, ctr.smsize) };
-    let map = StackMapParser::parse(slice).unwrap();
-    let live_vars = map.get(&retaddr.try_into().unwrap()).unwrap();
+    // Retrieve the live variables for this guard from this trace's stackmap.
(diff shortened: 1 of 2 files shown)


ptersilie:
@ltratt I couldn't measure any performance difference using `tpack.lua` but maybe you can have a look if this changes the things you're seeing in your #800 PR?

ltratt:
For `db.lua` before:

```
{
    "duration_compiling": 5.5279,
    "duration_deopting": 2.2666,
    "duration_jit_executing": 0.2,
    "duration_outside_yk": 0.144,
    "duration_trace_mapping": 3.3760,
    "traces_collected_err": 0,
    "traces_collected_ok": 11,
    "traces_compiled_err": 1,
    "traces_compiled_ok": 10
}
```

After:

```
{
    "duration_compiling": 5.5225,
    "duration_deopting": 0.354,
    "duration_jit_executing": 0.2,
    "duration_outside_yk": 0.140,
    "duration_trace_mapping": 3.3757,
    "traces_collected_err": 0,
    "traces_collected_ok": 11,
    "traces_compiled_err": 1,
    "traces_compiled_ok": 10
}
```

So about 6.5x better!

ptersilie:
Nice, there's probably some more performance we can tickle out of `deopt`, if we look closer, but at least it's not as egregious as it was before.

vext01:
Yes, yes, yes!

vext01:
Is this one ready to go?

ltratt:
I'm a complete idiot and commented on the wrong PR!

ltratt:
Please squash.

PR #804, opened 2023-08-02
Better document the stats module.

ltratt:


ykrt/src/jitstats.rs
```diff
@@ -17,7 +14,14 @@ use std::{
 };
 use strum::{Display, EnumCount, EnumIter, IntoEnumIterator};
 
+/// Record yk statistics if enabled. In non-testing mode this is only enabled if the end user
+/// defines the global variable `YKD_JITSTATS`. In testing mode this is always enabled, with output
```
vext01, line 18: s/global/environment/ ?

ykrt/src/jitstats.rs
```diff
@@ -17,7 +14,14 @@ use std::{
 };
 use strum::{Display, EnumCount, EnumIter, IntoEnumIterator};
 
+/// Record yk statistics if enabled. In non-testing mode this is only enabled if the end user
```
vext01, line 17: comma after `mode`?

ykrt/src/jitstats.rs
ltratt: Fixed in e83ddd3.

ykrt/src/jitstats.rs
ltratt: I think it's fine with or without?

ykrt/src/jitstats.rs
```diff
@@ -17,7 +14,14 @@ use std::{
...
+/// defines the global variable `YKD_JITSTATS`. In testing mode this is always enabled, with output
+/// being sent to `stderr`.
```
vext01: What's the justification for "always enabled by default in testing?". I'd expect it off by default like all the other things.

ltratt: Hmm, that's a good question. You're right, it probably shouldn't be always on.

If it's OK, I'll fix that in the next PR.

ykrt/src/jitstats.rs
vext01: Up to you, but I was jarred on first read.

vext01: thanks

ykrt/src/jitstats.rs
vext01: Please do :)

vext01:
Please squash

ltratt:
Squashed.

ykrt/src/jitstats.rs
vext01: In the next PR, can it also be a `0/1`-valued variable for consistency?

ltratt: That wouldn't make sense -- you have to give a filename (or `-` for stderr) as an argument.

vext01: ah yeah, my bad.

Issue #612, opened 2023-02-28
Properly handle failed trace compilation.


ptersilie:
Currently, when a trace fails to compile, we just ignore it and move on. Soon, we want to handle this better, and depending on why the compilation failed, never try this location again, or try it again later.

ltratt:
We haven't fully fixed this, but between https://github.com/ykjit/yk/pull/806 and https://github.com/ykjit/yk/pull/811 I would say we are now sufficiently close that this issue is no longer relevant.

PR #806, opened 2023-08-02
Mark locations that cause "too long" traces as `DontTrace`.

ltratt:
Once traces get beyond a certain size, the time for mapping and compilation becomes ever more punishing. For example, with yklua and db.lua, one trace in particular takes 6 seconds to map and compile because the PT trace generates 250,000 blocks!

This commit very crudely implements "abort compilation when a trace is too long". After a very quick experiment, 5000 blocks seemed a reasonable threshold, in the sense that a threshold of 10,000 blocks can end up with LLVM compilation becoming very slower. I don't know if compilation slows down linearly or not as the number of blocks increases, but the small numbers I have suggest it is superlinear. It looks like the mapping phase but that's misleading because "mapping" includes "mapping" and "IR generation" -- the latter seems to be the main slowdown.

Not only does this commit stop compiling a trace if it's too long, but it marks the `HotLocation` as `DontTrace` so there won't be an attempt to make a second trace. We should probably do something more RPython-esque such as "wait for a location to fail 5 times before marking it as DontTrace", but given how slow mapping and compilation are, we'll probably end up having to have a lower threshold.

The effect on db.lua and tpack.lua is noticable. Before this commit, db.lua:

```
"duration_compiling": 4.978,
"duration_deopting": 0.314,
"duration_jit_executing": 0.002,
"duration_outside_yk": 0.137,
"duration_trace_mapping": 3.709,
"trace_executions": 771,
"traces_collected_err": 0,
"traces_collected_ok": 11,
"traces_compiled_err": 1,
"traces_compiled_ok": 10
YKD_SERIALISE_COMPILATION=1 YKD_STATS=- src/lua tests/db.lua  9.24s user 0.14s system 98% cpu 9.475 total
```

After:

```
"duration_compiling": 4.955,
"duration_deopting": 0.313,
"duration_jit_executing": 0.001,
"duration_outside_yk": 0.145,
"duration_trace_mapping": 0.766,
"trace_executions": 771,
"traces_collected_err": 0,
"traces_collected_ok": 11,
"traces_compiled_err": 1,
"traces_compiled_ok": 10
YKD_SERIALISE_COMPILATION=1 YKD_STATS=- src/lua tests/db.lua  6.29s user 0.13s system 98% cpu 6.534 total
```

And tpack.lua before:

```
"duration_compiling": 13.403,
"duration_deopting": 0.055,
"duration_jit_executing": 0.000,
"duration_outside_yk": 0.087,
"duration_trace_mapping": 2.841,
"trace_executions": 5,
"traces_collected_err": 0,
"traces_collected_ok": 8,
"traces_compiled_err": 2,
"traces_compiled_ok": 6
YKD_SERIALISE_COMPILATION=1 YKD_STATS=- src/lua tests/tpack.lua  16.32s user 0.18s system 99% cpu 16.615 total
```

And tpack.lua after:

```
"duration_compiling": 13.345,
"duration_deopting": 0.055,
"duration_jit_executing": 0.000,
"duration_outside_yk": 0.080,
"duration_trace_mapping": 1.258,
"trace_executions": 5,
"traces_collected_err": 0,
"traces_collected_ok": 8,
"traces_compiled_err": 2,
"traces_compiled_ok": 6
YKD_SERIALISE_COMPILATION=1 YKD_STATS=- src/lua tests/tpack.lua  14.68s user 0.17s system 99% cpu 14.933 total
```

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -43,6 +43,11 @@ type AtomicHotThreshold = AtomicU32;
 pub type TraceFailureThreshold = u16;
 pub type AtomicTraceFailureThreshold = AtomicU16;
 
+/// How many blocks long can a trace be before we give up trying to compile it? Note that the
+/// slower our compiler, the lower this will have to be in order to give the perception of
+/// reasonable performance.
+/// FIXME: needs to be configurable.
+pub(crate) const DEFAULT_TRACE_TOO_LONG: usize = 5000;
 const DEFAULT_HOT_THRESHOLD: HotThreshold = 50;
 const DEFAULT_TRACE_FAILURE_THRESHOLD: TraceFailureThreshold = 5;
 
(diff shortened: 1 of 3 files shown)


ltratt:
Partly fixes https://github.com/ykjit/yk/issues/612.

ltratt:
The `map` function returns `IRTrace`, though I admit that seems to just be a `Vec` of blocks? So I'm not really sure:

```rust
fn map(self: Box<Self>, _tracer: Arc<dyn Tracer>) -> Result<IRTrace, InvalidTraceError>
```

ykrt/src/mt.rs
```diff
@@ -43,6 +43,10 @@ type AtomicHotThreshold = AtomicU32;
 pub type TraceFailureThreshold = u16;
 pub type AtomicTraceFailureThreshold = AtomicU16;
 
+/// How many blocks long can a trace be before we give up trying to compile it? Note that the
+/// slower our compiler, the lower this will have to be in order to give the perception of
+/// reasonable performance.
```
vext01, line 48: `FIXME: make configurable`?

ltratt: Fixed in e6af1d1.

ltratt:
> I think it'd be worth flamegraphing things before your change to see what is so costly about mapping.

I think someone who wrote the mapper might want to do that :) I will also time out before I can do anything.

ykrt/src/trace/hwt/mapper.rs
```diff
@@ -147,7 +147,10 @@ impl<'a> HWTMapper {
     ) -> Result<Vec<IRBlock>, HWTracerError> {
         let mut ret: Vec<IRBlock> = Vec::new();
 
-        for block in &mut trace_iter {
+        for (i, block) in &mut trace_iter.enumerate() {
+            if i > crate::mt::DEFAULT_TRACE_TOO_LONG {
```
vext01, line 151: Hrm. I looked to see if we can avoid doing *any* mapping at all if the trace is too long, but the decoder is lazy and we don't really know without decoding the trace...

You could express the acceptable trace length in "bytes of PT payload", but that wouldn't be portable between trace collectors...

i guess you had similar thoughts?

ltratt: I did. Clearly we'll need to do something better than this at some point (e.g. we could do an analysis of PT buffer lengths and their average number of blocks, and guess based on that). But that's a lot of complexity and if we optimise some of the parts of, for example, the mapper, we'd have to reanalyse everything again. So I thought: a simple, obviously crude, threshold is probably a tolerable starting point. It has the virtue of being easy to adjust!

ykrt/src/mt.rs
```diff
@@ -376,38 +380,43 @@ impl MT {
...
+                        }
+                    };
+                }
+                Err(_) => {
+                    // FIXME: Immediately marking a location as `DontTrace` is too brutal.
+                    // See: https://github.com/ykjit/yk/issues/612
+                    hl_arc.lock().kind = HotLocationKind::DontTrace;
+                    mt.stats.trace_compiled_err();
+                    #[cfg(feature = "yk_jitstate_debug")]
+                    print_jit_state("trace-compilation-aborted");
```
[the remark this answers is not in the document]
ltratt: I'd like to rethink the `print_jit_state` stuff entirely, but that's definitely something for future PRs!

ltratt:
OK to squash or ... ?

vext01:
Please squash.

ltratt:
Squashed.

ltratt:
@vext01 I think this one's just awaiting an r+?

PR #808, opened 2023-08-03
Remove `unsafe impl`s

ltratt:
This PR removes two dodgy `unsafe impl Sync/Send`s. The first (https://github.com/ykjit/yk/commit/9d7bff4faf1a2c8e158c8ccbfd2e8bd998b004e4) is no longer needed. The second (https://github.com/ykjit/yk/commit/1e70fc9fa6c246cf3ce50fe5291d83e4115c2de9) is a bit more work, but arguably more important.

### DIFF
--- a/ykrt/src/trace/mod.rs
+++ b/ykrt/src/trace/mod.rs
@@ -221,10 +221,13 @@ impl IRTrace {
     }
 }
 
-#[derive(Debug)]
+struct SendSyncConstPtr<T>(*const T);
+unsafe impl<T> Send for SendSyncConstPtr<T> {}
+unsafe impl<T> Sync for SendSyncConstPtr<T> {}
+
 struct Guard {
     failed: u32,
-    code: Option<*const c_void>,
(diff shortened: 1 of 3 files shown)


ykrt/src/trace/mod.rs
```diff
@@ -221,10 +221,13 @@ impl IRTrace {
     }
 }
 
-#[derive(Debug)]
+struct SendSyncPtr<T>(T);
```
vext01, line 224: Should we make it:

```
struct SendSyncPtr<T>(*const T);
```

So that we can't accidentally put non-pointers inside?

ltratt: Good idea! Done in 9c5a16b -- notice I've also changed the struct to `SendSyncConstPtr` because this can't be used with `*mut` pointers (if we have them at some point), and we'd need a second struct for that.

vext01:
LGTM. Please squash.

ltratt:
Squashed.

PR #810, opened 2023-08-09
Fix an FFI call.

vext01:
Assigned Jake, as Laurie and Lukas are away and Jake is a C++ aficionado :)

See individual commits.

### DIFF
--- a/yktracec/src/ykllvmwrap.cc
+++ b/yktracec/src/ykllvmwrap.cc
@@ -50,6 +50,8 @@ struct BitcodeSection {
   void *data;
   // The length of the LLVM bitcode section, in bytes.
   size_t len;
+
+  StringRef asStringRef() { return StringRef(static_cast<char *>(data), len); }
 };
 
 // If possible, return a string describing the location of an instruction in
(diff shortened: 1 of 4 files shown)


vext01:
Typo in the first commit message: "There's no much online".

Can fix after review.

yktracec/src/ykllvmwrap.cc
```diff
@@ -201,8 +201,8 @@ void initLLVM(void *Unused) {
 // Load the GlobalAOTMod.
 //
 // This must only be called from getAOTMod() for correct synchronisation.
-void loadAOTMod(struct BitcodeSection &Bitcode) {
-  auto Sf = StringRef((const char *)Bitcode.data, Bitcode.len);
+void loadAOTMod(struct BitcodeSection *Bitcode) {
+  auto Sf = StringRef((const char *)Bitcode->data, Bitcode->len);
```
jacob-hughes, line 205: I would tend to avoid doing C casts like this and use one of the C++ casts (`reinterpret_cast` / `bit_cast` / etc) as it's more explicit.

jacob-hughes:
LGTM other than the minor nit.

yktracec/src/ykllvmwrap.cc
vext01: [40a859c](https://github.com/ykjit/yk/pull/810/commits/40a859c3c4f13498375e4b242c2c6fbeec7842ee) removes the C-style cast.

This commit would stand alone.

vext01:
OK to fix typo in commit message?

jacob-hughes:
Please squash with updated commit.

vext01:
No squashing needed. Fixed commit message.

PR #811, opened 2023-08-15
Consistently record trace failure

ltratt:
This PR moves "increment trace failure" into a function (https://github.com/ykjit/yk/commit/ca7baa10a471647f96455973ebfee1aae7e17626) which means that we can then use this function when trace compilation fails (https://github.com/ykjit/yk/commit/75e9fbb18fb63c462342e9be18f01ecc1b04a526). This is still probably too simplistic, but it's less simplistic than before, and therefore nudges us in a better direction.

### DIFF
--- a/ykrt/src/location.rs
+++ b/ykrt/src/location.rs
@@ -10,7 +10,7 @@ use std::{
 };
 
 use crate::{
-    mt::{HotThreshold, TraceFailureThreshold},
+    mt::{HotThreshold, TraceFailureThreshold, MT},
     trace::CompiledTrace,
 };
 use parking_lot::Mutex;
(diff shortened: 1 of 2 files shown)


ykrt/src/location.rs
```diff
@@ -204,6 +204,21 @@ pub(crate) struct HotLocation {
     pub(crate) trace_failure: TraceFailureThreshold,
 }
 
+impl HotLocation {
+    /// Mark a trace at this `HotLocation` as having failed. The return value indicates whether
```
vext01, line 208: nitpick: "trace *starting* at this location"

ykrt/src/mt.rs
```diff
@@ -70,7 +70,7 @@ pub struct MT {
     /// How many worker threads are currently running. Note that this may temporarily be `>`
     /// [`max_worker_threads`].
     active_worker_threads: AtomicUsize,
-    tracer: Arc<dyn Tracer>,
+    tracer: Arc<Arc<dyn Tracer>>,
```
vext01, line 73: Go on then: why the double `Arc`?

ykrt/src/location.rs
ltratt: Fixed in bf7ff37.

ykrt/src/mt.rs
ltratt: Good question. Will check.

ltratt: It took me a while to remind myself what I was doing here: this was a midway point on a change to allow the `Tracer` to be changed *during* a yk run (at the moment this isn't useful, but one day it might be). But `Arc<Arc>` is pointless in and of itself -- I just forgot to complete things! d4ee063 completes the change.

vext01: You could have said that you were doing something very very clever, and I'd have believed you :)

vext01:
Please squash.

ykrt/src/mt.rs
ltratt: I admire your faith, but it is always more likely that I'm doing something very very stupid :p

ltratt:
Squashed.

PR #812, opened 2023-08-18
Add a `Compiler` trait to allow us to have multiple JIT compilers.

ltratt:
This commit minimally adds a `Compiler` trait which will in the future allow us to swap between different JIT compilers. At the moment this is slightly farcical because `IRTrace` `compile`s itself, and that `compile` method assumes an LLVM backend. This commit isn't aiming to fix that -- though hopefully we'll do so soon! -- but having a trait seems the right first step towards separating out the different concerns.

### DIFF
--- a/ykrt/build.rs
+++ b/ykrt/build.rs
@@ -1,3 +1,5 @@
 pub fn main() {
     ykbuild::apply_llvm_ld_library_path();
+
+    println!("cargo:rustc-cfg=jitc_llvm");
 }
--- a/ykrt/src/compile/jitc_llvm.rs
+++ b/ykrt/src/compile/jitc_llvm.rs
@@ -1,0 +1,24 @@
+//! An LLVM JIT backend. Currently a minimal wrapper around the fact that [IRTrace]s are hardcoded
+//! to be compiled with LLVM.
+
+use crate::{compile::Compiler, trace::IRTrace};
+use libc::c_void;
+use std::{error::Error, sync::Arc};
+use tempfile::NamedTempFile;
+
+pub(crate) struct JITCLLVM;
+
+impl JITCLLVM {
+    pub(crate) fn new() -> Result<Arc<Self>, Box<dyn Error>> {
+        Ok(Arc::new(JITCLLVM))
+    }
+}
+
+impl Compiler for JITCLLVM {
+    fn compile(
+        &self,
+        irtrace: IRTrace,
+    ) -> Result<(*const c_void, Option<NamedTempFile>), Box<dyn Error>> {
+        irtrace.compile()
+    }
+}
--- a/ykrt/src/compile/mod.rs
+++ b/ykrt/src/compile/mod.rs
@@ -1,0 +1,26 @@
+use crate::trace::IRTrace;
+use libc::c_void;
+use std::{error::Error, sync::Arc};
+use tempfile::NamedTempFile;
+
+#[cfg(jitc_llvm)]
+mod jitc_llvm;
+
+/// The trait that every JIT compiler backend must implement.
+pub(crate) trait Compiler: Send + Sync {
+    /// Compile an [IRTrace] into machine code.
+    fn compile(
+        &self,
+        irtrace: IRTrace,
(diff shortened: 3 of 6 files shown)


ykrt/build.rs
```diff
@@ -1,3 +1,5 @@
 pub fn main() {
     ykbuild::apply_llvm_ld_library_path();
+
+    println!("cargo:rustc-cfg=jitc_llvm");
```
vext01, line 4: Should this be a cargo feature instead?

ltratt: It's not obvious to me that there any advantages to one over t'other. Any suggestions?

ykrt/src/compile/jitc_llvm.rs
```diff
@@ -0,0 +1,24 @@
...
+
+pub(crate) struct JITCLLVM;
+
+impl JITCLLVM {
+    pub(crate) fn new() -> Result<Arc<Self>, Box<dyn Error>> {
+        Ok(Arc::new(JITCLLVM))
+    }
+}
+
+impl Compiler for JITCLLVM {
```
vext01, line 17: You've said yourself that we have to be very specific with names of things in this system. We have many compilation steps in our system, so should `Compiler` be `TraceNativeCodeGenerator` or something similarly very specific (but hopefully shorter)?

ltratt: I umed and ahed on this, but on reflect "AOT compiler" and "JIT compiler" seem like reasonable pairings to me. Do I like "JIT" as a term? Not really. But "TraceNative" doesn't conjure up a useful image to me, and just feels long for the sake of it.

Naming is a nightmare...

ykrt/src/trace/mod.rs
```diff
@@ -161,6 +162,15 @@ impl IRTrace {
         (di_tmp, di_fd, di_tmpname_c)
     }
 
+    #[cfg(test)]
```
vext01, line 165: Hah, filth :)

ykrt/src/mt.rs
```diff
@@ -73,6 +74,9 @@ pub struct MT {
     /// The [Tracer] that should be used for creating future traces. Note that this might not be
     /// the same as the tracer(s) used to create past traces.
     tracer: Mutex<Arc<dyn Tracer>>,
+    /// The [Compiler] that will be used for compiling future mapped traces. Note that this might
```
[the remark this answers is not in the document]
ltratt: Isn't a "mapped trace" an `IRTrace`? I could use `IRTrace` instead I suppose.

ykrt/src/trace/mod.rs
ltratt: Yep, this is horrendous!

vext01: (sorry, that was meant for `llvmbc_section`)

ykrt/build.rs
vext01: A cargo feature can be configured without source code changes.

ykrt/src/mt.rs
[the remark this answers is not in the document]
vext01: That's correct, and yes please.

ykrt/src/compile/mod.rs
```diff
@@ -0,0 +1,26 @@
+use crate::trace::IRTrace;
+use libc::c_void;
+use std::{error::Error, sync::Arc};
+use tempfile::NamedTempFile;
+
+#[cfg(jitc_llvm)]
+mod jitc_llvm;
+
+/// The trait that every JIT compiler backend must implement.
```
vext01, line 9: I think we can avoid some of the naming problems by making this comment a bit more specific? What is a "JIT compiler backend"? Is it the backend that takes a mapped trace to an IR, or the backend that takes an IR to native code?

We do hope this distinction disappears in the end of course...

Apologies if I nitpick too much.

vext01:
Actually, don't worry too much about naming. Our abstractions will change.

Ready to merge?

ltratt:
Please go ahead!

PR #813, opened 2023-08-19
Move more of the LLVM JIT compiler code into `jitc_llvm.rs`.

ltratt:
At a high-level, this commit inlines the call to `irtrace.compile` in:

```rust
impl Compiler for JITCLLVM {
    fn compile(
        &self,
        irtrace: IRTrace,
    ) -> Result<(*const c_void, Option<NamedTempFile>), Box<dyn Error>> {
      irtrace.compile()
    }
}
```

That has various knock-on effects which somewhat drown this out. Mostly these are rather minor, but note that `llvmbc_section`:

  1. Moves from ykutil to ykrt.
  2. Moves from using linker hacks to `dlsym` (I couldn't get the linker hacks to work with the rest of this commit).

Interestingly, this commit indirectly makes clearer the depth of the assumption that "LLVM is our only JIT", for example in deopting. That should help us continue to refactor things.

### DIFF
--- a/tests/src/bin/run_trace_compiler_test.rs
+++ b/tests/src/bin/run_trace_compiler_test.rs
@@ -4,7 +4,10 @@
 //! `trace_compiler` directory of this crate.
 
 use std::{collections::HashMap, convert::TryInto, env, error::Error, ffi::CString, fs::File};
-use ykrt::trace::{IRBlock, IRTrace};
+use ykrt::{
+    compile::default_compiler,
+    trace::{IRBlock, IRTrace},
+};
 
 const BBS_ENV: &str = "YKT_TRACE_BBS";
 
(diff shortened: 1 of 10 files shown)


ltratt:
Oops, since we have a number of unused things in master, I didn't notice that that c638822 introduced more! Fixed in 1757f0c. Will need squashing.

vext01:
LGTM.

(Later we should wrap dlsym(), divert all call sites through our wrapper, and annotate it `#[cached]`)

bors r+

vext01:
Sorry, please squash.

ltratt:
>  (Later we should wrap dlsym(), divert all call sites through our wrapper, and annotate it #[cached])

We would probably (at a minimum) need to somehow involve dlopen/dlclose in this, because I can imagine scenarios where we end up with a stale cache.

ptersilie:
I quickly skimmed through this assuming Edd has already done the major review work. But I can see where this is going and I think its a good idea (even if that means I have some annoying rebasing to do on my branch).

bors r+

PR #815, opened 2023-08-21
More splitting apart of HWT

ltratt:
This PR does a bit more splitting apart of HWT, in the sense of making it a more distinct component within ykrt. Each commit hopefully contains appropriate motivation for that change.


vext01:
I would have gone for `LLVMBlock` or `LLVMBlockID` instead of `TracedAOTBlock`, but it hardly seems to matter if we plan to kill LLVM in the JIT runtime.

ltratt:
> I would have gone for LLVMBlock or LLVMBlockID instead of TracedAOTBlock, but it hardly seems to matter if we plan to kill LLVM in the JIT runtime.

As you know, naming is not my strong point :) But I very deliberately wanted to make clear these were AOT blocks so they don't get muddled up with JIT LLVM blocks. They could have been (and in a future could be1) `AOTLLVMBlock` or similar. [FWIW, I think "BlockID" is incorrect because that suggests the struct is referencing a unique ID, when in fact multiple struct instances can record the same underlying block ID because of unrolling.]

vext01:
> But I very deliberately wanted to make clear these were AOT blocks

That's a good point.

> multiple struct instances can record the same underlying block ID because of unrolling

Not so sure about that though. Unrolling where? If LLVM did unrolling, then the IR reflects that, and the blocks inside the loop iterations are distinct (with unique IDs).

ltratt:
> >   multiple struct instances can record the same underlying block ID because of unrolling
> Not so sure about that though. Unrolling where? If LLVM did unrolling, then the IR reflects that, and the loop iterations are distinct (and their blocks with unique IDs).

If we record:

```c
for (...) {
  x
}
```

and the loop goes around twice, our trace will be `x x` i.e. we'll record the same underlying block ID twice.

vext01:
Right. But the underlying AOT block is indeed unique. There is no requirement for each AOT block to appear only once in a trace.

It doesn't matter that much, let's move on. If any of these names cause confusion later we can rename them again.

bors r+

vext01:
(Maybe you were thinking that each `IRBlock` uniquely identifies a position in the trace?)

ltratt:
> (Maybe you were thinking that each IRBlock uniquely identifies a position in the trace?)

I know that's not the case: I meant that by having "ID" in the name, you end up with an ambiguity over which is the unique thing. But, as you said, it's not a big deal: I'm just explaining my pea-brained motivation for not using "ID" in the name :)

vext01:
Right, gotcha.

PR #816, opened 2023-08-21
There's no need to store a `Tracer` object.

ltratt:
When we start tracing, all we need to remember is the `ThreadTracer`. If a `ThreadTracer` (or a mapper) wants access to the `Tracer` (e.g. for shared caching), it can store that instance itself because the `Tracer` is `self: Arc<Self>` and can thus store/pass references to itself.

[*Why* this was in the code I no longer remember. I suspect it was needed at some point in the past and no longer is. But, perhaps, it was never needed!]

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -210,7 +210,7 @@ impl MT {
                 match Arc::clone(&tracer).start_collector() {
                     Ok(tt) => THREAD_MTTHREAD.with(|mtt| {
                         promote::thread_record_enable(true);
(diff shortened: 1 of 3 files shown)


ltratt:
@vext01 Oops, rustfmt stupidity on my part. OK to squash?

vext01:
yup

ltratt:
Squashed (and sorry for not doing the rustfmt earlier!).

vext01:
no worries.

bors r+

PR #821, opened 2023-08-24
Remove `TraceDecoderBuilder`.

ltratt:
This is a lot of complexity that we currently, and probably indefinitely, have no use for. The important thing is that there is a `TraceDecoder` trait which means that it is possible to swap different decoders in/out whether or not there's a `Builder` or not.


ltratt:
I didn't think to check the benchmarks! With luck 43f8cb6 fixes things.

bors try

ltratt:
OK, that fixed it. OK to squash?

vext01:
go ahead.

ltratt:
Squashed.

PR #822, opened 2023-08-25
Simpler trace decoder API

ltratt:
This PR simplifies the trace decoding API. Previously, hwtracer had separate notions of a "trace" and a "trace decoder".

Although it wasn't obvious (because we only have one type of tracer and one type of decoder at the moment) this API was suboptimal because it allowed you to express at compile-time impossible combinations. For example, (let's pretend there are in fact multiple decoders!), I could have created a "PT trace" and a "CoreSight" decoder, and that would have compiled, and then done lord alone knows what at run-time. We partly solved this by forcing the client of hwtracer to know what type of trace it had created and to use the correct decoder, accidentally hardcoding details that we would have preferred to be internal to hwtracer into the client.

After a seemingly minor simplification (https://github.com/ykjit/yk/commit/4def543d4c159855251d566fc80936d41078b0bb) this PR thus moves `iter_blocks` to `Trace` itself (https://github.com/ykjit/yk/commit/1beeb8267028e2e18c4d202328b1fe1ad91f426d). There are various further simplifications / alterations we could make by going further in this direction (e.g. this PR does not give users a way of specifying which decoder they want), but I'm not yet sure which looks the most promising. This thus seems like a useful checkpoint for a PR.

### DIFF
--- a/hwtracer/src/collect/perf/mod.rs
+++ b/hwtracer/src/collect/perf/mod.rs
@@ -4,8 +4,9 @@ use super::PerfCollectorConfig;
 use crate::{
     c_errors::PerfPTCError,
     collect::{ThreadTracer, Tracer},
+    decode::ykpt::YkPTBlockIterator,
     errors::HWTracerError,
-    Trace,
+    Block, Trace,
 };
 use libc::{c_void, free, geteuid, malloc, size_t};
 use std::{convert::TryFrom, fs::File, io::Read, slice, sync::Arc};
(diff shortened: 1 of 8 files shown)


ltratt:
I've remembered to test `cargo bench` this time (no effect!), but since there may be other parts where this reaches, I'll do a `try` just to be sure I've caught everything relevant.

bors try

vext01:
Do you think we will ever have a choice of decoders for any given trace format?

ltratt:
> Do you think we will ever have a choice of decoders for any given trace format?

Yes, but probably only after a lot more refactoring. My working guess is that the best way to do this will be to flatten hwtracer into ykrt, and use some careful `enum`s and `#[cfg(...)]` tricks, but there's a lot of work between here and there, and even minor things can throw that plan off track.

ltratt:
Finding a good design won't be trivial because we have more than two factors that combine together (e.g. perf/non-perf, PT/CoreSight, ykpt/libt, etc). I do have a rough idea of where I'd like to get to, but it's hard to say if/when I'll get there.

PR #824, opened 2023-08-27
Rework hwtracer

ltratt:
This PR -- which has a scary number of commits, but doesn't change *vast* quantities of code, and deletes slightly more than it adds -- reworks hwtracer, specifically its internal structure. This allows us to make the external interface more complete, slightly simpler, and (mostly) in one place: see the revised [`hwtracer/src/lib.rs`](https://github.com/ltratt/yk/blob/f06f1b95bb4b62b3801ad98a7484ef34de9cc4cb/hwtracer/src/lib.rs) which contains the entire mainstream user interface (minus blocks which... I'm not sure what to do with yet). Every commit in this PR passes all tests: as that suggests, each commit aims to nudge the system gradually in the right direction, rather than do so in one single shot.

The changes in this PR are based on one fundamental observation: we can't neatly separate "collectors" from "decoders". Indeed, the existing code implicitly shows that the two are inevitably intertwined. For example, if you want to collect PT traces, you need a collector which knows about PT, and a decoder which knows about it. Unsurprisingly, the current "perf collector" knows a lot about PT.

The easiest way to see the changes this PR makes is to first look at hwtracer's previous file structure:

```
.clangd
.gitignore
Cargo.toml
README.md
build.rs
feature_checks
feature_checks/check_perf.c
feature_checks/check_perf_pt.c
src
src/block.rs
src/c_errors.rs
src/collect
src/collect/mod.rs
src/collect/perf
src/collect/perf/collect.c
src/collect/perf/mod.rs
src/decode
src/decode/mod.rs
src/decode/ykpt
src/decode/ykpt/mod.rs
src/decode/ykpt/packet_parser
src/decode/ykpt/packet_parser/mod.rs
src/decode/ykpt/packet_parser/packets.rs
src/errors.rs
src/lib.rs
src/llvm_blockmap.rs
```

and what this PR changes it to:

```
.clangd
.gitignore
Cargo.toml
README.md
build.rs
feature_checks
feature_checks/linux_perf.c
src
src/block.rs
src/errors.rs
src/lib.rs
src/llvm_blockmap.rs
src/perf
src/perf/collect.c
src/perf/collect.rs
src/perf/mod.rs
src/pt
src/pt/c_errors.rs
src/pt/mod.rs
src/pt/ykpt
src/pt/ykpt/mod.rs
src/pt/ykpt/packets.rs
src/pt/ykpt/parser.rs
```

In other words, there is no "collector" or "decoder" modules: "perf" and "pt" (and, if we had them, say "coresight" etc etc) are top-level concepts. The changes necessary to get here take up many of the commits.

https://github.com/ykjit/yk/pull/822 dropped the pretence that arbitrary decoders could work with arbitrary collectors, but didn't really propose a complete replacement. This PR does, in the `TracerKind` enum:

```rust
/// The kind of [Tracer] to be built by [TracerBuilder].
pub enum TracerKind {
    // If you add a new variant, don't forget to update `all_collectors` in the `test` mod later in
    // this file.
    #[cfg(all(linux_perf, pt))]
    /// An IntelPT tracer. Note that this currently uses the ykpt decoder.
    PT(perf::PerfCollectorConfig),
}
```

This is obviously rather simplistic at the moment (we have... one collector... and one decoder!). The idea is that we can add the combinations we support into this enum easily. Let's say we want to offer libipt as a decoder option. We could do something like:

```rust
// lib.rs
pub enum TracerKind {
  PT(perf::PerfCollectorConfig, perf::PtDecoder)
}

// pt/mod.rs
enum PtDecoder {
  YkPT,
  libipt
}
```

In other words, we can make sure the user can only statically express combinations that we definitely support.

I could go on, but the individual commits tell the story in more detail if you need it.

The other thing this PR does -- since prodding around the internal structure helped highlight this to me -- is to start to break apart the obviously non-portable code and feature gate it. This isn't complete -- for example, our "perf" module hard-codes PT support in all sorts of ways (I'm not even sure if the C code could reasonably be made applicable to non-PT things, but that's partly my ignorance of the perf interface). But now, at least, more things will fail to build at compile-time in ways that will help anyone extending hwtracer more clearly identify which parts of the system they need to change.

There is, definitely, much more that could be done -- but I think this PR is big enough already!


ltratt:
bors try

vext01:
Maybe we need to speak about this in-person.

I've not read any code yet, but I'm confused by:

> The changes in this PR are based on one fundamental observation: we can't neatly separate "collectors" from "decoders"

Why not? I find the separation of collector and decoder (and the filesystem hierarchy that comes with it) really intuitive.

You are right that we didn't couple encoders and decoders properly in the past, but I don't think it would be hard, would it?

```
enum TraceFormat {
  PT
  CoreSight,
  Software,
}


trait Collector {
  fn collects() -> Set<TraceFormat>;
}

trait Decoder {
  fn decodes -> Set<TraceFormat>;
}
```

Then it's easy to decide combinations that can work.

vext01:
(And I do feel bad objecting to this change, because it looks like you spent a lot of time on it...)

ltratt:
> I find the separation of collector and decoder (and the filesystem hierarchy that comes with it) really intuitive.

You're right, it is intuitive! Unfortunately it is also too inflexible, although you do have to use a bit of imagination to see that from the current code.

The challenge is that collectors and decoders are coupled and collectors can usefully share code with other collectors. Consider PT: you have to know something about PT to collect PT traces *and* to decode them, but quite a lot of the collector code isn't very PT specific.

Before this PR, hwtracer would arguably be better named "linuxperfpttracer", because "Linux + Perf + PT" is assumed absolutely everywhere, and explicitly hardcoded in many okaces too. In a sense, this PRs main contribution is to weaken that assumption + hardcoding. It doesn't do a perfect job, but I would argue that it does improve more things than it worsens.

Of course, the next question is: if we can't have a nice split of collectors and encoders, what can we replace them with? Imagine that we use your suggestion, where "trace format" (e.g. PT) is the main concept. That sounds OK, until we realise that the collector isn't PT specific. For example, I'm pretty sure that our "perf" support could (and, probably, should) be fairly generic over most (all?) trace formats. In other words, I suspect that a few `#ifdef`s in some C code can allow us to support both CoreSight and PT. But if "trace format" is the top-level concept, do both PT and CoreSight need to have from-scratch Perf support?

So, in some sense it's not rocket science to imagine moving from the builder API that https://github.com/ykjit/yk/pull/824/commits/413d1dfd73a42a0f358b56cd75677866b64fa1b1 creates to something like you're suggesting. If the API you're suggesting does end up being the right thing, it'll be easier to get there after this PR than before it (if that makes sense).

hwtracer/src/pt/mod.rs
```diff
@@ -1,3 +1,4 @@
+pub(crate) mod c_errors;
```
vext01, line 1: Not all of the c_errors are PT-specific, but we can re-share those parts later if we ever implement another trace collector.

ltratt: Yes, I agree. In general, I think our error enums are weirdly unuseful, because they're so fine-grained that we push the complexity of the errors onto callers. Ironically, I think enums with fewer cases would be better for us. But I haven't fully thought that through, and it feels like something for a future PR to consider.

vext01: I was trying to give better errors than "PT failed somehow" (one of the PT error codes) or "Something else in C failed" (an errno error code). Super generic errors waste time in debugging, tracking down where the error came from.

We might be able to do better, but it's not super important right now.

ltratt: I think, in general, `Box<dyn Error>` is the way to go, because you can put in an arbitrarily detailed string, but without giving the user the impression they can do much about the error. In the case of PT, we arguably only care about "an error that might go away if you try again" vs. "an error that's probably permanent".

hwtracer/build.rs
```diff
@@ -37,7 +37,7 @@ fn main() {
 
             #[cfg(target_arch = "x86_64")]
             {
-                println!("cargo:rustc-cfg=decoder_ykpt");
+                println!("cargo:rustc-cfg=ykpt");
```
vext01, line 40: This isn't a renaming I'd have done myself, but it's probably not worth arguing about.

vext01:
There's some good stuff in this PR mixed in with some stuff I'm not so hot about.

I still wish we'd kept the notion of a encoder separate from a decoder. The discussion yesterday didn't really sway me to be honest.

But to argue about it is to waste time, so unless @ptersilie wildy disagrees, I guess we move on?

ltratt:
I think this is a decision for you alone.

hwtracer/src/pt/mod.rs
vext01: You surprise me! Once upon a time you used to dislike `dyn Error` due to the dynamic dispatch overhead.

Although we are talking about error paths, some are recoverable error paths, so performance may still matter.

I don't know how much `dyn` slows us down here though.

ltratt: > Once upon a time you used to dislike dyn Error due to the dynamic dispatch overhead.

I don't think I ever said that. Yes, dynamic dispatch has overhead, but for errors it's mostly a non-issue.

PR #826, opened 2023-09-08
Kill unreachable deopt cases.

vext01:
Whilst debugging another issue, we realised that "direct" stackmap entries were only possible for the constructs that we explicitly excluded from the shadow stack, and those can only appear in the AOT control point frame (for now).

Rather than have untested and possibly incorrect direct record handlers, we instead remove them and crash if they arise when we don't expect them.

All tests pass, both here and in yklua.

### DIFF
--- a/ykrt/src/deopt.rs
+++ b/ykrt/src/deopt.rs
@@ -204,14 +204,10 @@ extern "C" fn ts_reconstruct(ctx: *mut c_void, module: LLVMModuleRef) -> LLVMErr
                 let _val = unsafe { registers.get(*reg) };
                 todo!();
             }
-            SMLocation::Direct(reg, off, _size) => {
-                // When using `llvm.experimental.deoptimize` then direct locations should always be
-                // in relation to RBP.
-                assert_eq!(*reg, 6);
-                let addr = unsafe { registers.get(*reg) as *mut u8 };
-                let addr = unsafe { addr.offset(isize::try_from(*off).unwrap()) };
-                let aot = &aotvals[i];
-                framerec.var_init(aot.val, aot.sfidx, addr as u64);
+            SMLocation::Direct(..) => {
+                // We use a (heap allocated) shadow stack for both the JIT and AOT stack, so
+                // `Direct` locations can't (currently) arise for JITted code.
+                todo!();
             }
             SMLocation::Indirect(reg, off, size) => {
                 let addr = unsafe { registers.get(*reg) as *mut u8 };
--- a/ykrt/src/frame/mod.rs
+++ b/ykrt/src/frame/mod.rs
@@ -296,29 +296,18 @@ impl FrameReconstructor {
                             registers[usize::try_from(*off - 1).unwrap()] = val;
                         }
                     }
-                    SMLocation::Direct(reg, off, _) => {
-                        if i == 0 {
-                            // Direct locations are pointers into the stack (e.g. alloca or GEP).
-                            // Normally, AOT and JIT have different stacks so copying them over
-                            // would be incorrect. However, since we are using a shadow stack which
-                            // is shared between AOT and JIT these values should be identical and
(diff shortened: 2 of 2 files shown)


ykrt/src/deopt.rs
```diff
@@ -204,14 +204,11 @@ extern "C" fn ts_reconstruct(ctx: *mut c_void, module: LLVMModuleRef) -> LLVMErr
...
-                // When using `llvm.experimental.deoptimize` then direct locations should always be
-                // in relation to RBP.
-                assert_eq!(*reg, 6);
-                let addr = unsafe { registers.get(*reg) as *mut u8 };
-                let addr = unsafe { addr.offset(isize::try_from(*off).unwrap()) };
-                let aot = &aotvals[i];
-                framerec.var_init(aot.val, aot.sfidx, addr as u64);
+            SMLocation::Direct(..) => {
+                // Normally a `Direct` location would be used for a local variable expressing a
+                // pointer to the frame. Since JITted code frames have had their stack "unified"
```
ptersilie, line 209: Maybe we can just directly say that we have a shadow stack that is shared between AOT and JIT, so readers don't need to read between the lines.

ykrt/src/deopt.rs
```diff
@@ -204,14 +204,11 @@ extern "C" fn ts_reconstruct(ctx: *mut c_void, module: LLVMModuleRef) -> LLVMErr
...
+                // with the AOT stack (a heap allocation), we don't see any `Direct` locations.
+                unreachable!();
```
ptersilie, line 211: Not sure it should be an unreachable, since in the future we will have direct locations (once we only move references values to the shadow stack). So maybe this should be a `todo`.

ykrt/src/frame/mod.rs
```diff
@@ -296,29 +296,18 @@ impl FrameReconstructor {
...
-                        let size = unsafe { LLVMABISizeOfType(layout, eltype) };
-                        // Direct locations are always be in regards to RBP.
-                        debug_assert_eq!(*reg, RBP_DWARF_NUM);
-                        let temp = unsafe { rbp.offset(isize::try_from(*off).unwrap()) };
-                        debug_assert!(*off < i32::try_from(rec.size).unwrap());
-                        unsafe {
-                            libc::memcpy(temp, val as *const c_void, usize::try_from(size).unwrap())
-                        };
+                    SMLocation::Direct(..) => {
+                        // Direct locations are pointers to the stack, stored on the stack (e.g.
```
ptersilie, line 300: They are not stored on the stack. They are direct pointers to values on the stack.

ykrt/src/frame/mod.rs
```diff
@@ -296,29 +296,18 @@ impl FrameReconstructor {
...
+                        // `alloca` or GEP). Our shadow stack unifies the (would have been
```
ptersilie, line 301: Drop the "would have been separate"?

ykrt/src/frame/mod.rs
vext01: Yeah, which themselves are on the stack :P

ptersilie: Ah yes, you're right.

ykrt/src/deopt.rs
ltratt: `unreachable` will panic so if we currently think we shouldn't get here, `unreachable` is fine (and can help the optimiser do a mildly better job).

vext01:
All comment addressed in 4bef80de.

ykrt/src/deopt.rs
vext01: Well I changed it to `todo!()` which also panics, so I guess that's OK too.

vext01:
> typos in the commit message

will fix when squashing.

ptersilie:
Please squash.

vext01:
splat and fixed typos in commit message and PR description.

PR #828, opened 2023-09-11
Implement side-traces.

ptersilie:
When a guard fails a specified amount of times (default: 50), we now compile a side-trace which runs from the guard failure to the control point. The next time the same guard fails we can now execute the side-trace instead. There's one caveat: when a side-trace runs all the way to the end, we currently need to deoptimise back to the control point, which then runs one iteration in the normal interpreter before executing a trace again. While this can be fixed by placing one additional stackmap before the control point, it's probably not worthwhile as we'll be soon moving to our own trace compiler.


ykrt/src/compile/mod.rs
```diff
@@ -38,15 +52,35 @@ struct SendSyncConstPtr<T>(*const T);
...
-    failed: u32,
-    code: Option<SendSyncConstPtr<c_void>>,
+pub struct Guard {
+    failed: AtomicU32,
+    ct: Mutex<Option<Arc<CompiledTrace>>>,
+}
+
+impl Guard {
+    pub fn inc(&self) {
+        let n = self.failed.load(Ordering::Relaxed);
```
ltratt, line 62: This has a race condition: you could load "3", another thread can store "4", then you store "4" i.e. you've lost an increment. You want `fetch_add`, which avoids this race.

ykrt/src/compile/mod.rs
```diff
@@ -38,15 +52,35 @@ struct SendSyncConstPtr<T>(*const T);
...
+        self.failed.store(n + 1, Ordering::Relaxed);
+    }
+
+    pub fn failcount(&self) -> u32 {
+        self.failed.load(Ordering::Relaxed)
+    }
+
+    pub fn setct(&self, ct: Arc<CompiledTrace>) {
+        let _ = self.ct.lock().insert(ct);
```
ltratt, line 71: Why the `let _`?

ykrt/src/compile/mod.rs
```diff
@@ -38,15 +52,35 @@ struct SendSyncConstPtr<T>(*const T);
...
+    }
+
+    pub fn getct(&self) -> Option<Arc<CompiledTrace>> {
+        self.ct.lock().as_ref().cloned()
```
ltratt, line 75: I get nervous with `clone`/`cloned` because if the inner type changes this will do something, but probably not what we want. If possible, can we do something like `self.ct.lock().map(|x| Arc::clone(x))` to avoid this?

ykrt/src/compile/mod.rs
```diff
@@ -60,21 +94,28 @@ pub struct CompiledTrace {
     /// Pointer to heap allocated live AOT values.
     aotvals: SendSyncConstPtr<c_void>,
     /// List of guards containing hotness counts or compiled side traces.
-    guards: Vec<Option<Guard>>,
+    pub guards: Vec<Guard>,
```
ltratt, line 97: `pub(crate)`

ykrt/src/compile/mod.rs
```diff
@@ -60,21 +94,28 @@ pub struct CompiledTrace {
...
     #[allow(dead_code)]
     di_tmpfile: Option<NamedTempFile>,
+    /// Reference to the HotLocation, required for side tracing.
+    pub(crate) hl: Weak<Mutex<HotLocation>>,
 }
 
 impl CompiledTrace {
     /// Create a `CompiledTrace` from a pointer to an array containing: the pointer to the compiled
     /// trace, the pointer to the stackmap and the size of the stackmap, and the pointer to the
     /// live AOT values.
```
ltratt, line 112: This comment is now out of date with the `hl` argument.

ykrt/src/deopt.rs
```diff
@@ -268,18 +277,82 @@ unsafe extern "C" fn __ykrt_deopt(
...
 
+    // Check if we have a side trace and execute it.
+    if guardid != SIDETRACE_LAST_GUARD_ID {
+        let guard = &ctr.guards[guardid];
+        if let Some(st) = guard.getct() {
+            let registers = Registers::from_ptr(rsp);
+            let live_vars = ctr.smap.get(&retaddr.try_into().unwrap()).unwrap();
+            let mut ykctrlpvars = Vec::new();
+            for (_i, locs) in live_vars.iter().skip(1).enumerate() {
+                assert!(locs.len() == 1);
```
ltratt, line 297: Should probably be `debug_assert`?

ltratt: Given the next line will crash (safely) if this fails, the `assert` isn't doing anything useful. Rather this should be a comment explaining why `locs[0]` won't trigger the `unwrap`.

ykrt/src/deopt.rs
```diff
@@ -268,18 +277,82 @@ unsafe extern "C" fn __ykrt_deopt(
...
+                let l = locs.get(0).unwrap();
```
ltratt, line 298: `let l = locs[0];` should work?

ykrt/src/deopt.rs
```diff
@@ -268,18 +277,82 @@ unsafe extern "C" fn __ykrt_deopt(
...
+                match l {
+                    SMLocation::Register(reg, _size, _off, _extra) => {
+                        let _val = unsafe { registers.get(*reg) };
+                        todo!();
+                    }
+                    SMLocation::Direct(reg, off, _size) => {
+                        // When using `llvm.experimental.deoptimize` then direct locations should always be
+                        // in relation to RBP.
```
ltratt, line 306: `debug_assert`

ykrt/src/deopt.rs
```diff
@@ -268,18 +277,82 @@ unsafe extern "C" fn __ykrt_deopt(
...
+                            1 => unsafe { ptr::read::<u8>(addr) as u64 },
+                            2 => unsafe { ptr::read::<u16>(addr as *mut u16) as u64 },
+                            4 => unsafe { ptr::read::<u32>(addr as *mut u32) as u64 },
+                            8 => unsafe { ptr::read::<u64>(addr as *mut u64) },
+                            _ => unreachable!(),
+                        };
+                        ykctrlpvars.push(v);
+                    }
+                    SMLocation::Constant(v) => {
+                        ykctrlpvars.push(*v as u64);
```
ltratt, line 325: Can we make this a platform independent size?

ykrt/src/deopt.rs
```diff
@@ -302,7 +375,30 @@ unsafe extern "C" fn __ykrt_deopt(
     // pass in variables from this scope via a struct which is passed into the function.
     LLVMOrcThreadSafeModuleWithModuleDo(moduleref, ts_reconstruct, infoptr as *mut c_void);
 
+    // We want to start side tracing only after we deoptimised so we don't trace the whole deopt
```
ltratt, line 378: Perhaps split this sentence in two, at least using a comma?

ykrt/src/deopt.rs
```diff
@@ -302,7 +375,30 @@ unsafe extern "C" fn __ykrt_deopt(
...
+    // routine which will later be costly to disassemble.
+    if guardid != SIDETRACE_LAST_GUARD_ID {
+        let guard = &ctr.guards[guardid];
+        if guard.failcount() >= ctr.mt.sidetrace_threshold() - 1 {
```
ltratt, line 382: What happens if `sidetrace_threshold` is 0?

ykrt/src/mt.rs
```diff
@@ -59,10 +60,32 @@ static SERIALISE_COMPILATION: LazyLock<bool> = LazyLock::new(|| {
     &env::var("YKD_SERIALISE_COMPILATION").unwrap_or_else(|_| "0".to_owned()) == "1"
 });
 
+#[derive(Debug, Copy, Clone)]
+pub struct SideTraceInfo {
```
ltratt, line 64: Let's always use `pub(crate)` unless we *have* to make things `pub`.

yktracec/src/jitmodbuilder.cc
```diff
@@ -459,6 +470,9 @@ class JITModBuilder {
   // compilation.
   CallStack CallStack;
 
+  // Flag showing whether this trace is a normal trace or a side trace.
```
ltratt, line 473: Better something like "Set to true for a sidetrace or false for a normal trace"?

yktracec/src/jitmodbuilder.cc
```diff
@@ -828,6 +842,10 @@ class JITModBuilder {
     assert(CurFrame);
     CurFrame->setResume(CurBBIdx, Instr, CurInstrIdx);
 
+    // Clone the CallStack, store it on the heap, and hardcode its pointer
```
ltratt, line 845: Why do we clone the whole callstack?

yktracec/src/jitmodbuilder.cc
```diff
@@ -1178,6 +1203,24 @@ class JITModBuilder {
     Builder.SetInsertPoint(LoopEntryBB);
   }
 
+  // Generate loads for the live variables passed into the side trace call.
+  // Unlike the normal trace we can't loop back to the top of the side-trace
+  // (the side-trace ends at the control point and this is were we need to
```
ltratt, line 1208: s/were/where/

ltratt:
This is a very important PR, which is good :) I would like to see three changes though:

1. Document all new structs/attributes/methods. I know this feels like a chore, and it can feel like everything is obvious, but our future selves might be even more stupid than us -- I am invariably grateful for comments my past self made.
2. We need more tests. This is too important for 1 small test to give us confidence. I guess that some existing tests are perhaps now compiling side traces and/or those tests can be extended/duplicated-with-changes to do so? For example, we need tests that: nest side traces; side traces in inlined functions; and so on. If some of those tests don't succeed, we don't necessarily need to fix them in this PR (we can mark them as "ignored"), but it will tell us what we need to do soon.
3. Think a bit about platform dependence and how we can reduce it. Whenever I see types like `u64` I wince inside because it generally means we're thinking "all the world's an x86". At the very least, any method which can't work on anything except x86 should have an `#[cfg(...)]` but in many cases it's no real extra effort to abstract over the platform. I'm not sure this PR introduces *too* much extra platform dependence, but there were a few bits I was unsure about, so I'd like reassuring :)

tests/c/side-trace.c
```diff
@@ -0,0 +1,103 @@
...
+//     jit-state: execute-side-trace
+//     28
+//     jit-state: deoptimise
+//     30
+//   stdout:
+//     exit
+
+// Testing side tracing functionality. The program immediately compiles a
+// trace, which runs for 10 iterations at which point the guard generated by
+// the if/else in function `foo` fails. After failing for 5 times, a side trace
```
vext01, line 57: Says after 5 times here, but you said 50 elsewhere?

ykrt/src/compile/mod.rs
ptersilie: Is it? We still need this `hl` for side-tracing. We've just changed it from an `Arc` to a `Weak`.

tests/c/side-trace.c
ptersilie: 50 is the default, but for this test I changed it to 5. See https://github.com/ykjit/yk/pull/828/files/3b64ba16a80c11718bf1ea730eb756982d48cd7d#diff-66116c893a3e9b6083a5eefd36f66447a0461fd33defc495275221ad16457e77R84

ykrt/src/deopt.rs
ptersilie: So `let l = locs[0];` will also fail savely, so there's no need for the assert?

ykrt/src/deopt.rs
ptersilie: No, this is always a u64 in LLVM stackmaps.

ykrt/src/deopt.rs
ptersilie: Whoops. Will change it to `>`.

ykrt/src/mt.rs
```diff
@@ -50,6 +50,7 @@ pub type AtomicTraceFailureThreshold = AtomicU16;
 /// FIXME: needs to be configurable.
 pub(crate) const DEFAULT_TRACE_TOO_LONG: usize = 5000;
 const DEFAULT_HOT_THRESHOLD: HotThreshold = 50;
+const DEFAULT_SIDETRACE_THRESHOLD: HotThreshold = 50;
```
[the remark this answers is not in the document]
ptersilie: Really, it's that low? When I had it at 5, `db.lua` was compiling A LOT of sidetraces.

yktracec/src/jitmodbuilder.cc
ptersilie: Will update the comment: we need it to initialise the jitmodbuilder when tracing after a guardfailure. Otherwise, when a guard fails inside a guard, we don't know where to continue tracing when that call returns.

ykrt/src/compile/mod.rs
ptersilie:
```
warning: unused return value of `Option::<T>::insert` that must be used
  --> ykrt/src/compile/mod.rs:71:9
   |
71 |         self.ct.lock().insert(ct);
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: if you intended to set a value, consider assignment instead
   = note: `#[warn(unused_must_use)]` on by default
help: use `let _ = ...` to ignore the resulting value
   |
71 |         let _ = self.ct.lock().insert(ct);
```

ykrt/src/compile/mod.rs
ltratt: I mean the comment needs to mention `hl`.

ykrt/src/deopt.rs
ltratt: Correct.

ykrt/src/deopt.rs
ltratt: Aha.

ykrt/src/mt.rs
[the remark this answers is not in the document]
ltratt: Yep, that's what you'd expect. I'm not saying 5 is the perfect number, but until we have a better reason to diverge from RPython, I'd be inclined to go with their number.

ykrt/src/compile/mod.rs
ptersilie: Will fix. But this made me realise if two threads have the same control point, all the hot counts are gonna be shared, so we'll reach the hot threshold much faster. Not saying this is a problem, just something I hadn't thought about before.

ykrt/src/compile/mod.rs
```diff
@@ -91,13 +132,21 @@ impl CompiledTrace {
         // We heap allocated this array in yktracec to pass the data here. Now that we've
         // extracted it we no longer need to keep the array around.
         unsafe { libc::free(data as *mut c_void) };
+        let mut guards = Vec::new();
```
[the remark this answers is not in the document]
ptersilie: It appears not, since that would required `Guard` to be cloneable, which it can't be due to the atomic field.

ykrt/src/compile/mod.rs
ptersilie: Fixed in b009e68cd

ykrt/src/compile/mod.rs
ptersilie: Fixed in b6f02d72f

ykrt/src/compile/mod.rs
ptersilie: Fixed in 223b81a08

ykrt/src/compile/mod.rs
ptersilie: Fixed in c515a462

ltratt: s/passed in references/references passed in/

ykrt/src/compile/mod.rs
ltratt: > But this made me realise if two threads have the same control point, all the hot counts are gonna be shared, so we'll reach the hot threshold much faster.

Yep. It's not a problem though!

ykrt/src/deopt.rs
ptersilie: `locs` is of type `LiveVar` which is a wrapper around a vec, so unfortunately not.

ltratt: ?

[Although my suggestion was wrong. I think `let l = &locs[0];` should work.]

ykrt/src/deopt.rs
ptersilie: Fixed in 24d4962b8

ykrt/src/deopt.rs
ptersilie: Fixed in ddb92565a7d

ykrt/src/deopt.rs
ptersilie: Fixed in cb78bfe39

ykrt/src/deopt.rs
ptersilie: Added comment in: 24d4962b8c

ykrt/src/deopt.rs
ptersilie: Fixed in c0a151a9e20e9

ykrt/src/compile/mod.rs
ptersilie: What I meant is "the references to MT and HotLocation that are passed in as arguments". Your change makes it sound like we are passing stuff into MT and HotLocation which isn't the case here.

ykrt/src/deopt.rs
ptersilie: `locs` is of type `LiveVar`, which is just a struct. You can't index into a struct.

ltratt: Oh! So then we should (as always) document why the `unwrap` can't fail.

ptersilie: https://github.com/ptersilie/yk/blob/e4b4d66bd59709c5955797a78787143d2d2f389d/yksmp/src/lib.rs#L38

ltratt: Ah!

ykrt/src/mt.rs
ptersilie:
```
warning: type `SideTraceInfo` is more private than the item `Compiler::compile`                                                                                                                                                         
  --> ykrt/src/compile/mod.rs:26:5                                                                                                                                                                                                      
   |                                                                                                                                                                                                                                    
26 | /     fn compile(                                                                                                                                                                                                                  
27 | |         &self,                                                                                                                                                                                                                   
28 | |         irtrace: MappedTrace,                                                                                                                                                                                                    
29 | |         sti: &SideTraceInfo,                                                                                                                                                                                                     
30 | |     ) -> Result<(*const c_void, Option<NamedTempFile>), Box<dyn Error>>;                                                                                                                                                         
   | |________________________________________________________________________^ method `Compiler::compile` is reachable at visibility `pub`                                                                                             
   |                                                                                                                                                                                                                                    
note: but type `SideTraceInfo` is only usable at visibility `pub(crate)`                                                                                                                                                                
  --> ykrt/src/mt.rs:64:1                                                                                                                                                                                                               
   |                                                                                                                                                                                                                                    
64 | pub(crate) struct SideTraceInfo {                                                                                                                                                                                                  
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                    
   = note: `#[warn(private_interfaces)]` on by default
```

ltratt: OK.

yktracec/src/jitmodbuilder.cc
ptersilie: Fixed in 2f184428628d

yktracec/src/jitmodbuilder.cc
ptersilie: Fixed in a7f07f7a358

yktracec/src/jitmodbuilder.cc
ptersilie: Fixed in 8933af0be67

ykrt/src/mt.rs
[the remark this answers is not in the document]
ptersilie: Fixed in 0229dcd9005

yktracec/src/jitmodbuilder.cc
ltratt: Maybe I'm not understanding what `CallStack` is, but why do we need to duplicate it? Is it the C call stack or the shadow stack or some frames or ... ?

ykrt/src/compile/mod.rs
ltratt: Fair enough, though "passed in references" doesn't quite parse in English.

ptersilie: Better? fb3470abdb

yktracec/src/jitmodbuilder.cc
ptersilie: It's the JITModBuilder call stack. Remember that we keep saying that JITModBuilder is not unlike an interpreter. Whenever we see a call in the trace, we need to push some info onto the `CallStack` that keeps track of where this call happened (and some other info) so that we can continue copying instructions at the correct block and instruction after we've finished inlining the call.

Now when we collect a side-trace, we are starting in the middle of a function. So when that function returns we have no idea where to continue collecting instructions from. PT only tells us the block but not the instruction. We do collect this information when collecting the parent trace, so we just need to clone it and then pass it to the side-trace.

Again, this is all a consequence of calls not being terminators in LLVM IR.

ltratt: Ah, I get it!

ptersilie:
Added another test which needs reviewing: [Add nested side-traces test.](https://github.com/ykjit/yk/pull/828/commits/04f925f63819ecdf6779be9af841099f1b839749)

You're allowed to say "I told you so" @ltratt

ptersilie:
I'm working on it. I'm not expecting any surprises with nested while loops since they are just unrolled anyway, but will add one. As for a side-trace inside an outlined function, I'm afraid to let you know that won't work, cause an outlined function won't have any guards. :wink:

ptersilie:
Another test for you to review @ltratt.

ltratt:
Please squash (and you'll need to rebase too!).

ptersilie:
Wait, can I remove those ugly comments from the commit message, please?

ltratt:
> Wait, can I remove those ugly comments from the commit message, please?

Yes, you can! Please force push an update.

ptersilie:
Done.

ltratt:
> default: 50

Didn't we change the default to 5?

ptersilie:
We did but I forgot to update the comment. Uno secundo!

ptersilie:
Fixed.

ptersilie:
Sorry, forgot to reformat after all the merge conflict stuff. Okay to squash?

ltratt:
Please squash.

ptersilie:
Squashed!

ptersilie:
Ok to squash?

ltratt:
Please squash.

ptersilie:
Squashed.

ptersilie:
Okay, I think we can leave that last commit as is unless you don't want to have a broken commit in there to not break `git bisect`.

ptersilie:
I also have to rebase...again!

ltratt:
I would prefer to squash that commit away so that, as you say, bisecting will work. Please squash (and rebase).

ptersilie:
Squashed and rebased!

ptersilie:
Let's try it before we do anything else.

bors try

ltratt:
If this passes, please squash.

ptersilie:
Squashed!

PR #829, opened 2023-09-11
Find successful prelink and postlink O2 passes. 

nmdis1999:
This PR aims introduces a python script which operates on a list of passes  (from full O2 or pre/postlink O2) and find passes which are `n` times successful  in building yklua interpreter and running yklua tests.

The divide and conquer algorithm preserves the order of input passes and keeps spliting the list of passes till it finds successful set of passes. Further, the two halves are processed paralelly on separate threads with a shared list to store the successful passes. This reduced the time for single run of `try_passes.py` script from ~450 sec to ~180 secs.

The changes to  ykllvm's `BackendUtil.cpp` makes sure that pre & post-link time passes can be added to `cflags` with a new pass (`newpm-passes`).
I changed `yk-config` now to read new list of string (i.e passes) with argumet `--prelink` or `--postlink` instead of environment variable.

### DIFF
--- a/bin/yk-config
+++ b/bin/yk-config
@@ -10,20 +10,21 @@ if [ ! -d "${DIR}/../ykrt" ]; then
     exit 1
 fi
 
-# The link-time optimisations that we can tolerate.
+# The post-link time optimisations that we can tolerate.
 #
 # FIXME: Once we know what breaks us, allow the user to use the regular -O0,
 # -O1, -O2, etc. clang flags like normal.
-LINKTIME_PASSES=instcombine
+POSTLINK_PASSES_STR=instcombine
 
 OUTPUT=""
 
 usage() {
     echo "Generate C compiler flags for building against the yk JIT.\n"
     echo "Usage:"
-    echo "    yk-config <mode> <--cc|--cxx|--ar|--ranlib|--cppflags|--cflags|--ldflags>\n"
+    echo "    yk-config <mode> [--prelink-pipeline <passes>] [--postlink-pipeline <passes>] <--cc|--cxx|--ar|--ranlib|--cppflags|--cflags|--ldflags>\n"
     echo "    Where <mode> is either 'debug' or 'release'."
 }
+
 
 handle_arg() {
     mode=$1
@@ -41,6 +42,26 @@ handle_arg() {
         --cxx) OUTPUT="${ykllvm_bin_dir}/clang++" ;;
         --ar) OUTPUT="${ykllvm_bin_dir}/llvm-ar" ;;
         --ranlib) OUTPUT="${ykllvm_bin_dir}/llvm-ranlib" ;;
+        --prelink-pipeline)
+            if [ -z "$2" ] || [ "$(echo $2 | cut -b 1,2)" = "--" ]; then
+                1>&2 echo "Error: Expected passes after --prelink-pipeline."
+                usage
+                exit 1
+            else
+                PRELINK_PASSES_STR=$2
+                shift
+            fi
+            ;;
+        --postlink-pipeline)
+            if [ -z "$2" ] || [ "$(echo $2 | cut -b 1,2)" = "--" ]; then
+                1>&2 echo "Error: Expected passes after --postlink-pipeline."
+                usage
+                exit 1
+            else
+                POSTLINK_PASSES_STR=$2
+                shift
+            fi
+            ;;
         --cflags)
             # Enable LTO.
             OUTPUT="${OUTPUT} -flto"
@@ -51,6 +72,9 @@ handle_arg() {
             # Without this, clang will slap `optnone` attributes on every
             # function, causing optimisations to skip them.
             OUTPUT="${OUTPUT} -Xclang -disable-O0-optnone"
+            if [ ! -z "${PRELINK_PASSES_STR}" ]; then
+                OUTPUT="${OUTPUT} -mllvm --newpm-passes=${PRELINK_PASSES_STR}"
+            fi
             case $mode in
                 debug) OUTPUT="$OUTPUT -g" ;;
             esac
@@ -111,7 +135,8 @@ handle_arg() {
             #
             # This pairs with `-Xclang -disable-O0-optnone`. See above.
             OUTPUT="${OUTPUT} -Wl,--mllvm=--yk-optnone-after-ir-passes"
-            OUTPUT="${OUTPUT} -Xlinker --lto-newpm-passes=${LINKTIME_PASSES}"
+
+            OUTPUT="${OUTPUT} -Xlinker --lto-newpm-passes=${POSTLINK_PASSES_STR}"
 
             # Have the `.llvmbc` and `.llvm_bb_addr_map` sections loaded into
             # memory by the loader.
@@ -171,8 +196,14 @@ mode=$1
 shift
 
 while [ $# -ne 0 ]; do
-    handle_arg $mode $1
-    shift
+    if [ -z "$2" ] || [ "$(echo $2 | cut -b 1,2)" = "--" ]; then
+        handle_arg $mode $1
+        shift
+    else
+        handle_arg $mode $1 $2
+        shift
+        shift
+    fi
 done
 
 echo ${OUTPUT}
--- a/ykrt/try_passes.py
+++ b/ykrt/try_passes.py
@@ -1,0 +1,203 @@
+#!/usr/bin/env python3
+
+import argparse, math, os, shutil, sys, time, queue 
+from dataclasses import dataclass
+from multiprocessing import Manager, Process, Queue, Value
+from subprocess import PIPE, Popen
+
+# Stages in an LTO pipeline where optimisation passes can happen.
+STAGES = "pre_link", "link_time"
+@dataclass
+class Pass:
+    name: str
+    # FIXME: implement support for pass parameters?
+    #args: list
+
+    def __str__(self):
+        return self.name
+
+@dataclass
+class PipelineConfig:
+    pre_link: list
+    link_time: list
+
+    def __str__(self):
+        pre_link = ",".join([ str(p) for p in self.pre_link ])
+        link_time = ",".join([ str(p) for p in self.link_time ])
+        return f"PipelineConfig(pre_link=[{pre_link}], link_time=[{link_time}])"
+
+def get_pipeline_config(is_prelink, ok_passes, try_passes):
+    """Returns pipeline configuration based on the flag type."""
+    
+    if not is_prelink:
+        return PipelineConfig([], ok_passes + try_passes)
+    else:
+        return PipelineConfig(ok_passes + try_passes, [])
+
+def get_opt_cmd(is_prelink):
+    """Returns command based on the flag type."""
+    
+    if not is_prelink:
+        return "opt -passes='lto<O2>' -print-pipeline-passes < /dev/null 2>/dev/null | tr ',' '\n'"
+    else:
+        return "opt -passes='lto-pre-link<O2>' -print-pipeline-passes < /dev/null 2>/dev/null | tr ',' '\n'"
+
+def log(logf, s):
+    logf.write(s)
+    logf.flush()
+
+def executable_exists(cmd):
+    return shutil.which(cmd) is not None
+
+def get_all_passes(is_prelink):
+    cmd = get_opt_cmd(is_prelink) 
+    p = Popen(cmd, shell=True, stdout=PIPE, close_fds=True)
+    sout, serr = p.communicate() 
+    assert(p.returncode == 0)
+
+    sout = sout.decode()
+    pass_descrs = [x.strip() for x in sout.strip().split("\n")]
+
+    passes = []
+    seen = set()
+     
+    for descr in pass_descrs:
+        parts = descr.split("<")
+        if parts[0] not in seen:
+            seen.add(parts[0])
+            passes.append(Pass(parts[0]))
+
+    print(f"Found {len(passes)} passes")    
+    return passes
+
+def test_pipeline(logf, pl):
+    sys.stdout.write(str(pl) + "...")
+    sys.stdout.flush()
+
+    log(logf, "\n\n" + str(pl) + "\n")  
+
+    # Make sure we don't run empty strings in pipeline.
+    assert (len(pl.pre_link) != 0 and len(pl.link_time) != 0), "Both prelink and postlink passes cannot be empty!!!"
+
+    env = os.environ
+    env["PRELINK_PASSES"] = ",".join([p.name for p in pl.pre_link])
+    env["LINKTIME_PASSES"] = ",".join([p.name for p in pl.link_time])
+
+    p = Popen("try_repeat 10 sh test.sh 2>&1", cwd=CWD, shell=True,
+              stdout=PIPE, close_fds=True, env=env)
+    
+    sout, _ =  p.communicate()
+    log(logf, sout.decode())
+
+    if p.returncode == 0:
+        print(" [OK]")
+        log(logf, str(pl) + ": OK\n")
+    else:
+        log(logf, str(pl) + " : FAILED\n")
+        print(" [FAIL]")
+    return p.returncode == 0
+
+def list_of_passes_to_str(passes):
+    return ",".join([str(p) for p in passes])
+
+def binary_split_worker(logf, ok_passes, passes_tasks, is_prelink, processing):
+    while True:
+        try:
+            # Wait for a task in the queue for worker_timeout seconds
+            try_passes = passes_tasks.get(block=True)
+            processing.value += 1
+        except queue.Empty:
+            print(f"\033[91mClosing Worker\033[0m")
+            break
+        else:
+            if try_passes == "CLOSE":
+                print(f"\033[33mClosing Worker\033[0m")
+                break
+            log(logf, f">>> Trying to add:\n{list_of_passes_to_str(try_passes)}\n\n")
+            
+            # Choose pipeline based on flag type.
+            config = get_pipeline_config(is_prelink, ok_passes, try_passes)
+
+            if test_pipeline(logf, config):
+                ok_passes.extend(try_passes)
+            elif len(try_passes) == 1:
+                return
+            else:
+                subset_len = len(try_passes)
+                subset1 = try_passes[:subset_len // 2] 
+                subset2 = try_passes[subset_len // 2:]
+
+                # Put the 2 halves of the split in the worker queue to be processed concurrently
+                passes_tasks.put(subset1)
+                passes_tasks.put(subset2)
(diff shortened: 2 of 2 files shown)


ltratt:
> deterministically successful

I'm not sure "deterministically successful" has a meaning. Rather, I think this should be "runs *N* times without failing"? In other words, we're really talking about something along the lines of "successful to a given probability".

> The divide and conquer algorithm preserves the order of input passes and keeps spliting the list of passes till it finds deterministcally successful set of passes.

If I have passes [A, B, C, D, E] what is the sequence of passes that will be tried? I guess (assuming each of these fails!) ABCDE, AB, DE, A, E?

bin/yk-config
```diff
@@ -16,6 +16,11 @@ fi
 # -O1, -O2, etc. clang flags like normal.
 LINKTIME_PASSES=instcombine
 
+# XXX hacked in quick
```
ltratt, line 19: What is this doing, and how can we do it in a way that's not a quick hack?

bin/yk-config
```diff
@@ -109,6 +118,7 @@ handle_arg() {
             #
             # This pairs with `-Xclang -disable-O0-optnone`. See above.
             OUTPUT="${OUTPUT} -Wl,--mllvm=--yk-optnone-after-ir-passes"
+
```
ltratt, line 121: Extraneous blank line.

ykrt/try_passes.py
```diff
@@ -0,0 +1,190 @@
+#!/usr/bin/env python3
+
+import argparse
```
ltratt, line 3: Conventionally it's nicer to merge imports (so long as they don't make the line too long). So `import argparse, math, os, shutil, sys, time` and then the `from` lines. Also try and sort everything alphabetically -- it helps maintenance over time.

ykrt/try_passes.py
```diff
@@ -0,0 +1,190 @@
...
+import os, sys
+from subprocess import Popen, PIPE
+from dataclasses import dataclass
+import math
+import shutil
+from multiprocessing import Process, Manager, Pool, Lock
+import time
+
+# Stages in an LTO pipeline where optimisation passes can happen.
+STAGES = "pre_link", "link_time"
```
ltratt, line 13: This isn't valid Python syntax is it? Or is it and I just don't know it?!

ykrt/try_passes.py
```diff
@@ -0,0 +1,190 @@
...
+CWD = None
```
ltratt, line 14: Whoah, global mutable variables! Except in very limited circumstances, we can't do this -- it will cause us all sorts of problems.

ykrt/try_passes.py
```diff
@@ -0,0 +1,190 @@
...
+        # search?
+        p1 = Process(target=attempt2_inner, args=(logf, ok_passes, subset1, flagType,))
+        p1.start()
+        p2 = Process(target=attempt2_inner, args=(logf, ok_passes, subset2, flagType,))
+        p2.start()
+        p1.join()
+        p2.join()
+
+
+# Use a binary tree to attempt to speed things up by accepting whole groups of
```
ltratt, line 148: s/binary tree/binary chop/ ?

ykrt/try_passes.py
vext01: Parens are optional in tuples.

```
>>> X = 1, 2
>>> X
(1, 2)
```

(I think I wrote this line.)

Should `link_time` be `post-link` to be consistent with LLVM? We should add a comment to explain their confusing terminology.

ltratt: I did not know that!

ykrt/try_passes.py
```diff
@@ -0,0 +1,190 @@
...
+    def __str__(self):
+        pre_link = ",".join([ str(p) for p in self.pre_link ])
+        link_time = ",".join([ str(p) for p in self.link_time ])
+        return f"PipelineConfig(pre_link=[{pre_link}], link_time=[{link_time}])"
+
+# Define the pipeline in advance.
+PIPELINE_CONFIG_MAP = {
+    'lto': lambda ok_passes, try_passes: PipelineConfig([], ok_passes + try_passes),
+    'prelink': lambda ok_passes, try_passes: PipelineConfig(ok_passes + try_passes, []),
+    'O2': lambda ok_passes, try_passes: PipelineConfig(ok_passes + try_passes, ok_passes + try_passes)
```
vext01, line 41: I don't think the default O2 pipeline has identical pre-link and post-link passes, does it?

nmdis1999: No, we are already fetching fetching`prelink` (with `-prelink`)  and `post-link` (with `-lto`) pipeline  seperately at O2 level. But I kept this line from last implementaton, can remove it if we don't need this.

ykrt/try_passes.py
```diff
@@ -0,0 +1,190 @@
...
+    pre_link: list
+    link_time: list
+
+    def __str__(self):
+        pre_link = ",".join([ str(p) for p in self.pre_link ])
+        link_time = ",".join([ str(p) for p in self.link_time ])
+        return f"PipelineConfig(pre_link=[{pre_link}], link_time=[{link_time}])"
+
+# Define the pipeline in advance.
+PIPELINE_CONFIG_MAP = {
```
vext01, line 38: Would this be better as a function?

ykrt/try_passes.py
```diff
@@ -0,0 +1,190 @@
...
+        log(logf, str(pl) + ": OK\n")
+    else:
+        log(logf, str(pl) + " : FAILED\n")
+        print(" [FAIL]")
+    return p.returncode == 0
+
+def list_of_passes_to_str(passes):
+    return ",".join([str(p) for p in passes])
+
+def attempt2_inner(logf, ok_passes, try_passes, flagType):
```
vext01, line 113: We should rename this function? The name was just me hacking around.

ltratt:
> But the script is supposed to break list of passes in following manner (considering every subset fails):
[A, B, C, D, E] -> [A, B] [C, D, E] -> [A] [B] [C, D, E] -> [C] [D, E] -> [D] [E]

Got it. I think for an upcoming PR we should investigate a [genetic algorithm](https://en.wikipedia.org/wiki/Genetic_algorithm) which is likely to find a much longer pipeline (while still respecting the ordering of the main pipeline) relatively quickly and efficiently. Not for this PR though.

nmdis1999:
ready for review.

ykrt/try_passes.py
```diff
@@ -69,11 +64,14 @@ def get_all_passes(flagType):
     assert(p.returncode == 0)
 
     sout = sout.decode()
-    pass_descrs = list(set([x.strip() for x in sout.strip().split("\n")]))
-    
+    # order is getting messed up because of using set, since set is
```
vext01: I think this comment describes the past state of the code a little more than necessary.

It could either be removed or rephrased to something like:

> We are careful to maintain the ordering of the passes.

Or similar.

ykrt/try_passes.py
```diff
@@ -69,11 +64,14 @@ def get_all_passes(flagType):
...
+    # unordered. 
+    pass_descrs = [x.strip() for x in sout.strip().split("\n")]
+
     passes = []
     seen = set()
-
+    
+    #pass_descrs = pass_descrs[:3]
```
vext01: remove this commented code?

ykrt/try_passes.py
```diff
@@ -93,7 +91,7 @@ def test_pipeline(logf, pl):
     env["PRELINK_PASSES"] = ",".join([p.name for p in pl.pre_link])
     env["LINKTIME_PASSES"] = ",".join([p.name for p in pl.link_time])
 
-    p = Popen("try_repeat 10 sh test.sh 2>&1", cwd=CWD, shell=True,
```
vext01: is the `cwd` keyword needed? Does it not default to `pwd`?

ykrt/try_passes.py
```diff
@@ -130,19 +128,27 @@ def attempt2_inner(logf, ok_passes, try_passes, flagType):
...
-        subset1 = try_passes[:subset_len] 
-        subset2 = try_passes[subset_len:]
+        subset_len = len(try_passes)
+        subset1 = try_passes[:subset_len // 2] 
+        subset2 = try_passes[subset_len // 2:]
         # XXX: We assume that whatever subsets we accept from this splitting
         # work in combination, but that may not be true. Not sure what we'd do
         # if they didn't work in combination though... restart the whole
         # search?
-        p1 = Process(target=attempt2_inner, args=(logf, ok_passes, subset1, flagType,))
```
vext01: Is the XXX comment above still true? Based on our chat last week, I believe so?

ykrt/try_passes.py
```diff
@@ -29,19 +27,28 @@ def __str__(self):
...
-    'O2': lambda ok_passes, try_passes: PipelineConfig(ok_passes + try_passes, ok_passes + try_passes)
-}
-
-OPT_CMD_MAP = {
-    'lto': "opt -passes='lto<O2>' -print-pipeline-passes < /dev/null 2>/dev/null | tr ',' '\n'",
-    'prelink': "opt -passes='lto-pre-link<O2>' -print-pipeline-passes < /dev/null 2>/dev/null | tr ',' '\n'",
-    'O2': "opt -O2 -print-pipeline-passes < /dev/null 2>/dev/null | tr ',' '\n'"
-}
-# Command to run for flagType i.e -lto, -prelink, -O2`.
+def get_pipeline_config(flagType, ok_passes, try_passes):
```
vext01, line 29: This is a nitpick: the rest of the code base uses lower-case underscored variable names, but we introduce a camel case name here.

ykrt/try_passes.py
```diff
@@ -29,19 +27,28 @@ def __str__(self):
...
+    """Returns the appropriate pipeline configuration based on the flagType."""
+    
+    if flagType == 'lto':
```
vext01, line 32: If you use a bool for the flag type then you don't need to have an else case.

ykrt/try_passes.py
```diff
@@ -50,28 +57,19 @@ def log(logf, s):
...
     sout = sout.decode()
-    # order is getting messed up because of using set, since set is
-    # unordered. 
     pass_descrs = [x.strip() for x in sout.strip().split("\n")]
 
     passes = []
     seen = set()
     
-    #pass_descrs = pass_descrs[:3]
+    # pass_descrs = pass_descrs[-3:]
```
vext01: remove?

bin/yk-config
```diff
@@ -56,9 +51,22 @@ handle_arg() {
             # Without this, clang will slap `optnone` attributes on every
             # function, causing optimisations to skip them.
             OUTPUT="${OUTPUT} -Xclang -disable-O0-optnone"
-            # XXX hacked in quick.
-            if [ -n "${PRELINK_PASSES}" ]; then
-                OUTPUT="${OUTPUT} -mllvm --newpm-passes=${PRELINK_PASSES}"
+            if [ ! -z "$2" ]; then
```
vext01, line 54: I was expecting a top-level `--custom-prelink-pipeline` (or similarly named) which just overides a default pipeline.

In other words, there would be a top-level entry in the `case` construct for your new flag, which does nothing more than set a variable to override the pipeline. Then when the user provide `--cflags` the new pipeline is used.

ykrt/try_passes.py
nmdis1999: Yes, we still make some combinations.

ykrt/try_passes.py
nmdis1999: This is to get path to `yklua` earlier this path was hardcoded.

ykrt/try_passes.py
nmdis1999: I don't understand, currently we have two options `lto` or `prelink` if user pass any other flag we raise an error. Do we instead want to have a default flagType set if the user pass anything other than these two options?

vext01: I'd replace the string `flagType` with a bool `pre_link`. There are only two possibilities, so a bool covers both nicely.

bin/yk-config
vext01: > wouldn't we need to inform the user ro pass this flag before --cflags

Yes, that was what I was expecting.

The sub-flag thing might be OK, although it does appear to complicate the implementation quite a bit.

ltratt: I actually don't know what line 56 does, so I'm not scared, but I am bamboozled ;)

That said, just because I'm stupid doesn't mean line 56 (or a minor variant using `&&`) isn't the right thing to do.

vext01: It's a sub-parser for the `yk-config --cflags` switch, so you can have `yk-config --cflags -prelink BLAH` to specify a custom pipeline. The `--` check is to decide if the next switch is a top-level argument or a sub-argument to `--cflags`.

I had expected a top-level `--custom-prelink-pipeline` switch, which is simple to add, but does mean that it must come **before** `--cflags` on the command line (unless you jump through hoops).

ltratt: I'm happy with whatever the two of you decide is best.

nmdis1999: @vext01 I feel sub parser would be more intutive for user, and I can add a comment explaining what line 56 is doing,  please lmk what you think for this?

vext01: What still bugs me a little is that there is no precedent for double hyphen for top-level args and single hyphen for sub-args. As a user I'd find this surprising.

FWIW, `yk-config`'s interface was based on `pkg-config` which has no sub-args.

If we really want sub-args, I think all args should be double-hyphen. That should be doable?

But honestly, if I were doing this, I'd have done the simpler implementation. I'll let you choose: simpler implementation or consistent parsing for sub-args.

nmdis1999: So I think the implementation wouldn't be much simpler, since we still need the string of passes after --prelink or --postlink. What I can do is add double hyphen for now. And see if I can implement both (pre and postlink) as args easily.

nmdis1999: Fixed here: [d2e295e](https://github.com/ykjit/yk/pull/829/commits/d2e295eef2863e9d606282d86616c8dc98200873)

ykrt/try_passes.py
nmdis1999: fixed here: [9e17fdb](https://github.com/ykjit/yk/pull/829/commits/9e17fdb52b14807eefd50bfe0335386fecfe9121)

ykrt/try_passes.py
nmdis1999: Fixed here: 9e17fdb52b14807eefd50bfe0335386fecfe9121

ykrt/try_passes.py
nmdis1999: do we need to change anything here?

ykrt/try_passes.py
nmdis1999: removed the comment.

ykrt/try_passes.py
```diff
@@ -31,11 +31,8 @@ def get_pipeline_config(flag_type, ok_passes, try_passes):
     
     if flag_type == 'lto':
         return PipelineConfig([], ok_passes + try_passes)
-    elif flag_type == 'prelink':
-        return PipelineConfig(ok_passes + try_passes, [])
     else:
-        raise ValueError(f"Unknown flag_type: {flag_type}")
-
+        return PipelineConfig(ok_passes + try_passes, [])
```
vext01, line 35: We can still get rid of the stringly-typed programming here.

If we rename `flag_type` to `is_prelink` which is a bool, then it can reduce to:

```
if prelink:
  ...
else
  ...
```

(don't forget to update callsites)

bin/yk-config
```diff
@@ -21,10 +21,11 @@ OUTPUT=""
 usage() {
     echo "Generate C compiler flags for building against the yk JIT.\n"
     echo "Usage:"
-    echo "    yk-config <mode> <--cc|--cxx|--ar|--ranlib|--cppflags|--cflags|--ldflags>\n"
+    echo "    yk-config <mode> [--prelink <passes>] [--postlink <passes>] <--cc|--cxx|--ar|--ranlib|--cppflags|--cflags|--ldflags>\n"
```
vext01, line 24: Great!

One tiny request: can we rename `--prelink` and `--postlink` to `--prelink-pipeline` and `--postlink-pipeline`?

Cheers.

vext01:
Just two minor comments, then this can go in.

nmdis1999:
> Just two minor comments, then this can go in.

Addressed here: [0fcb724](https://github.com/ykjit/yk/pull/829/commits/0fcb72454ea9656560c73ac2cc81e7584d7fa0d9)

vext01:
Looks good.

Do you want to squash and resolve conflicts now?

nmdis1999:
> Looks good.
> 
> Do you want to squash and resolve conflicts now?

ok!

nmdis1999:
> Looks good.
> 
> Do you want to squash and resolve conflicts now?

Squashed all commits.

ltratt:
Something seems to have gone slightly wrong as this PR seems to have updated the ykllvm submodule -- but I don't think this PR is related to ykllvm, or maybe I'm wrong?!

nmdis1999:
> Something seems to have gone slightly wrong as this PR seems to have updated the ykllvm submodule -- but I don't think this PR is related to ykllvm, or maybe I'm wrong?!

We had ykllvm changes from edd's fork. We need these changes to append prelink flags in `--cflags` through `newpm-passes` similar to llvm's existing `lto-newpm-passes`

ltratt:
Ah, then we'll need to do a PR on ykllvm first so that those commits can safely be incorporated in the yk repository. @vext01 Want to raise a ykllvm PR?

vext01:
https://github.com/ykjit/ykllvm/pull/80 is the PR for the ykllvm stuff.

But there is an important thing to take into acccount: passing the empty string to the pipeline cusomisation switches does not specify an empty pipeline! It specifies the default pipeline.

So if @nmdis1999's script ever needs to specify an empty pipeline, you shouldn't pass the empty string. Pass `verify` instead!

ltratt:
> So if @nmdis1999's script ever needs to specify an empty pipeline, you shouldn't pass the empty string. Pass verify instead!

I think it would be best to `assert` that we never have an empty pipeline in this script: if we ever encounter that case we can work out what to do later. Might be worth adding a quick commit for that now.

@nmdis1999 You'll want to add a commit to this PR which changes the ykllvm commit to the latest in ykllvm (8245ec330de1452a0ded2f8dba94b6afe131d85a).

vext01:
> I think it would be best to assert that we never have an empty pipeline in this script

good idea!

ykrt/try_passes.py
nmdis1999, line 1: change `or` to `and`.

vext01:
Please squash.

[We've spoken about when to force push and when not]

PR #831, opened 2023-09-13
Make the `Compiler::compile` trait method return a `CompiledTrace`.

ltratt:
Previously we made the caller of `Compiler::compile` construct a `CompiledTrace` which meant we leaked minor internal details such as `Option<NamedTempFile>`.

This commit makes the `compile` method have the simpler signature:

```rust
pub trait Compiler: Send + Sync {
    /// Compile an [MappedTrace] into machine code.
    fn compile(&self, mt: Arc<MT>, irtrace: MappedTrace) -> Result<CompiledTrace, Box<dyn Error>>;
    ...
}
```


vext01:
cool.

bors r+

ltratt:
> Yay, more merge conflicts!

Ha! This should be fairly minor overall, hopefully!

PR #833, opened 2023-09-14
Draft new IR on-disk serialisation format.

vext01:
**This is a draft **

I've sketched a rough on-disk serialisation format for our new IR.

Opening for discussion. Does this look like we are headed in the right direction?

I don't expect to get this right first time.

(Note that there will need to be a separate in-memory representation for the IR)

(I started this in ykdocs. I'm not sure if that's the best place for it?)

### DIFF
--- a/docs/src/internals/ir.md
+++ b/docs/src/internals/ir.md
@@ -1,0 +1,155 @@
+# Yk Intermediate IR
+
+**This is a WIP document. We are currently in the process of switching from
+using LLVM IR to our own IR. This document describes the new IR**
+
+## On-disk serialisation format.
+
+(syntax loosely based on
+https://www.llvm.org/docs/StackMaps.html#stack-map-format)
+
+```
+header {
+    magic: usize    // hard-coded `0xedd5f00d`
+    version: usize  // format version (currently 0)
+}
+toc {
+    funcs {
+        num_funcs: usize
+        func_offs[num_funcs] {      // indices of this array server as function IDs.
+            func_data_off: usize    // byte offset into `func_data`.
+        }
+        func_data_size: usize       // in bytes.
+                                    // allows deserialiser to quickly skip over
+                                    // `func_data`.
+    }
+    globals {
+        num_globals: usize
+        global_offs[num_globals] {  // indices of this array serve as global IDs.
+            global_data_off: usize  // byte offset into `global_data`.
+        }
+        global_data_size: usize     // in bytes.
+                                    // allows deserialiser to quickly skip over
+                                    // `global_data`.
+    }
+    types {
+        num_types: usize
+        type_offs[num_types] {      // indices of this array serve as type IDs.
+            type_data_off: usize    // byte offset into `type_data`
+        }
+        type_data_size: usize       // in bytes.
+                                    // allows deserialiser to quickly skip over
+                                    // `type_data`.
+    }
+    constants {
+        num_constants: usize
+        type_offs[num_types] {
+            constant_data_off: usize    // byte offset into `constant_data`
+        }
+    }
+}
+func_data[num_funcs] {
+    name: null_term_str
+    func_sig {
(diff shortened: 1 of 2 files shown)


docs/src/internals/ir.md
```diff
@@ -0,0 +1,155 @@
...
+using LLVM IR to our own IR. This document describes the new IR**
+
+## On-disk serialisation format.
+
+(syntax loosely based on
+https://www.llvm.org/docs/StackMaps.html#stack-map-format)
+
+```
+header {
+    magic: usize    // hard-coded `0xedd5f00d`
```
ptersilie, line 13: What this `magic` thing used for?

docs/src/internals/ir.md
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+    func_index {
+        num_blocks: usize           // value of 0 means this is a declaration.
+        block_offs[num_blocks] {
+            block_data_off: usize   // byte offset into `block_data`.
+        }
+    }
+    block_data[num_blocks] {
+        num_instrs: u32
```
ptersilie, line 66: Did you mean `usize`? If not, why do we use a `u32` here when we used a `usize` everywhere else.

ptersilie:
I can't quite envision how this would be used in practise, e.g. to get all the necessary info to make a trace. Would it be too much to ask to have an example run on how we would go about doing that? That might also highlight any potential problems with the approach.

docs/src/internals/ir.md
vext01: Yes, it was supposed to be a usize.

I toyed with the idea of using u32s so that we could identify a basicblock using only one pointer-sized field (func-idx: u32 + bbidx: u32), but deemed it a) limiting, and b) too clever for a first stab.

vext01:
This isn't what you'd manipulate to build a trace. The format you see here is immutable description of the AOT code only. It would be converted into in-memory data-structures, which we can then manipulate.

Were you expecting something else?

vext01:
> example

Not until I've figured out what the in-memory representation is.

ltratt:
Dumb question: I was expecting we would use `[repr(C)]` structs so that ykllvm could write them and we could "deserialise" them simply by creating a pointer to them. I'm not sure if that's what this PR is suggesting or not?

vext01:
> Dumb question: I was expecting we would use [repr(C)] structs so that ykllvm could write them and we could "deserialise" them simply by creating a pointer to them. I'm not sure if that's what this PR is suggesting or not?

Yes, we should be able to cast certain byte offsets within this payload into structs.

(I have questions, but let's see if we are on the same page first)

ptersilie:
> This isn't what you'd manipulate to build a trace. The format you see here is immutable description of the AOT code only. It would be converted into in-memory data-structures, which we can then manipulate.
> 
> Were you expecting something else?

What I meant is, I would like to see this being used to find the correct functions/blocks of a trace. This is supposed to be used to get this info without loading in the whole thing, right? But I can't quite fathom how this layout will allow us to do that.

vext01:
OK, so suppose you want block 3 of function 2 (i've tried hard to make it so that all things are identified by integers).

If we haven't already looked something up function 2, then we will have to find it.  The TOC tells us the offsets of all the functions in the `func_data` field -- suppose func 2 is at  `func_data+123`. We then decode the index of this function at `func_data+123` to learn the offsets of all of the blocks within the function 2's `block_data` (but we do not eagerly decode the blocks).

To get block 3, we look into the function's index to find the offset of the block's information in `block_data` -- say it's `block_data+456`. Then we can find the instructions of that block at `block_data+456`.

(the instructions for a block are not indexed in the format I laid out, so we have to load those eagerly. I figured the index would get too large)

So you see, we only looked at the data we needed.

Makes sense?

vext01:
I'm going to raise a new PR for this.

PR #834, opened 2023-09-14
Rework hwtracer's error interface.

ltratt:
Previously hwtracer exposed an `HWTracerError` with a large number of variants. https://github.com/ykjit/yk/pull/832 slightly reduced that number, but still there was no way that a user of hwtracer could be expected to work out what to do with most of the variants.

This PR therefore makes a radical change to `HWTracerError`:

```rust
#[derive(Debug, Error)]
pub enum HWTracerError {
    /// hwtracer does not support the requested configuration. Adjust the configuration and retry.
    #[error("Configuration error: {0}")]
    ConfigError(String),
    /// hwtracer has failed in such a way that there is no point retrying the operation.
    #[error("Unrecoverable error: {0}")]
    Unrecoverable(String),
    /// hwtracer has encountered a temporary error: it is possible (though not guaranteed!) that
    /// retrying the same operation again will succeed.
    #[error("Temporary error: {0}")]
    Temporary(TemporaryErrorKind),
}
```

In essence, `HWTracerError` now bundles all the errors it can encounter into four categories: configuration errors; unrecoverable errors; and temporary errors. In general, you probably can't do much about the first two categories other than show an error to the user and then exit. However, the last category allows you to make intelligent decisions about whether "let's try that again" makes sense. `TemporaryErrorKind` is:

```rust
#[derive(Debug)]
pub enum TemporaryErrorKind {
    /// Memory allocation failed.
    CantAllocate,
    /// The trace buffer has overflowed. Either record a smaller trace or increase the size of the
    /// trace buffer.
    TraceBufferOverflow,
    /// The trace was interrupted.
    TraceInterrupted,
}
```

The first commit (https://github.com/ykjit/yk/commit/505829a4151b011e31bc2726368f092c9f0ed75f) does most of the hard work and exposed two things. First, we sometimes returned errors for things that are either unfinished code or the breaking of internal invariants: I've made those `panic`s because the only thing anyone can do if they encounter one of those is tell us, and then we can fix it. Second, the `NoMorePackets` variant was used solely for control flow within ykpt: in other words, we had the somewhat bizarre situation that we exposed an error variant to make code within hwtracer easier to structure! So I left a temporary hack in and then https://github.com/ykjit/yk/commit/d381f242c084069d8664f8818750f8eb9bc085ae (somewhat brutally) removes the temporary hack. I'm sure a nicer design is possible, but since this is now kept internal to ykpt, I'm not too worried about making it nicer quite yet.

### DIFF
--- a/hwtracer/Cargo.toml
+++ b/hwtracer/Cargo.toml
@@ -14,6 +14,7 @@ ykaddr = { path = "../ykaddr" }
 intervaltree = "0.2.7"
 byteorder = "1.4.3"
 leb128 = "0.2.5"
+thiserror = "1"
 
 [target.'cfg(target_arch = "x86_64")'.dependencies]
 iced-x86 = { version = "1.18.0", features = ["decoder"]}
--- a/hwtracer/src/errors.rs
+++ b/hwtracer/src/errors.rs
@@ -1,82 +1,40 @@
-use libc::{c_int, strerror};
-use std::{
-    error::Error,
-    ffi::CStr,
-    fmt::{self, Display, Formatter},
-};
+use std::fmt::{self, Display, Formatter};
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum HWTracerError {
+    /// hwtracer does not support the requested configuration. Adjust the configuration and retry.
+    #[error("Configuration error: {0}")]
+    ConfigError(String),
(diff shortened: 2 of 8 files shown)


ltratt:
bors try

vext01:
I like it.

bors r+

PR #837, opened 2023-09-19
Catch unused variables in .buildbot.sh.

ltratt:
This uses `cargo rustc` to run through the crates in yk, turning unused variable warnings into errors.


ltratt:
bors try

ltratt:
bors try

ltratt:
bors try

ltratt:
@ptersilie This will need squashing, assuming you're OK with it!

ltratt:
[Might also be worth a quick glance from @vext01 ]

ptersilie:
I was surprised that `module` wasn't needed anymore. But I remember now this was only required to get type sizes for `Direct` stackmap entries, which we currently don't see. It may the case tho, that once we are more conservative with what we put onto the shadowstack, we'll see `Direct` entries again and will have to put that code back.

Please squash.

ltratt:
Squashed.

PR #839, opened 2023-09-19
Change many traits/structs from `pub` to `pub(crate)`.

ltratt:
Doing so makes clearer what's part of the public API and what isn't. Most of these are mechanical with the exception of introducing a `compile_for_tc_tests` test-only function (which is necessary for us to make the `Compiler` trait be `pub(crate)`).


ptersilie:
Looks good.

bors r+

PR #842, opened 2023-09-20
Start work on the new code generator.

vext01:
**Companion PR ~~coming soon~~: https://github.com/ykjit/ykllvm/pull/81**

We are aiming to replace LLVM as a trace compiler because it's too slow to get code out the door.

This first unit of work drafts a deserialiser for a (massively incomplete) AOT IR format. For now we model the top-level IR module, functions, blocks and instructions.

Currently there are no types, no globals, no constants, only one opcode: Nop, and there are no instruction arguments yet. Future PRs will expand the IR as we incrementally add functionality.

The serialisation format is really dumb to start with and we load in the whole module eagerly. We may need to optimise that later (adding the ability to only load what we need).

I've added a (temporary) environment variable `YKD_NEW_CODEGEN` to enable running the tests with the new codegen. Obviously nothing works yet.

(We can still unit test the IR and new codegen as we go without flipping the whole test suite over to it)

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -1,0 +1,220 @@
+//! Yk's AOT IR deserialiser.
+//!
+//! This is a parser for the on-disk (in the ELF binary) IR format used to express the
+//! ahead-of-time compiled interpreter.
+//!
+//! The `Display` implementations convert the in-memory data structures into a human-readable
+//! textual format.
+
+use deku::prelude::*;
+use std::{
+    error::Error,
+    ffi::CStr,
+    fmt::{self, Display},
+};
+
+/// A magic number that all bytecode payloads begin with.
+const MAGIC: u32 = 0xedd5f00d;
+/// The version of the bytecode format.
+const FORMAT_VERSION: u32 = 0;
+
+fn deserialise_string(v: Vec<u8>) -> Result<String, DekuError> {
+    let err = Err(DekuError::Parse("failed to parse string".to_owned()));
+    match CStr::from_bytes_until_nul(v.as_slice()) {
+        Ok(c) => match c.to_str() {
+            Ok(s) => Ok(s.to_owned()),
+            Err(_) => err,
+        },
+        _ => err,
+    }
+}
+
+/// An instruction opcode.
+#[deku_derive(DekuRead)]
+#[derive(Debug)]
+#[deku(type = "u8")]
+pub(crate) enum Opcode {
+    Nop = 0,
+}
+
+/// A bytecode instruction.
+#[deku_derive(DekuRead)]
+#[derive(Debug)]
+pub(crate) struct Instruction {
+    opcode: Opcode,
+}
+
+impl Display for Instruction {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?}", self.opcode)
+    }
+}
+
+/// A basic block containing bytecode instructions.
+#[deku_derive(DekuRead)]
+#[derive(Debug)]
+pub(crate) struct Block {
+    #[deku(temp)]
+    num_instrs: usize,
+    #[deku(count = "num_instrs")]
+    instrs: Vec<Instruction>,
+}
+
+impl Display for Block {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        for i in &self.instrs {
+            writeln!(f, "    {}", i)?;
+        }
+        Ok(())
+    }
+}
+
+/// A function.
+#[deku_derive(DekuRead)]
+#[derive(Debug)]
+pub(crate) struct Function {
+    #[deku(until = "|v: &u8| *v == 0", map = "deserialise_string")]
+    name: String,
+    #[deku(temp)]
+    num_blocks: usize,
+    #[deku(count = "num_blocks")]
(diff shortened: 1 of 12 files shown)


ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -0,0 +1,201 @@
...
+    fmt::{self, Display},
+};
+
+/// A magic number that all bytecode payloads begin with.
+const MAGIC: u32 = 0xedd5f00d;
+/// The version of the bytecode format.
+const FORMAT_VERSION: u32 = 0;
+
+fn deserialise_string(v: Vec<u8>) -> Result<String, DekuError> {
+    match unsafe { CStr::from_ptr(v.as_ptr() as *const i8) }.to_str() {
```
ltratt, line 22: Dumb question: if the input is a `Vec`, is that `Vec` the length of the C string? If so, we can avoid the `unsafe` `CStr` stuff?

ykrt/src/compile/jitc_yk/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+use ykaddr::addr::symbol_vaddr;
+
+mod aot_ir;
+
+pub(crate) struct JITCYk;
+
+impl Compiler for JITCYk {
+    fn compile(
+        &self,
+        _irtrace: MappedTrace,
```
ltratt, line 16: The interface for this has changed in master. You'll need to add:

```
        sti: Option<SideTraceInfo>,
        hl: Arc<Mutex<HotLocation>>,
```

I suggest `panic`ing if `sti.is_some()` at first.

ykrt/src/compile/mod.rs
```diff
@@ -25,6 +28,15 @@ pub trait Compiler: Send + Sync {
 }
 
 pub fn default_compiler() -> Result<Arc<dyn Compiler>, Box<dyn Error>> {
+    #[cfg(jitc_yk)]
+    // Transitionary env var to turn on the new code generator.
+    //
+    // This will be removed once the transition away from LLVM as a trace compiler is complete.
+    if let Ok(v) = env::var("YKD_NEW_CODEGEN") {
```
ltratt, line 35: `if let Ok("1") = env::var(...)` should work I think?

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: That's right. I think we can just use `String::from_utf8`?

(We could actually zero copy the into a `&CStr`, but I thought it was premature optimisation at this point)

ykrt/src/compile/jitc_yk/mod.rs
vext01: ok

ykrt/src/compile/mod.rs
vext01: Yeah. Dunno why I did that :laughing:

ykrt/src/compile/jitc_yk/aot_ir.rs
ltratt: >  I think we can just use String::from_utf8

You might need to chop off the nul-byte at the end?

vext01: You're right. The input is determined by `#[deku(until = ...)]` which says:

> Note: The last value which matches the predicate is read

Which means we do take the null terminator too.

https://docs.rs/deku/latest/deku/attributes/#until

vext01: Out of curiosity, how would you have done it? Is there a better way than `from_utf8`?

ltratt: Maybe https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_until_nul ?

ptersilie:
Looks like a reasonable first step to me.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Fixed in [d78198d](https://github.com/ykjit/yk/pull/842/commits/d78198d945a3243269e004101972aa301ae38ac9)

ykrt/src/compile/jitc_yk/mod.rs
vext01: Merged upstream (and resolved conflicts) in [84e625e](https://github.com/ykjit/yk/pull/842/commits/84e625ec6d6b2878620fdf81cc684c8fd0ff8cf2), then added the proposed check in [6e47039](https://github.com/ykjit/yk/pull/842/commits/6e4703968ee53ca81dfc67a6f6fd8a94a62fe6da)

vext01:
Ready for re-review. Note that I also synced ykllvm to get the compiler-side requisites.

vext01:
Sorry last minute typo fix. Ready.

ykrt/src/compile/mod.rs
ltratt: Dose `if let Ok("1") = env::var("YKD_NEW_CODEGEN").map(|x| x.as_str())` work?

ltratt: OK, sorry for leading you up the garden path!

ltratt:
Please squash.

vext01:
splat.

PR #844, opened 2023-09-21
Tidy up the representation of guards

ltratt:
This PR tidies up the internal representation of guards a little bit. Each commit hopefully contains the necessary justification.

### DIFF
--- a/yktracec/src/jitmodbuilder.cc
+++ b/yktracec/src/jitmodbuilder.cc
@@ -54,9 +54,6 @@ const char *PromoteRecFnName = "__yk_promote";
 // The name prefix used for blocks that are branched to when a guard succeeds.
 #define GUARD_SUCCESS_BLOCK_NAME "guardsuccess"
 
-// Special ID for the final guard inside a side-trace.
-const size_t SIDETRACE_LAST_GUARD_ID = numeric_limits<uint64_t>::max();
-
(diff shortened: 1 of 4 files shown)


yktracec/src/jitmodbuilder.cc
```diff
@@ -1714,7 +1711,7 @@ class JITModBuilder {
                   // main interpreter.
                   BasicBlock *FailBB =
                       getGuardFailureBlock(BB, CurBBIdx, CPInstr, CurInstrIdx,
-                                           SIDETRACE_LAST_GUARD_ID);
+                                           GuardCount);
```
ptersilie, line 1714: I hope this won't bite us later. Although I can't currently think of a way that this guard may be created before other guards. It could be an issue when we start patching traces, but then we need to rework the guards anyway, and there probably won't be such a thing as the last guard anymore. This I believe this is fine for now.

ptersilie:
This looks fine, just one comment, but that requires no action on your part.

yktracec/src/jitmodbuilder.cc
ltratt: If we do need that, I reckon it'll be fairly obvious what's gone wrong.

ltratt:
Oops! Formatting fixed in ab5303d. OK to squash?

ptersilie:
Go ahead.

ltratt:
Squashed.

PR #847, opened 2023-09-26
Flesh out (very roughly) instruction encoding for Yk IR.

vext01:
**Compiler change: ~~coming soon~~ https://github.com/ykjit/ykllvm/pull/82**

This adds to Yk IR: a few opcodes, an instruction encoding, basic type encoding, basic constant encoding.

Types and constants are not stored inline inside the instruction (because types and constants are arbitrarily large, and we don't want multiple copies of large things hanging around). Instructions instead store indices to type and constant tables.

Following LLVM's precedent, where a value is identified by the instruction that generates it, our local variables are a pair: the basic block index and instruction index of the instruction that creates the value.

This unit of work has highlighted some problems with using `Display` as a human-readbale IR formatter:

 - Because constants are not stored inline, to print the actual constant we need access to the constant table.

 - Similarly for types.

 - When we print an instruction, if the instruction generates a value, we don't have context (from the parent function and block) to print the variable's name.

This should be fixable by using a "printer" struct (with state) to do the printing instead of `Display`.

Another solution would be to have references to the things we need in `self`, but it's not clear that would be easy.

Sample snippet of human-readable IR at this stage:

```
func main {
  bb0:
    $_ = alloca const[0]
    $_ = call const[1], const[2]
    $_ = call const[3], const[4], $0_0, const[5]
    store $0_1, const[6]
    $_ = alloca const[0]
    $_ = alloca const[0]
    $_ = getelementptr $0_1, const[4]
    $_ = getelementptr $0_1, const[7]
    $_ = getelementptr $0_1, const[8]
    store $0_8, const[6]
    ...
```

As you can see, constants and generated values cannot be printed properly yet.

I didn't extend our tests just yet, as I want to fix the formatting first. This will be my next PR.

### DIFF
--- a/docs/src/internals/ir.md
+++ b/docs/src/internals/ir.md
@@ -5,14 +5,22 @@ using LLVM IR to our own IR. This document describes the new IR**
 
 ## On-disk serialisation format.
 
+### Module encoding.
+
 ```
 Module {
     // header
-    magic: u32      // hard-coded `0xedd5f00d`
+    magic: u32 = 0xedd5f00d
     version: u32    // format version (currently 0)
     // functions
     num_funcs: usize
     funcs: Function[num_funcs]
+    // types
+    num_types: usize
+    types: Type[num_types]
+    // constants
+    num_constants: usize
+    constants: Constant[num_constants]
 }
 
 Function {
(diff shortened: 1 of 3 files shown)


docs/src/internals/ir.md
```diff
@@ -28,5 +36,45 @@ Block {
...
+```
+
+
+### Type encoding.
+
+`Type` is encoded as one of the following:
+
+IntegerType {
+    type_kind: u8 = 0
+    num_bits: u32
```
ltratt, line 57: Isn't `num_bits` implied by `type_kind`?

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -102,6 +204,36 @@ impl Display for Function {
     }
 }
 
+// A fixed-width two's compliment integer.
+//
+// Signedness is not specified. That is down to the operations performed upon the integer to
```
ltratt, line 209: Second sentence doesn't parse?

docs/src/internals/ir.md
vext01: `Type` can only express the size of a type in bytes, whereas this is a measure of bits. I may be able to fix that. Bear with me.

vext01: Hold on, I misunderstood.

type_kind 0 just says "it's an integer". It's up to every type kind to communicate its size. Is that OK, or should we have a top-level struct that indicates the size in a generic fashion, then a sub-enum for the type-specific parts?

Maybe we can kick this down the road a bit until we implement some more types?

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Quite right. Fixed in 351c5fbc

docs/src/internals/ir.md
ltratt: I see what you mean about allowing arbitrary bit sizes. OK for me to worry about more details later.

ltratt:
Please squash.

vext01:
splat

PR #848, opened 2023-09-27
Add docstring in pass finder script and moving the script to a new folder.

nmdis1999:
This PR:
- added a docstring at top of the pass-finder script.
- created and moved the script to a new folder (`pass-finder`) under `ykrt` .
- fixed typo in assert from PR #829

### DIFF
--- a/ykrt/pass_finder/try_passes.py
+++ b/ykrt/pass_finder/try_passes.py
@@ -1,6 +1,26 @@
 #!/usr/bin/env python3
+"""
+usage: python3 try_passes.py <-lto -prelink] -- path path/to/test/oracle
 
-import argparse, math, os, shutil, sys, time, queue 
+This script aims to identify an LLVM AOT pipeline that satisfies a test oracle.
+A typical test oracle would be a script that builds an interpreter and runs a
+test suite. This script is required because some passes don't yet work with Yk.
+
+The user specifies at least one pipeline (i.e. -prelink or -lto). The algorithm
+then starts with the default clang -O2 pipeline which is sent to the test
+oracle for evaluation. Each time the oracle is invoked it is passed a list of
+prelink and postlink passes in the environment variable (PRELINK_PASSES and
+POSTLINK_PASSES). The oracle must ensure that these pipelines are used when
+compiling the interpreter (usually by passing the pipeline to the
+--pre-link-pipeline and --post-link-pipeline arguments of yk-config). If the
+oracle accepts the pipeline (it returns zero), then the passes are appended to
+an "accept list". If the oracle rejects the pipeline (it returns non-zero),
+then the algorithm divides the failing pipeline into two parts and tests the
+two halves independently (in parallel). The search continues recursively until
+the search space is exhausted. Upon termination the final accept list is
+printed.
+"""
+import argparse, os, shutil, sys, time, queue 
 from dataclasses import dataclass
 from multiprocessing import Manager, Process, Queue, Value
 from subprocess import PIPE, Popen
@@ -77,7 +97,7 @@ def test_pipeline(logf, pl):
     log(logf, "\n\n" + str(pl) + "\n")  
 
     # Make sure we don't run empty strings in pipeline.
-    assert (len(pl.pre_link) != 0 and len(pl.link_time) != 0), "Both prelink and postlink passes cannot be empty!!!"
+    assert (len(pl.pre_link) != 0 or len(pl.link_time) != 0), "Both prelink and postlink passes cannot be empty!!!"
 
     env = os.environ
     env["PRELINK_PASSES"] = ",".join([p.name for p in pl.pre_link])
@@ -176,6 +196,10 @@ def main(logf, is_prelink):
     binary_split(logf, passes, is_prelink)
 
 if __name__ == "__main__":
+    if '--help' in sys.argv:
+        print(__doc__)
+        exit(0)
+
     parser = argparse.ArgumentParser()
 
     group = parser.add_mutually_exclusive_group(required=True)


ykrt/pass_finder/try_passes.py
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/env python3
-
-import argparse, math, os, shutil, sys, time, queue 
+"""
+This script is designed to identify the sequences of passes that successfully
+build the yklua interpreter and execute its tests. If a sequence fails, the
+algorithm divides it into two sections and retries. The order of the passes, as
+initially received, is preserved throughout the process.
+"""
+import argparse, os, shutil, sys, time, queue
```
nmdis1999, line 8: I removed import `math` as it wasn't being used anywhere anymore.

ykrt/pass_finder/try_passes.py
```diff
@@ -77,7 +82,7 @@ def test_pipeline(logf, pl):
     log(logf, "\n\n" + str(pl) + "\n")  
 
     # Make sure we don't run empty strings in pipeline.
-    assert (len(pl.pre_link) != 0 and len(pl.link_time) != 0), "Both prelink and postlink passes cannot be empty!!!"
+    assert (len(pl.pre_link) != 0 or len(pl.link_time) != 0), "Both prelink and postlink passes cannot be empty!!!"
```
nmdis1999, line 85: so we will need this to be an `or` since the assert will fails  and exit with error message only when both pre_link and link_time pipeline are parsed with empty string, otherwise if any one of the prelink or link_time have non-empty string the script should proceed normally.

ykrt/pass_finder/try_passes.py
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/env python3
-
-import argparse, math, os, shutil, sys, time, queue 
+"""
+This script is designed to identify the sequences of passes that successfully
+build the yklua interpreter and execute its tests. If a sequence fails, the
+algorithm divides it into two sections and retries. The order of the passes, as
+initially received, is preserved throughout the process.
+"""
```
vext01, line 7: We can tweak this to be more precise:

 - It's not specific to `ylklua`: you can use any oracle.
 - We need to say what happens after the sequence is chopped in two?
 - We need to say what the starting sequence is.

ykrt/pass_finder/try_passes.py
```diff
@@ -176,6 +181,10 @@ def main(logf, is_prelink):
     binary_split(logf, passes, is_prelink)
 
 if __name__ == "__main__":
+    if '--doc' in sys.argv:
```
vext01, line 184: `--help` is more conventional

vext01:
In the future, it would have been good to have had the three changes as separate commits, but don't worry for this one.

[We could also do with a way to specify the path to the test oracle. This could be a follow-up PR]

ykrt/pass_finder/try_passes.py
nmdis1999: Does following sounds better?

> This script aims to identify the sequences of passes that successfully builds and executes oracle's tests 10 times. The algorithm starts with a list of passes such as [pass1, pass2, pass3, pass4, pass5], if a sequence fails to build and/or successfully run the oracle's test, the algorithm divides the list into two subsets (i.e [pass1, pass2] and [pass3, pass4, pass5])  and retries to build and run oracle's tests 10 times . The algorithm parallely process the subsets and has a limit of 10 process to prevent fork bomb. The order of the passes, as initially received, is preserved throughout the process.

vext01: Much better. I'll propose some tweaks/restructuring:

> This script aims to identify an LLVM AOT pipeline that satisfies a test oracle. A typical test oracle would be a script that builds an interpreter and runs a test suite. This script is required because some passes don't yet work with Yk.
>
> The algorithm starts with the default clang `-O2` pipeline which is sent to the test oracle for evaluation. If the oracle accepts the pipeline (it returns zero), then the passes are appended to an "accept list".  If the oracle rejects the pipeline (it returns non-zero), then the algorithm divides the failing pipeline into two parts and tests the two halves independently (in parallel). The search continues recursively until the search space is exhausted. Upon termination the final accept list is printed.

Now the only things we don't talk about are:
 - The inputs to the oracle. IIRC it's an environment variable?
 - The pre-link and post-link pipelines, and their role in all of this.

Can you try an edit to include those details?

[I expect we will go back and forth a few more times on this]

nmdis1999: Updated, should I talk more about what changes are made in the pipeline (i.e changes in `cflags` and `ldflags`)?

> usage: python3 try_passes.py [-lto] [-prelink] -- path path/to/test/oracle
> This script aims to identify an LLVM AOT pipeline that satisfies a test oracle. A typical test oracle would be a script that builds an interpreter and runs a test suite. This script is required because some passes don't yet work with Yk.
> The algorithm expects the default pre and/or post-link clang `-O2` pipeline which is sent to the test oracle for evaluation. After receiving a list of prelink and postlink passes, the test oracle sets an environment variable to enable these passes, overriding the default pipeline. If the oracle accepts the pipeline (it returns zero), then the passes are appended to an "accept list".  If the oracle rejects the pipeline (it returns non-zero), then the algorithm divides the failing pipeline into two parts and tests the two halves independently (in parallel). The search continues recursively until the search space is exhausted. Upon termination the final accept list is printed.

vext01: > The algorithm expects the default pre and/or post-link clang -O2 pipeline

I don't know what you mean by "expects" here. Don't you query `opt` to get the starting -O2 pipeline?

> pre and/or post-link clang -O2 pipeline

A user will ask: when is it "and" and when is it"or"?

> the test oracle sets an environment variable to enable these passes, overriding the default pipeline

The script doesn't actually enable or disable any passes itself, it just sets an environment variable and it's up to the oracle to ensure that those passes are respected. I think that need to be clearer.

> should I talk more about what changes are made in the pipeline (i.e changes in cflags and ldflags)?

We might not need that here, because it's the job of the oracle.

Sorry if this is annoying, but you can see the level of precision we are striving for.

nmdis1999: > > The algorithm expects the default pre and/or post-link clang -O2 pipeline
> 
> I don't know what you mean by "expects" here. Don't you query `opt` to get the starting -O2 pipeline?

> 
> > pre and/or post-link clang -O2 pipeline

> A user will ask: when is it "and" and when is it"or"?

I meant the algorithm "expects" user to specify  ***at least*** one option (i.e `lto` or `prelink`) and then query `opt` to start the pipeline. If it doesn't receieve either one of the two options then the script will end with an error.

> > the test oracle sets an environment variable to enable these passes, overriding the default pipeline
> 
> The script doesn't actually enable or disable any passes itself, it just sets an environment variable and it's up to the oracle to ensure that those passes are respected. I think that need to be clearer.

Yes, so rephrasing the line to:
> After receiving a list of prelink and postlink passes, the test oracle sets an environment variable, overriding the default pipeline


> Sorry if this is annoying, but you can see the level of precision we are striving for.

no problem.

vext01: > I meant the algorithm "expects" user to specify

OK, so we can say: "the user specifies..."

Which reminds me, we should add a usage line. You know when the tool prints stuff like:

```
usage: the_tool [-v] this_arg that_arg
```

> "After receiving a list of prelink and postlink passes, the test oracle sets an environment variable, overriding the default pipeline"

Good. Even better(?):

> Each time the oracle is invoked it is passed a list of prelink and postlink passes in the environment variable `BLAH`. The oracle must ensure that these pipelines used when compiling the interpreter (usually by passing the pipeline to the `--pre-link-pipeline` and `--post-link-pipeline` arguments of `yk-config`.

(something like that? adjust as you see fit)

ykrt/pass_finder/try_passes.py
nmdis1999: addressed here: 934d4f63bef0ffc771a4a25a8ccef762f0ec3d28

ykrt/pass_finder/try_passes.py
vext01: Great. This is very close now.

> usage: python3 try_passes.py [-lto] [-prelink] -- path path/to/test/oracle

Are the `-lto` and `-prelink` arguments "one or the other, but one is required"? If so I'd format it like `<-lto|-prelink>`.

> path/to/test/oracle

I thought the path to the oracle was hard-coded for now?

Also, please wrap the docstring at 80 characters.

nmdis1999: No, the path to interepreter is expected, the test file name is hardcodes. So we are using `test.sh` but the directory path is given by user.

vext01: OK, ideally later we should make it so the user specifies the whole path to the script.

vext01:
Just need to wrap the doc comment now.

nmdis1999:
I had to force push as I had merged master. Should I squash the commits @vext01 ?

vext01:
Next time just push the merge commit, and we can eliminate it during squashing.

Yes, please squash.

nmdis1999:
> Next time just push the merge commit, and we can eliminate it during squashing.
> 
> Yes, please squash.

Ok, squashed!

PR #850, opened 2023-09-27
Serialise types later.

vext01:
We must serialise types last, as constants could introduce new types.

Extended test to cover this possibility.

Had to use `const`s instead of enums with unit variants for the type and constant kinds, as `MyEnum::MyVariant as u8` doesn't work inside `#[deku(id = "...")]`.


vext01:
OK for me to force push the rebase/resolution?

ltratt:
Please force push a rebase.

vext01:
pushed.

PR #852, opened 2023-09-28
Plumb in a more capable way of printing IR.

vext01:
The problem with using `Display` for this was that it made it hard to pass around the necessary state required to print the various elements of the IR. For example:

 - stringifying constants and types requires table lookups into the parent module.

 - stringifying generated values requires us to know about where the instruction that generates the value lives in the block and instruction structure of the IR.

This is the result of trying several different designs, none of which are perfect, but this seems to be the least offensive.

This change introduces a `YkDispaly` trait for converting in-memory IR to human-readable form. It works similarly to `Display`, but the parent module is always passed in, thus solving the issues outlined above.

To print the local variable generated by an instruction, we need to know the block and instruction index of the instruction. Rather than compute it on-demand, we compute *all* variable names the first time an instruction is stringifed and cache the names inside each instruction. Because the IR is conceptually immutable, we get away with this.

The other approaches I dismissed:

 - Having an explicit printer struct: annoying for the user.

 - Having Deku put references to the parent structures in each IR element so that we can always walk upwards toward the `AOTModule`: wastes a lot of memory, even if you never print the IR.

Note that I still haven't implemented printing for everything. For example, printing constants and types still prints placeholder strings.

Follow up PRs will fix these. I just want to have the interface reviewed first.

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -1,17 +1,10 @@
 //! Yk's AOT IR deserialiser.
 //!
 //! This is a parser for the on-disk (in the ELF binary) IR format used to express the
-//! ahead-of-time compiled interpreter.
-//!
-//! The `Display` implementations convert the in-memory data structures into a human-readable
-//! textual format.
+//! (immutable) ahead-of-time compiled interpreter.
 
 use deku::prelude::*;
-use std::{
-    error::Error,
-    ffi::CStr,
-    fmt::{self, Display},
-};
+use std::{cell::RefCell, error::Error, ffi::CStr};
 
 /// A magic number that all bytecode payloads begin with.
 const MAGIC: u32 = 0xedd5f00d;
@@ -26,6 +19,26 @@ fn deserialise_string(v: Vec<u8>) -> Result<String, DekuError> {
             Err(_) => err,
         },
         _ => err,
+    }
+}
+
+/// A trait for converting in-memory data-structures into a human-readable textual format.
+///
+/// This is modelled on [`std::fmt::Display`], but a reference to the module is always passed down
+/// so that constructs that require lookups into the module's tables from stringification have
+/// access to them.
+///
+/// The way we implement this (returning a `String`) is inefficient, but it doesn't hugely matter,
+/// as the human-readable format is only provided as a debugging aid.
+pub(crate) trait IRDisplay {
+    /// Return a human-readable string.
+    fn to_str(&self, m: &AOTModule) -> String;
+
+    /// Print myself to stderr in human-readable form.
+    ///
+    /// This is provided as a debugging convenience.
+    fn dump(&self, m: &AOTModule) {
+        eprintln!("{}", self.to_str(m));
     }
 }
 
@@ -47,9 +60,9 @@ pub(crate) enum Opcode {
     Unimplemented = 255,
 }
 
-impl Display for Opcode {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", format!("{:?}", self).to_lowercase())
+impl IRDisplay for Opcode {
+    fn to_str(&self, _m: &AOTModule) -> String {
+        format!("{:?}", self).to_lowercase()
     }
 }
 
@@ -75,12 +88,9 @@ pub(crate) struct ConstantOperand {
     constant_idx: usize,
 }
 
-impl Display for ConstantOperand {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        // FIXME: print the constant properly.
-        // This requires access to the constant table which we don't have here, so we will have to
-        // re-architect printing.
-        write!(f, "const[{}]", self.constant_idx)
+impl IRDisplay for ConstantOperand {
+    fn to_str(&self, m: &AOTModule) -> String {
+        format!("{}", m.consts[self.constant_idx].to_str(m))
     }
 }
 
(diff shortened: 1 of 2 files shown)


vext01:
Gah, typo in commit message: `YkDispaly`

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -3,15 +3,10 @@
 //! This is a parser for the on-disk (in the ELF binary) IR format used to express the
 //! ahead-of-time compiled interpreter.
 //!
-//! The `Display` implementations convert the in-memory data structures into a human-readable
-//! textual format.
+//! Once decoded, the IR is immutable. Any mutations after the fact are UB.
```
ltratt, line 6: > Any mutations after the fact are UB.

Why do we allow mutations if it's UB? Why don't we hide anything which enables mutation?

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -29,6 +24,27 @@ fn deserialise_string(v: Vec<u8>) -> Result<String, DekuError> {
...
 }
 
+/// A trait for converting in-memory data-structures into a human-readable textual format.
+///
+/// This is modelled on [`std::fmt::Display`], but a reference to the module is always passed down
+/// so that constructs that require lookups into the module's tables from stringification have
+/// access to them.
+///
+/// The way we implement this (returning a `String`) is inefficient, but it doesn't hugely matter,
+/// as the human-readable format is only provided as a debugging aid, and isn't used in
```
ltratt, line 34: Remove ", and isn't used in 'production'" IMHO.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -29,6 +24,27 @@ fn deserialise_string(v: Vec<u8>) -> Result<String, DekuError> {
...
+/// "production".
+pub(crate) trait YkDisplay {
```
ltratt, line 36: I think this should be called "IRDisplay".

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -268,26 +313,46 @@ pub(crate) struct AOTModule {
...
+                for (inst_idx, inst) in bb.instrs.iter().enumerate() {
+                    if inst.opcode.generates_value() {
+                        *inst.name.borrow_mut() = Some(format!("{}_{}", bb_idx, inst_idx));
+                    }
+                }
+            }
+        }
+    }
+
+    // Note that we don't implement `YkDisplay` on the top-level `AOTModule`. Imagine it, it would
```
ltratt, line 330: This comment seems to be floating? Is it a doc comment or ...?

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Once we've passed it outside of this module it's opaque and immutable.

It's more a not for developers working within this crate, who will have free reign. Perhaps the module doc isn't the right place for my note?

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: It's a doc for the two following two methods. I didn't really know how to do it...

ykrt/src/compile/jitc_yk/aot_ir.rs
ltratt: >Perhaps the module doc isn't the right place for my note?

Agreed: it shouldn't be in the externally facing docs.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -29,6 +24,27 @@ fn deserialise_string(v: Vec<u8>) -> Result<String, DekuError> {
...
+///
+/// This is modelled on [`std::fmt::Display`], but a reference to the module is always passed down
+/// so that constructs that require lookups into the module's tables from stringification have
+/// access to them.
+///
+/// The way we implement this (returning a `String`) is inefficient, but it doesn't hugely matter,
+/// as the human-readable format is only provided as a debugging aid, and isn't used in
+/// "production".
+pub(crate) trait YkDisplay {
+    /// Compute a human-readable string of myself.
```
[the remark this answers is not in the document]
ltratt: I would just say "Return a human-readable string" myself.

ykrt/src/compile/jitc_yk/aot_ir.rs
ltratt: I totally failed to understand that's what it's referring to...

vext01:
Aside from those niggles, is the general approach OK do you think?

ltratt:
Fine with me!

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: I just killed the comment altogether (in 20b53e05b6db8e1a775d99eb084c9be36505016a). I hope it should be obvious, and we can't explain every little implementation detail.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Fixed in [5c20aed](https://github.com/ykjit/yk/pull/852/commits/5c20aed1293e5154a4f76c1162eae0fee68fbb1c)

vext01:
Ready for re-review.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Fixed in 20b53e05b6db8e1a775d99eb084c9be36505016a.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Fixed in 20b53e05b6db8e1a775d99eb084c9be36505016a

ykrt/src/compile/jitc_yk/aot_ir.rs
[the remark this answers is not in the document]
vext01: Fixed in 20b53e05b6db8e1a775d99eb084c9be36505016a

ltratt:
Please squash (updating the commit message as appropriate).

vext01:
splat.

PR #853, opened 2023-09-29
Improve printing of unimplemented things.

vext01:
Incremental changes to the way we print unimplemented things. See commits.

Compiler change ~~coming soon~~: https://github.com/ykjit/ykllvm/pull/84 .

Probably should sync the submodule once that's in.

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -44,7 +44,7 @@ pub(crate) trait IRDisplay {
 
 /// An instruction opcode.
 #[deku_derive(DekuRead)]
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
 #[deku(type = "u8")]
 pub(crate) enum Opcode {
     Nop = 0,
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -147,8 +161,24 @@ pub(crate) struct Instruction {
     name: RefCell<Option<String>>,
 }
 
+impl Instruction {
+    fn get_type<'a>(&self, m: &'a AOTModule) -> &'a Type {
+        &m.types[self.type_index]
```
ltratt, line 166: There's a clear dependency between `&self` and `&AOTModule`: if you `get_type` on the wrong module, this will do weird things. That should, at least, be documented IMHO. Personally, I would probably put `type(&self, instr: Instruction)` on `AOTModule` because it then becomes more obvious that you really need to think carefully about which `Instruction`s came from which `AOTModule`.

vext01: [ccc5797](https://github.com/ykjit/yk/pull/853/commits/ccc5797d7a2c66f524a1a367f1f33b0cdbdb7e41) does as you suggested,  then [bbe01b8](https://github.com/ykjit/yk/pull/853/commits/bbe01b88694d09d5e1f9f4382d24088b4195c986) does the same to another part of the IR.

I now wonder if we should have a `Typed` trait so that the AOT module doesn't need one method per kind of typed thing that you might want to get the type of. Then again, we've only needed to ask the type of those two things so far...

Thoughts? Shall we kick it down the road a bit?

ltratt: Yeah, probably worth punting for now.

ltratt:
@vext01 Can I suggest you run clippy on this PR, and fix any suggestions it has for this code? If there are no such suggestions, please squash.

vext01:
I've added a docstring that I missed earlier. If that's OK with you, I'll clippy this.

ltratt:
If clippy finds anything, please push a commit. Otherwise please squash.

ltratt:
Please squash.

vext01:
Do you mind if I also sync ykllvm here? It wouldn't cause failures not to (because woefully, the deserialiser never consumes any IR from ykllvm in CI), but we probably should.

It bothers me that we are not testing the serialiser... Maybe we need to look into our "run this test only on this codegen" thing we discussed in-person last week?

ltratt:
> Do you mind if I also sync ykllvm here? It wouldn't cause failures not to (because woefully, the deserialiser never consumes any IR from ykllvm in CI), but we probably should.

I think we should do that in another PR and, as I think you're suggesting, actually test something (even if incredibly minimally) that results from ykllvm!

vext01:
Even just derserialising the IR of "hello world" will cover a ton of the deserialiser, so I agree that we should do that soon!

So shall i sync ykllvm now, or later?

ltratt:
>  So shall i sync ykllvm now, or later?

In another PR where we actually consume some IR.

vext01:
Splatted down into one commit.

PR #854, opened 2023-10-02
Pass-finder scripts supports argument, moved yklua test suite to same folder as pass-finder script.

nmdis1999:
This PR addresses two changes:
1) Earlier pass-finder script stripped off arguments from passes, with changes from this commit we will now be able to use passes with arguments.
2) Moved yklua test suite to same folder as pass-finder script.

### DIFF
--- a/ykrt/pass_finder/try_passes.py
+++ b/ykrt/pass_finder/try_passes.py
@@ -30,9 +30,6 @@ STAGES = "pre_link", "link_time"
 @dataclass
 class Pass:
     name: str
-    # FIXME: implement support for pass parameters?
-    #args: list
-
     def __str__(self):
         return self.name
 
@@ -82,10 +79,10 @@ def get_all_passes(is_prelink):
     seen = set()
      
     for descr in pass_descrs:
-        parts = descr.split("<")
-        if parts[0] not in seen:
-            seen.add(parts[0])
-            passes.append(Pass(parts[0]))
+        if descr not in seen:
+            seen.add(descr)
+            passes.append(Pass(descr))
+    
 
     print(f"Found {len(passes)} passes")    
     return passes
--- a/ykrt/pass_finder/yklua_tests.sh
+++ b/ykrt/pass_finder/yklua_tests.sh
@@ -1,0 +1,32 @@
+#!/bin/sh
+
+set -e
+
+MODE=release
+
+if [ ! -z $PRELINK_PASSES ]; then
+  echo "yk-here"
+  yk-config ${MODE} --prelink-pipeline "${PRELINK_PASSES}" --cflags 
+else
+  yk-config ${MODE} --cflags
+fi
+
+if [ ! -z $LINKTIME_PASSES ]; then
+  yk-config ${MODE} --postlink-pipeline "${LINKTIME_PASSES}" --ldflags
+else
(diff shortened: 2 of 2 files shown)


ykrt/try_passes.py
```diff
@@ -62,10 +59,10 @@ def get_all_passes(is_prelink):
...
      
     for descr in pass_descrs:
-        parts = descr.split("<")
-        if parts[0] not in seen:
-            seen.add(parts[0])
-            passes.append(Pass(parts[0]))
+        if descr not in seen:
+            seen.add(descr)
+            passes.append(Pass(descr))
+
```
vext01, line 65: So we simply pass through the default arguments to each pass?

It wasn't what I expected, but it's probably a decent first step.

(I thought that we might try different sets of arguments, but let's not do that right now)

ykrt/try_passes.py
```diff
@@ -83,7 +80,7 @@ def test_pipeline(logf, pl):
     env["PRELINK_PASSES"] = ",".join([p.name for p in pl.pre_link])
     env["LINKTIME_PASSES"] = ",".join([p.name for p in pl.link_time])
 
-    p = Popen("try_repeat 10 sh test.sh 2>&1", cwd=CWD, shell=True,
+    p = Popen("try_repeat 1 sh test-new.sh 2>&1", cwd=CWD, shell=True,
```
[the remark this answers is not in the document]
vext01: Also, why `test-new.sh`?

ykrt/yklua_tests.sh
```diff
@@ -0,0 +1,33 @@
...
+else
+  yk-config ${MODE} --ldflags
+fi
+
+make clean && make YK_BUILD_TYPE=release
+
+cd tests
+
+LUA=../src/lua
+# XXX we should repeat, say, 5 or 10 times to try to rule out non-deterministic
```
vext01, line 25: Since there is repetition in the script that calls this one, do we need repetition here?

ykrt/try_passes.py
nmdis1999: oh, i have a question: where do we get these new arguments from if not from default?

ykrt/try_passes.py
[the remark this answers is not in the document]
nmdis1999: I didn't mean to commit this change, it was purely from testing purpose. Will revert this.

nmdis1999:
Had to merge master before addressing your comment @vext01 , [45543f1](https://github.com/ykjit/yk/pull/854/commits/45543f1c18368b2e3b73be341362b0d333f77fe5) consist of the changes you asked and I moved `yklua` test script to `pass-finder` folder under `ykrt`.

vext01:
The change looks OK I think, but it was harder than necessary to verify because you force pushed.

Please only force push when the reviewer is OK with it.

If you needed to merge master, then you can stack a merge commit on the old commits and then add any extra changes after that. This way I can see exactly what changed without having the changes mixed in with earlier changes.

Anyway, this one is now looking good. Please squash your commits.

nmdis1999:
Squashed [b7e6f9a](https://github.com/ykjit/yk/pull/854/commits/b7e6f9acdfe0257409f84eb66056bcc3cc5f4e06).

PR #857, opened 2023-10-02
Update dependencies.

ltratt:
Some of these are quite old, and are rather more specific than seems useful: `x.y.z` only really makes sense when we know for sure that we need a specific `.z` or `.y.z` version, when generally we don't really know. So if `.y` (or, when appropriate, `.x`) has increased, I've dropped the suffix(es), for simplicity's sake.

### DIFF
--- a/hwtracer/Cargo.toml
+++ b/hwtracer/Cargo.toml
@@ -7,9 +7,9 @@ license = "Apache-2.0 OR MIT"
 
 [dependencies]
 libc = "0.2.148"
-strum = { version = "0.25", features = ["derive", "strum_macros"] }
+strum = { version = "0.25", features = ["derive"] }
 strum_macros = "0.25"
-deku = { version = "0.16.0", features = ["std"] }
+deku = "0.16.0"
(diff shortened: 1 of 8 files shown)


vext01:
We've been here before.

The reason [we decided to use `x.y.z`](https://github.com/ykjit/yk/pull/210#issuecomment-760119522) is because the standard tools (`cargo update` and `cargo upgrade`) emit the full version numbers, and having to edit the version numbers after the fact is annoying and not necessary.

I think we should use the full version numbers. Semver is clear about the meaning of them, even if the `z` can be redundant.

ltratt:
I think this is something upon which reasonable people can differ. I did this upgrade by hand because I wanted to check each dependency individually. If someone uses `cargo upgrade` and gets `x.y.z` next time, I won't complain.

vext01:
> If someone uses cargo upgrade and gets x.y.z next time, I won't complain.

This is a good compromise. Thanks.

ltratt:
@vext01 I think this one will need kicking again.

PR #858, opened 2023-10-03
Add the void type.

vext01:
See commits.

Required compiler change ~~coming soon~~: https://github.com/ykjit/ykllvm/pull/85

### DIFF
--- a/docs/src/internals/ir.md
+++ b/docs/src/internals/ir.md
@@ -53,8 +53,12 @@ Constant {
 
 `Type` is encoded as one of the following:
 
-IntegerType {
+VoidType {
     type_kind: u8 = 0
+}
+
+IntegerType {
+    type_kind: u8 = 1
     num_bits: u32
 }
 
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -273,7 +273,8 @@ impl IRDisplay for IntegerType {
     }
 }
 
-const TYKIND_INTEGER: u8 = 0;
+const TYKIND_VOID: u8 = 0;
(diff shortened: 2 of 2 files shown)


ltratt:
> if the instruction's type is not void, it makes a value.

What does "makes a value" mean? "Produces a value" or ... ? And what do we mean by "value"? It would be good if we can define these terms here for ourselves so that we're not dependent on LLVM and its (sometimes quirky, often changing...) terminology.

vext01:
By "makes a value" I mean the instruction returns something, and thus defines a new local variable.

We can define stuff if we like, but right now it's all the same as LLVM. I'm also not 100% sure I'm ready to write that sort of stuff down until we've gotten a little further in the design. Can we delay it a bit?

ltratt:
> By "makes a value" I mean the instruction returns something, and thus defines a new local variable.

Are we saying that it literally defines a variable (as, I believe, LLVM does) or that its result must be assigned to a variable?

> We can define stuff if we like, but right now it's all the same as LLVM. I'm also not 100% sure I'm ready to write that sort of stuff down until we've gotten a little further in the design. Can we delay it a bit?

OK.

vext01:
> Are we saying that it literally defines a variable (as, I believe, LLVM does) 

Correct.

In LLVM IR, any instruction starting `%blah =` defines a variable, and the way you'd check that using the LLVM API is to ask if the instruction is of type void. I'm doing exactly the same here for now.

ltratt:
OK, let's worry about the design of this when we've got some more experience.

bors r+

vext01:
(we also need the compiler change to be merged first, and then we have to sync it here)

vext01:
synced ykllvm. ok to squash?

ltratt:
Please squash.

vext01:
splat.

PR #859, opened 2023-10-03
Fix memory leak due to `CompiledTrace` not dropping.

ptersilie:
We can test for leaks by enabled yk stats which are only printed when `MT` drops, which doesn't happen if we are leaking `CompiledTrace`s.

Also, while we are here make it very clear that we can never return from executing a side-trace.

### DIFF
--- a/ykrt/src/deopt.rs
+++ b/ykrt/src/deopt.rs
@@ -102,7 +102,9 @@ pub(crate) struct NewFramesInfo {
 compile_error!("__llvm_deoptimize() not yet implemented for this platform");
 
 /// The `__llvm__deoptimize()` function required by `llvm.experimental.deoptimize` intrinsic, that
-/// is called during a guard failure.
+/// is called during a guard failure. Forwards all its arguments to `__ykrt_deopt` which does the
+/// actual deoptimisation. See that function's description for more details. Note, that `ctr` will
+/// be cast into an `Arc` inside `__ykrt_deopt` and then dropped.
 #[cfg(target_arch = "x86_64")]
 #[naked]
 #[no_mangle]
@@ -265,7 +267,9 @@ extern "C" fn ts_reconstruct(ctx: *mut c_void, _module: LLVMModuleRef) -> LLVMEr
 ///
 /// The arguments are as follows:
 ///
-///   * `ctr`: The [CompiledTrace].
+///   * `ctr`: An `Arc<CompiledTrace>` that has been cast to a `*const CompiledTrace`. Will be cast
+///      back into an `Arc` by this function and then dropped before deoptimisation or execution of
+///      a side-trace.
 ///   * `frameaddr`: Address of the control point's frame.
 ///   * `aotvals`: Struct describing the location of the AOT live variables.
 ///   * `actframes`: Address and size of vector holding active AOT frame information needed to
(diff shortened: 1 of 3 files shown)


ptersilie:
@ltratt I could need a second pair of eyes on this. I believe this isn't sound (see `FIXME`), but I'm not entirely sure, and also don't quite know how to fix this.

ykrt/src/deopt.rs
```diff
@@ -338,8 +338,15 @@ unsafe extern "C" fn __ykrt_deopt(
...
                 _,
-                unsafe extern "C" fn(*mut c_void, *const CompiledTrace, *const c_void) -> (),
+                unsafe extern "C" fn(*mut c_void, *const CompiledTrace, *const c_void) -> !,
             >(st.entry());
+
+            // This `CompiledTrace` was cloned before passing it into this trace. Since the
+            // function `f` never returns, we need to make sure to drop `ct` here to avoid leaks.
+            // FIXME: If this is the last `Arc`, the `CompiledTrace` is freed at this point, and
+            // `st` will no longer be available when we run the side-trace below.
+            drop(ctr);
```
ltratt, line 348: I'm confident this is safe in the sense of just `ctr`, because Rust's type system has proven it safe (plus I checked manually, but i trust that less!).

However I think the `FIXME` is wrong: the `st` is a `Clone` of an `Arc` and stands alone. In other words, I don't think the `ctr` being dropped causes the `st` to be dropped.

ptersilie: Oh, right. Because `st` is an `Arc` it can't be dropped, which keeps the `CompiledTrace` alive, even if all it's `Arc`s have been dropped. So I assume as soon as `st` drops, the `CompiledTrace` drops too, but I'm not entirely sure how Rust knows when that happens.

ltratt: Rust doesn't know because of the `Arc::into_raw(st)` -- it's our job to turn it back into a non-raw `Arc` so that it can be dropped appropriately. I don't know if we do that anywhere though?

ptersilie: Ah wait now I get it. Since `st` is an `Arc` it can exist even if `CompiledTrace` is freed!

ptersilie: I will remove the `FIXME`.

ltratt: Note that `st` is also `Arc<CompiledTrace>` and that we then `Arc::into_raw` it. So it still needs to be `Arc::from_raw`ed at some point.

ptersilie:
Ok to squash?

ltratt:
Please squash (but, separately, please consider https://github.com/ykjit/yk/pull/859#discussion_r1344326109 in case we need to fix that in another PR).

ykrt/src/deopt.rs
ptersilie: > Note that `st` is also `Arc<CompiledTrace>` and that we then `Arc::into_raw` it. So it still needs to be `Arc::from_raw`ed at some point.

It will be dropped during deopt either by the `from_raw` at the beginning of `__ykrt_deopt` or by the new `drop` added in this PR, since the only way to leave a (side)trace is via deopt.

ptersilie:
Tweaked the comment a bit.

ltratt:
These all look good, but we should add to the docstring comment for (at least) ` __llvm_deoptimize` so that someone reading that knows that `*const CompiledTrace` is an `Arc` and will be `from_raw`ed and `drop`ed. If there are any other functions that's relevant to, we should make a similar adjustment.

ptersilie:
Have another look.

ltratt:
Please squash.

ptersilie:
Squashed.

PR #861, opened 2023-10-04
Use LLVM sanitisers as part of our CI tests.

ltratt:


ltratt:
bors try

vext01:
Does the sanitiser swap out libc?

vext01:
Ah, I suppose this is related to: https://github.com/ykjit/yk/pull/860

ltratt:
> Does the sanitiser swap out libc?

It does -- but it's not the only thing that could do so. `LD_PRELOAD`, for example, would break this test.

ltratt:
bors try

ltratt:
bors try

ltratt:
@vext01 This is now ready for consideration -- and one thing you can validly ask is "is this a good idea at all?". The good news is that it only adds a minute or two to buildbot execution time.

vext01:
This is the prose I hinted at last time we spoke about sanitisers:

> Use of sanitizers together with build scripts and procedural macros is technically possible, but in almost all cases it would be best avoided

We use both...

https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html#build-scripts-and-procedural-macros

ltratt:
@vext01 The bors try + buildbot changes worked, so ...

.buildbot.sh
```diff
@@ -94,6 +94,33 @@ for i in $(seq 10); do
     YKD_NEW_CODEGEN=1 RUST_TEST_SHUFFLE=1 cargo test
 done
 
+# Test with LLVM sanitisers
+RUSTFLAGS="-Z sanitizer=address" cargo test --target x86_64-unknown-linux-gnu
+# The thread sanitiser has a number of false positives by default, so we need a
+# suppression file to avoid those overwhelming us. This does mean that we might
+# suppress some true positives, but there's little we can do about that.
+suppressions_path=`mktemp`
+cat << EOF > $suppressions_path
```
vext01, line 103: If we decide to use the sanitisers, I'd just put the suppression stuff into a file and import it rather than generating it each time.

.buildbot.sh
```diff
@@ -94,6 +94,33 @@ for i in $(seq 10); do
...
+# The tests::EXTRA_LINK closure leads to a race being reported, but that seems
+# to be due to an internal detail of `LazyLock` rather than our code.
+race:tests::EXTRA_LINK::
```
vext01, line 106: The EXTRA_LINK stuff invokes the compiler without any of `yk-config`s flags. If we tack on the stuff required for the sanitiser, does it silence some of the spurious errors?

.buildbot.sh
ltratt: The advantage of doing it like this is that test + suppression have to evolve together. [Whether that turns out to be a good idea in the medium term isn't clear to me yet.]

.buildbot.sh
ltratt: I don't understand the question.

vext01:
> @vext01 The bors try + buildbot changes worked, so ...

I'm more worried about the time we night spend sifting through false positives and false negatives.

We can try it  and see how far we get, but based on the the Rust dev's outlook, I don't have high hopes either.

ltratt:
> We can try it and see how far we get, but based on the the Rust dev's outlook, I don't have high hopes either.

Yes, I view this as experimental, and it might need to be backed out, but it does seem worth a try.

.buildbot.sh
vext01: Do you need to add sanitiser flags here: https://github.com/ykjit/yk/blob/ec757fefde04c6b8df62605168922a71b0da884c/tests/src/lib.rs#L37 ?

ltratt: I don't understand this question either!

vext01:
Go ahead with the review, and let's see how practical it is over the next month or so.

.buildbot.sh
vext01: You mean the locality is better?

ltratt: Yes, while we're in the experimental phase.

.buildbot.sh
vext01: Do you need to add snaitiser flags to invocations of the C compiler for the code that will be linked to the rust code?

ltratt:
After a chat elsewhere I think this is ready for squashing @vext01 ?

vext01:
Yes. Sanisiters are on probation for now :)

Squash away.

vext01:
May as well sync the branch too.

ltratt:
Squashed (and synced).

ltratt:
Another network timeout: can you retry?

vext01:
bors r+

(I can reproduce this locally. Fails non-deterministically)

vext01:
> assertion failed: taken

Not this again...

bors r+

PR #864, opened 2023-10-05
Optimise deoptimisation.

ptersilie:
This PR attempts to improve the performance of deoptimisation. Since deoptimisation can happen frequently we want it to be as cheap as possible so it doesn't negate the performance gain from running traces.

These changes manage to improve deoptimisation in debug mode yk by about 2x. While a good improvement it is still way to high: according to `YK_STATS` we are spending 34x more time in deopt than running traces (from `db.lua`: 0.007s JIT exec; 0.24 deopt). Unfortunately, in relase-mode yk the speedup gained by this PR is even lower, down to 5%. The reason for this appear to be side-traces which trigger a lot of deoptimisations for little gain as they don't run very long (this might change once we use patching to make side-traces loop back to the beginning of the trace). We can also see that side-traces can be nested deeply putting even more pressure on deoptimisation: executing a side-trace always requires a some (though not all of) deoptimisation to run first. The following screenshot shows some nested side-traces for illustration:

image

I will have a quick look if I can optimise the side-trace part of deoptimisation a little and then either add it as a commit here or push it in a separate PR. In the meantime this is open for discussion.

### DIFF
--- a/ykrt/src/deopt.rs
+++ b/ykrt/src/deopt.rs
@@ -306,8 +306,9 @@ unsafe extern "C" fn __ykrt_deopt(
         if let Some(st) = guard.getct() {
             let registers = Registers::from_ptr(rsp);
             let live_vars = ctr.smap().get(&retaddr.try_into().unwrap()).unwrap();
-            let mut ykctrlpvars = Vec::new();
-            for (_i, locs) in live_vars.iter().skip(1).enumerate() {
+            debug_assert!(live_vars.len() > 0);
+            let mut ykctrlpvars = vec![0; live_vars.len() - 1];
+            for (i, locs) in live_vars.iter().skip(1).enumerate() {
                 assert!(locs.len() == 1);
                 let l = locs.get(0).unwrap();
                 match l {
@@ -321,7 +322,7 @@ unsafe extern "C" fn __ykrt_deopt(
                         assert_eq!(*reg, 6);
                         let addr = unsafe { registers.get(*reg) as *mut u8 };
                         let addr = unsafe { addr.offset(isize::try_from(*off).unwrap()) };
-                        ykctrlpvars.push(addr as u64);
+                        ykctrlpvars[i] = addr as u64;
                     }
                     SMLocation::Indirect(reg, off, size) => {
                         let addr = unsafe { registers.get(*reg) as *mut u8 };
@@ -333,10 +334,10 @@ unsafe extern "C" fn __ykrt_deopt(
                             8 => unsafe { ptr::read::<u64>(addr as *mut u64) },
                             _ => unreachable!(),
                         };
-                        ykctrlpvars.push(v);
+                        ykctrlpvars[i] = v;
                     }
                     SMLocation::Constant(v) => {
-                        ykctrlpvars.push(*v as u64);
+                        ykctrlpvars[i] = *v as u64;
                     }
                     SMLocation::LargeConstant(_v) => {
                         todo!();
--- a/ykrt/src/deopt.rs
+++ b/ykrt/src/deopt.rs
@@ -296,6 +296,7 @@ unsafe extern "C" fn __ykrt_deopt(
     // an `Arc` so it will be dropped at the end of this function. Unless, we are going to execute
     // a side-trace. In that case this function will not return and we need to drop `ctr` manually.
     let ctr = Arc::from_raw(ctr);
+    (*ctr).mt().stats.timing_state(TimingState::Deopting);
 
     let guardid = GuardId(guardid);
 
@@ -347,6 +348,7 @@ unsafe extern "C" fn __ykrt_deopt(
                 _,
                 unsafe extern "C" fn(*mut c_void, *const CompiledTrace, *const c_void) -> !,
             >(st.entry());
+            (*ctr).mt().stats.timing_state(TimingState::JitExecuting);
 
             // This `Arc<CompiledTrace>` was cloned before executing this trace. Since the
             // side-trace will not return, we need to drop `ctr` manually here to avoid leaks.
@@ -366,7 +368,6 @@ unsafe extern "C" fn __ykrt_deopt(
 
     #[cfg(feature = "yk_jitstate_debug")]
     print_jit_state("deoptimise");
-    (*ctr).mt().stats.timing_state(TimingState::Deopting);
 
     // Copy arguments into a struct we can pass into the ThreadSafeModuleWithModuleDo function.
     let mut info = ReconstructInfo {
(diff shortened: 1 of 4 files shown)


ykrt/src/frame/mod.rs
```diff
@@ -35,6 +36,10 @@ pub static AOT_STACKMAPS: LazyLock<Vec<SMEntry>> = LazyLock::new(|| {
     StackMapParser::get_entries(slice)
 });
 
+pub fn load_aot_stackmaps() {
```
ltratt, line 63: `pub(crate)`

ykrt/src/frame/mod.rs
```diff
@@ -33,7 +46,18 @@ pub static AOT_STACKMAPS: LazyLock<Vec<SMEntry>> = LazyLock::new(|| {
...
-    StackMapParser::get_entries(slice)
+    let (entries, numrecs) = StackMapParser::get_entries(slice);
+    let mut pinfos = Vec::new();
+    let mut records = Vec::new();
+    records.resize_with(usize::try_from(numrecs).unwrap(), || (Record::empty(), 0));
+    for entry in entries {
+        pinfos.push(entry.pinfo);
+        for r in entry.records {
+            let idx = usize::try_from(r.id).unwrap();
+            records[idx] = (r, pinfos.len() - 1);
```
ltratt, line 57: Are we sure that `pinfos.len()` is bigger than 0 at this point?

ltratt:
Subject to those minor comments, I'm inclined to merge this PR ASAP, rather than keeping on building on top of it. Even if it's only a small improvement it *is* an improvement, so I see no reason to keep the branch alive unnecessarily long.

ptersilie:
If we disable side-traces, the deopt/JIT rate is slightly better, though still shouldn't take up more time than running traces:
```
"duration_deopting": 0.009,
"duration_jit_executing": 0.001,
```

ptersilie:
> Subject to those minor comments, I'm inclined to merge this PR ASAP, rather than keeping on building on top of it. Even if it's only a small improvement it _is_ an improvement, so I see no reason to keep the branch alive unnecessarily long.

If you can wait an hour, I will quickly try to check a quick idea. Otherwise, this is good to go and I don't think I have currently any other ideas to speed this up.

ltratt:
I am patient, so an hour is fine :)

ptersilie:
Sorry, got sidetracked, but this is ready now. Didn't manage to improve this by much, but every little bit helps. I think overall the problem is that side-traces don't loop back which means we deopt a lot more than we need to. So this is probably as much I can do for now.

ykrt/src/deopt.rs
```diff
@@ -306,8 +306,8 @@ unsafe extern "C" fn __ykrt_deopt(
         if let Some(st) = guard.getct() {
             let registers = Registers::from_ptr(rsp);
             let live_vars = ctr.smap().get(&retaddr.try_into().unwrap()).unwrap();
-            let mut ykctrlpvars = Vec::new();
-            for (_i, locs) in live_vars.iter().skip(1).enumerate() {
+            let mut ykctrlpvars = vec![0; live_vars.len() - 1];
```
ltratt, line 309: Do we know that `live_vars.len()` is > 0?

ptersilie: It would require there to be no live variables whatsoever, which, while highly unlikely, Im not confident enough about to categorically rule out. So I have two possible fixes for you:
```rust
            let mut ykctrlpvars = vec![0; live_vars.len()];
```
or
```rust
debug_assert!(live_vars.len() > 0);
```

ltratt: You could also use `checked_sub` instead of `-` and/or add a check in the trace compiler that traces always have 1 live variable or more.

[I'm actually not really sure why the `-1` is here though! If there are *n* live vars, why would we allocate space for *n - 1*?]

ltratt:
@ptersilie Please note there are other comments on this PR too.

ykrt/src/deopt.rs
ptersilie: Wouldn't a `checked_add` be less performant? The reason we do `n-1` is because we skip the first element of `live_vars` as it isn't needed. But we could just use `n` to size the vec. That extra element isn't going to kill us.

ltratt: If we skip the first variable then presumably we *do* know there are > 0 variables? What is that first variable?

ykrt/src/frame/mod.rs
ptersilie: I checked and the only way for it to be `0` is if there's not a single stackmap in the program. Which in turn would require the program to have no branches. But that would be a pretty silly program and since it can't have loops we can't trace anything so will never end up in deopt.

ltratt: OK so I suggest:

1. adding a check to the trace compiler that this doesn't happen (i.e. borking in the trace compiler if someone tries to compile such a trace),
2. adding a comment here referencing the check in the compiler.

Note: *maybe* we can even detect this AOT and not at trace creation time? I dunno.

ptersilie: Wait, I got an even better answer. We literally push to `pinfos` just before this loop. So it's guaranteed to always have at least one entry.

ltratt: Aha!

ykrt/src/deopt.rs
ptersilie: Unlike stackmaps generated by stackmap calls, stackmaps generated by the `llvm_deoptimise` intrinsic has an additional live variable at the beginning containing 3 values CC, Flags, and NumDeopts. These values are of no interest to us so we just skip them here. What I'm not certain about is whether these values still get written to the stackmap if there are no other live variables. So we should probably add a check here.

ltratt: A `debug_assert` is always a good thing in my book!

ykrt/src/frame/mod.rs
ptersilie: Fixed in [6ac1a43](https://github.com/ykjit/yk/pull/864/commits/6ac1a43dc7f2f7a64999aeaa54e50f88bf958052)

ykrt/src/deopt.rs
ptersilie: Fixed in [358141e](https://github.com/ykjit/yk/pull/864/commits/358141e0ae1cf1d84faddccd16cd5463d5028895)

ptersilie:
Think that's all comments addressed.

ltratt:
Please squash.

ltratt:
@ptersilie Huh, did you force push after the r+?

ptersilie:
No, I didn't. This is probably due to Github being very slow to push the commits to the PRs. They had some issues yesterday and I didn't check if the force push showed before I commented "Squashed". Just try again.

PR #870, opened 2023-10-10
Add an "IR lowering" test suite.

vext01:
For each test you feed in textual LLVM IR and it gets compiled with ykllvm's clang. We then extract and deserialise the Yk IR, and use lang_tester to match against it.

We do have to be a bit careful when writing tests. Even at -O0, clang optimises away some LLVM instructions if it deems them redundant. If in doubt, make instructions depend on dynamic properties (e.g. argc or argv).

Nonetheless, this is much less painful that manually throwing bytes into a vector, as we've been doing in unit tests.

### DIFF
--- a/tests/langtest_ir_lowering.rs
+++ b/tests/langtest_ir_lowering.rs
@@ -1,0 +1,67 @@
+//! A suite for testing lowerings of LLVM IR to Yk IR.
+
+use lang_tester::LangTester;
+use regex::Regex;
+use std::{env, fs::read_to_string, path::PathBuf, process::Command};
+use tempfile::TempDir;
+use ykbuild::ykllvm_bin;
+
+const COMMENT: &str = ";";
+
+fn main() {
+    println!("Running IR lowering tests...");
+
+    let tempdir = TempDir::new().unwrap();
+    LangTester::new()
+        .test_dir("ir_lowering")
+        .test_file_filter(|p| p.extension().as_ref().and_then(|p| p.to_str()) == Some("ll"))
+        .test_extract(move |p| {
+            read_to_string(p)
+                .unwrap()
+                .lines()
+                .skip_while(|l| !l.starts_with(COMMENT))
+                .take_while(|l| l.starts_with(COMMENT))
+                .map(|l| &l[COMMENT.len()..])
+                .collect::<Vec<_>>()
+                .join("\n")
+        })
+        .test_cmds(move |p| {
+            let mut exe = PathBuf::new();
+            exe.push(&tempdir);
+            exe.push(p.file_stem().unwrap());
+
+            // We don't use yk-config here, as we are testing one very specific functionality that
+            // requires only one special flag.
+            let mut compiler = Command::new(ykllvm_bin("clang"));
+            compiler.args(&[
+                "-flto",
+                "-fuse-ld=lld",
+                "-O0",
+                "-o",
+                exe.to_str().unwrap(),
+                "-Wl,-mllvm=--yk-embed-ir",
+                p.to_str().unwrap(),
+            ]);
+
+            let md = env::var("CARGO_MANIFEST_DIR").unwrap();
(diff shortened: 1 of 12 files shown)


tests/langtest_ir_lowering.rs
```diff
@@ -0,0 +1,80 @@
...
+};
+use tempfile::TempDir;
+use ykbuild::ykllvm_bin;
+
+const COMMENT: &str = ";";
+
+fn main() {
+    println!("Running IR lowering tests...");
+
+    let filter: fn(&Path) -> bool = |p| {
```
ltratt, line 19: Why break this out in an `fn` rather than a closure in `test_file_filter`? I think you can do something simpler like `test_file_filter(|p| p.extension() == Some("ll"))`?

vext01: It's because I copied the code from another suite and didn't do the best job of minimising the code :)

Will fix.

ltratt: Maybe fix both at the same time, if they're in the yk repo?

vext01: I think [aef79a4](https://github.com/ykjit/yk/pull/870/commits/aef79a41a54ebfb3bb1d87efaa8535c2c20f76c2) is right?

It's a bit miserable because `OsStr` doesn't coerce with `str` as I'd hoped.

ltratt: That looks much better to me! Is it the only instance of this in the repo?

vext01:
One more tweak in 28cbba8552b2 also.

tests/langtest_ir_lowering.rs
vext01: 4610e93bd4b2 attempts to apply the same idioms, but it's not as clear cut for `langtest_c.rs`. Let me know what you think.

ltratt:
Please squash as appropriate.

vext01:
splat.

vext01:
>  error: extra arguments to `rustc` can only be passed to one target, consider filtering

I'm not sure why that's started failing now...

vext01:
bors try

vext01:
@ltratt it appears that `cargo check --tests` only works so long as the crate doesn't build a binary.

This is why you had special-cased the `tests` crate before, and now this PR adds a binary to `ykrt`, so that's broken too now.

Could be a rust bug, but I've failed to make a minimal example of the problem :confused: 

What do you think we should do?

ltratt:
Disable it for the new binary too? Even if we can't run it *everywhere*, anywhere we can run it is surely better than disabling everything.

vext01:
bors try

vext01:
bors try

vext01:
bors try

vext01:
bors ping

vext01:
bors try-

vext01:
bors try

vext01:
bors try

vext01:
bors try

vext01:
OK to squash?

ltratt:
Please squash.

vext01:
splat.

vext01:
This failed because of an assertion failure in the PT decoder.

Not something I want to get into now. If it keeps coming up, I can investigate.

OK to retry?

ltratt:
@vext01 Please file an issue with an error so this isn't forgotten.

vext01:
That's a network error, please retry.

> error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-nightly.toml.sha256' to '/ci/.rustup/tmp/pqkwjkp2_is5o88w_file': failed to make network request: error sending request for url (https://static.rust-lang.org/dist/channel-rust-nightly.toml.sha256): operation timed out: operation timed out

vext01:
what's up with this?

bors ping

vext01:
>  failed to write to an output file. Disk full?

Investigating...

PR #877, opened 2023-10-13
PT: Don't attempt to track whether packet generation is enabled.

vext01:
PGE and PGD are two kinds of PT packets that mean "Packet Generation Enable" and "Packet Generation Disable". PT emits these as you enable and disable tracing.

The decoder tracks whether we are currently in a portion of the trace where packet generation is enabled. Currently we always use a fresh perf fd to collect a trace, so we know to start the decoder with `self.pge=false`.

But going forwards we want to re-use perf fds. In this scenario, when you collect a trace using a "pre-owned" perf fd, you can't generally know the state of the packet stream, or whether to start the decoder with `self.pge=false` or `self.pge=true`.

That sounds problematic, but what I've realised is that whether or not packet generation is enabled has zero consequence to the decoder, so there's no need to track it!

This change removes `self.pge` and all of its uses from the decoder. Lo and behold, all tests still pass.

### DIFF
--- a/hwtracer/src/pt/ykpt/mod.rs
+++ b/hwtracer/src/pt/ykpt/mod.rs
@@ -219,9 +219,6 @@ pub(crate) struct YkPTBlockIterator<'t> {
     tnts: VecDeque<bool>,
     /// The compressed return stack.
     comprets: CompressedReturns,
-    /// When true, packet generation is enabled (we've seen a `TIP.PGE` packet, but no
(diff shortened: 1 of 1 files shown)


ltratt:
Hmm, yes, this doesn't look good for ThreadSanitizer. I will have a look at this separately but for now let's hope another try gets this PR through CI.

bors r+

Issue #784, opened 2023-07-17
ykcpython not building


ltratt:
In the [yk_update branch in my ckcpython fork](https://github.com/ltratt/ykcpython/tree/min_yk_update) I have tried minimially updating ykcpython to a new yk. It now builds but seems to bork when linking the final executable

```
/home/ltratt/yk/target/debug/ykllvm/bin/clang -fuse-ld=lld -Wl,--mllvm=--embed-bitcode-final -Wl,--mllvm=--disable-branch-fold -Wl,--mllvm=--disable-block-placement -Wl,--mllvm=--disable-early-taildup -Wl,--mllvm=--disable-tail-duplicate -Wl,--mllvm=--yk-disable-tail-call-codegen -Wl,--mllvm=--yk-no-fallthrough -Wl,--mllvm=--yk-patch-control-point -Wl,--mllvm=--yk-insert-stackmaps -Wl,--mllvm=--yk-block-disambiguate -Wl,--mllvm=--yk-extended-llvmbbaddrmap-section -Wl,--mllvm=--yk-stackmap-spillreloads-fix -Wl,--mllvm=--yk-shadow-stack -Wl,--mllvm=--yk-stackmap-add-locs -Wl,--mllvm=--yk-optnone-after-ir-passes -Wl,--lto-newpm-passes=instcombine -Wl,--mllvm=--yk-alloc-llvmbc-section -Wl,--mllvm=--yk-alloc-llvmbbaddrmap-section -Wl,--lto-basic-block-sections=labels -Wl,--export-dynamic -Wl,--mllvm=--yk-linkage -L/home/ltratt/yk/bin/../target/debug/deps -Wl,-rpath=/home/ltratt/yk/bin/../target/debug/deps -Wl,-rpath=/home/ltratt/yk/target/debug/ykllvm/lib -Wl,--disable-new-dtags -Wl,--no-rosegment    -lykcapi -Xlinker -export-dynamic -o python Programs/python.o libpython3.11.a -lykcapi -lcrypt -lpthread -ldl  -lutil -lm   -lm 
ld.lld: /home/ltratt/yk/ykllvm/llvm/lib/IR/Type.cpp:584: llvm::StringRef llvm::StructType::getName() const: Assertion `!isLiteral() && "Literal structs never have names"' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
[...]

PR #878, opened 2023-10-15
Thread local shadowstack

Pavel-Durov:
Dependant: 
https://github.com/ykjit/ykllvm/pull/94

Related:
https://github.com/ykjit/yk/issues/784

Allocate and set the local thread shadow stack in `thread_inttercept` module.


Pavel-Durov:
I'm unsure how I can assert in tests that thread shadow stacks are in deed thread local. 
Maybe I can check the allocation address of local thread variables?

ykrt/src/thread_intercept.rs
```diff
@@ -1,11 +1,54 @@
...
+            panic!("Unable allocate stack!")
+        }
+        // Set shadowstack addr with new allocated stack
+        *(shadowstack_addr as *mut *mut c_void) = stack_addr;
+        let thread_routine: &ThreadRoutine = (arg as *mut ThreadRoutine)
+            .as_ref()
+            .expect("Thread routine function and args data.");
+        // Call original thread routine
+        let result = (thread_routine.start_routine)(thread_routine.args);
+        free(stack_addr);
```
Pavel-Durov, line 35: Is there a benefit of using POSIX functions for thread cleanup such as `pthread_cleanup_push` and `pthread_cleanup_pop`?
These functions are not part of rust [libc crate](https://docs.rs/libc/latest/libc/)

ykrt/src/thread_intercept.rs
```diff
@@ -1,11 +1,54 @@
...
+    pub start_routine: extern "C" fn(*mut c_void) -> *mut c_void,
+}
+
+extern "C" fn wrap_thread_routine(arg: *mut c_void) -> *mut c_void {
+    let str = CString::new("shadowstack_0").unwrap();
+    unsafe {
+        //  Obtain address of a shadowstack_0 symbol
+        let shadowstack_addr: *mut libc::c_void = dlsym(null_mut(), str.as_ptr() as *const i8);
+        if shadowstack_addr.is_null() {
+            panic!("Unable to find shadowstack address!")
```
ltratt, line 21: Nit-picking but in general we don't need "!" in error messages: the system exiting will induce sufficient additional adrenaline levels in the user :)

ykrt/src/thread_intercept.rs
ltratt: Can we solve this issue by calling the "normal" `pthread_create` function? If not, then I guess we have to emulate whatever `pthread_create` does... At least we would probably have to add a wrapper around `pthread_cleanup_push` that panics, so that if any code does use it, it will die in a clear and obvious way.

ykrt/src/thread_intercept.rs
```diff
@@ -1,11 +1,54 @@
...
+
+#[derive(Debug)]
+struct ThreadRoutine {
+    pub args: *mut c_void,
+    pub start_routine: extern "C" fn(*mut c_void) -> *mut c_void,
+}
+
+extern "C" fn wrap_thread_routine(arg: *mut c_void) -> *mut c_void {
+    let str = CString::new("shadowstack_0").unwrap();
+    unsafe {
```
ltratt, line 17: In general I'm nervous about big `unsafe` blocks like this because it lets us put lots of code in without thinking about whether it's safe or not. Ideally I prefer to use `unsafe` only on individual lines that *must* be unsafe.

ykrt/src/thread_intercept.rs
```diff
@@ -1,11 +1,54 @@
...
-    _: extern "C" fn(*mut c_void) -> *mut c_void,
-    _: *mut c_void,
+    thread: *mut pthread_t,
+    attr: *const pthread_attr_t,
+    start_routine: extern "C" fn(*mut c_void) -> *mut c_void,
+    args: *mut c_void,
 ) -> c_int {
-    todo!("Allocate and set thread-local ShadowStack instance. No support for threads yet!");
+    unsafe {
+        let routine_args = &ThreadRoutine {
```
ltratt, line 48: The `let` isn't unsafe so can go outside of the `safe` block.

However, I'm not totally certain if passing a pointer to C code to something on the "Rust stack" is defined or undefined behaviour. You're going to have to do some reading up on UB with the C FFI in Rust to work that out I'm afraid!

ptersilie:
Apart from what @ltratt has already commented, this looks good. However, we need a test for this, as currently we don't really know if it's working as expected. You should be able to test this by creating a thread that creates some local variables, and then check in the parent function that the local variables haven't changed. Since currently all threads share the same shadow stack, you should quite quickly find a situation where the threads mess up the variables on the stack.

ykrt/src/thread_intercept.rs
Pavel-Durov: Ok. I can break it into separate, individual lines :)

ykrt/src/thread_intercept.rs
Pavel-Durov: I am not sure I understand what you mean by "calling the normal pthread_create function"
I thought that's what I'm doing here - calling POSIX `pthread_create` function with a wrapper function that will call the original `pthread_create` `start_routine`

ykrt/src/thread_intercept.rs
Pavel-Durov: removed 👉 d9ceb2e6d80eb0ced85945c289872a910074ddf4

ykrt/src/thread_intercept.rs
Pavel-Durov: done here 👉 ced50ed5a4004fe77b616dedcf25c07bea088d4d

ykrt/src/thread_intercept.rs
Pavel-Durov: I think in both cases It is an undefined behaviour since `routine_args` might be freed by the time the created thread will run.

I switched to use [Box::into_raw](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.into_raw) and [Box::from_raw](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from_raw) here 👉 ced50ed5a4004fe77b616dedcf25c07bea088d4d

Pavel-Durov:
@ptersilie I tried to create local variables in the thread routine and assert in the parent function the values of local variables. 
But it seems to be working even when the shadow stack is not set to be thread-local in LLVM.
Here's my test: 216d5f01d965011ad8ca73020722adea2363b69b

ptersilie:
I reckon the threads run so fast that you get away with sharing the shadow stack since by the time the next thread is started and writes to the stack, the previous one has printed its output already.

What I would try is set a variable in the main method, then start a thread which also assigns some variables, then still in main sleep for a second and then check the output. The sleeping will make sure the thread will have overwritten the stack before you print.

tests/c/pthread_create.c
```diff
@@ -1,30 +1,42 @@
...
   yk_mt_hot_threshold_set(mt, 0);
   YkLocation loc = yk_location_new();
 
-  int i = 3;
-  while (i > 0) {
+  int MAX_THREADS = 3;
+  int thread_args[3] = {1, 2, 3};
+  pthread_t thread_handles[MAX_THREADS];
+
+  int a = 0;
```
ptersilie, line 27: Also, try changing this to `int a = argc`. I'm not sure what optimisations we're currently running on top of `O0`, but one of them might statically work out that `a` is always 0 and remove the `if (a != 0)` case below entirely.

ykrt/src/thread_intercept.rs
ltratt: I *think* I see how this works, but it's not totally clear. We need to document clearly when we're passing around raw pointers like this so that we know whose role it is to free them: that means the return value(s) and parameter(s) will need `///` documenting.

ykrt/src/thread_intercept.rs
ltratt: I don't think I understood this fully. What I should have said is: let's add a wrapper around `pthread_cleanup_push` that panics if anyone uses it. That way we'll at least know if we ever need to support it! [I don't mind hard errors which make it clear where we need to look for the solution, but I am always very scared if errors can get through silently!]

Pavel-Durov: I think I understand what you mean, I can definetly add a wrapper for `pthread_cleanup_push` and `pthread_cleanup_pop` functions.

But my question was -  Should we use these `pthread_cleanup_push/pop` functions for YK-created threads cleanup? In this PR context, the the cleanup is `free(stack_addr)`.
I guess there can be an case where the thread is cancelled and we end up not freeing the allocated stack.

ltratt: > But my question was - Should we use these pthread_cleanup_push/pop functions for YK-created threads cleanup?

Good question! I don't think we need to since our wrapper automatically means we can do the "last" cleanup ourselves?

ykrt/src/thread_intercept.rs
Pavel-Durov: Updated 👉  bce94d1be08829164adf5879374b55f5062ffa01

tests/c/pthread_create.c
Pavel-Durov: ok, will give it a go!

tests/c/pthread_create.c
```diff
@@ -1,30 +1,42 @@
...
     yk_mt_control_point(mt, &loc);
-    pthread_t thread;
-    pthread_create(&thread, NULL, thread_function, NULL);
-    pthread_join(thread, NULL);
-    i--;
+    pthread_create(&thread_handles[i], NULL, thread_function, &thread_args[i]);
+    pthread_join(thread_handles[i], NULL);
   }
 
+  if (a != 0) {
```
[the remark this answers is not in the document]
Pavel-Durov: I tried & failed to override thread-global stack variables, my test file 👉  f642314b5f127fd666830a98b6b71085565a4e50
I'll try harder 💪

ykrt/src/thread_intercept.rs
```diff
@@ -1,11 +1,83 @@
...
+    let thread_routine = unsafe {
+        // Obtain ThreadRoutine struct
+        Box::from_raw(arg as *mut ThreadRoutine)
+    };
+
+    // Call original thread routine
+    let result = (thread_routine.start_routine)(thread_routine.args);
+
+    unsafe {
+        // Free allocated stack
```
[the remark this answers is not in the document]
Pavel-Durov: done

tests/c/pthread_create.c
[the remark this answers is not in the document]
Pavel-Durov: I was able to create a simple example with a good reproduction rate (when I run it I am able to get race conditions every time)
👉 ee405ad6580ec941fec1082b349e95ac2ba6fb81

tests/c/pthread_create.c
Pavel-Durov: covered in 👉 ee405ad6580ec941fec1082b349e95ac2ba6fb81

ykrt/src/thread_intercept.rs
Pavel-Durov: We decided to not handle this edge case of thread cancelation although we understand that it might be a potential issue the impact is not clear. I added `YKFIXME`  as a reminder.

tests/c/pthread_create.c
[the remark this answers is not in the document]
Pavel-Durov: Test output:
```shell
$ cargo test thread_local_stack.c -- -- nocapture
...
Pattern (error at line 1):
>> |704982704,704982704,704982704

Text (error at line 1):
>> |690086522,690183060,622836793,

```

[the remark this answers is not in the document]
ptersilie: Oh I see. You just spawn multiple threads, let them all run, and then check that each thread outputs the same. If we share the shadowstack, the outputs will be different. Good work!

ykrt/src/thread_intercept.rs
```diff
@@ -1,11 +1,83 @@
...
+#[derive(Debug)]
+struct ThreadRoutine {
+    pub args: *mut c_void,
+    pub start_routine: extern "C" fn(*mut c_void) -> *mut c_void,
+}
+
+extern "C" fn wrap_thread_routine(arg: *mut c_void) -> *mut c_void {
+    let str = CString::new("shadowstack_0").unwrap();
+    let shadowstack_symbol_addr = unsafe {
+        //  Obtain address of a shadowstack_0 symbol
```
ltratt, line 18: It would be more idiomatic to put these comments in the line *above* the `let`, which (hopefully) let's cargofmt put the code on a single line.

ykrt/src/thread_intercept.rs
```diff
@@ -1,11 +1,83 @@
...
+        dlsym(null_mut(), str.as_ptr() as *const i8)
+    };
+    if shadowstack_symbol_addr.is_null() {
+        panic!("Unable to find shadowstack address")
+    }
+    let stack_addr = unsafe {
+        // Allocate stack
```
ltratt, line 25: I think we can kill this comment.

ykrt/src/thread_intercept.rs
```diff
@@ -1,11 +1,83 @@
...
+        malloc(SHADOW_STACK_SIZE)
+    };
+    if stack_addr.is_null() {
+        panic!("Unable allocate stack")
```
ltratt, line 29: s/Unable allocate/Unable to allocate/

ykrt/src/thread_intercept.rs
```diff
@@ -1,11 +1,83 @@
...
+    }
+    unsafe {
+        // Set shadowstack symbol with new allocated stack
+        *(shadowstack_symbol_addr as *mut *mut c_void) = stack_addr;
+    }
+    let thread_routine = unsafe {
+        // Obtain ThreadRoutine struct
```
ltratt, line 36: Why do we have to `Box` this?

ykrt/src/thread_intercept.rs
Pavel-Durov: OK, killed 👉  7bb37f0ab0129a7eed413b2d3a0a0c5b3c950355

ykrt/src/thread_intercept.rs
Pavel-Durov: fixed 👉 b96630ebb6163802890c460913ef029aaaaccd5e

ykrt/src/thread_intercept.rs
Pavel-Durov: We're boxing `ThreadRoutine` cause we're passing a reference from Rust to C that will be used asynchronously.

If we don't Box it, we will have Rust stack-allocated reference, which might be dropped by the time it's used in `__wrap_pthread_create` since `__wrap_pthread_create` and `thread_function` run asynchronously.

I added an explanation in the wrapper function docstring 👉 https://github.com/Pavel-Durov/yk-fork/blob/thread-local-shadowstack/ykrt/src/thread_intercept.rs#L48.

Is there another way of doing it without Boxing?

ltratt: Ah, so the `arg` parameter needs to be carefully documented along the lines of "thread_function: *const c_void is a pointer to a heap value produced by Box::into_raw. This function will consume and drop that box."

ykrt/src/thread_intercept.rs
```diff
@@ -1,11 +1,83 @@
...
+
+/// This function intercepts the standard POSIX pthread_create `pthread_create` function and wraps it with custom
+/// YK functionality.
+///
+/// The function stores `start_routine` and `args` pointers in `ThreadRoutine` struct raw pointer.
+/// `ThreadRoutine` is passed to the `wrap_thread_routine` function as a raw pointer,
+/// that will use it to call the original routine with the original arguments.
+///
+/// The `ThreadRoutine` raw pointer is initialised by the `Box::into_raw` call.
+/// It will be dropped once it's re-constructed as a box from the raw pointer. i.e. when `Box::from_raw` is called.
```
ltratt, line 53: This sentence is correct but not useful, because it doesn't say *who* will call `Box::from_raw`.

Pavel-Durov: Added more info 👉 cfcd40160b15d79e6c0c2e7bf54951f620e92e3e

ykrt/src/thread_intercept.rs
Pavel-Durov: added 👉 165ad32010100033f9c9db7c28a4e377a560707e

ltratt:
Please squash.

Pavel-Durov:
This PR is dependent on this YKLVM change 👉 https://github.com/ykjit/ykllvm/pull/94
I can squash but its not yet ready to be merged

ltratt:
OK. Though IIUC @ptersilie has given you the bits you need to make https://github.com/ykjit/ykllvm/pull/94 work sufficiently for this PR?

Pavel-Durov:
added `emulated-tls` to `yk-config` 👉 a203496f9b62f9c48c6b3474eca4994872ea6059

Pavel-Durov:
updated 👉 147c821085a1e8c37bb926d1d1413d7001fccb39

Pavel-Durov:
Ready for another review @ptersilie, @ltratt 
Tested with related changes in https://github.com/ykjit/ykllvm/pull/94

Pavel-Durov:
Ready for another review @ptersilie, @ltratt 
Tested with related changes in https://github.com/ykjit/ykllvm/pull/94

bin/yk-config
```diff
@@ -90,7 +90,7 @@ handle_arg() {
             # Use lld.
             OUTPUT="${OUTPUT} -fuse-ld=lld"
             # Embed LLVM bitcode as late as possible.
-            OUTPUT="${OUTPUT} -Wl,--mllvm=--embed-bitcode-final"
+            OUTPUT="${OUTPUT} -Wl,--mllvm=--embed-bitcode-final -Wl,--mllvm=--emulated-tls"
```
ptersilie, line 93: We should put this flag on its own line and add a comment what it does: forces use of emulated TLS since our JIT doesn't support native TLS.

Pavel-Durov: done 👉 e55dc233f3affad262494786a9e04cdc4bcecb38

ptersilie:
This PR is currently on hold. While we have now swapped out the JIT for ORG which supports native TLS, there is currently a bug related to PIE and stackmaps: https://github.com/llvm/llvm-project/issues/75074.

ltratt:
After a discussion with @Pavel-Durov, we've decided to keep this open until either:

1. The new JIT is enabled, at which point we close this PR without merging.
2. The LLVM bug is fixed, at which point we merge this PR.

ltratt:
Condition (1) has been met so time to close this one.

PR #879, opened 2023-10-16
Resizing the aux buffer causes problems, so don't.

ltratt:
If the collector thread determines that the trace needed more room, it resizes the buffer and updates the capacity. However, something goes wrong that causes (at least) the pointer value to become corrupted in some way. There are several possible candidates: the updates are done non-atomically in a "collector" thread so may not synchronise with the being-collected thread; and updates are also not atomically synchronised with Rust code. My surface-level attempts to fix this haven't succeeded, so either there's more synchronisation needed than I currently realise, or there are other problems as well.

This commit thus simply disables trace buffer reallocation. For yk this isn't likely to be a big problem: we throw away big traces anyway. [Arguably the current buffer size is much bigger than we need and want, but that's something we can worry about separately.]

With this commit I can successfully run "yklua db.lua" in a loop without an error: without this commit, it nearly always dies (normally with a SEGFAULT in realloc/malloc).

Fixes https://github.com/ykjit/yklua/issues/38.

### DIFF
--- a/hwtracer/src/perf/collect.c
+++ b/hwtracer/src/perf/collect.c
@@ -270,24 +270,13 @@ bool read_aux(void *aux_buf, struct perf_event_mmap_page *hdr,
     new_data_size = (size - tail) + head;
   }
 
-  // Reallocate the trace storage buffer if more space is required.
   __u64 required_capacity = trace->len + new_data_size;
   if (required_capacity > trace->capacity) {
-    // Over-allocate to 2x what we need, checking that the result fits in
-    // the size_t argument of realloc(3).
-    if (required_capacity >= SIZE_MAX / 2) {
-      // We would overflow the size_t argument of realloc(3).
-      hwt_set_cerr(err, hwt_cerror_errno, ENOMEM);
-      return false;
-    }
-    size_t new_capacity = required_capacity * 2;
-    void *new_buf = realloc(trace->buf.p, new_capacity);
-    if (new_buf == NULL) {
-      hwt_set_cerr(err, hwt_cerror_errno, errno);
-      return false;
-    }
-    trace->capacity = new_capacity;
-    trace->buf.p = new_buf;
+    // FIXME: Reallocate the trace storage buffer if more space is required.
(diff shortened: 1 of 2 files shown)


ltratt:
@Pavel-Durov Can I check this fixes https://github.com/ykjit/yklua/issues/38 for you too?

hwtracer/src/perf/collect.c
```diff
@@ -270,24 +270,13 @@ bool read_aux(void *aux_buf, struct perf_event_mmap_page *hdr,
...
-    }
-    size_t new_capacity = required_capacity * 2;
-    void *new_buf = realloc(trace->buf.p, new_capacity);
-    if (new_buf == NULL) {
-      hwt_set_cerr(err, hwt_cerror_errno, errno);
-      return false;
-    }
-    trace->capacity = new_capacity;
-    trace->buf.p = new_buf;
+    // FIXME: Reallocate the trace storage buffer if more space is required.
```
vext01, line 275: Is this even a FIXME? Based on our discussion earlier, we may never want to reallocate?

ltratt: Agreed, we *probably* don't, but until we've measured things, I'm not totally sure.

vext01: We can leave it a FIXME for now I guess.

We didn't adjust the buffer size. Do we want to do that either here or in a follow up PR?

ltratt: > We didn't adjust the buffer size. Do we want to do that either here or in a follow up PR?

Not sure I understand this.

vext01: What I mean is, we no longer allow the buffer to grow, so is the initial buffer size appropriate.

I forget where the initial buffer size comes from. Does the user select it, or did we hard-code it?

ltratt: In practise it's set by constants in the C code, though users can override this.

vext01: OK, let's get this in. We can always tweak it later.

PR #880, opened 2023-10-16
Re-use perf file descriptors for trace collection.

vext01:
Laurie and Lukas realised that opening Linux perf (`perf_event_open()`) can be very slow.

To quote Laurie:

> it's bimodal: one peak tightly clustered around 4ms, another peak
> loosely clustered between 20-28ms

This change therefore makes the JIT re-use perf file descriptors (per-thread).

This required a small change to the state machine in the decoder. Previously we'd always expected a PSB+ sequence at the start of the trace, but with a re-used perf fd you won't get one until the next time `IA32_RTIT_STATUS.PacketByteCnt` exceeds the next emitted-bytes threshold.

In general Intel recommend that you always scan for a PSB when starting to decode a trace. We get away without doing this because:

 a) For a new perf fd, the packet stream is always aligned at the start
    of a PSB packet.

 b) Nothing we do causes the packet stream to become desynchronised, so
    for a re-used perf fd, the packet stream is always packet-aligned.

### DIFF
--- a/hwtracer/src/perf/collect.c
+++ b/hwtracer/src/perf/collect.c
@@ -20,6 +20,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <syscall.h>
+#include <threads.h>
 #include <time.h>
 #include <unistd.h>
 
@@ -35,6 +36,17 @@
 #ifndef INFTIM
 #define INFTIM -1
 #endif
+
+/*
+ * The thread's perf file descriptor and its associated underlying `mmap(2)`
+ * regions. The file descriptor is re-used for subsequent trace collections for
+ * the same thread.
+ *
+ * FIXME: These leak when a thread dies.
+ */
+static thread_local void *cache_base_buf = NULL;
+static thread_local void *cache_aux_buf = NULL;
+static thread_local int cache_perf_fd = -1;
 
 enum hwt_cerror_kind {
   hwt_cerror_unused = 0,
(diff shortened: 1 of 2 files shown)


vext01:
Raising as a draft, because I'm not sure if we should (or if it's possible) attempt to close and unmap the stuff upon the death of a thread.

@ltratt any thoughts?

ltratt:
@vext01 Looks good to me, and wonderfully minimal!

vext01:
So we are OK to let the fd and mmap regions leak when a thread dies? I imagine the kernel will forcibly close and unmap them anyway?

ltratt:
I doubt it will close them when a thread dies. So you're right, a FIXME is probably a good idea here.

My aim, in the next week or three, is to largely replace this code, so I view this PR as an excellent temporary hack that's validated Lukas's thesis, and is worth putting in now, but whose long-term consequences we can probably ignore.

vext01:
FWIW, found this:

https://man7.org/linux/man-pages/man3/pthread_cleanup_push.3.html

For now I'll push a FIXME.

ltratt:
> FWIW, found this:
>
>https://man7.org/linux/man-pages/man3/pthread_cleanup_push.3.html

See also https://github.com/ykjit/yk/pull/878#discussion_r1360350722 !

vext01:
FIXME pushed. Ready for @ptersilie's critical eye.

ptersilie:
Having a very uncritical eye today. So please squash.

vext01:
splat.

PR #888, opened 2023-10-26
Respect `CARGO_JOBS` when building ykllvm

ltratt:
This PR makes ykbuild pass down the value of `CARGO_JOBS` to ykllvm, so that you can easily restrict the number of jobs that building yk spawns without knowing that ykllvm is lots and lots of non-cargo-controlled code. Most of the hard work is in https://github.com/ykjit/yk/commit/4362533660e89098da6dd6d5e8f7187248b129e7, but when I looked at the code a bit more, I realised there's a long-standing weirdness that I ended up fixing in https://github.com/ykjit/yk/commit/6a6761bc2d5e866f7b927f02ee6b663509dc276d.

Note that this PR has one mild surprise that probably won't lead to actual surprise: if you're really unlucky `cargo` can be building `CARGO_JOBS` number of jobs at just the point that ykbuild kicks off `CARGO_JOBS` number of jobs in ykllvm. In practise this is unlikely to be an issue, because a) ykbuild is an unintentional synchronisation point in the build that almost immediately else blocks anything else from compiling until it's done b) the first ykbuild does is run ykllvm's config stuff, which takes about 10 secs to run, virtually guaranteeing that cargo's other jobs are in fact complete.

### DIFF
--- a/ykbuild/build.rs
+++ b/ykbuild/build.rs
@@ -136,8 +136,13 @@ fn main() {
         .current_dir(build_dir.as_os_str().to_str().unwrap());
 
     let mut build_cmd = Command::new("cmake");
+    let mut build_args = vec!["--build".into(), ".".into()];
+    if let Ok(jobs) = env::var("NUM_JOBS") {
+        build_args.push("-j".into());
+        build_args.push(jobs);
+    }
     build_cmd
-        .args(["--build", "."])
+        .args(build_args)
         .current_dir(build_dir.as_os_str().to_str().unwrap());
 
     let mut inst_cmd = Command::new("cmake");
@@ -180,10 +185,6 @@ fn main() {
                 .to_str()
                 .unwrap(),
         );
-
-        if generator == "Unix Makefiles" {
-            build_cmd.args(["-j", num_cpus::get().to_string().as_str()]);
-        }
 
         cfg_cmd.status().unwrap().exit_ok().unwrap();
     }
--- a/ykbuild/Cargo.toml
+++ b/ykbuild/Cargo.toml
@@ -14,6 +14,5 @@ tempfile = "3.8"
 
 [build-dependencies]
 fs4 = "0.6"
-num_cpus = "1.0"
 rerun_except = "1.0.0"
 which = "4.4"


ykbuild/build.rs
```diff
@@ -136,8 +136,13 @@ fn main() {
         .current_dir(build_dir.as_os_str().to_str().unwrap());
 
     let mut build_cmd = Command::new("cmake");
+    let mut args = vec!["--build".into(), ".".into()];
+    if let Ok(jobs) = env::var("CARGO_JOBS") {
```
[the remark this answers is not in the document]
ltratt: I'd an idiot.

I saw that both `CARGO_BUILD_JOBS` and `NUM_JOBS` are supported and... merged the two in my head? I don't know which of the two to prefer. Any thoughts?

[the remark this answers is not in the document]
ltratt: Aha, setting `CARGO_BUILD_JOBS` causes cargo to set `NUM_JOBS` in the build script! Adjusting now.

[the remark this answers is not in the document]
vext01: Do we only build ykllvm in the build target? I assume so.

That being the case `CARGO_BUILD_JOBS` seems best?

So if I run `cargo -j <blah>` cargo sets this var, and we pick it up. Correct?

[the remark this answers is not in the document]
vext01: Side-note: `NUM_JOBS` is an awfully generic and clash-inviting variable name :) Not that we can do anything about it.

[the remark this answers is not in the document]
ltratt: Fixed in https://github.com/ykjit/yk/pull/888/commits/c8a1126aea25a64f03f7090b795671fecf8b7af3. Even better, cargo seems to *always* set this: if the user doesn't set `CARGO_BUILD_JOBS` then `NUM_JOBS` is set to the number of CPUs. So we don't even need the `num_cpus` crate anymore!

This will require some squashing.

ltratt:
> Can the user still provide a different -j to what cargo uses, via YKB_YKLLVM_BUILD_ARGS?

Supporting that is hard as then we need to parse that more carefully. Also the current code (before this PR) arbitrarily sets `-j` in some cases, so this PR doesn't make things worse. I think I'd add code to deal with `-j` in `YKB_YKLLVM_BUILD_ARGS` in another PR.

vext01:
I'm not sure what the second commit is fixing. The commit message implies a bug, but what is it?

ltratt:
Honestly, the best I could do right now is squash this down, as the original two commits don't quite make sense any more, and are obscuring what the PR *should* have been about. It should have said "respect `NUM_JOBS`". I can squash it down if you'd like to make reviewing easier?

vext01:
>  I think I'd add code to deal with -j in YKB_YKLLVM_BUILD_ARGS in another PR.

It may not be worth your time. We could just say: if you want to do weird stuff, be a weirdo like Edd and build ykllvm yourself 😆 

That said, I think @Pavel-Durov was considering using your build system with distcc. I don't know if he got that far though... I tested that distcc and passing your own `-j` worked with the easy build system once upon a time.

vext01:
>  can squash it down if you'd like to make reviewing easier?

Squash it and I'll pretend I haven't seen the PR before :)

ltratt:
@vext01 Ready for (pretend it's the first time) (re)review.

ykbuild/build.rs
```diff
@@ -136,8 +136,13 @@ fn main() {
         .current_dir(build_dir.as_os_str().to_str().unwrap());
 
     let mut build_cmd = Command::new("cmake");
+    let mut build_args = vec!["--build".into(), ".".into()];
```
vext01, line 139: This is an optional nitpick, but more args use `into()` than don't, which would suggest the array would be better typed `&str` than `String`, then many of the `into()`s will go away.

ltratt: We can't do `&str` in the array because `env::var` returns a `String`.

ykbuild/build.rs
```diff
@@ -181,10 +186,6 @@ fn main() {
                 .unwrap(),
         );
 
-        if generator == "Unix Makefiles" {
```
vext01, line 184: So does `make` get a sensible default `-j` from elsewhere now?

ykbuild/build.rs
vext01: So borrow it :)

ltratt: In this case it's not in scope long enough to do it.

ykbuild/build.rs
ltratt: Yes, we pass `-j` to *whatever* the build command is, without special casing makefiles.

ykbuild/build.rs
vext01: Fair enough.

PR #890, opened 2023-10-27
Fixes related to recent ykllvm upstream merge.

ptersilie:
Once https://github.com/ykjit/ykllvm/pull/99 is merged I will update the submodule and then this is ready too.


hwtracer/src/llvm_blockmap.rs
```diff
@@ -132,6 +132,9 @@ impl BlockMap {
             let n_blks = leb128::read::unsigned(&mut crsr).unwrap();
             for _ in 0..n_blks {
                 let mut corr_bbs = Vec::new();
+                if _version > 1 {
```
ltratt, line 135: This was (probably) prefixed with `_` because it was unused before, but now it's used we should change the variable name to just `version`.

ptersilie: Fixed in [c309468](https://github.com/ykjit/yk/pull/890/commits/c3094689a2fb65bad59ea67598ae268f5dd8f209). Will squash together with the ykllvm submodule update later.

ltratt: Technically the ykllvm update should be a normal commit, and not done as part of squashing. But there's good news! Since the ykllvm merge will (hopefully) go through without problems, you can push a commit referencing the hash of your last commit to ykllvm now. If all goes well, then it saves us having to wait for ykllvm to merge.

ptersilie: I disagree that it should be a normal commit as that would break `git bisect`. Pushed ykllvm update in [0d112ca](https://github.com/ykjit/yk/pull/890/commits/0d112ca584831ef2a766e0232d46ef6a60994516). Ok to squash?

ltratt: Doing so doesn't break git bisect, I promise!

ltratt:
Please squash.

hwtracer/src/llvm_blockmap.rs
ptersilie: Are you sure about that? Let's say we have
```
commit 1 Update ykllvm
commit 2 ykllvm related fixes
```

If we `git bisect` and it gets to `commit 2` it will work, but when it gets to `commit 1` it won't since yk at that commit won't work with the updated ykllvm.

ltratt: If ykllvm is updated (*especially* if it's squashed), you have to update the hash. But if that doesn't happen (and it didn't happen in this pair of PRs), there's no problem.

ptersilie: I feel like we are talking about different things, cause I have no idea what you mean. I did update the `ykllvm` submodule. What I'm saying is that the `ykllvm` submodule update needs to happen in the same commit as the fixes for the `ykllvm` update. Otherwise if you do `git bisect` where there will be a commit where the submodule is updated, but the fixes haven't been added, which will then fail the tests in `yk` (probably `yk` won't even compile).

ptersilie:
Squashed.

Issue #818, opened 2023-08-22
Find a better way to outline external code than frame counting.


vext01:
Currently when the trace compiler hits a call to external code, it disassembles the raw instruction stream counting the depth of the call stack as it goes. Normal trace compilation resumes when we get back to the same stack depth as when we started.

This will fail if the instruction stream does stack gymnastics (e.g. emits not a `call` but a `push retaddr; jmp func`).

We need to find a better way. Ideally we'd periodically (after all control flow dispatches?) look at the value of the stack pointer to decide if we should stop outlining, but this will slow us down.

Last time I looked PT has a way to record the dynamic value of the stack pointer, but you can't control when such packets are emitted, and we could be notified too late.

(If we can solve this, I think we can remove the disassembler from the PT decoder)

vext01:
Just reading up on this again.

The facility I mentioned above is "PEBS output to PT". By the looks of it, PEBS is a mechanism for raising events when certain performance counters overflow. You can have those events (in some CPUs) written to PT packets which record the stack pointer *near* the time of the overflow.

Unless I've overlooked something, this isn't sufficient for our use case.

ltratt:
As discussed this morning offline, the problem here is when we call to unmappable code and it calls back into mappable code: we have to be sure that we've got back to mappable code without any unmappable code being "left on the callstack" as it were.

One way of tackling this is to realise that we can in most cases statically determine the successor blocks that can be seen after a call. So for example, if we saw a function call in block B, we know that the next blocks we can see are (say) C and D. [For C and D we'd also have to work out which blocks can come before them for the next check to be correct, but I'll try and keep things simple for this comment!] If we see C and D in the stream but haven't seen B, we know that we've got back to the "right point" and can continue. However, if we see B again, before we've seen C or D, there is still unmappable code on the callstack and we're recursing.

There are some challenges: for example, if a function call is the very first or very last thing in a function we won't have predecessor/successor blocks that we can statically determine. My inclination here would be to mark such functions as `do_not_trace` and/or say that if we encounter them during tracing that we abort the trace. Such cases *probably* don't happen very often -- we should easily be able to measure *how* often and then determine whether it's worth having clever support for them or not. [Technically the predecessor/successor analysis could be whole-program, though I imagine coding that up would be... fun!]

PR #891, opened 2023-11-02
Zero length call fixes for our PT decoder.

### LINKED ISSUE #874: Bug in compressed return logic?
On [one occasion](https://github.com/ykjit/yk/pull/870#issuecomment-1758293195) we saw:

```
19:44:38 ---- lang_tests::unroll_safe_implies_noinline.c stderr ----
19:44:38 
19:44:38 Pattern (error at line 2):
19:44:38    |...
19:44:38 >> |--- Begin aot ---
19:44:38    |...
19:44:38    |call void @never_aot_inline(i32 noundef ...
19:44:38    |...
19:44:38    ...
19:44:38 
19:44:38 Text (error at line 4):
19:44:38    |thread '<unnamed>' panicked at hwtracer/src/pt/ykpt/mod.rs:453:13:
19:44:38    |assertion failed: taken
19:44:38    |note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
19:44:38 >> |fatal runtime error: failed to initiate panic, error 5
19:44:38 
19:44:38 ---- lang_tests::unroll_safe_implies_noinline.c stdout ----
```

Suggest we expected a compressed return, but the chip did not?

vext01:
This is a handful of "zero-length calls" fixes and clarifications for compressed returns in our PT decoder.

To be clear: we have not (as far as I know) been bitten by any of the issues that these fixes solve. But that's not to say that we couldn't be in the future.

This sadly does **not** fix #874, but we may as well be as correct as we can.

Requires a compiler change. ~Coming soon~ https://github.com/ykjit/ykllvm/pull/100

### DIFF
--- a/hwtracer/src/llvm_blockmap.rs
+++ b/hwtracer/src/llvm_blockmap.rs
@@ -39,6 +39,8 @@ pub enum SuccessorKind {
 pub struct CallInfo {
     /// Offset of the call instruction
     callsite_off: u64,
+    /// Offset of the return address (should the call return conventionally).
+    return_off: u64,
     /// Offset of the target of the call (if known statically).
     target_off: Option<u64>,
 }
@@ -46,6 +48,10 @@ pub struct CallInfo {
 impl CallInfo {
     pub fn callsite_off(&self) -> u64 {
         self.callsite_off
+    }
+
+    pub fn return_off(&self) -> u64 {
+        self.return_off
     }
 
     pub fn target_off(&self) -> Option<u64> {
@@ -151,10 +157,13 @@ impl BlockMap {
                 let mut call_offs = Vec::new();
                 for _ in 0..num_calls {
                     let callsite_off = crsr.read_u64::<NativeEndian>().unwrap();
+                    let return_off = crsr.read_u64::<NativeEndian>().unwrap();
+                    debug_assert!(callsite_off < return_off);
                     let target = crsr.read_u64::<NativeEndian>().unwrap();
                     let target_off = if target != 0 { Some(target) } else { None };
                     call_offs.push(CallInfo {
                         callsite_off,
+                        return_off,
                         target_off,
                     })
                 }
--- a/hwtracer/src/pt/ykpt/mod.rs
+++ b/hwtracer/src/pt/ykpt/mod.rs
@@ -581,6 +581,7 @@ impl<'t> YkPTBlockIterator<'t> {
                     reposition = true;
                 }
                 iced_x86::FlowControl::Call => {
+                    // A *direct* call.
                     if inst.code() == iced_x86::Code::Syscall {
                         // Do nothing. We have disabled kernel tracing in hwtracer, so
                         // entering/leaving a syscall will generate packet generation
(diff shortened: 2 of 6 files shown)


ltratt:
Can we find a test (or more) which exercises this?

ptersilie:
Looks good to me, but @ltratt is closer to the matter so I'll leave the review to him.

vext01:
> Can we find a test (or more) which exercises this?

Actually yes, I probably could. At least for disassembly-based decoding.

vext01:
(I've also realised that I didn't sync ykllvm, but I can do that in a bit after I've done the additional testing)

vext01:
OK, this is amusing. I added a test that traces this:

```
 # A little function that returns the value of RIP via (what Intel calls) a "zero-length call".                                                                                                                                                                                                                        
zero_len_call:
    mov rax, 0                                                                                                                                                                                                                               
    call y
y:
    pop rax
    ret
```

Of course it breaks the outliner's frame counter, because there's no corresponding `ret` for `call y`. Should have seen it coming!

```
x86_64_zero_len_call: src/jitmodbuilder.cc:1465: Module *JITModBuilder::createModule(): Assertion `UR->StackAdjust > 0' failed.
```

So this is more evidence that we haven't yet seen zero-len calls when disassembling.

Related #818 

This means that it's not easy for me to test zero-length calls in disassembly-based decoding.

The other place we can see zero-length calls is in compiler assisted decoding. To test that we'd have to get LLVM to emit code like in the above test (not easy), or use llvm's inline asm which we don't yet support properly.

I suggest we move on?

ltratt:
@vext01 Do you mean "merge the PR without tests"?

vext01:
At best we could commit the test ignored, until the frame counter is gone.

I'd have to do some tidying up and find a way to only run a test on x86_64. Do you think it's worth it right now? Not sure.

ltratt:
Let's merge this one, but not raise/merge future PT decoder PRs unless we've managed to find a test for it?

bors r+

vext01:
> but not raise/merge future PT decoder PRs unless we've managed to find a test for it?

I think that's too strong. Fixing #818 is non-trivial. How about we add an issue instead?

ltratt:
I'm not sure what the issue to be raised would be, but I guess "yes, do it".

vext01:
I've just devised a test for one of the modes!

Look at this filth:

```
# A little function that returns the value of RIP via (what Intel calls) a
# "zero-length call".
zero_len_call:
    mov rax, 0
    call y
y:
    # Note that `push z` doesn't do what you think it would!
    lea rdi, [z]
    push rdi
    ret
z:
    pop rax
    ret
```

A `call` with no `ret` skews the frame counter by one, so we can restore equalibrium by doing a `ret` with no `call` afterwards :P

Will tidy up and raise a PR after lunch!

ltratt:
Please squash.

vext01:
Since we have the opportunity, we may as well sneak the test into this PR, where it belongs!

5c73b2f8e605 adds the test. LGTY?

vext01:
I've really messed up here. That merge commit isn't supposed to be there. Can I force push to fix?

ltratt:
Please squash.

vext01:
All squashed down for merge.

PR #892, opened 2023-11-02
Extra instcombine pass no longer works.

ptersilie:
Since syncing upstream LLVM some things in the pipeline appear to have changed, which means we no longer can run the instcombine pass here. We also can't pass the empty string here, since then the default pipeline runs which has optimisations we cannot yet deal with.

Hopefully @nmdis1999 work will reveal some passes we can put back in here and/or find ways to deal with the default
pipeline.

### DIFF
--- a/bin/yk-config
+++ b/bin/yk-config
@@ -10,11 +10,14 @@ if [ ! -d "${DIR}/../ykrt" ]; then
     exit 1
 fi
 
-# The post-link time optimisations that we can tolerate.
+# The post-link time optimisations that we can tolerate. At this moment this is
+# only the instcombine pass in the below configuration (copied over from the
+# default pipeline). Using `instcombine` on its own breaks, possibly due to the
+# missing `no-verify-fixpoint`.
 #
 # FIXME: Once we know what breaks us, allow the user to use the regular -O0,
 # -O1, -O2, etc. clang flags like normal.
-POSTLINK_PASSES_STR=instcombine
+POSTLINK_PASSES_STR="instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>"
 
 OUTPUT=""
 


vext01:
Does `function()` work (i.e. an empty function pass manager)?

ptersilie:
No: `/bin/yk-config: 17: Syntax error: "(" unexpected`

ltratt:
@ptersilie you'd need to put it in string quotes to avoid the shell complaining.

ptersilie:
Ah, but also no:
```
LLVM ERROR: unable to parse pass pipeline description 'function()': unknown function pass ''
```

ltratt:
OK. Then I think `verify` is OK.

bors r+

ptersilie:
`curl: (28) Failed to connect to sh.rustup.rs port 443 after 300697 ms: Timeout was reached`. Think we can just retry.

ltratt:
@ptersilie A lot of tests seem to have broken?

ptersilie:
Whoops. Investigating.

ptersilie:
Ready for rereview and squashing.

bin/yk-config
```diff
@@ -14,7 +14,7 @@ fi
 #
 # FIXME: Once we know what breaks us, allow the user to use the regular -O0,
 # -O1, -O2, etc. clang flags like normal.
-POSTLINK_PASSES_STR=verify
+POSTLINK_PASSES_STR="instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>"
```
vext01, line 17: In the commit message, I think you should say where this comes from and why it fixes it.

vext01:
One comment from me, otherwise LGTM.

bin/yk-config
ptersilie: I don't think I know the answer to either of those. Well, it comes from `opt`, so I presume it's the default pipeline way of running `instcombine`. Why it fixes it? :man_shrugging:

vext01: becuase `opt` has those arguments in the upstream pipeline, and one of them (the fixpoint one) suppresses the yklua build error.

ltratt: I assume the `no-` things turn parts of `instcombine` off and that's the fix?

ptersilie: Ah, it might be the `no-verify-fixpoint` that fixed it, cause that sounds like the error I got.

ltratt: Is `no-verify-fixpoint` on its own enough I wonder?

ptersilie: Fixed in [2aae465](https://github.com/ykjit/yk/pull/892/commits/2aae46516b3f3f5efcb416ef4e0aea7d8589aa5c). Personally, I don't get much out of this comment, but see if that works for you.

ltratt: > Is no-verify-fixpoint on its own enough I wonder?

Can we check this or, at least, raise an issue to investigate further? It might be useful to know what the minimal fix is.

ptersilie: I can check this on Monday. Just leave the PR open until then.

ltratt: Thanks!

ltratt: I wondered what `no-verify-fixpoint` was, because it sounds potentially bad (e.g. what if it's checking for correctness). Actually, it's not really bad https://reviews.llvm.org/D154579:

>In order to prevent accidental optimization regressions in the future, this implements a default-enabled verify-fixpoint option, which will make sure that the fix point has indeed been reached after a single iteration. This means that tests where this is not the case need to be explicitly annotated. The actual optimization pipeline will disable this option, as failure to reach the fix point is expected to happen there (in rare cases, as described above).

So really it's an internal check that `InstCombine` does a good optimisation job, but that's too strict a check, which is why it's disabled in the real pipeline. [There is an interesting philosophical debate to be had here about whether this should have been on/off by default, but that's outside our control!]

My guess based on this is that we might only need `no-verify-fixpoint` for our case, but now that I understand what it is, I think "do whatever the default pipeline does" (which has a couple of other things) is the right thing to do.

ltratt:
Please squash

ptersilie:
Oh, forgot to rebase. Ok to do so?

ltratt:
Please do. And can you also confirm (while bors is away) that you've run the full test suite etc?

ptersilie:
Rebased. Can confirm that tests and benchmarks run.

PR #893, opened 2023-11-03
Trace naming tweaks

ltratt:
This is a purely cosmetic PR: it tweaks some comments and changes the name of one struct. It's part of my longer-term ambition to change the tracing backends to do things like reuse collectors, but a) this PR stands on its own b) I want to check to see if there are violent objections to the renaming.


ykrt/src/trace/mod.rs
```diff
@@ -18,8 +18,9 @@ pub(crate) mod hwt;
 
 pub(crate) use errors::InvalidTraceError;
 
-/// A tracer is an object which can start / stop collecting traces. It may have its own
-/// configuration, but that is dependent on the concrete tracer itself.
+/// A `Tracer` is a front-end to an underlying tracer approach (e.g. hardware or software tracing).
+/// That backend may have its own configuration options, which is why `Tracer` does not have a
```
vext01, line 22: Mostly fine, but if you want to refer to "that backend", wouldn't it be better to have used the term "backend" prior?

I know that you are using "underlying tracing approach" and "backend" as synonyms, but maybe unified terminology would be clearer?

ykrt/src/mt.rs
```diff
@@ -27,7 +27,7 @@ use crate::print_jit_state;
 use crate::{
     compile::{default_compiler, CompiledTrace, Compiler, GuardId},
     location::{HotLocation, HotLocationKind, Location, TraceFailed},
-    trace::{default_tracer, RawTrace, ThreadTracer, Tracer},
+    trace::{default_tracer, RawTrace, TraceCollector, Tracer},
```
vext01, line 30: Do you think that it's obvious that a `TraceCollector` is thread-specific?

I could buy that. Maybe we could clarify that in a docstring?

ykrt/src/trace/hwt/mod.rs
```diff
@@ -1,4 +1,4 @@
-//! Hardware tracing via ykrustc.
+//! Hardware tracing via hwtracer.
```
vext01, line 1: Wow! There's a blast from the past.

ykrt/src/mt.rs
ltratt: I guess we could document that but I can't imagine what a non-thread-specific tracer would even look like. It feels a bit like overkill to me in the sense of "what other bonkers things should I tell you are bonkers"? But maybe I'm missing out on something.

ykrt/src/trace/mod.rs
ltratt: Good point. Fixed in 0cd2907.

ykrt/src/trace/hwt/mod.rs
ltratt: Indeed!

ykrt/src/mt.rs
vext01: It's not bonkers though, is it?

Linux perf has a mode to trace all processes and threads on a whole CPU core:

> In per-cpu mode all processes (or processes from the selected cgroup i.e. -G option, or processes selected with -p or -u) are traced. Each cpu has its own buffer. Inheritance is allowed.

ykrt/src/trace/mod.rs
```diff
@@ -18,9 +18,9 @@ pub(crate) mod hwt;
 
 pub(crate) use errors::InvalidTraceError;
 
-/// A `Tracer` is a front-end to an underlying tracer approach (e.g. hardware or software tracing).
-/// That backend may have its own configuration options, which is why `Tracer` does not have a
-/// `new` method.
+/// A `Tracer` is a front-end to a tracer backend (e.g. hardware or software tracing). The tracer
```
vext01, line 21: perfect, thanks!

ykrt/src/mt.rs
ltratt: Actually I was a good boy and the doc string is still:

```
/// Represents a thread which is currently tracing.
```

so done without a new commit!

vext01: OK, good enough for me.

vext01:
Please squash.

ltratt:
Squashed.

PR #896, opened 2023-11-06
Find successful passes with genetic algorithm [DRAFT]

nmdis1999:
This PR aims to find successful passes (i.e which pass all YK and YKLUA tests) with the help of genetic algorithm. The fitness function currently gives priority to pass length (i.e number of flags) to pick up most likely parents. The PR is currently in draft form and the description will be updated once it's ready for review with an explanation of what and how this algorithm is working with YK test suite.


ykrt/pass_finder/try_passes.py
```diff
@@ -66,53 +66,96 @@ def log(logf, s):
...
+    angle_stack = []    
+
+    i = 0  
+    while i < len(passes_string):
+        char = passes_string[i]
+
+        if char == '(':
+            paren_stack.append(char)
+        elif char == ')':
+            if paren_stack:
```
ltratt, line 82: I don't think `paren_stack` can ever be non-true unless there's a bug. So I'd do:

```python
elif char == ')':
  paren_stack.pop()
```

ykrt/pass_finder/try_passes.py
```diff
@@ -66,53 +66,96 @@ def log(logf, s):
...
-    sout, serr = p.communicate() 
-    assert(p.returncode == 0)
+def split_passes(passes_string):
+    parts = []           
+    temp_part = []      
+    paren_stack = []     
+    angle_stack = []    
+
+    i = 0  
+    while i < len(passes_string):
```
ltratt, line 76: Since we only `i += 1` in the loop I tihink this is equivalent to:

```python
for c in passes_string:
  ...
```

where `c` is what is currently called `char`.

ykrt/pass_finder/try_passes.py
```diff
@@ -66,53 +66,96 @@ def log(logf, s):
...
-    pass_descrs = [x.strip() for x in sout.strip().split("\n")]
+    return parts
 
+def get_all_passes(is_prelink):
+    #cmd = get_opt_cmd(is_prelink)
+    cmd = "opt -passes='lto-pre-link<O2>' -print-pipeline-passes < /dev/null 2>/dev/null"
+    sout = subprocess.run(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.DEVNULL)
+    sout = sout.stdout.decode('utf-8')
+    pass_descrs = split_passes(sout)
+    print(f"\033[91m{pass_descrs}\033[0m")
```
ltratt, line 113: What do the control codes do?

ykrt/pass_finder/try_passes.py
```diff
@@ -185,12 +228,111 @@ def binary_split(logf, passes, is_prelink):
...
+    ret, exec_time = test_pipeline(logf, config)
+    nFlags = len(try_passes) 
+    if ret:
+        try:
+            exec_time = float(exec_time)  # Convert exec_time to a float
+            # Return True for successful programs and execution time as a negative value
+            print(f"\033[92m returned {ret} and execution time {exec_time}\033[0m")
+            return  [nFlags, exec_time]
+        except ValueError:
+            print(f"Error converting exec_time to float: {exec_time}")
```
ltratt, line 247: Is there a reason for catching this error? Why wouldn't we be able to convert the time to a float?

ykrt/pass_finder/try_passes.py
nmdis1999: the codes prints statements in ansi color, makes it easier to find some line in huge stdout dump (I'll remove this in final version).

ykrt/pass_finder/try_passes.py
nmdis1999: I am converting the exec_time from `str` to `float` so this check makes sure if the converstion is wrong we get a message.

ltratt: Have you ever seen it go wrong? If so, that probably means we're in deep trouble! Basically I'm not a fan of defensive-because-we-don't-know programming. So I would prefer to just say `float(x)` and if it raises an exception, then we know we have a bug to fix, whereas the `try/except` is likely to hide programming mistakes.

nmdis1999: ah, I see. Then it makes sense to remove `try/except` :)

ykrt/pass_finder/try_passes.py
vext01: I've done this before too. I'm OK with it for this little script.

Issue #841, opened 2023-09-20
Non serialised compilation error


Pavel-Durov:
## Issue

Failing to running yklua tests with `YKD_SERIALISE_COMPILATION=0` 

YkLua related Issues:
https://github.com/ykjit/yklua/issues/43
https://github.com/ykjit/yklua/issues/44
https://github.com/ykjit/yklua/issues/46
https://github.com/ykjit/yklua/issues/48
https://github.com/ykjit/yklua/issues/52

> Errors and backtraces can be found in attached issues.

Example of errors:

```shell

lua: /yk/ykllvm/llvm/lib/IR/Value.cpp:1101: void llvm::ValueHandleBase::RemoveFromUseList(): Assertion `*PrevPtr == this && "List invariant broken"' failed.
...

lua: /yk/target/debug/ykllvm/include/llvm/ExecutionEngine/Orc/ThreadSafeModule.h:134: decltype(auto) llvm::orc::ThreadSafeModule::withModuleDo(Func &&) [Func = (lambda at src/ykllvmwrap.cc:408:30)]: Assertion `M && "Can not call on null module"' failed.
...

lua: /yk/ykllvm/llvm/lib/IR/Value.cpp:1101: void llvm::ValueHandleBase::RemoveFromUseList(): Assertion `*PrevPtr == this && "List invariant broken"' failed.
....
lua: /yk/ykllvm/llvm/lib/IR/Value.cpp:285: llvm::ValueName* llvm::Value::getValueName() const: Assertion `I != Ctx.pImpl->ValueNames.end() && "No name entry found!"' failed.
lua: /yk/ykllvm/llvm/lib/IR/Value.cpp:285: llvm::ValueName* llvm::Value::getValueName() const: Assertion `I != Ctx.pImpl->ValueNames.end() && "No name entry found!"' failed.
...
lua: /yk/ykllvm/llvm/include/llvm/ADT/ScopedHashTable.h:248: llvm::ScopedHashTableScope<K, V, KInfo, AllocatorTy>::~ScopedHashTableScope() [with K = llvm::MemoryLocation; V = unsigned int; KInfo = llvm::DenseMapInfo<llvm::MemoryLocation>; AllocatorTy = llvm::RecyclingAllocator<llvm::BumpPtrAllocatorImpl<>, llvm::ScopedHashTableVal<llvm::MemoryLocation, unsigned int> >]: Assertion `KeyEntry == ThisEntry && "Scope imbalance!"' failed.

[...]

PR #898, opened 2023-11-08
Make sure threads don't exit until all LLVM compilation threads complete.

ltratt:
This is a horrible hack: we track in an atomic integer how many worker threads are doing something (which we use as a proxy for "are doing something with LLVM) and, when someone calls `yk_mt_drop` in C, we wait for those worker threads to finish doing whatever they're doing.

The obvious horror with this hack is that we stop the application exiting even if the interpreter has finished. A less obvious horror is that `Mt::drop` isn't called while compilation threads are running, which is why we have to put this hack into `yk_mt_drop` in the C API. That bakes in an assumption about the C-ness of interpreters, which is fine for now, but clearly not going forwards.

To emphasise how hacky this is, all the code is under a flag `yk_llvm_sync_hack`. This will also make it easy to find all the places to rip out when we no longer need this hack.

### DIFF
--- a/ykrt/build.rs
+++ b/ykrt/build.rs
@@ -7,4 +7,7 @@ pub fn main() {
     println!("cargo:rustc-cfg=jitc_yk");
     // Always compile in the HWT tracer.
     println!("cargo:rustc-cfg=tracer_hwt");
+    // FIXME: This is a temporary hack because LLVM has problems if the main thread exits before
+    // compilation threads have finished.
+    println!("cargo:rustc-cfg=yk_llvm_sync_hack");
 }
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -85,6 +85,11 @@ pub struct MT {
     /// How many worker threads are currently running. Note that this may temporarily be `>`
     /// [`max_worker_threads`].
     active_worker_threads: AtomicUsize,
+    #[cfg(yk_llvm_sync_hack)]
+    /// A temporary hack which keeps track of how many compile (etc.) jobs are active, because LLVM
+    /// dies horribly if the main thread exits before those jobs are finished. This then marries up
+    /// with Self::.llvm_sync_hack().
+    active_worker_jobs: AtomicUsize,
     /// The [Tracer] that should be used for creating future traces. Note that this might not be
     /// the same as the tracer(s) used to create past traces.
     tracer: Mutex<Arc<dyn Tracer>>,
@@ -116,6 +121,8 @@ impl MT {
             job_queue: Arc::new((Condvar::new(), Mutex::new(VecDeque::new()))),
             max_worker_threads: AtomicUsize::new(cmp::max(1, num_cpus::get() - 1)),
             active_worker_threads: AtomicUsize::new(0),
+            #[cfg(yk_llvm_sync_hack)]
+            active_worker_jobs: AtomicUsize::new(0),
             tracer: Mutex::new(default_tracer()?),
             compiler: Mutex::new(default_compiler()?),
             stats: YkStats::new(),
@@ -176,6 +183,9 @@ impl MT {
         // new worker thread iff we aren't already running the maximum number of worker threads.
         // Once started, a worker thread never dies, waiting endlessly for work.
 
+        #[cfg(yk_llvm_sync_hack)]
+        self.active_worker_jobs.fetch_add(1, Ordering::Relaxed);
+
         let (cv, mtx) = &*self.job_queue;
         mtx.lock().push_back(job);
         cv.notify_one();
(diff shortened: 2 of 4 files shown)


ykrt/build.rs
```diff
@@ -7,4 +7,7 @@ pub fn main() {
     println!("cargo:rustc-cfg=jitc_yk");
     // Always compile in the HWT tracer.
     println!("cargo:rustc-cfg=tracer_hwt");
+    // FIXME: This is a temporary hack because LLVM has problems if the main thread exits before
+    // compilation threads have finished.
```
vext01, line 11: Do we have an issue we can link? If so, is the issue up to date with our current findings?

ykrt/src/mt.rs
```diff
@@ -85,6 +85,11 @@ pub struct MT {
     /// How many worker threads are currently running. Note that this may temporarily be `>`
     /// [`max_worker_threads`].
     active_worker_threads: AtomicUsize,
+    #[cfg(yk_llvm_sync_hack)]
+    /// A temporary hack which keeps track of how many compile (etc.) jobs are active, because LLVM
+    /// dies horribly if the main thread exits before those jobs are finished. This then marries up
+    /// with Self::.llvm_sync_hack().
+    active_worker_jobs: AtomicUsize,
```
vext01, line 92: Did you opt for atomics because they are cheaper than stashing join handles?

ykrt/src/mt.rs
```diff
@@ -172,10 +179,11 @@ impl MT {
 
     /// Queue `job` to be run on a worker thread.
     fn queue_job(self: &Arc<Self>, job: Box<dyn FnOnce() + Send>) {
+        #[cfg(yk_llvm_sync_hack)]
+        self.active_worker_jobs.fetch_add(1, Ordering::Relaxed);
         // We have a very simple model of worker threads. Each time a job is queued, we spin up a
         // new worker thread iff we aren't already running the maximum number of worker threads.
         // Once started, a worker thread never dies, waiting endlessly for work.
-
```
vext01, line 178: That should have been a separate commit.

I won't make you change it :smile:

ykrt/src/mt.rs
```diff
@@ -557,6 +569,14 @@ impl MT {
             }
         }
     }
+
+    #[cfg(yk_llvm_sync_hack)]
+    pub fn llvm_sync_hack(&self) {
```
vext01, line 574: I should be violently objecting to a spinlock, but since this really isn't designed to be permanent, I guess it's ok for now. Agree?

ykrt/src/mt.rs
ltratt: In retrospect, moving the comment would have been better. Fixed in 3598efc.

ykrt/src/mt.rs
ltratt: Join handles would be harder, because worker threads aren't necessarily active, so "join handles" would require a) a condvar call to wake them up b) a bool to let them know to exit. So the horrors of an (atomic) count are at least briefer horrrors.

ykrt/build.rs
ltratt: I guess https://github.com/ykjit/yk/issues/841 is an example, but probably not the only one. @Pavel-Durov ?

ykrt/src/mt.rs
ltratt: It's terrible but, as you say, temporary. Anything more sophisticated won't (I hope!) be worth the effort.

ptersilie:
Love it! Please squash.

How do we activate this for yklua for example?

ltratt:
> How do we activate this for yklua for example?

It's enabled by default, so you don't need to do anything other than use this branch's version of yk/ykllvm for yklua.

ykrt/build.rs
vext01: Maybe don't worry about it.

ykrt/src/mt.rs
vext01: > because worker threads aren't necessarily active

Dumb question: what do we mean by an "active thread"?

Perhaps a worker thread doesn't exit after it has performed a task? It hangs around waiting for a new task?

vext01:
> Love it! Please squash.

Hold up, we are still discussing :)

ykrt/src/mt.rs
ltratt: Worker threads wait for jobs: so "active" means "a worker thread that's actively doing something", excluding "worker threads waiting for a job to do".

vext01: Right. Now I understand your prior comment. Threads don't join after a compilation task finishes, so a join handle doesn't really cut it.

vext01:
Please squash.

ltratt:
Squashed.

PR #900, opened 2023-11-09
Find successful passes with genetic algorithm

nmdis1999:
This PR aims to find successful passes (i.e which pass all YK and YKLUA tests) with the help of genetic algorithm. The fitness function currently gives priority to execution time (the lesser the better) to pick up most likely parents.


ltratt:
@nmdis1999 Does this carry over the comments I made on the other PR?

nmdis1999:
@ltratt All but one, I have to change while loop (with `i`) to for. Will push this after cleaning some lines in next commit.

nmdis1999:
@ltratt  I am capturing timing information in `cargo_run.py` (688283f0babc84b3c29268144fe01fa50d27af0e) and then feeding that in `try_passes.py`'s genetic algorithm. I wondering if this is the right way or is there a way to capture just runtime of `cargo test` without the compilation step. I tried running `cargo build` before running `cargo test` but it didn't seem to do much.

ykrt/pass_finder/cargo_run.py
```diff
@@ -1,22 +1,22 @@
...
     os.chdir("/home/shreei/research/yk_pv")
+    #subprocess.run(["cargo build"], shell=True, env=env or os.environ)
     times = []
     for i in range(n):
         before = time.time() 
-        try:
-            c = subprocess.run(["timeout 60 cargo test"], shell=True, env=env or os.environ) 
-        except:
-            print(f"\033[92m @@@@@@@@@@@@@@\033[0m")
+        c = subprocess.run(["timeout 60 cargo test"], shell=True, env=env or os.environ)
```
ltratt, line 11: Ah, I think I see the challenge! `cargo test` is building Rust code with normal (non-ykllvm) LLVM.

What we want to benchmark is, for example, yklua running some benchmarks, because yklua is compiled with ykllvm.

nmdis1999: Can we rely on yklua's test for this? I remember some of the tests were running fine but always end with a SEG fault.

ltratt: In your case, you actually don't want the JIT to kick in: you want to just test the interpreter. So you'll want to set `yk_mt_hot_threshold_set` to a ludicrously high value (e.g. `1<<28` or something) in yklua. Then you should hopefully find that most benchmarks work fine. @Pavel-Durov can tell you at least a few which are known to work (even with the JIT on!).

nmdis1999:
A bit of explanation how genetic algorithm is assigning weight to list of passes. Genetic algorithm first generates a list of passes which is then used to re-compile/build YK and run YKLUA test via `cargo_run.py`.  Only YKLUA's test runtime is considered for assigning weight (or fitness score). 

Since `cargo_run.py` calls `run.sh` from YKLUA repository it is not visible in this PR. But here is the content of `run.sh`:

```
#!/bin/sh

set -e

# Check if YKLUA_PATH is set and is a valid directory.
if [ -z "$YKLUA_PATH" ] || [ ! -d "$YKLUA_PATH" ]; then
    echo "YKLUA_PATH directory does not exist."
    exit 1
fi

cd $YKLUA_PATH
MODE=debug

if [ ! -z $PRELINK_PASSES ]; then
  echo "yk-here"
  yk-config ${MODE} --prelink-pipeline "${PRELINK_PASSES}" --cflags 
else
  yk-config ${MODE} --cflags
fi

if [ ! -z $LINKTIME_PASSES ]; then
  yk-config ${MODE} --postlink-pipeline "${LINKTIME_PASSES}" --ldflags
else
  yk-config ${MODE} --ldflags
fi

make clean && make YK_BUILD_TYPE=debug

cd benchmark
SECS=120
LUA=../src/lua

OK=""
FAIL=""

exstatus=0
YKD_SERIALISE_COMPILATION=0 timeout -s9 ${SECS}s ${LUA} fannkuchredux.lua 3 \
    >/dev/null 2>&1 || exstatus=$?

if [ ${exstatus} -eq 0 ]; then
    echo "OK"
    OK="${OK}"
else
    if [ ${exstatus} -eq 137 ]; then
        echo "TIMEOUT"
    else
        echo "FAIL"
    fi
    FAIL="${FAIL}"
fi

echo "\n---"
echo "OK: ${OK}"
echo "FAIL: ${FAIL}"
```

ykrt/pass_finder/cargo_run.py
```diff
@@ -1,22 +1,37 @@
...
-    #subprocess.run(["cargo build"], shell=True, env=env or os.environ)
-    times = []
-    for i in range(n):
-        before = time.time() 
-        c = subprocess.run(["timeout 60 cargo test"], shell=True, env=env or os.environ) 
-        elapsed = time.time() - before
-        times.append(elapsed)
-    mean_time = sum(times) / len(times)
+    os.chdir(yk_path)
+    r = subprocess.run(["timeout 50 cargo test"], shell=True, env=env or os.environ)
```
ltratt, line 11: How long does the test suite normally take to run? [I can see that we might need a `timeout` because a bad pass might cause it to get stuck in a loop but I wonder if we need a fairly generous timeout of 2x the normal running time?]

ykrt/pass_finder/cargo_run.py
```diff
@@ -1,22 +1,37 @@
...
+    # assert r.returncode == 0
+    if r.returncode == 0:
+        os.chdir("/home/shreei/research/yklua")
```
ltratt, line 14: Not convinced that this will work on my machine :p

ykrt/pass_finder/cargo_run.py
```diff
@@ -1,22 +1,37 @@
...
+        times = []
+        for _ in range(n):
+            subprocess.run(["make clean"], shell=True, env=env)
+            c = subprocess.run(["make && sh test.sh"], shell=True, env=env or os.environ)
```
ltratt, line 18: What's the difference between `test.sh` and `run.sh`? I guess we also need a `timeout sh test.sh` for the same reasons as above? And probably also `timeout sh run.sh` below?

ykrt/pass_finder/try_passes.py
```diff
@@ -369,6 +376,6 @@ def main(logf, glogf, is_prelink):
     CWD = args.path
     print(f"PATH to interpreter: {CWD}")
 
-    with open("genetic.log", "w+") as glogf:
-        with open("passes.log", "w") as logf:
+    with open("/home/shreei/research/yk_pv/ykrt/pass_finder/genetic.log", "w+") as glogf:
```
ltratt, line 379: Not convinced this line will work on my machine either ;)

ltratt:
@nmdis1999 If/when this has successfully found some passes, please let us know.

nmdis1999:
@ltratt I ran the algorithm for 100 generation (and population size * 2), this ran between 4-5 days for both PRELINK and POSTLINK pass list.
Here is the best pass list the algorithm dumped based on execution time of the benchmark run:

**PRELINK**
```
forceattrs,openmp-opt,called-value-propagation,globalopt,function(invalidate<aa>),deadargelim,globaldce,verify
```

**POSTLINK**
```
annotation2metadata,forceattrs,function<eager-inv>(lower-expect,simplifycfg<bonus-inst-threshold=1;no-forward-switch-cond;no-switch-range-to-icmp;no-switch-to-lookup;keep-loops;no-hoist-common-insts;no-sink-common-insts>,sroa<modify-cfg>,early-cse<>),openmp-opt,ipsccp,called-value-propagation,function(mem2reg), function<eager-inv>(instcombine,simplifycfg<bonus-inst-threshold=1;no-forward-switch-cond;switch-range-to-icmp;no-switch-to-lookup;keep-loops;no-hoist-common-insts;no-sink-common-insts>),require<globals-aa>,require<profile-summary>,cgscc(devirt<4>(inline<only-mandatory>,inline,function-attrs,openmp-opt-cgscc,function<eager-inv>(sroa<modify-cfg>,early-cse<memssa>,speculative-execution,jump-threading,correlated-propagation,simplifycfg<bonus-inst-threshold=1;no-forward-switch-cond;switch-range-to-icmp;no-switch-to-lookup;keep-loops;no-hoist-common-insts;no-sink-common-insts>,instcombine,libcalls-shrinkwrap,tailcallelim,simplifycfg<bonus-inst-threshold=1;no-forward-switch-cond;switch-range-to-icmp;no-switch-to-lookup;keep-loops;no-hoist-common-insts;no-sink-common-insts>,reassociate,require<opt-remark-emit>,loop-mssa(loop-instsimplify,loop-simplifycfg,licm<no-allowspeculation>,loop-rotate,licm<allowspeculation>,simple-loop-unswitch<no-nontrivial;trivial>),simplifycfg<bonus-inst-threshold=1;no-forward-switch-cond;switch-range-to-icmp;no-switch-to-lookup;keep-loops;no-hoist-common-insts;no-sink-common-insts>,instcombine,loop(loop-idiom,indvars,loop-deletion,loop-unroll-full),sroa<modify-cfg>,vector-combine,mldst-motion<no-split-footer-bb>,gvn<>,sccp,bdce,instcombine,jump-threading,correlated-propagation,adce,memcpyopt,dse,loop-mssa(licm<allowspeculation>),coro-elide,simplifycfg<bonus-inst-threshold=1;no-forward-switch-cond;switch-range-to-icmp;no-switch-to-lookup;keep-loops;hoist-common-insts;sink-common-insts>,instcombine),coro-split)),name-anon-globals
```

ykrt/pass_finder/cargo_run.py
nmdis1999: I intend to compile YKLUA and base test cases with test.sh first (this is just to make sure the passes can build and run some YKLUA test cases), `run.sh` is used to run the benchmark test we choose (with Pavel's help) and use that as fitness score metrics.
And thanks I think both of these could use `timeout`.

ykrt/pass_finder/cargo_run.py
nmdis1999: Agreed, so the `cargo test` doesn't run for more than 12-15 secs. Changing this to `timeout 20 cargo test`.

nmdis1999:
@ltratt @vext01 this PR is ready for review now.

ykrt/pass_finder/try_passes.py
```diff
@@ -185,18 +211,114 @@ def binary_split(logf, passes, is_prelink):
...
+def evaluate_fitness(glogf, is_prelink, entity, passes, cwd):
+    try_passes = []
+    for i, bit in enumerate(entity):
+        if bit == 1:
+            try_passes.append(passes[i])
+    log(glogf, f"\ncurrently evaluating {try_passes}\n")
+    config = get_pipeline_config(is_prelink, try_passes)
+    ret, exec_time = test_pipeline(glogf, config, cwd) 
+    if ret:
+            exec_time = float(exec_time)  # Convert exec_time to a float
```
vext01, line 223: I'd remove the comment. It states the obvious.

ykrt/pass_finder/try_passes.py
```diff
@@ -185,18 +211,114 @@ def binary_split(logf, passes, is_prelink):
...
+    for generation in range(generations):
+        log(glogf, "=========================================================")
+        log(glogf, f"\ngeneration: {generation}\n") 
+        fitness_scores.clear()
+        # Evaluate fitness for each entity in the population
+        fitness_scores = [evaluate_fitness(glogf, is_prelink, entity, passes, cwd) for entity in population]
+        log(glogf, f"\nfitness score: {fitness_scores}\n")
+        log(glogf, "=========================================================")
+
+        # less the execution time is better
```
vext01, line 261: This sentence doesn't quite parse. How about "A lower execution time is better"?

ykrt/pass_finder/try_passes.py
```diff
@@ -185,18 +211,114 @@ def binary_split(logf, passes, is_prelink):
...
+        wt = [(1/t) for t in fitness_scores] 
+        print(f"\033[38;5;128m {fitness_scores}\033[0m")
```
vext01, line 263: Is it worth defining the colours as named constants at the top of the file?

ykrt/pass_finder/try_passes.py
```diff
@@ -185,18 +211,114 @@ def binary_split(logf, passes, is_prelink):
...
+        
+        # Check if we have reached the target fitness
+        if target_fitness in fitness_scores:
+            print(f"Target fitness reached in generation {generation + 1}!")
+            break
+        
+        # to randomly pick entities when the complete population fails
```
vext01, line 270: This is a real nitpick, but some comments start with an uppsercase letter, others don't. Should we choose one and stick with it?

(if you decide to fix this, please only change comments already in the the change set for this PR)

ykrt/pass_finder/try_passes.py
```diff
@@ -185,18 +211,114 @@ def binary_split(logf, passes, is_prelink):
         print(72 * "=")
         print(list_of_passes_to_str(ok_passes))
 
-def main(logf, is_prelink):
+def evaluate_fitness(glogf, is_prelink, entity, passes, cwd):
```
vext01, line 214: Can we add a short docstring to all newly introduced functions? It will help our future selves.

vext01:
This is looking good. All of my comments are minor.

ykrt/pass_finder/cargo_run.py
```diff
@@ -0,0 +1,59 @@
...
+        raise ValueError("YKLUA_PATH environment variable is not set")
+
+    if env is None:
+        env = os.environ.copy()
+
+    env['PATH'] = f"{yk_path}/bin:{env['PATH']}"
+    env['YK_BUILD_TYPE'] = 'debug'
+     
+    # Check if pre and post-link flags are set
+    prelink_passes = (env or os.environ).get('PRELINK_PASSES', 'Not Set')
```
ltratt, line 22: At this point `env` is guaranteed to be non-None, so not sure why the `or` is needed.

ykrt/pass_finder/cargo_run.py
```diff
@@ -0,0 +1,59 @@
...
+    curdir = os.getcwd()
+    os.chdir(yklua_path)
+
+    times = []
+    c_test = None 
+
+    for _ in range(n):
+        subprocess.run(["make clean"], shell=True, env=env)
+        c = subprocess.run(["make && timeout 30 sh test.sh"], shell=True, env=env or os.environ)
+        print(f"\033[95m returncode for c is {c.returncode} \033[0m\n")
```
ltratt, line 36: In general I'm allergic to chatty scripts as we tend to end up not being able to see the forest for the trees. I also suspect that we should be checking `c.returncode` here before bothering with `cargo test`? Running the latter if the former failed seems pointless?

ykrt/pass_finder/cargo_run.py
```diff
@@ -0,0 +1,59 @@
...
+       
+        os.chdir(yk_path)
+        r = subprocess.run(["timeout 50 cargo test"], shell=True, env=env or os.environ)
+        os.chdir(yklua_path) 
+
+        if c.returncode == 0 and r.returncode == 0:
+            before = time.time()
+            c_test = subprocess.run(["timeout 8 sh run.sh"], shell=True, env=env or os.environ)
```
ltratt, line 44: We should check the returncode, because if this `timeout`s, or fails for other reasons, we won't notice.

ykrt/pass_finder/cargo_run.py
```diff
@@ -0,0 +1,59 @@
...
+            elapsed = time.time() - before
+            times.append(elapsed)
+
+    if len(times) != 0:
+        mean_time = sum(times) / len(times)
+    else:
+        mean_time = 0
```
ltratt, line 51: What does a mean time of zero mean for this program?

ykrt/pass_finder/cargo_run.py
vext01: If Iti finds these messages useful for debugging and development, perhaps there should be a verbose flag (off by default)?

ykrt/pass_finder/try_passes.py
```diff
@@ -43,7 +43,7 @@ def __str__(self):
         link_time = ",".join([ str(p) for p in self.link_time ])
         return f"PipelineConfig(pre_link=[{pre_link}], link_time=[{link_time}])"
 
-def get_pipeline_config(is_prelink, ok_passes, try_passes):
+def get_pipeline_config(is_prelink, try_passes, ok_passes=[]):
```
ltratt, line 46: We can't use `[]` is a default argument in Python (well, unless they've finally fixed it in Python 3?), because it's a singleton that's shared amongst all calls to the function. The standard idiom is:

```python
def f(x = None):
  if x = None: x = []
```

vext01: They didn't fix this in Python 3:

```python
In [2]: def f(x=[]):
   ...:     x.append(1)
   ...:     print(x)
   ...: 

In [3]: f()
[1]

In [4]: f()
[1, 1]
```

It's OK as long as you don't mutate it, but I do tend to agree, that it's best to not pass lists at all.

ykrt/pass_finder/try_passes.py
```diff
@@ -66,53 +66,79 @@ def log(logf, s):
...
 
-def get_all_passes(is_prelink):
-    cmd = get_opt_cmd(is_prelink) 
-    p = Popen(cmd, shell=True, stdout=PIPE, close_fds=True)
-    sout, serr = p.communicate() 
-    assert(p.returncode == 0)
+def split_passes(passes_string):
+    parts = []           
+    temp_part = []      
+    paren_stack = []
```
ltratt, line 72: I think `paren_stack` is a bool in disguise. At most it's an interger in disguise because the values stored in this list aren't checked. Personally I would just have an integer `nesting = 0` and inc/dec it for both types of brackets. If someone gives an invalid string (e.g. switches closed brackets around), this will not work properly (but nor does the current code), but that can come under the "someone else shouldn't do that" category :)

ykrt/pass_finder/try_passes.py
```diff
@@ -66,53 +66,79 @@ def log(logf, s):
...
-    p = Popen("try_repeat 10 sh run_tests.sh 2>&1", cwd=CWD, shell=True,
-              stdout=PIPE, close_fds=True, env=env)
+    env["POSTLINK_PASSES"] = ",".join([p.name for p in pl.link_time])
     
-    sout, _ =  p.communicate()
-    log(logf, sout.decode())
+    ret, time = cargo_run.run_test(cwd, env=env)
 
-    if p.returncode == 0:
+    print(f"\033[92m return code for cargo_run: {ret} \033[0m")
```
ltratt, line 132: Same comment as earlier.

ykrt/pass_finder/try_passes.py
ltratt: I'd also like that!

ykrt/pass_finder/try_passes.py
```diff
@@ -185,18 +211,114 @@ def binary_split(logf, passes, is_prelink):
         print(72 * "=")
         print(list_of_passes_to_str(ok_passes))
 
-def main(logf, is_prelink):
+def evaluate_fitness(glogf, is_prelink, entity, passes, cwd):
+    try_passes = []
+    for i, bit in enumerate(entity):
```
ltratt, line 216: I think this can become something like:

```python
try_passes = [passes[i] for (i, x) in enumerate(entity) if x]
```

Warning: I might be wrong!

ykrt/pass_finder/try_passes.py
```diff
@@ -185,18 +211,114 @@ def binary_split(logf, passes, is_prelink):
...
+        fitness_scores = [evaluate_fitness(glogf, is_prelink, entity, passes, cwd) for entity in population]
+        log(glogf, f"\nfitness score: {fitness_scores}\n")
+        log(glogf, "=========================================================")
+
+        # less the execution time is better
+        wt = [(1/t) for t in fitness_scores] 
+        print(f"\033[38;5;128m {fitness_scores}\033[0m")
+        
+        # Check if we have reached the target fitness
+        if target_fitness in fitness_scores:
```
ltratt, line 266: Shouldn't this be a `>=` check rather than the (implicit) `==` it is now?

ykrt/pass_finder/try_passes.py
```diff
@@ -185,18 +211,114 @@ def binary_split(logf, passes, is_prelink):
...
+            print(f"Target fitness reached in generation {generation + 1}!")
+            break
+        
+        # to randomly pick entities when the complete population fails
+        if fitness_scores.count(float('inf')) == len(fitness_scores):
+            wt = [1 for _ in fitness_scores]
```
ltratt, line 272: Might be clearer as `wt = [1] * len(fitness_scores)`?

ykrt/pass_finder/try_passes.py
```diff
@@ -185,18 +211,114 @@ def binary_split(logf, passes, is_prelink):
...
+
+        # Select parents for reproduction (roulette wheel selection)
+        parents = []
+        for _ in range(population_size // 2):
+            parent1 = random.choices(population, weights=wt, k=1)[0]
+            parent2 = random.choices(population, weights=wt, k=1)[0]
```
ltratt, line 278: Is it a requirement for roulette wheel that the parents are different? I dunno!

ykrt/pass_finder/try_passes.py
```diff
@@ -206,23 +328,29 @@ def main(logf, is_prelink):
...
-    CWD = args.path
+    yk_path = os.environ.get('YK_PATH')
+    if yk_path is None:
+        raise ValueError("YK_PATH environment variable is not set")
+
+    yklua_path = os.environ.get('YKLUA_PATH')
+    if yklua_path is None:
+        raise ValueError("YKLUA_PATH environment variable is not set")
+
+
```
ltratt, line 349: Double blank line (I know, I know, I'm a nit-picker).

ykrt/pass_finder/try_passes.py
nmdis1999: we'll need seperarte entity for crossover and mutation. That's why they are different.

ykrt/pass_finder/try_passes.py
nmdis1999: this is already handled, when returncode `ret` is non zero `test_pipeline` returns `False, None`

ykrt/pass_finder/cargo_run.py
nmdis1999: mean time zero means one of the returncode's (checks) is non-zero (aka failed).

ykrt/pass_finder/try_passes.py
nmdis1999: @vext01 do we want docstring in all of the functions or only the genetic algorithm ones?

vext01: For now, only new functions you've added in this PR.

If you want to add others, then that should be a separate PR IMHO.

ykrt/pass_finder/try_passes.py
nmdis1999: addressed here: 8dab2e70489ece37fe6efa8ab40983248dc19b1d

ykrt/pass_finder/try_passes.py
nmdis1999: addressed here: 8dab2e70489ece37fe6efa8ab40983248dc19b1d

ykrt/pass_finder/try_passes.py
nmdis1999: Addressed here: 8dab2e70489ece37fe6efa8ab40983248dc19b1d

ykrt/pass_finder/cargo_run.py
nmdis1999: Addressed here: 6c3f62f2dd481f0a792a6e681a1d4f0fe6b336fc

ykrt/pass_finder/cargo_run.py
nmdis1999: Addressed here: 6c3f62f2dd481f0a792a6e681a1d4f0fe6b336fc

ykrt/pass_finder/cargo_run.py
nmdis1999: Addressed here: 6c3f62f2dd481f0a792a6e681a1d4f0fe6b336fc

ykrt/pass_finder/try_passes.py
nmdis1999: Addressed: 6c3f62f2dd481f0a792a6e681a1d4f0fe6b336fc

ykrt/pass_finder/try_passes.py
```diff
@@ -66,53 +66,79 @@ def log(logf, s):
...
+    if part:  
+        parts.append(part)
 
-    sout = sout.decode()
-    pass_descrs = [x.strip() for x in sout.strip().split("\n")]
+    return parts
 
+def get_all_passes(is_prelink):
+    cmd = get_opt_cmd(is_prelink)
+    sout = subprocess.run(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.DEVNULL)
```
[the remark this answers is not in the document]
nmdis1999: Addressed: 6c3f62f2dd481f0a792a6e681a1d4f0fe6b336fc

ykrt/pass_finder/try_passes.py
nmdis1999: Done: 6c3f62f2dd481f0a792a6e681a1d4f0fe6b336fc

ykrt/pass_finder/try_passes.py
nmdis1999: Done: 6c3f62f2dd481f0a792a6e681a1d4f0fe6b336fc

ykrt/pass_finder/try_passes.py
nmdis1999: Addressed: 6c3f62f2dd481f0a792a6e681a1d4f0fe6b336fc

ykrt/pass_finder/try_passes.py
nmdis1999: Addressed here: 0ebf4e6be0f59bb4b28fc16e91db87d6dcbe50f7

ykrt/pass_finder/try_passes.py
nmdis1999: Addressed here: afb33c3ac820c5a7164301dfcdc43ea4d1835fd0

ykrt/pass_finder/try_passes.py
nmdis1999: Addressed here: 7e2186b965e3a4b2fff39a563ffcf1a3cc36c8db

nmdis1999:
@ltratt @vext01 addressed all the comments. Ready for review :)

ykrt/pass_finder/try_passes.py
ltratt: So this `print` can't ever fire?

ykrt/pass_finder/try_passes.py
ltratt: Since they're randomly chosen, can't `parent1` and `parent2` end up being the same value?

ltratt:
We're getting there! @nmdis1999 Please have a quick look over the open chats (including one or two from last year). With luck we can get this moving soon (once we have some rough numbers from your benchmarking run).

ykrt/pass_finder/try_passes.py
nmdis1999: I have removed this `print` one of the commits I pushed yesterday. The returncode is being check in the next line with `if ret == 0:`

ykrt/pass_finder/cargo_run.py
nmdis1999: This has been addressed.

ykrt/pass_finder/try_passes.py
nmdis1999: This too.

ykrt/pass_finder/try_passes.py
nmdis1999: They can but not always (if I dare to say), however I think it is important for us to have `parent1` and `parent2` because the `crossover` function finds a pivot and takes first section (till pivot point) from `parent1` and rest from `parent2`. 
Also, these parents are appended in a parents list, which is then used to get children by crossover-ing and mutating.
I am not sure how we are supposed to operate `crossover` function if both parent are always same.

ykrt/pass_finder/cargo_run.py
ltratt: I don't see anything in https://github.com/ykjit/yk/commit/7e2186b965e3a4b2fff39a563ffcf1a3cc36c8db about paths with `shreei`?

ykrt/pass_finder/try_passes.py
ltratt: https://github.com/ykjit/yk/commit/0ebf4e6be0f59bb4b28fc16e91db87d6dcbe50f7 is better but i think you can simplify this to just `nesting` --- you don't gain anything at the moment from checking the nesting of parens and angle brackets separately.

ykrt/pass_finder/try_passes.py
ltratt: I'm pretty sure the change is incorrect. It's now:

```python
if target_fitness >= fitness_scores:
```

but I doubt that runs without a type error and, if it does, it won't do what we want.

What you want, I think, is to say "is there a fitness that's >= target_fitness". `any` is one way of doing this e.g.:

```python
>>> x=[1,2,3,4]
>>> any(y > 4 for y in x)
False
>>> any(y >= 4 for y in x)
True
```

nmdis1999: Addressed here: d2c78c9eba0a7d94101a39cef66404217c8e5306

ykrt/pass_finder/try_passes.py
nmdis1999: Addressed here: d2c78c9eba0a7d94101a39cef66404217c8e5306

ykrt/pass_finder/cargo_run.py
nmdis1999: Addressed here: d2c78c9eba0a7d94101a39cef66404217c8e5306

ykrt/pass_finder/cargo_run.py
nmdis1999: I couldn't find which commit I solved it it, but there is no hardcoded path anymore.

nmdis1999:
@ltratt addressed your comments. Ready for review/

ykrt/pass_finder/cargo_run.py
ltratt: If you can't find the commit, how do we know?

ykrt/pass_finder/try_passes.py
ltratt: So if the parents must be different, we have to do something to make that happen. For example, select parent1 then in a loop select a parent2 until parent2 != parent1.

ykrt/pass_finder/cargo_run.py
nmdis1999: Sorry, I found the commit: b179d7cd38f5c06aee45e10b01b907aac31ac5e3

ykrt/pass_finder/try_passes.py
nmdis1999: Changed here: 0edbc0d87347d1d384b8dd4545f5e0d77115fbcf

ltratt:
OK, so code review looks reasonable -- all we need to know now is whether it works! So we'll need performance numbers in here as and when you have them.

ltratt:
@nmdis1999 Please undo the force push (ironically you'll need to force push in order to do that!).

ltratt:
@nmdis1999 Quick questions:

1. Why sometimes CI and sometimes std dev? Let's pick one (ideally CI) and stick with it.
2. Can you define what "WITHOUT YKLLVM" (etc) mean? I can guess but I'm not 100% sure, so best that you tell me!
3. What is the best prelink and/or postlink set(s) of passes you've found so far?

tests/Cargo.toml
```diff
@@ -36,7 +36,7 @@ yktracec = { path = "../yktracec", features = ["yk_testing"] }
 
 [dev-dependencies]
 criterion = { version = "0.5.1", features = ["html_reports"] }
-lang_tester = "0.7.4"
+lang_tester = { version = "0.7.4", path = "../../lang_tester" }
```
ltratt, line 39: You'll need to make this point to the branch in Edd's github fork. Cargo has the settings you need to do this.

ykrt/pass_finder/cargo_run.py
```diff
@@ -25,20 +25,19 @@ def run_test(cwd, id, yk_path, yklua_path, env, n=1):
...
     times = []
     c_test = None 
-
     for _ in range(n):
         subprocess.run(["make clean"], shell=True, env=env)
         c = subprocess.run(["make && timeout 30 sh test.sh"], shell=True, env=env or os.environ)
         os.chdir(yk_path)
         if c.returncode == 0:
-            r = subprocess.run(["timeout 60 cargo test"], shell=True, env=env or os.environ)
+            r =  subprocess.run(f"timeout 60 cargo test", shell=True, env=env or os.environ)
```
ltratt, line 40: The extra space after `=` is unneccessary.

ykrt/pass_finder/setup_genetic.py
```diff
@@ -0,0 +1,71 @@
...
+                subprocess.run(f"sed -i -e 's/PRELINK_PASSES/PRELINK_PASSES_{i}/g' Makefile", shell=True) 
+                subprocess.run(f"sed -i -e 's/POSTLINK_PASSES/POSTLINK_PASSES_{i}/g' Makefile", shell=True)  
+
+            temp_directories.extend(temp_dir)
+            os.chdir(curr_dir)
+
+def setup(curr_dir, base_temp_dir, yk_path, yklua):
+    num_cores = multiprocessing.cpu_count() - 1
+    directories = []
+    # for i in range(num_cores):
```
ltratt, line 50: Commented code.

ltratt:
https://github.com/ykjit/yk/pull/900/commits/f42221a676821a6be679076e0b06c77b9722bd12 is going to be a problem: we can't commit the PR with those changes, because they break "normal" yk. @vext01 do you have any thoughts on how we could handle this in the master branch? It *might* be that we end up having to do this stuff in a branch/fork for a while? [I'd prefer to merge into master, but aware that it might be tricky.]

tests/Cargo.toml
vext01: Hold on, we don't want to merge that into `yk/master`. It fudges lang_tester to not match stdin/stdout.

What's the plan here?

vext01:
I suppose we could use a cargo feature for the kludged lang_tester?

ltratt:
It's not just lang_tester though: notice that this branch also disables a test *and* disables code in ykcapi.

vext01:
I suppose that could be feature gated too... I'm not really sure what the best way forward here is. gate or branch...

ltratt:
Please force push a rebase + squash (doesn't have to be to 1 commit though) against master.

Issue #903, opened 2023-11-15
Assertion failure in completion code


vext01:
See https://github.com/ykjit/yk/pull/902

vext01:
For the record, I ran the CI jobs in fresh docker instances overnight on our beefy machine. I got 30 runs with no completion crash. Hrm...

ltratt:
Fixed in #910.

PR #907, opened 2023-11-17
Reenable completion

vext01:
I can't reproduce #903 locally or via repeated forced CI jobs. This makes it hard for me to debug.

Based on the bug report I have, it would appear that we see a compiler invocation with a source file that has no extension. I don't see how that could happen.

This change re-enables the completion stuff and adds better error messages, so that if it crops up again we get some information that might indicate what's going on.

### DIFF
--- a/ykbuild/src/completion_wrapper.rs
+++ b/ykbuild/src/completion_wrapper.rs
@@ -93,9 +93,17 @@ impl CompletionWrapper {
 
             // We assume (and assert) that the source file is the last argument.
             let ccfile = buf.split(' ').last().unwrap();
-            assert!(["c", "cpp", "cxx", "cc"]
-                .iter()
-                .any(|e| e == &Path::new(ccfile).extension().unwrap().to_str().unwrap()));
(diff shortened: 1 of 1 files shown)


ltratt:
@Pavel-Durov was able to replicate this with a docker command doo-dah, so I think we should ask him how he did it and work from there. Since this bug has definitely happened multiple times including multiple days of failing nightly builds IIRC, I'm decidedly unkeen on merging this PR (sorry!).

Pavel-Durov:
I was able to reproduce it before (this Tuesday).

BUT today I tried and failed to reproduce it in the same way, based on https://github.com/ykjit/yk/commit/cda769ae01b4a0ffe255e815703fce7f0c3dc19d (before the merge of https://github.com/ykjit/yk/commit/252828404e128f4611692cee3e6febf37bc1a992)

At first, I thought that it might be related to the docker container version -  we don't use explicit image versioning in out docker files (we use `debian:bookworm`).

But the last update was 3 weeks ago (`bookworm-20231030`) so it is probably not that.

Not sure what's going on 😶

PR #911, opened 2023-11-23
Replace unsound StackAdjust code.

ptersilie:
With the recent change to ykllvm, which ensures no calls can appear in the entry block of a function, we no longer need to rely on the StackAdjust solution in order to detect callbacks from external functions (i.e. unmappable code). Instead, we now only need to check if an unmappable block is followed by an entry block. If it is, then the external function called back into mappable code; if not then we have returned from the external function.

### DIFF
--- a/yktracec/src/jitmodbuilder.cc
+++ b/yktracec/src/jitmodbuilder.cc
@@ -1426,73 +1426,42 @@ public:
       TraceLoc Loc = InpTrace[Idx];
 
       if (UnmappableRegion *UR = Loc.getUnmappableRegion()) {
-        // The trace entered a region of unmappable foreign code.
-        //
-        // As noted in the mapper and asserted in the JITModBuilder constructor,
-        // there are never two unmappable regions in a row, so the next block
-        // (if there is one) is guaranteed to be mappable. The question is: will
-        // the foreign code be *returning* into mappable code, or will it be
-        // *calling back* deeper into mappable code? The trace decoder keeps
-        // track of the stack effects of foreign code (assuming its control flow
-        // isn't bonkers), so we are able to use that information to decide.
-
-        if (InpTrace.Length() == Idx + 1) {
-          // This unmappable region is the end of the trace, so there's no
-          // need to maintain the stack any more.
-          continue;
-        }
-
-        assert(!CallStack.curMappableFrame());
-
-        // FIXME: think about what do do in the face of setjmp/longjmp. How
-        // would we find the stack adjustment value for such code?
-        if (UR->StackAdjust < 0) {
-          // The stack got smaller as a result of executing the foreign code,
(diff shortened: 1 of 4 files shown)


ltratt:
Does this mean we can also remove the `stackAdjust` field definition from the relevant class?

yktracec/src/jitmodbuilder.cc
```diff
@@ -1426,73 +1426,40 @@ class JITModBuilder {
...
-            assert(!CallStack.curMappableFrame());
+      IRBlock *IB = Loc.getMappedBlock();
+      assert(IB);
+      CurBBIdx = IB->BBIdx;
+
+      auto [F, BB] = getLLVMAOTFuncAndBlock(IB);
+
+      if (CallStack.curMappableFrame() == nullptr) {
+        if (!BB->isEntryBlock()) {
+          // The current block is not an entry block and we have just seen an
```
vext01, line 1440: Should we state the property of the IR that allows us to make this deduction: that an entry block can contain no calls (aside from potentially mallocing the shadow stack)?

ptersilie:
> Does this mean we can also remove the `stackAdjust` field definition from the relevant class?

Yes, I was going to do that in a follow up PR because there's a couple of places this needs to be ripped out of.

ltratt:
> Yes, I was going to do that in a follow up PR because there's a couple of places this needs to be ripped out of.

OK, I'm neutral about whether it's a second commit here or another PR. Please do as you see fit.

yktracec/src/jitmodbuilder.cc
ptersilie: [377beb7](https://github.com/ykjit/yk/pull/911/commits/377beb7322c9de0da57358f2f554fafd68e4a639)

ptersilie:
> OK, I'm neutral about whether it's a second commit here or another PR. Please do as you see fit.

We need to remove stuff from ykllvm as well, so it makes more sense as a separate PR pair.

yktracec/src/jitmodbuilder.cc
vext01: One small nitpick.

> This only works when `--yk-no-calls-in-entryblocks` flag is used

I think we should something more like:

> We rely on the `NoCallsOnEntryBlocks` pass to provide the necessary disambiguation for this to work: Since no calls can appear in an entry block... <explain>...

ptersilie: [e44242c](https://github.com/ykjit/yk/pull/911/commits/e44242cdc30c940f1d8c5362aa574a46cda52c82)

ltratt:
If @vext01 is OK with this, please squash.

vext01:
Please squash.

ptersilie:
Squashed.

PR #916, opened 2023-11-27
Don't conflate a call with an unknown target with an indirect call.

### LINKED ISSUE #874: Bug in compressed return logic?

vext01:
You can call an external symbol (with an address statically unknown) directly.

Requires a compiler change, ~coming soon~: https://github.com/ykjit/ykllvm/pull/106

I had hoped this might fix #874, but alas no. Nonetheless, this is more correct and worth including.

### DIFF
--- a/hwtracer/src/llvm_blockmap.rs
+++ b/hwtracer/src/llvm_blockmap.rs
@@ -43,6 +43,8 @@ pub struct CallInfo {
     return_off: u64,
     /// Offset of the target of the call (if known statically).
     target_off: Option<u64>,
+    /// Indicates if the call is direct (true) or indirect (false).
+    direct: bool,
 }
 
 impl CallInfo {
@@ -56,6 +58,10 @@ impl CallInfo {
 
     pub fn target_off(&self) -> Option<u64> {
         self.target_off
+    }
+
+    pub fn is_direct(&self) -> bool {
+        self.direct
     }
 }
 
@@ -161,10 +167,12 @@ impl BlockMap {
                     debug_assert!(callsite_off < return_off);
                     let target = crsr.read_u64::<NativeEndian>().unwrap();
                     let target_off = if target != 0 { Some(target) } else { None };
+                    let direct = crsr.read_u8().unwrap() == 1;
                     call_offs.push(CallInfo {
                         callsite_off,
                         return_off,
                         target_off,
+                        direct,
                     })
                 }
 
--- a/hwtracer/src/pt/ykpt/mod.rs
+++ b/hwtracer/src/pt/ykpt/mod.rs
@@ -282,20 +282,20 @@ impl<'t> YkPTBlockIterator<'t> {
             let target = call_info.target_off();
 
             if let Some(target_off) = target {
-                // This is a direct call.
+                // The address of the callee is known.
                 //
                 // PT won't compress returns from direct calls if the call target is the
                 // instruction address immediately after the call.
                 //
                 // See the Intel Manual, Section 33.4.2.2 for details.
-                if target_off != call_info.return_off() {
+                if !call_info.is_direct() || target_off != call_info.return_off() {
                     self.comprets
                         .push(CompRetAddr::AfterCall(call_info.callsite_off()));
                 }
                 self.cur_loc = ObjLoc::MainObj(target_off);
                 return Ok(Some(self.lookup_block_from_main_bin_offset(target_off)?));
             } else {
-                // This is an indirect call.
+                // The address of the callee isn't known.
                 self.comprets
                     .push(CompRetAddr::AfterCall(call_info.callsite_off()));
                 self.seek_tip()?;


ltratt:
Can we add a test for this, either in ykllvm or yk?

ltratt:
Please squash.

vext01:
splat.

vext01:
During benchmarking:

```
`Result::unwrap()` on an `Err` value: Temporary(TraceBufferOverflow)
```

retry?

ltratt:
Bad news: same failure.

Good news: same failure. Which suggests it might be deterministic and caused by this PR.

vext01:
More general question then: what should we do if the benchmark fails with a buffer overflow?

I guess it's ok? We just want to check it runs somewhat, right?

ltratt:
In general, I'd say "buffer overflow" means one of:

1. We've been really, really, really unlucky with the trace in a particular execution.
2. The trace buffer size is too small.
3. The benchmark isn't really suitable for CI.
4. We've hit a bug.

In this case my working theory is that #4 is closest to the truth. If, however, we just completely suppress "buffer overflow"s we'll never know which of the 4 we might be encountering.

vext01:
What's confusing is, to my knowledge, there should be no functional change for this PR. The only time I've seen a direct call to an unknown target is via emulated TLS, which isn't in-tree yet.

Will investigate.

vext01:
I've just reviewed the code.

The benchmark in question is this:
https://github.com/ykjit/yk/blob/d78120e32087186a2bea89492ed8ee531729c83e/tests/benches/collect_and_decode.c#L31-L38

We trace a max of 50 iterations of this loop and the PT aux buffer size is 64MB.

Nothing that I can see springs out as being buggy.

I don't know how malloc works in detail, but is it possible that we trace a long compaction in the allocator or something like that?

64MB is still fairly small as far as a PT buffer goes.

A rough benchmark of PT throughput:

```
$ time perf record -e intel_pt// python3 -c 'for i in range(10000000): pass'
[ perf record: Woken up 352 times to write data ]
[ perf record: Captured and wrote 354.559 MB perf.data ]
0.33user 0.43system 0:01.16elapsed 66%CPU (0avgtext+0avgdata 43656maxresident)k
0inputs+726216outputs (0major+3751minor)pagefaults 0swaps
$ du -h  perf.data
355M    perf.data
```

So about 300MB/s of PT payload.

A 64 MB buffer would allow about 0.2 seconds of tracing.

I'd have thought that 50 iterations of the above loop could be done in that time...

vext01:
Shall we try doubling the buffer size?

ltratt:
OK.

vext01:
done.

This commit should stay separate.

ltratt:
Oh, ah. I thought you were going to do it just for the benchmark in question. I wonder if that's doable or not? Because allocating 128MiB is a non-trivial cost.

vext01:
It's doable, if we add an C API call to set the aux buffer size. Do we want to go there?

vext01:
Just looking at how `yk_mt_hot_threshold_set()` works to emulate it for a `yk_mt_trace_buffer_size_set()`.

Obviously such an option would be PT-specific, yet `MT` is generic.

So is it OK to add a setting and say it does nothing if the PT backend isn't used? What do you think?

ltratt:
Do we know that doubling the buffer size actually fixes the problem for this benchmark? The fact that we didn't see the problem before this PR is a bit odd.

vext01:
We don't. But we also don't know the state of the CI server. It could have been loaded or something.

vext01:
There's no `bors try` any more, so I've forced a build:
https://ci.soft-dev.org/#/builders/1/builds/3011

vext01:
I've reverted the buffer size doubling.

We now know this PR is not the cause of the non-deterministic failure. I'm fixing that elsewhere.

If I squash this, can we retry and see if we can get it in?

vext01:
Squashed.

vext01:
With #917 in, hopefully this will merge now...

PR #918, opened 2023-12-04
Switch to ORC for LLVM-based code generation.

vext01:
This initial iteration leaks the JITed code and stackmaps. To do the correct thing we would store the `LLJIT` object (freshly released from its `unique_ptr` so as to stop it dropping) and carry it around until we want to delete the underlying memory (probably inside `CompiledTrace::drop`).

(Note that even with that fix, all JITted code is still immortal, because we store code in a way that Drop won't run anyway)

This change also disables code generation for the `trace_compiler` suite. I was having a hard time getting it to link, but then realised that for what this suite is designed to do, no code generation is required. In turn some other hacks could go.

`GlobalMappings` appears to not be needed any more. ORC seems to be able to find symbols in the current address space automatically. It's a largeish change, so I think we should kill this in a follow-up PR.

For more information on ORC, see:
https://llvm.org/docs/ORCv2.html


vext01:
I should check yklua still works with this change. Please wait until I have before merging.

yktracec/src/stackmap_oll_plugin.cc
```diff
@@ -0,0 +1,59 @@
+// LLVM ORC Object Linking Layer plugin for preserving and locating stackmaps.
+
+#include "stackmap_oll_plugin.h"
+
+#ifdef __linux__
```
ltratt, line 5: Presumably we need an `#else #error unsupported platform` or similar, to stop "undefined variable SMSectionName" confusion?

vext01:
> I should check yklua still works with this change

Our `test.sh` still works just fine.

yktracec/src/stackmap_oll_plugin.cc
vext01: Fixed in 21d0acad

ltratt:
Please squash.

vext01:
splat.

PR #919, opened 2023-12-05
Remove `globalmappings` and `faddrs` and associated parts

ltratt:
This PR gradually removes `faddrs` and friends. This turned out to be a bit more subtle than I first expected, partly because of some unexpected function-type-inference in C++ code. After one botched attempt, I found it easiest to do this in bite-sized chunks: I suspect that also makes things easier to review, so I've left each chunk as a (fully `cargo test`ed) commit.


ltratt:
> Having symbol addresses discovered by PT will avoid dlsym()s in the new code generator (I think), right?

Since we've removed `globalMappings` by going to Orc, `faddrs` is unused. At least AFAICS.

I must admit, I would really like to do a full `bors try` to check the whole CI suite: I've `cargo test`ed everything, but that's not all the tests I'd liked to have run.

vext01:
> faddrs is unused

For now yes, but I don't know how the new code generator will do calls. If (like MCJIT) dynasm (or whatever we use) can't resolve these symbols, we will need to provide the callee addresses ourselves.

I agree that we can kill `GlobalMappings` and the bit where we pass function addresses into jitmodbuilder, but we might want to keep the bit where a PT trace tells us the function addresses it discovered?

vext01:
> we might want to keep the bit where a PT trace tells us the function addresses it discovered?

since you didn't change hwtracer, i guess we are already leaving that in?

ltratt:
> but we might want to keep the bit where a PT trace tells us the function addresses it discovered?

Other backends won't discover this automatically, and since we don't even use the information at the moment, I'm not able to tell what we gain by keeping this around. I'm willing to believe that there is a good reason, but you'll have to spell it out to me in idiot terms!

vext01:
I've just reviewed the code and it doesn't work how I remembered!

I thought that ykpt told us the callee addresses, but no. The trace requires interpretation by a trace builder to know where the calls are so actually I think what you've done is probably right after all.

(We could teach the decoder to tell us about callee addresses later, but for now, it doesn't communicate that)

ltratt:
> since you didn't change hwtracer, i guess we are already leaving that in?

Yeah, at least for now, because I can imagine non-yk consumers *might* want that information in.

vext01:
Glad to see so much unused guff disappear actually :)

PR #920, opened 2023-12-06
Fix Clippy warnings.

ltratt:
Most of these, to my surprise, turned out to be unused imports. I've therefore added a check in CI so that we won't re-add these, or others, in the future.


vext01:
disk on CI server was full of old docker caches. cleaning...

PR #924, opened 2023-12-07
Start nudging us towards "trace decoding as an iterator".

ltratt:
This PR is partly a sanity check before I go further. In essence, it introduces a `TraceIterator` which in essence replaces `RawTrace`. Currently the `hwt` backend still does its mapping in a single step, but it at least exposes an iterator (even if it doesn't make much use of it).

This commit is clearly a half-way house to what we might eventually want. In particular it doesn't yet support "trace mapping went wrong half-way through" because `hwt` can't go wrong in that way -- yet. But, hopefully, it is a nudge in the right direction. It also slightly simplifies the existing code as well as making the API more flexible.

### DIFF
--- a/tests/c/unmapped_setjmp.c
+++ b/tests/c/unmapped_setjmp.c
@@ -5,8 +5,7 @@
 //     jit-state: start-tracing
 //     set jump point
 //     jumped!
-//     jit-state: stop-tracing
-//     jit-state: trace-compilation-aborted
+//     jit-state: stop-tracing-aborted
 //     ...
 
 // Check that we bork on a call to setjmp in unmapped code.
--- a/tests/c/setlongjmp.c
+++ b/tests/c/setlongjmp.c
@@ -4,8 +4,7 @@
 //   env-var: YKD_PRINT_JITSTATE=1
 //   stderr:
 //     ...
-//     jit-state: stop-tracing
-//     jit-state: trace-compilation-aborted
+//     jit-state: stop-tracing-aborted
 //     ...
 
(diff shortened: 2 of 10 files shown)


tests/c/unmapped_setjmp.c
```diff
@@ -5,8 +5,7 @@
 //     jit-state: start-tracing
 //     set jump point
 //     jumped!
-//     jit-state: stop-tracing
-//     jit-state: trace-compilation-aborted
+//     jit-state: stop-tracing-aborted
```
vext01, line 8: "stop tracing aborted" is confusing to me.

It sounds like we were unable to stop tracing, and are therefore still tracing.

Unless there's a good reason to, I'd keep the old behavior.

ltratt: There is a good reason: both the old message *and* the new message are wrong! "Trace compilation" wasn't aborted previously: mapping went wrong (not compilation). However, at the moment we can't tease that error out separately. So I don't like the new message, but it's equivalent incorrect to the old message, so my dislike remains, but it has neither grown or shrink ;)

vext01: agreed (perhaps with `start-trace-compilation`/`stop-trace-compilation`/`trace-compilation-aborted(mapfail)` for consistency with the `start`/`stop` scheme we have in place.)

If you plan to change this name very soon, I'm OK with this in the interim.

ltratt: I do plan to do so soon(ish), assuming you're happy with the overall direction of this PR.

PR #925, opened 2023-12-08
Rewrite jitmodbuilder inlining/outlining logic.

ptersilie:
With https://github.com/ykjit/ykllvm/pull/108 it is now easier to detect recursion and foreign callbacks:

In order to detect when to inline or outline a function call in a trace, we previously relied on a complicated call stack that tracked at which instruction we started outlining (or inlining) and whether we've processed an unmappable block or not. Returning from a call was also difficult due to the fact functions calls aren't terminators and we had to continue tracing within a duplicate block but had to make sure we continue after the instruction that started outlining. If this sounds complicated it's because it was. To make matters worse, this logic was spread through out `jitmodbuilder.cc`.

Luckily, we are now enforcing that 1) no call may appear inside an entry block and 2) calls are effectively terminators, meaning that there can only be at most one call per block. This makes it a lot easier to detect whether we called into or returned from unmappable code. This commit also moves the entire logic into a single place, making it easier to reason about and maintain.

### DIFF
--- a/bin/yk-config
+++ b/bin/yk-config
@@ -126,6 +126,9 @@ handle_arg() {
             # Emit stackmaps used for JIT deoptimisation.
             OUTPUT="${OUTPUT} -Wl,--mllvm=--yk-insert-stackmaps"
 
+            # Split blocks after function calls.
+            OUTPUT="${OUTPUT} -Wl,--mllvm=--yk-split-blocks-after-calls"
+
             # Ensure we can unambiguously map back to LLVM IR blocks.
             OUTPUT="${OUTPUT} -Wl,--mllvm=--yk-block-disambiguate"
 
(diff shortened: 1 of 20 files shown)


ptersilie:
I think it might be useful for @vext01 to have a look at this too, even if it's just to marvel at all the red lines and the removal of the overcomplicated `CallStack/ResumeAfter` implementation.

ptersilie:
One of the core parts of this PR is https://github.com/ykjit/yk/pull/925/files#diff-7f66eed9b5e12f3961e818d9eb07cef175c437b47ab3bf31ad3428bc57f1050eR1117, which is where the entire `CallStack` logic moved to now, and is among other things responsible for detecting callbacks from external functions.

vext01:
This also fixes intrinsics, right? Or is that a follow up PR?

ptersilie:
> This also fixes intrinsics, right? Or is that a follow up PR?

Yeah, I'm gonna do that separately. This PR is big enough as it is.

ptersilie:
> Is the ykllvm submodule hash pointing at tip of tree in the hopefully-just-to-be-merged ykllvm PR, or will it need updating?

It should be pointing to it yes.

ptersilie:
Missed some formatting. Ok to squash?

vext01:
Please squash.

ptersilie:
Squashed.

PR #928, opened 2023-12-13
Fix side-trace bug triggered by phi nodes.

ptersilie:
PHI nodes require the `LastBB` value to be set, so we can determine which value we need to pick when creating a trace. When creating a side-trace we forgot to reset the `LastBB` value, which resulted in a crash when the first block in a side-trace contained phi nodes.


ptersilie:
Ok to squash?

ltratt:
Please squash.

ptersilie:
Squashed.

PR #929, opened 2023-12-18
Detect inlined intrinsics by peeking at the trace.

ptersilie:
Our previous method of detecting whether an intrinsics was inlined or not relied on annotating intrinsic calls at compile time. For many intrinsics this turned out to be very hard if not impossible, depending on how the intrinsic is lowered.

Since we now only allow one function call per block, this makes detecting inlined intrinsics at runtime very easy. All we need to do is peek at the next block in the trace: if it is a mappable block, the intrinsic was inlined; if the block is unmappable it wasn't inlined (which means we need to copy the call into the JIT module).

https://github.com/ykjit/ykllvm/pull/112 removes the related stuff in ykllvm.

### DIFF
--- a/yktracec/src/jitmodbuilder.cc
+++ b/yktracec/src/jitmodbuilder.cc
@@ -1244,28 +1244,26 @@ public:
               continue;
             }
 
-            // Any intrinsic call which may generate machine code must have
-            // metadata attached that specifies whether it has been inlined or
-            // not.
-            MDNode *IMD = I->getMetadata("yk.intrinsic.inlined");
-            if (IMD == nullptr) {
-              dumpValueAndExit(
-                  "instrinsic is missing `yk.intrinsic.inlined` metadata", &*I);
-            }
-            ConstantAsMetadata *CAM =
-                cast<ConstantAsMetadata>(IMD->getOperand(0));
-            if (CAM->getValue()->isOneValue()) {
-              // The intrinsic was inlined so we don't need to expect an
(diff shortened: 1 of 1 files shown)


ptersilie:
I've also tried to apply the same technique to detect inline assembly containing calls. But it appears these currently break the mapper, since the blocks are not returned in the correct order. This might require a joint investigation with @vext01 to figure out what's going on there.

ltratt:
Are we able to add a test (even if only some sort of black box test) that checks this property holds?

ptersilie:
What property do you mean? We already have intrinsic tests, but they relied on the compile time annotations before. If they still pass then we know this new way works.

ltratt:
Should there be a PR (on ykllvm?) that removes the `yk.intrinsic.inlined` stuff then?

ptersilie:
Yes, I will rip out the annotation stuff next.

ltratt:
Can we do it (if necessary as a linked PR) here? Then the relationship with the two will be obvious.

ptersilie:
Sure, will add it as a dependency once ready.

ptersilie:
I've also updated ykllvm even if that dependency isn't strictly necessary for yk to work.

ltratt:
Please squash.

ptersilie:
Squashed but then noticed I forgot to format. So needs squashing again.

ltratt:
Please squash.

ptersilie:
Squashed.

Issue #909, opened 2023-11-20
`don't know how to handle operand` with try_repeat.


vext01:
Shook this out with try_repeat:

```
---- lang_tests::unroll_safe_implies_noinline.c status ----
Error

---- lang_tests::unroll_safe_implies_noinline.c stderr ----

--- Begin aot ---
source_filename = "ld-temp.o"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

%YkCtrlPointVars = type { ptr, ptr, ptr, ptr, ptr }
%struct.YkLocation = type { i64 }

@shadowstack_0 = global ptr null

; Function Attrs: noinline nounwind optnone uwtable
define dso_local void @never_aot_inline(i32 noundef %0) #0 {
  %2 = load ptr, ptr @shadowstack_0, align 8
  %3 = getelementptr i8, ptr %2, i32 0
  store i32 %0, ptr %3, align 4
  br label %4

4:                                                ; preds = %8, %1
  %5 = load i32, ptr %3, align 4
  %6 = add nsw i32 %5, -1
  store i32 %6, ptr %3, align 4
  %7 = icmp eq i32 %5, 0
  call void (i64, i32, ...) @llvm.experimental.stackmap(i64 4, i32 0, ptr %2, ptr %3, i1 %7)
  br i1 %7, label %12, label %8

8:                                                ; preds = %4
  %9 = load i32, ptr %3, align 4
  %10 = getelementptr i8, ptr %2, i32 16
  store ptr %10, ptr @shadowstack_0, align 8
  %11 = call i32 @call_me(i32 noundef %9) #5
  call void (i64, i32, ...) @llvm.experimental.stackmap(i64 3, i32 0, ptr %2, ptr %3, i32 %9)
  store ptr %2, ptr @shadowstack_0, align 8
  br label %4, !llvm.loop !6

12:                                               ; preds = %4
  br label %14

13:                                               ; preds = %14
  ret void

14:                                               ; preds = %12
  br label %13
}

declare dso_local i32 @call_me(i32 noundef) #1

; Function Attrs: noinline nounwind optnone uwtable
define dso_local i32 @main(i32 noundef %0, ptr noundef %1) #2 {
  %3 = alloca %YkCtrlPointVars, align 8
  %4 = call ptr @malloc(i64 1000000)
[...]

PR #930, opened 2023-12-19
Make a start on the JIT IR and the trace builder.

vext01:
The JIT IR is designed to be small. There are two kinds of instruction:
 - short instructions, with inlined operands.
 - long ones (unimplemented as of yet)

Ideally, function, block and instruction IDs (which are indices read from the on-disk AOT IR) would be either converted to references as we decode, but this would require changes to the decoder. We may have to ditch deku:

https://github.com/sharksforarms/deku/issues/383

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -267,7 +267,7 @@ pub(crate) struct Instruction {
 
 impl Instruction {
     /// Returns the operand at the specified index. Panics if the index is out of bounds.
-    pub(crate) fn get_operand(&self, index: usize) -> &Operand {
+    pub(crate) fn operand(&self, index: usize) -> &Operand {
         &self.operands[index]
     }
 
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -5,12 +5,23 @@
 
 use byteorder::{NativeEndian, ReadBytesExt};
 use deku::prelude::*;
-use std::{cell::RefCell, error::Error, ffi::CStr, fs, io::Cursor, path::PathBuf};
+use std::{
+    cell::RefCell,
+    error::Error,
+    ffi::CStr,
+    fs,
+    io::Cursor,
+    ops::{Deref, DerefMut},
+    path::PathBuf,
+};
 
 /// A magic number that all bytecode payloads begin with.
 const MAGIC: u32 = 0xedd5f00d;
 /// The version of the bytecode format.
 const FORMAT_VERSION: u32 = 0;
+
+/// The symbol name of the control point function (after ykllvm has transformed it).
+const CONTROL_POINT_NAME: &str = "__ykrt_control_point";
 
 fn deserialise_string(v: Vec<u8>) -> Result<String, DekuError> {
     let err = Err(DekuError::Parse("failed to parse string".to_owned()));
@@ -45,7 +56,7 @@ pub(crate) trait IRDisplay {
 
 /// An instruction opcode.
 #[deku_derive(DekuRead)]
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
 #[deku(type = "u8")]
 pub(crate) enum Opcode {
     Nop = 0,
@@ -82,12 +93,51 @@ impl IRDisplay for ConstantOperand {
 }
 
 #[deku_derive(DekuRead)]
-#[derive(Debug)]
-pub(crate) struct LocalVariableOperand {
+#[derive(Debug, Hash, Eq, PartialEq)]
+pub(crate) struct InstructionID {
     #[deku(skip)] // computed after deserialisation.
     func_idx: usize,
     bb_idx: usize,
     inst_idx: usize,
+}
+
+impl InstructionID {
+    pub(crate) fn new(func_idx: usize, bb_idx: usize, inst_idx: usize) -> Self {
+        Self {
+            func_idx,
+            bb_idx,
+            inst_idx,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct BlockID {
+    pub(crate) func_idx: usize,
+    pub(crate) bb_idx: usize,
+}
+
+impl BlockID {
+    pub(crate) fn new(func_idx: usize, bb_idx: usize) -> Self {
+        Self { func_idx, bb_idx }
+    }
+}
+
+#[deku_derive(DekuRead)]
+#[derive(Debug, Hash, Eq, PartialEq)]
+pub(crate) struct LocalVariableOperand(InstructionID);
+
+impl Deref for LocalVariableOperand {
+    type Target = InstructionID;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl DerefMut for LocalVariableOperand {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
 }
 
 impl IRDisplay for LocalVariableOperand {
@@ -161,6 +211,31 @@ pub(crate) enum Operand {
     Unimplemented(#[deku(until = "|v: &u8| *v == 0", map = "deserialise_string")] String),
 }
 
+impl Operand {
+    /// For a [Self::LocalVariable] operand return the instruction that defines the variable.
+    ///
+    /// Panics for other kinds of operand.
+    ///
+    /// OPT: This is expensive.
+    pub(crate) fn to_instr<'a>(&self, aotmod: &'a Module) -> &'a Instruction {
+        match self {
+            Self::LocalVariable(lvo) => {
+                &aotmod.funcs[lvo.func_idx].blocks[lvo.bb_idx].instrs[lvo.inst_idx]
+            }
+            _ => panic!(),
+        }
+    }
+
+    /// Return the `InstructionID` of a local variable operand. Panics if called on other kinds of
+    /// operands.
+    pub(crate) fn to_instr_id(&self) -> InstructionID {
+        match self {
+            Self::LocalVariable(lvo) => InstructionID::new(lvo.func_idx, lvo.bb_idx, lvo.inst_idx),
+            _ => panic!(),
+        }
+    }
+}
+
 impl IRDisplay for Operand {
     fn to_str(&self, m: &Module) -> String {
         match self {
@@ -188,6 +263,45 @@ pub(crate) struct Instruction {
     /// A variable name, only computed if the instruction is ever printed.
     #[deku(skip)]
     name: RefCell<Option<String>>,
+}
+
+impl Instruction {
+    /// Returns the operand at the specified index. Panics if the index is out of bounds.
+    pub(crate) fn get_operand(&self, index: usize) -> &Operand {
+        &self.operands[index]
+    }
+
+    pub(crate) fn opcode(&self) -> Opcode {
+        self.opcode
+    }
+
+    pub(crate) fn is_store(&self) -> bool {
+        self.opcode == Opcode::Store
+    }
+
+    pub(crate) fn is_gep(&self) -> bool {
+        self.opcode == Opcode::GetElementPtr
+    }
+
+    pub(crate) fn is_control_point(&self, aot_mod: &Module) -> bool {
+        if self.opcode == Opcode::Call {
+            // Call instructions always have at least one operand (the callee), so this is safe.
+            let op = &self.operands[0];
+            match op {
+                Operand::Function(fop) => {
+                    return aot_mod.funcs[fop.func_idx].name == CONTROL_POINT_NAME;
+                }
+                _ => todo!(),
+            }
+        }
+        false
+    }
+
+    /// Determine if two instructions in the (immutable) AOT IR are the same based on pointer
+    /// identity.
+    pub(crate) fn ptr_eq(&self, other: &Self) -> bool {
+        std::ptr::eq(self, other)
+    }
 }
 
 impl IRDisplay for Instruction {
@@ -249,8 +363,9 @@ impl IRDisplay for Instruction {
 pub(crate) struct Block {
     #[deku(temp)]
     num_instrs: usize,
+    // FIXME: unpub
     #[deku(count = "num_instrs")]
-    instrs: Vec<Instruction>,
+    pub instrs: Vec<Instruction>,
 }
 
 impl IRDisplay for Block {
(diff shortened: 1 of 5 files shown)


ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -161,6 +208,30 @@ pub(crate) enum Operand {
...
+            _ => None,
+        }
+    }
+
+    /// Return the `InstructionID` of a local variable operand. Must not be called on other
+    /// operands.
+    pub(crate) fn to_instr_id(&self) -> InstructionID {
+        match self {
+            Self::LocalVariable(lvo) => InstructionID::new(lvo.func_idx, lvo.bb_idx, lvo.inst_idx),
+            _ => unreachable!(),
```
ltratt, line 230: This arm is definitely reachable, but to do so would be UB, so we want to abort execution. This should really then be a `panic` IMHO.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -190,6 +261,43 @@ pub(crate) struct Instruction {
     name: RefCell<Option<String>>,
 }
 
+impl PartialEq for Instruction {
+    fn eq(&self, other: &Self) -> bool {
```
ltratt, line 265: I'm fine with using pointer equality but exposing it as `=` makes me a bit queasy because `=` is generally assumed to be a weaker condition than pointer equality. Maybe we should just expose a `ptr_eq` function to callers? If we ever need to change the pointer equality-ness, this will make very clear to us every position which needs changing.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -190,6 +261,43 @@ pub(crate) struct Instruction {
...
+        std::ptr::eq(self, other)
+    }
+}
+
+impl Instruction {
+    pub(crate) fn get_operand(&self, index: usize) -> &Operand {
+        self.operands.get(index).unwrap()
```
ltratt, line 272: Isn't this `self.operands[index]` but more verbose? Also because of the implicit `panic` I think we should document the `# Panics` in the docstring.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -190,6 +261,43 @@ pub(crate) struct Instruction {
...
+        self.opcode == Opcode::Store
+    }
+
+    pub(crate) fn is_gep(&self) -> bool {
+        self.opcode == Opcode::GetElementPtr
+    }
+
+    pub(crate) fn is_control_point(&self, aot_mod: &Module) -> bool {
+        if self.opcode == Opcode::Call {
+            let op = &self.operands[0];
```
ltratt, line 289: I guess `Opcode::Call` always has (at least?) one operand so this can't panic?

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -190,6 +261,43 @@ pub(crate) struct Instruction {
...
+            match op {
+                Operand::Function(fop) => {
+                    return aot_mod.funcs[fop.func_idx].name == "__ykrt_control_point";
```
ltratt, line 292: Probably best as a constant somewhere as we're bound to need it more than once.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -546,7 +655,21 @@ impl Module {
         *self.var_names_computed.borrow_mut() = true;
     }
 
-    /// Fill in the function index of local variable operands of instructions.o
+    pub(crate) fn func_index(&self, find_func: &str) -> Option<usize> {
+        // OPT: create a cache in the Module.
+        for (f_idx, f) in self.funcs.iter().enumerate() {
```
ltratt, line 660: I think this function can become just `self.funcs.iter().enumerate().find(|x| x.name == find_func).0` or similar.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -0,0 +1,400 @@
+//! The Yk JIT IR
+
+use std::fmt;
+use strum_macros::FromRepr;
+
+/// Number of bits used to encode an opcode.
+const OPCODE_SIZE: u64 = 8;
```
ltratt, line 7: I think this stuff should be gated behind a `#[cfg(x86)]` thing or similar. We could even gate the entire module for the time being.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -0,0 +1,400 @@
...
+const SHORT_OPERAND_KIND_MASK: u64 = 7;
+const SHORT_OPERAND_MASK: u64 = 0x3ffff;
+
+/// Bit fiddling for instructions.
+const INSTR_ISSHORT_SIZE: u64 = 1;
+const INSTR_ISSHORT_MASK: u64 = 1;
+const INSTR_OPCODE_MASK: u64 = 0xe;
+
+/// An instruction is identified by its index in the instruction vector.
+pub(crate) type InstructionID = usize;
```
ltratt, line 24: In general I've learnt not to use `type` like this, because it gives the illusion that `usize` and `InstructionID` can't be confused --- but Rust considers them exactly the same type. If we want to do this, a new type (e.g. `struct InstructionId(usize)`) actually gives you a degree of type safety.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -0,0 +1,400 @@
...
+
+#[derive(Debug, PartialEq)]
+pub enum Operand {
+    Long(LongOperand),
+    Short(ShortOperand),
+}
+
+impl Operand {
+    pub(crate) fn new(kind: OpKind, val: u64) -> Self {
+        if val < 2u64.pow(15) {
```
ltratt, line 71: Hang on, isn't this exceeding the range of `u64`? What is the aim here? I wonder if what was intended was to reserve `u64::MAX` as a special value?

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -1,44 +1,127 @@
...
-    fn build(self) -> Result<Module, Box<dyn Error>> {
+    /// Create the prolog of the trace.
+    fn create_trace_header(&mut self, blk: &aot_ir::Block) {
+        // Find trace input variables and emit `LoadArg` instructions for them.
+        let mut last_store = None;
+        let mut trace_input = None;
+        let mut input = Vec::new();
+        for inst in blk.instrs.iter().rev() {
+            if inst.is_control_point(self.aot_mod) {
+                trace_input = Some(inst.get_operand(3));
```
ltratt, line 58: How do we know what operand `3` is?

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -1,44 +1,127 @@
...
+            }
+        }
+    }
+
+    /// Walk over a traced AOT block, translating the constituent instructions into the JIT module.
+    fn process_block(&mut self, bid: aot_ir::BlockID) {
+        // unwrap safe: can't trace a block not in the AOT module.
+        let blk = self.aot_mod.block(&bid).unwrap();
+
+        // Decide how to translate each AOT instruction based upon it's opcode.
```
ltratt, line 84: s/it's/its/ (yes, I am a member of the [Apostrophe Protection Society](https://www.apostrophe.org.uk/) :p)

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: the first operand would be a callee, so yeah.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: @vext01 tells me that the intention is for this to be a 64-bit layout regardless of the platform. We should add a doc string to the module to that effect, just so that nervous people like me know.

vext01: This is just the number of bits used to encode an opcode. It isn't x86_64 specific.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: It's checking if the operand's payload can fit in a small operand, where we have only 15-bits of wiggle room.

(a long instruction, once implemented, would allow larger operands)

I missed the 15 in my "kill all magic numbers" pass too.

vext01: (2^15 should fit in a u64 though, or I've gone mad)

ltratt: Right, got it. I'll leave the comment here so we remember to resolve it.

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: We should define a named constant.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: >  (2^15 should fit in a u64 though, or I've gone mad)

I misread it because I'm so unfamiliar with that way of doing bit patterns -- mea culpa!

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: I dunno man, looks like you wanna kill that apostrophe! :laughing:

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Do you mind if I also rename this to just `operand()` to be more rusty?

ltratt: Go ahead!

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Fixed in f7a664064c4be6065593565929309acb3bf126cf

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Fixed in 9d1a0ff18619b4cdb5e42ceed710f2043369c8b4

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Fixed in 9f91ce460fb9314f10f5f741bae136f5659b388f.

I forgot the rename, will push another commit shortly.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Documented in c54d023c1e017ee88e4cb050bc0b6c0d256d2d47

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Fixed in a4a13b1bd36cc180c6757a28b0099726db519f10

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Fixed in 8284b7c11508a06eb6cd41ba90749b926c6045ca

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Documented in 731d1ae92873424f75dbb0244578b47ee7929872

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Fixed in 6cd71d2cc456d260ee0f69a64c6ca69f3ae2ae5a

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Added a constant in a34b500956cfb5a71d3a63fd2f4eb78e582ce7b7

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: Fixed in f4d906419449c702878e47a163a0ed1fc387b34f

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: Fixed in c02dd73a02dede5bd706e6567f1532367daf31d9

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Renamed in ac7f75e99b6094ce242655ec49c4bb51a3b00f8b.

Let's keep that commit separate, as it's partly existing code we are changing there.

vext01:
@ltratt last two commits ok for you?

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -89,7 +89,8 @@ pub enum Operand {
 
 impl Operand {
     pub(crate) fn new(kind: OpKind, val: u64) -> Self {
-        if val < 2u64.pow(u32::try_from(SHORT_OPERAND_VALUE_SIZE).unwrap()) {
+        // check if the operand's value can fit in a short operand.
+        if val & (u64::MAX << SHORT_OPERAND_VALUE_SIZE) == 0 {
```
ltratt, line 98: It's still not obvious to me what `SIZE` is (OK, OK, I know *now*, but I'll forget later, because it's unclear). Can we put all the maths in the constant at the top of the file when it will become much clearer?

vext01: Did https://github.com/ykjit/yk/pull/930/commits/eb47831171ace788c850f6fe9b6530fe8d219151 not work?

ltratt: OK the docs on `SIZE` help, but let's still stick all the maths in the constant (if nothing else, it makes clearer a) what the values actually are b) they're compile-time constants).

vext01: I think that would need to be another PR as there are lots of computations we'd want to move to the top.

ltratt: All of them are introduced in this PR aren't they? Or are they preexisting?

ltratt: OK, promise me you'll fix it then :)

ltratt:
Please squash.

vext01:
splat.

vext01:
> OK, promise me you'll fix it then :)

https://github.com/ykjit/yk/issues/931

vext01:
I agree that this failure seems unrelated. Maybe kick it again and see if it merges?

vext01:
> don't know how to handle operand:   %1 = load ptr, ptr @shadowstack_0, align 8

again...

ltratt:
Yep, so this isn't a transient error...

vext01:
It's probably this: https://github.com/ykjit/yk/issues/909

I can look into it in January.

vext01:
I just forced a build of this and it succeeded:
https://ci.soft-dev.org/#/builders/1/builds/3288

So I think the failure is in fact transient, maybe it is #909.

It's probably gonna need some try_repeat and thinking hard to fix.

What shall we do with this PR? Retry?

vext01:
(Note that #909 predates this PR)

ptersilie:
Actually, I was working on a bug before Christmas that might fix this. I still have to make a test for the PR. Let's see if that works before looking into this further.

vext01:
Since this PR appears unrelated to the failure, should we retry it?

PR #933, opened 2024-01-08
Fix outlining bug.

ptersilie:
This change fixes a bug where we forgot to reset the `OutlineBase` in `jitmodbuilder.cc` after resetting `Outlining` to `false`. This meant we continued outlining after seeing `yk_mt_control_point`, when we should be copying into the trace instead. This bug only occured when a call immediately preceeded the control point and we were tracing at least 2 rounds of the main interpreter loop. This scenario wasn't previously covered by our test suite.

### DIFF
--- a/tests/c/outlinebase_bug.c
+++ b/tests/c/outlinebase_bug.c
@@ -1,0 +1,78 @@
+// Run-time:
+//   env-var: YKD_PRINT_IR=jit-pre-opt
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_PRINT_JITSTATE=1
+//   stderr:
+//     ...
+//     --- Begin jit-pre-opt ---
+//     ...
+//     {{1}} = call i32 @fputc(...
+//     ...
+//     {{2}} = call i32 @fputc(...
+//     ...
+//     {{3}} = call i32 @fputc(...
+//     ...
+//     {{4}} = call i32 @fputc(...
+//     ...
+//     --- End jit-pre-opt ---
+//     ...
+//   stdout:
+//     exit
+
+// Test that outlining works as expected. Each trace in this test does two
+// rounds of the main interpreter loop, resulting in four `fprintf`s being
+// copied into the JIT module. However, if we forgot to reset the `OutlineBase`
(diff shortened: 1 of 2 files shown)


tests/c/outlinebase_bug.c
```diff
@@ -0,0 +1,79 @@
...
+  while (i > 0) {
+    int r = bar(argc + 2);
+    YkLocation *tmp;
+    if (i %2 == 0) {
+      tmp = &loc;
+    } else {
+      tmp = NULL;
+    }
+    yk_mt_control_point(mt, tmp);
+    fprintf(stderr, "r = %i\n", r);
```
ltratt, line 69: I don't think these two `fprintf`s are relevant to the test in the sense that their output isn't checked (let alone the inconsistent spaces either side of `=` :p)? Or, if they are, are they just `NOOPT_VAL` in disguise?

ptersilie: It's just something I can easily check if it appears in the JIT IR. I agree the content of the print doesn't matter too much.

ltratt: I suggest we kill them -- it's just extra cruft from a reader's perspective.

ptersilie: No, I need them to check that the JIT is doing the right thing. The test checks that there are four `fprintf`s in the trace. If there are only 2 then we've encountered the bug. But I can simplify them by passing the empty string or a single character as the print content is irrelevant.

ltratt: > But I can simplify them by passing the empty string or a single character as the print content is irrelevant.

Works for me!

ptersilie: Try [136382e](https://github.com/ykjit/yk/pull/933/commits/136382e8ea14abd106a69545927b81dcbe975b40)

ltratt:
Please squash.

ptersilie:
Squashed.

ptersilie:
Whoops, missing formatting. Ok to squash?

ltratt:
Please squash.

ptersilie:
Squashed.

PR #936, opened 2024-01-10
Bail when encountering inline assembler.

ptersilie:
We currently can't handle calls within inline assembler in the AOT module, as this messes with the blockmap, leading to the PT decoder producing nonsensical traces. As a temporary solution we just bail when we encounter any assembler that isn't the empty string (we need those for NOOPT_VAL).

### DIFF
--- a/tests/c/inline_asm_call.c
+++ b/tests/c/inline_asm_call.c
@@ -1,16 +1,21 @@
 // Run-time:
-//   env-var: YKD_PRINT_IR=jit-pre-opt
 //   env-var: YKD_SERIALISE_COMPILATION=1
-//   env-var: YKD_PRINT_JITSTATE=1
 //   status: error
+//   stderr:
+//     ...
+//     InlineAsm is currently not supported.
 
-// Check that we can handle inline asm properly (currently expectely fails
-// until we can deal with calls inside inline asm).
+// Check that we bail when we see non-empty inline asm.
 
 #include <assert.h>
 #include <stdlib.h>
 #include <yk.h>
 #include <yk_testing.h>
+
+int foo() {
+  fprintf(stderr, "foo\n");
+  return 11;
+}
 
 int main(int argc, char **argv) {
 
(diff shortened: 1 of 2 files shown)


ltratt:
Does this work with yklua?

ptersilie:
I've tested `db.lua`, `gb.lua`, and `tpack.lua`, and all worked fine.

ltratt:
Could we bail at AOT time for this? That way we'd know we could never encounter it.

ptersilie:
We could, but it's a bit more work if done properly: I would have to add a new pass for this (unless I add something to an unrelated pass), which is a bit tedious and currently seems like a distraction from more important matters.

ltratt:
Can't we just bork in ykllvm if we see "asm not empty string"? After all, AIUI, the problem is that at AOT we don't split blocks properly?

ptersilie:
The problem is that the pt decorder expects us to put information about calls into the blockmap. We can't do that for inline asm where the calls are invisible to us.

Yes, I could catch this at AOT time, but it is a lot more work as we need an extra pass, unless you are ok for me to hack this into any of the existing passes that walk the AOT module.

ltratt:
I guess I was thinking that the serialisation pass (which must serialise asm statements) can do the borking?

ptersilie:
AFAIC the LLVM IR we are currently using is serialised by some LLVM pass we don't control (we could hack a check in there though possibly). For our new IR we definitely can do that though.

ptersilie:
> OK OK, point taken :) Let's do it in this PR and just bork. But at some point we should move to this AOT.

Agreed.

ltratt:
Release mode will remove `debug_assert`s (but not `assert`s unless you explicitly tell it to).

ptersilie:
Those are in the C++ code which I believe are removed in release mode.

ltratt:
Ah, that I don't know!

ltratt:
That said, `assert` really means "the code is internally incorrect" but here we mean "the user gave us sensible input but we can't handle it". So it should probably `err` or similar.

ptersilie:
Okay, will fix it.

ptersilie:
Fixed.

ptersilie:
Just squashed but realised I forgot to format again. So will have to do another push.

ltratt:
In this case please format + squash in one go.

ptersilie:
Done.

PR #938, opened 2024-01-16
Draft attempt at statically typed IR.

ptersilie:
Here's our attempt at getting some static typing into our IR, which turned out to be much harder then we first thought. This is probably not perfect, but it's a start. Our next step will likely be to create amock optimisation pass, to see how difficult/annoying this IR is to work with.

We've documented our design decisions inside the docstrings, so the `Instruction` docstring is probably a good point to start reviewing this.

### DIFF
--- a/ykrt/src/compile/jitc_yk/jit_ir.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir.rs
@@ -12,12 +12,6 @@
 use std::fmt;
 use strum_macros::FromRepr;
 
-/// Number of bits used to encode an opcode.
-const OPCODE_SIZE: u64 = 8;
-
-/// Max number of operands in a short instruction.
-const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
-
 /// Bit fiddling.
 ///
 /// In the constants below:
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -104,6 +104,8 @@ impl<'a> TraceBuilder<'a> {
 
             // Insert the newly-translated instruction into the JIT module.
             self.jit_mod.push(jit_inst);
+
+            eprintln!("{}", self.jit_mod);
```
ptersilie, line 108: Whopps!

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -93,115 +57,81 @@ impl From<u64> for OpCode {
...
+/// encoded into a short instruction.
+///
+/// Instructions are immutable and must not be copied. If you wish to change an instruction,
+/// you must make a new one and replace the old one. If you wish to copy an instruction, you
+/// should make a new one with the same operands. There are several reasons for this:
+///
+///   1) modifying a (boxed) long instruction could lead to it become short, which may result in
+///      accidental memory leaks if we are not careful to reconsititute the [Box] and let it drop.
+///   2) copying a long instruction would make a hidden alias to a boxed pointer. This can lead to
+///      double frees or use after.
```
vext01, line 112: use after *frees*

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -93,115 +57,81 @@ impl From<u64> for OpCode {
...
+/// structure describing the operands.
+///
+/// The IR is designed such that (where possible) the most common kinds of operations can be
+/// encoded into a short instruction.
+///
+/// Instructions are immutable and must not be copied. If you wish to change an instruction,
+/// you must make a new one and replace the old one. If you wish to copy an instruction, you
+/// should make a new one with the same operands. There are several reasons for this:
+///
+///   1) modifying a (boxed) long instruction could lead to it become short, which may result in
```
ltratt, line 109: "lead to it become short" doesn't parse.

vext01: "lead *it*" would be better.

vext01:
At this point, we are mostly looking for comments on the overall design. Polish can come later.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: That wouldn't parse either. I assume that "modifying a boxed long instruction could lead to it becoming a short instruction" is what's intended?

vext01: yes

ltratt:
I might be missing something (some pointers would be good!) but is this basically type punning `Instruction`? How do we do pattern matching on an instruction?

vext01:
You can only really pattern match on an opcode and then based on the result convert to the correct specific instruction.

It's basically the same as `isa<T>` in LLVM.

As far as I (we?) can see, you can't do full pattern matching without bloating all instructions. We tried, trust us!

ltratt:
I believe we agreed yesterday to close this one. If anyone objects please feel free to reopen!

PR #939, opened 2024-01-16
Software tracer

Pavel-Durov:


ykrt/src/trace/sw/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+
+struct SWTTraceCollector {}
+
+impl TraceCollector for SWTTraceCollector {
+    fn stop_collector(self: Box<Self>) -> Result<Box<dyn TraceIterator>, InvalidTraceError> {
+        let mut aot_blocks: Vec<TracedAOTBlock> = vec![];
+        BASIC_BLOCKS.with(|mtt| {
+            for bb in mtt.borrow().iter() {
+                // TODO: Get function name from IR by function_index
+                let func_name = CString::new("test_func").expect("Failed to create CString");
```
Pavel-Durov, line 46: @vext01, @ptersilie how can I obtain IR function name by index?

bin/yk-config
```diff
@@ -144,6 +144,11 @@ handle_arg() {
             # Encode additional locations in stackmaps.
             OUTPUT="${OUTPUT} -Wl,--mllvm=--yk-stackmap-add-locs"
 
+            # Use software-tracer pass
+            # Note: We can't run yk-basicblock-tracer pass with HardwareTracer
+            if [ "${YKB_TRACER}" = "sw" ]; then
```
Pavel-Durov, line 149: @vext01 you were right about the "=" vs "==" issue here.

ykrt/src/trace/sw/mod.rs
vext01: You will have to iterate over the functions in the LLVM AOT IR module, take the n-th element and use `->getName()` on it.

Pavel-Durov: ok thanks!

Pavel-Durov: this module https://github.com/ykjit/yk/blob/master/ykrt/src/compile/jitc_yk/aot_ir.rs?

vext01: That's the new AOT IR, which isn't ready yet. You probably want to use the LLVM backend. Look for a `Module *` in C++ code in `yktracec`.

Pavel-Durov: got it

ykrt/src/trace/sw/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+    fn stop_collector(self: Box<Self>) -> Result<Box<dyn TraceIterator>, InvalidTraceError> {
+        let mut aot_blocks: Vec<TracedAOTBlock> = vec![];
+        BASIC_BLOCKS.with(|mtt| {
+            let mut func_store: HashMap<u32, CString> = HashMap::new();
+            let func_indices: Vec<u32> = mtt.borrow().iter().map(|x| x.function_index).collect();
+            let mut func_names: *mut IRFunctionNameIndex = std::ptr::null_mut();
+            let mut func_names_len: libc::size_t = 0;
+            let bc_section = crate::compile::jitc_llvm::llvmbc_section();
+            unsafe {
+                get_function_names(
```
Pavel-Durov, line 75: @vext01 here's my first attemp on obtaining funciton names by inxes using `ykllvmwrap` module.
When runned as:
```shell
YKB_TRACER=sw cargo test  ::simple_swt.c -- --nocapture 
```
It constructs the following trace:
```text
Mapped { func_name: "main", bb: 10 }
Mapped { func_name: "inc", bb: 0 }
Mapped { func_name: "inc", bb: 1 }
Mapped { func_name: "add1", bb: 0 }
Mapped { func_name: "add1", bb: 1 }
Mapped { func_name: "add1", bb: 3 }
Mapped { func_name: "add1", bb: 2 }
Mapped { func_name: "inc", bb: 2 }
Mapped { func_name: "inc", bb: 4 }
Mapped { func_name: "inc", bb: 3 }
Mapped { func_name: "main", bb: 11 }
Mapped { func_name: "main", bb: 8 }
Mapped { func_name: "main", bb: 9 }
```

ykrt/src/trace/sw/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+                    },
+                    func_indices.as_ptr(),
+                    func_indices.len(),
+                    &mut func_names,
+                    &mut func_names_len,
+                );
+                for entry in std::slice::from_raw_parts(func_names, func_names_len) {
+                    func_store.insert(entry.index, std::ffi::CStr::from_ptr(entry.name).to_owned());
+                }
+                free_function_names(func_names);
```
Pavel-Durov, line 88: This is kind of akward, but is there a better way to deallocate memmory allocated in C?

yktracec/src/ykllvmwrap.cc
```diff
@@ -491,3 +491,56 @@ extern "C" void *__yktracec_irtrace_compile_for_tc_tests(
...
+/// Represents an index and name of an LLVM IR function.
+extern "C" struct IRFunctionNameIndex {
+  unsigned int index;
+  const char *name;
+};
+
+/// Deallocates memory allocated for an array of `IRFunctionNameIndex`
+/// structures.
+extern "C" void free_function_names(IRFunctionNameIndex *function_names) {
+  delete[] function_names;
```
vext01, line 504: Where is the `new` that allocated what you are `delete`ing here?

But before you answer that: I think `get_function_names` is computing a (conceptual) map from index to function name, right?

Assuming so, isn't this the kind of thing that we will want to use time and time again? And since the AOT module doesn't change, and is the same for all threads, I'd be happy for this map to be computed once on first use, and kept around for future tracing sessions to use. You can just let the death of the process be the freeing step :P

[`std::call_once`](https://en.cppreference.com/w/cpp/thread/call_once) may help.

vext01:
Offline you said you still get this error:

> simple_swt: src/jitmodbuilder.cc:1139: Module *JITModBuilder::createModule(): Assertion `LastInst && isa<CallInst>(LastInst)' failed.

This comes from [here](https://github.com/ykjit/yk/blob/3df2f2cd9a65fb6af85534a61d2ea98e5d7f0f0b/yktracec/src/jitmodbuilder.cc#L1140)

The surrounding logic is saying: "if you came from a mappable block into an entry block, then the last instruction you saw must have been a call instruction", but this is invalid somehow. This suggest the trace is invalid control flow WRT the AOT module.

I suggest you dump the AOT module, and then look at the trace you printed above and follow execution through the AOT module to check it makes sense.

yktracec/src/ykllvmwrap.cc
Pavel-Durov: sounds good

Pavel-Durov: updated 👉 3a6d4b4386704ae8553725c55a9e5749c43717e2

tests/c/simple_swt.c
```diff
@@ -0,0 +1,39 @@
+// Run-time:
```
ltratt, line 1: Do we need this test? Isn't it covered in other tests that don't specify what the tracer is (so will be run in hwt or swt mode)?

ykrt/build.rs
```diff
@@ -1,13 +1,16 @@
...
     // Always compile in the LLVM JIT compiler.
     println!("cargo:rustc-cfg=jitc_llvm");
     // Always compile in our bespoke JIT compiler.
     println!("cargo:rustc-cfg=jitc_yk");
-    // Always compile in the HWT tracer.
-    println!("cargo:rustc-cfg=tracer_hwt");
     // FIXME: This is a temporary hack because LLVM has problems if the main thread exits before
     // compilation threads have finished.
     println!("cargo:rustc-cfg=yk_llvm_sync_hack");
+    if std::env::var("YKB_TRACER") == Ok("sw".to_owned()) {
```
ltratt, line 11: I think this should be something like:

```rust
match env::var("YKB_TRACER").as_ref() {
  Ok("swt") => ...,
  Ok("hwt") | Err(env::VarError::NotPresent) => ...
  Ok(x) => panic!("Unknown tracer {x}"),
  Err(_) => panic!("Invalid value for YKB_TRACER")
}
```

so that invalid values of `YKB_TRACER` aren't silently (confusingly) ignored.

At some point the default should probably move from hwt -> swt. But not yet I suppose!

ltratt, line 11: I think this should be something like:

```rust
match env::var("YKB_TRACER").as_ref() {
  Ok("swt") => ...,
  Ok("hwt") | Err(env::VarError::NotPresent) => ...
  Ok(x) => panic!("Unknown tracer {x}"),
  Err(_) => panic!("Invalid value for YKB_TRACER")
}
```

so that invalid values of `YKB_TRACER` aren't silently (confusingly) ignored.

At some point the default should probably move from hwt -> swt. But not yet I suppose!

ykrt/src/trace/sw/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+        vec_size: usize,
+        result: *mut *mut IRFunctionNameIndex,
+        len: *mut libc::size_t,
+    );
+}
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct IRFunctionNameIndex {
+    pub index: u32, //libc::c_int,
```
ltratt, line 31: Commented code. Also should this be `usize`?

ykrt/src/trace/sw/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+                for entry in std::slice::from_raw_parts(func_names, func_names_len) {
+                    func_store.insert(entry.index, std::ffi::CStr::from_ptr(entry.name).to_owned());
+                }
+            }
+            aot_blocks = mtt
+                .borrow()
+                .iter()
+                .map(|bb| TracedAOTBlock::Mapped {
+                    func_name: func_store.get(&bb.function_index).unwrap().to_owned(),
+                    bb: bb.block_index as usize,
```
ltratt, line 93: Because `as` is dangerous we require a comment justifying each use of it. But in this case I *think* you can use `usize::from(bb.block_index)` to get a type-safe version.

ltratt:
One generic, and shallow, comment: I think this should be `swt` to go with the existing `hwt` (or we should rename `hwt` to `hw`).

vext01:
(I'll hold off reviewing until the bug that was discussed in-person yesterday has been addressed)

ptersilie:
Here's the diff to fix that bug we were seeing:
```diff
diff --git a/llvm/lib/CodeGen/Yk/FixStackmapsSpillReloads.cpp b/llvm/lib/CodeGen/Yk/FixStackmapsSpillReloads.cpp
index 86d613000bbf..e2b1964c5613 100644
--- a/llvm/lib/CodeGen/Yk/FixStackmapsSpillReloads.cpp
+++ b/llvm/lib/CodeGen/Yk/FixStackmapsSpillReloads.cpp
@@ -103,6 +103,13 @@ bool FixStackmapsSpillReloads::runOnMachineFunction(MachineFunction &MF) {
         // removed during lowering?
         if (MI.getOpcode() != TargetOpcode::STACKMAP &&
             MI.getOpcode() != TargetOpcode::PATCHPOINT) {
+          MachineOperand Op = MI.getOperand(0);
+          if (Op.isGlobal() && Op.getGlobal()->getGlobalIdentifier() == "yk_trace_basicblock") {
+            // yk_trace_basicblock calls don't require stackmaps so we don't
+            // need to adjust anything here. In fact, doing so will skew any
+            // stackmap that follows.
+            continue;
+          }
           // If we see a normal function call we know it will be followed by a
           // STACKMAP instruction. Set `Collect` to `true` to collect all spill
           // reload instructions between this call and the STACKMAP instruction.
```

In short: stackmaps are buggy and report the wrong offset due to the register allocator mocking up the machine code after stackmaps have been emitted. To fix this we have a pass that cleans things up again after the register allocator has finished. This pass assumed that all calls are followed by a stackmap that needs to be fixed. However, the `yk_trace_basicblock` calls don't have (nor need) a stackmap following it. So the pass tried to fix a stackmap that wasn't there, and thus applied this fix to the next stackmap it could find. In this case the next stackmap was one belonging to a branch which doesn't need fixing.

If you apply this diff to ykllvm you should be running again. The diff will need to be pushed in its own PR, but probably needs some cleaning up first (e.g. the function name shouldn't be hardcoded, but extracted from your the swtracer pass header or wherever it is defined).

ltratt:
@ptersilie Possibly silly question: do we know that `MachineOperand Op = MI.getOperand(0);` always returns non-NULL? [I don't know the API here so slightly curious.]

ptersilie:
I would assume so, since we know that MI is a call instruction which I believe requires a target. But I guess it couldn't hurt to check `MI.getNumOperands()` first.

ltratt:
>I would assume so, since we know that MI is a call instruction which I believe requires a target. 

I suggest quickly verifying this. But in either case maybe add an `assert(MI.getNumOperands() > 0)` or similar?

Pavel-Durov:
> One generic, and shallow, comment: I think this should be `swt` to go with the existing `hwt` (or we should rename `hwt` to `hw`).

updated 👉 9bf03addf5c4ba840e0acc4ee202fd4e87684d81

tests/c/simple_swt.c
Pavel-Durov: Removed 👉 57aed88f9a8139d1f84651e3b83005de7373e86a

ykrt/build.rs
Pavel-Durov: updated 👉 54e80f246c153ae1f1ae91e0150b4c24cb7de16d

Pavel-Durov: updated 👉 54e80f246c153ae1f1ae91e0150b4c24cb7de16d

ykrt/src/trace/sw/mod.rs
Pavel-Durov: comment removed 👉 2690d9c5c195c0175655c32e4e342e7871701e56

Pavel-Durov: We can use `usize` here, but at some point, we'll have to cast it to `u32` anyway cause that's the type for indices coming from the LLVM pass.

ykrt/src/trace/sw/mod.rs
Pavel-Durov: no luck:
```
error[E0277]: the trait bound `usize: From<u32>` is not satisfied
```
I guess I will add a comment then :)

ykrt/src/trace/sw/mod.rs
ltratt: > or maybe all the indices should be usize anyway?

In general I'd start with `usize` unless you know that the space is definitely a problem.

vext01:
What's the status of this PR @Pavel-Durov? Thanks.

Pavel-Durov:
The status is:

The current implementation + @ptersilie [change](https://github.com/ykjit/yk/pull/939#issuecomment-1909980698) still has 10 tests failing with 5 error types. All seems to be failing deterministically.

Error  1:
```
src/jitmodbuilder.cc:1207: Module *JITModBuilder::createModule(): Assertion `PredFound' failed.
````
Tests: `nested_sidetrace.c`, `side-trace.c`, `sidetrace_while.c`


Error 2:
```
don't know how to handle operand: i32 %0
```
Tests: `reentrant.c`, `qsort.c`

Error 3:
```
jit-state: stop-tracing-aborted
```
Tests: `unmapped_setjmp.c`, `setlongjmp.c`

Error 4:
```
don't know how to handle operand:   %5 = call i64 %0(ptr noundef nonnull @.str) #8, !dbg !48
```
Tests: `funcptrarg_noir.c`

Error 5:
```
ld.lld: error: undefined symbol: __yktrace_hwt_mapper_blockmap_len
```
Tests: `blockmap.c`
 > This one we can skip now, thanks to `lang_tester` `ignore-if` feature

File: ykrt/src/trace/sw/mod.rs (hunk 8 above)
Comment (Pavel-Durov) [reply ykrt/src/trace/sw/mod.rs]: changed to `usize` 👉 7f2664dc28755524ed4461d141f2478d26c274ec

File: tests/c/blockmap.c [hunk 10]
```diff
@@ -1,3 +1,5 @@
+// # The function __yktrace_hwt_mapper_blockmap_len is compile-time guarded for hwt only.
+// ignore-if: test $YKB_TRACER == "swt"
```
ltratt, line 2: I think this should be `test $YKB_TRACER != "hwt"`.

Pavel-Durov: Yes, you are right
Actually, it should be:
```
test "$YKB_TRACER" != "hwt" && test "$YKB_TRACER" != ""
```
Since by default, `YKB_TRACER` is not set

Pavel-Durov: updated 👉 2c494bfc79f1a4741475c3464ab9f2885472aec3

ltratt: We should always be setting a variable so we know which tracer we're setting. I agree that we don't do that consistently right now.

Pavel-Durov: Ok, I can set it to be `test "$YKB_TRACER" != "hwt"` and update CI tests accordingly

ltratt: https://github.com/ykjit/yk/pull/953 might give you pointers as to where we want to move (but we're not all the way there yet).

yktracec/src/ykllvmwrap.cc
Pavel-Durov: I moved `called_once` and function match process to `swt` module 👉 ee2a93c05910594526ea8c4a9c61de121916f2da

ptersilie:
Here's another diff that fixes some of the tests, e.g. `qsort.c`, `funcptr_noit.c`, etc:
```diff
diff --git a/yktracec/src/jitmodbuilder.cc b/yktracec/src/jitmodbuilder.cc
index ed898cf81d82..d037d56432e9 100644
--- a/yktracec/src/jitmodbuilder.cc
+++ b/yktracec/src/jitmodbuilder.cc
@@ -291,12 +291,8 @@ class JITModBuilder {
       }
       if (!Outlining) {
         copyInstruction(&Builder, (Instruction *)&*CI, CurBBIdx, CurInstrIdx);
-        if (!IsSWTrace) {
-          // Since the software tracer doesn't trace into external functions, we
-          // don't need to outline.
-          OutlineBase = CallStack.size();
-          Outlining = true;
-        }
+        OutlineBase = CallStack.size();
+        Outlining = true;
       }
     } else {
       // Calling to a non-foreign function.
@@ -1257,7 +1253,15 @@ public:
               assert(Idx + 1 < InpTrace.Length());
               TraceLoc MaybeNextIB = InpTrace[Idx + 1];
               if (const IRBlock *NextIB = MaybeNextIB.getMappedBlock()) {
-                CF = AOTMod->getFunction(NextIB->FuncName);
+                if (IsSWTrace && CI->isIndirectCall() && NextIB->BBIdx > 0) {
+                  // With software tracing we don't get unmappable blocks when
+                  // an indirect call target is unmappable. But we can check
+                  // instead if the next block is an entry block. If it's not,
+                  // the call target is unmappable and can't be inlined.
+                  CF = nullptr;
+                } else {
+                  CF = AOTMod->getFunction(NextIB->FuncName);
+                }
               } else {
                 CF = nullptr;
               }
```

As for the side-trace tests, it took me quite a while to work out what was going on. In short, when starting a new trace, we don't reset the vector storing the last trace, so the vector just keeps growing. This does not only affect side-traces, but also any program where we trace multiple times.

Pavel-Durov:
Thanks for the diff @ptersilie I will give it a go and run more tests. 
Maybe it will unlock some of the lua benchmarks 🤞

ptersilie:
> Thanks for the diff @ptersilie I will give it a go and run more tests. Maybe it will unlock some of the lua benchmarks 🤞

The lua tests definitely run multiple traces so you'll also have to fix the trace vector bug first before they'll work again.

Pavel-Durov:
I also noticed that with the software tracer, two tests are not terminating: `unroll_safe_implies_noinline.c`, `unroll_safe_inlines.c`. Not sure how I missed it before.

ltratt:
> I also noticed that with the software tracer, two tests are not terminating: unroll_safe_implies_noinline.c, unroll_safe_inlines.c. Not sure how I missed it before.

@Pavel-Durov I'm not sure if you're asking for a (re)review on this PR or not: best to be clear if you are! [And if you aren't, please ignore me!]

yktracec/src/jitmodbuilder.cc
```diff
@@ -1266,6 +1293,8 @@ class JITModBuilder {
               // setjmp/longjmp, so for now simply abort this trace.
               // See: https://github.com/ykjit/yk/issues/610
               return nullptr;
+            } else if (S == "yk_trace_basicblock") {
```
ltratt, line 1296: I *think* this check (well, "filter") should be in https://github.com/ykjit/yk/pull/939/files#diff-b952f4a017fefda851303dbeff602e0064b56a0da470e4e64de9c35fd950c8e4: there's no need for jitmodbuilder to know about it, as it's an internal swt detail.

Pavel-Durov: Makes sense.
I'll remove it from `jitmodbuilder.cc`

Pavel-Durov:
> > I also noticed that with the software tracer, two tests are not terminating: unroll_safe_implies_noinline.c, unroll_safe_inlines.c. Not sure how I missed it before.
> 
> @Pavel-Durov I'm not sure if you're asking for a (re)review on this PR or not: best to be clear if you are! [And if you aren't, please ignore me!]

I am not asking for a review yet, cause it's not complete yet, still need to fix the failing tests.

yktracec/src/jitmodbuilder.cc
ptersilie: I don't see how this can be moved to `swt/mod.rs`. All this does is skip the `yk_trace_basicblock` call instruction so it isn't copied over into the JIT module. I don't think `swt/mod.rs` has a view of the AOT IR in order to do the same.

ltratt: I don't know the quickest/easiest way for swt/mod.rs to do this, but logically it should be its job. If we're not careful, jitmodbuilder (or its modern equivalent) will just end up knowing too much about the quirks of each tracing backend. [That said, we might not be able to fully get away this, and I don't want to torture ourselves over it. So we'll have to be pragmatic.]

ptersilie: In order to do this `swt/mod.rs` would have to iterate over all instructions and check if they are a call to `yk_trace_basicblock` and then remove them. That being said, I believe that call is always the first instruction of each block (@Pavel-Durov will need to double-check that), in which case you could just remove the first instruction of each block before handing things over to `jitmodbuilder`.

ltratt: > In order to do this swt/mod.rs would have to iterate over all instructions and check if they are a call to yk_trace_basicblock and then remove them.

Yes. So the real question is: is there a fast way to do this? I can't remember how swt identifies blocks. If it does it by address (somehow) it might be as simple as comparing against the symbol address of the `yk_trace_basicblock` function header or similar?

>That being said, I believe that call is always the first instruction of each block (@Pavel-Durov will need to double-check that), in which case you could just remove the first instruction of each block before handing things over to jitmodbuilder.

Or this! Though I think we'd want to leave lots of `debug_assert`s in to check this works OK.

ptersilie: Correction: the above isn't correct. If a block has phi-nodes or landing-pad instructions, the `yk_trace_basicblock` call will be inserted afterwards. So it can happen that it is not the first instruction of a block.

ptersilie: > Yes. So the real question is: is there a fast way to do this? I can't remember how swt identifies blocks. If it does it by address (somehow) it might be as simple as comparing against the symbol address of the `yk_trace_basicblock` function header or similar?

I don't see how that would help.  I had another look at `swt/mod.rs` and it doesn't even have a view of the instructions or the LLVM IR. It literally just has a vector of block indidces. Those indices are handed to jitmodbuilder which then extracts the LLVM IR blocks and processes them. There is no way that `swt/mod.rs` can alter instructions unless we change both swt and hwt to hand jitmodbuilder actual LLVM IR blocks instead of just the indices.

ptersilie: One way to fix this is to intercept the LLVM IR before it gets serialised and remove those calls then. This is doable for the llvm backend, and should be even easier for our new yk backend as we lower every LLVM instruction to our own AOT IR. So removing a call at the point is very simple.

Pavel-Durov: Oh, I overlooked how `jitmodbuilder` works 👀

ltratt: > One way to fix this is to intercept the LLVM IR before it gets serialised and remove those calls then. 

That's a good idea and has other benefits too (e.g. less serialised stuff)!

Pavel-Durov: If the main concern here is the separation of concerns, i.e. that jitmodbuilder (or its modern equivalent) will end up knowing too much. We can use an abstraction on the tracer implementation while having concrete implementation encapsulated in the tracers.

Something like:
```
// jitmodbuilder
if (is_tracable_function(yk_S) == false) {
	continue;
}

// swt
fn is_tracable_function(func_name CString) -> boolean {
	return func_name != "yk_trace_basicblock";
}

// hwt
fn is_tracable_function(func_name CString)  -> boolean {
	return true;
}
``` 
But intercepting the LLVM IR before it gets serialised sounds cleaner

ltratt: >  But intercepting the LLVM IR before it gets serialised sounds cleaner

Agreed.

yktracec/src/jitmodbuilder.cc
```diff
@@ -1228,7 +1246,16 @@ class JITModBuilder {
               assert(Idx + 1 < InpTrace.Length());
               TraceLoc MaybeNextIB = InpTrace[Idx + 1];
               if (const IRBlock *NextIB = MaybeNextIB.getMappedBlock()) {
-                CF = AOTMod->getFunction(NextIB->FuncName);
+                auto isNextBlockNotEntry = NextIB->BBIdx > 0;
+                if (IsSWTrace && CI->isIndirectCall() && isNextBlockNotEntry) {
```
ltratt, line 1250: I have been chewing on this part of the PR for a while and I think there might be an issue with it.

Imagine we have this program

```
fn f(x: Fn(...)) {
   x();
}

fn g() {
  ...
}
```

and we're tracing `f`. It calls a function pointed to by `x`in a way that we can't know statically. Imagine that we call some code which contains `f(&g)` --- then the trace would look something like:

```
start_f()
indirect_call(*x)
start_g()
...
```

and would be handled by line 1250 in your PR. But imagine that in another place I have code which contains `f(&h)` where `h` is unmappable and looks like:

```
fn h() {
  g()
}
```

This will also create a trace:

```
start_f()
indirect_call(*x)
start_g()
...
```

In other words, I don't *think* we can tell from the software trace whether some unknown code was called as part of the indirect function call as we get the same trace in either case. Warning: I might be wrong!

If I'm not wrong then we need to think of solutions. Two jump out at me:

1. Throw away any trace which contains an indirect function call. This is easy, but obviously unsatisfying.
2. Have the software tracer record the *value* of the function pointer. Then jitmodbuilder (or, probably, the trace mapper) can say:
      a. The pointer is to a function with IR and we can inline it easily.
      b. The pointer is to a function without IR and we can't inline it easily.

CCing @vext01 and @ptersilie  who might have additional thoughts.

yktracec/src/jitmodbuilder.cc
```diff
@@ -1109,14 +1115,27 @@ class JITModBuilder {
...
           if (CallStack.size() == OutlineBase) {
             Outlining = false;
             OutlineBase = 0;
           }
-          continue;
+          if (!IsSWTrace)
+            continue;
+        }
+        if (IsSWTrace) {
+          // In the software tracer external calls are invisible. So when an
```
ltratt, line 1131: I'm less sure than with https://github.com/ykjit/yk/pull/939/files#r1481895717, but I think this part of the code might also suffer from the same problem (i.e. mappable -> unmappable -> mappable code).

yktracec/src/jitmodbuilder.cc
```diff
@@ -1293,7 +1322,13 @@ class JITModBuilder {
           continue;
         }
 
-        if (Idx > 0 && Idx < InpTrace.Length() - 1) {
+        size_t LastBlockIndex;
```
ltratt, line 1325: Assuming https://github.com/ykjit/yk/pull/959 merges, this chunk of code can be removed.

yktracec/src/jitmodbuilder.cc
```diff
@@ -1109,14 +1115,27 @@ class JITModBuilder {
           continue;
         } else if (LastInst && isa<ReturnInst>(LastInst)) {
           LastInst = nullptr;
-          assert(CallStack.back()->getParent() == BB);
-          LastBB = BB;
+          if (!IsSWTrace) {
+            assert(CallStack.back()->getParent() == BB);
+          }
+          LastBB = CallStack.back()->getParent();
```
ltratt, line 1121: This is the main remaining line in the PR that I don't understand: any clues?

yktracec/src/jitmodbuilder.cc
ptersilie: If what you are saying is that we can't tell the difference betwen `g` being called directly via an indirect call or indirectly via another unmappable function, then yes that is correct. However, since an indirect call starts outlining, both cases would be handled the same:
1) See indirect call. Start outlining
2) See entry block of g: push callstack
3) See return from g: pop callstack. Now we stop outlining since we've reached f again
4) Continue copying instructions as normal.

yktracec/src/jitmodbuilder.cc
ptersilie: In the hwt we see blocks containing calls twice, once before the call happens, and then when we return from the call. This is due to how PT sees things. So this assert basically checks, that after returning from a call, we end up in the same block as before the call (this block is then skipped immediately as we've already processed it before the call). In the software tracer we don't see blocks twice, so when we return from a call, the next block we see is the block following the block containing the call. So this assert is incorrect in that case. It's just a small hwt sanity check, but it can probably be removed.

yktracec/src/jitmodbuilder.cc
ltratt: @ptersilie Good point, that is an option 3 that I hadn't considered. It means we lose the ability to inline some safe things, but does mean we don't inline unsafe things. So I think my option 1 is pointless for this PR: we should do option 3 (i.e. what you've suggested) in this PR (I guess that *is* what the PR does already?). And then we can try my option 2 in a later PR.

ptersilie: I just had another look at this code (particularly line 1250). So what this does for the hwt is to peek ahead in the trace and check if the next block is a mappable block. This way we know the indirect call is to something we can inline (if the next block is unmappable then we start outlining). The change in this commit was added to make swt do the same. However, thanks to @ltratt's example I now realise this is not correct. Peek ahead in the software tracer might give you a mappable entry block, however we don't know if there was an umappable call in between. So what that means for this PR is that we always have to outline indirect calls in the swt (at least for the time being). To do so I think we just need to turn that above condition into `if (IsSWTrace)`.

yktracec/src/jitmodbuilder.cc
ltratt: Got it. I think we should remove the `assert` from jitmodbuilder. I wonder if we can work out how to either a) `assert` the same thing in hwt b) get hwt to merge the blocks such that jitmodbuilder sees swt and hwt as doing the same thing? That is probably somewhat hard, but worth thinking about, as it will keep the code simpler if jitmodbuilder (or jitc_yk) can be ignorant of the tracing backend.

yktracec/src/jitmodbuilder.cc
Pavel-Durov: Removed on merge with master 👉 https://github.com/ykjit/yk/pull/939/commits/d05ddbfdd10504184b897a7bc440bae7e410c436

yktracec/src/jitmodbuilder.cc
Pavel-Durov: updated 👉 b7237000e85848d3b498eff3bf1ca9425236359c

ltratt: Works for me.

Pavel-Durov:
> Here's another diff that fixes some of the tests, e.g. `qsort.c`, `funcptr_noit.c`, etc:
> 
> ```diff
> diff --git a/yktracec/src/jitmodbuilder.cc b/yktracec/src/jitmodbuilder.cc
> index ed898cf81d82..d037d56432e9 100644
> --- a/yktracec/src/jitmodbuilder.cc
> +++ b/yktracec/src/jitmodbuilder.cc
> @@ -291,12 +291,8 @@ class JITModBuilder {
>        }
>        if (!Outlining) {
>          copyInstruction(&Builder, (Instruction *)&*CI, CurBBIdx, CurInstrIdx);
> -        if (!IsSWTrace) {
> -          // Since the software tracer doesn't trace into external functions, we
> -          // don't need to outline.
> -          OutlineBase = CallStack.size();
> -          Outlining = true;
> -        }
> +        OutlineBase = CallStack.size();
> +        Outlining = true;
>        }
>      } else {
>        // Calling to a non-foreign function.
> @@ -1257,7 +1253,15 @@ public:
>                assert(Idx + 1 < InpTrace.Length());
>                TraceLoc MaybeNextIB = InpTrace[Idx + 1];
>                if (const IRBlock *NextIB = MaybeNextIB.getMappedBlock()) {
> -                CF = AOTMod->getFunction(NextIB->FuncName);
> +                if (IsSWTrace && CI->isIndirectCall() && NextIB->BBIdx > 0) {
> +                  // With software tracing we don't get unmappable blocks when
> +                  // an indirect call target is unmappable. But we can check
> +                  // instead if the next block is an entry block. If it's not,
> +                  // the call target is unmappable and can't be inlined.
> +                  CF = nullptr;
> +                } else {
> +                  CF = AOTMod->getFunction(NextIB->FuncName);
> +                }
>                } else {
>                  CF = nullptr;
>                }
> ```
> 
> As for the side-trace tests, it took me quite a while to work out what was going on. In short, when starting a new trace, we don't reset the vector storing the last trace, so the vector just keeps growing. This does not only affect side-traces, but also any program where we trace multiple times.

Updated tracer 👉 b2d0c6db3fb52718ec24acba9482066d6f8dc5c0
Updated test 👉 96a7d70c8e033743c1dd0045700e866c36ea2dae

ltratt:
@Pavel-Durov How many tests aren't working after this?

Pavel-Durov:
> @Pavel-Durov How many tests aren't working after this?

Only `side-trace.c` was resurrected by resetting the vector (b2d0c6db3fb52718ec24acba9482066d6f8dc5c0).

Current Failing tests (some with new errors):
```
funcptrarg_hasir.c
interp_inside_call.c
interp_inside_call2.c
nested_sidetrace.c
nested_writetoptr.c
promote.c
promote_expr.c
promote_not_great.c
setlongjmp.c
sidetrace_phinode.c
sidetrace_while.c
unmapped_setjmp.c
unroll_safe_inlines.c
```

Pavel-Durov:
> I'm not super surprised that the `promote*` tests fail (#972 is the start of fixing that, but it won't be enough for you yet). I wonder if all the other ones have side traces or ... ? Maybe start picking one at a time, see if its failure makes sense, check with @vext01 or @ptersilie if it doesn't, and then loop to the next. We'll get there :)

Marked failing tests with `ignor-if` and added a comment with error type to keep track of the progress 👉 3037aeb80146d612cf80877ddd5ff8dd76280889

ltratt:
@Pavel-Durov If you merge in https://github.com/ykjit/yk/pull/976 (it will have a small merge conflict I'm afraid), I *think* the `promote_*` tests should now work?

Pavel-Durov:
> @Pavel-Durov If you merge in #976 (it will have a small merge conflict I'm afraid), I _think_ the `promote_*` tests should now work?

I did merge the latest master 👉 8fd40e0c17c1e314028624566a16f9ac3903562a + ea43365a1f8d5b0fe1442c15c5266d5c27abbf76

Outcome:
`promote_guard.c` works now

Others:
`promote.c`, `promote_expr.c`, `promote_guard.c`, `promote_not_great.c`

Fail at the unimplemented case in [deopt.rs#L57](https://github.com/ykjit/yk/blob/master/ykrt/src/deopt.rs#L57)
```
thread '<unnamed>' panicked at ykrt/src/deopt.rs:57:13:
not yet implemented
fatal runtime error: failed to initiate panic, error 5
```

ltratt:
Ah, the deopt stuff is @ptersilie territory :)

tests/c/funcptrarg_hasir.c
```diff
@@ -1,4 +1,4 @@
-// Error:
+// # Error:
```
ltratt: I don't understand what we've done to this test: why is so much of it commented out?

bin/yk-config
```diff
@@ -143,7 +143,11 @@ handle_arg() {
             OUTPUT="${OUTPUT} -Wl,--mllvm=--yk-shadow-stack"
             # Encode additional locations in stackmaps.
             OUTPUT="${OUTPUT} -Wl,--mllvm=--yk-stackmap-add-locs"
-
+            # Use software-tracer pass
+            # Note: We can't run yk-basicblock-tracer pass with HardwareTracer
```
ltratt, line 147: I think the "note line" is irrelevant: it's kind of obvious that the "swt" pass only makes sense with "swt".

tests/c/blockmap.c
```diff
@@ -1,3 +1,5 @@
+// # The function __yktrace_hwt_mapper_blockmap_len is compile-time guarded for hwt only.
+// ignore-if: test "$YKB_TRACER" != "hwt" && test "$YKB_TRACER" != ""
```
ltratt, line 2: Can `YKB_TRACER` ever be the empty string? I think `test "$YKB_TRACER" != "hwt"` is sufficient.

tests/c/side-trace.c
```diff
@@ -21,35 +21,35 @@
 //     9
 //     10
 //     jit-state: deoptimise
-//     12
+//     ...
```
ltratt, line 24: Why have these become `...`?

ykrt/src/trace/swt/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+
+#[derive(Debug, Eq, PartialEq)]
+struct TracingBlock {
+    function_index: usize,
+    block_index: usize,
+}
+
+thread_local! {
+    static BASIC_BLOCKS: RefCell<Vec<TracingBlock>> = RefCell::new(vec![]);
+    static FUNC_NAMES: RefCell<HashMap<usize, CString>> = RefCell::new(HashMap::new());
```
ltratt, line 19: What is `FUNC_NAMES`? Let's add (even brief) doc strings to these things so that dummies like me have some help :)

ykrt/src/trace/swt/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+    );
+}
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct IRFunctionNameIndex {
+    pub index: usize,
+    pub name: *const libc::c_char,
+}
+/// Inserts LLVM IR basicblock metadata into a thread-local BASIC_BLOCKS vector.
```
ltratt, line 36: Missing blank line before comment.

ykrt/src/trace/swt/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+                        }
+                    }
+                });
+
+                aot_blocks = tb
+                    .borrow()
+                    .iter()
+                    .map(|tb| match fnames.borrow_mut().get(&tb.function_index) {
+                        Some(name) => TraceAction::MappedAOTBlock {
+                            func_name: name.to_owned(),
```
ltratt, line 101: @vext01 Looking at this makes me realise that having to pass around (owned) `String`s is really inefficient. Would it be possible to just carry around the address of the function or some other ID that can later be converted to a string?

[Obviously we'd need to do this with hwt too. And it wouldn't hold up this PR but it's a question worth thinking about generally.]

ykrt/src/trace/swt/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+impl SWTracer {
+    pub fn new() -> Result<Self, Box<dyn Error>> {
+        Ok(SWTracer {})
+    }
+}
+
+struct SWTTraceRecorder {}
+
+impl TraceRecorder for SWTTraceRecorder {
+    fn stop(self: Box<Self>) -> Result<Box<dyn AOTTraceIterator>, InvalidTraceError> {
```
ltratt, line 70: We don't have to do so in this PR but the idea here is that `stop` should be *very* fast, and all the nasty slow stuff (iterating over things) is done in a custom `AOTTraceIterator`. I have a sort-of half-done version of hwt which does this, but it doesn't work yet (for boring reasons).

yktracec/src/jitmodbuilder.cc
```diff
@@ -1273,6 +1301,12 @@ class JITModBuilder {
               // setjmp/longjmp, so for now simply abort this trace.
               // See: https://github.com/ykjit/yk/issues/610
               return nullptr;
+            } else if (IsSWTrace && S == "yk_trace_basicblock") {
+              // YKFIXME: We don't want jitmodbuilder to be aware of any
```
ltratt, line 1305: This is a very opaque comment. Can we clarify?

tests/c/funcptrarg_hasir.c
Pavel-Durov: I think this is an out-of-date version of the file, I removed the comment in the latest commit: https://github.com/Pavel-Durov/yk-fork/blob/software-tracer/tests/c/funcptrarg_hasir.c

Originally I was adding comments to keep track of the failing tests.

bin/yk-config
Pavel-Durov: removed 👉 de4a1896b9c2d48a7f9aa1bba688520a82bcc562

tests/c/blockmap.c
Pavel-Durov: Updated 👉 b3119f15916eaf41311e46631b82cc177504a000

tests/c/side-trace.c
Pavel-Durov: I will check gain

ykrt/src/trace/swt/mod.rs
Pavel-Durov: Added 👉 b3cd5fd0b66ac9a0a8adfb0ab58aa0bbcbd4cedb

ykrt/src/trace/swt/mod.rs
Pavel-Durov: Addeed 👉 a34da68cc6fa56962f91198ff948c4ef0d50d535

yktracec/src/jitmodbuilder.cc
Pavel-Durov: Updated 👉 a79c916cc989e4bf8d3dda072e71fd30da3839f3
Is it clearer?

ykrt/src/trace/swt/mod.rs
Pavel-Durov: updated 👉 9a18474546bdd172c52c86bd11858a87899eb59e

tests/c/side-trace.c
Pavel-Durov: This change is not needed any more.
Updated 👉  a82d1e2d9fa9b0cbefc2b496a975dcfabac6e15e

Pavel-Durov:
There is still this comment - https://github.com/ykjit/yk/pull/939#discussion_r1501013662
Otherwise, I think this PR is ready for review

ykrt/src/trace/swt/mod.rs
ltratt: Ah, I think 9a18474 might make things worse, in the sense that it's pushing much more pressure onto the normal interpreter? I guess this slows things down; or, at least, it will later.

I'm inclined to revert 9a18474 for now, and then do something a little different in a future PR. By then I hope I've raised my hwt PR which might make clearer what I was thinking of here (or it might not!).

.buildbot.sh
```diff
@@ -131,11 +131,15 @@ export YKB_YKLLVM_BUILD_ARGS="define:CMAKE_C_COMPILER=/usr/bin/clang,define:CMAK
...
-for i in $(seq 10); do
-    RUST_TEST_SHUFFLE=1 cargo test --release
-    YKD_NEW_CODEGEN=1 RUST_TEST_SHUFFLE=1 cargo test --release
+tracers="hwt swt"
+
+for tracer in $tracers; do 
+    YKB_TRACER="${tracer}" cargo build --release -p ykcapi
+    echo "Running ${tracer} tests"
+    for i in $(seq 10); do
+        YKB_TRACER="${tracer}" RUST_TEST_SHUFFLE=1 cargo test --release
```
ltratt, line 140: Just realised: we should also test this in non-release mode (which should be a bit further up in the `.buildbot.sh` file).

ltratt:
@ptersilie Once the revert of https://github.com/ykjit/yk/pull/939/commits/9a18474546bdd172c52c86bd11858a87899eb59e has occurred, this is ready for you to take another look at.

@Pavel-Durov Quick check: does `cargo test` pass on this PR? I *think* it now does, but wanted to check!

Pavel-Durov:
> @Pavel-Durov Quick check: does `cargo test` pass on this PR? I _think_ it now does, but wanted to check!
`cargo test` works on this PR for both `YKB_TRACER=swt` and `YKB_TRACER=hwt`

Pavel-Durov:
> @ptersilie Once the revert of [9a18474](https://github.com/ykjit/yk/commit/9a18474546bdd172c52c86bd11858a87899eb59e) has occurred, this is ready for you to take another look at.

reverted 👉 68d876b0085a965a4e26a63323d718986ba97f89

ykrt/src/trace/swt/mod.rs
Pavel-Durov: reverted 👉 68d876b0085a965a4e26a63323d718986ba97f89

.buildbot.sh
Pavel-Durov: Updated 👉 9c5c476120b2b886d4a71cffa3f1760916f330b3

ykrt/build.rs
```diff
@@ -1,13 +1,20 @@
...
     println!("cargo:rustc-cfg=jitc_llvm");
     // Always compile in our bespoke JIT compiler.
     println!("cargo:rustc-cfg=jitc_yk");
-    // Always compile in the HWT tracer.
-    println!("cargo:rustc-cfg=tracer_hwt");
     // FIXME: This is a temporary hack because LLVM has problems if the main thread exits before
     // compilation threads have finished.
     println!("cargo:rustc-cfg=yk_llvm_sync_hack");
+    match env::var("YKB_TRACER") {
+        Ok(ref tracer) if tracer == "swt" => println!("cargo:rustc-cfg=tracer_swt"),
```
ptersilie, line 14: Can this be `Ok("swt")` instead or won't that work?

ptersilie:
Apart from one minor questions this looks good.

ykrt/build.rs
Pavel-Durov: I think the only way to do so will be to declare "swt" and "hwt" as variables and then match on that.

If I use `Ok("swt")` I get type mismatch: `expected String, found &str`
And I can't use `String::from` cause calls are not allowed in patterns.

ptersilie: You might be able to do `match env::var("YKB_TRACER").as_ref()` but it's not that important I guess. I leave it up to you if you want to try.

Pavel-Durov: That gives me an error as well: `expected &String, found &str`
I guess I could do something like:
```
match env::var("YKB_TRACER").as_ref().map(|s| s.as_str()) {
    Ok("swt") => println!("cargo:rustc-cfg=tracer_swt"),
     ...
}
```
But its also not great :)

ptersilie: Okay, nevermind. Let's not get held up by this. :-)

ptersilie:
Please squash.

Pavel-Durov:
squashed 👉 ad8576e83bc8896e622f19592f50e81130eb6e4a

Pavel-Durov:
Build failed cause of formatting in c:

fixed 👉 deaa9d0a0c508344fb612156f8a530a41c865399

@ptersilie can re-review?

ltratt:
Please squash.

Pavel-Durov:
squashed 👉 2288c201f37eda2500c965efe0406ffcaa5a1186

Pavel-Durov:
tests failed 🤔

Pavel-Durov:
I think there was just a mixup with `swt` and `hwt` build/test
This change fixes the buildbot 👉 d18037bf1eb6d22e383cdae6443c5c191382d306

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -204,7 +204,7 @@ impl InstructionID {
     }
 }
 
-#[derive(Debug, PartialEq)]
```
Pavel-Durov, line 207: Something went wrong here with this merge, there are a few unwanted changes here...
ykrt/src/compile/jitc_yk/aot_ir.rs
ykrt/src/compile/jitc_yk/jit_ir.rs
ykrt/src/compile/jitc_yk/trace_builder.rs

Pavel-Durov:
> @Pavel-Durov I'm confused as to what's changed, possibly because of the force push.

Yeah sorry, something went wrong with one of the master merges.

The main change from last merge attempt is this commit: d18037b

ltratt:
OK please remember not to force push in the future unless asked.

Please squash.

Pavel-Durov:
I was desperate (I know it's not a good excuse). I'll avoid doing it again.

squashed 👉 0df2295547f0e6be053904374b5dab880ea42505

PR #943, opened 2024-01-24
A new, statically typed, JIT IR.

ptersilie:
This reworks the JIT IR, incorporating the wisdom we've absorbed through studying PyPy and LuaJIT's JIT IRs.

The JIT IR is now a vector of word-sized instructions that can be matched and destructured using Rust's type system. This differs slightly from PyPy and LuaJIT, which use bytestreams, but we want static type-safety.

Any information that can't fit inline in the instruction word (e.g. operands and their types) are referenced using indices into an auxiliary vector containing the overflow information. This idea was borrowed from LuaJIT.

Indices are all u16, and thus can only represent a limited number of elements. This is fine, as we don't expect traces to be so large that the indices grow beyond what a u16 can represent. If they do we can just abort the trace and remain sound.

A note on mutability: you may mutate the instruction stream so long as you don't cause any index skew. It's OK to replace/mutate an instruction, but you must NOT remove instructions.

The IR is not finished and has a lot of dead code, hence we mark the whole module `#![allow(dead_code)]`. This will eventually be deleted.

Probable follow up PRs:
 - propagate errors when indices overflow (and test).
 - implement missing operand and instruction types (get enough in to get a simple test loop to translate to JIT IR).
 - generate code :)

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -23,6 +23,31 @@ const FORMAT_VERSION: u32 = 0;
 /// The symbol name of the control point function (after ykllvm has transformed it).
 const CONTROL_POINT_NAME: &str = "__ykrt_control_point";
 
+// Generate common methods for index types.
+macro_rules! index {
+    ($struct:ident) => {
+        impl $struct {
+            pub(crate) fn new(v: usize) -> Self {
+                Self(v)
+            }
+
+            pub(crate) fn to_usize(&self) -> usize {
+                self.0
+            }
+        }
+    };
+}
+
+#[deku_derive(DekuRead)]
+#[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq)]
+pub(crate) struct FuncIndex(usize);
+index!(FuncIndex);
+
+#[deku_derive(DekuRead)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) struct TypeIndex(usize);
+index!(TypeIndex);
+
 fn deserialise_string(v: Vec<u8>) -> Result<String, DekuError> {
     let err = Err(DekuError::Parse("failed to parse string".to_owned()));
     match CStr::from_bytes_until_nul(v.as_slice()) {
@@ -96,13 +121,13 @@ impl IRDisplay for ConstantOperand {
 #[derive(Debug, Hash, Eq, PartialEq)]
 pub(crate) struct InstructionID {
     #[deku(skip)] // computed after deserialisation.
-    func_idx: usize,
+    func_idx: FuncIndex,
     bb_idx: usize,
     inst_idx: usize,
 }
 
 impl InstructionID {
-    pub(crate) fn new(func_idx: usize, bb_idx: usize, inst_idx: usize) -> Self {
+    pub(crate) fn new(func_idx: FuncIndex, bb_idx: usize, inst_idx: usize) -> Self {
         Self {
             func_idx,
             bb_idx,
@@ -113,12 +138,12 @@ impl InstructionID {
 
 #[derive(Debug)]
 pub(crate) struct BlockID {
-    pub(crate) func_idx: usize,
+    pub(crate) func_idx: FuncIndex,
     pub(crate) bb_idx: usize,
 }
 
 impl BlockID {
-    pub(crate) fn new(func_idx: usize, bb_idx: usize) -> Self {
+    pub(crate) fn new(func_idx: FuncIndex, bb_idx: usize) -> Self {
         Self { func_idx, bb_idx }
     }
 }
@@ -149,7 +174,7 @@ impl IRDisplay for LocalVariableOperand {
 #[deku_derive(DekuRead)]
 #[derive(Debug)]
 pub(crate) struct TypeOperand {
-    type_idx: usize,
+    type_idx: TypeIndex,
 }
 
 #[deku_derive(DekuRead)]
@@ -167,7 +192,7 @@ impl IRDisplay for BlockOperand {
 #[deku_derive(DekuRead)]
 #[derive(Debug)]
 pub(crate) struct FunctionOperand {
-    func_idx: usize,
+    func_idx: FuncIndex,
 }
 
 /// An operand that is an argument to the parent function.
@@ -220,7 +245,7 @@ impl Operand {
     pub(crate) fn to_instr<'a>(&self, aotmod: &'a Module) -> &'a Instruction {
         match self {
             Self::LocalVariable(lvo) => {
-                &aotmod.funcs[lvo.func_idx].blocks[lvo.bb_idx].instrs[lvo.inst_idx]
+                &aotmod.funcs[lvo.func_idx.to_usize()].blocks[lvo.bb_idx].instrs[lvo.inst_idx]
             }
             _ => panic!(),
         }
@@ -241,8 +266,8 @@ impl IRDisplay for Operand {
         match self {
             Self::Constant(c) => c.to_str(m),
             Self::LocalVariable(l) => l.to_str(m),
-            Self::Type(t) => m.types[t.type_idx].to_str(m),
-            Self::Function(f) => m.funcs[f.func_idx].name.to_owned(),
+            Self::Type(t) => m.types[t.type_idx.to_usize()].to_str(m),
+            Self::Function(f) => m.funcs[f.func_idx.0].name.to_owned(),
             Self::Block(bb) => bb.to_str(m),
             Self::Arg(a) => a.to_str(m),
             Self::Unimplemented(s) => format!("?op<{}>", s),
@@ -254,7 +279,7 @@ impl IRDisplay for Operand {
 #[deku_derive(DekuRead)]
 #[derive(Debug)]
 pub(crate) struct Instruction {
-    type_index: usize,
+    type_index: TypeIndex,
     opcode: Opcode,
     #[deku(temp)]
     num_operands: u32,
@@ -271,8 +296,31 @@ impl Instruction {
         &self.operands[index]
     }
 
+    /// Return a slice of the remaining operands, starting from the index `from` (inclusive).
+    pub(crate) fn remaining_operands(&self, from: usize) -> &[Operand] {
+        &self.operands[from..]
+    }
+
     pub(crate) fn opcode(&self) -> Opcode {
         self.opcode
+    }
+
+    /// For a call instruction, return the callee.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the instruction isn't a call instruction.
+    pub(crate) fn callee<'a>(&self) -> FuncIndex {
+        debug_assert!(matches!(self.opcode, Opcode::Call));
+        let op = self.operand(0);
+        match op {
+            Operand::Function(fo) => fo.func_idx,
+            _ => panic!(),
+        }
+    }
+
+    pub(crate) fn type_index(&self) -> TypeIndex {
+        self.type_index
     }
 
     pub(crate) fn is_store(&self) -> bool {
@@ -289,7 +337,7 @@ impl Instruction {
             let op = &self.operands[0];
             match op {
                 Operand::Function(fop) => {
-                    return aot_mod.funcs[fop.func_idx].name == CONTROL_POINT_NAME;
+                    return aot_mod.funcs[fop.func_idx.0].name == CONTROL_POINT_NAME;
                 }
                 _ => todo!(),
             }
@@ -363,9 +411,8 @@ impl IRDisplay for Instruction {
 pub(crate) struct Block {
     #[deku(temp)]
     num_instrs: usize,
-    // FIXME: unpub
     #[deku(count = "num_instrs")]
-    pub instrs: Vec<Instruction>,
+    pub(crate) instrs: Vec<Instruction>,
 }
 
 impl IRDisplay for Block {
@@ -384,14 +431,14 @@ impl IRDisplay for Block {
 pub(crate) struct Function {
     #[deku(until = "|v: &u8| *v == 0", map = "deserialise_string")]
     name: String,
-    type_index: usize,
+    type_index: TypeIndex,
     #[deku(temp)]
     num_blocks: usize,
     #[deku(count = "num_blocks")]
     blocks: Vec<Block>,
 }
 
-impl Function {
+impl<'a> Function {
     fn is_declaration(&self) -> bool {
         self.blocks.is_empty()
     }
@@ -400,11 +447,20 @@ impl Function {
     pub(crate) fn block(&self, bb_idx: usize) -> Option<&Block> {
         self.blocks.get(bb_idx)
     }
+
+    #[cfg(test)]
+    pub(crate) fn new(name: &str, type_index: TypeIndex) -> Self {
+        Self {
+            name: name.to_string(),
+            type_index,
+            blocks: Vec::new(),
+        }
+    }
 }
 
 impl IRDisplay for Function {
     fn to_str(&self, m: &Module) -> String {
-        let ty = &m.types[self.type_index];
+        let ty = &m.types[self.type_index.to_usize()];
         if let Type::Func(fty) = ty {
             let mut ret = format!(
                 "func {}({}",
@@ -420,7 +476,7 @@ impl IRDisplay for Function {
                 ret.push_str(", ...");
             }
             ret.push(')');
-            let ret_ty = &m.types[fty.ret_ty];
+            let ret_ty = &m.types[fty.ret_ty.to_usize()];
             if ret_ty != &Type::Void {
                 ret.push_str(&format!(" -> {}", ret_ty.to_str(m)));
             }
@@ -445,7 +501,7 @@ impl IRDisplay for Function {
 //
 // Signedness is not specified.
 #[deku_derive(DekuRead)]
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub(crate) struct IntegerType {
     num_bits: u32,
 }
@@ -481,7 +537,7 @@ impl IRDisplay for IntegerType {
 }
 
 #[deku_derive(DekuRead)]
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub(crate) struct FuncType {
     /// The number of formal arguments the function takes.
     #[deku(temp)]
@@ -490,13 +546,28 @@ pub(crate) struct FuncType {
     #[deku(count = "num_args")]
     arg_tys: Vec<usize>,
     /// Type index of the function's return type.
-    ret_ty: usize,
+    ret_ty: TypeIndex,
     /// Is the function vararg?
     is_vararg: bool,
 }
 
+impl FuncType {
+    pub(crate) fn num_args(&self) -> usize {
+        self.arg_tys.len()
+    }
+
+    #[cfg(test)]
+    pub(crate) fn new(arg_tys: Vec<usize>, ret_ty: TypeIndex, is_vararg: bool) -> Self {
+        Self {
+            arg_tys,
+            ret_ty,
+            is_vararg,
+        }
+    }
+}
+
 #[deku_derive(DekuRead)]
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub(crate) struct StructType {
     /// The number of fields the struct has.
     #[deku(temp)]
@@ -548,7 +619,7 @@ const TYKIND_UNIMPLEMENTED: u8 = 255;
 
 /// A type.
 #[deku_derive(DekuRead)]
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 #[deku(type = "u8")]
 pub(crate) enum Type {
     #[deku(id = "TYKIND_VOID")]
@@ -601,7 +672,7 @@ impl IRDisplay for Type {
 #[deku_derive(DekuRead)]
 #[derive(Debug)]
 pub(crate) struct Constant {
-    type_index: usize,
+    type_index: TypeIndex,
     #[deku(temp)]
     num_bytes: usize,
     #[deku(count = "num_bytes")]
@@ -610,7 +681,7 @@ pub(crate) struct Constant {
 
 impl IRDisplay for Constant {
     fn to_str(&self, m: &Module) -> String {
-        m.types[self.type_index].const_to_str(self)
+        m.types[self.type_index.to_usize()].const_to_str(self)
     }
 }
 
@@ -661,17 +732,30 @@ impl Module {
         *self.var_names_computed.borrow_mut() = true;
     }
 
-    pub(crate) fn func_index(&self, find_func: &str) -> Option<usize> {
+    pub(crate) fn func_index(&self, find_func: &str) -> Option<FuncIndex> {
         // OPT: create a cache in the Module.
         self.funcs
             .iter()
             .enumerate()
             .find(|(_, f)| f.name == find_func)
-            .map(|(f_idx, _)| f_idx)
+            .map(|(f_idx, _)| FuncIndex(f_idx))
+    }
+
+    /// Look up a `FuncType` by its index.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the type index is either out of bounds, or the corresponding type is not a
+    /// function type.
+    pub(crate) fn func_ty(&self, func_idx: FuncIndex) -> &FuncType {
+        match self.types[self.funcs[func_idx.to_usize()].type_index.to_usize()] {
+            Type::Func(ref ft) => &ft,
+            _ => panic!(),
+        }
     }
 
     pub(crate) fn block(&self, bid: &BlockID) -> Option<&Block> {
-        self.funcs.get(bid.func_idx)?.block(bid.bb_idx)
+        self.funcs.get(bid.func_idx.to_usize())?.block(bid.bb_idx)
     }
 
     /// Fill in the function index of local variable operands of instructions.
@@ -684,7 +768,7 @@ impl Module {
                 for inst in &mut bb.instrs {
                     for op in &mut inst.operands {
                         if let Operand::LocalVariable(ref mut lv) = op {
-                            lv.func_idx = f_idx;
+                            lv.func_idx = FuncIndex(f_idx);
                         }
                     }
                 }
@@ -695,8 +779,8 @@ impl Module {
     /// Get the type of the instruction.
     ///
     /// It is UB to pass an `instr` that is not from the `Module` referenced by `self`.
-    fn instr_type(&self, instr: &Instruction) -> &Type {
-        &self.types[instr.type_index]
+    pub(crate) fn instr_type(&self, instr: &Instruction) -> &Type {
+        &self.types[instr.type_index.to_usize()]
     }
 
     // FIXME: rename this to `is_def()`, which we've decided is a beter name.
@@ -724,6 +808,21 @@ impl Module {
     }
 }
 
+#[cfg(test)]
+impl Module {
+    pub(crate) fn push_func(&mut self, func: Function) -> FuncIndex {
+        let idx = self.funcs.len();
+        self.funcs.push(func);
+        FuncIndex(idx)
+    }
+
+    pub(crate) fn push_type(&mut self, ty: Type) -> TypeIndex {
+        let idx = self.types.len();
+        self.types.push(ty);
+        TypeIndex(idx)
+    }
+}
+
 /// Deserialise an AOT module from the slice `data`.
 pub(crate) fn deserialise_module(data: &[u8]) -> Result<Module, Box<dyn Error>> {
     match Module::from_bytes((data, 0)) {
@@ -748,9 +847,9 @@ pub fn print_from_file(path: &PathBuf) -> Result<(), Box<dyn Error>> {
 #[cfg(test)]
 mod tests {
     use super::{
-        deserialise_module, deserialise_string, Constant, IntegerType, Opcode, FORMAT_VERSION,
-        MAGIC, OPKIND_BLOCK, OPKIND_CONST, OPKIND_FUNCTION, OPKIND_LOCAL_VARIABLE, OPKIND_TYPE,
-        OPKIND_UNIMPLEMENTED, TYKIND_FUNC, TYKIND_INTEGER, TYKIND_PTR, TYKIND_STRUCT,
+        deserialise_module, deserialise_string, Constant, IntegerType, Opcode, TypeIndex,
+        FORMAT_VERSION, MAGIC, OPKIND_BLOCK, OPKIND_CONST, OPKIND_FUNCTION, OPKIND_LOCAL_VARIABLE,
+        OPKIND_TYPE, OPKIND_UNIMPLEMENTED, TYKIND_FUNC, TYKIND_INTEGER, TYKIND_PTR, TYKIND_STRUCT,
         TYKIND_UNIMPLEMENTED, TYKIND_VOID,
     };
     use byteorder::{NativeEndian, WriteBytesExt};
@@ -1101,7 +1200,7 @@ func bar();
             // Construct an IR constant and check it stringifies ok.
             let it = IntegerType { num_bits };
             let c = Constant {
-                type_index: 0,
+                type_index: TypeIndex::new(0),
                 bytes,
             };
             assert_eq!(it.const_to_str(&c), expect);
--- a/ykrt/src/compile/jitc_yk/jit_ir.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir.rs
@@ -1,22 +1,13 @@
 //! The Yk JIT IR
 //!
 //! This is the in-memory trace IR constructed by the trace builder and mutated by optimisations.
-//!
-//! Design notes:
-//!
-//!  - This module uses `u64` extensively for bit-fields. This is not a consequence of any
-//!    particular hardware platform, we just chose a 64-bit field.
-//!
-//!  - We avoid heap allocations at all costs.
 
-use std::fmt;
-use strum_macros::FromRepr;
+// For now, don't swap others working in other areas of the system.
+// FIXME: eventually delete.
+#![allow(dead_code)]
 
-/// Number of bits used to encode an opcode.
-const OPCODE_SIZE: u64 = 8;
-
-/// Max number of operands in a short instruction.
-const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
+use super::aot_ir;
+use std::{fmt, mem, ptr};
 
 /// Bit fiddling.
 ///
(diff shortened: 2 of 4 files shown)


ltratt:
>  A note on mutability: you may mutate the instruction stream so long as you don't cause any index skew. It's OK to replace/mutate an instruction, but you must NOT remove instructions.

A thought. *Maybe* (but not for this PR at least!) we should wrap the `Vec` of instructions with a new type that doesn't expose instruction deletion? Obviously we'd still let trusted code get access to the underlying vec, but that would make it harder for our future selves to be careless and use `remove` or friends without thinking a bit about it.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -265,16 +265,33 @@ pub(crate) struct Instruction {
     name: RefCell<Option<String>>,
 }
 
-impl Instruction {
+impl<'a> Instruction {
```
ltratt, line 268: Since `Instruction` doesn't have a lifetime, I think this lifetime should be moved directly into those functions that need it (e.g. `get_callee`).

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -265,16 +265,33 @@ pub(crate) struct Instruction {
...
+    }
+
     pub(crate) fn opcode(&self) -> Opcode {
         self.opcode
     }
 
+    /// For a call instruction, return the callee.
+    ///
+    /// Panics if the instruction isn't a call.
+    pub(crate) fn get_callee(&self, m: &'a Module) -> &'a Function {
```
ltratt, line 286: `get_` is not very Rust-like and we should avoid it when possible: `callee` would be fine here I think.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -265,16 +265,33 @@ pub(crate) struct Instruction {
...
+        &self.operands[from..]
+    }
+
     pub(crate) fn opcode(&self) -> Opcode {
         self.opcode
     }
 
+    /// For a call instruction, return the callee.
+    ///
+    /// Panics if the instruction isn't a call.
```
ltratt, line 285: The convention is `#! Panics` in doc strings.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -391,7 +408,15 @@ pub(crate) struct Function {
     blocks: Vec<Block>,
 }
 
-impl Function {
+impl<'a> Function {
+    pub(crate) fn get_name(&self) -> &String {
```
ltratt, line 412: Pointless `get_` (I won't comment on the rest of these but you get the idea).

ltratt: [In general it's also more idiomatic to return `&str` than `&String`, but it's not a big deal.]

ptersilie:
> > A note on mutability: you may mutate the instruction stream so long as you don't cause any index skew. It's OK to replace/mutate an instruction, but you must NOT remove instructions.
> 
> A thought. _Maybe_ (but not for this PR at least!) we should wrap the `Vec` of instructions with a new type that doesn't expose instruction deletion? Obviously we'd still let trusted code get access to the underlying vec, but that would make it harder for our future selves to be careless and use `remove` or friends without thinking a bit about it.

Yes, we've thought about that too. I believe we have this already, since the JIT `Module` fields are private, so the only access is via accessor functions. Obviously, the `jit_ir.rs` file can still access the underlying vec directly. So if you want to avoid that too, then your solution would fix that.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
-pub(crate) struct InstructionID(usize);
-
-impl InstructionID {
-    pub(crate) fn new(v: usize) -> Self {
-        Self(v)
-    }
+const OPERAND_INDEX_MASK: u16 = 0x7fff;
+
+// The maximum size of an operand we can encode in an instruction is 15 bits.
+const MAX_OPERAND_IDX: u16 = 32768;
```
ltratt, line 23: It's impossible for someone stupid like to me to convert from decimal to a bit pattern. I personally normally prefer `0b111111` notation for this (unless there are "too many" bits for someone to count accurately) but hex `0x1234` notation or `x << y` shifting can work too. [Note that `OPERAND_INDEX_MASK` uses hex already!]

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
+
+// Pre-allocated type indices.
+const TYPE_INDEX_VOID: u16 = 0;
```
ltratt, line 26: These three variables might better be an enum with `repr(u16)`?

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
-impl From<u64> for OpKind {
-    fn from(v: u64) -> Self {
-        // unwrap safe assuming only valid discriminant numbers are used.
-        Self::from_repr(v).unwrap()
-    }
+/// A function type.
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct FuncType {
+    /// Type indices for the function's formal arguments.
+    arg_tys: Vec<TypeIndex>,
```
ltratt, line 57: I'm fine with using `ty` as the abbrevation, but then I wondered if we should use the same abbreviation in the struct names too? [I don't feel strongly about this, so feel free to ignore it!]

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
+impl FuncType {
+    pub(crate) fn new(arg_tys: Vec<TypeIndex>, ret_ty: TypeIndex, has_varargs: bool) -> Self {
+        Self {
+            arg_tys,
+            ret_ty,
+            has_varargs,
+        }
+    }
+
+    pub(crate) fn num_args(&self) -> usize {
```
ltratt, line 73: This should probably be `len_args` but actually it's a dangerous function because it doesn't account for var args. At the very least a doc string should point out this caveat.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
-    fn from(v: u64) -> Self {
-        // unwrap safe assuming only valid discriminant numbers are used.
-        Self::from_repr(v).unwrap()
+// FIXME: macro generate `new` and `get` for all index types.
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd)]
+pub(crate) struct InstrIndex(u16);
+
+impl InstrIndex {
+    pub(crate) fn new(v: usize) -> Self {
+        // FIXME: propagate error
```
ltratt, line 84: I'd suggest fixing this ASAP, at least in the next PR, because adding `Result`s can change the layout of code a surprising amount.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
+        Self(u16::try_from(v).unwrap())
+    }
+
+    pub(crate) fn get(&self) -> u16 {
```
ltratt, line 88: The Rust convention for this would probably be to name it `as_u16` or `to_u16`.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
+pub(crate) struct ExtraArgsIndex(u16);
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub(crate) struct TypeIndex(u16);
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub(crate) struct ConstIndex(u16);
+
+impl ConstIndex {
+    pub(crate) fn get(&self) -> u16 {
```
ltratt, line 106: Same comment as above.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
+///  1             15
+/// +---+--------------------------+
+/// | k |         index            |
+/// +---+--------------------------+
+/// ```
+///
+///  - `k=0`: `index` is a local variable index
+///  - `k=1`: `index` is a constant index
+///
+///  The IR can represent 2^{15} = 32768 locals, and as many constants.
```
ltratt, line 125: I think it can represent `2^15-1` i.e. 32767.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
+            }
         }
     }
 
-    fn raw(&self) -> u64 {
-        match self {
-            Self::Long(_) => todo!(),
-            Self::Short(op) => op.0,
+    /// Unpacks a [PackedOperand] into a [Operand].
+    pub fn get(&self) -> Operand {
```
ltratt, line 144: Probably `to_operand` is the right name.

It might be better to do this as a `From` (or maybe just an `Into`) implementation?

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
+            Self::Const(idx) => write!(f, "Const({})", idx.get()), // FIXME
         }
     }
+}
+
+// FIXME: this isn't the correct representation of a constant.
+// It should be a bag of bytes and a type.
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum Constant {
+    Usize(usize),
```
ltratt, line 176: Note that this `enum` is now two machine words big. That's not a problem for now, but given that most usize` constants will be tiny (my guess is that > 50% will be the value `0`!) wasteful. Also some constants won't have a fixed size (e.g. `String)`. So at some point we'll want something more compact for the common case. I'm not sure if that's what's meant by "bag of bytes and a type" or not.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
-            // Then mask it out.
-            Operand::Short(ShortOperand(op & SHORT_OPERAND_MASK))
-        } else {
-            todo!()
+impl CallInstruction {
+    pub(crate) fn new(m: &mut Module, target: FuncIndex, args: &[Operand]) -> CallInstruction {
+        let mut arg1 = PackedOperand::default();
+        let mut extra = ExtraArgsIndex::default();
+
+        if args.len() >= 1 {
```
ltratt, line 309: Probably clearer as:

```rust
let arg1 = if args.len() == 0 { PackedOperand::default() } else { ... };
```

?

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
-    pub(crate) fn create_load(op: Operand) -> Self {
-        if op.is_short() {
-            let mut instr = Instruction::new_short(OpCode::Load);
-            instr.set_short_operand(op, 0);
-            instr
+    /// It is undefined behaviour to provide an out-of-bounds index.
+    pub(crate) fn operand(&self, m: &Module, idx: usize) -> Option<Operand> {
+        #[cfg(debug_assertions)]
+        {
+            let func = m.get_func(self.target);
```
ltratt, line 328: I'm not sure there's a readable one-liner here. But maybe something like:

```rust
match m.get_type(m.get_func(self.target).tyidx) {
  Type::Func(ft) => assert!(ft.num_args > idx),
  _ => panic!()
}
```

is clearer?

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
-            OpCode::LoadArg => true,
-        }
-    }
+/// Describes a function.
+#[derive(Clone, Debug)]
+pub(crate) struct Function {
+    /// The symbol name for the function.
+    name: String,
+    /// The type of the function.
+    tyidx: TypeIndex,
```
ltratt, line 350: Elsewhere `ty` has implied "index" without `idx` as a suffix. In fairness I think `idx` as a suffix makes things clearer. But whichever we choose, consistency is key.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: Ah, later on I realise that these are the first 3 elements in a module's type index so an `enum` probably wouldn't be a great idea. A clearer doc comment here would be good.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -26,297 +17,401 @@ const SHORT_INSTR_MAX_OPERANDS: u64 = 3;
...
+        };
+        // We reserve some type slots for common types. This means we can identify them by index
+        // without having to look into the type table.
+        //
+        // FIXME: Add more common types.
+        //
+        // The order of these pushes are important and must correspond with `TYPE_INDEX_*`;
+        this.push_type(Type::Void); // TYPE_INDEX_VOID
+        this.push_type(Type::Ptr); // TYPE_INDEX_PTR
+        this.push_type(Type::Unimplemented); // TYPE_INDEX_UNIMPLEMENTED
```
ltratt, line 413: Why not `types: vec![Type::Void, Type::Ptr, Type::Unimplemented]` rather than 3 pushes?

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -333,6 +428,118 @@ impl Module {
...
+    /// Push a new constant into the constant table and return its index.
+    pub(crate) fn push_const(&mut self, constant: Constant) -> ConstIndex {
+        let idx = self.consts.len();
+        self.consts.push(constant);
+        ConstIndex(u16::try_from(idx).unwrap()) // FIXME: propagate error
+    }
+
+    /// Get the index of a type, inserting it in the type table if necessary.
+    pub fn const_index(&mut self, c: &Constant) -> ConstIndex {
+        // FIXME: can we optimise this?
```
ltratt, line 460: How often might this function be called? If "a lot" then "use a hash table" might end up being the right advice. But for now a linear search is probably OK.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -333,6 +428,118 @@ impl Module {
...
+        for (idx, tc) in self.consts.iter().enumerate() {
+            if tc == c {
+                // const table hit.
+                return ConstIndex(u16::try_from(idx).unwrap()); // FIXME: propagate error
+            }
+        }
+        // type table miss, we need to insert it.
+        self.push_const(c.clone())
```
ltratt, line 468: Are we sure that `clone` is a good idea here? If `Constant` is a big chunk of memory, this will have unpleasant performance and maybe other weird consequences.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -333,6 +428,118 @@ impl Module {
...
+                // type table hit.
+                return FuncIndex(u16::try_from(idx).unwrap()); // FIXME: propagate error
+            }
+        }
+        // type table miss, we need to insert it.
+        self.push_func(func.clone())
+    }
+
+    /// Look up a [Function] by index. Panics if an out of bounds index is supplied.
+    pub(crate) fn get_func(&self, idx: FuncIndex) -> &Function {
```
ltratt, line 506: Why do we need `func_index` and `get_func`? I'm a bit unsure why I'd ever call `func_index` for a given function more than once.

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -144,6 +169,7 @@ impl<'a> TraceBuilder<'a> {
                 }
             }
         }
+        eprintln!("{}", self.jit_mod);
```
ltratt, line 172: Debugging statement left in?

vext01:
> Obviously, the jit_ir.rs file can still access the underlying vec directly. So if you want to avoid that too, then your solution would fix that.

The module can always see in innards of things in the same module, so I think what we have is the best you will get.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Oops, that's probably a left-over from an earlier design.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: According to the rust docs, the convention is to have a "Panics" markdown level 1 heading. Is that what you meant?

Like:
```
/// Inserts an element at position `index` within the vector, shifting all
/// elements after it to the right.
///
/// # Panics
///
/// Panics if `index` is out of bounds.
```

Note sure if the `!` has significance, or is a typo.

https://rust-lang.github.io/api-guidelines/documentation.html

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Agreed.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: It's the maximum value an operand index can assume. I don't think it makes sense to store a conceptually integer decimal value as a bit pattern or hex.

It makes sense for masks to be hex though.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: I don't understand this initial comment. The constant is explicitly typed `u16`, so I'm not sure what value annotating a `repr` would add.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: This is something we are quite guilty of across our code base. We don't have very unified terminology.

Examples:
 - `idx` vs `index`
 -  `func` vs `function`
 - `const` vs `constant`
 
It bothers me.

Note that `type` is special, in that it's a keyword in the rust grammar.

I guess we should use `ty` consistently.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: Right so `x << y` might be best here as it makes clearer that this isn't a random value but is a clear power of 2-1? In particular, I think the max value here is 1 greater than it should be.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Good point.

How about we make it return one of the following discrims:

```
enum NumArgs {
   Fixed(usize),
   Varargs(usize), // this many static args + varargs.
}
```

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Indeed. One of us will work on this immediately (in another PR), while the other works on something else. We figured we can work in parallel a bit more after we get something to iterate upon in.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: There's the repo as a whole, but let's at least make this part of the codebase (at least the bits added by this PR) consistent.

[Later I realised that `ty` and `tyidx` should probably be different things.]

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: I considered adding a `Deref` to all index types via a macro. Thoughts?

ykrt/src/compile/jitc_yk/aot_ir.rs
ltratt: Oops, I meant `# Panics`!

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: We are counting how many locals and constants we can represent. Index 32767 is the 32768th index, as indices start from 0. Or am I being silly?

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Since the conversion happens in both directions, I guess a `From` would be best, then you get `Into` for free (IIRC)?

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: Ah, sorry, I meant `2^15-3` because of the "reserved" types. Not sure if that's worth mentioning here though.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: `Deref` is more meant for smart pointers IMHO.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: Deal.

ltratt:
> The module can always see in innards of things in the same module, so I think what we have is the best you will get.

This isn't about *enforcing* visibility, per se, but making it clearer to our future selves that we really really don't want to use `remove` and friends (even though they can bypass our wrapper if they want). Anyway, not a big deal, and not a worry for this PR, but maybe something to think about in the future.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Agreed that this could be more compact, but I think this part of the IR should be easy to change without breaking lots of other parts of the IR.

> I'm not sure if that's what's meant by "bag of bytes and a type" or not.

A design I've seen in the past for representing constants is:
```
struct Const {
  ty: Type,
  bytes: Vec<u8>,
}
```

And the type tells you how to interpret the bytes.

A more optimal version would be:
```
constant_data: Vec<u8>; // somewhere in the module

struct Const {
  ty: Type,
  bytes: &[u8]
}
```

Or:
```
```
constant_data: Vec<u8>; // somewhere in the module

struct Const {
  ty: Type,
  cdata_idx: usize,
  cdata_len: usize, // maybe not needed, but you get the idea.
}
```

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: I think I agree, since we have different structs for types and type indices.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: I think we can resolve this for now, and think more about it later. Because most constants are small (`0` is almost certainly the most common constant!), though, I would prefer they don't all have to heap allocate.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Every time you "intern" a type. We can't expect the API user to know if a type has already been interned, so we do the check for them.

From experience with LLVM, you throw types at a module quite frequently and willy-nilly, so perhaps we should in fact have some kind of cache, but should we do that as another PR?

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: See e.g. the `Entry` API in hashmaps.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: At the moment `get_func()` is used only in a debug assertion.

`func_index()` is used only once while creating the initial JIT IR.

I don't know the answer to your question, but I expect the API to refine itself as we discover the use-cases required of it.

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: yes :P

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: > At the moment get_func() is used only in a debug assertion.

Can we guard it with `#[cfg(debug_assertions)]` or similar?

> func_index() is used only once while creating the initial JIT IR.

Right, so my question is: do we ever call this for a function F more than once? I kinda guess "no", but maybe I'm wrong.

ltratt:
Overall this looks like a real improvement over #930 -- excellent! My detailed comments have, I think, covered pretty much all the high level points I might pull out:

* Using `ty` and (I think) `tyidx` consistently would help readability a lot.
* I'm not a big fan of "lookup and insert if not found" functions, because (as we've seen) we end up not really knowing what to do in the "not found" case: clone or ... ? Something like `HashMap`s `Entry` API might provide inspiration, but even that might not be ideal.
* `get_` function names are generally not Rust-y (with some limited exceptions).

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: > I think the max value here is 1 greater than it should be.

Yes, I just realised that this is bogus. It should be all ones in the least-significant bits! Will fix.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: As far as I can see `entry()` unconditionally takes ownership of the key you give it.

https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.entry

So if the key is a large struct, can you rely on the compiler eliding copying it into `entry()`'s frame? I don't think so.

If not, then you copy the key even if the entry already exists, right? So if you expect more hits than misses (no idea if that's true), the it would be prudent to take a reference instead?

ltratt: >  So if the key is a large struct, can you rely on the compiler eliding copying it into entry()'s frame?

Nope, you can never rely on that sort of thing I'm afraid.

> So if you expect more hits than misses (no idea if that's true), the it would be prudent to take a reference instead?

My guess is that in many cases you can write the code to know whether the function has been encountered yet or not. Maybe I'm wrong.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: The reserved types are still types that should be accounted for thought, right? I don't think we need to make any distinction between reserved and other types.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: That being the case, I think it's best to accept a reference, as it gives you the flexibility to pass a reference to something you've already looked up and not incur a copy.

ykrt/src/compile/jitc_yk/jit_ir.rs
ptersilie: > [Later I realised that `ty` and `tyidx` should probably be different things.]

Yes, so in this case it should be `arg_tyidx`, or just `tyidx`. I think it's important to differentiate between an actual `Type` or its `TypeIndex`.

ykrt/src/compile/jitc_yk/jit_ir.rs
ptersilie: No, `push_type` does a bit more than just pushing a value into a vector here.

ptersilie: Actually, in this case it doesn't matter, since we are not doing anything with the index that `push_type` returns. So I guess your suggestion works.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Fixed in 02631b74d6a455fe7aaad5183daf16098dd5797e

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Fixed in 4b6a16b0d2138270895891c980d0f5f6a1993839.

Other "gets" fixed in:
 - 3e325437f3d12587a91a9aa61ac9346d66f4a18b
 - 3dfe284a77794bbfce6010be7aec635db16d550c

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Fixed in 919c1b90e93cb5e4ae6bd050a3641b3a6d6381ef

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Fixed in 4b6a16b0d2138270895891c980d0f5f6a1993839, then removed in c8333b87fab7833bd6584ed4234b0d27b4c545db.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Fixed in c62a43fe486af2b46a2cff5828622d89de73ea06

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: The pre-allocated JIT type indices were removed by the change that shares AOT type with the JIT IR: c8333b87fab7833bd6584ed4234b0d27b4c545db

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Since consts are not shared, this comment also remains outstanding.

The JIT will surely want to make its own constants for constant folding.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: `func_index()` and `get_func()` were deleted by the "sharing with AOT" commit: c8333b87fab7833bd6584ed4234b0d27b4c545db

vext01:
Most of the commits are referenced in replies to your comments. From memory the other stuff you may wish to additionally review (that wasn't obsoleted by a later change):

 - a832c23c41544b138956a8714ba8adbbf53b579c
 - 02dfe208eb7799c26a80aa9c05c389aafd4fdf02
 - aafb8f2bd96565c7af1057028b65ec0ab3398761
 - 407aa5a67fea23f831fb81f61541aaa2a18190d7

Sorry this is a bit hard to review. It's been one of those where you make a small change and then other things that really "should be fixed now (tm)" come up.

Stuff for future PRs that I held back on for fear of overloading the reviewer further:
 - Add remaining index types to AOT IR where there are currently type-unsafe `usize`s.
 - Audit AOT IR for `ty` and `ty_idx` consistency.
 - error propagation.
 - Encapsulate instruction vector in a new type to communicate that you can't remove from it.
 - clippy

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: I remember a conversation about sharing the constants with the AOT IR, but allowing the JIT IR to add constants to its own table.

Do we still want to try that?

The other thing we could do is only share constants with the AOT IR and tackle it later when we come to write optimisation passes.

ykrt/src/compile/jitc_yk/aot_ir.rs
ltratt: I'm fine with the `get`s but I'm not a fan of `Deref` (https://github.com/ykjit/yk/pull/943#discussion_r1466154063) which isn't used for these sorts of purposes - we should have a `to_u16` (or similar) method IMHO to be Rusty. https://doc.rust-lang.org/std/ops/trait.Deref.html's documentation is worth a read:

> Warning: Deref coercion is a powerful language feature which has far-reaching implications for every type that implements Deref. The compiler will silently insert calls to Deref::deref. For this reason, one should be careful about implementing Deref and only do so when deref coercion is desirable. See [below](https://doc.rust-lang.org/std/ops/trait.Deref.html#when-to-implement-deref-or-derefmut) for advice on when this is typically desirable or undesirable.
>
>Types that implement Deref or DerefMut are often called “smart pointers” and the mechanism of deref coercion has been specifically designed to facilitate the pointer-like behaviour that name suggests.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: > I think we opted to not share constants with the AOT IR for now?

We agreed the opposite, but I'm not completely wedded to that (ideally we don't want to do things like copy large constant strings because it's a waste) provided you're reasonably sure that changing this later wouldn't be difficult.

ltratt:
Broadly speaking, this is definitely going in the right direction! I have left some comments (including in commits), but mostly nothing major (the `Deref` comment aside).

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: What luajit does, it has a constant array, and then stores only a size and len into it. We could do this?

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Right.

FWIW, I did test the coercion cases I cared about and they didn't seem to work (in a bad way).

Nonetheless, I'm OK with `to_u16()` and/or `to_usize()` depending on use case, so let's do that.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: The challenge with "constants" is that they're not all present in source code. If you write `x = 2 + 3` you'll probably generate at compile-time the constant `5` even though it's not in the source.

ykrt/src/compile/jitc_yk/aot_ir.rs
ltratt: I would like to get rid of all the `Deref`s and replace them with explicit methods.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: Yeah, but in retrospect we can punt on it for this PR.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: > I'm more interested in the kinds of constant folding that can only be done by the JIT due to speculation.

That's my point!

In that sense, AOT constants and JIT constants *could* (but don't have to be!) different things. It's not something we *have* to get right now, but we should remain open to both possibilities IMHO.

vext01: OK, well for now, the JIT has its own constant table and doesn't share any constants the AOT module. That probably wouldn't be hard to change later.

What I think will bother you is that, due to the above, we still have the "find it or insert it" idiom (that you didn't like) for constants. We were able to kill that idiom for functions by sharing with AOT, but that's not what we are doing here.

Another way around this is to duplicate constants at the cost of memory. I don't like it though. I imagine some constants will come up time and time again like `0` and `1` and `-1`. If we duplicate constants, we also can't compare them for equality by a simple index comparison.

ltratt: OK let's move on.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: I've killed the `Deref`s that were added in this PR in d1ddec800717eeef6e31bd96b533e99a4c8bfcf2.

More pre-existing ones to kill in a future PR (I now have a list of follow-up PRs in my notebook).

vext01:
I *think* I've addressed everything now.

There will be follow up PRs.

ltratt:
Please squash.

vext01:
squash and updated the commit message to reflect the sharing changes.

PR #944, opened 2024-01-30
Tidy up index types.

vext01:
See commits.


vext01:
The failure is unrelated to this change (a non-deterministic crash in the LLVM backed, sigh).

But I'm glad it crashed, because I spotted some related bits i missed.

Requires squashing if ok.

ltratt:
Please squash

vext01:
Notice "InlineAsm is currently not supported."

It'd be good if we could get the error to print the unsupported inline asm.

PR #947, opened 2024-02-01
Update to the latest lang_tester.

ltratt:
Note that we don't use `comment_prefix("#")` in the IR tests as test lines in those tests begin with "#"! We could use another character but, at the moment, we have no need to add comments to any of those tests.

### DIFF
--- a/tests/Cargo.toml
+++ b/tests/Cargo.toml
@@ -36,7 +36,7 @@ yktracec = { path = "../yktracec", features = ["yk_testing"] }
 
 [dev-dependencies]
 criterion = { version = "0.5.1", features = ["html_reports"] }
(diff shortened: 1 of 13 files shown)


vext01:
Thanks for this.

I wonder if we could/should use the new `ignore-if` to rid the test suite of the "magic test filenames" stuff we have.

From memory:
 - tests starting `debug_` are only run in debug builds.
 - tests ending `.newcg.c` are only run for the new JIT backend.

That'd be a separate PR, so let's merge this.

ltratt:
> I wonder if we could/should use the new ignore-if to rid the test suite of the "magic test filenames" stuff we have.

Yes, probably! Agreed that it would be a separate PR.

PR #950, opened 2024-02-01
Fix a conditional and add some `assert`s to check if we've got UB.

ltratt:
In several cases we ask for elements of a vector (e.g. `index + 1`) without knowing for sure if those elements exist. It might be that these extra checks are redundant -- perhaps we know (or once knew) for sure that they're OK. But since the code doesn't make clear that this is safe, it's better for us to validate the implicit assumption.

### DIFF
--- a/yktracec/src/jitmodbuilder.cc
+++ b/yktracec/src/jitmodbuilder.cc
@@ -654,6 +654,7 @@ class JITModBuilder {
     // so that we can insert an appropriate guard into the trace. A block must
     // exist at `InpTrace[TraceIdx + 1]` because the branch instruction must
     // transfer to a successor block, and branching cannot turn off tracing.
+    assert(TraceIdx + 1 < InpTrace.Length());
     IRBlock *NextIB = InpTrace[TraceIdx + 1].getMappedBlock();
     assert(NextIB);
     BasicBlock *NextBB;
@@ -940,6 +941,7 @@ class JITModBuilder {
 
     // Get the instruction index of CPCI in its parent block.
     size_t CPCIIdx = 0;
+    assert(CPCI->getParent() != NULL);
     for (Instruction &I : *CPCI->getParent()) {
       if (&I == cast<Instruction>(CPCI)) {
         break;
@@ -1014,6 +1016,7 @@ class JITModBuilder {
     // There should never be two unmappable blocks in a row in the trace
     // (because the mapper collapses them to save memory).
     for (size_t I = 0; I < IL - 1; I++) {
+      assert(I + 1 < InpTrace.Length());
       assert(InpTrace[I].getMappedBlock() || InpTrace[I + 1].getMappedBlock());
     }
 #endif
@@ -1255,7 +1258,7 @@ public:
             // code. As far as I can tell, there's only one intrinsics that
             // allows this: llvm.init.trampoline. Investigate if this is a
             // problem.
-            if (Idx < InpTrace.Length()) {
+            if (Idx + 1 < InpTrace.Length()) {
               if (UnmappableRegion *UR =
                       InpTrace[Idx + 1].getUnmappableRegion()) {
                 // The next block is unmappable, which means this intrinsic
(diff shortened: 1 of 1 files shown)


yktracec/src/jitmodbuilder.cc
```diff
@@ -654,6 +654,7 @@ class JITModBuilder {
     // so that we can insert an appropriate guard into the trace. A block must
     // exist at `InpTrace[TraceIdx + 1]` because the branch instruction must
     // transfer to a successor block, and branching cannot turn off tracing.
+    assert(TraceIdx + 1 < InpTrace.Length());
```
vext01, line 657: The comment explains that is safe, but we may as well keep the assertion in. Agree?

ltratt: Personally I think the `assert` complements the comment, but up to you!

yktracec/src/jitmodbuilder.cc
```diff
@@ -940,11 +941,13 @@ class JITModBuilder {
 
     // Get the instruction index of CPCI in its parent block.
     size_t CPCIIdx = 0;
+    assert(CPCI->getParent() != NULL);
```
vext01, line 944: I have a feeling this `CPCI->getParent() != NULL` is always true, because `F->user_back()` is an iterator-like interface, and iterators don't (to my knowledge) return `NULL` if the iterator is empty.

Maybe we need to check against `F->user_end()` instead?

yktracec/src/jitmodbuilder.cc
```diff
@@ -940,11 +941,13 @@ class JITModBuilder {
...
     for (Instruction &I : *CPCI->getParent()) {
       if (&I == cast<Instruction>(CPCI)) {
         break;
       }
       CPCIIdx++;
+      assert(CPCI->getParent() != NULL);
```
vext01, line 950: Line 945 would have crashed before this assertion could be checked. Move it up?

It's actually impossible that `CPCI` is unparented, since it's been extracted from the AOT module. No instruction in the AOT module is unparented by construction.

I'd just remove this assertion personally.

yktracec/src/jitmodbuilder.cc
```diff
@@ -1013,7 +1016,9 @@ class JITModBuilder {
     assert(InpTrace[0].getMappedBlock());
     // There should never be two unmappable blocks in a row in the trace
     // (because the mapper collapses them to save memory).
+    assert(IL > 0);
```
vext01, line 1019: This is a duplicate of line 1013.

yktracec/src/jitmodbuilder.cc
vext01: Keep it in.

yktracec/src/jitmodbuilder.cc
ltratt: Agreed. Fixed in a202260c.

yktracec/src/jitmodbuilder.cc
vext01: Hang on. I was misleading. Let me explain better.

`CPCI->getParent() != NULL` is the correct way to check if `CPCI` has a parent.

But `CPCI` itself is an iterator returned from `F->user_back();`. If `F` has no user, then `CPCI` is going to be the iterator's begin element(?) and calling `CPCI->getParent()` on it might cause a crash. I'm not 100% sure, but this sure feels fishy.

We could, in addition to this assertion, check `F->user_back() != F->user_begin()`?

Really we should check my above reasoning, but it would take me some time. Do you want me to?

ltratt: `assert(F->user_back() != F->user_begin());` leads to:

```
src/jitmodbuilder.cc:944:27: error: invalid operands to binary expression ('User *' and 'user_iterator' (aka 'user_iterator_impl<User>'))

  cargo:warning=  944 |     assert(F->user_back() != F->user_begin());
```

We don't want to spend too much time on this, but if you happen to know a quick fix to improve the `assert` it wouldn't hurt.

vext01: Your initial assertion was correct I think.

`user_back()` doesn't return an iterator, my bad.

vext01:
Please squash.

ltratt:
Squashed.

PR #951, opened 2024-02-01
Remove the LLVM JIT compiler ("trace compiler") test suite.

ltratt:
This is a very fragile test suite that percolates throughout the existing trace compiler in a way that makes it quite hard to make changes to it (e.g. to support software tracing).

In the short term we want to tweak the LLVM trace compiler to make it easier to support software tracing. In the medium term we want to remove the LLVM trace compiler entirely. Fixing the test suite is rather annoying because of its fragility, and not particularly beneficial to us, so it's easiest to remove it now and be done with it.

There are probably more bits we can remove and simplify, but this commit does enough of a job on its own to be a noticeable help.


yktracec/src/ykllvmwrap.cc
```diff
@@ -461,13 +461,6 @@ void *compileIRTrace(FN Func, char *FuncNames[], size_t BBs[], size_t TraceLen,
     rewriteDebugInfo(JITMod, TraceName, DebugInfoFD,
                      filesystem::path(DebugInfoPath));
 
-#ifdef YK_TESTING
```
vext01, line 464: A bit above, you should be able to kill `template <typename FN>` since there is now only one viable function to call.

vext01:
This is great! Look at all that complexity melt away!

yktracec/src/ykllvmwrap.cc
ltratt: I plan to have a deeper look at that in a follow-up PR, although I could do it as a commit here if you'd like?

vext01: as you wish.

ptersilie:
:+1:

PR #953, opened 2024-02-01
Use lang tester's `ignore-if` to avoid encoding configuration in test names

ltratt:
This PR, in three stages, removes the need to encode "only run this C test if X" using lang_tester's new `ignore-if`. Note that we have to use `set_env` so that the environment variables are available to `ignore-if`. Maybe lang_tester should grow the ability to have an environment just to run `ignore-if` commands in -- but that's a worry for another day.


tests/c/simple.newcg.c
```diff
@@ -1,3 +1,4 @@
+// ignore-if: test YK_JIT_COMPILER != "yk"
```
vext01, line 1: Can we just pick up `YKD_NEW_CODEGEN` here?

https://ykjit.github.io/yk/dev/runtime_config.html#ykd_new_codegen

vext01:
Great. Simplifies a lot of stuff!

Just one small comment.

tests/c/simple.newcg.c
ltratt: `YKD_NEW_CODEGEN` is like calling a road "new road" -- when you build another road should it be called "newer road"? The directory structure in ykrt/compiler gives these things names, so we should IMHO use those names and not a boolean `NEW_CODEGEN` env var.

vext01: Why don't we push a commit to rename the existing `YKD_NEW_CODEGEN` env var and use that instead? I don't see any benefit to having two environment variables for the same purpose.

ltratt: That's probably a good idea, but the `YK_` precedent is already present in the test infrastructure, so renaming `YKD_NEW_CODEGEN` feels like a separate PR to me (and best done by someone who knows exactly what `YKD_NEW_CODEGEN` does).

vext01: I agree. So this PR should follow precedent, and a follow up PR can do the rename.

vext01: (I can do the rename)

ltratt: Thank you!

ltratt:
@vext01 I *think* this is OK to merge now?

vext01:
I meant this PR should use the existing env var in the interim, but don't worry, I'll sort it in a follow up PR :P

vext01:
Is it possible that some bad logic is causing *all* tests to be run if `YKD_NEW_CODEGEN=1`?

ltratt:
@vext01 Hopefully that new commit fixes things.

ltratt:
Squashed.

PR #954, opened 2024-02-02
Move control point block removal to hwt

ltratt:
This PR moves the "ignore the first block" logic, which is only relevant for hwt, into hwt. The first commit does this brutally but makes what's going on clear; the second commit does so more elegantly and does a partial check that we've removed the right thing.

### DIFF
--- a/yktracec/src/jitmodbuilder.cc
+++ b/yktracec/src/jitmodbuilder.cc
@@ -1052,9 +1052,7 @@ public:
   Module *createModule() {
     size_t CurBBIdx;
     size_t CurInstrIdx;
-    // Iterate over the blocks of the trace. The first block contains the
-    // control point so we want to start copying instructions just after that.
-    for (size_t Idx = 1; Idx < InpTrace.Length(); Idx++) {
+    for (size_t Idx = 0; Idx < InpTrace.Length(); Idx++) {
       // Update the previously executed BB in the most-recent frame (if it's
       // mappable).
       TraceLoc Loc = InpTrace[Idx];
--- a/ykrt/src/trace/hwt/mapper.rs
+++ b/ykrt/src/trace/hwt/mapper.rs
@@ -134,7 +134,21 @@ impl HWTMapper {
     ) -> Result<Vec<TracedAOTBlock>, Box<dyn Error>> {
         let mut ret: Vec<TracedAOTBlock> = Vec::new();
 
-        for (i, block) in trace.iter_blocks().enumerate() {
+        let mut trace_iter = trace.iter_blocks();
+        // The first block contains the control point so we need to remove that.
+        match trace_iter.next() {
+            Some(Ok(x)) => {
+                // As a rough proxy for "check that we removed only the thing we want to remove",
+                // we know that the control point will be contained in a single mappable block.
+                assert!(matches!(
+                    self.map_block(&x).as_slice(),
+                    &[Some(TracedAOTBlock::Mapped { .. })]
+                ));
+            }
+            _ => unreachable!(),
+        }
+
+        for (i, block) in trace_iter.enumerate() {
             if i > crate::mt::DEFAULT_TRACE_TOO_LONG {
                 return Err("Trace too long".into());
             }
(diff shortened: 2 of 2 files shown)


yktracec/src/jitmodbuilder.cc
```diff
@@ -1052,9 +1052,7 @@ class JITModBuilder {
   Module *createModule() {
     size_t CurBBIdx;
     size_t CurInstrIdx;
-    // Iterate over the blocks of the trace. The first block contains the
```
vext01, line 1055: Shall we add a comment saying what the first block should be now?

You might be able to add an assertion that the first block is a successor of the block containing the control point?

ltratt: TBH, I don't know what the first block should be, nor do I know how to encode that assertion!

vext01: > I don't know what the first block should be

A successor of the block containing the control point call.

>  nor do I know how to encode that assertion

OK, not worth worrying about.

ltratt: Is it always a direct successor. I guess I assumed it could be a function call? Anyway, maybe not a big worry?

vext01: You're right, it could be an unmappable block inside a function call.

vext01:
Hah, the new trace compiler backend looks at the block you just removed to find the trace inputs.

This answers a question in my PR:

https://github.com/ykjit/yk/pull/956/files#diff-0019b0cfa54cfb8c953a80e8ab4e0ce900b5fada3100f573c59c09feadd32fc6R172-R174

ltratt:
Oh! What should we do?

vext01:
If you want to keep this design, quickish fix: search the AOT IR for the control point call.

Better fix: extend ykllvm to encode where to find the control point call into the AOT IR. i.e. at compile time.

ptersilie:
This change doesn't affect the way `jitmodbuilder.cc` currently searches for the control point call (see https://github.com/ykjit/yk/blob/master/yktracec/src/jitmodbuilder.cc#L933). Of course this won't work if there's multiple control point calls (note the `FIXME`), but its probably not worth fixing here.

We should merge this and then do the correct thing in the new trace builder, i.e. search for the control point block manually, which should be the predecessor of the first block in the trace.

ptersilie:
@ltratt Do you want to squash?

ltratt:
I think the two commits are OK separately.

ltratt:
@ptersilie Now I remember the problem: I don't think the new code gen works with this change!

But that said, the error is a weird PT error I've never seen before (this is one for @vext01 to look at):

```
12:28:44    |jit-state: start-tracing
12:28:44 >> |thread '<unnamed>' panicked at hwtracer/src/pt/ykpt/mod.rs:453:13:
12:28:44    |assertion failed: taken
12:28:44    |note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
12:28:44    |fatal runtime error: failed to initiate panic, error 5
```

So maybe @ptersilie try rerunning CI and see if we get to the "genuine" error I think this PR might cause?

vext01:
That's an error I've only ever seen when compressed returns are enabled. We disabled them a while back. When did you get this error? Today?

vext01:
Right, in this PR. Investigating...

vext01:
I re-ran the build and got the failure I'd expect (not the PT 'taken' error):
https://ci.soft-dev.org/#/builders/1/builds/3560/steps/2/logs/stdio

It is very curious that the "taken" thing comes up with compressed returns disabled. This suggests a PT packet parsing bug elsewhere.

I few month ago I spent 3 days scrutinising the compressed return logic and matching it up with the Intel manual. I was unable to find anything "dodgy".

vext01:
I suggest we don't worry about the "taken" bug for now. It's very very rare, and we have an issue up for it, which I've just updated with our new information.

ltratt:
@vext01 @ptersilie I suggest one of you try merging again (I expect the PR to fail but not for the PT reasons).

vext01:
I linked to the error I expect above. We'd need to teach the new trace builder to find the control point from the successor block.

PR #955, opened 2024-02-02
Encapsulate the trace vector.

ptersilie:
Since references to instructions inside a trace are represented by their index, certain operations, like removal and insertion of instructions (except appending at the end), are disallowed. In order to make it more difficult to accidentally do that, this commit encapsulates the vector storing the instructions of a trace, only exposing operations that are safe to do.

Additional notes: I kinda cheated here, since I manually implemented an `iter` method which just returns the underlying vec's `iter`. If we wanted to iterate over the trace vector using a simple for loop, e.g. `for instr in tracevec`, then we would have to implement `Iterator`, which is a teeny bit more involved. But I decided to wait and see how we are going to access a trace, once we moved to the optimisation/compilation stages for our new IR, to see if it's really needed.

### DIFF
--- a/ykrt/src/compile/jitc_yk/jit_ir.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir.rs
@@ -7,6 +7,7 @@
 #![allow(dead_code)]
 
 use super::aot_ir;
+use super::tracevec::TraceVec;
 use std::{fmt, mem, ptr};
 
 /// Bit fiddling.
@@ -383,7 +384,7 @@ pub(crate) struct Module {
     /// The name of the module and the eventual symbol name for the JITted code.
     name: String,
     /// The IR trace as a linear sequence of instructions.
-    instrs: Vec<Instruction>,
+    instrs: TraceVec,
     /// The extra argument table.
     ///
     /// Used when a [CallInstruction]'s arguments don't fit inline.
@@ -401,7 +402,7 @@ impl Module {
     pub fn new(name: String) -> Self {
         Self {
             name,
-            instrs: Vec::new(),
+            instrs: TraceVec::new(),
             extra_args: Vec::new(),
             consts: Vec::new(),
         }
--- a/ykrt/src/compile/jitc_yk/mod.rs
+++ b/ykrt/src/compile/jitc_yk/mod.rs
@@ -51,6 +51,7 @@ static PHASES_TO_PRINT: LazyLock<HashSet<IRPhase>> = LazyLock::new(|| {
 pub mod aot_ir;
 pub mod jit_ir;
 mod trace_builder;
+pub mod tracevec;
 
 pub(crate) struct JITCYk;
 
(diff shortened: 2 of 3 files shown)


ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -383,7 +384,7 @@ pub(crate) struct Module {
     /// The name of the module and the eventual symbol name for the JITted code.
     name: String,
     /// The IR trace as a linear sequence of instructions.
-    instrs: Vec<Instruction>,
+    instrs: TraceVec,
```
ltratt, line 387: Should this be `InstrVec` or `VecInstrs` or similar?

ptersilie:
Another advantage of the wrapper is that we can later have a `remove` method on it, which simply inserts a `NOP` or similar, so we don't have to do that manually (e.g. from within an optimisation pass).

ykrt/src/compile/jitc_yk/jit_ir.rs
ptersilie: Either sounds fine to me. Maybe `InstrVec` is slightly better?

vext01:
LGTM!

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: I prefer `InstrVec`

ltratt:
I wonder if we should use `TiVec` as @vext01 did in another recent PR so that when (as we will surely want to) we use indices to look things up, we have to use a specific index type? https://crates.io/crates/typed-index-collections

[I don't feel strongly about this, but it sounds like it might be useful here.]

ltratt:
Hmm, though `TiVec` would let you `remove` elements. So maybe we need a new type `GrowOnlyVec<T, K>` that's otherwise like `TiVec`? We could even bung that in a crate if we really wanted to!

vext01:
Me an Lukas discussed this the other day and decided to wait and see how the trace is accessed.

If we don't juggle indices very much, it would only add complexity (although not much).

I don't object if someone wants to do that though.

ltratt:
We are bound to juggle indices when optimising because you often have to look forward/backwards a number of elements, and that's a pain to do with iterators.

vext01:
Ok, well it shouldn't be a difficult change.

ptersilie:
So what's the actionable change here now?

vext01:
Replace the inner vector with a `TiVec<InstrIndex, Instruction>` and fix fallout.

ptersilie:
Done.

vext01:
Well that was easy enough.

Are we still doing the renaming to `InstrVec`?

ltratt:
I've been chewing on this PR and I now think we shouldn't merge it:

1. It doesn't model much of a `Vec`. Scaling up to do things like `[...]` and slices and iterations is more work and, indeed, much harder than this PR.
2. Probably the most common mistake we're going to make is mixing up indexes between structures. `TiVec` will save us from that. Accidentally calling `remove` does feel like a less common mistake we'll make.

I take the blame for suggesting this PR -- sorry!

ptersilie:
I did voice those exact concerns when I started working on this, but I may have not been able to convey them properly without making the actual changes. So I agree with you that this is more hassle than it's worth and shouldn't be merged.

PR #956, opened 2024-02-02
Compile error propagation

vext01:
See commits. Main work is the second.

### DIFF
--- a/ykrt/src/compile/mod.rs
+++ b/ykrt/src/compile/mod.rs
@@ -9,9 +9,7 @@ use parking_lot::Mutex;
 use std::slice;
 use std::{
     collections::HashMap,
-    env,
-    error::Error,
-    fmt,
+    env, fmt,
     sync::{
         atomic::{AtomicU32, Ordering},
         Arc, Weak,
@@ -28,6 +26,15 @@ pub(crate) mod jitc_llvm;
 #[cfg(jitc_yk)]
 pub mod jitc_yk;
 
+/// A failure to compile a trace.
+#[derive(Debug, thiserror::Error)]
+pub enum CompilationError {
+    #[error("Unrecoverable error: {0}")]
+    Unrecoverable(String),
+    #[error("Temporary error: {0}")]
+    Temporary(String),
+}
+
 /// The trait that every JIT compiler backend must implement.
 pub(crate) trait Compiler: Send + Sync {
     /// Compile a mapped trace into machine code.
@@ -37,10 +44,10 @@ pub(crate) trait Compiler: Send + Sync {
         irtrace: Vec<TracedAOTBlock>,
         sti: Option<SideTraceInfo>,
         hl: Arc<Mutex<HotLocation>>,
-    ) -> Result<CompiledTrace, Box<dyn Error>>;
+    ) -> Result<CompiledTrace, CompilationError>;
 }
 
-pub(crate) fn default_compiler() -> Result<Arc<dyn Compiler>, Box<dyn Error>> {
+pub(crate) fn default_compiler() -> Result<Arc<dyn Compiler>, CompilationError> {
     #[cfg(jitc_yk)]
     // Transitionary env var to turn on the new code generator.
     //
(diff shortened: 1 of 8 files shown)


ykrt/src/compile/mod.rs
```diff
@@ -28,6 +27,50 @@ pub(crate) mod jitc_llvm;
...
+    Temporary(TemporaryErrorKind),
+}
+
+impl CompilationError {
+    pub fn new_jitir_index_overflow(which: String) -> Self {
+        Self::Temporary(TemporaryErrorKind::JITIRIndexOverflow(which))
+    }
+}
+
+pub type CompilationResult<T> = Result<T, CompilationError>;
```
ltratt, line 45: I must admit I really dislike `type` aliases used in this way: they save very little typing (in this case 6 characters) but now I have to know what they alias. So I find they make the code worse. I appreciate this is a personal quirk of mine so feel free to ignore if you disagree.

CompilationResult<T>
Result<T, CompilationError>

ykrt/src/compile/mod.rs
```diff
@@ -28,6 +27,50 @@ pub(crate) mod jitc_llvm;
 #[cfg(jitc_yk)]
 pub mod jitc_yk;
 
+/// A failure to compile a trace.
+#[derive(Debug, thiserror::Error)]
+pub enum CompilationError {
+    #[error("Unrecoverable error: {0}")]
+    Unrecoverable(String),
+    #[error("Temporary error: {0}")]
+    Temporary(TemporaryErrorKind),
```
ltratt, line 36: Do we gain anything useful from `TemporaryErrorKind`? My guess is that, at least at the moment, we will not find its contents valuable other than printing them out for debugging. Perhaps in most cases we even want a `temporary_compilation_error!()` macro which expands to `CompilationError::Temporary(format!("Line {}", line!())` or similar? It cuts down on us having to type stuff but means that when we're debugging we'll know *which* temporary error caused a problem.

ykrt/src/compile/mod.rs
vext01: I think this is one of the few cases where a type alias is acceptable. It's an idiom in use in the rust std lib, so I'm actually OK with it.

@ptersilie do you have any preference?

ltratt: Yes it is used in the standard library, but it really makes a mess there too IMHO.

ykrt/src/compile/mod.rs
vext01: I think you are right. We don't actually care what's inside. I just copied precedence from [elsewhere](https://github.com/ykjit/yk/blob/7dc01b9ccc79bf629a2136cf545bc6db35c3ffd0/hwtracer/src/errors.rs#L15).

Maybe that should be updated at some point too (but not now).

vext01: Fixed in https://github.com/ykjit/yk/pull/956/commits/ab849eb9d74e139a62c5466dca6f2c283162f1e5

ykrt/src/compile/mod.rs
vext01: Changed in https://github.com/ykjit/yk/pull/956/commits/b169db4fbe9dcbb8e3b3bad5f4eeffd0e0b3ddd1

ltratt:
Please squash.

vext01:
splat.

ltratt:
@vext01 Those are genuine (if minor) errors, which will need fixing.

vext01:
I guess this is the way to fix it? https://github.com/ykjit/yk/pull/956/commits/55ded243bb6bc95a455e01cc92a0f9a9d8926dcc

ltratt:
> I guess this is the way to fix it?

Nope :) You'll need to conditionally compile the whole match arm.

vext01:
Yeah, i was hoping not to...

vext01:
Someone meh fix in https://github.com/ykjit/yk/pull/956/commits/da2a6133ff707b501a823dce2fad87499fb70c55

ltratt:
Hmm, I stand corrected. https://github.com/ykjit/yk/commit/55ded243bb6bc95a455e01cc92a0f9a9d8926dcc is less evil. Please revert da2a613.

vext01:
OK to squash?

ltratt:
Please squash.

vext01:
splat.

PR #957, opened 2024-02-06
First block bug

vext01:
This fixes the bug where the new JIT compiler expects the first block in the trace to contain the control point.

This requires an upcoming change from Laurie before it can be merged.


vext01:
This should be good to go now. OK to squash?

ltratt:
Please squash.

vext01:
splat.

PR #959, opened 2024-02-07
Move "ignore the very last block" into the hwt mapper.

ltratt:
The last PT-mapped block contains unmappable code that stops tracing. Other backends will not have this block, so handling it in jitmodbuilder is wrong. This commit removes this block in `hwt/mappper.rs` instead.

### DIFF
--- a/yktracec/src/jitmodbuilder.cc
+++ b/yktracec/src/jitmodbuilder.cc
@@ -1293,7 +1293,7 @@ public:
           continue;
         }
 
-        if (Idx > 0 && Idx < InpTrace.Length() - 1) {
+        if (Idx < InpTrace.Length()) {
           // Stores into YkCtrlPointVars only need to be copied if they appear
           // at the beginning or end of the trace. Any YkCtrlPointVars stores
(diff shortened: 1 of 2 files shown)


yktracec/src/jitmodbuilder.cc
```diff
@@ -1314,8 +1314,8 @@ class JITModBuilder {
                 I++;
                 CurInstrIdx++;
               }
-              assert(InpTrace.Length() > 2);
-              if (Idx == InpTrace.Length() - 2) {
+              assert(InpTrace.Length() > 1);
```
ptersilie, line 1317: I believe this assert is no longer needed. It probably was incorrect in the first place as we only want to make sure that `InpTrace.Length() - 1 ` is valid. But we already know that since we wouldn't have entered the loop if the length was 0.

ltratt: Agreed. Removed in b7cb83ae3252d5401e3d5fe706b7b1fadfddbff7.

ptersilie:
Please squash if needed.

PR #962, opened 2024-02-08
Remove unused method.

ltratt:
Newer Rust nightlies spot that this is unused, causing our `.buildbot.sh` to fail.

This is responsible for the recent failure of https://github.com/ykjit/yk/pull/959 to pass CI.

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -108,13 +108,6 @@ fn deserialise_into_ti_vec<I, T>(v: Vec<T>) -> Result<TiVec<I, T>, DekuError> {
 pub(crate) trait IRDisplay {
     /// Return a human-readable string.
     fn to_str(&self, m: &Module) -> String;
-
-    /// Print myself to stderr in human-readable form.
-    ///
-    /// This is provided as a debugging convenience.
-    fn dump(&self, m: &Module) {
-        eprintln!("{}", self.to_str(m));
-    }
 }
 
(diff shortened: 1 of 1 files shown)


vext01:
We use this method for debugging, so I don't want to remove it. Can we mark it used somehow?

ltratt:
Given that `eprintln!("{}", x.to_str())` is equivalent, this feels like a really low value method to leave lurking as dead code...

ltratt:
[TBH, I'm not even sure why we have `to_str` instead of just `impl Display` or `Debug`, but that's a slightly separate issue.]

vext01:
> TBH, I'm not even sure why we have to_str instead of just impl Display or Debug

Because you require the module in order to print the instructions sensibly and those traits don't provision for passing in user data.

> Given that eprintln!("{}", x.to_str()) is equivalent...

I'm going to push back.

When I'm debugging something difficult, the harder it is for me to do print debugging, the less productive I am.

Typing (and remembering that this is how you have to do it):

```
eprintln!("{}", x.to_str(m))
```

insead of:
```
x.dump(m)
```

Can really distract from the (usually difficult) problem I'm trying to solve. And this frustrates me.

I know you don't attach much weight to LLVM design decisions, but they do the same and (IMHO) got it right. In LLVM it's:

```
X.dump();
```

instead of:

```
X.print(errs());
```

Please allow us this :)

ltratt:
Please can we unblock CI first then have this (old) argument in a later PR?

ltratt:
@vext01 is going to take a different approach to this. Closing.

PR #968, opened 2024-02-08
Slightly fiddle with `yk_promote`.

ltratt:
This PR slightly fiddles with `yk_promote`, including making it properly generic (though we currently only have 1 implementation!). If this is OK, I will probably then try and move `promote.rs` from `yktracec` to `ykrt`, which seems like a more natural home for it (but maybe difficult for reasons I don't yet know).

[`yktracec` is an annoying crate because any changes to it cause long compile times. I might try and address that separately at some point, but ultimately I'm not sure we want this crate to exist as a separate thing anyway.]

### DIFF
--- a/ykcapi/yk.h
+++ b/ykcapi/yk.h
@@ -67,11 +67,9 @@ YkLocation yk_location_new(void);
 // will occur.
 void yk_location_drop(YkLocation);
 
-// Promote a value to a constant.
-//
-// Yk will attempt to promote the argument to a constant when the call is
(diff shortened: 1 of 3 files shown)


vext01:
I hadn't heard of `_Generic` before this. I must admit, I didn't realise the preprocessor knew a jot about types!

This all looks OK, I think.

Do you want to squash?

ltratt:
`_Generic` is C11 and hasn't seen wide-spread use AFAIK, but it makes things like this much easier!

Squashed.

PR #971, opened 2024-02-09
New codegen backend: Start emitting code.

vext01:
This makes a start at emitting X86_64 code from the JIT IR.

Obviously this is non-functional at this point (it's currently not even called from the JIT pipeline), but should serve as something we can iterate upon and at least unit test in isolation.

Many things missing:
 - Trace input handling.
 - Correct allocation sizes.
 - Stackmaps
 - Debugger support.
 - Loads more testing.
 - Cross-arch testing support.

This PR is already large, so I've held off doing those for now.

Raising as a draft, as I think this could perhaps use some more tests.


ykrt/src/compile/jitc_yk/codegen/x86_64.rs
```diff
@@ -0,0 +1,252 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::CodeGen;
+    use crate::compile::jitc_yk::{aot_ir, codegen::tests::match_asm, jit_ir};
+
+    #[test]
+    fn simple_codegen() {
+        let mut aot_mod = aot_ir::Module::default();
```
vext01, line 227: @ltratt @ptersilie  take a look at this test. Here we have to build an AOT module so that the JIT module can refer to parts of it. It's a bit awkward.

We could make the JIT module independent of the AOT module at the cost of copying. Is it worth the performance hit for testing?

ykrt/src/compile/jitc_yk/codegen/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+    frame_off: usize,
+}
+
+impl LocalAlloc {
+    fn new(frame_off: usize) -> Self {
+        Self { frame_off }
+    }
+}
+
+impl LocalAlloc {
```
ltratt, line 42: Merge the `impl`s?

ykrt/src/compile/jitc_yk/codegen/mod.rs
```diff
@@ -0,0 +1,218 @@
+//! The JIT's Code Generator.
+
+// FIXME: eventually delete.
+#![allow(dead_code)]
+
+use super::{jit_ir, CompilationError};
+use dynasmrt::ExecutableBuffer;
+#[cfg(any(debug_assertions, test))]
+use dynasmrt::{AssemblyOffset, DynasmApi};
+#[cfg(any(debug_assertions, test))]
```
ltratt, line 10: Why always `debug_assertions` or `test`?

ykrt/src/compile/jitc_yk/codegen/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+}
+
+impl LocalAlloc {
+    fn frame_off(&self) -> usize {
+        self.frame_off
+    }
+}
+
+/// The result of running the code generator.
+pub(crate) struct CodeGenOutput {
```
ltratt, line 49: Wouldn't `CompiledTrace` or `ExecutableTrace` or similar be clearer?

ykrt/src/compile/jitc_yk/codegen/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+        let tp = tp.set_comments(&self.comments);
+        tp.to_string()
+    }
+}
+
+/// The code generator.
+pub(crate) struct CodeGen<'a> {
+    jit_mod: &'a jit_ir::Module,
+    #[cfg(target_arch = "x86_64")]
+    asm: dynasmrt::x64::Assembler,
```
ltratt, line 73: I think we're immediately baking in `x86_64` here in such a way that we won't unbake it later. So either we call this `X86CodeGen` and don't pretend it's anything else, or we split things into machine independent and machine dependent parts now and get it (as) right (as we can).

ykrt/src/compile/jitc_yk/codegen/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+    /// Abstract stack pointer, as a relative offset from `RBP`. The higher this number, the larger
+    /// the JITted code's stack. That means that even on a host where the stack grows down, this
+    /// value grows up.
+    asp: usize,
+    /// Maps each JIT IR local variable to where the JIT will store it at the machine level.
+    ///
+    /// Conceptually this is a hashmap, but since the keys are integers and there can never be
```
ltratt, line 80: I think it's conceptually a vector!

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
```diff
@@ -0,0 +1,252 @@
...
+/// We choose callee-save registers so that we don't have to worry about storing/restoring them
+/// when we do a function call to external code.
+static WR0: Rq = Rq::R12;
+static WR1: Rq = Rq::R13;
+static WR2: Rq = Rq::R14;
+
+/// The X86_64 SysV ABI requires a 16-byte aligned stack prior to any call.
+const SYSV_CALL_STACK_ALIGN: usize = 16;
+
+impl CodeGen<'_> {
```
ltratt, line 29: Ah, I see that you're effectively extending `CodeGen` here. The problem with this is that it means you can only ever have one code gen compiled in. That means we can't ever have e.g. a slow & fast code gen that we swap between at run-time. It might make testing more complex too?

Probably the machine-dependent part should `impl` a `trait` of some kind.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
```diff
@@ -0,0 +1,252 @@
...
+        let alloc = LocalAlloc::new(self.asp);
+        self.local_allocs.push(alloc);
+        alloc
+    }
+
+    fn store_local(&mut self, l: &LocalAlloc, reg: Rq, size: usize) {
+        debug_assert!(l.frame_off <= self.asp);
+
+        match i32::try_from(l.frame_off) {
+            Ok(off) => match size {
```
ltratt, line 129: I suggest:

```rust
Ok(8) => ...
Ok(4) => ...
Err(...)
```

is clearer.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
ltratt: > Is it worth the performance hit for testing?

If it's the difference between "testing is intolerable" and "testing is tolerable" then I think "yes".

ltratt:
This is a decent start! My main suggestion (other than the testing aspect) is that the PR currently uses `impl` extension in a way that's not very idiomatic and also very constraining. I think we probably want something like:

```
struct MachineIndependentStuff { ... }

impl MachineIndependentStuff { ... }

trait MachineDependent {
  fn emit_blah(...)
}
```

and then `x86` `impl`s that `trait`. This is a very rough sketch, though, obviously!

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
vext01: @ptersilie what do you think?

ykrt/src/compile/jitc_yk/codegen/mod.rs
vext01: - tests use the asmprinter to match the textual representation
 - it's also useful for debugging
 - it costs to make the disassembly, so I opted to not have this functionality in release builds, but we could re-think that?

ykrt/src/compile/jitc_yk/codegen/mod.rs
vext01: There's already a struct called `CompiledTrace` elsewhere, and it's not just that, it contains some stuff required for doing a disassembly with "comments" which can be matched in tests and are useful for eyeballing disassemblies.

ykrt/src/compile/jitc_yk/codegen/mod.rs
vext01: There's a comment to that effect someplace, and this is what I meant with "cross-arch testing" in the PR description.

As I said, I don't want to tackle this in the first PR, but if you think it's worthwhile (and I do think it is), we should address this in a follow up PR.

ykrt/src/compile/jitc_yk/codegen/mod.rs
vext01: It maps `InstrIdx`s to their allocations, so I consider this an optimised map.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
vext01: Yep, as discussed above and in comments in the code. One for a follow up PR.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
vext01: I agree. Not sure why I did it like that.

ykrt/src/compile/jitc_yk/codegen/mod.rs
ltratt: OK... but I think we can still find a better name than `Output`?

ykrt/src/compile/jitc_yk/codegen/mod.rs
ltratt: OK we're probably in the category of bikeshedding terminology here :)

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
ltratt: I think we should improve this in this PR, because it's quite a fundamental decision.

vext01:
I agree with your concerns about a statically compiled-in backend. I don't like it because it means we can only test one codegen backend on any given arch, when in reality we can test a lot of it, just not execution of another arch.

If it's OK with you, I think that should be another unit of work.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
vext01: OK.

ltratt:
> I agree with your concerns about a statically compiled-in backend. I don't like it because it means we can only test one codegen backend on any given arch, when in reality we can test a lot of it, just not execution of another arch.

I really think we need to improve this in this PR.

ykrt/src/compile/jitc_yk/codegen/mod.rs
vext01: Well, it's executable code plus (maybe if compiled-in) debugging/testing information.

That's quite hard to sum up in a pithy type name, hence the rather generic name.

`CodeArtifact`? `ExectableBufferWrapper`? `CodeGenFinished`?

ykrt/src/compile/jitc_yk/codegen/mod.rs
vext01: Just to expand on this: the expensive part of making a disassembly is adding the comment lines that:

 - make it easy/possible to match only the desired part of the trace using fm tests.
 - make a trace easier for a human to read when debugging or amending tests.

E.g. a disassembly looks something like this:
```
; prologue
mov rsp, rbp
sub rsp, 64
; %0 = loadarg 0
mov ...
; %1 = load %0
mov ...
mov ...
```

Without the comment lines:
 - testing would be hard because you might match code that isn't what you want to match. comment lines serve as an anchor to an instruction of interest.
 - reading a trace is hard.

The comment lines are stored in a hashmap, and I think we don't want to be doing a load of hashing in release mode.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
vext01: Hang on, that doesn't make sense. We are not matching the innards of the `Result`, but the `size` argument of the function.

The size and offset are independent here.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
ptersilie: I dunno. I seems a bit backwards to me to affect your performance for the sake of making testing easier. But if it makes us test less because it's such a hassle then that's also bad. Maybe we could have a general AOT module that we then use for all the tests. Or we make a helper that makes it easier to construct the AOT module. Maybe some syntax where you make a trace and it then generates the corresponding AOT module automatically.

I dunno, since @ltratt was so adamant we never ever copy from the AOT if we can help it, I think he will have to make that decision.

ltratt: I didn't think about the impact of testing. Given the choice between "make testing plausible" and "squeeze every last drop of performance" I would go for "make testing plausible".

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
ltratt: Ah, oops! Although you could do `match (i32::try_from(...), size)` if you wanted, but it might not be clearer than what you currently have.

ykrt/src/compile/jitc_yk/codegen/mod.rs
ltratt: Got it.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
vext01: Yeah, I think i prefer to do it the unclever way this time :)

ykrt/src/compile/jitc_yk/codegen/mod.rs
vext01: This does raise an annoying consequence that is niggling: it means we can't reliably test release builds, and that seems backwards.

I wonder what we should do?

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
ptersilie: Some things, e.g. constant strings, constant arrays, etc, will be very expensive to copy though.

ltratt: They won't be *very* expensive :) And at the moment testability is going to save *us* more time.

vext01: OK, well that's definitely a separate PR, as there are many tentacles.

It will also cause more of the "search for this thing, or insert it" idiom, which @ltratt has expressed distaste for.

For example, types will no longer be shared between AOT and JIT modules, and there will need to be a mapping between the now separate elements in each module kind. When you have a handle on an AOT type and want to get the corresponding JIT type, can we avoid that idiom somehow?

ykrt/src/compile/jitc_yk/codegen/mod.rs
vext01: Fixed in https://github.com/ykjit/yk/pull/971/commits/4c25a723383a8c577fad6d1f339b8dd37bd0c88e

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
vext01: Fixed in https://github.com/ykjit/yk/pull/971/commits/4c25a723383a8c577fad6d1f339b8dd37bd0c88e

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
ltratt: > there will need to be a mapping between the now separate elements in each module kind

I don't get this bit.

ptersilie: So that's a no on generating the AOT/JIT module automatically from some description for testing?

ykrt/src/compile/jitc_yk/codegen/mod.rs
vext01: Fixed in https://github.com/ykjit/yk/pull/971/commits/01814fe81a9f9325407386e5693a0fdbf764aa5c

ykrt/src/compile/jitc_yk/codegen/mod.rs
ltratt: I think `CompiledTrace` is better than all of those, but I'm in danger of bikeshedding so I will resolve for now. It might become more obvious to us in the future.

vext01:
Note: I've note modularised any of the stuff that might be common between different codegen backends, e.g. the abstract stack and allocation bits. Thinking of kicking that down the line until the time comes for another backend.

ltratt:
> Thinking of kicking that down the line until the time comes for another backend.

I suggest that when it's easy to do so we should break things out: and it will often be as easy to break it out as not. But it doesn't have to be a big concern.

ykrt/src/compile/jitc_yk/codegen/mod.rs
vext01: Problem is there's already a `CompiledTrace` in this crate, very close to this code, and that could get confusing. It's `ykrt::compile::CompiledTrace`.

Maybe they should be merged at some point.

ltratt: > Maybe they should be merged at some point.

Interesting idea. Agreed, maybe. But we can worry about that a bit later.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
ptersilie: > > there will need to be a mapping between the now separate elements in each module kind
> 
> I don't get this bit.

I think he means that in order to not copy the same type multiple types you need to know if you've already copied it and then return the copy instead. To do that you probably need some kind of mapping, e.g. `Types[AOTType] = JITType`

vext01: > So that's a no on generating the AOT/JIT module automatically from some description for testing?

It sounds like it could be fiddly to be honest.

vext01: We could use a hashmap, yes, but I thought you wouldn't like it. If that's OK then it simplifies some other paces where that idiom is in use too (separate PR).

vext01:
> I suggest that when it's easy to do so we should break things out

It's easy soonish. We won't know if I got it right until we have another backend of course.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
ltratt: The other possibility is to have two modes: a testing mode where we copy everything; and a non-testing mode where we reference things. Ultimately "make it testable" is our goal. We won't get everything right, but if that's our goal, at least we'll know what to prioritise.

vext01: OK, so I think we should roll will "make it testable". I've made a note for two future PRs based on this discussion. Thanks.

vext01:
> What I meant is: as we add new functionality, let's break it out

So that means break it out right now in this PR. I'll start on that after lunch.

vext01:
Those last 3 commits do the "splitting out" and kill some unnecessary x86 guards.

If this looks good I can squash and sync the branch.

ykrt/src/compile/jitc_yk/codegen/abs_stack.rs
```diff
@@ -0,0 +1,81 @@
+//! The abstract stack.
+
+/// This data structure keeps track of an abstract stack pointer for a JIT frame during code
+/// generation. The abstract stack pointer is zero-based, so the stack pointer value also serves as
+/// the size of the stack.
+///
+/// The implementation is platform agnostic: as the stack gets bigger, the stack pointer grows
```
ltratt, line 7: s/the stack pointer/abstract stack pointer/

vext01: Fixed in https://github.com/ykjit/yk/pull/971/commits/6527226c2f1d6f3958634ca168bcc435395057d9

vext01:
OK to squash?

ltratt:
Please squash.

vext01:
splat.

vext01:
I need to update my local Rust :)

vext01:
Fixed. OK to squash?

ltratt:
Please squash.

vext01:
splat.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -92,6 +99,18 @@ macro_rules! index_16bit {
                 self.0.into()
             }
         }
+
+        // impl From<usize> for $struct {
```
ptersilie, line 103: Left over comments?

vext01: ug, yes.

vext01:
As per offline discussion, disable asan for now. See #981

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Fixed in https://github.com/ykjit/yk/pull/971/commits/3eb9bac52fff8d291f2540a76e2644b5520bbe76

vext01:
OK to squash?

vext01:
splat.

PR #972, opened 2024-02-12
Make promotion a non-global variable.

ltratt:
Although it's not obvious at first `promote.rs`'s use of a `thread_local!` is a simple global variable. That is problematic because it means that the compilation process has to run in the same thread as the tracing thread which only works if serialised compilation is turned on. If you ran a program which uses promotion without serialised compilation turned on you got a "promotion buffer underflowed" panic.

This PR makes the promotion array a non-global variable, so that serialised compilation is no longer required. This PR makes the change in the most bone-headed minimalistic manner possible: as this shows, this isn't a good fit with our current architecture and makes the structure a lot messier. I intend addressing this soon but at least this PR unbreaks the system.

### DIFF
--- a/ykrt/src/promote.rs
+++ b/ykrt/src/promote.rs
@@ -1,64 +1,6 @@
-//! Software constant recorder.
-//!
-//! This is used for promoting values to constants when compiling a trace.
+//! Trace promotion: promote values to constants when recording and compiling a trace.
 
-use std::{cell::RefCell, collections::VecDeque};
-
-// FIXME: Try to move this into ykrt in the existing thread local.
-thread_local! {
-    /// The thread's value recorder.
-    ///
-    /// This has to be specific to a thread to avoid cross-talk from other threads which may be
-    /// executing functions with promoted arguments at the same time.
-    pub static VAL_REC: RefCell<ValueRecorder> = RefCell::new(ValueRecorder::default());
-}
-
-/// A value recorder observes and records constant values during tracing. The trace compiler
-/// queries this when building a trace to replace promoted values with the observed constants.
-#[derive(Debug, Default)]
-pub struct ValueRecorder {
-    // `true` when recording promotions.
-    record_enable: bool,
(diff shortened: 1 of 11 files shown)


ykrt/src/promote.rs
```diff
@@ -1,64 +1,6 @@
...
-/// For the current thread, enable or disable constant recording.
-pub fn thread_record_enable(record: bool) {
-    VAL_REC.with_borrow_mut(|r| {
-        r.record_enable(record);
-    })
-}
-
-/// For the current thread, remove and return the oldest recorded value.
-#[no_mangle]
-pub extern "C" fn __yk_lookup_promote_usize() -> usize {
```
vext01, line 59: Do we need to delete this in a header too?

vext01:
This probably already performs better than before since we no longer need to call the `__yk_` stuff to access the promotions.

ykrt/src/promote.rs
ltratt: Good point! [Not a header but your overall point is correct.] Fixed in ba15f75c.

vext01:
Please squash.

ltratt:
FWIW, I think the performance difference will be very, probably unmeasurably, small. So this is just about getting the structure right, not performance.

ltratt:
Squashed.

vext01:
```
10:21:03   --- stderr
10:21:03   thread 'main' panicked at ykbuild/src/completion_wrapper.rs:100:25:
10:21:03   source file has no extension: '--version', buf='/ci/ykllvm/build/../inst/bin/clang --version'
10:21:03   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
10:21:03 warning: build failed, waiting for other jobs to finish...
```

any ideas?

ltratt:
Nope, I'm baffled by that one, but will keep investigating.

ltratt:
I cannot replicate this locally or with `tryci`. I don't know what to do other than shrug my shoulders and suggest trying merging again.

PR #973, opened 2024-02-13
Rename collection to recording

ltratt:
Needs #972 to be merged first.

[Warning: I will admit that "collection" as a term has never quite resonated with me, so I'm biased in favour of this commit from the off.]
 
Trace "collection", to me, implies a slightly passive view of what's going on, but promotion is a much more active view of what's going on. In particular, some tracing backends can potentially "decode" (another term we might want to revisit one day) a trace in such a way that they successively generate AOT blocks and promotions in one go. That means that we're going to have to rename the `TracedAOTBlock` enum, probably to something like `RecordedTraceThingy` where "recorded" might be an AOT block or a promoted variable (or ... who knows what else in the future!).

Note that I have chosen not to rename anything in the `hwtracer` crate. That crate is a bit separate from everything else and I'm not sure I want to rethink what it's doing at the moment.

### DIFF
--- a/tests/c/stats1.c
+++ b/tests/c/stats1.c
@@ -5,10 +5,10 @@
 //     {
 //       ...
 //       "trace_executions": 1,
-//       "traces_collected_err": 0,
-//       "traces_collected_ok": 1,
 //       "traces_compiled_err": 0,
-//       "traces_compiled_ok": 1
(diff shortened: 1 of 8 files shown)


vext01:
I don't object to "recorder".

vext01:
please.

PR #974, opened 2024-02-13
Make the compilation wrapper more resilient.

vext01:
- Don't assume the source file is the last argument.
 - Don't assume there will be a source file.

### DIFF
--- a/ykbuild/src/completion_wrapper.rs
+++ b/ykbuild/src/completion_wrapper.rs
@@ -10,6 +10,8 @@ use std::{
 };
 use tempfile::TempDir;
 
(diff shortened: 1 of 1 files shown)


ykbuild/src/completion_wrapper.rs
```diff
@@ -84,38 +86,35 @@ impl CompletionWrapper {
     /// Call when the build is done to generate the `compile_commands.json` file.
     pub fn generate(self) {
         let mut entries = Vec::new();
-
```
vext01, line 87: This wasn't intentional, but I guess you are going to like it :)

ltratt: It's the best part :p

PR #976, opened 2024-02-16
Move data recorded with `yk_promote` into `MTThread`.

ltratt:
I've slowly come to realise that calls to `yk_promote` could (but don't currently!) end up being in compiled in two different ways:

  1. A literal call to `yk_promote_*`.
  2. An inlined call to a side-band recording mechanism (e.g. `PTWRITE`).

Because I hadn't teased these two apart in my mind, my previous attempt had kind-of tried to merge the two together in one API. It's now clear that was entirely misguided on my part.

This commit can be seen as partly undoing my previous attempt: literal calls to `yk_promote_*` now record the data in `MTThread` which is a thread local. However, when tracing stops, the promotion data is now extracted from `MTThread` so it can safely be moved to another thread.

That means we can now handle #1 above: but we don't yet have an API that can handle #2.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -263,19 +263,29 @@ impl MT {
                 match Arc::clone(&tracer).start_recorder() {
                     Ok(tt) => THREAD_MTTHREAD.with(|mtt| {
                         *mtt.thread_tracer.borrow_mut() = Some(tt);
+                        *mtt.promotions.borrow_mut() = Some(Vec::new());
                     }),
                     Err(e) => todo!("{e:?}"),
                 }
             }
             TransitionControlPoint::StopTracing(hl_arc) => {
-                // Assuming no bugs elsewhere, the `unwrap` cannot fail, because `StartTracing`
+                // Assuming no bugs elsewhere, the `unwrap`s cannot fail, because `StartTracing`
                 // will have put a `Some` in the `Rc`.
-                let thrdtrcr = THREAD_MTTHREAD.with(|mtt| mtt.thread_tracer.take().unwrap());
+                let (thrdtrcr, promotions) = THREAD_MTTHREAD.with(|mtt| {
+                    (
+                        mtt.thread_tracer.take().unwrap(),
+                        mtt.promotions.take().unwrap(),
+                    )
+                });
                 match thrdtrcr.stop() {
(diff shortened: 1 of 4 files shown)


ykrt/src/mt.rs
```diff
@@ -286,12 +296,21 @@ impl MT {
             TransitionControlPoint::StopSideTracing(hl_arc, sti, parent) => {
                 // Assuming no bugs elsewhere, the `unwrap` cannot fail, because `StartTracing`
                 // will have put a `Some` in the `Rc`.
-                let thrdtrcr = THREAD_MTTHREAD.with(|mtt| mtt.thread_tracer.take().unwrap());
+                let (thrdtrcr, promotions) = THREAD_MTTHREAD.with(|mtt| {
+                    (
+                        mtt.thread_tracer.take().unwrap(),
+                        mtt.promotions.take().unwrap(),
```
vext01, line 302: we should say why the unwraps are safe.

vext01:
If I understand correctly, you move the promotions from a trace recorder (local to a thread by definition) into a thread local mtthread?

At the risk of sounding like a dummy, what is the advantage of this? Just more convenient, or something more profound?

ykrt/src/mt.rs
ltratt: The existing comment on line 296 explains.

ltratt:
> If I understand correctly, you move the promotions from a trace recorder (local to a thread by definition) into a thread local mtthread?

I originally thought that all tracing backends could work out when `yk_promote_*` was called and handle those in the same way as in-band (e.g. PTWRITE) data. It's now clear to me that this is possible for some backends (e.g. swt) but not others (hwt). Thus there has to be a generic outside-the-tracer mechanism for handling `yk_promote_*`.

ykrt/src/mt.rs
vext01: Oh right. I thought that was referring to something else because it's singular. Maybe wack and `s` on the end?

(Top-notch pedantry right here!)

vext01:
OK

ykrt/src/mt.rs
ltratt: Fixed in ef0b3e11.

ltratt:
Squashed.

PR #984, opened 2024-02-22
Semi-unbreak trace mapping stats.

ltratt:
At some point in refactoring, "trace mapping" costs had been ascribed to a single `assert` which meant that in `YKD_STATS` we always ended up with 0 seconds spent in mapping.

In reality, trace mapping is split into two parts. The first (and, at the moment, the only thing that takes any time) is `stop()` which in hwt takes a long time to run. So this commit effectively measures the costs of calling `stop`.

As and when hwt and other backends truly move trace mapping costs to an iterator, these timing probes will need to be moved again. But at least I know have a half-decent idea of what the current costs are!

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -277,8 +277,10 @@ impl MT {
                         mtt.promotions.take().unwrap(),
                     )
                 });
+                self.stats.timing_state(TimingState::TraceMapping);
                 match thrdtrcr.stop() {
                     Ok(utrace) => {
+                        self.stats.timing_state(TimingState::None);
                         #[cfg(feature = "yk_jitstate_debug")]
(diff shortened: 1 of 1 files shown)


ykrt/src/mt.rs
```diff
@@ -277,8 +277,10 @@ impl MT {
                         mtt.promotions.take().unwrap(),
                     )
                 });
+                self.stats.timing_state(TimingState::TraceMapping);
                 match thrdtrcr.stop() {
                     Ok(utrace) => {
+                        self.stats.timing_state(TimingState::None);
```
vext01, line 283: why don't we time just stop and use an intermediate let? This way we don't need multiple "turn off timing" calls.

PR #985, opened 2024-02-22
Nudge towards trace mapping iteration

ltratt:
This PR is two only semi-related changes:

1. Nudge hwt towards trace mapping iteration (https://github.com/ykjit/yk/commit/033e57f39646249e6ac80c589fe124603ee4c695).
2. That made me realise that true mapping iteration could raise an error during iteration, so make it possible to express that (https://github.com/ykjit/yk/commit/987e922dd790617638a16c61c470f6b0ceb3868b).

I'm raising this now, as https://github.com/ykjit/yk/commit/987e922dd790617638a16c61c470f6b0ceb3868b will force a (hopefully small) change in @Pavel-Durov's swt branch.

### DIFF
--- a/ykrt/src/compile/jitc_llvm.rs
+++ b/ykrt/src/compile/jitc_llvm.rs
@@ -36,7 +36,13 @@ impl Compiler for JITCLLVM {
         sti: Option<SideTraceInfo>,
         hl: Arc<Mutex<HotLocation>>,
     ) -> Result<CompiledTrace, CompilationError> {
-        let irtrace = aottrace_iter.0.collect::<Vec<_>>();
+        let mut irtrace = Vec::new();
+        for ta in aottrace_iter.0 {
+            match ta {
+                Ok(x) => irtrace.push(x),
+                Err(_) => todo!(),
+            }
+        }
         let (func_names, bbs, trace_len) = self.encode_trace(&irtrace);
 
         let llvmbc = llvmbc_section();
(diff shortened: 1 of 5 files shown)


ykrt/src/compile/jitc_llvm.rs
```diff
@@ -36,7 +36,13 @@ impl Compiler for JITCLLVM {
         sti: Option<SideTraceInfo>,
         hl: Arc<Mutex<HotLocation>>,
     ) -> Result<CompiledTrace, CompilationError> {
-        let irtrace = aottrace_iter.0.collect::<Vec<_>>();
+        let mut irtrace = Vec::new();
+        for ta in aottrace_iter.0 {
+            match ta {
+                Ok(x) => irtrace.push(x),
+                Err(_) => todo!(),
```
vext01, line 43: Why not populate the `todo!()`s? Should be a simple `.map_err()`?

ltratt: I have no way of testing these clauses, and I am not keen on adding more untested code to yk. I would like to address hwt's lack of testing, which will hopefully shake out `todo`s and ensure we actually do the right thing.

vext01: To test this you can provide virtual addresses that are not in the block map. Probably no small change though, so perhaps another PR at some point?

ltratt: Yes, we definitely need to think how we can test hwt more thoroughly (e.g. unit tests with PT data). But that's out of scope for now IMHO.

PR #988, opened 2024-02-22
Handle more bytecodes.

ptersilie:
This gets us a bit further in the small sample trace we are using.

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -379,6 +379,11 @@ pub(crate) struct Instruction {
 }
 
 impl Instruction {
+    /// Return the number of operands in this instruction.
+    pub(crate) fn operands_len(&self) -> usize {
+        self.operands.len()
+    }
+
     /// Returns the operand at the specified index. Panics if the index is out of bounds.
     pub(crate) fn operand(&self, idx: usize) -> &Operand {
         &self.operands[idx]
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -16,6 +16,7 @@ const FORMAT_VERSION: u32 = 0;
 
 /// The symbol name of the control point function (after ykllvm has transformed it).
 const CONTROL_POINT_NAME: &str = "__ykrt_control_point";
+const STACKMAP_CALL_NAME: &str = "llvm.experimental.stackmap";
 
 // Generate common methods for index types.
 macro_rules! index {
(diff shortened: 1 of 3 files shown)


ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -328,7 +328,9 @@ impl<'a> TraceBuilder<'a> {
                 }
                 None => {
                     // UnmappableBlock block
-                    todo!();
+                    // Ignore for now. May be later used to make sense of the control flow. Though
```
vext01, line 339: "may be" -> "maybe"

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -328,7 +328,9 @@ impl<'a> TraceBuilder<'a> {
                 }
                 None => {
                     // UnmappableBlock block
-                    todo!();
```
ltratt, line 295: Why have we removed this? I know we want to do one day, but we can't do so yet can we?

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -678,6 +693,20 @@ pub(crate) struct FuncType {
     is_vararg: bool,
 }
 
+impl FuncType {
+    pub fn arg_ty_idxs(&self) -> &Vec<TypeIdx> {
```
ltratt, line 697: `pub(crate)` unless we *really* have to `pub` these (which I don't think we do).

ykrt/src/compile/jitc_yk/trace_builder.rs
ltratt: I think it's correct as-is.

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: No real test case will be testable if we crash out here.

vext01:
LGTM, bar a typo.

ykrt/src/compile/jitc_yk/trace_builder.rs
ltratt: Wouldn't it be better to leave it in until/unless that does happen?

ptersilie: If we don't ignore unmappable traces, we can't continue dealing with the current sample trace we got. In the future we would have to put outlining logic here probably.

ykrt/src/compile/jitc_yk/aot_ir.rs
ptersilie: `pub(crate)` didn't work. Let me try again.

ykrt/src/compile/jitc_yk/trace_builder.rs
ltratt: OK.

ykrt/src/compile/jitc_yk/aot_ir.rs
ptersilie: I could have sworn that I was getting an error at some point, but maybe that was unrelated. Seems to be working now. Fixed in [c8ff9d9](https://github.com/ykjit/yk/pull/988/commits/c8ff9d93537605107934da4253109b6ce76beb20)

ptersilie:
Ok to squash?

ltratt:
Please squash.

ptersilie:
Squashed.

PR #990, opened 2024-02-22
Fail earlier on long PT traces

ltratt:
This is "maybe a PR on the stage to something better or maybe a PR for consideration that we then throw away before doing some better". Probably more the former than the latter but that's up to you!

The basic problem is that trace mapping in hwt can be painfully slow for long traces (I *think* mapping time is linear for trace length, but validating that hypothesis is more annoying than I want to do now). For example if we trace `db.lua` most traces take 1-3ms to map, but one takes 250ms --- and then exceeds MT's trace limit! This commit allows us to avoid mapping traces that are too long at all. So for `db.lua` on master:

```json
{
    "duration_compiling": 12.618,
    "duration_deopting": 0.011,
    "duration_jit_executing": 0.006,
    "duration_outside_yk": 0.059,
    "duration_trace_mapping": 0.436,
    "trace_executions": 168,
    "traces_compiled_err": 0,
    "traces_compiled_ok": 39,
    "traces_recorded_err": 0,
    "traces_recorded_ok": 39
}
```

and on this PR:

```json
{
    "duration_compiling": 12.747,
    "duration_deopting": 0.011,
    "duration_jit_executing": 0.006,
    "duration_outside_yk": 0.060,
    "duration_trace_mapping": 0.203,
    "trace_executions": 168,
    "traces_compiled_err": 0,
    "traces_compiled_ok": 39,
    "traces_recorded_err": 1,
    "traces_recorded_ok": 39
}
```

Note that "duration_trace_mapping" has gone from 0.436s -> 0.203s.

I think we probably want to do more here. For example, a "too long" trace could be a temporary thing, so retrying could make sense. And, perhaps, what we want to do is see if a loop is important enough that we're willing to gradually increase the PT buffer size from the fairly low limits in https://github.com/ykjit/yk/commit/13a92b3504bf4fb71c8133d5c4b709452ae53e78.


hwtracer/src/perf/mod.rs
```diff
@@ -3,9 +3,9 @@ use std::alloc::Layout;
 
 pub(crate) mod collect;
 
-const PERF_DFLT_DATA_BUFSIZE: size_t = 64; // Pages
-const PERF_DFLT_AUX_BUFSIZE: size_t = 64 * 1024 * 1024; // MiB
-const PERF_DFLT_INITIAL_TRACE_BUFSIZE: size_t = 1024 * 1024; // MiB
+const PERF_DFLT_DATA_BUFSIZE: size_t = 4; // Pages
```
vext01, line 6: I'm not certain it was correct to change the first two buffer sizes (which from memory are ring buffers unrelated the the total size of a trace). I will have to dig into the perf docs tomorrow.

ltratt: I was hoping you would tell me if this is a good idea or not!

vext01: OK, so `DATA_BUFSIZE` and `AUX_BUFSIZE` are used for low-level linux-perf interfacing. These are the buffers that the kernel will fill during trace recording, and we get a notification when they are getting full, at which point we have to copy the data out and update the tail pointer (they are conceptually ring buffers).

The AUX buffer is the one containing the PT packets. The DATA buffer contains side-band information (and we don't use or need that for our purposes).

The size of these buffers isn't related to the overall size of a trace. By making them smaller, you only increase the chance of an overflow. I'd suggest putting them back for now (unless it causes a performance problem).

(Note to self, check if we "consume" the DATA buffer, even if we don't use it, because failing to do so can still cause an overflow).

vext01: (Reading: https://github.com/intel/libipt/blob/master/doc/howto_capture.md)

ltratt: > By making them smaller, you only increase the chance of an overflow

That's exactly what I intended!

vext01: Why would you want that though?

Suppose the AUX buffer is 1MB big and we are willing to records traces of up to 64MB.

All it would take is a little system load for an underflow during the first 1MB of data.

vext01: I also notice that the docs say:

> Note that the size of both the AUX and the DATA area has to be a power of two pages. The DATA area needs one additional page to contain the perf_event_mmap_page.

The AUX buffer comment says `MiB` not `pages` so we are probably allocating a lot more than we thought.

ltratt: A PT trace that's really big is likely to produce an AOT trace far longer than we want to compile.

vext01: Right, but the length of the PT trace is nothing to do with the AUX buffer.

It's a bit like in your audio player, you can configure the buffer size, but that doesn't dictate the length of the tracks you can play in any way. If you make the buffer too short, you will get skips as the buffer underruns.

vext01: Let me investigate the bogus `MiB` unit.

vext01: - Revert the data and aux buffer sizes.
 - Add comments to all of the buffer size consts describing them. I can help with this if you are not sure.

About the MiB unit, I see what happens now. We convert the MiBs into pages, then later there is a runtime check that this is a power of two. So it's actually OK, if unexpected.

vext01: One thing that is a bit odd though is that whilst the AUX bufsize in MiB is OK for a default page size (4096), it can't be suitable for arbitrary page sizes. I wonder if we should round up to the next power of two pages (separate PR)?

ltratt: Not sure. Do I really want PERF_DFLT_DATA_BUFSIZE to be 8x bigger than PERF_DFLT_INITIAL_TRACE_BUFSIZE?

vext01: I've just reverted it to what it was. I thought that was the plan?

Neither of the `PERF_*` bits have anything to do with the length of a trace, so we shouldn't be changing them in this PR.

ltratt: Let's take a step back. What I wanted to do by reducing these values was causing "too long" traces to result in a "PT overflow". Where by "too long" I meant "much shorter than what we used to consider too long". I don't know how each of these values interacts though, so perhaps the values you're suggesting do what I want?

vext01: I understand, but reducing `PERF_DFLT_DATA_BUFSIZE` and `PERF_DFLT_AUX_BUFSIZE` isn't the way to do it. Those buffers could be very small and still rarely result in PT overflows. It depends how quickly we service a request to empty the ring buffers. Making them smaller only makes it less likely to overflow, but gives no guarantees. How likely would also be system dependent...

ltratt:
@vext01 I spent a while reading the PT docs and also our code, and I believe I spotted two problems (and one opportunity) which I've fixed in https://github.com/ykjit/yk/pull/990/commits/ff9338417c89743cd92adf067d9f604f6e7cb6d1. The comment there hopefully explains what's going on.

vext01:
OK, I've first tried to verify some of the claims from your commit message.

Let's start with:

>  So we were `mmap`ing *massive* AUX buffers --- 256GiB!

With this diff applied:
```diff
diff --git a/hwtracer/src/perf/collect.c b/hwtracer/src/perf/collect.c
index 4d008fc3..8ebcc40f 100644
--- a/hwtracer/src/perf/collect.c
+++ b/hwtracer/src/perf/collect.c
@@ -578,6 +578,8 @@ hwt_perf_init_collector(struct hwt_perf_collector_config *tr_conf,
   // Allocate the AUX buffer.
   //
   // Mapped R/W so as to have a saturating ring buffer.
+  fprintf(stderr, "aux_size: %llu\n", base_header->aux_size);
+  fprintf(stderr, "data_size: %llu\n", base_header->data_size);
   if (!cache_aux_buf)
     cache_aux_buf = mmap(NULL, base_header->aux_size, PROT_READ | PROT_WRITE,
                          MAP_SHARED, tr_ctx->perf_fd, base_header->aux_offset);
```

We see:
```
aux_size: 67108864                                         
data_size: 262144
```

The len argument to mmap(2) is measured in bytes, so the aux buf is 67108864 / 1024 / 1024 = 64MiB.

That sounds reasonable to me and matches the comment:
```
const PERF_DFLT_AUX_BUFSIZE: size_t = 64 * 1024 * 1024; // MiB
```

So one of us has made a mistake. The question is, who?

hwtracer/src/perf/collect.rs
```diff
@@ -125,7 +125,7 @@ impl PerfThreadTracer {
         // `stop_collector` needs to return a Box<Tracer> anyway, so it's no big deal.
         //
         // Note that the C code will mutate the trace's members directly.
-        let mut trace = Box::new(PerfTrace::new(tracer.config.initial_trace_bufsize)?);
```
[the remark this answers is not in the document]
vext01: I must admit, I'm struggling to find the use-site of `initial_trace_bufsize`, so you might be right that something has gone wrong there. Investigating.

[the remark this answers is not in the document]
ltratt: I think that PT requires only the DATA and AUX buffers. I don't think there's a third.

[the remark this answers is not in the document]
vext01: You have to copy the packets out of the AUX buffer into somewhere, that's the third. The AUX buffer is typically small, and will be overwritten lots of times for a single tracing session.

You could allocate a large AUX buffer and never respond to overflow events, but that'd be... unconventional.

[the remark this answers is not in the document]
ltratt: We effectively ripped that out in fdbea0cab674d7a3426cae871a94785ac4678083.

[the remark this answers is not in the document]
vext01: Oh...

First the title of the commit is wrong:

> Resizing the aux buffer causes problems, so don't. 

The aux buffer cannot be resized. It's talking about the overall "trace buffer".

I'm OK with not reallocating the trace buffer, but the above diff shouldn't pass the data buffer size. If the trace buffer size is truly static, then we should just rename `initial_trace_bufsize` to `trace_bufsize`.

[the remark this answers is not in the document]
ltratt: Maybe I'm an idiot, but I don't know what the "trace" buffer is: when I looked PT only talks about DATA and AUX.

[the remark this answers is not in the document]
vext01: It's explained in the bullet list above. Does it make sense?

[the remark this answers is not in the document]
ltratt: No.

hwtracer/src/perf/mod.rs
```diff
@@ -6,24 +6,31 @@ pub(crate) mod collect;
 const PERF_DFLT_DATA_BUFSIZE: size_t = 4096; // 16MiB
 /// The default perf aux buffer size, measured in pages. Must be a power of 2.
 const PERF_DFLT_AUX_BUFSIZE: size_t = 16; // 64KiB
+/// The size of the final memory returned, measured in bytes. There are no constraints on size or
```
vext01, line 9: On the lines above the comments should have appended "(for a 4096-byte page)"? Maybe I'm being too pedantic?

hwtracer/src/perf/collect.c
```diff
@@ -1,5 +1,9 @@
 #define _GNU_SOURCE
 
+// FIXME: This collector partially, but not fully, deals with overflow in the
+// DATA/AUX buffers. It would probably be better to either never handle iit
```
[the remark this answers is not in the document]
vext01: Aux overflow handled [here](https://github.com/ykjit/yk/blob/8dc5ccee5ae32bccf17f2cbfe0804bcc66f8f279/hwtracer/src/perf/collect.c#L220)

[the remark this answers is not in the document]
vext01: Also up to you if you want to fix "iit". I know we are not worrying too much about typos in comments RN.

hwtracer/src/perf/mod.rs
ltratt: This stuff is sufficiently broken that I don't think the extra pedantry really helps TBH.

hwtracer/src/perf/collect.c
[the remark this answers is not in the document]
ltratt: Fixed in 282189b0.

vext01:
Thanks for your patience. Please squash.

ltratt:
Squashed.

vext01:
Just needs a cfmt.

ltratt:
I'm an idiot. Fixed in 9e252d74.

ltratt:
Squashed.

PR #993, opened 2024-02-26
Handle add instructions and integer constants.

ptersilie:
Depends on https://github.com/ykjit/ykllvm/pull/117.

There's possibly some discussion here. In LLVM each binary operation has their own class, but they are all part of the superclass `BinaryOperator`. In this PR we have a single `BinaryOperation` instruction in the AOT IR, but for the JIT IR we currently separate them out into different instructions, e.g. `AddInstruction`, `SubInstruction`, etc. There's an argument to be made to have a single `BinaryOp` instruction in the JIT IR too. We do have enough space to fit two operands and the operator into 64 bits. Whatever we decide it can be changed either here or in a future PR.

PR #996, opened 2024-02-27
Handle icmp and conditional branches.

ptersilie:
Ignore the first commit, which is from https://github.com/ykjit/yk/pull/993. This PR needs rebasing once that PR is in.

### DIFF
--- a/ykrt/src/compile/jitc_yk/jit_ir.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir.rs
@@ -327,6 +327,8 @@ pub enum Instruction {
     Store(StoreInstruction),
     StoreGlobal(StoreGlobalInstruction),
     Add(AddInstruction),
+    Icmp(IcmpInstruction),
+    Guard(GuardInstruction),
 }
 
 impl Instruction {
@@ -345,6 +347,8 @@ impl Instruction {
             Self::Store(..) => false,
             Self::StoreGlobal(..) => false,
             Self::Add(..) => true,
+            Self::Icmp(..) => true,
+            Self::Guard(..) => false,
         }
     }
 
@@ -363,6 +367,8 @@ impl Instruction {
             Self::Store(..) => None,
             Self::StoreGlobal(..) => None,
             Self::Add(..) => todo!(),
+            Self::Icmp(..) => todo!(),
+            Self::Guard(..) => None,
         }
     }
 
@@ -397,6 +403,8 @@ impl fmt::Display for Instruction {
             Self::Store(i) => write!(f, "{}", i),
             Self::StoreGlobal(i) => write!(f, "{}", i),
             Self::Add(i) => write!(f, "{}", i),
+            Self::Icmp(i) => write!(f, "{}", i),
+            Self::Guard(i) => write!(f, "{}", i),
         }
     }
 }
@@ -420,6 +428,8 @@ instr!(Call, CallInstruction);
 instr!(PtrAdd, PtrAddInstruction);
 // FIXME: Use a macro for all binary operations?
 instr!(Add, AddInstruction);
+instr!(Icmp, IcmpInstruction);
+instr!(Guard, GuardInstruction);
 
 /// The operands for a [Instruction::Load]
 ///
@@ -715,6 +725,76 @@ impl AddInstruction {
 impl fmt::Display for AddInstruction {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "Add {}, {}", self.op1(), self.op2())
+    }
+}
+
+/// The operand for a [Instruction::Icmp]
+///
+/// # Semantics
+///
+/// Compares two operands.
+///
+#[derive(Debug)]
+pub struct IcmpInstruction {
+    op1: PackedOperand,
+    op2: PackedOperand,
+}
+
+impl IcmpInstruction {
+    pub(crate) fn new(op1: Operand, op2: Operand) -> Self {
+        Self {
+            op1: PackedOperand::new(&op1),
+            op2: PackedOperand::new(&op2),
+        }
+    }
+
+    fn op1(&self) -> Operand {
+        self.op1.get()
+    }
+
+    fn op2(&self) -> Operand {
+        self.op2.get()
+    }
+}
+
+impl fmt::Display for IcmpInstruction {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Icmp {}, {}", self.op1(), self.op2())
+    }
+}
+
+/// The operand for a [Instruction::Guard]
+///
+/// # Semantics
+///
+/// Guards a trace against diverging execution. The remainder of the trace will be compiled under
+/// the assumption that (at runtime) the guard condition is true. If the guard condition is false,
+/// then execution may not continue, and deoptimisation must occur.
+///
+#[derive(Debug)]
+pub struct GuardInstruction {
+    /// The condition to guard against.
+    cond: PackedOperand,
+    /// The expected outcome of the condition.
+    expect: bool,
+}
+
+impl GuardInstruction {
+    pub(crate) fn new(cond: Operand, expect: bool) -> Self {
+        GuardInstruction {
+            cond: PackedOperand::new(&cond),
+            expect,
+        }
+    }
+
+    fn cond(&self) -> Operand {
+        self.cond.get()
+    }
+}
+
+impl fmt::Display for GuardInstruction {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Guard {}, {}", self.cond(), self.expect)
     }
 }
 
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -398,11 +398,6 @@ pub(crate) struct Instruction {
 }
 
 impl Instruction {
-    /// Return the number of operands in this instruction.
-    pub(crate) fn operands_len(&self) -> usize {
-        self.operands.len()
-    }
-
     /// Returns the operand at the specified index. Panics if the index is out of bounds.
     pub(crate) fn operand(&self, idx: usize) -> &Operand {
         &self.operands[idx]
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -229,7 +229,7 @@ impl BlockID {
 
 #[deku_derive(DekuRead)]
 #[derive(Debug, Hash, Eq, PartialEq)]
-pub(crate) struct LocalVariableOperand(InstructionID);
+pub(crate) struct LocalVariableOperand(pub(crate) InstructionID);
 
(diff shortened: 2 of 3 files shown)


ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -717,6 +728,74 @@ impl fmt::Display for AddInstruction {
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Icmp {}, {}", self.op1(), self.op2())
+    }
+}
+
+/// The operand for a [Instruction::Guard]
+///
+/// # Semantics
+///
+/// Guards a trace against diverging execution.
```
vext01, line 770: I'd add some guard blurb like:

> The remainder of the trace will be compiled under the assumption that (at runtime) the guard condition is true. If the guard condition is false, then execution may not continue, and deoptimisation must occur.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -717,6 +728,74 @@ impl fmt::Display for AddInstruction {
...
+pub struct GuardInstruction {
+    /// The condition to guard against.
+    cond: PackedOperand,
+    /// The expected outcome of the condition.
+    expect: bool,
+}
+
+impl GuardInstruction {
+    pub(crate) fn new(cond: Operand, expect: bool) -> Self {
+        GuardInstruction {
```
vext01, line 782: debug assert that the condition is a bool?

vext01:
LGTM. A couple of questions.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -717,6 +728,74 @@ impl fmt::Display for AddInstruction {
     }
 }
 
+/// The operand for a [Instruction::Icmp]
+///
+/// # Semantics
+///
+/// Compares two operands.
```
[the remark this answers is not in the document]
ptersilie: Not sure actually. I think it could also be floating points: https://llvm.org/doxygen/classllvm_1_1CmpInst.html#a7bd850492d35a34f0fe419e5555997a9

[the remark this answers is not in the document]
vext01: That'd be a [`fcmp`](https://llvm.org/docs/LangRef.html#fcmp-instruction) in LLVM.

[`icmp`](https://llvm.org/docs/LangRef.html#icmp-instruction) is for integers.

I assume we will copy this idiom?

ykrt/src/compile/jitc_yk/jit_ir.rs
ptersilie: Does LLVM guarantee that? `BranchInst.getCondition()` returns `Value`.

vext01: It's signature is `Value *getCondition()`, so I guess so.

ykrt/src/compile/jitc_yk/jit_ir.rs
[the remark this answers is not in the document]
ptersilie: Ah, I see. `icmp` can still take a vec of ints though.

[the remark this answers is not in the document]
vext01: Yeah, it's not uncommon for llvm instructions to have vectored modes.

Really we should be asserting that we don't see them when lowering our IR.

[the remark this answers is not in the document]
ptersilie: Why would be assert this? Are we not expecting to seen them ever?

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Well a "bool" is an `i1` in LLVM IR (and in ours too, right)?

Maybe I meant to say: check the condition is a 1-bit integer?

ykrt/src/compile/jitc_yk/jit_ir.rs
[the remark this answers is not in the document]
vext01: There's no reason they couldn't appear in the AOT LLVM IR, right? I bet there are ways to get clang to generate that kind of code.

ykrt/src/compile/jitc_yk/jit_ir.rs
ptersilie: Yes, but I what I meant is that I haven't seen LLVM enforcing this. Since `BranchInst` just takes a `Value` that could also be a `i8, i32, i64` etc.

ptersilie: Ah never mind. It's defined in the langref: https://llvm.org/docs/LangRef.html#br-instruction

ptersilie: Unfortunately, we don't store the type of the condition (why waste precious bytes when we know icmp returns `i1`) and we know that the AOT module must be verified so we know that condition must be `i1`.

vext01: There are two LLVM instructions that might influence our decision:

 - icmp
 - br

They both deal with `i1`:

# [conditional `br`](https://llvm.org/docs/LangRef.html#br-instruction)

```
br i1 <cond>, label <iftrue>, label <iffalse>
```

> The conditional branch form of the ‘br’ instruction takes a single ‘i1’

# [`icmp`](https://llvm.org/docs/LangRef.html#icmp-instruction)

```
<result> = icmp <cond> <ty> <op1>, <op2>   ; yields i1 or <N x i1>:result
```

It probably takes a generic `Value` so that you can branch on locals or globals etc.

(We are talking about the *type* of a Value)

vext01: I understand, but a) the codegen tests construct IR that didn't come from LLVM, and b) aren't assertions for testing our assumptions after all?

I won't die on this hill, but seems worthwhile to me. It's up to you.

ykrt/src/compile/jitc_yk/jit_ir.rs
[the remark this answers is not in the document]
ptersilie: Yes, so why would be assert this then? We should deal with those cases in the codegen? It should probably be a todo in the codegen, no?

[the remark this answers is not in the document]
vext01: Because I doubt we lower them properly currently. Let's consider this a separate issue.

What I'm interested in here is, is it valid in our IR to compare an i32 with a u64, for example?

This is invalid in LLVM IR:

> They must also be identical types.

So I suppose it's easiest to inherit that for our IR too? In which case I think we should assert the operand types are the same.

ykrt/src/compile/jitc_yk/jit_ir.rs
ptersilie: I started to implement this and realised that a) it's a bit of a pain to do this in the trace builder and b) this isn't the right place to do this. Since we can give the codegen a hand written trace, it should be the codegen to assert this, which would also cover mistakes hand written AOT IR. Though at some point it might be useful to have an AOT IR verifier as well.

vext01: OK

ykrt/src/compile/jitc_yk/jit_ir.rs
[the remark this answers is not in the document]
vext01: So I guess we would assert this in the codegen too?

ykrt/src/compile/jitc_yk/jit_ir.rs
ptersilie: Fixed in [47682cf](https://github.com/ykjit/yk/pull/996/commits/47682cfe2dc8f9266fa615c6291ac265f348cf8c)

ykrt/src/compile/jitc_yk/jit_ir.rs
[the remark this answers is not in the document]
ptersilie: I think in this case the argument from the other comment applies here too: asserting this in the codegen means we cover both handwritten AOT IR and handwritten traces (while asserting this in the trace builder only covers AOT IR).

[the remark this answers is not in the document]
vext01: ok, consider this resolved.

vext01:
Please squash and sync

ptersilie:
Squashed and rebased.

ptersilie:
Ok to squash?

vext01:
```
error: method `operands_len` is never used
```

vext01:
ah you fixed it. yes please.

ptersilie:
Squashed.

vext01:
```
20:45:58 Pattern (error at line 10):
20:45:58    ...
20:45:58    |...
20:45:58    |--- End aot ---
20:45:58    |...
20:45:58 >> |not yet implemented...
20:45:58    |...
20:45:58 
20:45:58 Text (error at line 190):
20:45:58    ...
20:45:58    |note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
20:45:58    |thread '<unnamed>' panicked at ykrt/src/mt.rs:561:51:
20:45:58    |called `Result::unwrap()` on an `Err` value: Any { .. }
```

ptersilie:
Hmm, I think in order to fix this I will have to merge https://github.com/ykjit/yk/pull/998 into this PR.

ltratt:
#998 says it depends on this PR being merged so have we got a circular dependency?!

ptersilie:
> #998 says it depends on this PR being merged so have we got a circular dependency?!

No. It's just that this PR has an error that #998 fixes. We could try to merge #998 first and hope it doesn't give me any merge conflicts. Thought it would probably be less work to just cherry-pick the one commit from #998 into this PR.

ltratt:
Cherry picking is perfectly acceptable as it doesn't require force pushing.

ptersilie:
> Cherry picking is perfectly acceptable as it doesn't require force pushing.

But it means we'll have to close #998 as that is then no longer needed.

ptersilie:
Cherry-picked [e6db72b](https://github.com/ykjit/yk/pull/996/commits/e6db72b24ce5be5bba9bfd226c0dc46a22855eff) from #998 to fix the error.

ptersilie:
I believe this is ready now.

ptersilie:
I think these commits are fine as they are. I assume we don't really care too much about some commits erroring out since we won't be needing to git bisect past this point in the future.

ptersilie:
While I agree with you in general, this stuff is so highly work in progress that I really doubt we'll ever need to bisect past this point. We don't even have a single working test yet.

But I can try and reorder them.

ltratt:
Please force push a reordering. We don't have to kill ourselves to do so in this case, but it is a service to our future selves to make sure that every commit passes `cargo test`.

ptersilie:
It's giving me loads of conflicts. So this might take a while. I still believe it's not the worth the hassle, but it seems we disagree on that.

ltratt:
OK let's not bother this time, but in the future let's make "`cargo test` works on every commit our goal".

Issue #1002, opened 2024-02-29
hwtracer's `iter_blocks` can free memory still in use


ltratt:
`iter_blocks` returns an `Iterator` which uses a `*mut u8`, but the lifetime of the `*mut u8` is not tied to the `Iterator`. It is possible (I found this out the hard way!) for the underlying block to be freed while the PT parser is still operating. Due to the way `iter_blocks` is called on the master branch this might not cause problems there, but it may be a/the cause of our occasional mystery segfaults.

PR #1003, opened 2024-03-01
Fix the lifetime of a "trace".

ltratt:
This one took me a while to hunt down and even longer to fix. In essence, `PerfTrace` allocates a block of memory M and hands it to `YKPTBlockIterator`, but `PerfTrace` frees M itself. In a fully synchronous program this happens to work, but as soon as the iterator is passed off to some other bit of code (e.g. another thread), the trace can be freed while the PT parser is still active.

The fundamental problem is that the trace is represented by a `*mut u8` which disguises the implicit lifetime in this structure. One fix would be to change `iter_blocks` to:

```rust
fn iter_blocks<'a>(&'a self) -> Box<dyn Iterator<Item = Result<Block, HWTracerError>> + 'a Send>;
```

but that makes using `iter_blocks` rather hard (since `PerfTrace` has to be kept alive too), and is also not really the API we want: we never want to iterate over a trace's blocks more than once.

This commit thus changes `iter_blocks` to:

```rust
fn iter_blocks(self: Box<Self>) -> Box<dyn Iterator<Item = Result<Block, HWTracerError>> + Send>;
```

Ideally we might prefer `self: Self` (i.e. without `Box<Self>`) but then this method isn't object safe, so it's (almost?) impossible to use.

`PerfTrace` then has to hand responsibility for `free`ing the block to `YKPTBlockIterator`. Theoretically there is still a tiny possibility of this not being correct (see one of the `FIXME`s) but addressing that theoretical possibility would probably require a huge restructure that I don't want to undertake now. This commit, at the very least, makes it much more difficult for us to hit segfaults.

Fixes https://github.com/ykjit/yk/issues/1002.

### DIFF
--- a/hwtracer/src/perf/collect.rs
+++ b/hwtracer/src/perf/collect.rs
@@ -10,7 +10,7 @@ use crate::{
     Block, ThreadTracer, Trace, Tracer,
 };
 use libc::{c_void, free, geteuid, malloc, size_t};
-use std::{fs::read_to_string, slice, sync::Arc};
+use std::{fs::read_to_string, sync::Arc};
 
 #[cfg(pt)]
 extern "C" {
@@ -137,16 +137,18 @@ impl PerfThreadTracer {
 
 /// A wrapper around a manually malloc/free'd buffer for holding an Intel PT trace. We've split
 /// this out from PerfTrace so that we can mark just this raw pointer as `unsafe Send`.
+///
+/// FIXME: Marking a pointer `Send` is necessary but not sufficient: we also need some sort of
+/// memory barrier to guarantee that all writes from thread A are seen by thread B. That said, x86
+/// (and currently this code can only sensibly run on x86) has a strong memory model makes it
+/// relatively unlikely that this will be a problem in practise. There is, though, a tiny chance
+/// that a super-smart compiler could try and optimise things across threads.
 #[repr(C)]
 #[derive(Debug)]
-struct PerfTraceBuf(*mut u8);
+#[cfg(target_arch = "x86_64")]
+pub(crate) struct PerfTraceBuf(pub(crate) *mut u8);
 
-/// We need to be able to transfer `PerfTraceBuf`s between threads to allow background
-/// compilation. However, `PerfTraceBuf` wraps a raw pointer, which is not `Send`, so nor is
-/// `PerfTraceBuf`. As long as we take great care to never: a) give out copies of the pointer to
-/// the wider world, or b) dereference the pointer when we shouldn't, then we can manually (and
-/// unsafely) mark the struct as being Send.
-unsafe impl Send for PerfTrace {}
+unsafe impl Send for PerfTraceBuf {}
 
 /// An Intel PT trace, obtained via Linux perf.
 #[repr(C)]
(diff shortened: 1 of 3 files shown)


vext01:
> This commit thus changes iter_blocks to...

OK, so to clarify, we make iterating consume the perf data, thus decoupling the lifetimes of `PerfTrace` and its iterator?

> Ideally we might prefer self: Self (i.e. without Box<Self>) but then this method isn't object safe

Does the `Self: Sized` trick work here? Sometimes you get away with it.

ltratt:
> OK, so to clarify, we make iterating consume the perf data, thus decoupling the lifetimes of PerfTrace and its iterator?

Kind of. Just consuming `PerfTrace` isn't enough, because `PerfTrace`'s `drop` will still cause `free` (and we need that to happen in case we don't call `iter_blocks`). But when `iter_blocks` runs, it "fishes" out the real trace, replaces it with `null` and moves ownership of the real thing to `YKPTBlockIterator`.

>>    Ideally we might prefer self: Self (i.e. without Box) but then this method isn't object safe
>Does the Self: Sized trick work here? Sometimes you get away with it.

Er, did I try that? Dunno. Not sure it's worth worrying about now either TBH!

hwtracer/src/perf/collect.rs
```diff
@@ -137,16 +137,17 @@ impl PerfThreadTracer {
 
 /// A wrapper around a manually malloc/free'd buffer for holding an Intel PT trace. We've split
 /// this out from PerfTrace so that we can mark just this raw pointer as `unsafe Send`.
+///
+/// FIXME: Marking a pointer `Send` is necessary but not sufficient: we also need some sort of
+/// memory barrier to guarantee that all writes from thread A are seen by thread B. That said, x86
+/// (remember that this code can only run on x86) has a strong memory model makes it relatively
```
vext01, line 143: In theory this part of the code isn't platform specific. We don't have to worry about it now, because we don't support other arches, but I think we'd probably still [use perf to collect (e.g.) coresight traces](https://docs.kernel.org/trace/coresight/coresight-perf.html).

It's been a while since I've thought about concurrent memory model stuff. Is the problematic read/write the pointer value itself, or what it points to?

ltratt: What it points to.

FWIW, `hwtracer` probably needs to be rewritten from scratch before we can port it. We've learnt so much since we originally wrote this, and the internal structure has now been pushed and pulled so much that I find working with it very difficult. We can probably let it retire with grace at some point and make a simpler, safer, alternative. But not yet :)

vext01: Agreed.

Can we just change:

> remember that this code can only run on x86

To something like:

> currently the system only supports x86, so...

And then hopefully someone will come along re-write hwtracer in rust and we can all retire :)

vext01: > What it points to.

Maybe it isn't possible, but could we not use a memory clobber somewhere?

ltratt: It's probably possible, but I wasn't even quite sure what the right place to do this is. I think working that out might be in danger of throwing good money after bad.

vext01: ok

vext01: (so just the comment re-word then?)

ltratt: Fixed in bf6f6585. Also added a `cfg` guard for x86 to enforce the comment!

ltratt: bf6f6585 has now shown up

vext01:
Please squash.

ltratt:
Squashed.

PR #1004, opened 2024-03-01
Fix hidden panics

vext01:
See commit messages.

By the way @ltratt, after this fix, I still see stats in the `--nocapture` output of passing tests where we have no requested stats. e.g.:

```
$ cargo test dont_trace_execution_of_a_trace -- --nocapture
...
running 1 test
{
    "duration_compiling": 0.000,
    "duration_deopting": 0.000,
    "duration_jit_executing": 0.000,
    "duration_outside_yk": 0.000,
    "duration_trace_mapping": 0.000,
    "trace_executions": 0,
    "traces_compiled_err": 0,
    "traces_compiled_ok": 0,
    "traces_recorded_err": 0,
    "traces_recorded_ok": 0
}
test mt::tests::dont_trace_execution_of_a_trace ... ok
```

A fix for that should be a separate PR, but any ideas?

### DIFF
--- a/ykrt/src/frame/mod.rs
+++ b/ykrt/src/frame/mod.rs
@@ -3,7 +3,9 @@
 
 use llvm_sys::{core::*, prelude::LLVMValueRef};
 use object::{Object, ObjectSection};
-use std::{ffi::c_void, ptr, sync::LazyLock, thread};
+#[cfg(not(test))]
+use std::thread;
+use std::{ffi::c_void, ptr, sync::LazyLock};
 use ykaddr::obj::SELF_BIN_MMAP;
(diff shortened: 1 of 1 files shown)


ltratt:
@vext01 look at `ykstats.rs`:

```rust
impl YkStats {
    #[cfg(not(test))]
    pub fn new() -> Self {
        if let Ok(p) = env::var("YKD_STATS") {
            Self {
                inner: Some(Mutex::new(YkStatsInner::new(p))),
            }
        } else {
            Self { inner: None }
        }
    }

    #[cfg(test)]
    pub fn new() -> Self {
        Self {
            inner: Some(Mutex::new(YkStatsInner::new("-".to_string()))),
        }
    }
```

PR #1006, opened 2024-03-04
Quality of life improvements

vext01:
The first commit is already in a PR pending merge. Please ignore that one.

Remaining commits are small items. Please see commit messages.

Curious to see what you think of https://github.com/ykjit/yk/commit/8474f0017598c0f295df7440981af7937dd046f1

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64.rs
@@ -118,7 +118,11 @@ impl<'a> X64CodeGen<'a> {
             jit_ir::Instruction::Load(i) => self.codegen_load_instr(instr_idx, &i),
             jit_ir::Instruction::PtrAdd(i) => self.codegen_ptradd_instr(instr_idx, &i),
             jit_ir::Instruction::Store(i) => self.codegen_store_instr(&i),
-            _ => todo!(),
+            jit_ir::Instruction::LoadGlobal(_) => todo!(),
+            jit_ir::Instruction::StoreGlobal(_) => todo!(),
+            jit_ir::Instruction::Call(_) => todo!(),
+            jit_ir::Instruction::Icmp(_) => todo!(),
+            jit_ir::Instruction::Guard(_) => todo!(),
         }
     }
 
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64.rs
@@ -412,10 +412,7 @@ impl<'a> AsmPrinter<'a> {
 mod tests {
     use super::{CodeGen, X64CodeGen, STACK_DIRECTION};
     use crate::compile::jitc_yk::{
-        codegen::{
-            reg_alloc::{RegisterAllocator, SpillAllocator},
-            tests::match_asm,
-        },
+        codegen::{reg_alloc::RegisterAllocator, tests::match_asm},
         jit_ir::{self, IntegerType},
     };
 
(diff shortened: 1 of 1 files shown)


ltratt:
Can you wait until the other PR merges and force push an update please?

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -295,7 +293,8 @@ impl<'a> TraceBuilder<'a> {
         };
         let expect = *succ_bb == nextbb.block_idx();
         let guard = jit_ir::GuardInstruction::new(jit_ir::Operand::Local(cond), expect);
-        Ok(self.jit_mod.push(guard.into()))
+        let _ = self.jit_mod.push(guard.into());
```
ptersilie, line 296: This is slightly odd on instructions that don't define a variable like here. Though I don't currently have a suggestion on how to make this better.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
```diff
@@ -118,7 +118,11 @@ impl<'a> X64CodeGen<'a> {
             jit_ir::Instruction::Load(i) => self.codegen_load_instr(instr_idx, &i),
             jit_ir::Instruction::PtrAdd(i) => self.codegen_ptradd_instr(instr_idx, &i),
             jit_ir::Instruction::Store(i) => self.codegen_store_instr(&i),
-            _ => todo!(),
```
ltratt, line 120: It's normally easier to do `x => todo!("{x:?}")` IMHO but it's not a big deal.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -165,6 +165,14 @@ impl InstrIdx {
     pub(crate) fn instr<'a>(&'a self, jit_mod: &'a Module) -> &Instruction {
         jit_mod.instr(*self)
     }
+
+    /// Make a local variable [Operand] from a local variable index.
+    pub(crate) fn to_operand(&self, m: &Module) -> Operand {
+        // If the instruction doesn't define a local variable, then it doesn't make sense to make
+        // an operand of it.
+        debug_assert!(self.instr(m).is_def());
```
ltratt, line 173: IMHO this should be `#Panics` or "undefined behaviour" doc string too, as if you call this function incorrectly you will get a surprise I think?

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
vext01: I just wanted to see at a glance what I need to work on next.

ykrt/src/compile/jitc_yk/trace_builder.rs
ltratt: Presumably a `push_no_rtn` or similar function that panics if it doesn't define a variable?

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
ltratt: Yes, the idiom I suggested does just that, but with a lot less typing :)

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: Indeed. I don't have a suggestion either. But there are some upsides that may pay off this oddity?

ltratt: FWIW, there are no upsides to this, and it looks to me like it swallows errors, so there are some downsides? Maybe I've misunderstood.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: It won't cause UB, it just means that you can pass an operand of void type, which doesn't make much sense. When would you ever need to do that?

I considered always checking and panicking, but `self.instr(m).is_def()` comes at a small runtime cost.

ltratt: My point is more: I think functions must document (even briefly) their preconditions and (when possible) error in some sensible way if those are violated. Otherwise what happens is our future selves forget the preconditions, call the function in the wrong way, and then spend days debugging because they got the wrong end of the stick.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
vext01: Your suggestion assumes that the codegen is linked to the pipeline though, which it is not. I'd have to write a test to cover the next case, which would require me to eyeball which case isn't yet covered :P

ltratt: I'm about 99.9% sure this isn't correct. So long as `x` implements `Debug`, the idiom I suggested does everything you suggest, but doesn't force you to manually expand it. I use this idiom all the time for exactly the reasons you're suggesting :)

Anyway, it's not important, and not worth changing for this PR, but it might save you time in the future!

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: No upsides! OK, my commit message failed to explain then.

Consider this hunk:
```diff
@@ -651,11 +643,15 @@ mod tests {
         let i64_ty_idx = jit_mod
             .type_idx(&jit_ir::Type::Integer(IntegerType::new(64)))
             .unwrap();
-        jit_mod.push(jit_ir::LoadTraceInputInstruction::new(0, i64_ty_idx).into());
-        jit_mod.push(jit_ir::LoadTraceInputInstruction::new(64, i64_ty_idx).into());
-        let op1 = jit_ir::Operand::Local(jit_ir::InstrIdx::new(0).unwrap());
-        let op2 = jit_ir::Operand::Local(jit_ir::InstrIdx::new(1).unwrap());
-        jit_mod.push(jit_ir::AddInstruction::new(op1, op2).into());
+        let ti_op1 = jit_mod
+            .push(jit_ir::LoadTraceInputInstruction::new(0, i64_ty_idx).into())
+            .unwrap()
+            .to_operand(&jit_mod);
+        let ti_op2 = jit_mod
+            .push(jit_ir::LoadTraceInputInstruction::new(64, i64_ty_idx).into())
+            .unwrap()
+            .to_operand(&jit_mod);
+        let _ = jit_mod.push(jit_ir::AddInstruction::new(ti_op1, ti_op2).into());
         let patt_lines = [
             "...",
             "; Add %0, %1",
```

Before we push the instructions, then manually magic up the operands with the correct instruction indices that correspond with the pushed instructions. That is a) laborious, b) error prone.

After, the pushes themselves give us everything we need to make the operands automatically.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
vext01: What I'm saying is, because the codegen isn't linked in the pipeline, no (non-unit) test can trigger your todo!().

ltratt: If so, it can't trigger your todos either. Anyway, let's move on.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: I'm not sure it's a precondition. Maybe we should just delete the assertion.

@ptersilie what do you think?

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: It kind of does. It errors if the instruction index that would result wasn't expressible in the bits we have.

That's only really a problem if a) the instruction defined a value (it may not), b) the value gets used (in which case the error would either be consumed, or the user would magic up another `InstrIdx` with the same value and consume the same error that would have otherwise been generated).

I'm not sure there's a better way, but I'm open to suggestions.

vext01: I'm getting cold feet about this change now.

Maybe we should back that commit out, and you just have to be careful with what `InstrIdx`s you magic up? It only really affects unit tests anyway. The trace builder doesn't work in a way that can take advantage of this idiom anyway.

vext01: Another idiom we could use in tests is something like:
```
self.push(some_instruction);
let instr_idx = self.previous_instr_idx();
```

of course this is more error prone too...

ykrt/src/compile/jitc_yk/jit_ir.rs
ptersilie: Hm, I think `debug_assert` is fine here. You can make an operand from a non-def instruction, it just doesn't make sense. Do we ever need the instridx of a non-def instruction, or can we just make sure we never get the instruction index of a non-def instruction?

ltratt: If we don't know we need to do something, I'd rather error if someone does it. If they then work out that they can generalise the function, they'll be able to do so knowing the full context.

ykrt/src/compile/jitc_yk/trace_builder.rs
ltratt: If we only need the index in tests, add a second function with `#[cfg(test)]` or similar perhaps?

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: How would we prefer to do this?
 - change `debug_assert` to `assert` and take some runtime hit.
 - keep it `debug_assert` and document that it's UB to make a void operand (a slight untruth).

ltratt: I've said my peace so please do whatever you think best (inc "keep as-is").

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: There's currently one non-test call site where the function could be useful.

I think it'd be confusing to have a different API available for testing so, if it's OK with you I will just revert this change.

ltratt: OK.

vext01: Reverted in 6409770a974abce2a137b21c190a99069f73c8ce

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Since I've decided to revert this commit, this function too disappears, and this conversation is resolved.

Revert in 6409770a974abce2a137b21c190a99069f73c8ce

vext01:
OK to squash and rebase?

ltratt:
Please squash.

vext01:
squashed.

PR #1008, opened 2024-03-04
Make trace mapping a true iterator

ltratt:
This PR turns hwt mapping into a true iterator. The most obvious benefit is that this allows mapping to happen in the compilation thread rather than pausing the interpreter (@Pavel-Durov you may want to make a similar change to swt). It will also allow us, if we later want to, to put trace mapping into a separate thread altogether.

The main commit is https://github.com/ykjit/yk/commit/934f3db4bfce6d968d20712e5002241f30175354; https://github.com/ykjit/yk/commit/e14cf139859a1446ff48b58ccb06119a1f23ca9e is a minor optimisation and simplification that has a bit more churn than I anticipated but still seems worthwhile.

### DIFF
--- a/tests/c/trace_too_long.c
+++ b/tests/c/trace_too_long.c
@@ -1,0 +1,46 @@
+// # This tests that traces that generate too many blocks cause "trace too
+// # long" warnings. This can be very slow (e.g. on swt), so ignore it except
+// # where we know it'll run # fast enough.
+// ignore-if: test "$YKB_TRACER" != "hwt"
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_PRINT_JITSTATE=1
+//   stderr:
+//     ...
+//     jit-state: stop-tracing
+//     jit-state: trace-compilation-aborted: Trace too long
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+int main(int argc, char **argv) {
+  YkMT *mt = yk_mt_new(NULL);
+  yk_mt_hot_threshold_set(mt, 0);
+  YkLocation loc1 = yk_location_new();
+  YkLocation loc2 = yk_location_new();
+
+  int i = 2000;
+  NOOPT_VAL(loc1);
+  NOOPT_VAL(loc2);
+  NOOPT_VAL(i);
+  YkLocation *loc = &loc1;
+  while (i > 0) {
+    yk_mt_control_point(mt, loc);
+    if (i == 2000)
+      loc = &loc2;
+    else if (i == 2)
+      loc = &loc1;
+    fprintf(stdout, "i=%d\n", i);
+    i--;
+  }
+  printf("exit");
+  NOOPT_VAL(i);
+  yk_location_drop(loc1);
+  yk_location_drop(loc2);
+  yk_mt_drop(mt);
+  return (EXIT_SUCCESS);
+}
--- a/tests/c/trace_too_long_hwt.c
+++ b/tests/c/trace_too_long_hwt.c
@@ -1,0 +1,45 @@
+// # This test checks that an overflowing PT buffer is caught at the point
+// # where a trace is stopped, not after trace mapping. It therefore only works
+// # on hwt.
+// ignore-if: test "$YKB_TRACER" != hwt
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_PRINT_JITSTATE=1
+//   stderr:
+//     ...
+//     jit-state: stop-tracing-aborted: Trace too long
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+int main(int argc, char **argv) {
+  YkMT *mt = yk_mt_new(NULL);
+  yk_mt_hot_threshold_set(mt, 0);
+  YkLocation loc1 = yk_location_new();
+  YkLocation loc2 = yk_location_new();
+
+  int i = 100000;
+  NOOPT_VAL(loc1);
+  NOOPT_VAL(loc2);
+  NOOPT_VAL(i);
+  YkLocation *loc = &loc1;
+  while (i > 0) {
+    yk_mt_control_point(mt, loc);
+    if (i == 100000)
+      loc = &loc2;
+    else if (i == 2)
+      loc = &loc1;
+    fprintf(stdout, "i=%d\n", i);
+    i--;
+  }
+  printf("exit");
+  NOOPT_VAL(i);
+  yk_location_drop(loc1);
+  yk_location_drop(loc2);
+  yk_mt_drop(mt);
+  return (EXIT_SUCCESS);
+}
--- a/ykrt/src/trace/hwt/mapper.rs
+++ b/ykrt/src/trace/hwt/mapper.rs
@@ -36,38 +36,7 @@ impl Iterator for HWTTraceIterator {
         while self.upcoming.len() < 2 {
             match self.hwt_iter.next() {
                 Some(Ok(x)) => {
-                    let mapped = map_block(&x);
-                    if mapped.is_empty() {
-                        // The block is unmappable. Note that we collapse consecutive unmappable
-                        // blocks.
-                        if self.upcoming.last() != Some(&TraceAction::new_unmappable_block()) {
-                            self.upcoming.push(TraceAction::new_unmappable_block());
-                            self.tas_generated += 1;
-                        }
-                    } else {
-                        for b in map_block(&x) {
-                            match b {
-                                Some(b) => {
-                                    if self.upcoming.last() != Some(&b) {
-                                        self.upcoming.push(b);
-                                        self.tas_generated += 1;
-                                    }
-                                }
-                                None => {
-                                    // Part of an hwtracer block mapped to a machine block in the
-                                    // LLVM block address map, but the machine block has no
-                                    // corresponding AOT LLVM IR blocks.
-                                    //
-                                    // FIXME: https://github.com/ykjit/yk/issues/388
-                                    // We *think* this happens because LLVM can introduce extra
-                                    // `MachineBasicBlock`s to help with laying out machine code.
-                                    // If that's the case, then for our purposes these extra blocks
-                                    // can be ignored. However, we should really investigate to be
-                                    // sure.
-                                }
-                            }
-                        }
-                    }
+                    self.map_block(&x);
                 }
                 Some(Err(HWTracerError::Unrecoverable(x)))
                     if x == "longjmp within traces currently unsupported" =>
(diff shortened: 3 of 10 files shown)


vext01:
Typo in first commit message "mappend"

tests/c/trace_too_long.c
```diff
@@ -0,0 +1,46 @@
...
+int main(int argc, char **argv) {
+  YkMT *mt = yk_mt_new(NULL);
+  yk_mt_hot_threshold_set(mt, 0);
+  YkLocation loc1 = yk_location_new();
+  YkLocation loc2 = yk_location_new();
+
+  int i = 2000;
+  NOOPT_VAL(loc1);
+  NOOPT_VAL(loc2);
+  NOOPT_VAL(i);
```
vext01, line 28: Is the duplicated `NOOPT_VAL(i)` something deviously clever, or a typo?

tests/c/trace_too_long.c
```diff
@@ -0,0 +1,46 @@
...
+  NOOPT_VAL(i);
+  YkLocation *loc = &loc1;
+  while (i > 0) {
+    yk_mt_control_point(mt, loc);
+    if (i == 2000)
```
vext01, line 33: This is an unusual use of the JIT, but I guess there's no reason why it shouldn't be allowed...

tests/c/trace_too_long.c
```diff
@@ -0,0 +1,46 @@
...
+// # fast enough.
+// ignore-if: test "$YKB_TRACER" != "hwt"
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_PRINT_JITSTATE=1
+//   stderr:
+//     ...
+//     jit-state: stop-tracing
+//     jit-state: trace-compilation-aborted: Trace too long
+
```
vext01, line 11: What is the intention of this test? To check that long traces are aborted I think? It's conventional to add a short comment saying what the test is checking.

tests/c/trace_too_long_hwt.c
```diff
@@ -0,0 +1,46 @@
+// # This test checks that an overflowing PT buffer is caught at the point
```
vext01, line 1: Is this test the same as the last one, just with a higher loop bound?

ykrt/src/trace/hwt/mapper.rs
```diff
@@ -1,37 +1,119 @@
...
+/// Map the *machine* blocks of the specified trace into LLVM IR blocks.
+///
+/// Each entry in the returned trace is either a "mapped block" identifying a successfully
+/// mapped LLVM IR block, or an unsuccessfully mapped "unmappable block" (an unknown region of
+/// code spanning at least one machine block).
 pub(crate) struct HWTTraceIterator {
-    trace: std::vec::IntoIter<TraceAction>,
+    hwt_iter: Box<dyn Iterator<Item = Result<Block, HWTracerError>> + Send>,
+    /// The next [TraceAction]`s we will produce when `next` is called. We need this intermediary
+    /// to allow us to deduplicate mapped/unmapped blocks. This will be empty on the first
```
vext01, line 18: > deduplicate mapped/unmapped

Do we de-duplicate mapped blocks too? I did not know that (or I did once, but have forgotten that detail).

tests/c/trace_too_long.c
ltratt: I don't think "unusual" makes sense in the context of a test: this is a way of determinstically emulating an "unfortunate" trace.

ykrt/src/trace/hwt/mapper.rs
```diff
@@ -1,37 +1,119 @@
...
+    tas_generated: usize,
 }
 
 impl AOTTraceIterator for HWTTraceIterator {}
 
 impl Iterator for HWTTraceIterator {
     type Item = Result<TraceAction, AOTTraceIteratorError>;
     fn next(&mut self) -> Option<Self::Item> {
-        self.trace.next().map(|x| Ok(x))
+        while self.upcoming.len() < 2 {
```
vext01, line 31: Why 2?

ykrt/src/trace/hwt/mapper.rs
ltratt: You implemented it :p https://github.com/ykjit/yk/blob/master/ykrt/src/trace/hwt/mapper.rs#L190

ykrt/src/trace/hwt/mapper.rs
```diff
@@ -1,37 +1,119 @@
...
+                                    // `MachineBasicBlock`s to help with laying out machine code.
+                                    // If that's the case, then for our purposes these extra blocks
+                                    // can be ignored. However, we should really investigate to be
+                                    // sure.
+                                }
+                            }
+                        }
+                    }
+                }
+                Some(Err(HWTracerError::Unrecoverable(x)))
```
vext01, line 36: I think the above arm should be lifted out into a function. The locality of this next arm confused me for a moment: there's so much code above that I forgot that we are in a `match`.

ykrt/src/trace/hwt/mapper.rs
ltratt: To allow deduplication: you always need to have 1 element in this buffer in order to know if the next is a duplicate.

tests/c/trace_too_long.c
ltratt: Fixed in 3fafef6.

tests/c/trace_too_long.c
ltratt: Lengthened description in 3fafef6.

ykrt/src/trace/hwt/mapper.rs
```diff
@@ -1,37 +1,119 @@
...
+                    if x == "longjmp within traces currently unsupported" =>
```
vext01, line 37: ^ hrm. Does this motivate not using a plain string error? (not for this PR obviously).

tests/c/trace_too_long_hwt.c
ltratt: In essence, yes: in order to overflow PT we need a *loooooong* trace.

ykrt/src/trace/hwt/mapper.rs
```diff
@@ -1,37 +1,119 @@
...
+                {
+                    return Some(Err(AOTTraceIteratorError::LongJmpEncountered));
+                }
+                Some(Err(_)) => todo!(),
```
vext01, line 41: If we hit this todo, what has happened? Add a comment perhaps?

ykrt/src/trace/hwt/mapper.rs
ltratt: This is clearly horrible, but there's only so much of the rest of the codebase I want to change in one go...

ykrt/src/trace/hwt/mapper.rs
ltratt: I think a comment wouldn't add anything here.

ykrt/src/trace/hwt/mapper.rs
```diff
@@ -135,95 +217,3 @@ fn map_block(block: &hwtracer::Block) -> Vec<Option<TraceAction>> {
...
-                    TraceAction::Promotion => todo!(),
-                }
-            }
-        } else {
-            for irblock in irblocks.into_iter() {
-                if let Some(irb) = irblock {
-                    match ret.last() {
-                        Some(last) if &irb != last => ret.push(irb),
-                        Some(_) => {
-                            // The `BlockDisambiguate` pass in ykllvm ensures that no
```
vext01, line 191: We seem to have lost this comment in the churn. It's an important one worth keeping IMHO.

ykrt/src/trace/hwt/mapper.rs
ltratt: Keep going to a later commit and I think you'll be happy?

ykrt/src/trace/hwt/mapper.rs
```diff
@@ -135,95 +217,3 @@ fn map_block(block: &hwtracer::Block) -> Vec<Option<TraceAction>> {
...
-            // The block is unmappable. Insert a IRBlock that indicates this, but only if the
-            // trace isn't empty (we never report the leading unmappable code in a trace). We
-            // also take care to collapse consecutive unmappable blocks into one.
-            if let Some(last) = ret.last_mut() {
-                match last {
-                    TraceAction::MappedAOTBlock { .. } => {
-                        ret.push(TraceAction::new_unmappable_block());
-                    }
-                    TraceAction::UnmappableBlock => (),
-                    TraceAction::Promotion => todo!(),
```
vext01, line 182: The new code also doesn't appear handle (or for now, crash) on promotions in the same way? Have we masked a todo!?

ykrt/src/trace/mod.rs
```diff
@@ -57,7 +57,13 @@ pub(crate) trait TraceRecorder {
     fn stop(self: Box<Self>) -> Result<Box<dyn AOTTraceIterator>, InvalidTraceError>;
 }
 
-/// An iterator which [TraceRecord]s use to process a trace into [TraceAction]s.
+/// An iterator which [TraceRecord]s use to process a trace into [TraceAction]s. The iterator must
+/// respect the following:
+///
+/// 1. The control point call must not appear at the head of the trace.
```
vext01, line 63: I think I know what you mean, but it's worth clarifying.

The first item handed out by the iterator should be the (sole?) immediate successor of the block containing the control point call that caused the trace to be recorded?

ykrt/src/trace/hwt/mapper.rs
ltratt: I'm not sure why this was a `todo`, since the mapper should just pass it through? In the new setup that's what happens anyway...

ykrt/src/trace/mod.rs
ltratt: If I could parse that sentence, I might agree :p

ykrt/src/trace/mod.rs
```diff
@@ -57,7 +57,13 @@ pub(crate) trait TraceRecorder {
...
+/// 2. Consecutive `TraceAction`s must not compare equal (i.e. the iterator must have deduplicated
+///    consecutive `TraceAction`s).
+/// 3. The "stop tracing" call must not appear at the tail of the trace.
```
vext01, line 66: More precisely: the unmappable block containing the body of the "stop tracing" function must not appear at the tail of the trace. It's OK (and expected) for the prior unmappable block containing the call-site to the "stop tracing" function to be at the tail of the trace. I think that's right?

ltratt: You tell me!

vext01: I believe what I've said is correct. Did the old code strip any mappable blocks from the end? If not, I think that proves it.

vext01:
The PR is motivated by the need to speed up the interpreter by not blocking during mapping, but we didn't observe any speed ups? Do we know why?

ykrt/src/trace/hwt/mapper.rs
ltratt: Fixed in 86f0eb7.

tests/c/trace_too_long.c
vext01: What I don't get is: if the intention of the test is to make a really long trace, why do we need two meta-tracers?

ltratt:
> The PR is motivated by the need to speed up the interpreter by not blocking during mapping, but we didn't observe any speed ups?

Not exactly. Since the LLVM backend mostly crashes without syncrhonised compilation turned on, we can't really test non-synchronised compilation right now. The new backend should solve these problems!

ykrt/src/trace/hwt/mapper.rs
vext01: OK, we really need to put that comment back in :)

tests/c/trace_too_long.c
ltratt: Did you notice the subtle difference in the way the tests signify failure? PT signifies failure in `stop`, hwt mapping in "mapping". We *could* unify those two things, but that would be more work. So we currently need (and needed before this PR!) two separate tests.

ykrt/src/trace/hwt/mapper.rs
ltratt: See https://github.com/ykjit/yk/pull/1008#discussion_r1512524230 and b95786d4feaea78efb099575ff02bd0e055ff4e6.

ykrt/src/trace/hwt/mapper.rs
vext01: I'm not sure what it means to hit that error.

ykrt/src/trace/mod.rs
ltratt: In fairness, the code in this PR a) does what you're suggesting b) documents it in detail at line 48. So I think we're good?

ykrt/src/trace/hwt/mapper.rs
vext01: If you believe it's correct, then that's fine.

ykrt/src/trace/hwt/mapper.rs
ltratt: Nor I. When we hit the `todo` we'll find out. That's the glory of `todo`s :)

tests/c/trace_too_long.c
vext01: My apologies, I wasn't clear.

What I meant to say is: why do you need two *locations* to make a long trace? I'm not saying you don't, but it isn't obvious to me.

ykrt/src/trace/hwt/mapper.rs
ltratt: Well at the moment hwt can't generate `TraceAction::Promotion` so the point's moot!

ykrt/src/trace/hwt/mapper.rs
vext01: Thanks

tests/c/trace_too_long.c
ltratt: You need to start tracing and then not have it stop for ages. You can do this in various ways: I randomly chose to do it in this way, in part because I don't think the optimiser can easily worm around this. That said, I wouldn't read too much into it other than "the test does test the thing we want".

ykrt/src/trace/mod.rs
vext01: I'm just adding more detail, as this is very nuanced code and we are likely going to have to revisit these comments later, for ourselves and others. The facts I know are:

 - The mappable block (call it `Bc`) containing the call-site of the control point doesn't appear in the trace.
 - The unmappable body of the control point is removed.
 - The first block given out is the successor block of `Bc`.

ltratt: Fixed in d5b916e.

ltratt:
I *think* that's everything, but let me know!

vext01:
Good. Please squash and fix typo in the commit message :)

PR #1010, opened 2024-03-06
Fix warnings in swt.

ltratt:
Also update `.buildbot.sh` so these problems don't recur.

### DIFF
--- a/ykrt/src/trace/swt/mod.rs
+++ b/ykrt/src/trace/swt/mod.rs
@@ -1,8 +1,7 @@
 //! Software tracer.
 
-use crate::frame::BitcodeSection;
-
 use super::{errors::InvalidTraceError, AOTTraceIterator, TraceAction, TraceRecorder};
+use crate::{frame::BitcodeSection, mt::DEFAULT_TRACE_TOO_LONG};
 use std::sync::Once;
 use std::{cell::RefCell, collections::HashMap, error::Error, ffi::CString, sync::Arc};
 
@@ -99,10 +98,8 @@ impl TraceRecorder for SWTTraceRecorder {
                     }
                 });
 
-                aot_blocks = tb
-                    .borrow()
-                    .iter()
-                    .map(|tb| match fnames.borrow_mut().get(&tb.function_index) {
+                aot_blocks.extend(tb.borrow().iter().map(|tb| {
+                    match fnames.borrow_mut().get(&tb.function_index) {
                         Some(name) => TraceAction::MappedAOTBlock {
                             func_name: name.to_owned(),
                             bb: tb.block_index,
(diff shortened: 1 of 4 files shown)


ykrt/src/trace/swt/mod.rs
```diff
@@ -99,10 +98,8 @@ impl TraceRecorder for SWTTraceRecorder {
                     }
                 });
 
-                aot_blocks = tb
-                    .borrow()
-                    .iter()
-                    .map(|tb| match fnames.borrow_mut().get(&tb.function_index) {
+                aot_blocks.extend(tb.borrow().iter().map(|tb| {
```
vext01, line 101: This looks like a semantic change?

vext01:
Are we sure this only fixes warnings?

ykrt/src/trace/swt/mod.rs
ltratt: Because of `let mut aot_blocks: Vec<TraceAction> = vec![]` above, it's semantically equivalent. [Or at least I think it is!]

ltratt:
> Are we sure this only fixes warnings?

?

ykrt/src/trace/swt/mod.rs
```diff
@@ -111,20 +108,23 @@ impl TraceRecorder for SWTTraceRecorder {
...
                             tb.function_index
                         ),
-                    })
-                    .collect();
+                    }
+                }));
             })
         });
-        Ok(Box::new(SWTraceIterator::new(aot_blocks)))
+        if aot_blocks.len() > DEFAULT_TRACE_TOO_LONG {
```
vext01, line 115: What about this stuff? Looks like some semantic fixes.

That's fine, but I was expecting only fixing warnings :)

ltratt: It's both. The warning was correct: the backend *has* to handle "trace too long" and it wasn't referencing the `DEFAULT_TRACE_TOO_LONG` variable.

ykrt/src/trace/swt/mod.rs
ltratt: It has to be defined earlier so that it can be referenced in the `with` closure. There are probably nicer ways of doing this, but this is the only short-term fix I can see.

vext01: ah right.

ykrt/src/trace/swt/mod.rs
ltratt: I had to look this one up to remind myself!

ltratt:
It's become outdated by other PRs. Fixing.

ltratt:
Fixed in 42f04aa. Note that these new changes are identical to those proposed in https://github.com/ykjit/yk/pull/1022/. I clearly need to go through and fix these more thoroughly soon, but I'm not sure either of these PRs is the time to do it.

vext01:
Please squash.

ltratt:
Squashed.

ltratt:
@ptersilie Aha, `Temporary(TraceBufferOverflow)` is the error I've been tracking elsewhere (I ran *10,000* tests earlier without triggering this -- it's only the second time I've seen it!). I'm about 99% certain it's not related to this PR so can we retry the merge?

ltratt:
Hmph, same warning. I think we should retry again. I'll raise a PR to fix the `TraceBufferOverflow` elsewhere.

ltratt:
I was wrong: it's a benchmark that's failing because it doesn't work with swt. https://github.com/ykjit/yk/pull/1010/commits/21134e6b4264b009b2949cc87ca659e96ea1ba84 simply doesn't run those benchmarks on swt. We'll have to look at those again.

OK to squash?

ptersilie:
Please squash.

ptersilie:
> @ptersilie Aha, `Temporary(TraceBufferOverflow)` is the error I've been tracking elsewhere (I ran _10,000_ tests earlier without triggering this -- it's only the second time I've seen it!). I'm about 99% certain it's not related to this PR so can we retry the merge?

Is that something I should know about?

ltratt:
Squashed.

PR #1011, opened 2024-03-07
Type check calls (and some unrelated bits)

vext01:
The third commit could use discussion. Should we enforce that a `TypeIdx` uniquely identifies a `Type`? We've not built the system with that in mind, but it may not take much to build that assumption in soonish. It would make type checks very cheap...

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64.rs
@@ -249,7 +249,11 @@ impl<'a> X64CodeGen<'a> {
         let op2 = inst.op2();
 
         // The types must match. If the IR is well-formed, this is guaranteed.
-        debug_assert_eq!(op1.type_(self.jit_mod), op2.type_(self.jit_mod));
+        debug_assert_eq!(
+            op1.type_(self.jit_mod),
+            op2.type_(self.jit_mod),
+            "attempt to add different types"
+        );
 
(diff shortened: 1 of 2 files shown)


vext01:
The only reason I can think of is that it's cheaper to get a type index for a type if you don't have to check if it's already extant. But you likely void that time cost later, when you can't type check by indices, and have to compare (via the `Eq` trait) `Type` values (which aren't always word-sized).

I think I'm in favour of strengthening the assumption, if you are. We can add debug_asserts to check that it isn't violated.

ltratt:
> I think I'm in favour of strengthening the assumption, if you are

Yes, we should strengthen the assumption IMHO. [Where "strengthen" can simply mean "document the guarantee".]

vext01:
Yep. Do you want a fresh PR for that, or shall I stack it on here?

ltratt:
Please do whichever you think best.

vext01:
https://github.com/ykjit/yk/pull/1011/commits/ae7e7a9cf05928f7ea2baa7c571bc479826342df documents it and adds an assertion.

The positioning of the (debug) assertion makes it expensiveish. Is that OK?

We could instead do the check once at the end, but that might be too late. If someone inserts duplicate types, we might get weird crashes before we get to the assertion...

ltratt:
In general I'm not super worried about the cost of a `debug_assert`.

Please squash.

vext01:
Thanks. Squashed.

PR #1012, opened 2024-03-07
Store the currently running trace in `MTThread`.

ltratt:
Previously we passed around `Arc<CompiledTrace>` as `*const` so that side-tracing/side-traces could access it. This is a very delicate, unsafe, `into_raw`/`from_raw` dance.

This commit is based on the realisation that side-tracing/side-traces by definition run in the same trace that the parent trace was just running in. We can therefore store the currently running trace in `MTThread`, and when side-tracing occurs, we pluck it out, and it becomes the parent trace. Doing so reduces the quantity of unsafe code -- in particular, unsafe code that's hard to audit when other changes are made. It does introduce slightly higher costs, because we have to access `MTThread` more often, but hopefully that's not really measurable.


vext01:
Looks like a good simplification.

Can two interpreter threads still run the same trace under this change?

ltratt:
They should be able to. Not that we can very easily test that...

vext01:
I'll let the future test this hypothesis :)

PR #1015, opened 2024-03-07
Allow downcasting from `CompiledTrace`.

ltratt:
At the moment we pass `CompiledTrace` to a backend which makes it difficult to deal with, because the backend really wants the concrete implementation. This PR allows that to happen with safe code. What we really want to do is upcast an `Arc<dyn CompiledTrace>` to `Arc<dyn Any>` but we have to do so mechanically as trait upcasting isn't yet stable (https://github.com/ykjit/yk/pull/1015/commits/b49f042ea604ef142aa8eefcd353de640c1539c1). This then allows us to downcast (https://github.com/ykjit/yk/pull/1015/commits/ba1876ff73f1450427e86686671142ee5adbc3d0).

Because it's fairly obviously (I think... hope!) an LLVM backend-only thing, I have only pushed `smap` down from `CompiledTrace` to `LLVMCompiledTrace` as an example. @ptersilie If you like the look of this PR you can then use it as an example of how to push other methods out of `CompiledTrace` as you realise that they're relevant to `LLVMCompiledTrace` but not jitc_yk's implementation of `CompiledTrace`.

[Note that this PR starts with a couple of PRs which make `deopt.rs` jitc_ykllvm-only. I should have done this a while ago, but it didn't really occur to me until now.]


ptersilie:
This looks good. :+1: for making `jitc_llvm` a directory. It's been bothering me for a while. :-)

PR #1016, opened 2024-03-08
Allow yk to use cargo profiles other than just "debug|release".

ltratt:
As I start to profile yk more, I find myself continually adding a temporary profile that turns debugging info in release mode on. After a while I realised that there's no reason we can't a) provide such a profile by default b) have yk make using it easy. This commit does just that. Tested working as hoped/expected with yklua.

### DIFF
--- a/bin/yk-config
+++ b/bin/yk-config
@@ -24,20 +24,20 @@ OUTPUT=""
 usage() {
     echo "Generate C compiler flags for building against the yk JIT.\n"
     echo "Usage:"
-    echo "    yk-config <mode> [--prelink-pipeline <passes>] [--postlink-pipeline <passes>] <--cc|--cxx|--ar|--ranlib|--cppflags|--cflags|--ldflags>\n"
-    echo "    Where <mode> is either 'debug' or 'release'."
+    echo "    yk-config <profile> [--prelink-pipeline <passes>] [--postlink-pipeline <passes>] <--cc|--cxx|--ar|--ranlib|--cppflags|--cflags|--ldflags>\n"
+    echo "    Where <profile> is a Rust cargo profile starting with either 'debug' or 'release'."
 }
 
 
 handle_arg() {
-    mode=$1
+    profile=$1
     shift
 
     if [ "x${YKB_YKLLVM_BIN_DIR}" != "x" ]; then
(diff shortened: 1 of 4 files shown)


bin/yk-config
```diff
@@ -24,20 +24,20 @@ OUTPUT=""
 usage() {
     echo "Generate C compiler flags for building against the yk JIT.\n"
     echo "Usage:"
-    echo "    yk-config <mode> [--prelink-pipeline <passes>] [--postlink-pipeline <passes>] <--cc|--cxx|--ar|--ranlib|--cppflags|--cflags|--ldflags>\n"
-    echo "    Where <mode> is either 'debug' or 'release'."
+    echo "    yk-config <profile> [--prelink-pipeline <passes>] [--postlink-pipeline <passes>] <--cc|--cxx|--ar|--ranlib|--cppflags|--cflags|--ldflags>\n"
+    echo "    Where <profile> is a Rust cargo profile starting with either 'debug' or 'release'."
```
vext01, line 28: Do we need to update the list of possible values, or does the new profile not use a new subdir in `target`?

ltratt: It creates/uses a new directory in `target`.

bin/yk-config
```diff
@@ -197,21 +197,21 @@ if [ $# -eq 0 ]; then
 fi
 
 case $1 in
-    debug|release);;
-    *) 1>&2 echo "unknown mode: $1\n"
+    debug*|release*);;
+    *) 1>&2 echo "Profile '$1' does not start with 'debug' or 'release'.\n"
```
vext01, line 201: Should it be: `debug|release|release-with-debug)`? That way we reject (for example) `debug-with-donuts`.

bin/yk-config
vext01: OK, so should we enumerate the valid profiles?

```
Where <profile> is a Rust cargo profile: 'debug'/'release'/'release-with-debug'."
```

vext01:
LGTM, with only minor comments.

bin/yk-config
ltratt: The PR deliberately allows `debug-with-donuts` in case you want to do so!

vext01: Ah, in case you define a target in your local config? gotcha.

ltratt: You've got it :)

PR #1020, opened 2024-03-08
Ensure that worker threads don't keep `MT` alive.

ltratt:
If a worker thread (in non-synchronised-compilation) started working, it gained a strong reference to an `MT` instance which stopped the `MT` instance from being dropped. That in turn meant that `YKD_STATS` weren't printed.

This commit fixes things by only handing a weak reference to worker threads, ensuring they can't spuriously keep an `MT` instance alive. The new test `stats4.c` ensures we can't regress in this regard.

### DIFF
--- a/tests/c/stats4.c
+++ b/tests/c/stats4.c
@@ -1,0 +1,35 @@
+// Run-time:
+//   env-var: YKD_STATS=-
+//   stderr:
+//     {
+//       ...
+//       "traces_compiled_ok": ...
+//       ...
+//     }
+
+// This tests that spun-up compile workers still cause stats to be printed out
+// (i.e. that, because of reference counting, they don't keep `Mt` alive
+// forever).
+
+#include <assert.h>
+#include <setjmp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+int main(int argc, char **argv) {
+  YkMT *mt = yk_mt_new(NULL);
+  yk_mt_hot_threshold_set(mt, 0);
+  YkLocation loc = yk_location_new();
+
+  for (uint64_t i = 0; i < 2; i += 1) {
+    yk_mt_control_point(mt, &loc);
+  }
+
+  yk_location_drop(loc);
+  yk_mt_drop(mt);
+  return (EXIT_SUCCESS);
+}
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -209,18 +209,23 @@ impl MT {
                 return;
             }
 
-            let mt = Arc::clone(self);
+            self.stats.timing_state(TimingState::None);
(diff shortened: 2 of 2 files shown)


tests/c/stats4.c
```diff
@@ -0,0 +1,35 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+int main(int argc, char **argv) {
+  YkMT *mt = yk_mt_new(NULL);
+  yk_mt_hot_threshold_set(mt, 0);
```
vext01, line 25: Would a threshold of 8 maximise the possibility that the main thread exits before a compilation thread?

vext01:
Just to check: there's no scenario where the main thread exits before a compilation thread? This would cause a dangling weak ref I think.

tests/c/stats4.c
ltratt: I'm not sure I understand the question, but there will always be scenarios where something can kill the system (e.g. with `exit` on the main thread) in a way that won't cause stats to be printed. This PR fixes one of those scenarios, but we probably can't fix all of them.

vext01: Hrm, I think we crossed wires.

I think the intention of this test is to show that if a compilation thread is still running when the main thread dies, that stats are still printed.

If that's the case, then a threshold of 0 gives the compilation thread 9 iterations worth of time to run before the main thread exits.

But if you set it to 8, you give it 0 iterations of time to run before the main thread exits.

The latter is more likely to capture what you want to test, isn't it? In the former you give the compilation thread more time to finish, and it may exit before the main thread.

vext01: Understood, but we want to eventually have a fast code generator. If it works with 8, then let's maximise the chances that we test the property we want to, and use 8?

Maybe a more direct approach is to use a loop with only 2 iterations and set the threshold to 1?

ltratt: I think we're over thinking this tiny change, and trying to predict the future when we can just wait and see.

ltratt: Just to be clear: this test fails without the associated change. So the test is testing *something* meaningful.

vext01: I'm not certain you've understood my explanations.

Do you agree with the following statements?

 - Which thread exits first is non-deterministic.
 - You want to test the case where the main thread exits first.
 - The more time the main thread runs (the remained of the loop iterations after the compilation thread has started), the lower the probability that the main thread will exit first.

If yes, to all of the above, I think you can make a more robust test by simply lowering the loop bound to `hot_threshold + 1`.

ltratt: Changed in c4aa14d.

vext01:
Thanks. Please squash!

ltratt:
Squashed.

ltratt:
@vext01 That's an interesting failure but, I believe, unrelated to this PR.

vext01:
We've seen that one before: https://github.com/icedland/iced/issues/383

PR #1022, opened 2024-03-09
Fiddle with frame parts

ltratt:
This PR started with an `unsafe` audit of frame reconstruction, which is in the first 4 commits. I then realised that's not very useful given the code will probably be replaced soon, but the commits don't hurt so I've left them in.

The real meat of the PR is recognising that the `frame` module is mostly jitc_llvm specific, so https://github.com/ykjit/yk/commit/5937d9e3fc77ba86e7d88a957d57f5f69d551032 moves the code into jitc_llvm. I then extracted what I believe is the non-jitcllvm specific part into a top-level module https://github.com/ykjit/yk/commit/fc09599cc53d68eec6b8120fa7d3afe1311a28f1.


ptersilie:
This looks fine. Two thoughts:

1) Most stuff in `frame/mod.rs` can be reused for other backends. The only thing that might differ is how the JIT live variables are represented. But we could find a common representation and update this for the LLVM backend too. We can leave the file in `jitc_llvm` for now, but I'll likely soon copy it back to a more general place with small adjustments.

2) You're right that AOT stackmaps will need to be accessed by other JIT backends as well. However, since we have to create new stackmaps for the JIT, we might want to use the same representation at AOT time. Though that's a larger change and won't happen until way later.

ltratt:
> But we could find a common representation and update this for the LLVM backend too. We can leave the file in jitc_llvm for now, but I'll likely soon copy it back to a more general place with small adjustments.

That's fine with me!

ptersilie:
> Most of this is specific to the LLVM backend in the sense that it's
really another part of `depot.rs`, so so this is a more logical place
for it.

Some typos here: "depot.rs" and "so so"

ltratt:
OK I can force push an update with a fixed commit description when you would like me to.

ptersilie:
Please do.

ltratt:
Force pushed an update.

ptersilie:
```
12:46:44 error[E0432]: unresolved import `crate::frame`
12:46:44  --> ykrt/src/trace/swt/mod.rs:3:12
12:46:44   |
12:46:44 3 | use crate::frame::BitcodeSection;
12:46:44   |            ^^^^^
12:46:44   |            |
12:46:44   |            unresolved import
12:46:44   |            help: a similar path exists: `crate::compile::jitc_llvm::frame`
12:46:44 
12:46:44 warning: unused import: `crate::trace::Tracer`
12:46:44    --> ykrt/src/trace/swt/mod.rs:125:9
12:46:44     |
12:46:44 125 |     use crate::trace::Tracer;
12:46:44     |         ^^^^^^^^^^^^^^^^^^^^
12:46:44     |
12:46:44     = note: `#[warn(unused_imports)]` on by default
12:46:44 
12:46:44 warning: unused imports: `AOTTraceIteratorError`, `AOTTraceIterator`, `TraceAction`
12:46:44    --> ykrt/src/trace/swt/mod.rs:126:24
12:46:44     |
12:46:44 126 |     use crate::trace::{AOTTraceIterator, AOTTraceIteratorError, TraceAction};
12:46:44     |                        ^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^
12:46:44 
12:46:44 warning: unused import: `std::sync::Arc`
12:46:44    --> ykrt/src/trace/swt/mod.rs:127:9
12:46:44     |
12:46:44 127 |     use std::sync::Arc;
12:46:44     |         ^^^^^^^^^^^^^^
```

ptersilie:
Yes, please go ahead.

ltratt:
Squashed.

PR #1024, opened 2024-03-12
Handle and codegen globals.

ptersilie:
This PR changes the `LoadGlobal` instruction into a `LookupGlobal` instruction which allows us to get rid of special global instructions, e.g. `StoreGlobal`, `LoadGlobal`, etc.

However, this comes with a slight performance penalty in the codegen. `LookupGlobal` simply copies a constant pointer into a register and then immediately uses that register. So we would get something like:
```
mov rax, 0x123abc
mov [rax], 0x1
```
rather than

```
mov [0x123abc], 0x1
```

To fix this we need to change the JIT IR to allow globals inside operands (we don't want to implement a special global version for each instruction, e.g. LoadGlobal/StoreGlobal/etc). The easiest way to do this is to make globals a subclass of constants, similarly to what LLVM does.

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -951,6 +951,15 @@ impl IRDisplay for GlobalDecl {
     }
 }
 
+impl GlobalDecl {
+    pub(crate) fn is_threadlocal(&self) -> bool {
+        self.is_threadlocal
+    }
+    pub(crate) fn name(&self) -> &str {
+        &self.name
+    }
+}
+
 /// A bytecode module.
 ///
 /// This is the top-level container for the AOT IR.
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64.rs
@@ -133,7 +133,6 @@ impl<'a> X64CodeGen<'a> {
             jit_ir::Instruction::PtrAdd(i) => self.codegen_ptradd_instr(instr_idx, &i),
             jit_ir::Instruction::Store(i) => self.codegen_store_instr(&i),
             jit_ir::Instruction::LoadGlobal(_) => todo!(),
-            jit_ir::Instruction::StoreGlobal(_) => todo!(),
             jit_ir::Instruction::Call(i) => self.codegen_call_instr(instr_idx, &i)?,
             jit_ir::Instruction::Icmp(i) => self.codegen_icmp_instr(instr_idx, &i),
             jit_ir::Instruction::Guard(i) => self.codegen_guard_instr(&i),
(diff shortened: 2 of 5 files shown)


ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -951,6 +951,15 @@ impl IRDisplay for GlobalDecl {
     }
 }
 
+impl GlobalDecl {
+    pub(crate) fn is_threadlocal(&self) -> bool {
+        self.is_threadlocal
+    }
+    pub(crate) fn name(&self) -> &String {
```
ltratt, line 958: In general `&String` is better as `&str`.

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -461,6 +451,7 @@ impl<'a> TraceBuilder<'a> {
                 Err(_) => todo!(),
             }
         }
+        println!("{}", self.jit_mod);
```
ltratt, line 454: Debugging print?

ptersilie: [c6caac0](https://github.com/ykjit/yk/pull/1024/commits/c6caac08ce1373a510414471896a0a7b80891b64)

ltratt:
Please squash.

ptersilie:
Squashed.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
```diff
@@ -343,6 +342,20 @@ impl<'a> X64CodeGen<'a> {
...
+        &mut self,
+        inst_idx: jit_ir::InstrIdx,
+        inst: &jit_ir::LookupGlobalInstruction,
+    ) {
+        let decl = inst.decl(self.jit_mod);
+        if decl.is_threadlocal() {
+            todo!();
+        }
+        let sym_addr = symbol_vaddr(&CString::new(decl.name()).unwrap()).unwrap();
+        dynasm!(self.asm ; mov Rq(WR0.code()), QWORD i64::try_from(sym_addr).unwrap());
```
vext01, line 355: I think the unwrap is safe, but we should comment why.

vext01:
Sorry, I had one late comment :)

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
ptersilie: [7f1ac54](https://github.com/ykjit/yk/pull/1024/commits/7f1ac545c4c0eea6d29789638c1720c4c95d57fa)

ptersilie:
@vext01 Ok to squash?

vext01:
Please squash.

ptersilie:
Whoops its trying to compile a comment. Fixed. Ok to squash?

ltratt:
Please squash.

ptersilie:
Squashed.

ltratt:
Please force push a rebase.

ptersilie:
Rebased.

PR #1027, opened 2024-03-12
Remove an old assert.

### LINKED ISSUE #960: Is `bb: usize::MAX` a valid value?
A question for @vext01. In `trace/mod.rs` we have:

```rust
pub enum TracedAOTBlock {
    /// A sucessfully mapped block.
    Mapped {
        /// The name of the function containing the block.
        ///
        /// PERF: Use a string pool to avoid duplicated function names in traces.
        func_name: CString,
        /// The index of the block within the function.
        ///
        /// The special value `usize::MAX` indicates unmappable code.
        bb: usize,
    },
    /// One or more machine blocks that could not be mapped.
    ///
    /// This usually means that the blocks were compiled outside of ykllvm.
    Unmappable,
}
```

AFAICS `bb = usize::MAX` is pointless since it's equivalent to TracedAOTBlock::Unmappable? Indeed, if I add an assert to `TracedAOTBlock::new_mapped`:

```rust
impl TracedAOTBlock {
    pub fn new_mapped(func_name: CString, bb: usize) -> Self {
        assert_ne!(bb, usize::MAX);
        Self::Mapped { func_name, bb }
    }
```

all our tests pass, so i guess the "special value" comment reflects some past state that's no longer relevant?

ltratt:
This has been in the code long enough that I feel confident that it will never trigger.

Fixes #960.

### DIFF
--- a/ykrt/src/trace/mod.rs
+++ b/ykrt/src/trace/mod.rs
@@ -107,10 +107,6 @@ pub enum TraceAction {
 
 impl TraceAction {
     pub fn new_mapped_aot_block(func_name: CString, bb: usize) -> Self {
-        // At one point, `bb = usize::MAX` was a special value, but it no longer is. We believe
-        // that no part of the code sets/checks for this value, but just in case there is a
-        // laggardly part of the code which does so, we've left this `assert` behind to catch it.
-        debug_assert_ne!(bb, usize::MAX);
         Self::MappedAOTBlock { func_name, bb }
     }
 


ltratt:
@vext01 That's a not-exactly-transient error that's not related to this PR. Can we retry?

vext01:
Looks like miscompilation

ltratt:
No, it's a PT trace overrunning. It turns out there are two ways this can happen and we only gracefully handle one of them so far.

vext01:
Can we kick this one now?

Issue #1039, opened 2024-03-17
Missing an API/approach for fatal errors


ltratt:
It's now reasonably clear to me (and I suppose I should have realised this earlier!) that we have two very different types of errors floating around:

1. "Something went wrong in a way that's reasonable and that the system can sensibly carry on from". For example we can't currently compile traces with `longjmp` in, but we still want to keep compiling traces etc.
2. "Something went wrong in a way that is unreasonable and the best thing we can do is exit immediately, tell the user what happened, and hope they can fix it". For example, if yk is compiled into an interpreter that is missing yk's special ELF sections, there's zero chance of the system working as intended, and it's better to panic than try to carry on.

We currently don't deal with this split very well for two completely different reasons:

a. We often use `return Err` for the second case above, when something like `unwrap` would be better.
b. In a multi-threaded context, killing one thread doesn't necessarily take the whole system down in the expected way.

We need to think of some sort of systematic approach and/or API sooner rather than later, before it becomes torturous to unpick this later.

ltratt:
We are making progress on this, to the point that I think having it as a generic issue is no longer useful. Closing.

PR #1040, opened 2024-03-17
Rethink `CompilationError`.

ltratt:
Before this PR `CompilationError` looked thus:

```rust
pub enum CompilationError {
    #[error("Unrecoverable error: {0}")]
    Unrecoverable(String),
    #[error("Temporary error: {0}")]
    Temporary(String),
}
```

However I noticed that `mt.rs` didn't do anything different based on whether an error was `Unrecoverable` or `Temporary`. Then when I looked in jitc_yk (in particular) I realised that we were using `CompilationError` as a catch-all that meant that it was unlikely mt could usefully do anything different with `Unrecoverable`/`Temporary`.

What I eventually realised (and this partly led to https://github.com/ykjit/yk/issues/1039) is that we overloaded `CompilationError` in jitc_yk for two completely different things:

1. We currently can't compile this trace (e.g. because it has `longjmp`), but another trace starting from this same location might be compilable, so try again.
2. An internal system invariant has broken.

The problem really comes with the second of those, which is fundamentally different to the first. For example, if we can't load yk's ELF sections, the entire system is broken, and there is no point in continuing execution. There are a number of other points where if an error happens, it means that we've made a mistake (there or elsewhere) and there is also no point in continuing execution. By percolating an `Err` upwards, however, we make it very hard to debug precisely where the invariant was broken.

This PR starts, crudely, to split apart these two classes of errors. It tries to make it so that invariant-breaking causes the system to die at the first point that we notice things going wrong. That, eventually, makes it clear that `CompilationError` should be reserved for "we couldn't compile this trace because it did reasonable thing *x* that we can't handle".

Although it won't be obvious, I (in a sense) started with what became the final commit (https://github.com/ykjit/yk/commit/72162f01dcb735bdbaebdc860b8b49dc1c4da8d1), gradually noticing places where we were using `CompilationError` for more general invariants of one sort or another. Some of those general invariants are, IMHO, fatal, so they're now `unwrap`s: some maybe (although I'm not fully convinced, but we can think again for a future PR) could be recoverable so they've become `Box<dyn Error>`. I fixed all of these things one-by-one (`git stash` was my friend): all of the commits in this PR pass all tests. Eventually at the end, I could do the "big" change in the final commit.

This PR definitely isn't perfect, and it probably makes a few things worse (which we can fix in the future if we need to), but I think overall it makes more things better, and is thus worth it. It also makes it easier to write code in jitc_yk due to the simpler API, and will definitely make debugging some things easier, which should be a useful productivity win overall.


vext01:
I'm glad you did this. I've been grappling with this distinction for a while, but due to time pressure haven't felt able to do this audit.

Should I review this commit by commit, or as one change?

ltratt:
> Should I review this commit by commit, or as one change?

As you prefer (but you might find "one change" easiest).

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -70,11 +70,13 @@ impl<'a> LocalNumbers<'a> {
...
     /// Panics if the instruction isn't present in the module.
-    fn idx(&self, instr: &Instruction) -> Result<InstrIdx, CompilationError> {
+    fn idx(&self, instr: &Instruction) -> InstrIdx {
         // FIXME: This is inefficient.
         for (idx, candidate) in self.m.instrs().iter().enumerate() {
             if ptr::addr_eq(instr, candidate) {
-                return Ok(InstrIdx::new(idx)?);
+                // The `unwrap` cannot fail because we won't have expanded `m.instrs()` beyond the
+                // ability of `InstrIdx` to represent an index.
+                return InstrIdx::new(idx).unwrap();
```
ptersilie, line 79: It's reasonable to assume we won't have more than 24^2 instructions, but I'm not sure we check that anywhere at the moment. In theory we could have a reaaaally long trace and then panic here. Ideally, we bail before that happens though.

ykrt/src/compile/jitc_llvm/mod.rs
```diff
@@ -284,7 +282,7 @@ impl Compiler for JITCLLVM {
             // The LLVM backend is now legacy code and is pending deletion, so it's not worth us
             // spending time auditing all of the failure modes and categorising them into
             // recoverable/temporary. So for now we say any error is temporary.
-            Err(CompilationError::Temporary("llvm backend error".into()))
+            Err(CompilationError("llvm backend error".into()))
```
vext01, line 285: I can't see which variant of `CompilationError` is being constructed any more. Which is it making?

ptersilie: https://github.com/ykjit/yk/pull/1040/commits/72162f01dcb735bdbaebdc860b8b49dc1c4da8d1#diff-fa66cad0aec598e0ed07d8166314f32521aca9b43e97016babccc058de8ddd00R26

ptersilie:
This looks good to me.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -1130,21 +1129,17 @@ impl Module {
...
 /// Deserialise an AOT module from the slice `data`.
-pub(crate) fn deserialise_module(data: &[u8]) -> Result<Module, CompilationError> {
-    match Module::from_bytes((data, 0)) {
-        Ok(((_, _), mut modu)) => {
-            modu.compute_local_operand_func_indices();
-            Ok(modu)
-        }
-        Err(e) => Err(CompilationError::Unrecoverable(e.to_string())),
-    }
+pub(crate) fn deserialise_module(data: &[u8]) -> Result<Module, Box<dyn Error>> {
```
vext01, line 1132: In the description you said:

> This PR starts, crudely, to split apart these two classes of errors. It tries to make it so that invariant-breaking causes the system to die at the first point that we notice things going wrong

I'm therefore not sure I understand this PR. Why don't you just panic here instead of bubbling an error up?

ykrt/src/compile/jitc_llvm/mod.rs
vext01: Ah. Github led me astray. If you click the type it shows the old definition.

ykrt/src/compile/jitc_yk/aot_ir.rs
ltratt: It's a minor design choice in this particular regard IMHO.

vext01: I don't understand. You said you want to crash ASAP for cases where the system can't operate, right? Here is one such case, but you opted to bubble up an error. So have I misunderstood something?

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: The JIT IR was explicitly designed so that index overflows could be temporary errors.

ykrt/src/compile/jitc_yk/aot_ir.rs
ltratt: In this case it just goes up one level. Personally I would inline this function, but I suggest not overthinking it.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Hang on. The comment actually says something interesting.

Is it true though? Does pushing an instruction beyond the index limit actually fail? I was under the impression it'd only fail if you try to materialise such an index. I don't think `TiVec` is aware of the fact that the index size is actually reduced (it relies on `usize` conversions, as I recall).

ykrt/src/compile/mod.rs
```diff
@@ -20,13 +22,8 @@ pub(crate) mod jitc_llvm;
 pub mod jitc_yk;
 
 /// A failure to compile a trace.
-#[derive(Debug, thiserror::Error)]
```
vext01, line 23: So this error type is now always a recoverable error? If so, shall we say that in a comment?

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: With this diff we get to the panic:
```
diff --git a/ykrt/src/compile/jitc_yk/jit_ir.rs b/ykrt/src/compile/jitc_yk/jit_ir.rs
index f0f24fda..66ab9f5b 100644
--- a/ykrt/src/compile/jitc_yk/jit_ir.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir.rs
@@ -1279,7 +1279,7 @@ impl Module {
         let int8_type_idx = types.len().into();
         types.push(Type::Integer(IntegerType::new(8)));
 
-        Self {
+        let mut this = Self {
             name,
             instrs: Vec::new(),
             extra_args: Vec::new(),
@@ -1290,7 +1290,14 @@ impl Module {
             int8_type_idx,
             func_decls: TiVec::new(),
             global_decls: TiVec::new(),
+        };
+
+        for _ in 0..33554432 { // 2^25
+            this.push(LoadTraceInputInstruction::new(0, this.ptr_type_idx()).into());
         }
+        panic!("got here ok");
+
+        this
     }
 
     /// Returns the type index of [Type::Void].
```

Which shows that the instruction backing vector can happily grow larger than an `InstrIdx` can express when encoded into the JIT IR.

We can either make `push()` fail (but does anything ever shortcut `push()`?), or we can just always propagate an error when an instruction index is created. In theory it's OK for the vector to grow bigger than `InstrIdx` as long as you never need to reference one of the large indices.

I don't know what's best, but the comment is incorrect at present.

ltratt: > We can either make push() fail

This is what we should have done (and if we didn't, we should do).

vext01: Go for it.

vext01: (I reckon all of the other index types will have the same issue, so check all the `Module::push_*` methods also?)

ykrt/src/compile/mod.rs
ltratt: There's no concept of "recoverable" or "unrecoverable" now: we just couldn't compile this particular trace. What the caller does with the failure is up to them. [In practise, mt will/should retry tracing the location a few times, and then eventually give up. It doesn't actually get that right now, though I have a half-finished branch which fixes that.]

vext01: Got it. Conceptually recoverable.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: Fixed in 0e549469.

ltratt:
@vext01 OK to squash?

ltratt:
Oh actually I don't think this needs squashing! So if you're OK with it, please feel free to merge.

vext01:
Not yet, I still don't understand something about https://github.com/ykjit/yk/commit/0e5494697cc0569f3353b71232e34197303fc6c7

vext01:
Now you can squash!

PR #1042, opened 2024-03-18
Remove broken test and the constant(s) it relies upon.

ltratt:
This test is very broken, using APIs that haven't existed for ages. While we do need a varargs test, this isn't really a useful basis for that.


vext01:
Why don't we update the test?

ltratt:
The test needs rewriting from scratch and, AFAIK, neither of our backends supports it anyway.

ltratt:
I'm not keen on raising an issue on this as it'll just get forgotten. TBH an ignored test that is broken is effectively invisible. If someone wants to fix it, go ahead, but at the moment it's pointlessly keeping alive an `#define` for no good reason: and *that* `#define` caused someone else to add a *separate* `#define` because they assumed it was necessary!

PR #1044, opened 2024-03-19
Event based testing

ltratt:
Because yk tries to do as much work as possible in threads, many of our tests have *had* to use `YKD_SERIALISED_COMPILATION` in order that they can guarantee that they're testing (e.g.) compiled code.

This commit adds a crude, but effective, ability for a test thread to suspend itself (without consuming undue CPU!) until a set of yk events have happened.

The `simple_non_serialised` test is an example of this:

```c
while (i > 0) {
  if (i == 3)
    __ykstats_wait_until(mt, test_compiled_event);
  ...
}

bool test_compiled_event(YkCStats stats) {
  return stats.traces_compiled_ok == 1;
}
```

In essence, `__ykstats_wait_until` runs the `test_compiled_event` whenever a yk event such as "a trace has been compiled" has occurred. If `test_compiled_event` returns false, the thread will be parked until another event has occurred. If and when `test_compiled_event` returns true, `__ykstats_wait_until` itself returns.

At the moment the "events" one can wait on are crude: you can detect, for example, the number of traces that have been compiled, but not that a *particular* location has been compiled. Will this turn out to be flexible enough in the long run? Time will tell. In the interim, this relatively simple mechanism should allow us to write tests that more accurately model yk running in the real-world.

### DIFF
--- a/tests/c/simple_non_serialised.c
+++ b/tests/c/simple_non_serialised.c
@@ -1,0 +1,54 @@
+// Run-time:
+//   env-var: YKD_PRINT_JITSTATE=1
+//   env-var: YKD_STATS=1
+//   stderr:
+//     jit-state: start-tracing
+//     i=4
+//     jit-state: stop-tracing
+//     i=3
+//     jit-state: enter-jit-code
+//     i=2
+//     i=1
+//     jit-state: deoptimise
+//   stdout:
+//     exit
+
+// Check that basic trace compilation in a thread works.
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+bool test_compiled_event(YkCStats);
+
+int main(int argc, char **argv) {
+  YkMT *mt = yk_mt_new(NULL);
+  yk_mt_hot_threshold_set(mt, 0);
+  YkLocation loc = yk_location_new();
+
+  int res = 9998;
+  int i = 4;
+  NOOPT_VAL(loc);
+  NOOPT_VAL(res);
+  NOOPT_VAL(i);
+  while (i > 0) {
+    yk_mt_control_point(mt, &loc);
+    if (i == 3)
+      __ykstats_wait_until(mt, test_compiled_event);
+    fprintf(stderr, "i=%d\n", i);
+    res += 2;
+    i--;
+  }
+  printf("exit");
+  NOOPT_VAL(res);
+  yk_location_drop(loc);
+  yk_mt_drop(mt);
+  return (EXIT_SUCCESS);
+}
+
+bool test_compiled_event(YkCStats stats) {
+  return stats.traces_compiled_ok == 1;
+}
--- a/ykcapi/yk_testing.h
+++ b/ykcapi/yk_testing.h
@@ -4,6 +4,7 @@
 #include <stdint.h>
 #include <stdio.h>
 
+// This function will only exist if the `hwt` tracer is compiled in to ykrt.
 size_t __yktrace_hwt_mapper_blockmap_len();
 
 // Blocks the compiler from optimising the specified value or expression.
--- a/ykcapi/yk_testing.h
+++ b/ykcapi/yk_testing.h
@@ -3,6 +3,38 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <yk.h>
+
+/// The statistics passed to the `test` function in `ykstats_wait_until`.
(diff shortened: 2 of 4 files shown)


tests/c/simple_non_serialised.c
```diff
@@ -0,0 +1,54 @@
...
+  YkLocation loc = yk_location_new();
+
+  int res = 9998;
+  int i = 4;
+  NOOPT_VAL(loc);
+  NOOPT_VAL(res);
+  NOOPT_VAL(i);
+  while (i > 0) {
+    yk_mt_control_point(mt, &loc);
+    if (i == 3)
```
ptersilie, line 39: Can I end up in a dead lock if I get this `i` wrong? i.e. it waits forever for the trace to compile but since I haven't reached the threshold yet that never happens? I can see myself walking into this one a lot when making tests. This is not a "can we fix this" comment, but rather a "do I need to watch out for this in the future" comment.

ltratt: Yes, if you want to wait for an event to complete, you need to make sure that you did whatever is necessary *for* that event to happen. There's no particular magic here, in that sense.

ptersilie:
Looks good to me.

vext01:
Typo in description:

>  until another event is occurred

vext01:
Love the idea!

Questions having not read any code yet:
 - What happens if the event never comes? Do we need a timeout?
 - Does the stats callback get traced and compiled-in to the trace?

ltratt:
We don't know if we need a timeout. Wait and see.

There's nothing clever about the stats, so it *could* be inlined (it won't, or at least not very deeply so, in practise as it goes via a Rush shim, but who knows if that'll change).

ykcapi/yk_testing.h
```diff
@@ -3,7 +3,40 @@
...
+  /// How many traces were recorded successfully?
+  uint64_t traces_recorded_ok;
+  /// How many traces were recorded unsuccessfully?
+  uint64_t traces_recorded_err;
+  /// How many traces were compiled successfully?
+  uint64_t traces_compiled_ok;
+  /// How many traces were compiled unsuccessfully?
+  uint64_t traces_compiled_err;
+  /// How many times have traces been executed? Note that the same trace can
+  /// count arbitrarily many times to this.
```
vext01, line 20: I'm not sure what is meant by:

 > Note that the same trace can count arbitrarily many times to this.

ltratt: 1 trace can be executed n times where n>=0.

ltratt: Yes.

vext01:
> There's nothing clever about the stats, so it could be inlined

That being the case, why doesn't this deadlock for this test?

If you make a trace T, and that trace's native code itself calls the wait logic, isn't that a case where you can't progress (unless another thread is also compiling traces?)

ltratt:
The index is chosen such that the iteration that's being traced is the iteration *before* we wait. See the test output to get a sense of this.

vext01:
Ah gotcha.

I suspect (but I'm not sure) that later we might  want to filter out calls to your waiting logic in the trace compiler. Let's wait and see.

vext01:
This LGTM.

If @ptersilie is happy we can merge.

vext01:
Hrm, the new test failed:

```
10:34:47 ---- lang_tests::simple_non_serialised.c stderr ----
10:34:47 
10:34:47 Pattern (error at line 5):
10:34:47    ...
10:34:47    |i=4
10:34:47    |jit-state: stop-tracing
10:34:47    |i=3
10:34:47 >> |jit-state: enter-jit-code
10:34:47    |i=2
10:34:47    |i=1
10:34:47    |jit-state: deoptimise
10:34:47    ...
10:34:47 
10:34:47 Text (error at line 5):
10:34:47    ...
10:34:47    |i=4
10:34:47    |jit-state: stop-tracing
10:34:47    |i=3
10:34:47 >> |i=2
10:34:47    |i=1
```

Maybe there's another source of non-determinism that only comes out on the CI machine?

ltratt:
Fixed (I think) in https://github.com/ykjit/yk/pull/1044/commits/1b03db7ab56ed50fe55a8663341fcad46c6365ad. Interestingly, we were always *technically* vulnerable to this flaw before, but we got away with it by chance because of a) `YKD_SERIALISED_COMPILATION` and (I think) x86's strong memory model!

ltratt:
@vext01 This has survived extensive test runs so I think is ready for merging.

vext01:
Please, fmt, squash and force push.

ltratt:
I'm an idiot! Squashed.

PR #1049, opened 2024-03-21
Don't write `YKD_STATS` output to a file called 1.

ltratt:
This tests needs to turn `YKD_STATS` on for the event based logging to work, and it's better if that output doesn't appear on stderr, as it muddies the test. However `YKD_STATS=1` writes to a file called `1`! This commit changes it so that `YKD_STATS` output is written to `/dev/null`, satisfying both constraints.

### DIFF
--- a/tests/c/simple_non_serialised.c
+++ b/tests/c/simple_non_serialised.c
@@ -1,6 +1,6 @@
 // Run-time:
 //   env-var: YKD_PRINT_JITSTATE=1
-//   env-var: YKD_STATS=1
+//   env-var: YKD_STATS=/dev/null
 //   stderr:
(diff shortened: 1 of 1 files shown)


vext01:
If I were nitpicking, I'd add a `ignore-if` if `/dev/null` doesn't exist (other platforms).

I won't die on on near that hill :) What do you think?

ltratt:
I bet you even Windows supports `/dev/null` this these days. I'm not going to worry about platforms I can't test on: what file would I even use in the test?

vext01:
I meant you could skip the test, but maybe it's actually better to crash anyway. That way we can devise a test when the time comes.

Happy to merge this once CI issues are resolved.

PR #1051, opened 2024-03-21
Rework globals

vext01:
This implements an idea Laurie had for locating global symbols.

ykllvm change ~~coming soon~~ (and should go in first): https://github.com/ykjit/ykllvm/pull/122

See commits.

### DIFF
--- a/tests/c/simple.newcg.c
+++ b/tests/c/simple.newcg.c
@@ -18,7 +18,7 @@
 //     ...
 //     --- End jit-pre-opt ---
 //     ...
-//     deopt...
+//     jit-state: enter-jit-code
 //     ...
 
 // Check that basic trace compilation works.
@@ -48,6 +48,10 @@ int main(int argc, char **argv) {
   NOOPT_VAL(i);
   while (i > 0) {
     yk_mt_control_point(mt, &loc);
+    // FIXME: ideally we'd print to stderr so as to have the output interleaved
+    // with jit-state prints, but we can't yet handle the `stderr` constant in
+    // `fputs("i", stderr)`.
+    puts("i");
     res += 2;
     i--;
   }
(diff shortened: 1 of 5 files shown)


ltratt:
LGMT.

ptersilie:
In regards to testing this: with https://github.com/ykjit/yk/pull/1030 being merged now we get a normal panic when the guard fails, so it should flush the puts. Have you tried this?

tests/c/simple.newcg.c
```diff
@@ -18,7 +18,7 @@
 //     ...
 //     --- End jit-pre-opt ---
 //     ...
-//     deopt...
```
ptersilie, line 21: Could you leave this "deopt" string in here so we know if we break guards and we don't deopt anymore?

vext01: Not in this PR, because the trace doesn't loop.

My next PR brings this back.

ltratt:
@vext01 Have you seen https://github.com/ykjit/yk/pull/1051#issuecomment-2011973596 ?

vext01:
Yes, but the answer is the same as the other comment: the test doesn't make it to deopt, since the trace doesn't loop.

The test is expected to do 2 iterations of the loop in JITted code, but we don't yet have the ability to do >1 iteration. My next PR will fix this.

ltratt:
Got it! Please force push a rebase against a master.

vext01:
rebased.

vext01:
(requires the ykllvm change to merge)

tests/c/simple.newcg.c
ptersilie: It shouldn't loop anyway. For me it deopts immediately.

ptersilie:
> Yes, but the answer is the same as the other comment: the test doesn't make it to deopt, since the trace doesn't loop.
> 
> The test is expected to do 2 iterations of the loop in JITted code, but we don't yet have the ability to do >1 iteration. My next PR will fix this.

Why do I get the `deopt` panic then? For me it doesn't loop. Or did this PR introduce changes that now make this loop. In which case, why don't we change it back so it deopts in the first execution of the trace.

ltratt:
I'm interpreting this discussion as something that can easily be changed in a future PR if necessary, so I can merge.

PR #1055, opened 2024-03-22
Various mt.rs changes

ltratt:
This PR collects together various changes to `mt.rs`. I still have more things to audit and check, but this already feels sufficiently like a jumble sale that it's probably worth offering this for comments before going any further.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -497,7 +497,9 @@ impl MT {
         })
     }
 
-    /// Perform the next step to `loc` in the `Location` state-machine for a guard failure.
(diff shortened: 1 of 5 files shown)


ykrt/src/mt.rs
```diff
@@ -464,10 +464,10 @@ impl MT {
                         // This thread is tracing something, so bail out as quickly as possible
                         return TransitionControlPoint::NoAction;
                     }
-                    match loc.count() {
+                    match loc.inc_count() {
```
ptersilie, line 462: Took me a while to confirm that this change is semantically the same.

PR #1058, opened 2024-03-25
Make the state machine encoded in `MTThread` explicit.

ltratt:
Previously `MTThread` was effectively:

```rust
pub(crate) struct MTThread {
  running_trace: RefCell<Option<Arc<dyn CompiledTrace>>>,
  tracing: RefCell<Option<Arc<Mutex<HotLocation>>>>,
  thread_tracer: RefCell<Option<Box<dyn TraceRecorder>>>,
  promotions: RefCell<Option<Vec<usize>>>,
```

It's far from obvious but these 4 fields encode a state machine about the current thread's state: and there are 3 separate states in that state machine! This was far from obvious to me when I started: I spent quite a while confused about the exact relationship between these fields before realising this.

This commit breaks the statemachine out into an `enum` which is then stored in a field in `MTThread`:

```rust
enum MTThreadState {
    Interpreting,
    Tracing {
        hl: Arc<Mutex<HotLocation>>,
        thread_tracer: Box<dyn TraceRecorder>,
        promotions: Vec<usize>,
    },
    Executing(Arc<dyn CompiledTrace>),
}
```

This was quite a bit more painful to do than the diff might suggest.

On the plus side, for example, we'd accidentally duplicated the `HotLocation` in one part of the code, so we had two different variables storing references to the same `HotLocation`. This commit fixes that.

On the negative side, we've forced a bit more responsibility onto whoever calls a `transition_*` function. I think I might spy a future opportunity to simplify some of this stuff, but that's something to tackle in the future. This commit already introduces plenty enough churn!

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -260,7 +260,7 @@ impl MT {
                     f(ctrlp_vars, frameaddr);
                 }
             }
-            TransitionControlPoint::StartTracing => {
+            TransitionControlPoint::StartTracing(hl) => {
                 #[cfg(feature = "yk_jitstate_debug")]
                 print_jit_state("start-tracing");
                 let tracer = {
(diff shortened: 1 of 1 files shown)


ykrt/src/mt.rs
```diff
@@ -629,57 +649,73 @@ impl Drop for MT {
     }
 }
 
+/// [MTThread]'s major job is to record what state in the "interpreting/tracing/executing"
+/// state-machine this thread is in. This enum contains the states.
+enum MTThreadState {
+    /// This thread is executing in the normal interpreter: it is not executing a trace or
+    /// recording a trace.
+    Interpreting,
```
ptersilie, line 657: The name "Interpreting" might be a bit confusing in that it appears to suggest that this thread is interpreting something, when in fact the parent thread is interpreting and this thread is really doing nothing, right?

ltratt: It means this thread is doing something that isn't a) tracing or b) executing a trace. "Interpreting" was the best term I could think of for that "something" (so far, at least).

ykrt/src/mt.rs
```diff
@@ -629,57 +649,73 @@ impl Drop for MT {
...
+        /// [HotLocation] the trace started at; for a side-trace, tracing started elsewhere.
+        hl: Arc<Mutex<HotLocation>>,
+        /// What tracer is being used to record this trace? Needed for trace mapping.
+        thread_tracer: Box<dyn TraceRecorder>,
+        /// Records the content of data recorded via `yk_promote`.
+        promotions: Vec<usize>,
+    },
+    /// This thread is executing a trace. Note that the `dyn CompiledTrace` serves two different purposes:
+    ///
+    /// 1. It is needed for side traces and the like.
```
ptersilie, line 670: "It is needed for trace execution, side traces, and the like"?

ltratt: It's not needed for trace execution AFAIK.

ykrt/src/mt.rs
```diff
@@ -629,57 +649,73 @@ impl Drop for MT {
...
+    /// 2. It allows another thread to tell whether the thread that started tracing a [Location] is
+    ///    still alive or not by inspecting its strong count (if the strong count is equal to 1
+    ///    then the thread died while tracing). Note that this relies on thread local storage
+    ///    dropping the [MTThread] instance and (by implication) dropping the [Arc] and
+    ///    decrementing its strong count. Unfortunately, there is no guarantee that thread local
+    ///    storage will be dropped when a thread dies (and there is also significant platform
+    ///    variation in regard to dropping thread locals), so this mechanism can't be fully relied
+    ///    upon: however, we can't monitor thread death in any other reasonable way, so this will
+    ///    have to do
```
ptersilie, line 679: Missing `.`. Only fix this if you're pushing other commits anyway.

ptersilie:
Looks good, bar some minor comments.

ykrt/src/mt.rs
ptersilie: Does it not contain the trace that is being executed?

ykrt/src/mt.rs
ptersilie: Yeah, just wanted to clear that up. I don't have a better suggestion at the moment. Maybe "Idling"?

ykrt/src/mt.rs
ptersilie: Ah sorry, I get it now. The trace is already being executed. Ignore this.

ykrt/src/mt.rs
ltratt: Fixed in e0042c05ebc90f7c65df54bfb205e7963d4ff796.

ptersilie:
Please squash.

ltratt:
Squashed.

PR #1059, opened 2024-03-26
Codegen constants.

vext01:
The first commit is the main work.

(The ykllvm sync is just routine, so we inherit the filtering out of debug instructions).

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64.rs
@@ -16,6 +16,7 @@ use super::{
 #[cfg(any(debug_assertions, test))]
 use crate::compile::jitc_yk::jit_ir::JitIRDisplay;
 use crate::compile::CompiledTrace;
+use byteorder::{NativeEndian, ReadBytesExt};
 use dynasmrt::{
     components::StaticLabel, dynasm, x64::Rq, AssemblyOffset, DynasmApi, DynasmError,
     DynasmLabelApi, ExecutableBuffer, Register,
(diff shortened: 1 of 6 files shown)


ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -16,11 +18,88 @@ use std::{
...
+// This is simple and can be shared across both IRs.
 pub(crate) use super::aot_ir::Predicate;
 
+/// A fixed-width integer type.
+///
+/// Signedness is not specified. Interpretation of the bit pattern is delegated to operations upon
+/// the integer.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(crate) struct IntegerType {
+    num_bits: u32,
```
ltratt, line 30: Do we really need `u32` bits?!

vext01: Probably not, but that's how it is everywhere else in our IRs. We semi-inherited this from llvm. They use `unsigned`, which on x86_64 is u32...

We can change it, but I suggest not in this PR. Maybe a task for the wolf?

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
```diff
@@ -300,13 +300,21 @@ impl<'a> X64CodeGen<'a> {
...
     /// Load a constant into the specified register.
     fn load_const(&mut self, reg: Rq, cidx: jit_ir::ConstIdx) {
-        match self.jit_mod.const_(cidx) {
-            jit_ir::Constant::U32(v) => {
-                dynasm!(self.asm; mov Rq(reg.code()), DWORD *v as i32)
+        let cst = self.jit_mod.const_(cidx);
+        let mut bytes = cst.bytes().as_slice();
+        let size = cst.type_idx().type_(self.jit_mod).byte_size().unwrap();
+        debug_assert_eq!(bytes.len(), size);
+        use byteorder::{NativeEndian, ReadBytesExt};
```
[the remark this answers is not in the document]
vext01: Yes, that should be moved to the top. I just do this when hacking and forgot to move it.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: Fair enough, this is something we can rejig later.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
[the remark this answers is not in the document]
vext01: Fixed in e8b27389

vext01:
Fixed the imports thing, and found a much simpler way to compute the bytes for constant integers (the two other commits).

OK to squash?

ltratt:
Please squash.

vext01:
splat.

PR #1060, opened 2024-03-27
Tracing function runtime patch

Pavel-Durov:
+ Move tracing function from `ykcapi` to `ykrt`.
+ Add runtime patching with a single `ret` instruction to the tracing function when jit transition between tracing and stop-tracing (including side traces).


ykrt/src/trace/swt/patch.rs
```diff
@@ -0,0 +1,134 @@
+use libc::{mprotect, size_t, sysconf, PROT_EXEC, PROT_READ, PROT_WRITE};
```
ltratt, line 1: Let's add a docstring at the top of the file explaining what the high-level intent is. And also, probably, a big fat warning that it's dangerous, but improves performance.

ykrt/src/trace/swt/patch.rs
```diff
@@ -0,0 +1,134 @@
...
+
+fn is_x86() -> bool {
+    std::env::consts::ARCH == "x86" || std::env::consts::ARCH == "x86_64"
+}
+
+/// This function is used to patch the `yk_trace_basicblock`
+/// function with a single `ret` (0xC3) instruction.
+#[cfg(tracer_swt)]
+pub(crate) unsafe fn patch_trace_function() {
+    if !is_x86() {
```
ltratt, line 76: See `codegen/mod.rs` for how to do this (better) at compile-time: guard `mod` with `#[cfg(target_arch = "x86_64")]`.

ykrt/src/trace/swt/patch.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    let func_address = ((function_ptr as usize) & !(page_size - 1)) as *mut c_void;
+    let page_size_aligned = (((function_ptr as usize) + mem::size_of_val(&function_ptr))
+        - (func_address as usize)) as usize;
+
+    // Set function memory region to be writable
+    let result = mprotect(
+        func_address,
+        page_size_aligned,
+        PROT_READ | PROT_WRITE | PROT_EXEC,
```
ptersilie, line 53: You don't need the `PROT_EXEC` here since you are changing it back to executable again below. In fact, on some systems you are not allowed to mark a region writeable and executable at the same time.

ykrt/src/trace/swt/patch.rs
```diff
@@ -0,0 +1,79 @@
...
+
+#[cfg(tracer_swt)]
+use crate::yk_trace_basicblock;
+
+#[cfg(tracer_swt)]
+static ORIGINAL_INSTRUCTIONS_INIT: Once = Once::new();
+#[cfg(tracer_swt)]
+static mut ORIGINAL_INSTRUCTIONS: [u8; 1] = [0; 1];
+#[cfg(tracer_swt)]
+static mut PATCH_INSTRUCTIONS: [u8; 1] = [0xC3];
```
ptersilie, line 13: Is this a nop?

Pavel-Durov: yes, 0xC3 is a return instruction

ykrt/src/trace/swt/patch.rs
Pavel-Durov: Updated 
👉 86e404b3b7c0c9f633b9c49f748bc65811d2186e
👉 0e88a29e6ccb330d899d869631035d66300a8bd4

ykrt/src/trace/swt/patch.rs
Pavel-Durov: Updated 👉 3725c7189fb0159b52a412da595f760370cf2ff4

Pavel-Durov: Sounds good.
But when I remove `PROT_EXEC` I get some of the tests failing in a non-deterministic way 🤔 
I guess its because somehow test execution overlaps and this runtime patching takes effect while other tests try to execute these instructions.
I get the same result when I run the tests sequentially as:
```shell
YKB_TRACER=swt cargo test -- --test-threads=1
```

ykrt/src/trace/swt/patch.rs
Pavel-Durov: Added comments with info 👉 799644e64d38f56c9447bcac350e86ce0b4e6182

ykrt/src/trace/swt/patch.rs
Pavel-Durov: added 👉 0bd8f3617e95057590fafe45a72eecf6cd96a922

ykrt/src/trace/swt/patch.rs
ltratt: I think https://github.com/ykjit/yk/pull/1060/commits/d572cabf3deeef89567a60ca671ac93d08d6743f is meant.

ykrt/src/trace/swt/patch.rs
ltratt: I don't think we want/need https://github.com/ykjit/yk/commit/86e404b3b7c0c9f633b9c49f748bc65811d2186e: what we want to do is `cdg` either the `mod` statement or only those functions which are arch specific. So https://github.com/ykjit/yk/commit/0e88a29e6ccb330d899d869631035d66300a8bd4 is probably not a bad start.

Pavel-Durov: With 86e404b3b7c0c9f633b9c49f748bc65811d2186e I thought to not limit `swt` to x86 so if we run it on other architectures - it will still work just without the runtime patching.
But I'm probably overthinking it cause we support only x86 anyway :)

Pavel-Durov: I think this commit makes it more clear why we need conditional compilation here 👉 b2aee3dd12b8144c7e0f1cf3ab7c8be19029950b

ltratt: > With https://github.com/ykjit/yk/commit/86e404b3b7c0c9f633b9c49f748bc65811d2186e I thought to not limit swt to x86 so if we run it on other architectures - it will still work just without the runtime patching.

Right, but I think the thing to do there would be to guard *inside* those functions, otherwise mt.rs has to know about swt's hardware acceleration (or not). b2aee3dd12b8144c7e0f1cf3ab7c8be19029950b seems reasonable enough to me.

ykrt/src/trace/swt/patch.rs
ltratt: > In fact, on some systems you are not allowed to mark a region writeable and executable at the same time.

I think Lukas's point is that you can `mprotect(PROT_READ | PROT_WRITE)` or `mprotect(PROT_READ | PROT_EXEC)` but you can't combine `PROT_EXEC` and `PROT_WRITE`: you have to write the page and only then mark it as executable.

Pavel-Durov:
Benchmarks completed.

# Benchmarks for JIT (threshold=5)
| YK                      | Range (min … max)    | Time (mean ± σ)     | Runs |
| ----------------------- | -------------------- | ------------------- | ---- |
| swt + runtime-patch     | 3.140 s …  5.266 s   | 3.895 s ±  0.654 s  | 50   |
| swt                     | 3.221 s …  5.163 s   | 4.035 s ±  0.638 s  | 50   |
| swt + nop             | 288.6 ms … 520.1 ms  | 420.4 ms ±  92.6 ms | 50   |
| hwt                     | 3.444 s …  5.503 s   | 4.453 s ±  0.775 s  | 50   |


# Benchmarks - No JIT (threshold=9999999)
| YK                      | Range (min … max)    | Time (mean ± σ)     | Runs |
| ----------------------- | -------------------- | ------------------- | ---- |
| swt + runtime-patch     | 248.4 ms … 257.6 ms  | 252.8 ms ±   2.1 ms | 50   |
| swt + nop              | 44.3 ms …  57.0 ms   | 54.3 ms ±   3.0 ms  | 50   |
| swt                     | 244.9 ms … 268.4 ms  | 259.2 ms ±   4.5 ms | 50   |
| hwt                     | 35.8 ms …  44.5 ms   | 39.4 ms ±   1.0 ms  | 50   |

## Summary

Looks like the runtime patching **doesn't** give the same performance gain as we've seen with nop benchmarks (x4 boost), there's probably another overhead involved.
Or maybe we need to find the right threshold value for `yk_mt_hot_threshold_set` to see the impact?

Command used:
```shell
YKD_SERIALISE_COMPILATION=1 hyperfine -w 2 -m 50 './src/lua ./tests/closure.lua'
```

ltratt:
Let's try this on a few other benchmarks, including those that make longer traces (with short traces we wouldn't necessarily expect to see much difference).

Pavel-Durov:
Disabling trace compilation:
```diff
diff --git ykrt/src/mt.rs ykrt/src/mt.rs
index 8e9f77c4..c02b6420 100644
--- ykrt/src/mt.rs
+++ ykrt/src/mt.rs
@@ -630,11 +630,11 @@ impl MT {
             // spin up a new thread for each compilation. This is only acceptable because a)
             // `SERIALISE_COMPILATION` is an internal yk testing feature b) when we use it we're
             // checking correctness, not performance.
-            thread::spawn(do_compile).join().unwrap();
+            // thread::spawn(do_compile).join().unwrap();
             return;
         }
 
-        self.queue_job(Box::new(do_compile));
+        // self.queue_job(Box::new(do_compile));
     }
 
``` 
Does show the difference as before  - roughly x4:
`yk+swt master`
```shell
Benchmark 1: ./src/lua ./stats/matrix/matrix.lua 50
  Time (mean ± σ):     330.8 ms ±   7.3 ms    [User: 424.2 ms, System: 36.4 ms]
  Range (min … max):   321.6 ms … 337.4 ms    10 runs
```

`yk-swt/tracing-function-runtime-patch`:
```shell
Benchmark 1: ./src/lua ./stats/matrix/matrix.lua 50
  Time (mean ± σ):      1.444 s ±  0.344 s    [User: 1.540 s, System: 0.031 s]
  Range (min … max):    1.102 s …  1.784 s    10 runs
```

ykrt/src/trace/swt/patch.rs
Pavel-Durov: Updated :point_right: 40c0696e76ca685f7334716461d13816e852ac3e
I get that :)
My issue was with failing the test, but I can't reproduce it anymore :no_mouth:

ykrt/src/trace/swt/patch.rs
```diff
@@ -0,0 +1,149 @@
...
+//!
+//! This module performs low-level memory operations and modifies the
+//! execution flow of functions at runtime. Improper use can lead to
+//! undefined behaviour, memory corruption, or crashes.
+
+use libc::{mprotect, size_t, sysconf, PROT_EXEC, PROT_READ, PROT_WRITE};
+use std::mem;
+use std::{ffi::c_void, sync::Once};
+
+#[cfg(tracer_swt)]
```
ltratt, line 27: Are all these `#[cfg]`s necessary?

ykrt/src/trace/swt/mod.rs
```diff
@@ -15,6 +15,8 @@ use std::{
     sync::{Arc, LazyLock},
 };
 
+pub(crate) mod patch;
```
ltratt, line 18: This should *probably* be guarded for only x86?

ykrt/src/trace/swt/patch.rs
Pavel-Durov: not needed. removed :point_right: 5ead8930eb6a4375069233f5aec35775b8c740f2

ykrt/src/trace/swt/patch.rs
Pavel-Durov: I understand that I can't combine `PROT_WRITE` and `PROT_EXEC` but..

I again get this issue with failing tests if I set `PROT_READ | PROT_WRITE` before calling`copy_nonoverlapping` (patching the function):
```shell
Exited due to signal: 11
```
But when I set it as `PROT_READ | PROT_WRITE | PROT_EXEC` it works fine.

ltratt: The normal trick here is: set to `PROT_READ|PROT_WRITE` do the writes; then set it to `PROT_READ | PROT_EXEC`. That way you never have `PROT_WRITE` *and* `PROT_EXEC` set at once.

Pavel-Durov: I understand and I wish it would work for me but I get segfault when I do it.
Will dig deeper.

Pavel-Durov: I think I'm going mad :upside_down_face: 
Its working fine with `PROT_READ | PROT_WRITE` on patch and `PROT_READ | PROT_EXEC` restore.
updated :point_right: 6f4bdc3a3a5d4ac15a130d7e3a877a8f9af062c3

ykrt/src/trace/swt/patch.rs
```diff
@@ -0,0 +1,143 @@
...
+///
+/// * `function_ptr` - A usize representing the memory address of the function to be patched.
+/// * `code` - A constant pointer to a u8 vector where the new instructions are located.
+/// * `size` - A size_t indicating the number of bytes to copy from `code`.
+///
+unsafe fn patch_function(function_ptr: usize, code: *const u8, size: size_t) {
+    let page_size = sysconf(libc::_SC_PAGESIZE) as usize;
+
+    let func_address = ((function_ptr as usize) & !(page_size - 1)) as *mut c_void;
+    let page_size_aligned = (((function_ptr as usize) + mem::size_of_val(&function_ptr))
```
ltratt, line 67: We can use `std::alloc::Layout::align` and friends to do this calculation for us.

I'm also not sure what the difference between `function_ptr` and `func_address` is.

Pavel-Durov: >I'm also not sure what the difference between function_ptr and func_address is. 

My understanding/intention here is that `function_ptr` is just the pointer to the function while `func_address` is the page address (maybe it should be renamed)

Pavel-Durov: updated :point_right: 8e00241271862bd3afdccf0f4c308217938f5098

ltratt: This might get lost but there are some comments on https://github.com/ykjit/yk/commit/8e00241271862bd3afdccf0f4c308217938f5098.

Pavel-Durov: oh I did missed that! sorry

ykrt/src/trace/swt/patch.rs
```diff
@@ -65,27 +65,22 @@ unsafe fn patch_function(function_ptr: usize, code: *const u8, size: size_t) {
...
-        Err(e) => {
-            panic!(
-                "Failed to create layout for fuction instuction patch: {:?}",
-                e
-            );
-        }
+    // This unwrap should be safe since we are using a page that is
+    // based on function_ptr with a known location.
+    let layout = Layout::from_size_align(start_offset, page_size)
+        .expect("Failed to create layout for function memory page");
```
ltratt, line 72: Let's just use `unwrap` as all these `expect`s really bloat the code, especially as they shouldn't ever trigger. [As this suggests we almost never use `expect`: it does have a place, but it's really rare for us.]

ykrt/src/trace/swt/patch.rs
```diff
@@ -70,16 +70,13 @@ unsafe fn patch_function(function_ptr: usize, code: *const u8, size: size_t) {
...
         .expect("Failed to create layout for function memory page");
 
-    // Set function memory page as writable
-    let result = mprotect(page_address, layout.size(), PROT_READ | PROT_WRITE);
-    if result != 0 {
-        panic!("Failed to change memory protection to be writable");
-    }
+    // Set function memory page as writable.
+    // Ignoring mprotect call failure.
+    mprotect(page_address, layout.size(), PROT_READ | PROT_WRITE);
```
ltratt, line 76: We don't want to ignore the return code: we just want to abort changing the machine code. So we need something like `if mprotect(...) { return ... }` and then recover from there.

ykrt/src/trace/swt/patch.rs
Pavel-Durov: Got it. 
updated :point_right: 0428dc1696e940437ea26afd96954097e5316878

ltratt: Can we get rid of the other `expect`s introduced in this PR too please?

ykrt/src/trace/swt/patch.rs
ltratt: Just to check: is just `return` enough for the system to keep running? If so, let's add a comment to that effect.

Alternatively, if we think "if `mprotect` has failed, we've got something wrong", then we might want to `panic` if `mprotect` fails.

ykrt/src/trace/swt/patch.rs
Pavel-Durov: I don't see any other `expect`s only a single `unwrap`.
Am I missing it somehow?

ykrt/src/trace/swt/patch.rs
Pavel-Durov: I think it's ok to return if `mrotect` as it should be transactional and not leave the memory in corrupted state on errro. Unless we consider the "runtime patching" as a must-have for SWT, then we should definitely `panic`.

I tested it with different invalid memory addresses that resulted in `mrotect` returning a non-0 result and the test suite passed.
We still risk changing memory that will not necessarily cause `mrotect` to return with an error but will cause some other part of the process to have invalid memory.

Pavel-Durov: Actually, I think this runtime patching is an integral part of SWT.
Added panic :point_right: 6f0abb36df22b19df4880e289a96538387e9e392

ltratt:
Hmm, I've just realised a possible blocker with this: multi-threading. That's going to require some very careful handling.

Pavel-Durov:
> Hmm, I've just realised a possible blocker with this: multi-threading. That's going to require some very careful handling.

Yes, that would be a problem with instruction patching but I thought we only support serialised compilation in YK?

ltratt:
yk is definitely multi-threaded but we tend to temporarily turn that off because of problems with llvm. When the new JIT codegen is ready, we'll be back to fully multi-threaded.

Let's have a chat about this PR in a day or two. It's still very useful!

ltratt:
@Pavel-Durov I think this PR is probably never going to be resurrected, at least not in the sense of "merged into master". If you agree, I suggest we close this.

PR #1064, opened 2024-03-29
Tidy SWT and improve mapping performance

ltratt:
This PR makes a number of changes, mostly small, to swt. The main ones are:

1. Making `stop` cheap: before this introduced a noticeable pause, particularly the first time it was called (as function names were gathered). This PR gives swt a "genuine" iterator, which means that mapping (including function name gathering) happens in a non-interpreting thread. `stop` is also now very cheap: it replaces a `Vec`, does a couple of cheap checks, and returns. https://github.com/ykjit/yk/commit/ba7fc1b60f5194737084b224699e3569b933b83e
2. Make function name gathering global rather than thread local. Now that mapping is done in a (there might be more than one!) compiler thread, we don't want to do function name gathering more than once. https://github.com/ykjit/yk/commit/b19a63ec4d1e021945484e30c6f55cfc8aaca409

Each commit should be complete and correct in its own right, and it might be easiest to see the story evolve commit-by-commit.


ykrt/src/trace/swt/mod.rs
```diff
@@ -73,77 +100,47 @@ struct SWTTraceRecorder {}
...
+    }
+}
+
+impl Iterator for SWTraceIterator {
+    type Item = Result<TraceAction, AOTTraceIteratorError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.bbs
+            .next()
+            .map(|tb| match FUNC_NAMES.get(&tb.function_index) {
```
Pavel-Durov, line 133: Just to make sure I understand how LazyLock works - is the first lazy initialisation of FUNC_NAMES will be on this line?

ltratt: Yes (meaning: it happens in the compiler thread).

Pavel-Durov: Will there be any benefit in moving this initialisation before tracing starts?

ltratt: >  Will there be any benefit in moving this initialisation before tracing starts?

Yes, possibly! But I haven't benchmarked it (yet) so decided not to. But conceptually we would do `thread::spawn(|| FUNC_NAMES.get())` in `SWTracer::new`. I would only do that if it shows up as a noticable cost, though, as creating threads isn't cheap either.

Pavel-Durov:
LGTM 
tests are passing as`YKB_TRACER=swt cargo test`
please squash

ltratt:
No squashing needed IMHO (all the commits stand alone), so please feel free to merge as-is.

PR #1066, opened 2024-04-02
Safer globalvars

ltratt:
See ykllvm PR https://github.com/ykjit/ykllvm/pull/127.

### DIFF
--- a/ykrt/src/compile/jitc_yk/jit_ir.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir.rs
@@ -158,6 +158,7 @@ const MAX_OPERAND_IDX: u16 = (1 << 15) - 1;
 
 /// The symbol name of the global variable pointers array.
 const GLOBAL_PTR_ARRAY_SYM: &str = "__yk_globalvar_ptrs";
+const GLOBAL_PTR_LEN_SYM: &str = "__yk_globalvar_len";
 
 /// [Instruction] to [InstrIdx] mapping used for stringifying instructions.
 ///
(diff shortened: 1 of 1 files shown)


ltratt:
I could very easily have misunderstood things. But this raises the obvious question: if we have this info elsewhere, why do we have the `__yk_globalvar_ptrs` symbol at all?

vext01:
Because only ykllvm can give us an array of pointers to the globals. During codegen it will emit pointers with the appropriate relocations so that at load-time the pointers become valid for the current address space.

The other method is to find the offsets of the globals in the file and read the PLT to find where the containing object was loaded. But that's pretty much what dlsym does, and we are trying to avoid loader queries.

vext01:
Just to add: statically we know how many globals there are, but we obviously cannot know their dynamic load addresses.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -20,6 +20,27 @@ use ykaddr::addr::symbol_vaddr;
...
+/// present if `YKD_NEW_CODEGEN=1`. When the new codegen becomes the only codegen, we can move from
+/// `dlsym` to:
+///
+/// ```rust
+/// extern "C" {
+///     /// The address of this variable (note: not the contents of the variable!) points to the
+///     /// beginning of an array of `usize`s.
+///     static __yk_globalvar_ptrs: usize;
+///     /// How many elements are there in `__yk_globalvar_ptrs`?
+///     static __yk_globalvar_len: i64;
```
vext01, line 35: I'm OK with the `extern "C"` stuff as long as we are OK with hacking/mocking around it being absent in Rust unit tests.

I still think that `__yk_globalvar_len` can be removed, and then the compiler change isn't needed any more.

ltratt: I tend to agree and I will probably close this PR and its ykllvm companion. Juggling other branches first, though.

PR #1074, opened 2024-04-03
Minor tweaks to `aot_ir.rs`.

ltratt:
This PR contains a number of minor (minor!) tweaks to `aot_ir.rs` after a code read. Of these, only https://github.com/ykjit/yk/commit/b8da221001c4d5002e056c927b7f8ad1f66550e2 is slightly substantive.

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -103,18 +103,12 @@ fn deserialise_into_ti_vec<I, T>(v: Vec<T>) -> Result<TiVec<I, T>, DekuError> {
 
 /// A trait for converting in-memory data-structures into a human-readable textual format.
 ///
-/// This is modelled on [`std::fmt::Display`], but a reference to the module is always passed down
-/// so that constructs that require lookups into the module's tables from stringification have
-/// access to them.
-///
-/// The way we implement this (returning a `String`) is inefficient, but it doesn't hugely matter,
-/// as the human-readable format is only provided as a debugging aid.
+/// This is analogous to [std::fmt::Display], but:
+///   1. Takes a reference to a [Module] so that constructs that require lookups into the module's
+///      tables from stringification have access to them.
+///   2. Returns a [String], for ease of use.
 pub(crate) trait AotIRDisplay {
     /// Return a human-readable string.
-    ///
-    /// FIXME: this isn't as efficient as it could be. We could pass down a mutable string
-    /// reference like `JitIRDisplay` does to avoid all the string allocations when stringifiying
-    /// each constituent IR element.
     fn to_string(&self, m: &Module) -> String;
 
     /// Print myself to stderr in human-readable form.
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -375,7 +375,10 @@ impl Operand {
     /// Returns the [Type] of the operand.
     pub(crate) fn type_<'a>(&self, m: &'a Module) -> &'a Type {
         match self {
-            Self::LocalVariable(_) => self.to_instr(m).type_(m),
+            Self::LocalVariable(_) => {
+                // The `unwrap` can't fail for a `LocalVariable`.
+                self.to_instr(m).type_(m).unwrap()
+            }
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -542,7 +545,7 @@ impl AotIRDisplay for Instruction {
         }
 
         let mut ret = String::new();
-        if self.is_def(m) {
+        if let Some(_) = self.type_(m) {
```
vext01: I don't understand the motivation here. It makes the code less readable, doesn't it?

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -103,18 +103,12 @@ fn deserialise_into_ti_vec<I, T>(v: Vec<T>) -> Result<TiVec<I, T>, DekuError> {
 
 /// A trait for converting in-memory data-structures into a human-readable textual format.
 ///
-/// This is modelled on [`std::fmt::Display`], but a reference to the module is always passed down
-/// so that constructs that require lookups into the module's tables from stringification have
-/// access to them.
-///
-/// The way we implement this (returning a `String`) is inefficient, but it doesn't hugely matter,
-/// as the human-readable format is only provided as a debugging aid.
+/// This is analogous [std::fmt::Debug], but:
```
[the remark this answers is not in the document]
ltratt: True, one could argue this either way.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -377,7 +377,7 @@ impl Operand {
         match self {
             Self::LocalVariable(_) => {
                 // The `unwrap` can't fail for a `LocalVariable`.
-                self.to_instr(m).type_(m).unwrap()
```
vext01: This method is called `type_` because it returns a `Type`, so if you want to rename it to `type_def`, then it should return a `TypeDef`.

But then I'm not sure if we want to rename `Type` to `TypeDef`. That doesn't sit well for some reason.

ltratt: This function is now `is_def` and `type_` in one go: in other words, it returns the type of the definition if there is one.

ltratt: Perhaps `def_type` would be clearer?

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: I guess it's not a big deal.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: That's confusing because `def_type()` is already used for instructions that define a type. An operand cannot define a type.

Taking a step back, what's up with `type_`? Is it the trailing underscore that bothers you?

ltratt: Notice that (before this commit) `type_` can *only* be used if `is_def` returns true.

ykrt/src/compile/jitc_yk/aot_ir.rs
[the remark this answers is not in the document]
vext01: Sorry to be a pain, but since the output is intended for human consumption, and the trait is called `AotIRDisplay` (i.e. contains the word "Display", because it's modelled on `Display`), I think it'd be better to explain in terms of `Display`

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Yes, I think I'm ok with all of that, just perhaps not the name `type_def()`.

ltratt: So `def_type` is better?

vext01: I suggest we keep it `type_`, in line with all of the other `type_` functions on similar types that do the same.

ltratt: Changed in af5864a2.

vext01:
please squash

ltratt:
Squashed.

PR #1076, opened 2024-04-03
Fix trace iteration.

ptersilie:
The code responsible for iterating and processing the blocks in a trace accidentally threw away the first block. The single test we have was working purely by accident.

### DIFF
--- a/ykrt/src/compile/jitc_yk/trace_builder.rs
+++ b/ykrt/src/compile/jitc_yk/trace_builder.rs
@@ -24,6 +24,10 @@ struct TraceBuilder<'a> {
     cp_block: Option<aot_ir::BlockID>,
     // Index of the first traceinput instruction.
     first_ti_idx: usize,
+    // Was the last instruction we've processed a return?
+    last_instr_return: bool,
+    // Was the last block we processed mappable or not?
+    last_block_mappable: bool,
 }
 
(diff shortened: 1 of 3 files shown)


ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -503,13 +511,40 @@ impl<'a> TraceBuilder<'a> {
...
-        let mut trace_iter = ta_iter.peekable();
+
         while let Some(tblk) = trace_iter.next() {
             match tblk {
                 Ok(b) => {
                     match self.lookup_aot_block(&b) {
                         Some(bid) => {
                             // MappedAOTBlock block
+
+                            if bid.is_entry() {
```
vext01, line 522: I don't know if my eyes are wonky, but does `cargo fmt` make this `if` shift left one column?

vext01:
Looks feasible, but I'd have to study the code hard to fully vouch for it.

What do you think @ltratt ?

ykrt/src/compile/jitc_yk/trace_builder.rs
ltratt: It's 4 spaces: time to go to spec savers? :p

vext01: :facepalm:

PR #1077, opened 2024-04-03
Vararg calls.

vext01:
See commits.

Required ykllvm change ~~soon~~: https://github.com/ykjit/ykllvm/pull/128

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64.rs
@@ -6,7 +6,7 @@
 
 use super::{
     super::{
-        jit_ir::{self, InstrIdx, Operand, Type},
+        jit_ir::{self, FuncDecl, InstrIdx, Operand, Type},
         CompilationError,
     },
     abs_stack::AbstractStack,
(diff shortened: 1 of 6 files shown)


ykrt/src/compile/jitc_yk/codegen/x86_64.rs
```diff
@@ -638,6 +642,33 @@ impl<'a> X64CodeGen<'a> {
...
 
+    /// Codegen a (non-varargs) call.
+    pub(super) fn codegen_call_instr(
+        &mut self,
+        inst_idx: InstrIdx,
+        inst: &jit_ir::CallInstruction,
+    ) -> Result<(), CompilationError> {
+        let func_decl_idx = inst.target();
+        let func_type = func_decl_idx.func_type(self.jit_mod);
+        let args = (0..(func_type.num_args()))
```
ptersilie, line 653: Can't we also use `inst.num_args` here? It should be the same as `func_type.num_args()`.

vext01: You have a keen eye :)

We could, but I prefer it this way, as it emphasises that these are static arguments of the function type, not the (potentially dynamic, for a varargs call) arguments of a call-site.

vext01: The type of the instruction differs, so you couldn't do it without adding traits.

PR #1079, opened 2024-04-04
`new` is conventionally the first `fn` in an `impl`.

ltratt:


ltratt:
That's an interesting failure mode: I managed to delete this branch (my mistake) which also cancelled the PR I think! @vext01 if you want to readd this to the queue I'd be grateful (and I'll fix the problem that caused this).

PR #1081, opened 2024-04-04
Inline sub-structs of `enum Operand`.

ltratt:
It now seems fairly clear that most of the `struct *Operand` structs aren't serving a useful structural purpose, in that their variants in `enum Operand` adequately disambiguate things. This PR inlines those structs.


ptersilie:
Sweet.

PR #1085, opened 2024-04-05
Start tweaking the AOT IR

ltratt:
I think these are mostly uncontroversial with the possible exception of https://github.com/ykjit/yk/commit/05e4551a7102a950d7d5ff51706e9b09746d5a63.

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -858,19 +858,6 @@ impl Type {
             Self::Unimplemented(s) => format!("?cst<{}>", s),
         }
     }
-
-    /// Returns a reference to the inner [StructType] if this is a `Self::Struct`.
-    ///
-    /// # Panics
-    ///
-    /// Panics if `self` isn't a `Self::Struct`.
-    pub(crate) fn as_struct(&self) -> &StructType {
-        if let Self::Struct(st) = self {
-            st
-        } else {
-            panic!();
-        }
-    }
 }
 
 impl AotIRDisplay for Type {
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -42,17 +42,11 @@ macro_rules! index {
     };
 }
 
+/// An index into [Module::funcs].
 #[deku_derive(DekuRead)]
 #[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq)]
 pub(crate) struct FuncIdx(usize);
 index!(FuncIdx);
-
-impl FuncIdx {
-    /// Return the [FuncType] for this [FuncIdx] in `m`.
-    pub(crate) fn func_type<'a>(&self, m: &'a Module) -> &'a FuncType {
-        m.func(*self).func_type(m)
-    }
-}
 
 #[deku_derive(DekuRead)]
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -721,7 +721,8 @@ impl FuncType {
             is_vararg,
         }
     }
-    pub(crate) fn arg_ty_idxs(&self) -> &Vec<TypeIdx> {
+
+    pub(crate) fn arg_ty_idxs(&self) -> &[TypeIdx] {
         &self.arg_ty_idxs
     }
 
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -306,7 +306,7 @@ impl Operand {
     /// operands.
     pub(crate) fn to_instr_id(&self) -> InstructionID {
         match self {
-            Self::LocalVariable(iid) => InstructionID::new(iid.func_idx, iid.bb_idx, iid.inst_idx),
+            Self::LocalVariable(iid) => iid.clone(),
             _ => panic!(),
         }
     }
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -100,7 +100,10 @@ impl<'a> TraceBuilder<'a> {
         // If this unwrap fails, we didn't find the call to the control point and something is
         // profoundly wrong with the AOT IR.
         let trace_inputs = trace_inputs.unwrap();
-        let trace_input_struct_ty = trace_inputs.operand(0).type_(self.aot_mod).as_struct();
+        let trace_input_struct_ty = match trace_inputs.operand(0).type_(self.aot_mod) {
```
vext01, line 103: This is a bit laborious, hence the helper.

ltratt: Yes, but the helper *really* confused me, because it's very unobvious that an `Idx` can do such things.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -302,9 +303,9 @@ impl Operand {
         }
     }
 
-    /// Return the `InstructionID` of a local variable operand. Panics if called on other kinds of
-    /// operands.
-    pub(crate) fn to_instr_id(&self) -> InstructionID {
+    /// Return a reference to the instruction iff this operand is a local variable operand. Panics
```
vext01, line 300: Hrm, not sure about this. "Reference" is too overloaded.

"Return a reference" is easily misunderstood to mean return a Rust reference.

A bit of background on the existing name. Ideally it'd be `InstrIdx`, but:
 - a) we already have a type of this name, which is an index local to a basic block.
 - b) it's not really just an index, is it? it's a triple of indices.

Question: What's up with `InstructionID`?

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -83,7 +83,7 @@ impl<'a> TraceBuilder<'a> {
         // PHASE 1:
         // Find the control point call and extract the trace inputs struct from its operands.
         let mut inst_iter = blk.instrs.iter().enumerate().rev();
-        for (_, inst) in &mut inst_iter {
+        while let Some((_, inst)) = inst_iter.next() {
```
vext01, line 86: How about:
```
for (_, inst) in inst_iter {
```

(just delete mut?)

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -470,7 +470,12 @@ impl<'a> TraceBuilder<'a> {
                 args.push(self.handle_operand(arg)?);
             }
             let jit_func_decl_idx = self.handle_func(inst.callee())?;
-            let instr = if !inst.callee().func_type(self.aot_mod).is_vararg() {
+            let instr = if !self
```
vext01, line 473: Again, this makes the code much harder to read at a glance.

The helper serves to lift away the complexity from the call-site, making it easier to comprehend.

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: @ptersilie thoughts?

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: I must admit that I find `while let` harder to read, but I won't die on this hill.

ykrt/src/compile/jitc_yk/trace_builder.rs
ltratt: Oops, sorry, I confused my helpers! The reason for this one is because `as_struct` bypasses the `enum` in a way that just seems weird.

ykrt/src/compile/jitc_yk/trace_builder.rs
ltratt: [I confused this one earlier.]

This helper *really* confused me, because it's the only index type that can return a non-index thing. It doesn't help that we have `Func`, `FuncIdx`, and `FuncType` in the mix too.

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: By "bypasses the enum", do you mean, hides the match?

ltratt: Yes: in order to read the code you need to know it's the enum, read the doc of `as_struct` and... that's actually harder than expanding the code here.

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: >  because it's the only index type that can return a non-index thing

It just makes the IR easier to work with, don't you think? It means we don't have to explicitly convert indices to what they refer to all the time.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: I don't mind a renaming, but `Ref` and "reference" do more harm than good I feel. Sorry.

ykrt/src/compile/jitc_yk/trace_builder.rs
ltratt: This is the only helper on an index and I spent ages baffled so "no" vote from me!

vext01: There are many similar helpers on index types in the JIT IR.

We need to move on. What shall we do:
 - Remove all helpers and inline them at all call-sites
 - Keep the helpers.

ltratt: This helper is bad because it's on an `Idx` type. The other helpers are more a matter of taste.

vext01:
Revert 05e4551a7 and please squash.

ltratt:
Squashed (removing the `Id` -> `Ref` commit, which luckily had no other ramifications).

ltratt:
I don't want to inline all the helpers!

vext01:
Just the ones on the index types, I mean. That was your objection, right?

ltratt:
>  Just the ones on the index types, I mean. That was your objection, right?

As I said: there was only one such helper function. There are no others.

PR #1093, opened 2024-04-09
Start homogenising logging

ltratt:
This PR is scarier than it looks, and it could have been scarier, but I think it's already grown enough for one PR! In essence, this PR tries incrementally to make the way we log things more homogeneous. I didn't fully realise the task I'd taken on here, but the high-level outcomes are:

1. Rename `YKD_PRINT_*` to `YKD_LOG_*`.
2. Allowing `YKD_LOG_*` to take a path as a first argument (rather than, say, "is defined") so that users can log to something other than stderr if they need to.
3. Putting more functionality behind a compile-time feature (which is now renamed `ykd` to reflect the `YKD_*` environment variable; it might better be thought of as `ykd_all` as some `YKD_*` features (notably `YKD_LOG_STATS`) are available without it).

I've also tried to simplify the internal structure / implementation and improve the docs somewhat. There is quite a bit more I could do here, and I might get around to it soon-ish. So while this PR doesn't move us to perfection -- indeed, it introduces a couple of minor rough edges as well as fixing many more -- it does move us in a better direction overall I hope.


ykrt/src/jitstate.rs
```diff
@@ -10,15 +10,19 @@ mod jitstate {
 
 #[cfg(feature = "ykd")]
 mod jitstate {
-    use std::{env, sync::LazyLock};
+    use std::{env, fs, sync::LazyLock};
 
-    static JITSTATE_DEBUG: LazyLock<bool> =
-        LazyLock::new(|| env::var("YKD_LOG_JITSTATE").is_ok());
+    static JITSTATE_DEBUG: LazyLock<Option<String>> =
+        LazyLock::new(|| env::var("YKD_LOG_JITSTATE").ok());
```
vext01: I wonder if the `.ok()`s in this file should be unwraps?

They are very unlikely to fail, but if they do, we'd be confused as to why not logging is showing up.

vext01:
Just that one comment really.

(I'd have preferred "jit-state" over "jitstate", but it doesn't seem worth worrying about)

PR #1094, opened 2024-04-10
Move the top-level `Module` to the top of the source file.

ltratt:
Since `Module` is the most important, and the top-most container, struct in this file, follow convention and move it to the top of the file. It's certainly a lot easier to understand with this ordering than starting with `IntegerType`!


ltratt:
Unusually, github's diff algorithm has really butchered this one --- if you use a better diff algorithm it's much more obvious that this has just moved things from one part of the file!

vext01:
I don't have any real opinion on this. I know that some people prefer types to appear in order of structural depth. By that convention Module would come somewhere near the bottom.

Since the advent of LSPs, I don't really treat files as linear objects any more. I just jump around them with no idea where I actually am in the flie :P

PR #1096, opened 2024-04-15
Fix deoptimisation for inlined calls.

ptersilie:


ykrt/src/compile/jitc_yk/codegen/x86_64.rs
```diff
@@ -64,34 +64,80 @@ extern "C" fn __yk_deopt(frameaddr: *const c_void, deoptid: usize, jitrbp: *cons
...
+                    tmp = unsafe { tmp.byte_add(REG64_SIZE) };
+                }
+                let val = registers[usize::from(*reg)];
+                unsafe { ptr::write(tmp as *mut u64, val) };
             }
         }
 
+        // Now write all live variables to the new stack in the order they are listed in the AOT
+        // stackmap.
         for aotvar in rec.live_vars.iter() {
```
ltratt, line 141: I *think* varidx can be changed to `for (varidx, aotvar) in rec.live_vars.iter().enumerate()` and we can remove the definition at the top of the function and `varidx += 1` below?

ptersilie: Sadly not. We want to increment `varidx` for each variable in each frame. Look where `varidx` is defined: https://github.com/ykjit/yk/pull/1096/files/28eaa49b399e203459f7d0c35c46853b155b83b4#diff-5d16a3250e41186bccaf33f5bf5d3c0a368044a529f1ee60958ee80a93531686R96

I'm not a fan of incrementing `varidx` like this either, and it already lead to a bug earlier because I forgot about a `continue`, which is why it was moved further up. If you have a better idea I'm all ears.

ykrt/src/compile/jitc_yk/codegen/x86_64.rs
```diff
@@ -107,14 +153,16 @@ extern "C" fn __yk_deopt(frameaddr: *const c_void, deoptid: usize, jitrbp: *cons
...
 
             let aotloc = if aotvar.len() == 1 {
                 aotvar.get(0).unwrap()
             } else {
                 todo!("Deal with multi register locations");
             };
             match aotloc {
-                SMLocation::Register(reg, _size, __off, __extra) => {
+                SMLocation::Register(reg, _size, _off, _extra) => {
+                    // FIXME: Deal with additional locations stored in `off` and `extra`.
```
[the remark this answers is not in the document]
ptersilie: I'm afraid not. `off` and `extra` already have values in the `simplecall.newcg.c` test. But writing data to the respective reg/stack doesn't affect the test. So implementing this here doesn't tell me if I got it right. I rather wait until I got a failing test before implementing this.

ptersilie:
Good this failed, since I haven't squashed yet. Looking into it.

ptersilie:
Fixed.

ltratt:
Please squash.

PR #1097, opened 2024-04-17
Give compiled traces a unique integer ID.

ltratt:
In essence this PR fixes this `FIXME`:

> // FIXME: the XXX below should be a thread-safe monotonically incrementing integer.

Note that in testing mode, the ID is meaningless because -- at least so far! -- the JIT IR doesn't need to depend on `MT` and it does make testing easier for that to continue to be the case.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -11,7 +11,7 @@ use std::{
     marker::PhantomData,
     mem,
     sync::{
-        atomic::{AtomicU16, AtomicU32, AtomicUsize, Ordering},
+        atomic::{AtomicU16, AtomicU32, AtomicU64, AtomicUsize, Ordering},
         Arc,
     },
     thread,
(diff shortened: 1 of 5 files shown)


ykrt/src/mt.rs
```diff
@@ -93,6 +93,10 @@ pub struct MT {
     /// The [Compiler] that will be used for compiling future `IRTrace`s. Note that this might not
     /// be the same as the compiler(s) used to compile past `IRTrace`s.
     compiler: Mutex<Arc<dyn Compiler>>,
+    /// A monotonically increasing integer that semi-uniquely identifies each compiled trace. This
+    /// is only useful for general debugging purposes, and must not be relied upon for semantic
+    /// correctness, because the IDs will wrap when the underlying `u64` overflows.
```
ptersilie, line 98: Hah, I would assume you would have long run out of memory to compile further traces before you overflow this. :D

ltratt: Yes, this is paranoia writ large! I even briefly considered making it a `u128` before doing the maths. But, still, since we don't *need* to rely on it for semantic correctness, we might as well document that. That's one footgun our great-great-grandchildren won't encounter ;)

ptersilie: You are very optimistic about the longevity of our software. :stuck_out_tongue:

PR #1104, opened 2024-04-19
Use a known-working version of Rust nightly.

ltratt:
Rust nightly, as of yesterday, breaks yk. We *think* this is because of https://github.com/rust-lang/rust/issues/124150 --- if so, hopefully we can switch back to "nightly nightly" in a day or two.


DianQK:
Could you explain more? #120614 was already landed two weeks ago.

ltratt:
@dianqk Ug, yes, you're right: I misread the merge date on that PR. Sorry!

So I don't really know why rust nightly as of 2 days ago works and current rust nightly doesn't...

DianQK:
So I think this is a separate issue; perhaps we can identify the problem and create an issue for it.

ltratt:
It could well be a bug at our end!

PR #1105, opened 2024-04-19
Make the AOT IR "static".

vext01:
Currently an AOT IR instruction is an opcode and a collection of generic operands. Whilst this is easy to lower to, it isn't easy to work with beyond that. You have to know which operand index is used for what.

This change makes an AOT IR instruction an enum with named, specialised fields. This is more type-safe and should be easier to work with.

Requires a ykllvm change ~~(coming soon)~~: https://github.com/ykjit/ykllvm/pull/133

(I'll point out some areas which require discussion)

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -31,7 +31,7 @@
 
 use byteorder::{NativeEndian, ReadBytesExt};
 use deku::prelude::*;
-use std::{cell::RefCell, error::Error, ffi::CString, fs, path::PathBuf};
+use std::{error::Error, ffi::CString, fs, path::PathBuf};
 use typed_index_collections::TiVec;
 
 /// A magic number that all bytecode payloads begin with.
@@ -44,11 +44,15 @@ const CONTROL_POINT_NAME: &str = "__ykrt_control_point";
 const STACKMAP_CALL_NAME: &str = "llvm.experimental.stackmap";
 const LLVM_DEBUG_CALL_NAME: &str = "llvm.dbg.value";
 
+/// The argument index of the trace inputs (live variables) struct at call-sites to the control
+/// point call.
+const CTRL_POINT_ARGIDX_INPUTS: usize = 2;
+
 // Generate common methods for index types.
 macro_rules! index {
     ($struct:ident) => {
         impl $struct {
-            #[allow(dead_code)] // FIXME: remove when constants and func args are implemented.
+            #[allow(dead_code)]
             pub(crate) fn new(v: usize) -> Self {
                 Self(v)
             }
@@ -108,6 +112,7 @@ pub(crate) struct GlobalDeclIdx(usize);
 index!(GlobalDeclIdx);
 
 /// An index into [FuncType::arg_ty_idxs].
+/// ^ FIXME: no it's not! But it should be!
 #[deku_derive(DekuRead)]
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub(crate) struct ArgIdx(usize);
@@ -182,38 +187,22 @@ impl AotIRDisplay for BinOp {
     }
 }
 
-/// An instruction opcode.
-#[deku_derive(DekuRead)]
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
-#[deku(type = "u8")]
-pub(crate) enum Opcode {
-    Nop = 0,
-    Load,
-    Store,
-    Alloca,
-    Call,
-    Br,
-    CondBr,
-    Icmp,
-    Ret,
-    InsertValue,
-    PtrAdd,
-    BinOp,
-    Unimplemented = 255,
-}
-
-impl AotIRDisplay for Opcode {
-    fn to_string(&self, _m: &Module) -> String {
-        format!("{:?}", self).to_lowercase()
-    }
-}
-
 /// Uniquely identifies an instruction within a [Module].
 #[deku_derive(DekuRead)]
 #[derive(Clone, Debug, Hash, Eq, PartialEq)]
 pub(crate) struct InstructionID {
-    #[deku(skip)] // computed after deserialisation.
-    func_idx: FuncIdx,
+    /// The index of the parent function.
+    ///
+    /// This is computed after deserialisation and relies on
+    /// `Module::compute_local_operand_func_indices() to correctly identify and update all local
+    /// variable operands, setting this field to `Some`. By using `Option` we can detect the case
+    /// where an operand is missed. Without `Option`, deku would insert a default 0 FuncIdx,
+    /// leading to general chaos.
+    ///
+    /// FIXME: I think we can avoid computing this after the fact by having ykllvm include the func
+    /// index in the serialised IR for all local variable operands.
+    #[deku(skip)]
+    func_idx: Option<FuncIdx>,
     bb_idx: BBlockIdx,
     inst_idx: InstrIdx,
 }
@@ -221,7 +210,7 @@ pub(crate) struct InstructionID {
 impl InstructionID {
     pub(crate) fn new(func_idx: FuncIdx, bb_idx: BBlockIdx, inst_idx: InstrIdx) -> Self {
         Self {
-            func_idx,
+            func_idx: Some(func_idx),
             bb_idx,
             inst_idx,
         }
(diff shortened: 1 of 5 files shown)


ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -277,41 +286,20 @@ impl AotIRDisplay for Predicate {
...
-const OPKIND_ARG: u8 = 5;
-const OPKIND_GLOBAL: u8 = 6;
-const OPKIND_PREDICATE: u8 = 7;
-const OPKIND_BINOP: u8 = 8;
-const OPKIND_UNIMPLEMENTED: u8 = 255;
-
 #[deku_derive(DekuRead)]
 #[derive(Debug)]
 #[deku(type = "u8")]
 pub(crate) enum Operand {
```
vext01, line 292: When we discussed this design, we were under the impression that `Operand` would have only two variants:
 - constant
 - local

But alas, we can't really do that. Take for example a call instruction. It has a vector of `Operand` and expects to be able to reference all manner of things like globals, functions (for function pointers), arguments to the parent function (which in our IR are distinct from local variables).

The JIT IR works around this by having special opcodes that "convert" anything not a local or a constant to a local. We could have done the same here too, but since we don't have the same space constraints in the AOT IR, I suggest we just keep the slightly larger Operand enum.

Notice that we were however able to delete a lot of other stuff:
 - type
 - block
 - predicate
 - binop
 - unimplemented

Since those things are not values in llvm, they are not operands in our IR.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -1174,318 +1235,9 @@ pub fn print_from_file(path: &PathBuf) -> Result<(), Box<dyn Error>> {
...
-        d.push(0); // null terminator.
-    }
-
-    /// Note that this test only checks valid IR encodings and not for valid IR semantics. For
-    /// example, nonsensical instruction arguments (incorrect arg counts, incorrect arg types etc.)
-    /// are not checked.
-    ///
-    /// FIXME: implement an IR validator for this purpose.
-    #[test]
-    fn deser_and_display() {
```
vext01, line 1197: I opted to delete this test because:

 - it's a paint to port
 - it always has been a pain to maintain
 - it doesn't test much that isn't already tested
 - it simplifies the operand enum, since we don't need to define `OPKIND_*` constants.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -385,122 +376,160 @@ impl AotIRDisplay for Operand {
...
+    /// This is used when stringifying the instruction.
     ///
-    /// Panics if the instruction isn't a call instruction.
-    pub(crate) fn callee(&self) -> FuncIdx {
-        debug_assert!(matches!(self.opcode, Opcode::Call));
-        let op = self.operand(0);
-        match op {
-            Operand::Func(func_idx) => *func_idx,
-            _ => panic!(),
+    /// FIXME: This is very slow and could be optimised.
```
vext01, line 446: The value being computed here used to be computed and cached (post-serialise) inside every single instruction the defined a value. Since the API user would have to see that fluff, I've opted to remove the cache and compute it every time.

Since it's only used for stringification, I hope that's OK.

vext01: (If we wanted to bring back the cache, we could use a hashmap in the module, but it'd require peppering lifetimes all over `Module` -- I didn't want to junk up this PR with that, and it may never really be an issue anyway)

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -74,12 +78,26 @@ macro_rules! index {
 pub(crate) struct FuncIdx(usize);
 index!(FuncIdx);
 
+impl FuncIdx {
```
ltratt, line 81: Putting these functions on the *indexes* feels really weird (we got rid of this in a recent PR). How difficult would it be to put these functions on `Module`, which is where they belong?

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -182,46 +207,30 @@ impl AotIRDisplay for BinOp {
...
+    /// This is computed after deserialisation and relies on
+    /// `Module::compute_local_operand_func_indices() to correctly identify and update all local
+    /// variable operands, setting this field to `Some`. By using `Option` we can detect the case
+    /// where an operand is missed. Without `Option`, deku would insert a default 0 FuncIdx,
+    /// leading to general chaos.
+    ///
+    /// FIXME: I think we can avoid computing this after the fact by having ykllvm include the func
+    /// index in the serialised IR for all local variable operands.
+    #[deku(skip)]
+    func_idx: Option<FuncIdx>,
```
ltratt, line 225: Agreed: this doesn't feel quite right, but it's probably OK for now.

ykrt/src/compile/jitc_yk/aot_ir.rs
ltratt: The performance of this isn't a problem.

ltratt:
Modulo the index issue (which I think is easily fixed?), this is definitely a step in the right direction: thanks!

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -108,6 +132,7 @@ pub(crate) struct GlobalDeclIdx(usize);
 index!(GlobalDeclIdx);
 
 /// An index into [FuncType::arg_ty_idxs].
+/// ^ FIXME: no it's not! But it should be!
```
ptersilie, line 135: I believe it's currently used for both looking up an argument type in the `FuncType` as well as looking up an argument passed into a function call.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: I will fix this one in a follow up PR. I have a plan.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: The problem is that the vector isn't a `TiVec` yet. I will fix this in a follow-up PR.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: OK, so fa3c3cd3 inverts this logic, but let me show you why I still think this is worse.

Before, you can easily understand what is happening when composing lookups (as happens fairly often) by reading the line lineary left to right. But when we invert the logic it becomes harder to parse because you have to parse from the inside out.

For example (--- before, +++ after):
```diff
- cidx.const_(m).type_idx().type_(m)
+ m.type_(m.const_(*cidx).type_idx())
```

(Note also the extra deref)

And this is exasperated further when the arguments are not trivial:
```diff
- ft.arg_ty_idxs()[usize::from(*arg_idx)].type_(m)
+ m.type_(ft.arg_ty_idxs()[usize::from(*arg_idx)])
```

I think what we had before is a nicer API, but I won't argue further if you still disagree.

ltratt: The original API was equivalent to saying `0.lookup(&vec![])` rather than the more idiomatic `vec![].lookup(0)`, which is why I don't like the original.

In terms of the update, why not give each function a helper on `Module` so you have `func_type(...)` (etc.) rather than just `type_`? You end up with same terseness as the original, but without the unidiomatic API.

vext01: Your way we shall do it ;)

ltratt:
No rebasing yet please: we can still address https://github.com/ykjit/yk/pull/1105#discussion_r1572374374 before doing so. [Or explain why the suggestion in that comment is a bad idea.]

ltratt:
> Fixing it will require me to analyse all the combinations of usage patterns of composed lookups.

If you want to do it in another PR, that's fine. But in theory it's as easy as: temporarily remove `type_` so you catch all calls, then convert e.g. `m.type_(m.func(*func_idx).type_idx)` to `m.func_type(func_idx))`.

ltratt:
Offline discussion results in: please squash.

vext01:
Resolving the conflicts in [this part](https://github.com/ykjit/yk/pull/1105/commits/9bfc015f12fc9ad419613b3337445b0aee3fcc58#diff-0019b0cfa54cfb8c953a80e8ab4e0ce900b5fada3100f573c59c09feadd32fc6L383) was brutal, but I think I got it right.

The tests pass...

PR #1108, opened 2024-04-22
Fix bug in deopt routine.

ptersilie:
Also remove the fixed rust version from .buildbot.sh again which was added as a workaround.


vext01:
LGTM

PR #1114, opened 2024-04-23
Expand binary operations in the JIT IR.

ltratt:
In the AOT IR, compressing these into one seems reasonable, but it now looks likely that:

1. Having separate `*Instruction` structs is a good idea (we can pass them to functions as parameters, which we can't if they're just discriminants in an `enum`).
2. If we have those, we logically also will benefit from having one `*Instruction` struct per binary operation.

Mostly this commit transforms the old `BinaryOperation` struct into a macro. Currently only the `Add` instruction is "fully" (sort of) implemented.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -177,7 +177,6 @@ impl<'a> X64CodeGen<'a> {
         self.comment(self.asm.offset(), inst.to_string(self.jit_mod).unwrap());
 
         match inst {
-            jit_ir::Instruction::BinOp(i) => self.codegen_binop_instr(instr_idx, i),
             jit_ir::Instruction::LoadTraceInput(i) => {
                 self.codegen_loadtraceinput_instr(instr_idx, i)
             }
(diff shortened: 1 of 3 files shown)


vext01:
> In the AOT IR, compressing these into one seems reasonable

What's different about the AOT IR? My gut feeling is that it would make more sense to have the same strategy for both the AOT and JIT IRs unless there is a good reason, but one isn't immediately obvious to me.

> We can pass them to functions as parameters, which we can't if they're just discriminants in an enum

This surprises me a little. We discussed a little while ago and decided against this strategy. Why the change of heart?

FWIW, I think I slightly prefer this approach because having separate structs means you can (as you said) pass the thing around as a whole, but you can also implement different traits on the separate structs. The downside is, more structs = more boilerplate, which often leads to the need for macros.

The only (potential) issue with expanding them like you have is that it uses more of our (limited) opcode space.

I wonder if we could get the best of both worlds with:

```
enum Instruction {
  ...
  BinOp(BinOpInstructionInner)
  ...
}

#[repr(u8)]
enum BinOpInstructionInner {
  Add { lhs, rhs },
  Sub { lhs, rhs },
  ...
}
```

This way all binary operations consume one opcode, but you get the structural integrity that you are aiming for too. The instruction should fit I think: 8 + 8 + 24 + 24 = 64. This could also be applied to the AOT IR I think.

ltratt:
You convinced me that having separate `Instruction` structs is The Right Thing. If so, "compressing" binary ops contradicts this idea, hence expanding it. We *could* make the AOT IR similar, but I think that's unnecessary gold plating right now. The JIT IR is more important because how we structure it affects the optimiser etc.

vext01:
> You convinced me that having separate Instruction structs is The Right Thing. If so, "compressing" binary ops contradicts this idea.

I think they are separate concepts. You could use different structs per-instruction, but still use one unexpanded binop, as demonstrated above. What are you thoughts on that approach?

(I'm not going against your proposal, just checking we've fully thought it through)

vext01:
> So "compressed" bin ops make the optimiser harder.

Note the sub-enum in the example. From the POV of the optimiser, does that not serve the same purpose as using a separate opcode?

PR #1115, opened 2024-04-24
Implement a new outlining algorithm.

ptersilie:
Depends on https://github.com/ykjit/ykllvm/pull/139 (and requires a submodule update once that's in).

The previous algorithm to detect when to start and stop outlining was complicated and required the existence of unmappable blocks in the trace. This put an an extra burden on the software tracer which would have to simulate those unmappable blocks, a sometimes difficult task.

The new algorithm is much simpler and works without unmappable blocks being present. It works as follows:

1) We detect a function that requires outlining. We note the successor block
   `succ` of the current block by looking at the current block's branch
   instruction. When we see this block outlining is finished.
2) Continue processing blocks. For each block check the following:
   a) is this block an entry block to the same function in which we started
      outlining? If yes, then we have recursion: increment a counter.
   b) is this block an exit block (i.e. it contains a return instruction) of
      the same function in which we started outlining? If yes, we are returning
      from a recursive call: decrement the counter.
   c) is the recursion counter zero and this block is the successor block
      `succ`? If yes, then stop outlining.
   d) if none of the above is true, skip this block entirely.

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -366,7 +366,10 @@ pub(crate) enum Instruction {
         args: Vec<Operand>,
     },
     #[deku(id = "5")]
-    Br,
+    Br {
+        /// The block this branch points to.
+        succ: BBlockIdx,
+    },
     #[deku(id = "6")]
     CondBr {
         cond: Operand,
@@ -432,7 +435,7 @@ impl Instruction {
         match self {
             Self::Alloca { .. } => Some(&Type::Ptr),
             Self::BinaryOp { lhs, .. } => Some(lhs.type_(m)),
-            Self::Br => None,
+            Self::Br { .. } => None,
             Self::Call { callee, .. } => {
                 // The type of the newly-defined local is the return type of the callee.
                 if let Type::Func(ft) = m.type_(m.func(*callee).type_idx) {
@@ -530,7 +533,7 @@ impl AotIRDisplay for Instruction {
                 binop.to_string(m),
                 rhs.to_string(m)
             )),
-            Self::Br => ret.push_str("br"),
+            Self::Br { succ } => ret.push_str(&format!("br bb{}", usize::from(*succ))),
             Self::Call { callee, args } => {
                 let args_s = args
                     .iter()
@@ -588,6 +591,13 @@ pub(crate) struct BBlock {
     num_instrs: usize,
     #[deku(count = "num_instrs", map = "map_to_tivec")]
     pub(crate) instrs: TiVec<InstrIdx, Instruction>,
+}
+
+impl BBlock {
+    // Returns true if this block is terminated by a return, false otherwise.
+    pub fn is_return(&self) -> bool {
+        matches!(self.instrs.last().unwrap(), Instruction::Ret { .. })
+    }
 }
 
 impl AotIRDisplay for BBlock {
--- a/ykrt/src/compile/jitc_yk/trace_builder.rs
+++ b/ykrt/src/compile/jitc_yk/trace_builder.rs
@@ -1,6 +1,6 @@
 //! The trace builder.
 
-use super::aot_ir::{self, AotIRDisplay, InstrIdx, Module};
+use super::aot_ir::{self, AotIRDisplay, BBlockId, InstrIdx, Module};
 use super::jit_ir;
 use crate::compile::CompilationError;
 use crate::trace::{AOTTraceIterator, TraceAction};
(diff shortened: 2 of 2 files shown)


ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -530,7 +530,7 @@ impl AotIRDisplay for Instruction {
                 binop.to_string(m),
                 rhs.to_string(m)
             )),
-            Self::Br => ret.push_str("br"),
+            Self::Br { .. } => ret.push_str("br"),
```
vext01, line 533: Can you include the new operand in the stringified instruction?

vext01:
Are there any tests we could port from the old codegen to test this?

ykrt/src/compile/jitc_yk/aot_ir.rs
ptersilie: [11ce067](https://github.com/ykjit/yk/pull/1115/commits/11ce067b6c840c08332b4c99b63137277d90eeba)

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -43,10 +43,10 @@ pub(crate) struct TraceBuilder<'a> {
     last_block_mappable: bool,
     // Inlined calls.
     frames: Vec<Frame<'a>>,
-    // Are we outlining?
-    outlining: bool,
-    // Frame index at which we started outlining.
-    outline_base: usize,
+    // The block at which to stop outlining.
+    outline_target_blk: Option<BBlockId>,
+    // Current stack of recursive calls to the function in which outlining was started.
```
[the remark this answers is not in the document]
ptersilie: [9fd5513](https://github.com/ykjit/yk/pull/1115/commits/9fd5513b43a710928f0c1210a4c319da402688a1)

ptersilie:
> Are there any tests we could port from the old codegen to test this?

Currently it's only meant to replace the existing algorithm in `trace_builder.rs` and not supposed to add functionality. That's coming next when I implement recursion detection.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -530,7 +530,7 @@ impl AotIRDisplay for Instruction {
                 binop.to_string(m),
                 rhs.to_string(m)
             )),
-            Self::Br { .. } => ret.push_str("br"),
+            Self::Br { succ: _succ } => ret.push_str("br"),
```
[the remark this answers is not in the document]
ptersilie: [e3178c0](https://github.com/ykjit/yk/pull/1115/commits/e3178c08bfececc1d0810432356642d0082428f5)

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -366,7 +366,7 @@ pub(crate) enum Instruction {
         args: Vec<Operand>,
     },
     #[deku(id = "5")]
-    Br,
+    Br { succ: BBlockIdx },
```
ltratt, line 369: Can we document what `succ` is? I know what it is now, but I guarantee I'll forget soon :)

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -590,6 +590,13 @@ pub(crate) struct BBlock {
     pub(crate) instrs: TiVec<InstrIdx, Instruction>,
 }
 
+impl BBlock {
+    // Returns true if this block contains a return, false otherwise.
```
ltratt, line 594: Should be `///` (not `//`).

Also I think it would be better to say "if this block is terminated by a return"?

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -37,16 +37,12 @@ pub(crate) struct TraceBuilder<'a> {
...
-    last_block_mappable: bool,
     // Inlined calls.
     frames: Vec<Frame<'a>>,
-    // Are we outlining?
-    outlining: bool,
-    // Frame index at which we started outlining.
-    outline_base: usize,
+    // The block at which to stop outlining.
+    outline_target_blk: Option<BBlockId>,
+    // Current count of recursive calls to the function in which outlining was started.
```
ltratt, line 44: May be worth adding "will be 0 if `self.outline_target_blk` is `None`"?

ltratt:
I only have some minor doc comments. Please feel free to address them in a single commit (or, if you disagree, feel free to do that in one fell swoop too!). You'll probably want to push a commit with the new ykllvm hash at the same time?

ltratt:
I only have some minor doc comments. Please feel free to address them in a single commit (or, if you disagree, feel free to do that in one fell swoop too!). You'll probably want to push a commit with the new ykllvm hash at the same time?

ptersilie:
All done. Ok to squash?

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -366,7 +366,10 @@ pub(crate) enum Instruction {
         args: Vec<Operand>,
     },
     #[deku(id = "5")]
-    Br { succ: BBlockIdx },
+    Br {
+        /// The block this branch points to.
```
ltratt, line 370: I'm a bit confused: `succ` normally means "the block after the current block in the layout" but the comment suggests it's the target branch?

ptersilie: Well, LLVM calls them successor blocks. They are the successor blocks of the current block in terms of control flow, not in terms of their index in the block vector.

ltratt: Ouch, not great terminology, but at least I now know!

ltratt:
Please squash.

ykrt/src/compile/jitc_yk/aot_ir.rs
ptersilie: I can rename it to `target_bb` if you like. I'm not fussed.

ltratt: Using LLVM's terminology here does make sense, so let's keep it.

ptersilie:
Squashed.

PR #1116, opened 2024-04-25
Remove the `JitIRDisplay` trait.

ltratt:
What I slowly realised is that most of this functionality is more concise if implemented on the "top level" `enum`s rather than the itty-bitty `struct`s. Doing this allows us to simplify a lot of the stringification code, both reducing the number of lines of code, and also flattening several layers of indirection.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -13,8 +13,6 @@ use super::{
     reg_alloc::{LocalAlloc, RegisterAllocator, StackDirection},
     CodeGen,
 };
-#[cfg(any(debug_assertions, test))]
-use crate::compile::jitc_yk::jit_ir::JitIRDisplay;
 use crate::compile::CompiledTrace;
 use byteorder::{NativeEndian, ReadBytesExt};
 use dynasmrt::{
@@ -174,7 +172,10 @@ impl<'a> X64CodeGen<'a> {
         inst: &jit_ir::Instruction,
     ) -> Result<(), CompilationError> {
         #[cfg(any(debug_assertions, test))]
-        self.comment(self.asm.offset(), inst.to_string(self.jit_mod).unwrap());
+        self.comment(
+            self.asm.offset(),
+            inst.display(instr_idx, self.jit_mod).to_string(),
+        );
 
         match inst {
             jit_ir::Instruction::LoadTraceInput(i) => {
--- a/ykrt/src/compile/jitc_yk/jit_ir.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir.rs
@@ -10,9 +10,8 @@ use super::aot_ir;
 use crate::compile::CompilationError;
 use num_traits::{PrimInt, ToBytes};
 use std::{
-    error::Error,
     ffi::{c_void, CStr, CString},
-    mem, ptr,
+    fmt, mem, ptr,
 };
 use typed_index_collections::TiVec;
 use ykaddr::addr::symbol_to_ptr;
@@ -238,17 +237,6 @@ impl Module {
         self.instrs.len()
     }
 
-    /// Stringify the module.
-    pub(crate) fn to_string(&self) -> Result<String, Box<dyn Error>> {
-        (self as &dyn JitIRDisplay).to_string(self)
-    }
-
-    /// Print the [Module] to `stderr`.
-    pub(crate) fn dump(&self) -> Result<(), Box<dyn Error>> {
-        eprintln!("{}", self.to_string()?);
-        Ok(())
-    }
-
     /// Returns a reference to the instruction stream.
     pub(crate) fn instrs(&self) -> &Vec<Instruction> {
         &self.instrs
@@ -405,32 +393,24 @@ impl Module {
     }
 }
 
-impl JitIRDisplay for Module {
-    fn to_string_impl(
-        &self,
-        m: &Module,
-        s: &mut String,
-        nums: &LocalNumbers<'_>,
-    ) -> Result<(), Box<dyn Error>> {
-        s.push_str("; compiled trace ID #");
-        s.push_str(&self.ctr_id.to_string());
-
-        s.push_str("\n\n; globals\n");
+impl fmt::Display for Module {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "; compiled trace ID #{}\n\n; globals\n", self.ctr_id)?;
         for g in &self.global_decls {
-            g.to_string_impl(m, s, nums)?;
-            if g.is_threadlocal() {
-                s.push_str("    ; thread local");
-            }
-            s.push('\n');
+            let tl = if g.is_threadlocal() {
+                "    ; thread local"
+            } else {
+                ""
+            };
+            write!(
+                f,
+                "@{}{tl}\n",
+                g.name.to_str().unwrap_or("<not valid UTF-8>")
+            )?;
         }
-
-        s.push_str("\nentry:\n");
-        for (idx, instr) in self.instrs().iter().enumerate() {
-            s.push_str("    ");
-            instr.to_string_impl(m, s, nums)?;
-            if idx != m.len() - 1 {
-                s.push('\n');
-            }
+        write!(f, "\nentry:")?;
+        for (i, instr) in self.instrs().iter().enumerate() {
+            write!(f, "\n    {}", instr.display(InstrIdx::from(i), self))?;
         }
 
         Ok(())
@@ -494,18 +474,6 @@ impl IntegerType {
     }
 }
 
-impl JitIRDisplay for IntegerType {
-    fn to_string_impl(
-        &self,
-        _m: &Module,
-        s: &mut String,
-        _nums: &LocalNumbers<'_>,
-    ) -> Result<(), Box<dyn Error>> {
-        s.push_str(&format!("i{}", self.num_bits()));
-        Ok(())
-    }
-}
-
 /// The declaration of a global variable.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub(crate) struct GlobalDecl {
@@ -546,19 +514,6 @@ impl GlobalDecl {
     }
 }
 
-impl JitIRDisplay for GlobalDecl {
-    fn to_string_impl(
-        &self,
-        _m: &Module,
-        s: &mut String,
-        _nums: &LocalNumbers<'_>,
-    ) -> Result<(), Box<dyn Error>> {
-        s.push('@');
-        s.push_str(self.name.to_str().unwrap_or("<not valid UTF-8>"));
-        Ok(())
-    }
-}
-
 /// Bit fiddling.
 ///
 /// In the constants below:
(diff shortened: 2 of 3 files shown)


vext01:
I'm OK with this as long as you put the `dump()`s back. I use this all the time during development and debugging to quickly inspect various parts of the IR.

The beauty of using the trait was that `dump()` comes for free without additional boilerplate.

ltratt:
I haven't removed `dump`, even though I deeply dislike it, because I don't want that argument again :)

vext01:
It looks like you kept the top-level one but removed it for all other IR structures.

https://github.com/ykjit/yk/pull/1116/files#diff-d66ba8eec98a7d4ee4f268ff4459098b17150480892e69b748f2b50dfd5d2cb9L643

I know you don't like dump, but I really, really to rely on it for quick debugging.

Typing `eprintln("{}", opnd.to_string())` really distracts from what I was doing.

`opnd.dump()` really helps.

ltratt:
I don't see `dump` being removed? I see it being moved but not in an interesting way.

vext01:
I applied the following diff to master:
```diff
diff --git a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
index 7a1dd257..55be52e4 100644
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -353,6 +353,8 @@ impl<'a> X64CodeGen<'a> {
         let lhs = inst.lhs();
         let rhs = inst.rhs();
 
+        lhs.dump(self.jit_mod);
+
         // Operand types must be the same.
         debug_assert_eq!(lhs.type_(self.jit_mod), rhs.type_(self.jit_mod));
```

and cargo check works.

When I apply it to your branch:
```
error[E0599]: no method named `dump` found for enum `jit_ir::Operand` in the current scope
   --> ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs:356:13
    |
356 |         lhs.dump(self.jit_mod);
    |             ^^^^ method not found in `Operand`
    |
   ::: ykrt/src/compile/jitc_yk/jit_ir.rs:910:1
    |
910 | pub(crate) enum Operand {
    | ----------------------- method `dump` not found for this enum
    |
    = help: items from traits can only be used if the trait is implemented and in scope
note: `AotIRDisplay` defines an item `dump`, perhaps you need to implement it
   --> ykrt/src/compile/jitc_yk/aot_ir.rs:146:1
    |
146 | pub(crate) trait AotIRDisplay {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

You removed it from everything apart from `Module`.

ltratt:
Oh, I see, because there was a default impl in `JITIrDisplay`.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -174,7 +172,20 @@ impl<'a> X64CodeGen<'a> {
         inst: &jit_ir::Instruction,
     ) -> Result<(), CompilationError> {
         #[cfg(any(debug_assertions, test))]
-        self.comment(self.asm.offset(), inst.to_string(self.jit_mod).unwrap());
+        {
+            if let Some(dt) = inst.def_type(self.jit_mod) {
```
vext01, line 176: This feels like it's in the wrong place. Shouldn't this be encapsulated inside a struct in the JIT IR?

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -939,22 +793,16 @@ pub(crate) enum Type {
...
-    ) -> Result<(), Box<dyn Error>> {
+impl fmt::Display for Type {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Self::Void => s.push_str("void"),
-            Self::Integer(it) => it.to_string_impl(m, s, nums)?,
-            Self::Ptr => s.push_str("ptr"),
-            Self::Func(ft) => ft.to_string_impl(m, s, nums)?,
+            Self::Void => write!(f, "void"),
+            Self::Integer(it) => write!(f, "i{}", it.num_bits()),
```
vext01, line 800: So if I have just the `IntegerType` I can't print it any more? I'd need to have it wrapped in `Type`?

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -939,22 +793,16 @@ pub(crate) enum Type {
...
+            Self::Ptr => write!(f, "ptr"),
+            Self::Func(_) => todo!(),
```
vext01, line 802: We lost the ability to print a function type?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ltratt: Good question! The basic answer is because an `Instruction` doesn't know its index in the trace. The previous `LocalNumbers` struct was one way around this, but quite complex.

I suppose one way around this would be a method on (probably) `Instruction` called something like `to_string_with_idx` that would do the `%0 = ` bit?

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: Currently we don't use this in any tests so I punted on it until we do...

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: `IntegerType` still implements `Debug`.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
vext01: Hrm. I don't really want to have to think about whether to call `to_string_with_idx` or `to_string`.

ltratt: We *could* I suppose do the old "lookup the instruction index in the `Module`" trick, but it feels really fragile to me. It took me ages to follow through the code to realise how it worked. But I don't feel strongly about it.

ltratt: Let me quickly see if I can knock up a prototype.

vext01:
I'm really not sure about this:

The design goals I was aiming for when implementing stringification (based on experience with working with LLVM IR):

 - 1) make it as easy as possible to stringify IR constructs.
 - 2) any IR construct can be stringified via a uniform interface.
 - 3) An IR construct should look the same whether it is printed in isolation or as part of something larger.

 - we degraded 1 slightly, by moving from `Thing.dump(m)` to `dbg!(Thing.to_string(m))`. I'm OK with this TBH.
 - we lost 2 in this PR. Many constructs lost their `to_string()` and can only be printed as a part of a parent (which may not even exist, if you are in the process of building IR).
 - we lost 3, with `Instruction`, which can only print the lhs of the assignment if printed as part of the module.

Do you agree with the above?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
vext01: Before you spend time on this, check the comment below about the overall approach.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: It's more about debugging. We should be able to print any IR construct in-situ.

ltratt:
I think only point (2) is relevant (I could fix (3)), so let's focus on that, because it's also indicative of something bigger. I don't yet have a firm opinion on it TBH, but here's what I've realised putting together this PR.

The IR is a mix of "semi-dumb" leaf structs and "very unleaf-like functionality". By "leaf struct", I mean things that are just indexes (e.g. an underlying `u16`). However, these leaf structs have methods on them which can reach into outer structs to make them look like they're part of a graph structure.

When we start making those leaf structs' methods reach into outer structs, we end up having to do a lot of computation/storage, that often duplicates things we hold elsewhere. Your point (3) is an example of that: we have the index of an `Instruction`, but it's outside the `Instruction`; so if we want to do super-nice printing of `Instruction`, it has to do a linear search through a `Module` to find its index.

Note that this is different to a more traditional IR which often will keep genuine references to other structs around (I think that's what LLVM does) such that the struct itself is genuinely part of a graph.

My vague intuition is that it's better to keep the functionality in leaf structs leaf-esque. That's what Cranelift seems to do, although I haven't fully explored every part of it. But I don't yet feel super strongly about it: the (fairly significant) code simplification in this PR is suggestive that this approach might be better, but as you point out, there is some loss of functionality too. There are times when shorter, simpler code, is worth a loss in functionality. Is this one of those times? I can't say for sure either way yet.

ptersilie:
>     * we lost 3, with `Instruction`, which can only print the lhs of the assignment if printed as part of the module.

We need to keep this in as it's impossible to debug an error in trace_builder (and later in optimisation passes) where the error is something like `Don't know how to handle 'load %8'`, as it doesn't tell us where this instruction is (it could be in one of many functions or there could be multiple versions inside a single trace). Knowing the LHS, e.g. `%2 = load %8` makes it much easier to work out where the instruction appears and then debug the error.

vext01:
> We need to keep this in as it's impossible to debug an error in trace_builder

That's what I feared.

Too may times, I've got an instruction and the module it came from side-by-side and I'm backwards tracing the data flow. Without the `%x = ` this becomes hostile to debug. Don't you agree?

ltratt:
To be clear: (3) is easily fixed.

vext01:
So I think you are saying you can bring back 2 and 3? If so, great!

ltratt:
No: I'm saying "please ignore (1) and (3)". And then I'm saying "(2) is a much bigger issue than it first looks; thoughts welcome".

vext01:
Isn't 2 just a matter of keeping all the `to_string()`s around, rather than inlining them into their parents?

ltratt:
Putting `to_string` everywhere will remove much (though not all) of the simplification of the PR. Maybe that's what we should do, but I'd be grateful for thoughts on the deeper issue I outlined in https://github.com/ykjit/yk/pull/1116#issuecomment-2076882156.

ltratt:
0a166412 fixes (3).

vext01:
> My vague intuition is that it's better to keep the functionality in leaf structs leaf-esque.

Apart from `Instruction`, what other leaf constructs need to reach outside?

ltratt:
Need? Unclear. But things like `Operand` and `Constant` did/do this too.

ptersilie:
> [0a16641](https://github.com/ykjit/yk/commit/0a1664129df6807499af4d34425ec4a25b3f94e4) fixes (3).

I'm happy with this change.

ltratt:
Looking at Cranelift (e.g. https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/codegen/src/ir/extfunc.rs#L323) made me realise we could (probably should) use a somewhat well-known Rust trick for "things that you want to `Display` but that need extra info to do so". 044e1032 does this, but it might not be obvious at first what's going on.

The basic idea does involve boilerplate, but leads to a surprisingly easy to remember API:

1. Every type implements `Display`.
2. *Except* those types that need "extra information" to display themselves: those types have a function `display(...) -> DisplayableStruct` where `DisplayableStruct: Display` and the `display` function takes in the "extra information".

So for example an `Instruction` needs (at least) its surrounding `Module` to print itself (and, just to emphasise the point, I've also passed in its `InstrIdx` which gets rid of the horror of a `ptr_eq`). So we now have:

```rust
impl Instruction {
    pub(crate) fn display<'a>(
        &'a self,
        instr_idx: InstrIdx,
        m: &'a Module,
    ) -> DisplayableInstruction<'a> {
        DisplayableInstruction {
            instr: self,
            instr_idx,
            m,
        }
    }
}

pub(crate) struct DisplayableInstruction<'a> {
    instr: &'a Instruction,
    instr_idx: InstrIdx,
    m: &'a Module,
}

impl fmt::Display for DisplayableInstruction<'_> {
        if let Some(dt) = self.instr.def_type(self.m) {
            write!(f, "%{}: {dt} = ", self.instr_idx.to_u16())?;
        }
        ...
    }
}
```

Why is this an improvement on what we had before? The nifty thing is that "definitely leaf nodes" directly `Display`, making it obvious that they can't reference parent/outer things. "Stored as leaf nodes but in order to display them nicely they need to access parent/outer things" gain a `display(...)`. method, making it hard to mistake them for "definitely leaf nodes".

It might seem like a minor thing, but I think this is the way out of our "is it a tree or graph?" conundrum: the `display(...)` function makes clear when we're converting from "stored as a tree"  to "compute as part of a graph". As a nice side effect it means that we end up with standard `to_string` methods that don't take paramaters for all types.

@vext01 @ptersilie See what you think. If you think you can tolerate this, I will either a) adjust this PR to follow this thought through for all types b) raise a new PR that does that. [I suspect altering this PR is easier, but I wouldn't know until I tried.]

ptersilie:
Works for me.

ltratt:
https://github.com/ykjit/yk/pull/1116/commits/f0c6acecf62455de22fce32a5734bee4f895ebf0 removes the remaining `to_string` methods in favour of the `Displayable*` idiom. To my surprise, this hasn't added much code, which reassures me that this probably is the right direction for us to go in. If you're OK with this, I'd say this PR can be considered done.

vext01:
Looking good!

> Can you please put back stringification of FuncType?

^ Would it be OK to do this?

> Talking of "across the board", the icing on the cake for the API you propose would be the ability to blindly call display() on any piece of IR, no matter how trivial the struct/enum is. I know we said doing it for indices would be almost silly, but since we use macros to generate index functionality anyway, we may as well sneak a display() in there and offer a consistent API for all.

Any comments on this?

ltratt:
>>   Can you please put back stringification of FuncType?
>^ Would it be OK to do this?

I'd like to wait until we have a need for this, because it's currently unclear to me whether it should go on `Type` or `FuncType` and until we have a use I can't really tell.

ltratt:
>>    Talking of "across the board", the icing on the cake for the API you propose would be the ability to blindly call display() on any piece of IR, no matter how trivial the struct/enum is. I know we said doing it for indices would be almost silly, but since we use macros to generate index functionality anyway, we may as well sneak a display() in there and offer a consistent API for all.
>>Any comments on this?

I'm not keen on this as it would add a *lot* of boilerplate for not much utility AFAICS (though I guess we could use a macro for it). My current rule is "try printing `x` with `println!("{x}")`; if that doesn't work then you need `println!("{}", x.display(...)`.

vext01:
> I'm not keen on this

OK, let's see how we go.

> I'd like to wait until we have a need for this

The need already exists! development/debugging.

ltratt:
> The need already exists! development/debugging.

I don't really agree: we only had one artificial test for it that tested it in isolation. When we actually see it in incoming IR it'll probably be more obvious if we can put it on `Type` or (as before) on `FuncType`. [I'd *like* it to go on `Type` but I can't quite tell if `FuncType` is better yet.]

vext01:
> When we actually see it in incoming IR

It's already present in every JIT module that does a call. [Every function declaration has a function type (via a `TypeIdx`)](https://github.com/ykjit/yk/blob/6b35f06f2f3a2bb24fc7d1203a053cd901ee7ca0/ykrt/src/compile/jitc_yk/jit_ir.rs#L987)

If you really don't want to implement it, I can do it next time I need it.

ltratt:
All I can say is the `todo` isn't hit in this PR! Can you give me an input that does hit it? I'll happily implement it if so!

vext01:
Never mind. Please squash.

ltratt:
Squashed.

PR #1119, opened 2024-04-25
Add DeoptSafepoint instruction 

jacob-hughes:
This solves the issue where deserialisation and handling of calls from within the yk runtime were special-cased to handle LLVM's stack-map intrinsic. This was brittle and created complex nested pattern matching.

Now, calls to LLVM's stack-map intrinsics are serialised to disk as `DeoptSafepoint` instructions, where the live variables are easily retrievable, and useless duplication (such as the function name) is removed.

I think naming this instruction `Stackmap` is a misnomer because a stack-map is just the name of the data structure on disk. I've named it `DeoptSafepoint` because what this instruction really represents is a call which says "at this moment in time, I know the live variables". In other VMs this is usually known as a safepoint -- at least in the GC world! I personally think this fits better, but I don't mind changing it as you know the system better than I do.

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -41,7 +41,6 @@ const FORMAT_VERSION: u32 = 0;
 
 /// The symbol name of the control point function (after ykllvm has transformed it).
 const CONTROL_POINT_NAME: &str = "__ykrt_control_point";
-const STACKMAP_CALL_NAME: &str = "llvm.experimental.stackmap";
 const LLVM_DEBUG_CALL_NAME: &str = "llvm.dbg.value";
 
 /// The argument index of the trace inputs (live variables) struct at call-sites to the control
@@ -434,6 +433,16 @@ pub(crate) enum Instruction {
         /// The resulting type of the operation.
         dest_type_idx: TypeIdx,
     },
+    #[deku(id = "13")]
+    DeoptSafepoint {
+        id: Operand,
+        // FIXME: this is never used, do we need to serialize this?
+        num_shadow_bytes: Operand,
+        #[deku(temp)]
+        num_lives: u32,
+        #[deku(count = "num_lives")]
+        lives: Vec<Operand>,
+    },
     #[deku(id = "255")]
     Unimplemented(#[deku(until = "|v: &u8| *v == 0", map = "map_to_string")] String),
 }
@@ -489,6 +498,7 @@ impl Instruction {
             }
             Self::Store { .. } => None,
             Self::Cast { dest_type_idx, .. } => Some(m.type_(*dest_type_idx)),
+            Self::DeoptSafepoint { .. } => None,
             Self::Unimplemented(_) => None,
             _ => todo!("{:?}", self),
         }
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -406,6 +405,16 @@ pub(crate) enum Instruction {
         binop: BinOp,
         rhs: Operand,
     },
+    #[deku(id = "12")]
```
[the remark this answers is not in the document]
vext01: BTW deku does not warn of re-using the same ID! I should raise an issue.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -406,6 +405,16 @@ pub(crate) enum Instruction {
...
+    DeoptSafepoint {
+        id: Operand,
+        // FIXME: this is never used, do we need to serialize this?
+        num_shadow_bytes: Operand,
```
vext01, line 412: kill it.

vext01:
LGTM with minor comments.

ykrt/src/compile/jitc_yk/aot_ir.rs
ptersilie, line 412: No, we won't need this. We can get rid of it. But given we will merge deopt into the call/br instructions they are attached to soon, there's probably no need to do this now.

ptersilie:
Great, this makes the stackmap stuff much more readable.

We need to squash away that first commit I think cause otherwise git bisect won't work past this point. But there's a comment from Edd that we need to fix first.

ykrt/src/compile/jitc_yk/aot_ir.rs
ptersilie: No, we won't need this. We can get rid of it. But given we will merge deopt into the call/br instructions they are attached to soon, there's probably no need to do this now.

ykrt/src/compile/jitc_yk/aot_ir.rs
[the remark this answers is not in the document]
jacob-hughes: I'll bump this to 13 and push an update once you merge your PR.

jacob-hughes:
Thanks. I don't think I can squash this down just yet anyway. I have to wait until my ykllvm PR merges first.

ltratt:
@jacob-hughes You can merge in your PR (though I don't think force pushing is right yet, as you need to make some code changes right?).

ykrt/src/compile/jitc_yk/aot_ir.rs
jacob-hughes: I will remove this as part of our new changes because removing this would require a new PR to ykllvm.

jacob-hughes:
I'm ready to squash. I need to bump the deku ID but I need to do that as part of the squash because it is a merge conflict with the base branch. Happy for me to do so?

ykrt/src/compile/jitc_yk/aot_ir.rs
ltratt: Works for me.

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #1122, opened 2024-04-26
Rename various things in the JIT IR

ltratt:
The JIT IR was previously rather inconsistent in how it named things. After documenting the "blessed" conventions in https://github.com/ykjit/yk/commit/5baa56c9b719cf5ea66424a60ababd2a9b002fd8, this PR then goes about making things more homogeneous. To my surprise, simply shortening various structs/functions shaves off 100LoC in the repo (mostly in the x86 backend), which suggests that the shorter names really are worthwhile!

This is bound to cause some conflicts for somebody, so it may amuse you to know that due to my own incompetence, I just spent 20 minutes rebasing this against master. I'll never learn...

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/mod.rs
@@ -19,7 +19,7 @@ pub(crate) mod x86_64;
 pub(crate) trait CodeGen<'a> {
     /// Instantiate a code generator for the specified JIT module.
     fn new(
-        jit_mod: &'a jit_ir::Module,
+        m: &'a jit_ir::Module,
         ra: Box<dyn RegisterAllocator>,
     ) -> Result<Box<Self>, CompilationError>
     where
--- a/ykrt/src/compile/jitc_yk/jit_ir.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir.rs
@@ -66,11 +66,11 @@ pub(crate) struct Module {
     /// A [TyIdx] describes an index into this.
     types: TiVec<TyIdx, Ty>,
     /// The type index of the void type. Cached for convinience.
-    void_type_idx: TyIdx,
+    void_ty_idx: TyIdx,
     /// The type index of a pointer type. Cached for convinience.
-    ptr_type_idx: TyIdx,
+    ptr_ty_idx: TyIdx,
     /// The type index of an 8-bit integer. Cached for convinience.
-    int8_type_idx: TyIdx,
+    int8_ty_idx: TyIdx,
     /// The function declaration table.
     ///
     /// These are declarations of externally compiled functions that the JITted trace might need to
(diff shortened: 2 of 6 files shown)


ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -1,6 +1,22 @@
-//! The Yk JIT IR
+//! The JIT's Intermediate Representation (IR).
 //!
-//! This is the in-memory trace IR constructed by the trace builder and mutated by optimisations.
+//! This is the IR created by [trace_builder] and which is then optimised. The IR can feel a little
+//! odd at first because while it fundamentally represents a graph, the in-memory representation is
```
vext01, line 4: Why is it a graph again? It's not to do with control flow, is it? Is it because constants can be shared by multiple instructions?

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -1,6 +1,22 @@
...
+//! use a number of abbreviations/conventions to keep the length of source down to something
+//! manageable (in alphabetical order):
+//!
+//!  * `const_`: a "constant"
+//!  * `decl`: a "declaration" (e.g. a "function declaration" is a reference to an existing
+//!    function somewhere else in the address space)
+//!  * `m`: the name conventionally given to the shared [Module] instance (i.e. `m: Module`)
+//!  * `Idx`: "index"
+//!  * `Instr`: "instructions"
+//!  * `Ty`: "type"
```
vext01, line 19: Surprisingly, I actually don't object to any of these names :)

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: > It's not to do with control flow, is it?

There is an element of this, because of looping, but in practise mostly:

> Is it because constants can be shared by multiple instructions?

Yes, mostly due to this sort of thing.

ltratt:
> I'm not sure that LoC is a metric for "worthwhileness".

Agreed --- but since the new names are, IMHO, equally as clear as the old names, it does seem a partial proxy for usefulness. I don't want to overstate it, though!

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -64,14 +64,14 @@ pub(crate) struct Module {
     consts: TiVec<ConstIdx, Constant>,
     /// The type table.
     ///
-    /// A [TypeIdx] describes an index into this.
-    types: TiVec<TypeIdx, Type>,
+    /// A [TyIdx] describes an index into this.
```
vext01, line 67: Commit message is inaccurate? Should it read "TypeIdx to **TyIdx**"?

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: I wonder if it might be worth clarifying, maybe giving an example? If I wasn't sure, then newcomers won't be sure.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: Oops, you're right, https://github.com/ykjit/yk/pull/1122/commits/668ac40acb8075cddea3f72bb3316bd0e44210f9 is incorrectly titled.

ykrt/src/compile/jitc_yk/codegen/mod.rs
```diff
@@ -19,7 +19,7 @@ pub(crate) mod x86_64;
 pub(crate) trait CodeGen<'a> {
     /// Instantiate a code generator for the specified JIT module.
     fn new(
-        jit_mod: &'a jit_ir::Module,
```
vext01, line 22: I **think** this one is OK *here* in the JIT IR, but in other places where you have both kinds of modules around, it could be very confusing.

Hopefully the fact that you are in `jit_ir.rs` should be enough to know that a `m` is a JIT `m` and not an AOT `m`?

ltratt: > Hopefully the fact that you are in jit_ir.rs should be enough to know that a m is a JIT m and not an AOT m?

Exactly my thinking!

vext01:
I think this is OK.

Head's up @ptersilie, merge conflicts incoming.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: Slightly changed the wording in 8c9f71bc9b47e1046de268d5710bf3a400fdc557 to focus on the most important parts.

ltratt:
I have tried to minimise merge conflicts, I promise!

ptersilie:
Looks good to me.

ltratt:
Squashed and commit message updated.

PR #1123, opened 2024-04-29
Implement code generation for switch instructions.

vext01:
This is a naive codegen that can doubtless be improved, but it will get us going.

Had to change the trace builder's handle_type() to take a reference to a type rather than an index, because switches contain u64 constants whose type may not have a TypeIdx yet.

ykllvm change ~~coming...~~: https://github.com/ykjit/ykllvm/pull/145/files

### DIFF
--- a/ykrt/src/compile/jitc_yk/trace_builder.rs
+++ b/ykrt/src/compile/jitc_yk/trace_builder.rs
@@ -152,7 +152,8 @@ impl<'a> TraceBuilder<'a> {
                         // FIXME: we should check at compile-time that this will fit.
                         match u32::try_from(aot_field_off) {
                             Ok(u32_off) => {
-                                let input_ty_idx = self.handle_type(aot_field_ty)?;
+                                let input_ty_idx =
+                                    self.handle_type(self.aot_mod.type_(aot_field_ty))?;
                                 let load_ti_instr =
                                     jit_ir::LoadTraceInputInst::new(u32_off, input_ty_idx).into();
                                 // If this take fails, we didn't see a corresponding store and the
@@ -239,6 +240,25 @@ impl<'a> TraceBuilder<'a> {
                 } => self.handle_cast(&bid, inst_idx, cast_kind, val, dest_type_idx),
                 aot_ir::Instruction::Ret { val } => self.handle_ret(&bid, inst_idx, val),
                 aot_ir::Instruction::DeoptSafepoint { .. } => Ok(()),
+                aot_ir::Instruction::Switch {
+                    test_val,
+                    default_dest,
+                    case_values,
+                    case_dests,
+                } => {
+                    let sm = &blk.instrs[InstrIdx::new(inst_idx - 1)];
+                    debug_assert!(sm.is_safepoint());
+                    self.handle_switch(
+                        &bid,
+                        inst_idx,
+                        sm,
+                        nextbb.as_ref().unwrap(),
+                        test_val,
+                        default_dest,
+                        case_values,
+                        case_dests,
+                    )
+                }
                 _ => todo!("{:?}", inst),
             }?;
         }
@@ -289,7 +309,7 @@ impl<'a> TraceBuilder<'a> {
         &mut self,
         aot_const: &aot_ir::Constant,
     ) -> Result<jit_ir::Constant, CompilationError> {
-        let jit_type_idx = self.handle_type(aot_const.type_idx())?;
+        let jit_type_idx = self.handle_type(self.aot_mod.type_(aot_const.type_idx()))?;
         Ok(jit_ir::Constant::new(
             jit_type_idx,
             Vec::from(aot_const.bytes()),
@@ -326,18 +346,18 @@ impl<'a> TraceBuilder<'a> {
     }
 
     /// Translate a type.
-    fn handle_type(&mut self, aot_idx: aot_ir::TypeIdx) -> Result<jit_ir::TyIdx, CompilationError> {
-        let jit_ty = match self.aot_mod.type_(aot_idx) {
+    fn handle_type(&mut self, aot_type: &aot_ir::Type) -> Result<jit_ir::TyIdx, CompilationError> {
+        let jit_ty = match aot_type {
             aot_ir::Type::Void => jit_ir::Ty::Void,
             aot_ir::Type::Integer(it) => jit_ir::Ty::Integer(jit_ir::IntegerTy::new(it.num_bits())),
             aot_ir::Type::Ptr => jit_ir::Ty::Ptr,
             aot_ir::Type::Func(ft) => {
                 let mut jit_args = Vec::new();
                 for aot_arg_ty_idx in ft.arg_ty_idxs() {
-                    let jit_ty = self.handle_type(*aot_arg_ty_idx)?;
+                    let jit_ty = self.handle_type(self.aot_mod.type_(*aot_arg_ty_idx))?;
                     jit_args.push(jit_ty);
                 }
-                let jit_retty = self.handle_type(ft.ret_ty())?;
+                let jit_retty = self.handle_type(self.aot_mod.type_(ft.ret_ty()))?;
                 jit_ir::Ty::Func(jit_ir::FuncTy::new(jit_args, jit_retty, ft.is_vararg()))
             }
             aot_ir::Type::Struct(_st) => todo!(),
@@ -354,7 +374,7 @@ impl<'a> TraceBuilder<'a> {
         let aot_func = self.aot_mod.func(aot_idx);
         let jit_func = jit_ir::FuncDecl::new(
             aot_func.name().to_owned(),
-            self.handle_type(aot_func.type_idx())?,
+            self.handle_type(self.aot_mod.type_(aot_func.type_idx()))?,
         );
         self.jit_mod.func_decl_idx(&jit_func)
     }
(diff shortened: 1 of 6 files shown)


ltratt:
> switches contain u64 constants whose type may not have a TypeIdx yet.

Dummy question: how does this occur?

vext01:
In the AOT instruction I encoded all the case integers as raw u64s rather than as generic `Operand`s. I did this so we don't have to match over possibilities that we know cannot happen.

But this also means that we never call `handle_operand()` on them, and thus 64-bit integers may not yet have a type index in the JIT IR yet.

ltratt:
> switches contain u64 constants whose type may not have a TypeIdx yet.

But `switch` doesn't have to take just `u64`s... or does it?

vext01:
An LLVM switch can only operate on integers.

Quoting the language reference:

> The ‘switch’ instruction uses three parameters: an integer comparison value ‘value’, a default ‘label’ destination, and an array of pairs of comparison value constants and ‘label’s. The table is not allowed to contain duplicate constant entries.

For simplicity of implementation, I fetch and use the case values as u64s.

But I wonder if these are bignums, and if so I should add an assertion (to ykllvm) to check they are not bigger than what u64 can express.

ltratt:
I doubt they can be bignums, but I expect they can be arbitrary bit sizes, and signed. The signedness *probably* doesn't matter, but the arbitrary bit size definitely will. At the very least we'd need a big `assert`/`todo` in ykllvm to catch this when it occurs.

vext01:
There's a getter in LLVM's `ConstantInt` that returns an `APInt` (arbitrary precision integer) which can express an arbitrarily large integer, so I think I do need the assertion.

ltratt:
That's not how I'd interpret https://llvm.org/doxygen/classllvm_1_1ConstantInt.html#ab240d0d30dfa9b392ef9d813f3f9e4be. That's more of a "cast" between two internal-to-LLVM types.

I think you just need https://llvm.org/doxygen/classllvm_1_1ConstantInt.html#ab240d0d30dfa9b392ef9d813f3f9e4be.

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -615,12 +642,115 @@ impl<'a> TraceBuilder<'a> {
...
+                        .to_owned()
+                        .make_constant(&mut self.jit_mod, *cv)?;
+                    let jit_const_opnd =
+                        jit_ir::Operand::Const(self.jit_mod.const_idx(&jit_const)?);
+
+                    // Do the comparison.
+                    let jit_test_val = self.handle_operand(test_val)?;
+                    let cmp = jit_ir::IcmpInst::new(
+                        jit_test_val,
+                        jit_ir::Predicate::Equal,
```
ltratt, line 726: I was expecting "not equal" here because that gives a slightly more intuitive notion of why the guard exists. It's not a big deal though.

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -615,12 +642,115 @@ impl<'a> TraceBuilder<'a> {
...
+            None => {
+                // The default case was traced.
+                //
+                // We need a guard that expresses that `test_val` was not any of the `case_values`.
+                // We encode this in the JIT IR like:
+                //
+                //     member = case_val1 || ... || case_valN
+                //     guard member == 0
+                //
+                // OPT: This could be much more efficient if we introduced a special "none of these
```
ltratt, line 706: I don't think a "none of these values" values is a good idea. It's really tricky to do this correctly and it will often bloat memory.

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -615,12 +642,115 @@ impl<'a> TraceBuilder<'a> {
...
+                // values" guard. It could codegen the comparisons into a bitfield and efficiently
+                // bitwise OR a whole word's worth of comparison outcomes at once.
+                //
+                // OPT: Also depending on the shape of the cases you may be able to optimise. e.g.
```
ltratt, line 710: I agree that this is a more plausible optimisation.

vext01:
I'm not sure what you mean. Did you mean to link to the same page twice?

I've just read the `APInt` source code, it's definitely a bignum implementation. [One of the constructors accepts an array of `uint64_t`](https://llvm.org/doxygen/classllvm_1_1APInt.html#a4a46ba6ad1c259b7fa9bc638ebb0a2f8).

So if [switch cases are ConstantInt](https://llvm.org/doxygen/classllvm_1_1SwitchInst.html#a9adb20f50fc2756ec2976d89ff396af2) and [ConstantInt uses APInt internally](https://llvm.org/doxygen/classllvm_1_1ConstantInt.html#af7e1934ed72a405ef073ea5f9bbe828e), then switch cases may be larger than 64-bit.

So until we are willing to take on bignums, the serialiser needs:
```
assert(Case.getCaseValue()->getBitWidth() <= 64);
```

Agree?

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: That's the semantics we have to implement I'm afraid.

If you have:
```
switch x {
case 0:
case 147:
case 5:
case 8888:
    break;
default:
    blah...
}
```

And we trace the default branch, then we must guard that `x` wasn't 0, 147, 5, of 8888.

vext01: Note that this PR already (inefficiently) implements these semantics.

ltratt: ?

ltratt: My point is we can't make a "value that is none of these values". Nowt to do with semantics :)

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: We compare all of the cases that weren't taken, or their 0/1 decisions together and guard that *that* was false.

You could apply demorgan's law to invert the logic, but me and lukas agreed that this way around is less confusing.

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: Yes, but it's not general. It's only feasible if you know the cases are a straight (gapless) run.

ltratt: Obviously, yes.

ltratt:
> switch cases may be larger than 64-bit.

Yes, that was what I was arguing :) This is not quite the same as "big num" though (at least in my mind), but that may just be a terminological thing.

>assert(Case.getCaseValue()->getBitWidth() <= 64);

Yes, this is what I meant.

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -615,12 +642,115 @@ impl<'a> TraceBuilder<'a> {
...
+                let jit_cond = self.jit_mod.push_and_make_operand(cmp_inst.into())?;
+
+                // Guard the result of the comparison.
+                self.create_guard(&jit_cond, bb == next_bb.bb_idx(), sm)?
+            }
+            None => {
+                // The default case was traced.
+                //
+                // We need a guard that expresses that `test_val` was not any of the `case_values`.
+                // We encode this in the JIT IR like:
```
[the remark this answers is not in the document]
vext01: I didn't understand this, but I think we resolved this offline?

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: I didn't understand this, but I think we resolved this offline?

ykrt/src/compile/jitc_yk/trace_builder.rs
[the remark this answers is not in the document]
ltratt: No, this is a bit different. The comment doesn't include the equality check (the actual implementation does!). So, as written it suggests that:

```c
switch ... {
  case 0: ..
  ...
}
```

is translated to `0 || ...` but it actually translates to `input = 0 || ...` which is subtly different.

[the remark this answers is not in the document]
vext01: The check would lower to `res = val == 0 ||... ; guard(!res)`. i.e. we or together the result of a numerical comparison against the case value, not the case value itself.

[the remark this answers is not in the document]
ltratt: Yes, but notice that the `//` comment doesn't include the `==0` bit.

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: Ah! Gotcha!

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -473,9 +497,12 @@ impl<'a> TraceBuilder<'a> {
         ptr: &aot_ir::Operand,
         type_idx: &aot_ir::TypeIdx,
     ) -> Result<(), CompilationError> {
-        let instr =
```
vext01, line 476: I didn't mean to alter this section of code. Will back out.

vext01: Sorry, ignore that, there's a valid change to `handle_type()`.

ykrt/src/compile/jitc_yk/trace_builder.rs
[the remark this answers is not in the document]
vext01: Fixed comment in ec9fbf55cb23406597a1f21a99501366ced66f41

ltratt:
Please squash.

ltratt:
Please squash an ykllvm update at the same time as squashing.

vext01:
rebased (including ykllvm sync)

ltratt:
Please force push an update with a ykllvm update.

vext01:
Done.

PR #1126, opened 2024-04-29
Fold stack-map calls into the instruction they track

jacob-hughes:
Calls to the llvm.experimental.stackmap intrinsic happen before conditional branches, and after regular function calls. This stack-map call encodes a 'safepoint' -- a record of all the live variables so that these instructions can safely deoptimise if a guard fails.

Conceptually, safepoints 'belong' to these instructions, and having to remember whether to go forward or back to find them is just LLVM complexity which doesn't fit with our mental model how AOT IR should look.

This commit refactors this and makes safepoints easier to work with. In essence it does two thing:

1) Removes the `DeoptSafepoint` instruction in the AOT IR, folding safepoint info into the relevant `CondBr` and `Call` instructions.
2) Renames uses of stack-map (an implementation detail) in the AOT IR to safepoint.

### DIFF
--- a/ykrt/src/compile/jitc_yk/trace_builder.rs
+++ b/ykrt/src/compile/jitc_yk/trace_builder.rs
@@ -1,6 +1,6 @@
 //! The trace builder.
 
-use super::aot_ir::{self, AotIRDisplay, BBlockId, FuncIdx, InstrIdx, Module};
+use super::aot_ir::{self, AotIRDisplay, BBlockId, FuncIdx, Module};
 use super::jit_ir;
 use crate::compile::CompilationError;
 use crate::trace::{AOTTraceIterator, TraceAction};
@@ -10,12 +10,12 @@ use std::{collections::HashMap, ffi::CString};
 const TRACE_FUNC_CTRLP_ARGIDX: u16 = 0;
 
 /// A TraceBuilder frame. Keeps track of inlined calls and stores information about the last
-/// processed stackmap, call instruction and its arguments.
+/// processed safepoint, call instruction and its arguments.
 struct Frame<'a> {
     // Index of the function of this frame.
     func_idx: Option<FuncIdx>,
-    /// Stackmap for this frame.
-    sm: Option<&'a aot_ir::Instruction>,
+    /// Safepoint for this frame.
+    safepoint: Option<&'a aot_ir::DeoptSafepoint>,
     /// JIT arguments of this frame's caller.
     args: Vec<jit_ir::Operand>,
 }
@@ -23,10 +23,14 @@ struct Frame<'a> {
 impl<'a> Frame<'a> {
     fn new(
         func_idx: Option<FuncIdx>,
-        sm: Option<&'a aot_ir::Instruction>,
+        safepoint: Option<&'a aot_ir::DeoptSafepoint>,
         args: Vec<jit_ir::Operand>,
     ) -> Frame<'a> {
-        Frame { func_idx, sm, args }
+        Frame {
+            func_idx,
+            safepoint,
+            args,
+        }
     }
 }
 
@@ -200,7 +204,7 @@ impl<'a> TraceBuilder<'a> {
                     self.handle_load(&bid, inst_idx, ptr, type_idx)
                 }
                 // FIXME: ignore remaining instructions after a call.
-                aot_ir::Instruction::Call { callee, args } => {
+                aot_ir::Instruction::Call { callee, args, .. } => {
                     // Get the branch instruction of this block.
                     let nextinst = blk.instrs.last().unwrap();
                     self.handle_call(inst, &bid, inst_idx, callee, args, nextinst)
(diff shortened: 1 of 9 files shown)


ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -227,18 +231,22 @@ impl<'a> TraceBuilder<'a> {
...
-                    let sm = &blk.instrs[InstrIdx::new(inst_idx - 1)];
-                    debug_assert!(sm.is_safepoint());
-                    self.handle_condbr(sm, nextbb.as_ref().unwrap(), cond, true_bb)
+                aot_ir::Instruction::CondBr {
+                    cond,
+                    true_bb,
+                    safepoint,
+                    ..
+                } => {
+                    // let sm = &blk.instrs[InstrIdx::new(inst_idx)];
```
ltratt, line 240: Debugging code.

jacob-hughes: Oops! Fixed here (05031f57ebc11a634e897872cc8c3d95cce179df)

ltratt:
All fine with me but will leave @ptersilie to do a more expert review.

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -10,23 +10,27 @@ use std::{collections::HashMap, ffi::CString};
 const TRACE_FUNC_CTRLP_ARGIDX: u16 = 0;
 
 /// A TraceBuilder frame. Keeps track of inlined calls and stores information about the last
-/// processed stackmap, call instruction and its arguments.
+/// processed safepoint, call instruction and its arguments.
```
ptersilie, line 13: This comment isn't entirely correct anymore. There's no "last processed" stackmap anymore since it's part of the call now. But I'll probably rewrite this whole thing anyway soon, so there's no need to change it now.

ptersilie:
This looks great! However, I'm still a bit unconvinced that renaming stackmaps to safepoint is the right choice, since LLVM still uses the term stackmaps. It stores the stackmaps in the binary, the safepoint ID is actually a stackmap ID to lookup a stackmap in the binary. And we use a stackmap parser to parse it. I don't think "stackmap" is just an implementation detail (even if it should be). At least that's the impression I get from reading the docs: https://llvm.org/docs/StackMaps.html#stackmap-format

jacob-hughes:
Here's where I see the difference in terminology:

- [Stackmap] The data structure in the binary which maps frames to their live variables.
- [Stackmap entry] A single mapping in the stackmap, keyed by the stackmap ID.
- [Safepoint] A place in the code for which there is a **stackmap entry**. A safepoint carries with it a bunch of live variables, and the corresponding stackmap entry tells us what to do with them.

I found references to stackmaps from inside the IR confusing, because it makes it sound like they're a dynamic property and that there are many of them, when actually the stackmap (singular) is just a table. 

> since LLVM still uses the term stackmaps

Personally I don't like that they do this. In fact, in GC API stuff -- where they make heavy use of stackmaps -- they address this confusion, and refer explicitly to the parts in code as _statepoints_. For all intents and purposes, a statepoint is what we call a safepoint, but they go one step further to disambiguate it because safepoints have another meaning in GC. See this snippet from the [docs](https://llvm.org/docs/Statepoints.html#id5):

> Note that the term ‘safepoint’ is somewhat overloaded. It refers to both the location at which the machine state is parsable and the coordination protocol involved in bring application threads to a point at which the collector can safely use that information. The term “statepoint” as used in this document refers exclusively to the former. [1]

Personally I think we should make this change because it disambiguates stackmaps, which I think are quite an overloaded term. I think maybe you're more familiar with LLVM internals so it wasn't as confusing to you as it was to me:)

ltratt:
I'm neutral (i.e. too stupid to understand) on the terminology debate. If we do decide to tweak things then I'd suggest:

1. Go the whole hog (i.e. if we should be using statepoints as a term, let's do that).
2. Document in a doc-string/markdown-file somewhere sensible exactly what terminology we're using and what it means (e.g. a bit like https://github.com/ykjit/yk/commit/172a3428fed631d6a743cabe747e6dc06f26cee2).

ptersilie:
Okay, you've convinced me. I've been calling the concept of storing information about the stack as "a stackmap" for so long that I struggled changing my mindset. But what you say makes sense now. I'm not super sold on "safepoint" ("statepoint" is probably a more accurate term as it describes the state of the stack at a certain point in time, but that's already used for GC), but it will do for now.

jacob-hughes:
I also prefer statepoint, but avoided it so as not to muddle it with the LLVM GC API stuff.

jacob-hughes:
Are you happy for me to squash this? I can address Laurie's comment but I think it makes sense to do it in a different PR as it ends up touching files which were not part of this one.

jacob-hughes:
Rebased and squashed. Can merge once https://github.com/ykjit/ykllvm/pull/147 lands.

jacob-hughes:
Ok this one is ready to go now if you're happy.

jacob-hughes:
This is a bit of a weird one. Locally this works but it seems to fail on CI (and tryci) after going through a test loop a few times. However, there's nothing in it that suggests flakiness (it's just an IR diff error). I have to investigate this one.

ltratt:
It might be a `try_repeat` job...

ptersilie:
The safepoint id has changed.

jacob-hughes:
On some of the test runs, this is the first safepoint in the trace, on others it is the second. Therefore, we must wildcard match on the safepoint ID: 38cea645b50275073cbd6c5d3df9abd6ec336ff4. Let me know if you're happy and I'll squash

ltratt:
@jacob-hughes You *might* be able to use a more specific test here if it's appropriate? If it's just the initial 0/1 that varies you might get away with something like this:

```
switch ${{10_1}}, bb{{bb14}}, [300 -> bb{{bb11}}, 299 -> bb{{bb12}}] [safepoint: ${{safepoint_id}}i64, ($0_0, $0_1, $0_3, $0_4, $0_5, $0_6, $10_1)]
```

ltratt:
Please squash.

jacob-hughes:
Squashed

ptersilie:
Huh, this is interesting that the id alternates. I think iterating over an LLVM module's functions is nondeterministic, which would explain this (i.e. we don't always process functions in the stackmaps adding pass in the same order). But I've never seen this happen in practise until now.

PR #1128, opened 2024-04-30
Vecs to sets

ltratt:
This PR moves several "vecs that are really sets" in the JIT IR to sets. As well as providing better performance, this also caught several duplicate methods and other inconsistencies, which I wasn't really expecting, but which help make the PR a notable code reduction

### DIFF
--- a/ykrt/src/compile/jitc_yk/jit_ir.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir.rs
@@ -72,13 +72,13 @@ pub(crate) struct Module {
     ptr_ty_idx: TyIdx,
     /// The type index of an 8-bit integer. Cached for convenience.
     int8_ty_idx: TyIdx,
-    /// The function declaration table.
+    /// The function declaration pool.
     ///
     /// These are declarations of externally compiled functions that the JITted trace might need to
     /// call.
     ///
     /// A [FuncDeclIdx] is an index into this.
-    func_decls: TiVec<FuncDeclIdx, FuncDecl>,
+    func_decls: IndexSet<FuncDecl>,
     /// The global variable declaration table.
     ///
     /// This is a collection of externally defined global variables that the trace may need to
@@ -141,7 +141,7 @@ impl Module {
             void_ty_idx,
             ptr_ty_idx,
             int8_ty_idx,
-            func_decls: TiVec::new(),
+            func_decls: IndexSet::new(),
             global_decls: TiVec::new(),
             guard_info: TiVec::new(),
             #[cfg(not(test))]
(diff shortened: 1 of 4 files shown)


ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -57,10 +58,10 @@ pub(crate) struct Module {
     ///
     /// An [ExtraArgsIdx] describes an index into this.
     extra_args: Vec<Operand>,
-    /// The constant table.
+    /// The constant pool.
     ///
     /// A [ConstIdx] describes an index into this.
-    consts: TiVec<ConstIdx, Constant>,
```
vext01, line 63: The commit message says:
>  We previously used a `Vec` (well, a `TiVec`) in a manner that was effectively a set (e.g. using `position`)

A set is unordered, so it makes little sense to speak of the position of an element in a set. The constant table is a mapping.

What I think this change conceptually does is makes insertion happen by hashing the value (not the key, as in a traditional mapping), thus removing the need to do searches for pre-existing values upon insertions, right?

If so, great! I just disagree with the commit message.

ltratt:
> A set is unordered

It doesn't *have* to be unordered: that's why we use [`IndexSet`](https://docs.rs/indexmap/2.2.6/indexmap/set/struct.IndexSet.html).

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -303,25 +298,7 @@ impl Module {
...
     /// Panics if the index is out of bounds.
     pub(crate) fn global_decl(&self, idx: GlobalDeclIdx) -> &GlobalDecl {
-        &self.global_decls[idx]
-    }
-
-    /// Get the index of a global, inserting it into the global declaration table if necessary.
-    ///
-    /// `aot_idx` is the [aot_ir::GlobalDeclIdx] index of the global in the AOT module. This is
-    /// needed to find the global variable's address in the global variable pointers array.
-    pub(crate) fn global_decl_idx(
```
vext01, line 355: Surely by not using an AOT index here you are looking up the wrong function address?

ltratt:
Are you talking about the commit message for https://github.com/ykjit/yk/pull/1128/commits/c270c6fbd3aeb68a6f26204e0a8de7ce06d4f61b? Doesn't it say what you're suggesting already?

vext01:
Yep, that one.

I disagree that the constant pool is a set. It's conceptually a mapping.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: Note that `_aot_idx` was not used in this function!

ltratt:
I wrote a response out, but then realised we're in the weeds of pedantry on both sides. I don't think this minor point on a commit message is worth worrying about, but if you think it's important, tell me exactly what words will make you happy, and I'll do it.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Ah right. The AOT index is inside `GlobalDecl`. Yeah I agree, this is just noise then.

vext01:
Too many hills :)

PR #1130, opened 2024-04-30
Remove vacall

ltratt:
In essence this semi-experimental PR flattens `VACall` into `Call`. That does simplify a lot of code, though it places more heap pressure for calls with exactly 1 argument (which now push an `Operand` into `m.extra_args`). My guess is that's such a small change that we couldn't benchmark it if we tried, and given that @ptersilie wants to add indirect calls, simplifying this code is probably worth whatever tiny performance penalty we might have to pay.

@ptersilie I did sort-of hope that I could encode an `enum CallKind` in `Call` but that doesn't work with [`repr(packed)`] on a surrounding struct (see [this SO post](https://stackoverflow.com/questions/77377069/how-to-pack-a-rust-enum-into-its-minimal-size)). So I think you'll want to split this into `DirectCall` (what's currently `Call`) and `IndirectCall`.

### DIFF
--- a/ykrt/src/compile/jitc_yk/jit_ir.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir.rs
@@ -27,7 +27,7 @@ use indexmap::IndexSet;
 use num_traits::{PrimInt, ToBytes};
 use std::{
     ffi::{c_void, CStr, CString},
-    fmt, mem, ptr,
+    fmt, mem,
 };
 use typed_index_collections::TiVec;
 use ykaddr::addr::symbol_to_ptr;
@@ -1298,10 +1298,10 @@ impl LookupGlobalInst {
 pub struct CallInst {
     /// The callee.
     target: FuncDeclIdx,
-    /// The first argument to the call, if present. Undefined if not present.
-    arg1: PackedOperand,
-    /// Extra arguments, if the call requires more than a single argument.
-    extra: ExtraArgsIdx,
+    /// How many arguments in [Module::extra_args] is this call passing?
+    num_args: u16,
+    /// At what index do the contiguous operands in [Module::extra_args] start?
+    extra_idx: ExtraArgsIdx,
 }
 
 impl CallInst {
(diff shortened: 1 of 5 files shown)


ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -1298,10 +1298,10 @@ impl LookupGlobalInst {
...
     /// The callee.
     target: FuncDeclIdx,
-    /// The first argument to the call, if present. Undefined if not present.
-    arg1: PackedOperand,
-    /// Extra arguments, if the call requires more than a single argument.
-    extra: ExtraArgsIdx,
+    /// How many arguments in [Module::extra_args] is this call passing?
+    num_args: u16,
+    /// At what index do the contiguous operands in [Module::extra_args] start?
+    extra_idx: ExtraArgsIdx,
```
vext01, line 1292: This reminds me: Having tripped over the meaning of this field a couple of times in the past, I wanted to rename it to `first_extra_arg_idx` (or similar), but actually what you are proposing here opens us to even better naming.

If no call opcode has inline args any more, then in a sense every arg is "extra" and we may as well kill the "extra" terminology altogether.

So (maybe in another PR) we can rename anything `ExtraArg`-related to just `CallArg`? Then `extra_idx` in this strict becomes `first_arg_idx` (or similar)?

What do you think? Like I said, we don't have to do this now.

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -1298,10 +1298,10 @@ impl LookupGlobalInst {
 pub struct CallInst {
     /// The callee.
     target: FuncDeclIdx,
-    /// The first argument to the call, if present. Undefined if not present.
```
vext01, line 1301: Looks like we could even drop the `#[repr(packed)]` now too?

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: I agree that this can be improved, and it can be done in another PR if we want.

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: Unfortunately doing so tips is into 8, not 7, bytes.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: Oh one more thing that we may want to think about. For a non-vararg call we could even remove the `num_args` field as it's already known from the function signature.

I guess you knew that and decided to cache it here? That's fine, I don't mind much either way.

vext01: Sorry, I've just seen that the next commit merges this code with vararg calls, so it makes complete sense to store the number of args.

ptersilie:
Looks good to me and it will make adding indirect calls easier.

PR #1132, opened 2024-05-01
Codegen PHI nodes.

vext01:
For this we introduce a new JIT opcode "assign", which is used to select the correct incoming value at trace construction time.

ykllvm PR ~~coming soon~~: https://github.com/ykjit/ykllvm/pull/148


ltratt:
Can we/should we have a test which unrolls loops? That will also use phi nodes and might be a useful test. If it doesn't yet work, then it might be worth making it the subsequent PR(s)?

vext01:
> Can we/should we have a test which unrolls loops? 

No idea if that works or not, but I can investigate.

ltratt:
@vext01 Shall I kick this one off for a merge anyway? We can fix things in a follow-up PR if necessary.

vext01:
Sure, if that's OK by you.

PR #1135, opened 2024-05-02
Remove more `type_` methods in the JIT IR.

ltratt:
This is, I believe, the final push: there is now just `Method::_type` left in the JIT IR.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -271,7 +271,10 @@ impl<'a> X64CodeGen<'a> {
         let rhs = inst.rhs();
 
         // Operand types must be the same.
-        debug_assert_eq!(lhs.type_(self.m), rhs.type_(self.m));
+        debug_assert_eq!(
+            self.m.type_(lhs.ty_idx(self.m)),
+            self.m.type_(rhs.ty_idx(self.m))
+        );
 
         self.load_operand(WR0, &lhs); // FIXME: assumes value will fit in a reg.
         self.load_operand(WR1, &rhs); // ^^^ same
@@ -292,7 +295,10 @@ impl<'a> X64CodeGen<'a> {
         let rhs = inst.rhs();
 
         // Operand types must be the same.
-        debug_assert_eq!(lhs.type_(self.m), rhs.type_(self.m));
+        debug_assert_eq!(
+            self.m.type_(lhs.ty_idx(self.m)),
+            self.m.type_(rhs.ty_idx(self.m))
+        );
 
         self.load_operand(WR0, &lhs); // FIXME: assumes value will fit in a reg.
         self.load_operand(WR1, &rhs); // ^^^ same
@@ -413,7 +419,7 @@ impl<'a> X64CodeGen<'a> {
             let op = &args[i];
             // We can type check the static args (but not varargs).
             debug_assert!(
-                i >= fty.num_args() || op.type_(self.m) == fty.arg_type(self.m, i),
+                i >= fty.num_args() || self.m.type_(op.ty_idx(self.m)) == fty.arg_type(self.m, i),
                 "argument type mismatch in call"
             );
             self.load_operand(reg, op);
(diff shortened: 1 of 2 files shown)


ltratt:
Whoops, I pressed that on the wrong repo!

ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -811,24 +811,6 @@ impl Operand {
         }
     }
 
-    /// Returns the type of the operand.
```
vext01, line 814: What's the motivation for removing this? I don't mind it. It makes finding the types of things shorter.

ltratt: It's part of the general "make the user aware that the IR is a tree and you should get info from `Module`" push. Note that the methods removed in this PR either 1) weren't used 2) entirely duplicated existing `ty_idx` functions.

vext01: I'm kinda still luke warm on this.

It hurts my head more to ask "whats the type index of this, and what's the type for this type index" instead of "what the type of this".

A compromise would be to make the function simply do: " self.m.type_(op.ty_idx(self.m))" -- that big match isn't necessary.

vext01: > make the user aware that the IR is a tree

Dumb question: why does the API user need to care about this? Abstraction vs. implementation details, no?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -413,7 +419,7 @@ impl<'a> X64CodeGen<'a> {
             let op = &args[i];
             // We can type check the static args (but not varargs).
             debug_assert!(
-                i >= fty.num_args() || op.type_(self.m) == fty.arg_type(self.m, i),
+                i >= fty.num_args() || self.m.type_(op.ty_idx(self.m)) == fty.arg_type(self.m, i),
```
vext01, line 422: Should a follow up commit rename `type_` to `ty`, as per our new conventions?

ykrt/src/compile/jitc_yk/jit_ir.rs
ltratt: Well we've done it everywhere else... I think it's better that we say "return indexes" consistently now. Or, if we want to have helper methods, to think carefully about why/when. Just having one or two left in feels like the worst of all worlds to me, I must admit.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ltratt: Good point. Yes!

ltratt: Well, follow-up PR.

ykrt/src/compile/jitc_yk/jit_ir.rs
vext01: OK.

Initially I'd aimed to make the API abstracted, instead of exposing the underlying data structures, but I guess you can't really do a full job of abstraction anyway.

ltratt: > Dumb question: why does the API user need to care about this? Abstraction vs. implementation details, no?

Good question! Normally I agree that you want to hide implementation details, but in this case we want to expose them, because it's *really* easy to treat the IR as if it's stored as a graph. That causes us to put lots of spaghetti-helpers everywhere that are extremely hard to follow. An extreme example of this was highlighted by https://github.com/ykjit/yk/pull/1116. Basically it's better, IMHO, to have the IR structs be as "dumb" as possible, making the user remember that `Module` actually stores all the interesting bits. This causes a bit more boilerplate but since it's very uniform, it actually causes less cognitive burden in the long run (IMHO).

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
vext01: OK

PR #1137, opened 2024-05-02
A more consistent Clippy approach

ltratt:
This PR does three main things:

1. An audit for no-longer-needed Clippy hints.
2. Fix Clippy warnings, many of which have crept into the code since the last Clippy PR 10 days (!) ago https://github.com/ykjit/yk/pull/1111.
3. Experiment with a way of stopping so many Clippy things slipping into the code.

The problem with (3) is that the people who do (2) are mostly not the people introducing things that need (3). In other words, Clippy isn't helping all of us equally, and some folk are introducing the same easily-spotted-by-Clippy issues repeatedly.

https://github.com/ykjit/yk/commit/5212b9bab59b0baeb7c0a5c9e90eba80c8a70daa thus *enforces* Clippy from now on: it uses [cargo-diff-tools](https://github.com/fpoli/cargo-diff-tools/) to raise an error when a PR introduces a change which triggers a Clippy hint. Note that this is slightly clever: if Clippy changes (which it does, and often!) this will suppress warnings on code not changed by a PR. So it will encourage us not to merge code that obviously violates Clippy.

Does this commit have the potential to be annoying? Definitely! Maybe too annoying? Possibly. If it is too annoying, we can revert this commit later. But I think it's worth trying: I find that perhaps 80% of Clippy lints actively improve code; 10% are neutral; and 10% are wrong. We can always suppress that latter category with `allow(...)` if we need to.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -20,6 +20,8 @@ export PATH="${CARGO_HOME}"/bin/:"$PATH"
 
 rustup toolchain install nightly --allow-downgrade --component rustfmt
 
+# Formatting problems are frequent in PRs, and easy to fix, so try and catch
+# those before doing anything complicated.
 cargo fmt --all -- --check
 
 # We now need a copy of ykllvm. Building this is quite slow so if there's a
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -79,13 +79,9 @@ YKLLVM_BIN_DIR=$(pwd)/../inst/bin
 export YKB_YKLLVM_BIN_DIR="${YKLLVM_BIN_DIR}"
 cd ../../
 
-# Check that clang-format is installed.
-PATH=${YKB_YKLLVM_BIN_DIR}:${PATH} clang-format --version
-# Check C/C++ formatting using xtask.
+# Check C/C++ formatting. First we run `xtask cfmt`...
 PATH=${YKB_YKLLVM_BIN_DIR}:${PATH} cargo xtask cfmt
-# This is used to check clang-tidy output, but the dirty submodule from building
-# ykllvm is also shown.
-# FIXME: Add build/ to .gitignore in ykllvm
+# ... and then we see if it caused any changes (i.e. caused the git repo to become dirty).
 git diff --exit-code --ignore-submodules
 
 for tracer in ${TRACERS}; do
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -14,11 +14,9 @@ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > rustup.sh
 sh rustup.sh --default-host x86_64-unknown-linux-gnu \
     --default-toolchain nightly \
     --no-modify-path \
-    --profile minimal \
+    --profile default \
     -y
 export PATH="${CARGO_HOME}"/bin/:"$PATH"
-
-rustup toolchain install nightly --allow-downgrade --component rustfmt
 
 # Formatting problems are frequent in PRs, and easy to fix, so try and catch
 # those before doing anything complicated.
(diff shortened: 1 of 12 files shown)


.buildbot.sh
```diff
@@ -83,10 +83,6 @@ cd ../../
 PATH=${YKB_YKLLVM_BIN_DIR}:${PATH} clang-format --version
 # Check C/C++ formatting using xtask.
 PATH=${YKB_YKLLVM_BIN_DIR}:${PATH} cargo xtask cfmt
-# This is used to check clang-tidy output, but the dirty submodule from building
-# ykllvm is also shown.
-# FIXME: Add build/ to .gitignore in ykllvm
-git diff --exit-code --ignore-submodules
```
vext01, line 87: Does `cargo xtask cfmt` exit non-zero if there were formatting changes then?

ltratt: What's that got to do with `git diff`?

ltratt: I really don't get how `git diff` on its own in CI can do anything useful.

ltratt: How can there be unstaged changes in CI?

vext01: `cargo xtask cfmt` does in-place changes if formatting is required.

ltratt: OK so I was completely mislead by the comment. Fixed in 325bf652.

[I also removed the `clang-format` check which is pointless and also very slightly mislead me.]

vext01: OK.

I'm not 100% sure why the submodule would be dirty, but let's assume it could happen :)

vext01:
Please squash.

ltratt:
Squashed.

.buildbot.sh
vext01: Yeah, but the `ykllvm` dir is explicitely excluded from the formatting run:
https://github.com/ykjit/yk/blob/9f01171281f4d5114a0fa7d12363f9a81102f6b6/xtask/src/main.rs#L10

vext01: Well it doesn't hurt to exclude the submodule I guess.

ltratt:
```
cargo-clippy-diff master -- --all-features -- -D warnings
13:04:10 fatal: ambiguous argument 'master': unknown revision or path not in the working tree.
13:04:10 Use '--' to separate paths from revisions, like this:
13:04:10 'git <command> [<revision>...] -- [<file>...]'
13:04:10 Error: `git diff` terminated with exit status 128
```

@vext01 Hmm, OK this problem did worry me a bit: I'm not quite sure what the right branch to compare against is. Do you know what branches are visible inside a CI job?

ltratt:
OK try merging this one and we can see.

ltratt:
OK to squash?

ltratt:
We won't know if it worked until we go further... Let's try!

ltratt:
@vext01 You were implicitly right: I should have tried something more debugging-y first. The new commit I've pushed will exit even if it's successful so we can avoid squashing even if we go for many tries. Please try "mergnig" (knowing that it will actually fail)?

vext01:
Not seeing any new commits since the last failure.

But anyway, you may as well use tryci for this until it works.

ltratt:
Oops, now actually pushed. Unfortunately `tryci` isn't useful here, because it's copying a local copy of the rep, so in this one regard it's actually different to buildbot.

ltratt:
See what you think of 2943659c.

ltratt:
Oops and https://github.com/ykjit/yk/pull/1137/commits/ba4cd62b1640f2510072b74fe6f5a9ba73b04962.

vext01:
Seems reasonable. Do you want to squash?

ltratt:
Squashed.

ltratt:
@vext01 I'm going to do something and ask your forgiveness in advance: I'm going to "merge" something where I've put an `exit` in to buildbot.sh so it can't succeed. I think I'm going to have to do several commits to work out what's really going on here.

vext01:
Please squash

ltratt:
Squashed.

PR #1140, opened 2024-05-05
Clippy now wants to know all the `cfg` flags we've added.

ltratt:
Without specifying these -- and, where they have values, their values -- Clippy now generates warnings. Unfortunately, adding these in means that one old Buildbot check no longer works (it always fails because "yk_testing" is now always part of Cargo's output). So we have to forego those.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -106,14 +106,6 @@ for tracer in ${TRACERS}; do
 
     # Error if Clippy detects any warnings introduced in lines changed in this PR.
     cargo-clippy-diff origin/master -- --all-features -- -D warnings
-
-    # There are some feature-gated testing/debugging switches which slow the JIT
-    # down a bit. Check that if we build the system without tests, those features
-    # are not enabled.
-    cargo -Z unstable-options build --build-plan -p ykcapi | \
-      awk '/yk_testing/ { ec=1 } END {exit ec}'
-    cargo -Z unstable-options build --build-plan -p ykrt | \
-      awk '/yk_testing/ { ec=1 } /yk_jitstate_debug/ { ec=1 } END {exit ec}'
 done
 
 # Run the tests multiple times on hwt to try and catch non-deterministic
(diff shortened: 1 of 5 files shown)


vext01:
After a little reading, there's a couple of alternatives that would let us keep the `yk_testing` check:

 - Disable the lint (`#[allow(unexpected_cfgs)]`).
 - Use features instead of cfgs.
 - Parse the build plan smarter (with `jq` maybe)?

It's not clear to me why this change causes the `yk_testing` feature to be pulled in.

Also `cargo -Z unstable-options build --build-plan -p ykcapi |awk '/yk_testing/ { ec=1 } END {exit ec}` exits with status 1 on master, before any of your changes, so I'm not really sure that your changes have caused this.

Could it be that the new rust actually is pulling in the `yk_testing` feature when we don't want it to?

ltratt:
> It's not clear to me why this change causes the yk_testing feature to be pulled in.

It isn't pulling it in, but our previous "grep the output" check now finds "oh, `yk_testing` is a feature that *can* be enabled".

vext01:
> It isn't pulling it in, but our previous "grep the output" check now finds "oh, yk_testing is a feature that can be enabled".

Understood, however, it appears unrelated to your changes, right? Just `rustup update` caused that to happen, even on master?

If so, yeah, we probably can't keep the checks in unless we were to parse the json better.

ltratt:
> Understood, however, it appears unrelated to your changes, right?

Not exactly. We are forced to make the changes in this PR: those changes invalid our previous grep/awk check.

vext01:
>  those changes invalid our previous grep/awk check.

But it happens on master too after `rustup update`.

Let's push on. We can potentially bring back this check later. It's semi-important.

vext01:
```
15:47:34 + awk /yk_testing/ { ec=1 } /yk_jitstate_debug/ { ec=1 } END {exit ec}
15:47:34     Blocking waiting for file lock on package cache
15:47:35     Blocking waiting for file lock on package cache
15:47:41 + status=1
```

A timeout with parallelism?

ltratt:
Mea culpa. OK to squash?

vext01:
Squash away.

ltratt:
Squashed.

PR #1145, opened 2024-05-10
Start "testing" of variants of instructions we don't yet support.

vext01:
I'm just ensuring they are lowered to the unsupported instruction opcode to prevent silent miscompilation.

In a future PR, I will audit all of the LLVM instructions that we lower, looking for things we should do the same with.


vext01:
This requires https://github.com/ykjit/ykllvm/pull/153

vext01:
This is ready to go now @ptersilie

PR #1146, opened 2024-05-13
`let ... else` is shorter in most cases than `_ => panic`.

ltratt:


ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -296,12 +296,10 @@ impl Operand {
...
     /// OPT: This is expensive.
     pub(crate) fn to_instr<'a>(&self, aotmod: &'a Module) -> &'a Instruction {
-        match self {
-            Self::LocalVariable(iid) => {
-                &aotmod.funcs[iid.func_idx].bblocks[iid.bb_idx].instrs[iid.inst_idx]
-            }
-            _ => panic!(),
-        }
+        let Self::LocalVariable(iid) = self else {
+            panic!()
```
vext01, line 300: Does rustfmt allow this to be a one-liner even?

ltratt: This is after rustfmt has done its thing so unfortunately "no".

PR #1147, opened 2024-05-13
Remove `num_args` vars

ltratt:


ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -428,9 +428,8 @@ impl<'a> X64CodeGen<'a> {
         // FIXME: floating point args
         // FIXME: non-SysV ABIs
         let fty = self.m.func_type(func_decl_idx);
-        let num_args = args.len();
-
-        if num_args > ARG_REGS.len() {
+        debug_assert_eq!(fty.num_args(), args.len());
```
vext01, line 431: I'm questioning this assertion. Is this code path used for varargs calls?

ltratt: Dunno. The main test passes...

vext01: Either:
 - The assertion is correct and we can delete code that deals with varargs.
 - The assertion is incorrect and test coverage isn't good enough.

I can't merge this in good faith without deeper investigation.

ltratt: FWIW I think that putting an a "too tough" assertion that passes our test suite is not a bad thing: someone who comes up with a future test that invalidates the assertion will know to fix it. I'd argue that this assertion is therefore good as-is, but it's up to you!

vext01: I'd rather investigate now. I'll do the work :)

vext01: > The main test passes...

After checking out your branch and running tests, many fail because they can no longer codegen calls to `printf` and `fprintf`.

The assertion is incorrect, and I also believe these comments to be incorrect is saying they only handle non-varargs calls:
 - https://github.com/ykjit/yk/blob/a277d04582a4b43edc0e99e0b74b6bf0f86ffc0c/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs#L476
 - https://github.com/ykjit/yk/blob/a277d04582a4b43edc0e99e0b74b6bf0f86ffc0c/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs#L489

I'm gonna CC @ptersilie though, because he knows this section of code better than I do.

ptersilie: > because he knows this section of code better than I do

Do I? Well in this case I agree that this `debug_assert` won't work, because this function also handles vararg calls. And in those cases the argument number of the func type may not be the same as the arg number of the call.

ltratt: I'm an idiot and forgot to set `YKD_NEW_CODEGEN=1`. 1 week away...

ltratt: Assertion fixed in 88ae1d74.

vext01: Can't see any new commits. Did you push?

ltratt: Which comment are you referring to? This one?

```rust
// We can type check the static args (but not varargs).
```

vext01: The two comments linked in https://github.com/ykjit/yk/pull/1147#discussion_r1598255400

ltratt: Got it! Fixed in 9b36f4ba.

vext01:
Thanks, please squash.

ltratt:
Squashed.

PR #1148, opened 2024-05-13
Simplify push methods

ltratt:
This simplifies a number of `push_` methods in the JIT IR.

### DIFF
--- a/ykrt/src/compile/jitc_yk/jit_ir.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir.rs
@@ -188,26 +188,14 @@ impl Module {
 
     /// Push an instruction to the end of the [Module].
     pub(crate) fn push(&mut self, inst: Inst) -> Result<InstIdx, CompilationError> {
-        match InstIdx::new(self.insts.len()) {
-            Ok(x) => {
-                self.insts.push(inst);
-                Ok(x)
-            }
-            Err(e) => Err(e),
-        }
+        InstIdx::new(self.insts.len()).inspect(|_| self.insts.push(inst))
     }
 
     pub(crate) fn push_indirect_call(
         &mut self,
         inst: IndirectCallInst,
     ) -> Result<IndirectCallIdx, CompilationError> {
-        match IndirectCallIdx::new(self.indirect_calls.len()) {
-            Ok(x) => {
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/jitc_yk/jit_ir.rs
```diff
@@ -253,9 +241,7 @@ impl Module {
     ///
     /// If `args` would overflow the index type.
     fn push_args(&mut self, args: Vec<Operand>) -> Result<ArgsIdx, CompilationError> {
-        let idx = self.args.len();
-        self.args.extend(args);
-        ArgsIdx::new(idx)
+        ArgsIdx::new(self.args.len()).inspect(|_| self.args.extend(args))
```
ptersilie, line 237: This may be shorter, but I find it more difficult to think about. The old version is quite clear: Extend args, but return the previous length. In this new version I find this harder to see. But I'm not going to die on this hill if you think this is still better.

ltratt: The downside to this change is that you need to know what `inspect` does, but it's a useful function, and I think the terseness is just about worth it on balance.

ptersilie: As I said, I don't feel super strongly about it.

ltratt: I forgot to say: there is an advantage to the new versions. It only mutates the self struct if the index is representable. In the old version, you got back a CompilationError but the struct had been mutated. So in that small way, the new version is functionally different (and IMHO better).

ptersilie: I did notice that, but thought since in those cases we have to abort the trace it doesn't matter too much that the vec was already changed. But fundamentally you're right.

vext01: I'm struggling to understand this. Where is `inspect()` defined?

ltratt: `Result::inspect` in the standard library.

vext01: Huh. Clever, but strange :)

PR #1152, opened 2024-05-14
Dockerfile speedups

ltratt:
This PR uses insights from https://www.marcobehler.com/guides/docker-images and https://stackoverflow.com/a/72851168 to speed up using Docker. Between the two commits in this PR, tryci can go from taking ~75s to starting to about 15s. I suspect this will also speedup our use with buildbot.

### DIFF
--- a/.buildbot_dockerfile_debian
+++ b/.buildbot_dockerfile_debian
@@ -1,20 +1,19 @@
 FROM debian:latest
+WORKDIR /ci
+RUN apt-get update && \
+  apt-get -y install clang-15 make curl procps file git cmake python3 \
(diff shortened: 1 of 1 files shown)


.buildbot_dockerfile_debian
```diff
@@ -1,20 +1,22 @@
 FROM debian:latest
+WORKDIR /ci
+RUN --mount=target=/var/lib/apt/lists,type=cache,sharing=locked \
+    --mount=target=/var/cache/apt,type=cache,sharing=locked \
+  rm -f /etc/apt/apt.conf.d/docker-clean && \
```
vext01, line 5: So what's this line for? The blog post doesnt' seem to mention it.

PR #1156, opened 2024-05-14
Perform similar renamings to AOT IR as were previously done to the JIT RI

ltratt:
This largely "ports" the changes to the JIT IR from https://github.com/ykjit/yk/pull/1122 to the AOT IR.

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -241,8 +241,8 @@ index!(BBlockIdx);
 /// An index into [BBlock::instrs].
 #[deku_derive(DekuRead)]
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
-pub(crate) struct InstrIdx(usize);
-index!(InstrIdx);
+pub(crate) struct InstIdx(usize);
+index!(InstIdx);
 
 /// An index into [Module::consts].
 #[deku_derive(DekuRead)]
@@ -319,15 +319,15 @@ impl Display for BinOp {
 /// Uniquely identifies an instruction within a [Module].
 #[deku_derive(DekuRead)]
 #[derive(Clone, Debug, Hash, Eq, PartialEq)]
-pub(crate) struct InstructionID {
+pub(crate) struct InstID {
     /// The index of the parent function.
     func_idx: FuncIdx,
     bb_idx: BBlockIdx,
-    inst_idx: InstrIdx,
+    inst_idx: InstIdx,
 }
 
-impl InstructionID {
-    pub(crate) fn new(func_idx: FuncIdx, bb_idx: BBlockIdx, inst_idx: InstrIdx) -> Self {
+impl InstID {
+    pub(crate) fn new(func_idx: FuncIdx, bb_idx: BBlockIdx, inst_idx: InstIdx) -> Self {
         Self {
             func_idx,
             bb_idx,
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -1,33 +1,28 @@
-//! Yk's AOT IR deserialiser.
+//! The AOT Intermediate Representation (IR).
 //!
-//! This module contains the data structures for the AOT IR and a parser to read it from its
-//! serialised format.
+//! This is the IR created by ykllvm: this module contains both the IR itself and a deserialiser
+//! from ykllvm's output into the IR itself. The IR can feel a little odd at first because we store
+//! indexes into vectors rather than use direct references. This allows us to squeeze the amount of
+//! the memory used down (and also bypasses issues with representing graph structures in Rust, but
```
vext01, line 6: A nitpick that may or may not be worth fixing: squeeze the amount of ~~the~~ memory used down.

vext01: By the way, I've made no attempt to make the AOT IR small. The true reason for the indices is because you can't serialise references.

ltratt: Since the index types are smaller than a word, we get squeeziness for free :) FWIW, I think the "the" thing is probably below our current threshold for change, but happy to go with what you think.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -1,33 +1,28 @@
...
+//! that's slightly accidental).
 //!
-//! The AOT IR accurately reflects the structure and semantics of the AOT binary. As such, it must
-//! not be mutated after the fact.
+//! The IR uses two general terminological conventions:
+//!   * A "definition" is something for which the IR contains complete knowledge.
+//!   * A "declaration" is something compiled externally for which we only know the minimal
+//!   external structure (e.g. a function signature), with the body being present elsewhere in the
```
vext01, line 12: Another optional nitpick: For variable declarations, you'd have an "initialiser" not a "body". Bodies only really make sense when you talk about functions.

ltratt: I did wonder about this, but couldn't think of a term that is a superset of these. I think it's kind-of clear as-is, but am happy to take a suggestion for an improvement.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: I was about to agree, but I've just remembered that in the AOT IR, the indices are all the same size as a pointer, so there is no space saving.

E.g.
```
pub(crate) struct FuncIdx(usize);
```

I think(?) for the AOT IR you don't really want reduced index sizes: since we encode a whole CFG, you could much more easily blow the limits of (say) 16/24-bit indices.

So I guess in light of that, we should delete the part about space saving?

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Probably not worth worrying about at this early stage.

vext01:
Thanks for this. I think this will be much less confusing for us!

ykrt/src/compile/jitc_yk/aot_ir.rs
ltratt: Agreed. Fixed in https://github.com/ykjit/yk/pull/1156/commits/86d540ea8dfb2d3c41f8225a09fa66147a976fc2.

vext01:
Please squash!

ltratt:
Squashed.

PR #1169, opened 2024-05-19
Fix tests.

ltratt:
These should always have had the `%` character in the patterns, but this wasn't noticed because of a bug in fm.


vext01:
Adding `%` is probably OK here as long as you don't expect to match a constant or global variable.

Note that the matcher uses only `{{}}` and we use `%{{}}` to match a local variable (because locals are prefixed `%`).

So tests containing things like `{{4}}` are actually fine, but could be more general than necessary. If you see what I mean?

If the tests pass, I'm OK with it. We probably meant to match local variables anyway.

ltratt:
Without this PR the tests literally no longer pass when fm is fixed. The tests should *always* have had a `%` prefix, but we got away without it solely because of the fm bug.

vext01:
Oh! In that case I think I've been leaning on this bug for what I thought was correct behavior!

ltratt:
OK, I need to work out how to run some more tests that I didn't notice, and which now fail. Bear with.

ltratt:
6a1c5deb fixes another test (this time a missing `$`). OK to squash?

vext01:
ok

ltratt:
Squashed.

PR #1177, opened 2024-05-21
Deserialise unimplemented constants.

vext01:
Requires a ykllvm change: https://github.com/ykjit/ykllvm/pull/164

Before, we serialised unimplemented LLVM constants as a zero-sized Yk constants.

This turned out to be a big mistake, as you don't find out there's a problem until the constant is used and you then have to go on a wild goose chase to find out what's wrong.

This change introduces an explicit "unimplemented constant", which crashes when you ask of it: "what's your type?" or "can I have your bytes?". The panic message then tells you what's up right away.

This changes a yklua `db.lua` crash from:

```
thread '<unnamed>' panicked at ykrt/src/compile/jitc_yk/jit_ir.rs:880:17:
assertion `left == right` failed
  left: 0
 right: 8
```

into:

```
unimplemented const: ptr null
```

To make this happen, I had to make the existing `struct Const` into an `enum Const` containing an `Unimplemented` variant.

I had hoped to express an unimplemented constant at the operand level (serialising an unimplemented *operand*) but this isn't easy. Because a constant index and the constant bytes are serialised at different times, you don't know a constant is unimplemented until after you've already encoded it as an instruction operand. So the "unimplementedness" has to be in `Const` somewhere -- that's what I've done here.

### DIFF
--- a/tests/ir_lowering/struct.ll
+++ b/tests/ir_lowering/struct.ll
@@ -1,9 +1,9 @@
 ; Dump:
 ;   stdout:
 ;     ...
-;     func main() {
+;     func main($arg0: {0: i32, 64: i64}) {
 ;       bb0:
-;         $0_0: {0: i32, 64: i64} = insertvalue const_struct, 100i32
+;         $0_0: {0: i32, 64: i64} = insertvalue $arg0, 100i32
 ;         ret
 ;     }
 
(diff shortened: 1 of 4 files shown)


ltratt:
@vext01 Looks like some tests are broken?

vext01:
Yes, because `icmp ptr` was implemented in the meantime.

https://github.com/ykjit/yk/pull/1177/commits/0880b15f396a60294cd7b60f959fd3702e00e19b fixes this.

ltratt:
Please squash.

vext01:
squashed

ltratt:
Please force push a rebase.

vext01:
Conflicts fixed.

PR #1178, opened 2024-05-21
We only need to encode one IR.

vext01:
Doesn't speed things up much I'm afraid.


ltratt:
Still, it's a useful thing to do -- thanks!

PR #1179, opened 2024-05-21
Add a tentative optimisation module and example optimisation.

ltratt:
The aim of this commit isn't to turn the JIT compiler into a fully optimising compiler, but to explore what we need to do to make it practical to write optimisations. To do that, I've implemented a (very) simple dead-code elimination pass that (to keep things simple for now) only really works with `Add` instructions. This does work for the one test I've implemented, but it's (ahem) not exactly production quality!

This has ended up being a more valuable exercise than I expected because it has immediately highlighted several factors:

  1. Instructions need to be `Clone` and `PartialEq`.

  2. For testing purposes (but not non-testing!) we need some sort of instruction that consumes an operand and tells (a.k.a. "lies") to the optimiser that the instruction has a side effect. One way of thinking of this is that it's equivalent to a "print instruction that doesn't print". I've tentatively called it `Consume` but am open to other naming suggestions.

  3. We need a way of writing human-friendly syntax for test inputs (we have a way -- which I haven't used -- of using human-friendly syntax for test outputs, but that's only half the problem). The "solution" to this is semi-obvious, but it's not a trivial amount of work.

  4. It's probably going to be better adding functions to instructions to express things like "does this instruction produce a side effect" rather than encoding that knowledge in every pass via (tediously long) `match` statements.

  5. We might want to think about things like adding a "rewrite operand" function/visitor to `Inst` rather than encoding it in each pass. I'm less sure about that, though.

I doubt these are the only things that we'll come across as we implement optimisations, but it seems like a useful starting point. So this commit is both, to some extent, a request for comments, but also "it's probably worth thinking about merging this as it can be a very early guide as to how we get the system ready for more powerful optimisations".

### DIFF
--- a/ykrt/src/compile/jitc_yk/mod.rs
+++ b/ykrt/src/compile/jitc_yk/mod.rs
@@ -22,6 +22,7 @@ use ykaddr::addr::symbol_to_ptr;
 pub mod aot_ir;
 mod codegen;
 pub mod jit_ir;
+mod opt;
 mod trace_builder;
 
 pub(crate) struct JITCYk;
@@ -70,6 +71,14 @@ impl Compiler for JITCYk {
                 "--- Begin jit-pre-opt ---\n{jit_mod}\n--- End jit-pre-opt ---",
             ));
         }
+
+        // FIXME: This should be enabled when optimisations can cope with all the inputs we see.
+        // let jit_mod = opt::opt(jit_mod)?;
+        // if should_log_ir(IRPhase::PostOpt) {
+        //     log_ir(&format!(
+        //         "--- Begin jit-post-opt ---\n{jit_mod}\n--- End jit-pre-opt ---",
+        //     ));
+        // }
 
         let cg = Box::new(Self::default_codegen(&jit_mod)?);
         let ct = cg.codegen()?;
--- a/ykrt/src/compile/jitc_yk/opt/lvn.rs
+++ b/ykrt/src/compile/jitc_yk/opt/lvn.rs
@@ -1,0 +1,115 @@
+#[cfg(test)]
+use crate::compile::jitc_yk::jit_ir::TestUseInst;
+use crate::compile::{
+    jitc_yk::jit_ir::{AddInst, Inst, InstIdx, Module, Operand},
+    CompilationError,
+};
+use typed_index_collections::TiVec;
+
+/// Perform (very very) simple dead-code elimination.
+pub(super) fn lvn(mut m: Module) -> Result<Module, CompilationError> {
+    // Which instructions have an effect on the wider world?
+    let mut used_ops = vec![false; m.len()];
+    for inst in m.iter_mut_insts() {
+        match inst {
+            Inst::LoadTraceInput(_) => (),
+            Inst::Add(x) => {
+                if let Operand::Local(y) = x.lhs() {
+                    used_ops[usize::from(y)] = true;
+                }
+                if let Operand::Local(y) = x.rhs() {
+                    used_ops[usize::from(y)] = true;
+                }
+            }
+            #[cfg(test)]
+            Inst::TestUse(x) => {
+                if let Operand::Local(y) = x.operand() {
+                    used_ops[usize::from(y)] = true;
+                }
+            }
+            x => todo!("{x:?}"),
+        }
+    }
+
+    // Rewrite the sequence of instructions, eliminating instructions whose results are unused, and
+    // performing the necessary adjustments to operands in the light of those instructions being
+    // removed.
+    fn operand_off(op: Operand, offs: &[InstIdx]) -> Operand {
+        match op {
+            Operand::Local(x) => Operand::Local(offs[usize::from(x)]),
+            Operand::Const(_) => op,
+        }
+    }
+
+    let mut off = 0;
+    let mut insts = TiVec::<InstIdx, _>::new();
+    let mut offs = Vec::new();
+    for (i, inst) in m.iter_insts().enumerate() {
+        // The `unwrap` can't fail because `i - off <= i` (by definition) and `i` is a
+        // representable `InstIdx`.
+        offs.push(InstIdx::new(i - off).unwrap());
+        if !used_ops[i] {
+            match inst {
+                #[cfg(test)]
+                Inst::TestUse(_) => (),
+                Inst::Add(_) => {
+                    off += 1;
+                    continue;
+                }
+                x => todo!("{x:?}"),
+            }
+        }
+        let inst = match inst {
+            Inst::LoadTraceInput(_) => inst.clone(),
+            Inst::Add(x) => {
+                AddInst::new(operand_off(x.lhs(), &offs), operand_off(x.rhs(), &offs)).into()
+            }
+            #[cfg(test)]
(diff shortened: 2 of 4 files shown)


ykrt/src/compile/jitc_yk/opt/lvn.rs
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    let mut off = 0;
+    let mut insts = TiVec::<InstIdx, _>::new();
+    let mut offs = Vec::new();
+    for (i, inst) in m.iter_insts().enumerate() {
+        // The `unwrap` can't fail because `i - off <= i` (by definition) and `i` is a
+        // representable `InstIdx`.
+        offs.push(InstIdx::new(i - off).unwrap());
```
ptersilie, line 50: The code to update the offsets looks a bit tedious. I wonder how annoying this will be for more complex optimisations and whether there's a way to make this easier for the user (i.e. optimisation writer)?

ltratt: I also wonder that: it's point (5) above.

ykrt/src/compile/jitc_yk/mod.rs
```diff
@@ -71,6 +72,14 @@ impl Compiler for JITCYk {
             ));
         }
 
+        // FIXME: This should be enabled when optimisations can cope with all the inputs we see.
+        // let jit_mod = opt::opt(jit_mod)?;
+        // if should_log_ir(IRPhase::PostOpt) {
+        //     log_ir(&format!(
+        //         "--- Begin jit-post-opt ---\n{jit_mod}\n--- End jit-pre-opt ---",
```
vext01, line 79: Random thought: later we probably will want the ability to print (and diff) individual passes at a more fine-grained level.

E.g.
```
--- Begin jit-opt-in-progress (after const_prop) ---
```

Not actionable now, obviously.

ykrt/src/compile/jitc_yk/opt/lvn.rs
```diff
@@ -0,0 +1,115 @@
...
+use crate::compile::jitc_yk::jit_ir::ConsumeInst;
+use crate::compile::{
+    jitc_yk::jit_ir::{AddInst, Inst, InstIdx, Module, Operand},
+    CompilationError,
+};
+use typed_index_collections::TiVec;
+
+/// Perform (very very) simple dead-code elimination.
+pub(super) fn lvn(mut m: Module) -> Result<Module, CompilationError> {
+    // Which operations are used?
```
vext01, line 11: s/operations/operands?/

ykrt/src/compile/jitc_yk/opt/lvn.rs
```diff
@@ -0,0 +1,115 @@
...
+    let mut used_ops = vec![false; m.len()];
+    for inst in m.iter_mut_insts() {
```
vext01, line 13: This looks like a def/use analysis.

We could generalise and abstract this into an API and then it could be shared with the register allocator, which will also need to know this kind of stuff (e.g. a local is born upon its definition, a local dies after it's last use).

It's a shame we will have to manually write the code to find operands in instructions. I guess that was one advantage of the "dynamic" IR: the operands were just there is a list. I'm not suggesting we go back to a dynamic IR though!

vext01:
I think the name `Consume` could be misleading, because it almost sounds like the operand is consumed by the instruction, when really it's only *used*.

I wonder if `TestUse` or `BlackBox` would be a better name?

ykrt/src/compile/jitc_yk/opt/lvn.rs
ltratt: Should actually be "instructions"!

ykrt/src/compile/jitc_yk/opt/lvn.rs
ltratt: > This looks like a def/use analysis.

Mostly, but this code is not meant to be a usable optimisation pass: it's meant to show us how *an* optimisation pass looks in our code. The analysis is incomplete beyond the very simple example.

ltratt:
`TestUse` isn't a bad shout.

ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -0,0 +1,10 @@
+use super::jit_ir::Module;
+use crate::compile::CompilationError;
+
+mod lvn;
```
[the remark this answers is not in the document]
ltratt: Whatever you want it to :) I just needed a name.

ltratt:
> I've had similar requirements for other unit tests. In those I took the operand out of the trace input struct. I don't know if the same strategy could be used here?

No, because that only tells you that there's an unknown input: it doesn't tell you that you *use* the resulting value.

>>    We need a way of writing human-friendly syntax for test inputs
>Before doing that, I think we should review and make consistent, our textual IRs. The human-readable IR format is very ad-hoc right now.

Whatever we need to do here, we need to do it ASAP.

ykrt/src/compile/jitc_yk/opt/lvn.rs
ptersilie: I guess we could just define a function `operands()` on each instructions which returns all operands?

ykrt/src/compile/jitc_yk/opt/mod.rs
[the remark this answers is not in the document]
ptersilie: I was wondering this too.

ykrt/src/compile/jitc_yk/opt/lvn.rs
ltratt: That's definitely one possibility. Honestly, I'm currently unsure what the best approach might be.

ykrt/src/compile/jitc_yk/opt/lvn.rs
vext01: "Instructions" makes the most sense IMHO.

We are scanning for local variable operands, and then caching a list of instructions which define those operands.

ykrt/src/compile/jitc_yk/opt/mod.rs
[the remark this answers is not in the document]
vext01: Assuming that file is just for that one specific optimisation, `dce` (Dead Code Elimination)?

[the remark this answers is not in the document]
ltratt: I suggest not bikeshedding this name right now.

ykrt/src/compile/jitc_yk/opt/lvn.rs
ltratt: Fixed in eeccf229.

ykrt/src/compile/jitc_yk/mod.rs
ltratt: Hmm, maybe. I think this should be a very internal detail, though, so it may not actually be worth doing in this way. Time will tell I suppose!

ltratt:
> TestUse isn't a bad shout.

Fixed in 7d7ce0fe.

ltratt:
@ptersilie OK to squash for you too?

ptersilie:
Yes, please go ahead.

ltratt:
Squashed.

PR #1184, opened 2024-05-22
Review human-readable JIT IR, in preparation for having it parsable.

vext01:
This change hopefully makes the IR more consistent and concise. If users/developers are going to be writing textual IR by hand, we may as well make is easier for them.

 - Make all names lower case, using underscores where it feels right. e.g. `PtrAdd` -> `ptr_add`, `LookupGlobal` -> `lookup_global`.

 - Shorten some names too, notably:
   - Use LLVM short predicate names, because people know them. e.g. `eq`, `sgt` and `ule` (equal, signed-greater-than, unsigned-less-than-or-equal).
   - `IndirectCall` -> `icall`.
   - `LoadTraceInput` -> `load_ti`
   - `SignExtend` -> `sext`
   - `ZeroExtend` -> `zext`
   - `TraceLoopStart` -> `tloop_start` (and updated debug comments and dynasm label names for consistency)

This mostly only touches the JIT IR for now. In one place a construct was shared among both IRs, so that changed for both.

I'll do the same review for the AOT IR later.

While here, also document the `display()` idiom.


ltratt:
This looks good except I thought we'd agreed to kill `Icmp` which is a weird artefact of our current IR and not something that has any use for a human reader?

vext01:
I'm only handling renamings here. I'm happy to change icmp, but since that's a unit of work in its own right, that'd be better as a new PR?

As I understand, you want to expand the icmp instruction into separate opcodes, one per predicate, is that right?

vext01:
Lukas may also have opinions on some of these changes, so I'll assign him too if that's ok?

ltratt:
>I'm only handling renamings here. I'm happy to change icmp, but since that's a unit of work in its own right, that'd be better as a new PR?
>
>As I understand, you want to expand the icmp instruction into separate opcodes, one per predicate, is that right?

A new PR is fine for this. There are two aspects:

1. How we pretty print `ICmp`s. I can't see any utility in adding `ICmp ... Eq` over just `Eq` so I suggest we ditch the `ICmp` aspect.
2. At some point we change the actual IR.

These two things can be done separately: I'd rather get (1) done now but (2) can wait for a bit.

>Lukas may also have opinions on some of these changes, so I'll assign him too if that's ok?

I think they're pretty uncontroversial overall.

ptersilie:
Looks good to me.

vext01:
Ah, I see. Well we can do (1) now.

Just to check, later, when we change the IR structures, you are OK with spending the opcode space on separate instructions? LLVM defines some 20 predicates (including float preds). I'm not sure we will need them all...

https://llvm.org/doxygen/classllvm_1_1CmpInst.html#a2be3583dac92a031fa1458d4d992c78b

ltratt:
> Just to check, later, when we change the IR structures, you are OK with spending the opcode space on separate instructions?

Yes.

ltratt:
@vext01 Correct, but do you want to do that in this PR?

vext01:
If it's OK with you, I'd like to tackle that later.

ltratt:
OK but you'll need to force push that new commit away please: I stopped the merge because of the new commit.

vext01:
squashed.

PR #1185, opened 2024-05-22
Add a basic JIT IR parser.

ltratt:
This uses grmtools to parse pretty-printed JIT IR concrete syntax back into JIT IR. The parser is woefully incomplete, only handling enough for the toy optimisation in lvn.rs, and not dealing with anything clever with local operand numbers and so on.

However, this is already enough to prove a useful point: we can now turn the horror that was the (sole) test in lvn.rs to:

```
fm_match(
    "
  entry:
    %0: i16 = load_ti 0, i16
    %1: i16 = load_ti 16, i16
    %2: i16 = add %0, %1
    %3: i16 = add %0, %1
    %4: i16 = add %0, %3
    test_use %3
    test_use %4
",
    |m| lvn(m).unwrap(),
    "
  ...
  entry:
    %0: i16 = load_ti 0, i16
    %1: i16 = load_ti 16, i16
    %2: i16 = add %0, %1
    %3: i16 = add %0, %2
    test_use %2
    test_use %3
",
);
```

Much more will need to be done on this before it can be used elsewhere, but this feels like the minimum useful to be worthy of considered for merging.

### DIFF
--- a/ykrt/Cargo.toml
+++ b/ykrt/Cargo.toml
@@ -42,6 +42,9 @@ default-features = false
 features = ["read_core", "elf"]
 
 [build-dependencies]
+cfgrammar = "0.13"
+lrlex = "0.13"
+lrpar = "0.13"
 regex = "1.9"
 ykbuild = { path = "../ykbuild" }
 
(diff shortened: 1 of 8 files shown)


ptersilie:
Neat.

PR #1187, opened 2024-05-22
Implement srem and trunc.

ptersilie:


ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -825,6 +859,35 @@ impl<'a> X64CodeGen<'a> {
...
+            matches!(from_type, jit_ir::Ty::Integer(_)) || matches!(from_type, jit_ir::Ty::Ptr)
+        );
+        // You can only truncate a bigger integer to a smaller integer.
+        debug_assert!(from_size > to_size);
+
+        // FIXME: assumes the input and output fit in a register.
+        self.load_operand(WR0, &from_val);
+        debug_assert!(to_size <= REG64_SIZE);
+
+        // FIXME: There's no instruction on x86_64 to mov from a bigger register into a smaller
```
vext01, line 881: I *think* that if you trunc to a byte boundary, you can just do nothing and leave the top bytes undefined. The IR type should dictate how many bytes should be read/written when the value is later used?

If you trunc to a non-byte value, you will be forced to zero some bits though. Right?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -1779,5 +1842,53 @@ mod tests {
...
+                .insert_ty(jit_ir::Ty::Integer(IntegerTy::new(32)))
+                .unwrap();
+            let op = m
+                .push_and_make_operand(jit_ir::LoadTraceInputInst::new(0, i32_ty_idx).into())
+                .unwrap();
+            m.push(jit_ir::TruncInst::new(&op, i8_ty_idx).into())
+                .unwrap();
+            let patt_lines = "
+                ...
+                ; %0: i32 = load_ti 0, i32
```
ltratt, line 1884: I'm about to raise a PR which removes the trailing `, T` from `load_ti` (see https://github.com/ykjit/yk/commit/24bcb9fec9eb5d3180f2fbe15438d197ff007b0c). It might be easiest to remove those from this PR, as *probably* because of the ykllvm pr, the load_ti PR will end up going in first. You'll only want to edit the `load_ti`s introduced in this PR, though, which I think is just this one and the one two lines below.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ptersilie: That matches my understanding.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ptersilie: Ok, I'll wait for you stuff to go in first then.

ltratt: If you want you can do this preemptively now with a commit. By the looks of things it's the only comment we have anyway!

ptersilie:
Pulled in upstream and fixed the test. Ok to squash?

ltratt:
Please squash.

ptersilie:
Squashed.

PR #1194, opened 2024-05-23
Test unsupported PHIs and casts.

vext01:
Requires https://github.com/ykjit/ykllvm/pull/168

### DIFF
--- a/tests/ir_lowering/unsupported_variants.ll
+++ b/tests/ir_lowering/unsupported_variants.ll
@@ -26,6 +26,9 @@
 ;      bb4:
 ;         unimplemented <<  %{{25}} = ptrtoint ptr %{{ptr}} to i8>>
 ;         unimplemented <<  %{{26}} = ptrtoint <8 x ptr> %{{ptrs}} to <8 x i8>>>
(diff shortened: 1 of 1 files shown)


tests/ir_lowering/unsupported_variants.ll
```diff
@@ -26,6 +26,9 @@
 ;      bb4:
 ;         unimplemented <<  %{{25}} = ptrtoint ptr %{{ptr}} to i8>>
 ;         unimplemented <<  %{{26}} = ptrtoint <8 x ptr> %{{ptrs}} to <8 x i8>>>
+;         unimplemented <<  %{{_}} = sext <4 x i32> %{{_}} to <4 x i64>>>
```
ltratt, line 29: Is there a reason we're using `_` as much as we are?

vext01: In this context the variable names are irrelevant.

For example, when testing a:
```
sext_vec = sext <4 x i32> %nums to <4 x i64>
```

It's sufficient to check for an `sext` of type `<4 x i32>`.

ltratt: OK.

PR #1196, opened 2024-05-24
More flexible and less forgiving JIT IR parser

ltratt:
The core of this PR is https://github.com/ykjit/yk/commit/2f5b210a6e7d5116f7ce4c243d6ca244a9246ef8, but the first two commits are the necessary prep to make that third commit plausible.

### DIFF
--- a/ykrt/src/compile/jitc_yk/jit_ir.y
+++ b/ykrt/src/compile/jitc_yk/jit_ir.y
@@ -29,10 +29,10 @@ Insts -> Result<Vec<ASTInst>, Box<dyn Error>>:
 
 Inst -> Result<ASTInst, Box<dyn Error>>:
     "LOCAL_OPERAND" ":" Type "=" "LOAD_TI" "INT" {
-      Ok(ASTInst::LoadTraceInput{type_: $3?, off: $6?.span()})
+      Ok(ASTInst::LoadTraceInput{assign: $1?.span(), type_: $3?, off: $6?.span()})
     }
   | "LOCAL_OPERAND" ":" Type "=" "ADD" Operand "," Operand  {
-      Ok(ASTInst::Add{type_: $3?, lhs: $6?, rhs: $8?})
+      Ok(ASTInst::Add{assign: $1?.span(), type_: $3?, lhs: $6?, rhs: $8?})
     }
   | "TEST_USE" Operand { Ok(ASTInst::TestUse($2?)) }
   ;
--- a/ykrt/src/compile/jitc_yk/jit_ir_parser.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir_parser.rs
@@ -11,7 +11,7 @@ use fm::FMBuilder;
 use lrlex::{lrlex_mod, DefaultLexerTypes, LRNonStreamingLexer};
 use lrpar::{lrpar_mod, NonStreamingLexer, Span};
 use regex::Regex;
-use std::{error::Error, sync::OnceLock};
+use std::{collections::HashMap, error::Error, sync::OnceLock};
 
 lrlex_mod!("compile/jitc_yk/jit_ir.l");
 lrpar_mod!("compile/jitc_yk/jit_ir.y");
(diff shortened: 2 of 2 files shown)


vext01:
> This commit makes it possible to have JIT IR inputs which do not have operand assignment [0, 1, 2, 3, ...]: operands can now be defined in any order and the parser will map them appropriately.

I'm not sure what this means. Can you give an example?

ltratt:
> I'm not sure what this means. Can you give an example?

Please check the tests in the commit: I think it'll be clearer then.

vext01:
I mean, the tests look like things that should parse, but I'm still unsure what this commit does.

What is an input that doesn't have operand assignment?

We could just be working with different terminology, but I'm not sure.

vext01:
So I think we are saying the commit allows for local variables to be named non-sequentially?

(I was confused by the use of "operand" here. A local variable isn't an operand. An operand is a use of a local variable as an argument to an instruction)

But that aside, (until now at least) the variable name must be `%` followed by the instruction index of the instruction that defines it. As such, variable names aren't sequential. Instructions in the JIT IR don't have a name field, so the name is just derived from the instruction index.

So the test from this commit:
```
          entry:
            %7: i16 = load_ti 0
            %3: i16 = load_ti 1
            %19: i16 = add %7, %3
```

Actually isn't valid JIT IR. You can choose to parse it, but you will be parsing a different language than the JIT IR.

ltratt:
> So I think we are saying the commit allows for local variables to be named non-sequentially?

Yes, so we can extract JIT IR fragments and aren't forced to rewrite names.

> Actually isn't valid JIT IR. You can choose to parse it, but you will be parsing a different language than the JIT IR.

Yes and no. From an internal data-structure POV you're correct: but when I'm writing textual input, I don't care about indices. That representation *could* change and the names, IMHO, do not need to carry a semantic value.

vext01:
> Yes and no. From an internal data-structure POV you're correct: but when I'm writing textual input, I don't care about indices.

We could move forward like this if:
 - we are ok with the data structures only being able to express a subset of what the textual language can.
 - we don't care round-tripping

@ptersilie do you have any opinions on this?

ltratt:
@vext01 I have a phrasing that might resonate better with you. All this PR does is parse a (textual) tree to another (abstract syntax) tree that is equal modulo α-renaming (i.e. the two trees are semantically equivalent). Make sense?

vext01:
@ltratt I'm OK with this.

Could we try to reword the commit message to this effect?

vext01:
Well the last commit message still doesn't make sense to me.

Based on our discussion, I think you want to say something like:

> This commit changes the JIT IR parser to allow inputs where local variable names aren't necessarily the instruction index number. This makes it easier for a human to write textual JIT IR. This does mean that if you parse, and then re-stringify a module, you may not get the same textual output, but it doesn't matter since the IR will still be semantically equal (just local variables may have different names).

ltratt:
Force pushed an updated commit message.

vext01:
I'm starting to feel like I'm being difficult, but the message still doesn't make sense.

> This commit makes it possible to have JIT IR inputs which do not have operand assignment [0, 1, 2, 3, ...] in order: operands can now be defined in any order and the parser will map them appropriately. 

Specifically:

> operand assignment

doesn't make sense. I think we are talking about *local variables*, not operands.

> [0, 1, 2, 3, ...]

Local variable names weren't neccesarily sequential before this change, e.g.:

```
%0 = add ...
call @f()
%2 = add ...
```

Here `%1` is skipped because the instruction at index 1 doesn't define a local variable.

Something like this would be much clearer?

> This commit changes the JIT IR parser to allow inputs where local variables aren't necessarily named using the defining instruction's index number.

ltratt:
> Local variable names weren't neccesarily sequential before this chang

Incorrect: that was what the parser accepted in its first version, as I simplified the problem.

I feel that the updated commit message, which uses the standard CS terminology of alpha-renaming, is adequately clear, even if it's not perfect (but which commit message is?.

vext01:
> Incorrect: that was what the parser accepted in its first version, as I simplified the problem.

Ah gotcha.

I'm going to merge this, but I still don't understand what an "operand assignment" is.

PR #1200, opened 2024-05-27
Parse enough of the JIT IR to rewrite all of the x86 backend test suite

ltratt:
Bit-by-bit, this PR parses enough of the JIT IR to replace all of the hand-crafted IR in the x86 backend's tests. Perhaps the easiest way to see what this PR enables is to look at that file at the end of this commit in https://github.com/ykjit/yk/blob/e738602b9bd9afa48bcb59f8db67c9f7c511af85/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs#L1174.

There is more that can be done in future PRs, and this PR is in places a bit messy (sometimes highlighting mild inconsistencies in the JIT IR), but it moves the JIT IR parser from a woefully incomplete to into something that we can realistically use for tests -- and it makes writing tests based on JIT IR hugely easier than before!


vext01:
Didn't spot anything wrong here.

ltratt:
@vext01 We should be able to retry this one now.

PR #1202, opened 2024-05-28
Be explicit about integer truncation.

ltratt:
This commit stops `jit_ir::IntegerTy::make_constant` truncating integer values. Currently this function is as dangerous as Rust's `as`, but hides behind an innocuous function name and a doc string which I at least managed to misread multiple times:

> The byte size of `T` must be at least as big as `self.byte_size()`. If the byte size of `T`
> is larger, `val` will be truncated to `self.byte_size()` bytes.

Both sentences are congruous with the code, but I read them many, many times before realising what they meant. In particular, I assumed the first sentence meant to say "less than or equal", and then I assumed that the second sentence was probably incorrect. Clearly I was wrong in both cases!

This commit makes `make_constant` neither extend or truncate: you now have to pass it in the *correctly sized* Rust integer type or else it will (at least in debug mode) panic. This is, inevitably, more annoying to use, but it now places the burden on the caller of `make_constant` to do truncation/extension at their end, rather than having `make_constant` do it silently.

### DIFF
--- a/ykrt/src/compile/jitc_yk/trace_builder.rs
+++ b/ykrt/src/compile/jitc_yk/trace_builder.rs
@@ -699,7 +699,8 @@ impl<'a> TraceBuilder<'a> {
         let mut jit_ptr = self.handle_operand(ptr)?;
         if const_off != 0 {
             let co_ty = jit_ir::IntegerTy::new(32);
-            let co_const = co_ty.make_constant(&mut self.jit_mod, const_off)?;
+            let co_const =
+                co_ty.make_constant(&mut self.jit_mod, i32::try_from(const_off).unwrap())?;
             let co_opnd = jit_ir::Operand::Const(self.jit_mod.insert_const(co_const)?);
             jit_ptr = self
                 .jit_mod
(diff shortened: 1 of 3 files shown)


ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -798,9 +799,21 @@ impl<'a> TraceBuilder<'a> {
...
+                    16 => jit_int_type
+                        .to_owned()
+                        .make_constant(&mut self.jit_mod, u16::try_from(val).unwrap())?,
+                    32 => jit_int_type
+                        .to_owned()
+                        .make_constant(&mut self.jit_mod, u32::try_from(val).unwrap())?,
+                    64 => jit_int_type
+                        .to_owned()
+                        .make_constant(&mut self.jit_mod, u64::try_from(val).unwrap())?,
+                    _ => unreachable!(),
```
vext01, line 815: Is that unreachable or `todo!()`? I think you can switch on any arbitrary integer type.

Not that it matters much, they both stop the program with a crash.

ltratt: Yes, they're effectively equivalent as they both panic: at the moment this branch is literally unreachable, if I read ykllvm's serialiser correctly.

vext01: I think it should be theoretically possible, but let's just merge it for now.

vext01:
OK

ltratt:
Squashed.

ltratt:
Please force push a rebase + squash (doesn't have to be to 1 commit though) against master.

PR #1205, opened 2024-05-28
Add an explicit well-formedness check for the JIT IR

ltratt:
This PR adds an explicit well-formedness check for the JIT IR. This is currently always called when `Module::from_str` is used, but can be called elsewhere if desired. It takes the basic checks from the x86 backend and expands and generalises them. It's useful enough that it caught one error in JIT IR input (in a parser test), which is encouraging (I'm ignoring the worrying part that I didn't spot this earlier!).

As part of this, I thoroughly confused myself with call arguments and function parameters, so I have snuck in a commit to use the standard naming of these things (https://github.com/ykjit/yk/commit/41dfca912978959b61ba59f322a8943ab36fd707) which might help me next time I have to wrap my head around this stuff.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -10,7 +10,7 @@
 
 use super::{
     super::{
-        jit_ir::{self, FuncDeclIdx, InstIdx, Operand, Ty},
+        jit_ir::{self, BinOp, FuncDeclIdx, InstIdx, Operand, Ty},
         CompilationError,
     },
     abs_stack::AbstractStack,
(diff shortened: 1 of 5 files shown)


vext01:
Yep, I like this.

Small things to think about:
 - you could use `verify` in place of `well_formed` to shorten stuff if you wanted.
 - is it worth enumerating what is checked in the doc comment? I reckon the code does a good job of describing what's checked already.
 - `assert_well_formed` is already quite large. Later we could split it down into stuff like `verify_module()`, `verify_func()`, `verify_block()` etc.

None of these are all that important, so I will follow your lead.

ltratt:
I think we'll end up using "verify" for other things: in general "verify" is a stronger property than "well-formed". I agree it's a mouthful: if it turns out to be a pain we can definitely reconsider the name.

>  Is it worth enumerating what is checked in the doc comment?

I did that on line 5, though I missed out the binary operations case. Fixed in b69c159d.

vext01:
> I did that on line 5

Sorry, I wasn't clear. I meant "I saw you had, but is it worth it?" and my follow up if "yes" was "binops are missing" :)

I'll interpret the addition of binops to mean that we want to keep the comment and merge this. Cancel the merge if you change your mind.

vext01:
My bad. Yes, please squash.

ltratt:
Squashed.

PR #1207, opened 2024-05-30
Use "lhs/rhs" in `ICmp` to match `BinOp`.

ltratt:
We could have chosen "left" and "right" but my sense is that "lhs" and "rhs" are the more common terminology, so we might as well standardise on that.


vext01:
Thanks

PR #1213, opened 2024-05-30
jit_ir: Expand `ptr_add` into `ptr_add` and `dyn_ptr_add`.

vext01:
Before the JIT IR `ptr_add` was used for both constant and dynamic pointer offsetting.

After this change, `ptr_add` is strictly a constant pointer offsetting, and `dyn_ptr_add` is a dynamic pointer offsetting.

The latter is now basically a multiply operation in disguise. We might have been able to re-use the `mul` opcode, but I find it useful to have all of LLVM's nuanced `getelementptr` contained in a specific opcode.

Due to encoding constraints, we don't support the full range of offsets that LLVM could emit, so we may need to revisit this later depending upon how many traces use large/small offsets/indices/sizes and thus get rejected by the trace compiler.


ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -1594,9 +1608,59 @@ impl PtrAddInst {
...
+    pub(crate) fn new(ptr: Operand, num_elems: Operand, elem_size: u16) -> Self {
+        Self {
+            ptr: PackedOperand::new(&ptr),
+            elem_size,
+            num_elems: PackedOperand::new(&num_elems),
+        }
+    }
+
+    pub(crate) fn ptr(&self) -> Operand {
+        let ptr = self.ptr;
```
vext01, line 1653: I've just learned that if you want to do an unaligned read of a copyable type, you can do so without using `unsafe` like this.

Note that merging the lines wouldn't compile.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -1594,9 +1608,59 @@ impl PtrAddInst {
...
+    /// The pointer to offset
+    ptr: PackedOperand,
+    /// The element size.
+    ///
+    /// Depending upon the platform, LLVM `getelementptr` may allow larger element sizes than what
+    /// this field can express. Due to space constraints we only accept a reduced range of values
+    /// and traces requiring values outside of this range cannot be JIT compiled.
+    elem_size: u16,
+    /// The (dynamic) number of elements.
+    num_elems: PackedOperand,
```
ltratt, line 1640: If I understand correctly, the way to implement `DynPtrAdd` is (roughly) `m.type_(self.ptr).byte_size() * self.num_elems`?

ltratt: I was asking a question not making a statement, because I don't know the answer :) I guess your question implies the answer to my question is "no"?

vext01: No, you can't assume that array elements are pointer-sized.

This is why in the LLVM API [GetElementPtrInst::collectOffset()](https://llvm.org/doxygen/classllvm_1_1GetElementPtrInst.html#a600fb2213a893d3db4d03ea9fe6b9e63), returns a vector of dynamic indices that are pairs of `Value, APInt`. The former is the dynamic number of elements, and the `APInt`s the constant byte-size of the elements of the array being indexed.

Then in the [language reference](https://llvm.org/docs/LangRef.html#id237) they explain:

> the index itself is sign-extended or truncated, and then multiplied by the type allocation size (that is, the size rounded up to the ABI alignment) of the currently indexed type. The offsets are then added to the low bits of the base address up to the index type width, with silently-wrapping two’s complement arithmetic.

A lot of detail, but I hope it roughly makes sense?

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -1594,9 +1608,59 @@ impl PtrAddInst {
...
+#[repr(packed)]
+pub struct DynPtrAddInst {
+    /// The pointer to offset
+    ptr: PackedOperand,
+    /// The element size.
+    ///
+    /// Depending upon the platform, LLVM `getelementptr` may allow larger element sizes than what
+    /// this field can express. Due to space constraints we only accept a reduced range of values
+    /// and traces requiring values outside of this range cannot be JIT compiled.
+    elem_size: u16,
```
ltratt, line 1638: In the pretty-printing this is the third element, but it's the second in the struct. I don't mind which we do, but consistency will help my pea brain :)

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -686,39 +686,33 @@ impl<'a> TraceBuilder<'a> {
...
 
         // Now apply any dynamic indices.
         //
         // Each offset is the number of elements multiplied by the byte size of an element.
-        let usize_bitsize = u32::try_from(mem::size_of::<usize>() * 8).unwrap(); // always fits.
-        let count_ty = jit_ir::IntegerTy::new(usize_bitsize);
-        for (count, size) in dyn_elem_counts.iter().zip(dyn_elem_sizes) {
-            let count_opnd = self.handle_operand(count)?;
+        for (num_elems, elem_size) in dyn_elem_counts.iter().zip(dyn_elem_sizes) {
+            let num_elems_opnd = self.handle_operand(num_elems)?;
```
ltratt, line 701: No need to change anything in this PR, but one of the nice things about Rust is to overwrite the variable, which stops us worrying about new (longer) variable names. So I'd just do `let num_elems = self.handle_operand(num_elems)?` here (and we do that in various other places).

vext01:
I only changed the JIT opcode, so no ykllvm change required.

(The AOT PtrAdd contains all the information (the dynamic and static elements) for one LLVM `getelementptr` -- we are able to do so because we don't have any space constraints for now).

ltratt:
> (The AOT PtrAdd contains all the information (the dynamic and static elements) for one LLVM getelementptr -- we are able to do so because we don't have any space constraints for now).

Ah, got it!

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: Yeah true, since I'm gonna be squashing later anyway could sneak that in now.

ltratt: I won't object if you do!

vext01: Fixed in e8233ae45a468ab7797b1a58ffcc09d6951d6ba7

ltratt:
Please squash and rebase.

vext01:
squashed.

PR #1215, opened 2024-06-03
Move JIT IR constants from "bag of bytes" to usable Rust types.

ltratt:
The "bag of bytes" approach has 3 problems:

  1. It's impractical to work. I tried writing a JIT analysis and the query "is this `Const` an integer 1" requires writing platform specific code. That's what triggered me to write this commit.

  2. All constants have a fixed size know at AOT: making the JIT treat them as variable sized things is unnecessary.

  3. The "struct pointing to a vec" thing is memory inefficient.

This commit thus moves `Const` from a "struct pointing to a vector" (which also isn't very efficient!) to an `enum`:

```rust
pub(crate) enum Const {
    I8(i8),
    I16(i16),
    I32(i32),
    I64(i64),
    Ptr(*const ()),
 }
```

This is more memory efficient than the previous representation and easier to work with: see e.g. the changes to `load_const` in trace_builder.

As noted in (2) above, it's not necessary for the AOT IR to consider these things variable sized, but this commit only deals with the JIT IR case. It would make sense to ram this change back through the AOT IR and ykllvm (i.e. removing the "bag of bytes" approach there and moving to a similar constant-sized enum, which would save some of the byte-vec dance that we have at JIT time), but from my perspective, I only care about users of the JIT IR, so it's not relevant for this commit.

Note that I had to delete the original `stringify_const_ptr` test because I couldn't work out how to make it platform independent (alas, Rust's integer types are an issue here), and the rest of the test is of a simple property that `stringify_const_ptr2` was adequately testing anyway.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -18,7 +18,6 @@ use super::{
     CodeGen,
 };
 use crate::compile::{jitc_yk::jit_ir::IndirectCallIdx, CompiledTrace};
-use byteorder::{NativeEndian, ReadBytesExt};
 use dynasmrt::{
     components::StaticLabel, dynasm, x64::Rq, AssemblyOffset, DynasmApi, DynasmError,
     DynasmLabelApi, ExecutableBuffer, Register,
@@ -827,29 +826,13 @@ impl<'a> X64CodeGen<'a> {
 
     /// Load a constant into the specified register.
     fn load_const(&mut self, reg: Rq, cidx: jit_ir::ConstIdx) {
-        let cst = self.m.const_(cidx);
-        let mut bytes = cst.bytes().as_slice();
-        let size = self.m.type_(cst.ty_idx()).byte_size().unwrap();
-        debug_assert_eq!(bytes.len(), size);
-        match size {
-            8 => {
-                let val = bytes.read_i64::<NativeEndian>().unwrap();
-                dynasm!(self.asm; mov Rq(reg.code()), QWORD val);
-            }
-            4 => {
-                let val = bytes.read_i32::<NativeEndian>().unwrap();
-                dynasm!(self.asm; mov Rq(reg.code()), DWORD val);
-            }
-            2 => {
-                let val = bytes.read_i16::<NativeEndian>().unwrap();
-                dynasm!(self.asm; mov Rw(reg.code()), WORD val);
-            }
-            1 => {
-                let val = bytes.read_i8().unwrap();
-                dynasm!(self.asm; mov Rq(reg.code()), val as i32);
-            }
-            _ => todo!("{}", size),
-        };
+        match self.m.const_(cidx) {
+            jit_ir::Const::I8(x) => dynasm!(self.asm; mov Rq(reg.code()), i32::from(*x)),
+            jit_ir::Const::I16(x) => dynasm!(self.asm; mov Rw(reg.code()), WORD *x),
+            jit_ir::Const::I32(x) => dynasm!(self.asm; mov Rq(reg.code()), DWORD *x),
+            jit_ir::Const::I64(x) => dynasm!(self.asm; mov Rq(reg.code()), QWORD *x),
+            jit_ir::Const::Ptr(_) => todo!(),
+        }
     }
 
     fn store_local(&mut self, l: &LocalAlloc, reg: Rq, size: usize) {
--- a/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
@@ -30,7 +30,6 @@ mod well_formed;
 use super::aot_ir;
 use crate::compile::CompilationError;
 use indexmap::IndexSet;
-use num_traits::{PrimInt, ToBytes};
 use std::{
     ffi::{c_void, CString},
     fmt, mem,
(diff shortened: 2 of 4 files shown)


vext01:
I see what you are aiming at, but I don't think this will work in the long run:

 - How will we represent arbitrary-width integers, like `i1`, `i7`, `i77` and `i256`  etc? I know we don't support those now, but we will need to be able to at some point.
 - How will we represent constant aggregate user types, like constant structs, constant arrays, and aggregates of structs and arrays?

As far as I can see, a bag of bytes is the only way.

ltratt:
> How will we represent arbitrary-width integers, like i1, i7, i77 and i256 etc? I know we don't support those now, but we will need to be able to at some point.

We can make a variant something like `ArbBint{bit_len: i32, vec: Vec<u8>}` or similar. I'm also quite happy to put off trying to deal with them until we have to!

> How will we represent constant aggregate user types, like constant structs, constant arrays, and aggregates of structs and arrays?

Those will be other enum variants, and some of them might point to be a bag of bytes (like the arbint above). That's totally fine: after all, we don't have to coerce a struct into an integer.

Put another way: we have to do lots of analysis on integer types, so special casing them (which is what this PR does) is going to happen somewhere in the code. The change I'm suggesting in this PR is definitely not the only way of handling this. For example, we could have method(s) along the lines of `fn to_integer(&self) -> Option<usize>` to handle the "do analysis on integers" case. Maybe that will even prove to be the better approach -- time will tell I suspect -- but even then we really don't want to represent fixed-sized integers as bags of bytes.

vext01:
OK, so the subset of types that we explicitly enumerate here, do we consider those optimisations, and the general case (arbitrarily-shaped types) as the slow path?

ltratt:
> do we consider those optimisations

I wouldn't *quite* put it like that, but "sort of, yes". I would say it's more "these are the base types that we have to use everywhere, so forcing every user to deal with platform-specific hard-to-get-right conversions every time they want to use them isn't great". So for me it's a combination of "correctness, ergonomics, and optimisation" -- but optimisation is in that list :)

> and the general case (arbitrarily-shaped types) as the slow path?

Yes, this bit I agree with. In particular, most of the slow path will never need to be thought of as an integer.

vext01:
I think I'm OK with this.

One snag is that because we will then have a non-canonical way to express certain constant integers, we will have to be very careful when comparing values. E.g. There would be two ways to store any given (e.g.) `u16`: once via the inline `u16` and again via the bag 'o bytes. So a structural disequality test, though tempting, would be incorrect.

If we are OK with this, I'll do a full review.

ltratt:
> One snag is that because we will then have a non-canonical way to express certain constant integers, we will have to be very careful when comparing values.

Nearly everything can be represented in multiple ways (e.g. you can represent `x + 3` in the IR as `x + 1 + 1 + 1`, so this doesn't worry me. If you want `I16` optimisations, use the `I16` enum variant, not the `ArbInt` variant. Doing the latter wouldn't affect correctness, but would affect optimisation, but I don't see many people being falling into the trap of using `ArbInt` when it's not needed.

vext01:
Yes, I just wanted to check that we've thought about this before we continue.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -827,29 +826,13 @@ impl<'a> X64CodeGen<'a> {
...
-            }
-            1 => {
-                let val = bytes.read_i8().unwrap();
-                dynasm!(self.asm; mov Rq(reg.code()), val as i32);
-            }
-            _ => todo!("{}", size),
-        };
+        match self.m.const_(cidx) {
+            jit_ir::Const::I8(x) => dynasm!(self.asm; mov Rq(reg.code()), i32::from(*x)),
+            jit_ir::Const::I16(x) => dynasm!(self.asm; mov Rw(reg.code()), WORD *x),
```
vext01, line 831: Complete bikeshed at this stage of development, but here I think it'd be better to use a 32-bit operand to avoid x86 [register stalls](https://fiigii.com/2020/02/16/Does-register-selection-matter-to-performance-on-x86-CPUs/#Partial-Register-Stall).

And the `DWORD` is redundant I think?

Tests seem to pass with:

```diff
diff --git a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
index 804f0909..faf5e087 100644
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -828,8 +828,8 @@ impl<'a> X64CodeGen<'a> {
     fn load_const(&mut self, reg: Rq, cidx: jit_ir::ConstIdx) {
         match self.m.const_(cidx) {
             jit_ir::Const::I8(x) => dynasm!(self.asm; mov Rq(reg.code()), i32::from(*x)),
-            jit_ir::Const::I16(x) => dynasm!(self.asm; mov Rw(reg.code()), WORD *x),
-            jit_ir::Const::I32(x) => dynasm!(self.asm; mov Rq(reg.code()), DWORD *x),
+            jit_ir::Const::I16(x) => dynasm!(self.asm; mov Rd(reg.code()), i32::from(*x)),
+            jit_ir::Const::I32(x) => dynasm!(self.asm; mov Rq(reg.code()), *x),
             jit_ir::Const::I64(x) => dynasm!(self.asm; mov Rq(reg.code()), QWORD *x),
             jit_ir::Const::Ptr(_) => todo!(),
 ```

ltratt: I copied all this stuff over from the old code and have no idea about x86! It would probably be good if someone who actually knows x86 goes over this stuff in another PR.

vext01: OK.

PR #1216, opened 2024-06-03
GDB support for JITted code.

vext01:
As requested, a work-in-progress (but working) branch to give us GDB support for our JITted code.

Unfinished:
 - C code is messy, needs tidy up.
 - Rust code has no error propagation.
 - GDB plugin not yet hooked into cargo build.
 - Can't break on jitted code symbols for some reason.

The (sparse) documentaton for the GDB JIT interface is here: https://sourceware.org/gdb/current/onlinedocs/gdb.html/JIT-Interface.html

In short, GDB defines and API and ABI for notifying it that new code is appearing. It's bother clever and scary. You define a global variable (I know, I know) with a special name that gdb reads when you call __jit_debug_register_code(). We have no control over the type or scope of this global: gdb expects it a certain way, and we have to play ball.

The docs also say you must synchronise modifications to the internals of the global if you are multi-threaded, which we are. Since we can't wrap the variable in a Rust mutex, we have to use an "out-of-band" mutex and very carefully ensure it stays locked while mutations can still occur.

All tests passing.

### DIFF
--- a/ykrt/build.rs
+++ b/ykrt/build.rs
@@ -1,7 +1,11 @@
 use cfgrammar::yacc::YaccKind;
 use lrlex::{CTLexerBuilder, DefaultLexerTypes};
 
-use std::env;
+use {
+    std::env,
+    std::io::{self, Write},
+    std::process::{exit, Command},
+};
 
 pub fn main() {
     ykbuild::apply_llvm_ld_library_path();
@@ -41,4 +45,15 @@ pub fn main() {
         .unwrap()
         .build()
         .unwrap();
+
+    // Build the gdb plugin.
+    env::set_current_dir("yk_gdb_plugin").unwrap();
+    let out = Command::new("make").output().unwrap();
+    if !out.status.success() {
+        io::stderr().write_all(&out.stdout).unwrap();
+        io::stderr().write_all(&out.stderr).unwrap();
+        exit(1);
+    }
+    env::set_current_dir("..").unwrap();
+    println!("cargo::rerun-if-changed=yk_gdb_plugin/yk_gdb_plugin.c");
 }
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -17,14 +17,18 @@ use super::{
     reg_alloc::{LocalAlloc, RegisterAllocator, StackDirection},
     CodeGen,
 };
+#[cfg(any(debug_assertions, test))]
+use crate::compile::jitc_yk::gdb::GdbCtx;
 use crate::compile::{jitc_yk::jit_ir::IndirectCallIdx, CompiledTrace};
 use dynasmrt::{
     components::StaticLabel, dynasm, x64::Rq, AssemblyOffset, DynasmApi, DynasmError,
     DynasmLabelApi, ExecutableBuffer, Register,
 };
+#[cfg(any(debug_assertions, test))]
+use indexmap::IndexMap;
 use std::sync::Arc;
 #[cfg(any(debug_assertions, test))]
-use std::{cell::Cell, collections::HashMap, error::Error, slice};
+use std::{cell::Cell, error::Error, slice};
 use ykaddr::addr::symbol_to_ptr;
 
 mod deopt;
@@ -75,11 +79,13 @@ pub(crate) struct X64CodeGen<'a> {
     ra: Box<dyn RegisterAllocator>,
     /// Deopt info.
     deoptinfo: Vec<DeoptInfo>,
+    /// Maps assembly offsets to comments.
+    ///
     /// Comments used by the trace printer for debugging and testing only.
     ///
     /// Each assembly offset can have zero or more comment lines.
     #[cfg(any(debug_assertions, test))]
-    comments: Cell<HashMap<usize, Vec<String>>>,
+    comments: Cell<IndexMap<usize, Vec<String>>>,
 }
 
 impl<'a> CodeGen<'a> for X64CodeGen<'a> {
@@ -98,7 +104,7 @@ impl<'a> CodeGen<'a> for X64CodeGen<'a> {
             ra,
             deoptinfo: Vec::new(),
             #[cfg(any(debug_assertions, test))]
-            comments: Cell::new(HashMap::new()),
+            comments: Cell::new(IndexMap::new()),
         }))
     }
 
@@ -148,11 +154,26 @@ impl<'a> CodeGen<'a> for X64CodeGen<'a> {
         // This unwrap cannot fail if `commit` (above) succeeded.
         let buf = self.asm.finalize().unwrap();
 
+        #[cfg(any(debug_assertions, test))]
+        let (comments, gdb_ctx) = {
+            use super::super::gdb;
+            let comments = self.comments.take();
+            let gdb_ctx = gdb::register_jitted_code(
+                self.m.ctr_id(),
+                buf.ptr(AssemblyOffset(0)),
+                buf.size(),
+                &comments,
+            )?;
+            (comments, gdb_ctx)
+        };
+
         Ok(Arc::new(X64CompiledTrace {
             buf,
             deoptinfo: self.deoptinfo,
             #[cfg(any(debug_assertions, test))]
-            comments: self.comments.take(),
+            comments,
+            #[cfg(any(debug_assertions, test))]
+            gdb_ctx,
         }))
     }
 }
@@ -875,7 +896,9 @@ pub(super) struct X64CompiledTrace {
     ///
     /// Used for testing and debugging.
     #[cfg(any(debug_assertions, test))]
-    comments: HashMap<usize, Vec<String>>,
+    comments: IndexMap<usize, Vec<String>>,
+    #[cfg(any(debug_assertions, test))]
+    gdb_ctx: GdbCtx,
 }
 
 impl CompiledTrace for X64CompiledTrace {
@@ -915,12 +938,12 @@ impl CompiledTrace for X64CompiledTrace {
 #[cfg(any(debug_assertions, test))]
 struct AsmPrinter<'a> {
     buf: &'a ExecutableBuffer,
-    comments: &'a HashMap<usize, Vec<String>>,
+    comments: &'a IndexMap<usize, Vec<String>>,
 }
 
 #[cfg(any(debug_assertions, test))]
 impl<'a> AsmPrinter<'a> {
-    fn new(buf: &'a ExecutableBuffer, comments: &'a HashMap<usize, Vec<String>>) -> Self {
+    fn new(buf: &'a ExecutableBuffer, comments: &'a IndexMap<usize, Vec<String>>) -> Self {
         Self { buf, comments }
     }
 
--- a/ykrt/src/compile/jitc_yk/gdb.rs
+++ b/ykrt/src/compile/jitc_yk/gdb.rs
@@ -1,0 +1,275 @@
+//! This implements support for the GDB JIT Compilation Interface described here:
+//! https://sourceware.org/gdb/onlinedocs/gdb/JIT-Interface.html
+//!
+//! This allows gdb to recognise our JITted code, so that we can have higher-level information
+//! (than just raw asm) displayed when debugging traces.
+
+use super::CompilationError;
+use deku::prelude::*;
+use indexmap::IndexMap;
+use std::{
+    ffi::{c_char, c_int, CString},
+    io::Write,
+    ptr,
+    sync::Mutex,
+};
+use tempfile::NamedTempFile;
+
+const TRACE_SYM_PREFIX: &str = "__yk_compiled_trace";
+
+/// JITted code actions.
+///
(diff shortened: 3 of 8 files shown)


ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -106,6 +106,11 @@ impl Module {
         Self::new_internal(ctr_id, global_decls_len)
     }
 
+    #[cfg(any(debug_assertions, test))]
+    pub(crate) fn id(&self) -> u64 {
```
ltratt, line 110: Remember that this ID is not guaranteed to be stable: I think it's worth documenting that (again) on this function (see the doc string comment on the attribute).

ykrt/src/compile/jitc_yk/gdb.rs
```diff
@@ -0,0 +1,241 @@
...
+    let old_first = unsafe { (*desc).first_entry };
+
+    // Make a new linked list node.
+    let mut new_ent = Box::new(JitCodeEntry {
+        next_entry: old_first,
+        prev_entry: ptr::null_mut(),
+        symfile_addr: payload.as_ptr() as *const i8,
+        symfile_size: u64::try_from(payload.len()).unwrap(),
+    });
+    let new_ent_ptr = &mut *new_ent as *mut JitCodeEntry;
```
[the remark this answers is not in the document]
vext01: It's a good question. One I'm not sure the answer of.

The next and prev fields of the linked list nodes are `*mut` because we absolutely do need to mutate them, and I think that forces us to make `&mut`s here?

Is there a way around this?

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
vext01: Must admit I missed that detail.

Would it be hard to make this an overflowing sequential integer? `gdb_c_test` assumes this for this feature:

```
  -b, --num-breaks <NUM_BREAKS>  Set breakpoints at the first `N` compiled traces
```

(this assumes I can get break points on JIT code working. It doesn't seem to work right now).

ykrt/src/compile/jitc_yk/gdb.rs
```diff
@@ -0,0 +1,241 @@
...
+
+/// Due to the way the gdb JIT API works, we are unable to have Rust handle the synchronisation of
+/// `__jit_debug_descriptor` for us.
+unsafe impl Send for JitDescriptor {}
+unsafe impl Sync for JitDescriptor {}
+unsafe impl Send for JitCodeEntry {}
+unsafe impl Sync for JitCodeEntry {}
+
+/// Guards `__jit_debug_descriptor`.
+static LOCK: Mutex<()> = Mutex::new(());
```
jacob-hughes, line 92: You might not be able to wrap `__jit_debug_descriptor` inside a `Mutex` because gdb expects a certain layout, but you can tie it to a Mutex using a `with`-style API. I think this will be better because you can: a) remove a lot of the `unsafe` blocks associated with accessing the global static; and b) ensure you never forget to lock and unlock accesses to `__jit_debug_descriptor`. I've mocked up a little example of what I mean here:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=57000cb92db0673bf5f4d9ba398b3ba3

vext01: I remember discussing this with you a few months back. Seems like a good idea, but I may defer it for a second PR if that's OK?

ltratt: I also had the same worry Jake has: I don't think the `Mutex` is guaranteed to lock what you want it to lock. I didn't have a good solution, though. Jake's seems easy-ish to implement?

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ltratt: I think the current guarantees are the correct ones, and the gdb backend should just accept that things might go a bit wonky if you run things for a billion years or so.

ykrt/src/compile/jitc_yk/gdb.rs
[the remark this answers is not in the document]
ltratt: Have you tried just getting rid of the `&mut`? I *think* `let new_ent_ytr = new_ent as *mut JitCodeEntry` might still let everything compile? If not... you've got to think again I'm afraid. You simply can't have multiple `&mut` references to one thing without being at very high risk of miscompilation due to UB.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ltratt: [You may notice from my mild attempt at humour that I don't think this is an issue in practise for the gdb backend, but the `id` function should document the guarantee -- or lack thereof -- anyway.]

ykrt/src/compile/jitc_yk/gdb.rs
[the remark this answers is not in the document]
vext01:
```
error[E0606]: casting `&JitCodeEntry` as `*mut JitCodeEntry` is invalid
   --> ykrt/src/compile/jitc_yk/gdb.rs:198:23
    |
198 |     let new_ent_ptr = &*new_ent as *mut JitCodeEntry;
```

but:
```
let new_ent_ptr = &*new_ent as *const JitCodeEntry as *mut JitCodeEntry;
```

Works, although it feels like a lie.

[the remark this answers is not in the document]
vext01: > Can you not use Box::into_raw?

I could, but then I'd have to store the pointer for reconstitution later. As it stands, freeing of the box is automatic.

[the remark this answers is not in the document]
ltratt: I was suggesting:

```rust
let new_ent_ytr = new_ent as *mut JitCodeEntry;
```

but you tried 

```rust
let new_ent_ptr = &*new_ent as *mut JitCodeEntry;
```

?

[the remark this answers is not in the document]
ltratt: @jacob-hughes Good point. `Box::into_raw()` is probably the right thing to do here. It avoids `&mut` entirely. And yes you have to free it manually, but as soon as you've got raw pointers, you're more-or-less guaranteeing to do the moral equivalent of "free" anyway.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ltratt: > I guess that's ok, since we'd never gdb a rust test?

Hmm, we might want to do that at some point, but far into the future. Maybe just make `#[cfg(test)]` not compile in the GDB code so that we stop our future selves being confused? Or at least `FIXME` or similar perhaps?

ykrt/src/compile/jitc_yk/gdb.rs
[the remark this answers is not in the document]
vext01: OK, let's do the into_raw() thing then. Ironically I started doing it that way, but decided against it!

For completeness of discussion:
```
error[E0605]: non-primitive cast: `Box<JitCodeEntry>` as `*mut JitCodeEntry`
   --> ykrt/src/compile/jitc_yk/gdb.rs:198:23
    |
198 |     let new_ent_ptr = new_ent as *mut JitCodeEntry;
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ an `as` expression can only be used to convert between primitive types or to coerce to a specific trait object
```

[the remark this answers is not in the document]
ltratt: > OK, let's do the into_raw() thing then. Ironically I started doing it that way, but decided against it!

Cool. The neat thing about doing it this way is that it removes the `&` references to the box so it's harder to accidentally shoot oneself in the foot.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
vext01: We could get the gdb plugin to generate its own sequential numbers, then the problem is uniform for all modes of operation. Since gdb is single-threaded (right?) I don't think we'd even have to synchronise updates to the counter.

(If we did that though, I'm unsure what role `ctr_id` plays any more)

ltratt: We're over-thinking this IMHO: I just think the `id` function in `jit_ir`, which isn't gdb specific, should document what it does. The gdb backend is, in practise, *fine* with the guarantee for the foreseeable future: I just don't want the `id` function to be misused by someone else in the medium term.

ykrt/src/compile/jitc_yk/gdb.rs
vext01: OK, I'll do it from the get-go then.

vext01:
I'm happy that none of the issues that were pointed out were show stoppers. So I'll continue.

vext01:
@ltratt just pushed a tidy-up commit. Would it be easier for you to review if I squashed it?

ykrt/build.rs
```diff
@@ -41,4 +45,18 @@ pub fn main() {
         .unwrap()
         .build()
         .unwrap();
+
+    // Build the gdb plugin.
+    env::set_current_dir("yk_gdb_plugin").unwrap();
+    let out = Command::new("make").output().unwrap();
```
ltratt, line 51: This might sound silly, but we'll probably notice the cost of this. Because all we're really doing is using it to check "is the output of `cc` older than the `.c` input" it's going to be *much* quicker to do this test ourselves in Rust code. I'm not going to worry about it for this PR, but at some point when you have a spare 10 minutes, maybe fix?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -10,19 +10,21 @@
...
     super::{
-        gdb::GdbCtx,
         jit_ir::{self, BinOp, FuncDeclIdx, InstIdx, Operand, Ty},
         CompilationError,
     },
     abs_stack::AbstractStack,
     reg_alloc::{LocalAlloc, RegisterAllocator, StackDirection},
     CodeGen,
 };
+#[cfg(any(debug_assertions, test))]
```
ltratt, line 20: Not something for this PR, but I wondered if this should be enabled only when `yk_testing` is enabled?

ltratt:
Please de-draft and squash.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ltratt: True.

vext01:
squashed.

I'm going to add some comments of my own. There are some parts I want to draw attention to.

ykrt/build.rs
vext01: I'm not sure what you mean. Which test?

Do you mean, don't invoke make, but call the compiler directly?

(I experimented with using the `cc` crate here, but it doesn't work well for this scenario because no rust code is actually linked to what we are building)

ykrt/src/compile/jitc_yk/gdb.rs
```diff
@@ -0,0 +1,275 @@
...
+    ///
+    /// This is a ptr to a Rust `Box` that must be reconstituted to be freed.
+    #[allow(dead_code)]
+    jit_code_entry: *mut JitCodeEntry,
+    /// The serialised payload of the symbol file.
+    #[allow(dead_code)]
+    payload: Vec<u8>,
+}
+
+unsafe impl Send for GdbCtx {}
```
vext01, line 166: These are new since you last saw this PR. This was caused by using `Box::into_raw()` (as we discussed earlier), which means be have to store a raw pointer instead of a Box.

I haven't been able to convince myself it's absolutely safe.

ykrt/build.rs
vext01: Ah wait. I  you mean the timstamp check?

That's handled below by rerun-if-changed.

ykrt/src/compile/jitc_yk/gdb.rs
ltratt: *Ideally* we'd only Send/Sync the pointer, and not the other parts of the struct. A lazy way of doing this would be to store the `*mut` as a `usize` and then you wouldn't need the `unsafe impl`s!

ltratt:
@vext01 This one failed.

vext01:
On it.

ykrt/build.rs
vext01: Ok. We can optimise it later, but even with `make` in the mix, it's in the order of `0.05` seconds.

ltratt: Agreed.

ltratt:
Please squash.

vext01:
squashed.

vext01:
OK to squash? I've checked that's the last clippy-diff warning.

ltratt:
Please squash.

vext01:
squashed.

PR #1217, opened 2024-06-03
Translate phi nodes to the local variable, not an intermediate variable.

ltratt:
Previously a phi node would be converted to an assignment. So, roughly:

```
%2 = phi %0, %1
print %2
```

would, if the `%1` case was taken, be translated to:

```
%2 = %1
print %2
```

but there's no need for the assignment: we can reference the local variable as often as we want.

This commit thus maps a phi node to its local variable without an intermediate assignment. In our running example we thus translate the code directly into:

```
print %1
```

I've also beefed up the phi-node tests considerably, including checking that they work for values returned from functions (which require intermediate variables, though in a different sense to the rest of this commit), and strengthening the text output we actually check for.

### DIFF
--- a/tests/c/phi1.newcg.c
+++ b/tests/c/phi1.newcg.c
@@ -9,12 +9,12 @@
 //     jitstate: stop-tracing
 //     --- Begin aot ---
 //     ...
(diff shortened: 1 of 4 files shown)


ltratt:
Closing in favour of a more comprehensive PR.

PR #1225, opened 2024-06-05
Canonicalise and optimise

ltratt:
This PR adds a (very! very!) simple canonicalisation and (more realistic than before but still very! very!) simple optimisation pass to help guide us as we further alter the system. Doing so has helped me better understand where the JIT IR isn't quite in the right shape (see the commit message in https://github.com/ykjit/yk/commit/535be4b23668ea3c544c4ebb69840f9819cf9650). I still think this PR is worth merging because it will help motivate changes we make to the rest of the system.


ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -923,6 +923,32 @@ impl Const {
...
+    ///
+    /// # Panics
+    ///
+    /// If `x` doesn't fit into the underlying integer type.
+    pub(crate) fn i64_to_int(&self, x: i64) -> Const {
+        match self {
+            Const::I8(_) => Const::I8(i8::try_from(x).unwrap()),
+            Const::I16(_) => Const::I16(i16::try_from(x).unwrap()),
+            Const::I32(_) => Const::I32(i32::try_from(x).unwrap()),
+            Const::I64(_) => Const::I64(i64::try_from(x).unwrap()),
```
vext01, line 948: I think `try_from` can be removed for the `I64` case? Doesn't matter much.

ltratt: It'll be optimised away in release mode. But this method is going to have to change soon, anyway, I think...

vext01:
LGTM.

ltratt:
I'm an idiot and forgot to check with Clippy. OK to squash?

vext01:
go ahead.

ltratt:
Squashed.

vext01:
force push formatted commits :)

ltratt:
I'm an idiot instance number 2. Force pushed.

PR #1231, opened 2024-06-05
Add IR support for volatile memory accesses.

vext01:
We only flag the accesses for now, because we have no optimisations that might like to try to remove dead loads and stores.


ltratt:
Excellent!

PR #1233, opened 2024-06-06
Implement support for 1-bit constant integers.

vext01:
Needed for events.lua.

This is hard to test since we don't have an AOT IR parser yet.

I've been unable to get clang to emit a 1-bit constant integer, but I have a cvise running now to try and find a C test at least.

In the interest of moving onwards, do we want to merge this now, and a test can be added later if we manage to reduce one?


ltratt:
>Needed for events.lua.
[...]
> I've been unable to get clang to emit a 1-bit constant integer, but I have a cvise running now to try and find a C test at least.

I'm not sure I can reconcile these two statements: are we saying "we see 1-bit integers in yklua but we don't know what C code causes them to be emitted"?

vext01:
Correct.

ltratt:
Agreed that a test would be good but we can still go ahead in the interim.

PR #1235, opened 2024-06-07
Move to one `Const::Int` type.

ltratt:
This moves us from multiple `Const::Int{8,16,...}` types to just `Const::Int(ty_idx, val)`. This is easier to use and more closely models what other IRs (e.g. LLVM and Cranelift) do: in essence, integers up to (and including) 64 bits are easy to handle. Bigger sized ints are still something we don't deal with: they'll probably become a `Const::ArbInt` type one day.

A downside to this change is that one now express nonsense integers (e.g. an "8 bit" integer with 9 bits set). The well_formedness check catches those.

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -1255,7 +1255,7 @@ pub(crate) fn const_int_bytes_to_string(num_bits: u32, bytes: &[u8]) -> String {
 #[deku_derive(DekuRead)]
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub(crate) struct IntegerTy {
-    num_bits: u32,
+    pub(crate) num_bits: u32,
 }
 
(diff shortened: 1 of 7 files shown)


vext01:
> A downside to this change is that one now express nonsense integers (e.g. an "8 bit" integer with 9 bits set). The well_formedness check catches those.

Another downside is that we potentially use more memory that is necessary, but I doubt it'll be an issue in practice.

vext01:
You can force push a fix.

ltratt:
Force pushed an update due to Clippy failures.

> Another downside is that we potentially use more memory that is necessary, but I doubt it'll be an issue in practice.

This commit doesn't change memory usage.

[FWIW, compared to the "bag of bytes", we save significant memory (a minimum of: two words + whatever overhead the memory allocator imposes)!]

vext01:
force push a fix :)

ltratt:
Rebased.

PR #1237, opened 2024-06-10
Mul and proxy

ltratt:
After clearing the mess I introduced with canonicalisation (https://github.com/ykjit/yk/commit/4160b063b68f67aac1438762aad752252c2a315f) this PR introduces better `Mul` optimisation, which also requires proxy instructions (https://github.com/ykjit/yk/commit/958c8bf77b06613a92501d1d31d15c263c2d1917). The second commit message (hopefully!) thoroughly explain the motivation for this. It might, though, be easiest to look at the tests (https://github.com/ykjit/yk/blob/958c8bf77b06613a92501d1d31d15c263c2d1917/ykrt/src/compile/jitc_yk/opt/simple.rs#L70) first, as that gives a good intuition to the effect of proxies.


ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -397,14 +397,41 @@ impl fmt::Display for Module {
...
+        let mut inst_iter = self.iter_inst_idxs();
+        while let Some(inst_i) = inst_iter.next(self) {
+            write!(f, "\n    {}", self.insts[inst_i].display(inst_i, self))?;
         }
 
         Ok(())
     }
 }
 
+/// An iterator over instruction indices. This skips "forwarding" instructions: in other words, it
```
ptersilie, line 409: "forwarding"/"proxy"?

ltratt: Fixed in 1a0d2290e3f31999c68a1ffe3d7fdf591dadd98a.

ptersilie:
I have nothing profound to say. This seems like the right idea and we'll probably find out quickly if this works well enough or if we need to find another approach.

Please squash,

ltratt:
Squashed.

PR #1241, opened 2024-06-11
Fix sdiv implementation.

ptersilie:
This fixes a bug with the sdiv implementation. The reason for the bug was that I misunderstood some details about `idiv`.

### DIFF
--- a/tests/c/sdiv.newcg.c
+++ b/tests/c/sdiv.newcg.c
@@ -5,10 +5,14 @@
 //   env-var: YKD_LOG_JITSTATE=-
 //   stderr:
 //     jitstate: start-tracing
-//     sdiv 10922
-//     sdiv2 715827882
-//     sdiv3 1431655764
-//     sdiv4 *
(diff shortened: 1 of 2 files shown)


ltratt:
Can we add tests to x86_64/mod.rs about the generated code? Those are more robust tests than the lang_tests for these sorts of thing (and the JIT IR parser & fm incombination make creating such tests pretty easy).

ptersilie:
I can, but it will require me to add at least 8 unit tests to get the same coverage as the langtest has now.

ptersilie:
I improved the langtest a bit and simplified the implementation for byte-sized values. I also added a unit test. Let me know if you want me to add unit tests for the other 7 cases.

ltratt:
In general I would love it if we unit tested all the cases, because that will really help us going forwards, but I will leave that to your discretion until yklua is running. Please squash if you want.

ptersilie:
Squashed (I kinda wanna push on with things and look into testing later).

PR #1242, opened 2024-06-11
Optimise `ICmp` when both sides are constants

ltratt:
The ultimate aim of this PR is to optimise `ICmp`s when both sides are constants (https://github.com/ykjit/yk/commit/e98a2c904d1908e1a9da7fee7e669c57bcb639f8), but doing so for all predicates requires quite a bit of preparatory work.


ptersilie:
Nice. I presume this is also in preparation to another optimisation that removes guards?

ltratt:
> I presume this is also in preparation to another optimisation that removes guards?

Correct! That will require some other stuff to be done first, which is why I'd prefer to do it in another PR.

PR #1243, opened 2024-06-11
Implement udiv.

ptersilie:
Needs rebasing after #1241 has gone in.


ltratt:
Please rebase that one when the #1241 has gone in.

PR #1245, opened 2024-06-11
Optimise multiply chains

ltratt:
The main part of this commit is https://github.com/ykjit/yk/commit/0145e7ba45913a694052c9d4731b764a2e8be5fa, which allows us to optimise chains of multiplications of constants, but first of all I had to fix a "what was I thinking?" bug (https://github.com/ykjit/yk/commit/4209c61cb914eb83aaadc70dee9011328d2a1192).

### DIFF
--- a/ykrt/src/compile/jitc_yk/opt/simple.rs
+++ b/ykrt/src/compile/jitc_yk/opt/simple.rs
@@ -45,7 +45,7 @@ fn opt_mul(
                 } else if y == 1 {
                     // Replace `x * 1` with `x`.
                     m.replace(inst_i, Inst::ProxyInst(mul_inst));
-                } else if y % 2 == 0 {
+                } else if y & (y - 1) == 0 {
                     // Replace `x * y` with `x << ...`.
                     let shl = u64::from(y.ilog2());
                     let new_const = Operand::Const(m.insert_const(old_const.u64_to_int(shl))?);
@@ -207,10 +207,7 @@ mod tests {
             %2: i64 = mul %0, 4i64
             %3: i64 = mul %0, 4611686018427387904i64
             %4: i64 = mul %0, 9223372036854775807i64
-            black_box %1
-            black_box %2
-            black_box %3
-            black_box %4
+            %5: i64 = mul %0, 12i64
         ",
             |m| simple(m).unwrap(),
             "
@@ -221,10 +218,7 @@ mod tests {
             %2: i64 = shl %0, 2i64
             %3: i64 = shl %0, 62i64
             %4: i64 = mul %0, 9223372036854775807i64
-            black_box %1
-            black_box %2
-            black_box %3
-            black_box %4
+            %5: i64 = mul %0, 12i64
         ",
         );
     }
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/jitc_yk/opt/simple.rs
```diff
@@ -45,7 +45,7 @@ fn opt_mul(
                 } else if y == 1 {
                     // Replace `x * 1` with `x`.
                     m.replace(inst_i, Inst::ProxyInst(mul_inst));
-                } else if y % 2 == 0 {
+                } else if y & (y - 1) == 0 {
```
ptersilie: Clever!

ykrt/src/compile/jitc_yk/opt/simple.rs
```diff
@@ -34,24 +34,50 @@ fn opt_mul(
...
-        | (Operand::Const(mul_const), Operand::Local(mul_inst)) => {
+        (Operand::Local(mut mul_inst), Operand::Const(mul_const))
+        | (Operand::Const(mul_const), Operand::Local(mut mul_inst)) => {
             let old_const = m.const_(mul_const);
-            if let Some(y) = old_const.int_to_u64() {
-                if y == 0 {
+            if let Some(old_val) = old_const.int_to_u64() {
+                let mut new_val = old_val;
+                // If we've got `%2: mul %1, xi8` then see if `%1` is of the form `mul %0, yi8`: if so
+                // we've got a chain that's `%2: %0 x*y`. We can thus "skip" the intermediate `mul`
```
ptersilie, line 43: I think the chain should be `%2: %0*y*x`.

ptersilie:
Just a typo. I see that due to the structure of traces, this will naturally merge multiple (>2) multiplications despite the optimisation only looking at 2 at a time. However, I believe a mul without constants in the middle can block some merging, e.g.
```
%1 = %0 * 2;
%2 = %1 * %1;
%3 = %2 * 2;
```
which otherwise could be optimised to:
```
%2 = %1 * %1;
%3 = %2 * 4;
```

ykrt/src/compile/jitc_yk/opt/simple.rs
ltratt: `*` is commutative, so the two are equivalent.

ltratt:
> However, I believe a mul without constants in the middle can block some merging, e.g.

Correct. I suspect we'll want to do a more clever tracking mechanism to spot such things, but we can get there bit-by-bit.

ykrt/src/compile/jitc_yk/opt/simple.rs
ptersilie: Fine, but you're still missing a `*`.

ltratt: Fixed in d668e29917b0dfc697329338a83520193b1e82fa.

ptersilie:
Please squash.

ltratt:
Squashed.

PR #1247, opened 2024-06-13
Enable deoptimisation of inlined constants.

ptersilie:
I'm raising this as a DRAFT, as I'm not quite sold on the design yet and would like to have some additional eyes on this. There's two main issues:
1) We need to introduce a `LocalAlloc::Const`, which doesn't make much sense conceptually.
2) We currently don't adjust for the size of the constant (this can be easily fixed by adding the size to the enum variant)

---

Due to function inlining (and later other optimisations) it can happen that a variable in AOT becomes a constant in the JIT. Currently, neither the trace builder nor deoptimisation can handle such cases, which this commit fixes, by inserting a `ProxyConst` instruction and reference this instead of the direct constant.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
@@ -31,6 +31,7 @@ pub(crate) enum LocalAlloc {
     ///
     /// FIXME: unimplemented.
     Register,
+    ConstInt(u64),
 }
 
 impl LocalAlloc {
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -19,7 +19,10 @@ use super::{
 };
 #[cfg(any(debug_assertions, test))]
 use crate::compile::jitc_yk::gdb::GdbCtx;
-use crate::compile::{jitc_yk::jit_ir::IndirectCallIdx, CompiledTrace};
+use crate::compile::{
+    jitc_yk::jit_ir::{Const, IndirectCallIdx},
+    CompiledTrace,
+};
 use dynasmrt::{
     components::StaticLabel, dynasm, x64::Rq, AssemblyOffset, DynasmApi, DynasmError,
     DynasmLabelApi, ExecutableBuffer, Register,
@@ -817,7 +820,20 @@ impl<'a> X64CodeGen<'a> {
         let mut locs: Vec<LocalAlloc> = Vec::new();
         let gi = inst.guard_info(self.m);
         for lidx in gi.lives() {
-            locs.push(*self.ra.allocation(*lidx));
+            match self.m.inst(*lidx) {
+                jit_ir::Inst::ProxyConst(c) => {
+                    // The live variable is a constant (e.g. this can happen during inlining), so
+                    // it doesn't have an allocation. We can just push the actual value instead
+                    // which will be written as is during deoptimisation.
+                    match self.m.const_(*c) {
+                        Const::Int(_, c) => locs.push(LocalAlloc::ConstInt(*c)),
+                        _ => todo!(),
+                    };
+                }
+                _ => {
+                    locs.push(*self.ra.allocation(*lidx));
+                }
+            }
         }
 
         // FIXME: Move `frames` instead of copying them (requires JIT module to be consumable).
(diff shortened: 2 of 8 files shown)


ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
```diff
@@ -31,6 +31,7 @@ pub(crate) enum LocalAlloc {
     ///
     /// FIXME: unimplemented.
     Register,
+    Const(u64),
```
ltratt, line 34: I think this needs to be `ConstInt` as we'll have other other `Const` things (e.g. pointers) later.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -817,7 +820,21 @@ impl<'a> X64CodeGen<'a> {
         let mut locs: Vec<LocalAlloc> = Vec::new();
         let gi = inst.guard_info(self.m);
         for lidx in gi.lives() {
-            locs.push(*self.ra.allocation(*lidx));
+            match self.m.inst(*lidx) {
+                jit_ir::Inst::ProxyConst(c) => {
```
ltratt, line 824: When we see a guard with a constant, we can just not emit the guard entirely. [The optimiser will remove such guards, but if optimisations are turned off, we will get such guards. But there's no point codegenning them.] That will simplify this method a lot I think!

ltratt:
I have a couple of comments that are probably worth addressing, but overall I'd say: let's just do this. It might turn out to be imperfect, and then we can iterate on it in-tree.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ptersilie: This is not the condition of the guard. It's a live variable that needs to be tracked and deoptimised.

ltratt: Ah!

ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
ptersilie: Are you fine with just changing this to `ConstInt` and not worry about the different sizes for now?

ltratt: I'm fine fleshing it out later, I assume we have an `assert` (or similar) somewhere that means we can only get to this point for `u64` sized things? What we don't want to do is compile code that silently overwrites (say) two adjacent `u8`s with a `u64`.

ltratt:
The other thing we'll want to add to this PR is at least one unit test to x86_64/mod.rs for the const case.

ptersilie:
> The other thing we'll want to add to this PR is at least one unit test to x86_64/mod.rs for the const case.

Sure. Can the JIT IR parser deal with `ProxyConst`s or do I need to add that first?

ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
ptersilie: We don't. I don't want to add an `assert` that will likely fail lua tests. We are doing the same with loading an 8bit local. It just gets zero extended to the full register. So I think writing the whole u64 to the register is no different.

ltratt:
> Can the JIT IR parser deal with ProxyConsts or do I need to add that first?

Good point. No, it doesn't deal with them because I originally thought "they can't appear in the output therefore they shouldn't be parsed". But I think that was wrong: they should be parsed even though they can't appear in the output. I would suggest `%0: i8 = 1i8`esque syntax for this.

[I think we're tending in a direction where `ProxyConst` will become `AssignConst`, but we can worry about that later.]

ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
ltratt: > So I think writing the whole u64 to the register is no different.

What happens when we deopt a heap value?

ptersilie:
Yeah I'm also starting to think a normal assign instruction isn't a bad idea. It would make the special `LocalAlloc::Const` case go away too.

ltratt:
For this PR I think I would stick with `ProxyConst` just because there will be more knock-on effects from changing it than are first apparent, and such a change is probably best done in isolation.

ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
ltratt: If that's the case why do we need to worry about the size of things in a guard?

ptersilie: I'm not sure what you mean. I don't think that's something we worry about. This discussion was specifically about this line https://github.com/ykjit/yk/pull/1247/files#diff-150b793f9518b19505892d1496e8641dc1de27061360bfb755cb1c85159f3eaeR910. Which happens whenever we load a local that is actually a `ProxyConst`.

ltratt: Resolved offline.

ltratt:
I think it's just `Const` -> `ConstInt` and the unit test (and JIT IR parser) change needed to get this in.

ptersilie:
This should address the remaining issues.

ltratt:
Please squash and mark the PR as "ready for review".

ptersilie:
Squashed.

ptersilie:
Force pushed some C formatting changes.

ptersilie:
Sorry, forgot to run the other tests. Doing so revealed that the instruction printing needed changing to deal with proxy stuff.

ltratt:
Hmm, not totally convinced by that new patch, but I'm also unsure what the right thing should be! Let's get it in and work out what to do later.

Please squash.

ptersilie:
Yeah, I must admit I wasn't too sure either.

Squashed.

PR #1248, opened 2024-06-13
Implement shl.

ptersilie:
This adds two more lua tests to our list of working tests, bringing that number up to 12 out of 18.

Needs rebasing after #1247 has gone in.


ptersilie:
Rebased.

PR #1249, opened 2024-06-14
Adjust Yk tests to handle Mem2Reg Changes

nmdis1999:
This PR aims to fix yk test cases so that we can check the right IR when enabling mem2reg pass. There is another test case `arithmetic.newgc.c` which fails at tuntime with error: 
```
thread '<unnamed>' panicked at ykrt/src/compile/jitc_yk/aot_ir.rs:502:13:
explicit panic
```

here is the reduced test case which reporduce this error:
```
#include <yk_testing.h>
void main(int argc) {
  YkMT *mt = yk_mt_new(NULL);
  while (4) {
    fprintf(stderr, "", argc);
  }
}
```


tests/c/switch_non_default.newcg.c
```diff
@@ -9,8 +9,6 @@
 //     jitstate: stop-tracing
 //     --- Begin aot ---
 //     ...
-//     switch %{{10_1}}, bb{{bb14}}, [300 -> bb{{bb11}}, 299 -> bb{{bb12}}] [safepoint: {{safepoint_id}}, (%0_0, %0_1, %0_3, %0_4, %0_5, %0_6, %10_1)]
```
ltratt, line 12: Removing this line means the test isn't really testing anything. So *something* should go here instead, right?

ltratt:
@ptersilie @vext01 Any ideas what this function should do instead of `panic`?

```rust
    /// Return the `InstID` of a local variable operand. Panics if called on other kinds of
    /// operands.
    pub(crate) fn to_inst_id(&self) -> InstID {
        let Self::LocalVariable(iid) = self else {
            panic!()
        };
        iid.clone()
    }
```

tests/c/phi3.newcg.c
```diff
@@ -1,40 +1,7 @@
 // ignore-if: test $YK_JIT_COMPILER != "yk" -o "$YKB_TRACER" = "swt"
 // Run-time:
-//   env-var: YKD_LOG_IR=-:aot,jit-pre-opt
 //   env-var: YKD_LOG_JITSTATE=-
 //   env-var: YKD_LOG_STATS=/dev/null
-//   stderr:
```
ptersilie, line 6: This seems a bit of a drastic change. This essentially removes everything we're trying to test. What was failing about this test?

ptersilie:
> @ptersilie @vext01 Any ideas what this function should do instead of `panic`?
> 
> ```rust
>     /// Return the `InstID` of a local variable operand. Panics if called on other kinds of
>     /// operands.
>     pub(crate) fn to_inst_id(&self) -> InstID {
>         let Self::LocalVariable(iid) = self else {
>             panic!()
>         };
>         iid.clone()
>     }
> ```

It makes no sense to call `to_inst_id` on anything that isn't an instruction. So `panic` seems like the right choice here? Are we seeing this panic? I noticed some constant propagation happening in the changes, so it could be related to that: we assume something is a function, but it's a constant instead. We should print `self` here to see why type we are getting here.

tests/c/phi3.newcg.c
nmdis1999: oh this is an accidental push, will correct this.

ltratt:
> Are we seeing this panic?

Yes, see the opening message to this PR, which shows the panic occurring. I agree that it's quite probably because something is becoming a constant. My guess is that addressing that isn't rocket science, and anything we can turn from a variable into a constant at AOT is a bonus!

vext01:
> @ptersilie @vext01 Any ideas what this function should do instead of panic?

I think panic is correct here.

The function is designed to return the instruction that defines the local variable "referenced" by the operand. Instructions only define local variables, so it makes no sense to call this on any other kind of operand.

ltratt:
I'm confused: it sounds like we're saying "panic is the right thing" as if we're right and the LLVM IR is wrong? Can we not think about how we might potentially be able to alter some part of the system to do The Right Thing? After all, what this PR is really showing is "when you turn on optimisations, LLVM IR is of a different shape than we've previously considered". I'm not saying we should necessarily fix it right now, but maybe we can give @nmdis1999 a hint of what to look at?

vext01:
The LLVM IR is unlikely to be incorrect. It's more likely a lowering bug.

To understand better, I think you'd need to look at the LLVM IR and our IR side by side. You might want to put print statements in the Yk IR serialiser to see which paths are taken and how certain LLVM constructs are lowered.

ltratt:
> It's more likely a lowering bug.

Aha!

@nmdis1999 do you think you can follow Edd's debugging suggestion? It sounds like it's *probably* a bug in ykllvm: the good news is that you have a nice, small, test case to use to hunt the problem down.

nmdis1999:
> #include <yk_testing.h>
> void main(int argc) {
>   YkMT *mt = yk_mt_new(NULL);
>   YkLocation loc = yk_location_new();
>   while (4) {
>     yk_mt_control_point(mt, &loc);
>     fprintf(stderr, "", argc);
>   }
> }

Printing self gives following output:
```
add 5
sub 3
mul 12
add 4
sub 2
mul 9
add 3
sub 1
mul 6
add 2
sub 0
mul 3
exit
[ykrt/src/compile/jitc_yk/aot_ir.rs:501:40] self = LocalVariable(
    InstID {
        func_idx: FuncIdx(
            2,
        ),
        bb_idx: BBlockIdx(
            0,
        ),
        inst_idx: InstIdx(
            0,
        ),
    },
)
[ykrt/src/compile/jitc_yk/aot_ir.rs:501:40] self = LocalVariable(
    InstID {
        func_idx: FuncIdx(
            2,
        ),
        bb_idx: BBlockIdx(
            1,
        ),
        inst_idx: InstIdx(
            2,
        ),
    },
)
[ykrt/src/compile/jitc_yk/aot_ir.rs:501:40] self = LocalVariable(
    InstID {
        func_idx: FuncIdx(
            2,
        ),
        bb_idx: BBlockIdx(
            0,
        ),
        inst_idx: InstIdx(
            4,
        ),
    },
)
[ykrt/src/compile/jitc_yk/aot_ir.rs:501:40] self = LocalVariable(
    InstID {
        func_idx: FuncIdx(
            2,
        ),
        bb_idx: BBlockIdx(
            0,
        ),
        inst_idx: InstIdx(
            3,
        ),
    },
)
[ykrt/src/compile/jitc_yk/aot_ir.rs:501:40] self = LocalVariable(
    InstID {
        func_idx: FuncIdx(
            2,
        ),
        bb_idx: BBlockIdx(
            0,
        ),
        inst_idx: InstIdx(
            1,
        ),
    },
)
[ykrt/src/compile/jitc_yk/aot_ir.rs:501:40] self = Arg {
    func_idx: FuncIdx(
        2,
    ),
    arg_idx: ArgIdx(
        0,
    ),
}
```
I have narrowed down this problem, it seems that something goes wrong when we use `yk_mt_drop` function in this program. So, fwiu the crash only happens after the loop has been processed. @vext01 @ltratt @ptersilie

ltratt:
@nmdis1999 Hang on, originally you got a crash in `aot_ir.rs` but now you're saying the crash happens at runtime? [Note: `yk_mt_drop` is not recognised as a special function by any part of ykllvm/yk.]

nmdis1999:
I did get crash in `aot_ir.rs` I mean that the I see the output of `for` loop on stdout and then a crash. Sorry for `yk_mt_drop` line, I think I misunderstood how it works.

ptersilie:
What's the crash? Does it have an error/line number? Looks like it might be related to `Arg` which we might not deal with properly at the moment.

nmdis1999:
> What's the crash? Does it have an error/line number? Looks like it might be related to `Arg` which we might not deal with properly at the moment.

@ptersilie  Here:thread '<unnamed>' panicked at ykrt/src/compile/jitc_yk/aot_ir.rs:502:13:
explicit panic

And yes, it is due to an `Arg` I have added the dbg print for that line in above comment.

nmdis1999:
@ptersilie aot IR:
```
# IR format version: 0
# Num funcs: 14
# Num consts: 9
# Num global decls: 6
# Num types: 21
global_decl @shadowstack_0
global_decl @stderr
global_decl @.str
global_decl @.str.1
global_decl @.str.2
global_decl @.str.3

func foo(%arg0: i32) -> i32 {
  bb0:
    %0_0: ptr = load @shadowstack_0
    br bb1
  bb1:
    %1_0: i32 = add %arg0, 3i32
    br bb3
  bb2:
    ret %1_0
  bb3:
    br bb2
}

func llvm.dbg.value(%arg0: ?ty<metadata>, %arg1: ?ty<metadata>, %arg2: ?ty<metadata>);

func main(%arg0: i32, %arg1: ptr) -> i32 {
  bb0:
    %0_0: ptr = alloca {0: i32, 64: ptr, 128: ptr, 192: ptr, 256: ptr}, 1, 8
    %0_1: ptr = call malloc(1000000i64)
    *@shadowstack_0 = %0_1
    %0_3: ptr = alloca {0: i64}, 1, 8
    %0_4: ptr = ptr_add %0_1, 0
    br bb1
  bb1:
    %1_0: ptr = ptr_add %0_1, 16
    *@shadowstack_0 = %1_0
    %1_2: ptr = call yk_mt_new(0x0)
    br bb2
  bb2:
    *@shadowstack_0 = %0_1
    %2_1: ptr = ptr_add %0_1, 16
    *@shadowstack_0 = %2_1
    call yk_mt_hot_threshold_set(%1_2, 0i32)
    br bb3
  bb3:
    *@shadowstack_0 = %0_1
    %3_1: ptr = ptr_add %0_1, 16
    *@shadowstack_0 = %3_1
    %3_3: i64 = call yk_location_new()
    br bb4
  bb4:
    *@shadowstack_0 = %0_1
    *%0_3 = %3_3
    *%0_4 = 2i32
    br bb5
  bb5:
    %5_0: i32 = load %0_4
    br bb6
  bb6:
    br bb7
  bb7:
    %7_0: i32 = load %0_4
    %7_1: i1 = sgt %7_0, 0i32
    condbr %7_1, bb8, bb13 [safepoint: 0i64, (%arg0, %0_0, %0_1, %0_3, %0_4, %1_2, %7_1)]
  bb8:
    %8_0: ptr = ptr_add %0_1, 16
    *@shadowstack_0 = %8_0
    %8_2: ptr = ptr_add %0_0, 0
    *%8_2 = %arg0
    %8_4: ptr = ptr_add %0_0, 8
    *%8_4 = %0_1
    %8_6: ptr = ptr_add %0_0, 16
    *%8_6 = %0_3
    %8_8: ptr = ptr_add %0_0, 24
    *%8_8 = %0_4
    %8_10: ptr = ptr_add %0_0, 32
    *%8_10 = %1_2
    %8_12: ptr = call llvm.frameaddress.p0(0i32)
    call __ykrt_control_point(%1_2, %0_3, %0_0, %8_12)
    br bb9
  bb9:
    *@shadowstack_0 = %0_1
    %9_1: i32 = load %0_4
    %9_2: i32 = add %9_1, %arg0
    %9_3: i32 = sub %9_1, %arg0
    %9_4: i32 = mul %9_1, %arg0
    %9_5: i32 = mul %9_4, 3i32
    %9_6: ptr = load @stderr
    %9_7: ptr = ptr_add %0_1, 16
    *@shadowstack_0 = %9_7
    %9_9: i32 = call fprintf(%9_6, @.str, %9_2)
    br bb10
  bb10:
    *@shadowstack_0 = %0_1
    %10_1: ptr = load @stderr
    %10_2: ptr = ptr_add %0_1, 16
    *@shadowstack_0 = %10_2
    %10_4: i32 = call fprintf(%10_1, @.str.1, %9_3)
    br bb11
  bb11:
    *@shadowstack_0 = %0_1
    %11_1: ptr = load @stderr
    %11_2: ptr = ptr_add %0_1, 16
    *@shadowstack_0 = %11_2
    %11_4: i32 = call fprintf(%11_1, @.str.2, %9_5)
    br bb12
  bb12:
    *@shadowstack_0 = %0_1
    %12_1: i32 = load %0_4
    %12_2: i32 = add %12_1, -1i32
    *%0_4 = %12_2
    br bb7
  bb13:
    %13_0: ptr = load @stderr
    %13_1: ptr = ptr_add %0_1, 16
    *@shadowstack_0 = %13_1
    %13_3: i64 = call fwrite(@.str.3, 5i64, 1i64, %13_0)
    br bb14
  bb14:
    *@shadowstack_0 = %0_1
    br bb16
  bb15:
    ret 0i32
  bb16:
    br bb15
}

func yk_mt_new(%arg0: ptr) -> ptr;

func yk_mt_hot_threshold_set(%arg0: ptr, %arg1: i32);

func llvm.dbg.declare(%arg0: ?ty<metadata>, %arg1: ?ty<metadata>, %arg2: ?ty<metadata>);

func yk_location_new() -> i64;

func yk_mt_control_point(%arg0: ptr, %arg1: ptr);

func fprintf(%arg0: ptr, %arg1: ptr, ...) -> i32;

func fwrite(%arg0: ptr, %arg1: i64, %arg2: i64, %arg3: ptr) -> i64;

func malloc(%arg0: i64) -> ptr;

func __ykrt_control_point(%arg0: ptr, %arg1: ptr, %arg2: ptr, %arg3: ptr);

func llvm.frameaddress.p0(%arg0: i32) -> ptr;

func llvm.experimental.stackmap(%arg0: i64, %arg1: i32, ...);
```

tests/c/indirect_call.newcg.c
```diff
@@ -9,8 +9,6 @@
 //     jitstate: stop-tracing
 //     --- Begin jit-pre-opt ---
 //     ...
-//     %{{1}}: i32 = icall %{{2}}(%{{3}})
```
nmdis1999, line 12: @ltratt mem2reg optimise away the register which loads for `%3`.
For example: the entry block before mem2reg has `%3`, and `%11 ` loads it.
```
entry:
    %0: ptr = arg(0)
    %1: ptr = load_ti 40
    %2: ptr = load_ti 32
    %3: ptr = load_ti 24
    %4: ptr = load_ti 16
    %5: ptr = load_ti 8
    %6: ptr = load_ti 0
    tloop_start:
    %8: ptr = lookup_global @shadowstack_0
    *%8 = %6
    %10: ptr = load %2
    %11: i32 = load %3
    %12: ptr = ptr_add %6, 16
    %13: ptr = lookup_global @shadowstack_0
    *%13 = %12
    %15: i32 = icall %10(%11)
    %16: ptr = lookup_global @shadowstack_0
    *%16 = %6
```
However, when we enable `mem2reg` `%3` is removed so we don't have `%11` which was loading it and thus `icall` is optimised out.
```
entry:
   %0: ptr = arg(0)
   %1: ptr = load_ti 24
   %2: ptr = load_ti 16
   %3: ptr = load_ti 8
   %4: ptr = load_ti 0
   tloop_start:
   %6: ptr = lookup_global @shadowstack_0
   *%6 = %4
   %8: i32 = load %2
   %9: ptr = ptr_add %4, 16
   %10: ptr = lookup_global @shadowstack_0
   *%10 = %9
```

so seems like removing this line is right thing to do if we are turning on mem2reg?

tests/c/dyn_ptradd_multidim.newcg.c
```diff
@@ -14,9 +14,9 @@
 //     --- End aot ---
 //     --- Begin jit-pre-opt ---
 //     ...
-//     %{{16}}: ptr = dyn_ptr_add %{{3}}, %{{_}}, {{64}}
```
nmdis1999, line 17: @ltratt we need to change the register value else the test fails when we turn on `mem2reg`.

ltratt: What does "change the register value" mean? The numbers inside `{{...}}` are for fm, and have no semantic meaning other than "the same name must have the same value each time it is referenced" (please read fm's docs on name matching).

tests/c/indirect_call.newcg.c
ltratt: Huh, this looks like a bug to me, but the bug *might* be in the test.

Can you try adding `NOOPT_VAL(fn)` just before the `while` loop and see if the `icall` is still optimised away?

nmdis1999: @ltratt adding `NOOPT_VAL(fn)` fixes `icall` being optimised out. Thanks.

tests/c/dyn_ptradd_multidim.newcg.c
nmdis1999: Sorry, one of the load is optimised out and everything shifts up so dyn_ptr_add check starts from %{{15}} instead of  %{{16}}. We can instead change the check to do `%{{_}}: ptr = dyn_ptr_add %{{3}}, %{{_}}, {{64}}` instead so if any loads/stores are optimised out we don't need to change the value at `{{}}`?

ltratt: Have you read the fm documentation?

nmdis1999: I read the fm docs (https://docs.rs/fm/0.3.0/fm/index.html), am not sure what am I missing? If a load is optimised out then wouldn't we have lesser number of numerical names with assigned values? And now `%16` is assigned with different value than we expected?

ltratt: What happens if you revert these lines? Does the test pass or fail?

nmdis1999: the test fails

ltratt: I'm afraid you'll have to debug why (or explain to me why I'm wrong!) as the changed test seems identical to me. AFAICS the two tests are identical modulo alpha-renaming, which shouldn't affect fm's matching.

vext01: One thing that I do when I can't figure out why something doesn't match, is I re-paste+amend the lines in question from the actual output. Usually the problem makes itself apparent.

nmdis1999: @ltratt We have multiple lines with `dyn_ptr_add` in the test (one below the other). Now, if the trace changes such that : 
```
%{{16}}: ptr = dyn_ptr_add %{{3}}, %{{_}}, {{64}}
```
becomes 
```
%{{15}}: ptr = dyn_ptr_add %{{3}}, %{{_}}, {{64}}
```
and 
```
%17: ptr = dyn_ptr_add %{{16}}, %{{_}}, {{16}}
```
becomes
```
%16: ptr = dyn_ptr_add %{{15}}, %{{_}}, {{16}}
```
fwiu the test thinks both `dyn_ptr_add` assignments are same and picks the second assignment to start the check.
So, instead of checking:
```
  -> %15: ptr = dyn_ptr_add %3, %12, 64
       %16: ptr = dyn_ptr_add %15, %13, 16
       %17: ptr = dyn_ptr_add %16, %14, 4
```
it starts the check at:
```
     %15: ptr = dyn_ptr_add %3, %12, 64
 -> %16: ptr = dyn_ptr_add %15, %13, 16
     %17: ptr = dyn_ptr_add %16, %14, 4

```

ltratt: The numbers in the brackets have no semantic meaning as such. You can replace every instance of `{{16}}`, for example, with `{{banana}}` (i.e. alpha rename) and you should have the same test. If that isn't the case, it's a bug (probably in fm).

[Warning: I think your change is just alpha renaming, but maybe I'm misreading it.]

nmdis1999: It looks like this is a `fm` bug. Should I file a bug report?

ltratt: Yes but you will need to provide a minimal test case in order for me to look at it.

nmdis1999:
ready for review @ltratt @vext01

tests/c/phi3.newcg.c
```diff
@@ -20,7 +20,7 @@
 //       %{{15_2}}: i32 = call g() ...
 //     ..~
 //     bb{{_}}:
-//       %{{_}}: i32 = phi bb{{_}} -> %{{15_2}}, bb{{_}} -> %{{13_2}}
+//       %{{_}}: i32 = phi bb{{_}} -> %{{13_2}}, bb{{_}} -> %{{15_2}}
```
ltratt, line 23: This test has changed more than is first obvious: notice that line 20 previously referenced `15_2`, but now it's referencing `13_2` (both of these are defined earlier in the test). Do we know why?

tests/c/simple_interp_loop2.newcg.c
```diff
@@ -13,7 +13,6 @@
 //     --- Begin jit-pre-opt ---
 //     ..~
 //     guard %{{_}}, true
-//     %{{_}}: ptr = load %{{8}}
```
ltratt, line 16: Do we know why this has disappeared?

tests/c/phi3.newcg.c
nmdis1999: Looks like mem2reg when iterating over basic blocks and since basic block 13 comes before 15 it renames the phi statement. The test case is failing because we are also checking the value of `%15_2` and `%13_2` which have different values, so the test expects same value in the phi statement.

tests/c/simple_interp_loop2.newcg.c
nmdis1999: `%8` is loaded at the begining of the entry block, seems like `mem2reg` removed redudant load here.

tests/c/phi3.newcg.c
ltratt: This is an observation but not an explanation: notice that `13_2` is the result of calling `f` but `15_2` is the result of calling `g`. We need a more considered explanation of why this change is OK. What we don't want to do is introduce an optimisation that works for our test cases just by accident. You will probably need to look at the full trace IR for this example and then use that to give an explanation.

tests/c/simple_interp_loop2.newcg.c
ltratt: I understand the observation, but we need to understand *why* this is correct.

vext01: Based on my understanding of mem2reg, this seems legit.

The purpose of mem2reg is to replace allocations (and loads and stores from/to them) with SSA variables and PHI nodes (if necessary).

https://llvm.org/docs/Passes.html#mem2reg-promote-memory-to-register

tests/c/phi3.newcg.c
nmdis1999: Phi nodes select the right value based on which predecessor block the control flow came from, regardless of the order. Since mem2reg adds/reconstructs phi nodes it changes the order in which it sees the basic blocks. But this should not be a problem for correctness as the order in phi node doesn't matter anyways.

ltratt: OK that makes sense.

PR #1251, opened 2024-06-16
Minimally allow `cargo check` to work on non-Linux.

ltratt:
Although yk can only run on Linux at the moment, `cargo check` fails on non-Linux, which means that rust-analyzer fails. Having got sufficiently fed up with not having any LSP help, I finally got around to working out the minimal (hopefully) changes necessary for `cargo check` to succeed on OpenBSD (but `cargo test` doesn't work, and I'm not worried about that for now).

Most of these changes are probably good things full stop -- it makes clearer that `hwt` can only work on x86_64/Linux, for example. One annoying thing is that Cargo workspaces can't be conditionally compiled, so `hwtracer` must always be compilable. It's easy to make its `lib.rs` not compile anything useful, but harder for `build.rs`. To add insult to injury `build::main` must always be a top-level function. I've therefore resorted to a slightly evil "inner mod" trick for `hwtracer/build.rs`

Still, overall, these changes aren't too intrusive, and they will probably help us when eventually we want to get yk actually working on non-Linux.

### DIFF
--- a/hwtracer/build.rs
+++ b/hwtracer/build.rs
@@ -1,36 +1,45 @@
-use rerun_except::rerun_except;
-use std::env;
-use std::path::PathBuf;
-use ykbuild::{completion_wrapper::CompletionWrapper, ykllvm_bin};
+#[cfg(not(target_os = "linux"))]
+mod inner {
+    use rerun_except::rerun_except;
 
-const FEATURE_CHECKS_PATH: &str = "feature_checks";
-
-/// Simple feature check, returning `true` if we have the feature.
-///
-/// The checks themselves are in files under `FEATURE_CHECKS_PATH`.
-fn feature_check(filename: &str, output_file: &str) -> bool {
-    let mut path = PathBuf::new();
-    path.push(FEATURE_CHECKS_PATH);
-    path.push(filename);
-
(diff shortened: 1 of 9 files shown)


vext01:
That binary prints the embedded LLVM IR. Since we will be removing the LLVM backend, we could just remove it?

vext01:
I don't object to this, but how are we going to make sure it always "cargo check"s on OpenBSD? Are you OK with manually checking this periodically?

(I keep meaning to try out [distant](https://distant.dev/) which I believe would allow you to run a nvim locally, but do all the building/analysing on a remote vim)

ltratt:
> I don't object to this, but how are we going to make sure it always "cargo check"s on OpenBSD? Are you OK with manually checking this periodically?

I am very fine with doing so! This is very much intended to be just a "best effort", rather than a "supported", thing right now.

> That binary prints the embedded LLVM IR. Since we will be removing the LLVM backend, we could just remove it?

Should I also remove `bin/ir_dump`?

vext01:
> Should I also remove bin/ir_dump

If it just wraps the binary, then yeah, go ahead.

ltratt:
Squashed.

PR #1269, opened 2024-06-27
Tell the user where in the input there was an error.

ltratt:
This is rather more informative than panicing.


vext01:
Can't argue with this.

PR #1273, opened 2024-06-28
Display and parse a guard's live variables.

ltratt:
Before we printed and parsed only things like:

```
guard true, %20
```

without knowing what live variables the guard has in mind. This PR both displays a guards live variables and allows them to be parsed in e.g.:

```
guard true, %20, [%0, %2, %7]
```

This has already highlighted that some guards include their condition SSA var in the live var list, and some don't. I don't know why: I'll raise a separate issue for that.


ykrt/src/compile/jitc_yk/jit_ir/jit_ir.y
```diff
@@ -62,22 +62,22 @@ Inst -> Result<ASTInst, Box<dyn Error>>:
     "*" Operand "=" Operand { Ok(ASTInst::Store{tgt: $2?, val: $4?, volatile: false}) }
   | "*" Operand "=" Operand "," "VOLATILE" { Ok(ASTInst::Store{tgt: $2?, val: $4?, volatile: true}) }
   | "BLACK_BOX" Operand { Ok(ASTInst::BlackBox($2?)) }
-  | "GUARD" "TRUE" "," Operand {
-      Ok(ASTInst::Guard{operand: $4?, is_true: true})
+  | "GUARD" "TRUE" "," Operand "," "[" LocalsList "]" {
```
ptersilie, line 65: We possibly want this to be `OperandsList` later so we can track constants too. I believe constants can end up here via constant propagation.

ptersilie:
> This has already highlighted that some guards include their condition SSA var in the live var list, and some don't. I don't know why

To me this just means that the guard condition just happens to be a live variable or not. For example
```
// Condition dies immediately after if and does not need to be deoptimised.
if x == 1 { 
   ...
}

// Condition is live after `if` and needs to be deoptimised.
int c = x == 1;
if c {
...
}
printf("%d", c);
```

ltratt:
That second idiom seems a very unlikely one in real code to me... Perhaps a more realistic possibility is that the compiler is identifying duplicate conditions? Dunno.

ptersilie:
Really? I think it can happen quite often when you pass arguments that you run checks on and foward them to other functions or use them for control flow.
```
int foo(bool b) {
   if (b) {
      ...
   }
   ...
   if(b) {
      ...
   }
   bar(b);
}
```

ltratt:
> I think it can happen quite often when you pass arguments that you run checks on and foward them to other functions or use them for control flow.

Aha, yes, using the `bool` type like that makes sense! I'm an idiot.

PR #1276, opened 2024-07-01
Implement (some) floating point comparisons.

vext01:
Requires: https://github.com/ykjit/ykllvm/pull/178

Float comparisons are a bit more involved than integer comparisons.

Here's some of the highlights (and my working understanding of them):

 - There are some special values that make the ordering a bit more interesting, e.g. +inf, -inf, NaN, +0 and -0.

 - There is a large encoding space for representing NaNs and NaNs are, in float terminology, "unordered". i.e. in the partial order of floats a NaN cannot be compared to any other element. Amusingly, two NaNs are never considered equal.

 - This plays into comparison operations. You can get the unordered result if at least one of the things you want to compare are NaN.

 - NaN comes in two variants: quiet and signalling, which basically mean "silently proceed, despite NaN" or "crash". In our x86 codegen we use `ucomis{s,d}` and `comis{s,d}` respectively for this. The latter would cause the chip to raise a floating point exception.

 - But what happens when we get the unordered result for a comparison and we are not crashing? We still need an answer, were they equal or not? We do as clang does, and follow the IEE-754 float spec:
   - Any "not equal" comparison involving NaN is true.
   - All other comparisons are false.

I hope I've got this all right :)


vext01:
(After this `attrib.lua` gets a little further)

vext01:
I'll force push to fix the conflicts, if that's ok?

tests/c/fcmp_double.newcg.c
```diff
@@ -0,0 +1,85 @@
...
+  double a = 1.1, b = 2.2, nana = NAN;
+  double aa = a, nanb = nana;
+  double zero = 0.0, negzero = -0.0;
+  NOOPT_VAL(loc);
+  NOOPT_VAL(i);
+  NOOPT_VAL(a);
+  NOOPT_VAL(aa);
+  NOOPT_VAL(b);
+  NOOPT_VAL(nana);
+  NOOPT_VAL(nanb);
```
ltratt, line 64: Do we need to `NOOPT_VAL` 'zero' and 'negzero'?

tests/c/fcmp_float.newcg.c
```diff
@@ -0,0 +1,85 @@
...
+
+int main(int argc, char **argv) {
+  YkMT *mt = yk_mt_new(NULL);
+  yk_mt_hot_threshold_set(mt, 0);
+  YkLocation loc = yk_location_new();
+
+  int i = 3;
+  float a = 1.1, b = 2.2, nana = NAN;
+  float aa = a, nanb = nana;
+  float zero = 0.0, negzero = -0.0;
```
ltratt, line 57: `NOOPT_VAL` ?

tests/c/fcmp_double.newcg.c
vext01: Yes, we do. Well spotted.

vext01:
https://github.com/ykjit/yk/pull/1276/commits/a9baf295184ef01c188fd1d0f73b9ed7b6d73de8 fixes the missing NOOPT_VALs.

vext01:
OK to force push a squash/rebase?

ltratt:
Please go ahead.

vext01:
ready

vext01:
force pushed a fix for formatting. go ahead.

PR #1277, opened 2024-07-01
Largely complete dead code elimination

ltratt:
This PR largely completes dead code elimination in the final commit (https://github.com/ykjit/yk/commit/49374b42bef49df52786d4f2c4303baaf5bc256a). On the way to doing that, I realised that I had to more carefully think about how we handle `Proxy*` instructions (https://github.com/ykjit/yk/commit/b47ddc07e4c2211076d3cd09e5efdde3672b4f0c). With this, I can successfully run yklua and eliminate dead code in its traces.

### DIFF
--- a/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
@@ -247,7 +247,7 @@ impl Module {
 
     /// Iterate, in order, over all `InstIdx`s of this module (including `Proxy*` and `Tombstone`
     /// instructions).
-    pub(crate) fn iter_inst_idxs(&self) -> impl DoubleEndedIterator<Item = InstIdx> {
+    pub(crate) fn iter_all_inst_idxs(&self) -> impl DoubleEndedIterator<Item = InstIdx> {
         (0..self.insts.len()).map(|x| InstIdx::new(x).unwrap())
     }
 
--- a/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
@@ -251,11 +251,13 @@ impl Module {
         (0..self.insts.len()).map(|x| InstIdx::new(x).unwrap())
     }
 
-    /// An iterator over instruction indices. This skips `Proxy*` and `Tombstone` instructions: in
-    /// other words, it produces monotonically increasing, but potentially non-consecutive, instruction
-    /// indices.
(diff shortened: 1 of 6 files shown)


vext01:
Superficially looks OK to me.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -1,9 +1,45 @@
...
+//!      memory used down (and also bypasses issues with representing graph structures in Rust, but
+//!      that's slightly accidental).
+//!
+//!   2. The IR is designed to allow most manipulations to be performed without requiring
+//!      reallocation.
+//!
+//! This second property has some surprising consequences:
+//!
+//!   1. We have to introduce `Proxy*` instructions which semi-transparently "forward" an
+//!      instruction's results on. For example if we IR in memory stored along these lines:
```
ptersilie, line 16: "results on"? Also "if we HAVE IR in memory".

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -1,9 +1,45 @@
...
+//!        ```
+//!
+//!      i.e. we want to pretend that the `%1` doesn't exist and that the reference to `%1` is
+//!      actually just a constant.
+//!
+//!   2. We never store [Operand]s directly: we only ever store [PackedOperand]s. Whenever we want
+//!      to get an [Operand] we *must* use [PackedOperand::unpack]. `unpack` does the necessary
+//!      deproxying.
+//!
+//!   3. In nearly all cases we should view the IR as deproxied: unless you explicitly intend to
```
ptersilie, line 40: If we really never want this, why don't we make `Module::inst` the no proxy version and `Module::inst_with_proxies` the other?

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ltratt: Fixed in 85ef34ee.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ltratt: I'm pretty sure we can come up with better names but I'd rather get a bit more experience with the API before worrying *too* much about it.

ltratt:
OK to squash?

ptersilie:
Please squash.

ltratt:
Squashed.

ltratt:
Force pushed a rebase.

PR #1281, opened 2024-07-02
Move more mapping into a thread

ltratt:
The first commit simply shuffles code around in a file. The main commit here (with description) is https://github.com/ykjit/yk/commit/7e289f679911090399edda16ddd8fae658daae46.

### DIFF
--- a/ykrt/src/trace/hwt/mapper.rs
+++ b/ykrt/src/trace/hwt/mapper.rs
@@ -29,33 +29,11 @@ impl AOTTraceIterator for HWTTraceIterator {}
 
 impl HWTTraceIterator {
     pub fn new(trace: Box<dyn Trace>) -> Result<Self, TraceRecorderError> {
-        let mut hwti = HWTTraceIterator {
+        Ok(Self {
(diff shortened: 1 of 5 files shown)


ykrt/src/trace/hwt/mapper.rs
```diff
@@ -190,6 +168,31 @@ impl HWTTraceIterator {
 impl Iterator for HWTTraceIterator {
     type Item = Result<TraceAction, AOTTraceIteratorError>;
     fn next(&mut self) -> Option<Self::Item> {
+        if self.tas_generated == 0 {
```
vext01, line 171: Unrelated to this PR, the name `tas_generated` gave me hassle (what's a "tas"? when is it generated?).

Turns out it's a counter! Should it be called `num_actions`?

Anyway, this is probably bikeshedding, and unactionable in this PR.

ltratt: "Trace ActionS generated". Agreed another name could help, but not sure it's a big deal.

PR #1284, opened 2024-07-03
Rework codegen trait

ltratt:
The main part of the PR is https://github.com/ykjit/yk/commit/4988bf7705fe23f34b8ee65336baef15a3145b63.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -10,11 +10,11 @@
 
 use super::{
     super::{
-        jit_ir::{self, BinOp, InstIdx, Operand, Ty},
+        jit_ir::{self, BinOp, InstIdx, Module, Operand, Ty},
         CompilationError,
     },
     abs_stack::AbstractStack,
-    reg_alloc::{LocalAlloc, RegisterAllocator, StackDirection},
+    reg_alloc::{spill_alloc::SpillAllocator, LocalAlloc, RegisterAllocator, StackDirection},
     CodeGen,
 };
 #[cfg(any(debug_assertions, test))]
@@ -31,9 +31,9 @@ use dynasmrt::{
 };
 #[cfg(any(debug_assertions, test))]
 use indexmap::IndexMap;
-use std::sync::Arc;
 #[cfg(any(debug_assertions, test))]
-use std::{cell::Cell, error::Error, slice};
+use std::{cell::Cell, slice};
+use std::{error::Error, sync::Arc};
 use ykaddr::addr::symbol_to_ptr;
 
 mod deopt;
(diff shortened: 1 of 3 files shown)


vext01:
Out of interest, what's the expensive part of bringing up a codegen?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -88,8 +88,24 @@ const STACK_DIRECTION: StackDirection = StackDirection::GrowsDown;
 #[inline(never)]
 pub extern "C" fn __yk_break() {}
 
+/// A simple front end for the X86_64 code generator.
+pub(crate) struct X64CodeGenFrontEnd;
```
vext01, line 92: Maybe a bikeshed too far, but this name confused me a bit, because a code generator is itself a backend. I know what you mean though, this is the frontend of the backend, right?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -88,8 +88,24 @@ const STACK_DIRECTION: StackDirection = StackDirection::GrowsDown;
...
+
+impl CodeGen for X64CodeGenFrontEnd {
+    fn codegen(&self, m: Module) -> Result<Arc<dyn CompiledTrace>, CompilationError> {
+        let ra = Box::new(SpillAllocator::new(StackDirection::GrowsDown));
+        X64CodeGen::new(&m, ra)?.codegen()
```
vext01, line 97: Actually I'm not sure what this change does. I thought you were going to cache the code generator so that you don't have to reinstantiate it every time, but it looks like we are doing that anyway here?

Is this change just setting us up to do that in a future PR, or?

ltratt: Yes, this is mostly setup: the point is that the `CodeGen` trait is now "correct" but this PR only minimally changes the x86_64 backend.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ltratt: This is a minimal hack to nudge us in the right direction: the names aren't good, but the structure isn't quite right either. We can gradually improve things in future PRs.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
vext01: So just for clarification: this PR just makes a place for shared caches to live in the `Frontend` struct?

ltratt: I tend to to think of it as "the `CodeGen` trait no longer prevents caching", but ultimately it's the same thing, yes.

vext01:
```
15:13:46 error: methods called `new` usually return `Self`
15:13:46    --> ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs:102:5
15:13:46     |
15:13:46 102 | /     pub(crate) fn new() -> Result<Arc<dyn CodeGen>, Box<dyn Error>> {
15:13:46 103 | |         Ok(Arc::new(Self))
15:13:46 104 | |     }
15:13:46     | |_____^
15:13:46     |
15:13:46     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#new_ret_no_self
15:13:46     = note: `-D clippy::new-ret-no-self` implied by `-D warnings`
15:13:46     = help: to override `-D warnings` add `#[allow(clippy::new_ret_no_self)]`
```

That's a lint we could probably turn off globally in a config file?

ltratt:
This is a place where you can see some unfortunate parts of type inference not working: you can't return the object directly but `?` tricks the inferencer into casting to `dyn T`. Fixed in 43b01f3.

vext01:
please squash

ltratt:
Squashed.

PR #1285, opened 2024-07-04
Minimally deal with const ints of different sizes in the x86 backend.

ltratt:
This isn't really a complete solution, because at the moment we store every constant integer in 64 bits worth of space, even if it's smaller. If/when we use less space, we might get some fun bugs as we find it breaks assumptions elsewhere, but it's difficult for us to deal with that until we encounter it.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
@@ -31,7 +31,12 @@ pub(crate) enum LocalAlloc {
     ///
     /// FIXME: unimplemented.
     Register,
-    ConstInt(u64),
+    /// A constant integer `bits` wide (see [jit_ir::Const::ConstInt] for the constraints on the
+    /// bit width) and with value `v`.
+    ConstInt {
+        bits: u32,
+        v: u64,
+    },
     ConstFloat(f64),
 }
 
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -1003,7 +1003,12 @@ impl<'a> X64CodeGen<'a> {
                     // it doesn't have an allocation. We can just push the actual value instead
                     // which will be written as is during deoptimisation.
                     match self.m.const_(*c) {
-                        Const::Int(_, c) => locs.push(LocalAlloc::ConstInt(*c)),
+                        Const::Int(tyidx, c) => {
+                            let Ty::Integer(bits) = self.m.type_(*tyidx) else {
+                                panic!()
+                            };
+                            locs.push(LocalAlloc::ConstInt { bits: *bits, v: *c })
+                        }
                         _ => todo!(),
                     };
                 }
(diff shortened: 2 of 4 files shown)


ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -1095,10 +1100,10 @@ impl<'a> X64CodeGen<'a> {
                 }
             }
             LocalAlloc::Register => todo!(),
-            LocalAlloc::ConstInt(c) => {
-                // FIXME: Adjust for different constant sizes. Requires storing the size in the
-                // variant.
-                dynasm!(self.asm; mov Rq(reg.code()), QWORD *c as i64);
+            LocalAlloc::ConstInt { bits, v } => {
+                // FIXME: Adjust for different constant sizes?
```
vext01, line 1104: I know you didn't introduce this, but I think this FIXME can be removed. If the type is smaller than 64-bits then the top bits are just undefined and two's compliment ensures any truncation maintains the same numeric value.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -1095,10 +1100,10 @@ impl<'a> X64CodeGen<'a> {
...
+                assert_eq!(*bits, 64);
```
vext01, line 1105: I think this assertion can therefore be removed too?

ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
```diff
@@ -31,7 +31,12 @@ pub(crate) enum LocalAlloc {
     ///
     /// FIXME: unimplemented.
     Register,
-    ConstInt(u64),
+    /// A constant integer `bits` wide (see [jit_ir::Const::ConstInt] for the constraints on the
+    /// bit width) and with value `v`.
+    ConstInt {
+        bits: u32,
```
vext01, line 37: I wonder if we should store a type index instead, then we can just compare the type indices in the well-formed check.

vext01:
Is there any functional change?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ltratt: At some point we'll want to handle these differently, so if we remove the assert, things get swallowed. It may be that this proves too annoying, though, and we have to get rid of it, but it passes yk's test suite so...

ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
ltratt: The integer type only stores the bit size, so storing the `TyIdx` would just force us to look things up unnnecessarily. I *think* it's fine (and probably best) to store just the number of bits in the codegen.

ltratt:
No functional change.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
vext01: Doesn't it crash programs that use constant integers that are not 64-bit? Or are all constant integers promoted to 64-bit?

ltratt: I expected it to crash in the test suite. The fact that it doesn't means that either the test suite isn't good enough or we promote things somewhere. I guess we'll find out soon enough.

vext01: OK. Will merge once I'm done with CI fiddling.

PR #1289, opened 2024-07-05
Terminology tweaks

ltratt:
This is a slightly half-hearted PR where I wanted to do something big, got bogged down in details, and then decided to salvage something minor.

The first commit somewhat addresses the comment from #1284 that the X86 structs weren't well named.

The second commit starts to make the register allocation terminology a bit clearer. Doing so has proven unintentionally useful: it makes clear that the `RegisterAllocator` trait fundamentally assumes that SSA variables have a single location, which means one can't express "store in a reg, spill to stack, reload from a reg" and so on. Doing something about that is going to be quite some work, hence this not-even-halfway-house PR.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/mod.rs
@@ -23,7 +23,7 @@ pub(crate) trait CodeGen: Send + Sync {
 
 pub(crate) fn default_codegen() -> Result<Arc<dyn CodeGen>, Box<dyn Error>> {
     #[cfg(target_arch = "x86_64")]
-    return Ok(x86_64::X64CodeGenFrontEnd::new()?);
+    return Ok(x86_64::X86_64CodeGen::new()?);
 
     #[cfg(not(target_arch = "x86_64"))]
     return Err("No code generator available for this platform".into());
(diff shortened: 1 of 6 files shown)


vext01:
I'm luke warm on the second commit.

I do agree that "register allocation" is a bad name, given that you typically allocate stack too as part of that process. "storage allocation" might have been a better name, but hey ho.

I don't really have a problem with `LocalAllocation`. You can allocate a register or a stack slot for a local. So it seems accurate.

That said, `VarLocation` is fine too.

Unless any of that changes your mind, I'm happy to merge.

ltratt:
One of the challenges is that in a "real" register allocator, the location of a variable changes over time (as you spill/unspill it). It's another reason I don't like to think of its "allocation" which sounds like a static thing. I'm unsure the names I'm suggesting are the best, but I think they are a mild improvement on what we have. At least, they helped make it clear to me where we have limitations in our current API.

vext01:
I'll merge this.

> One of the challenges is that in a "real" register allocator, the location of a variable changes over time (as you spill/unspill it). It's another reason I don't like to think of its "allocation" which sounds like a static thing.

I dunno about that. When I am allocated a seat on a plane, I don't expect to sit there forever :)

ltratt:
Looks like a transient PT error? Maybe retry?

PR #1294, opened 2024-07-10
Sort all numeric match statements to ascending order.

ltratt:
We previously used a mix of ascending and descending order, the latter of which I will cheerfully admit has driven me potty when doing experiments in this code in recent days. This commit homogenises on the (I believe more conventional) ascending order.


vext01:
This likely happened because 8 is most common and was implemented first :)

PR #1295, opened 2024-07-10
Less shadowstack

ptersilie:
Requires https://github.com/ykjit/ykllvm/pull/181.

We no longer adjust the shadowstack for external functions as there's no need to
do so. While this doesn't have a measurable performance gain, it does clean up
the IR quite a bit. So this

```
%1 = ptr_add shadowstack, 16
*shadowstack = %1
call external_func()
*shadowstack = %orig_shadowstack
```

turns into

```
call external_func()
```

The change broke a few tests which this commit fixes.


ptersilie:
I've added another unrelated commit that adds two new tests. These should have gone into the side-tracing PR but I forgot to add them back then.

ptersilie:
Ok to squash and rebase?

ltratt:
Please squash.

ptersilie:
Squashed and rebased.

PR #1296, opened 2024-07-11
Execute the trace inside the AOT frame.

ptersilie:
This change resets the stack back to the AOT frame before executing a trace. Effectively, this pretends the control point never happened, placing the AOT frame and trace frame right next to each other. This will allow us to easily access the AOT frame from within the trace to read out live variables, which in turn enables us to get rid of the trace inputs, in a future commit.

Depends on: https://github.com/ykjit/ykllvm/pull/182

### DIFF
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -58,13 +58,15 @@ pub extern "C" fn __ykrt_control_point(
     ctrlp_vars: *mut c_void,
     // Frame address of caller.
     frameaddr: *mut c_void,
+    // Stackmap id for the control point.
+    smid: u64,
 ) {
     debug_assert!(!ctrlp_vars.is_null());
     if !loc.is_null() {
         let mt = unsafe { &*mt };
         let loc = unsafe { &*loc };
         let arc = unsafe { Arc::from_raw(mt) };
-        arc.control_point(loc, ctrlp_vars, frameaddr);
+        arc.control_point(loc, ctrlp_vars, frameaddr, smid);
         forget(arc);
     }
 }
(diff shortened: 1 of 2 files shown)


ykcapi/src/lib.rs
```diff
@@ -58,13 +58,14 @@ pub extern "C" fn __ykrt_control_point(
     ctrlp_vars: *mut c_void,
     // Frame address of caller.
     frameaddr: *mut c_void,
+    smid: u64,
```
ltratt, line 61: Let's document this, because I don't think I'll remember it.

I think it's also worth documenting (at least somewhere in yk) that ykllvm explicitly writes out a `u64`. I wrote a long comment suggesting this should be `size_t` before realising that ykllvm really does write out a `u64` :)

ykrt/src/mt.rs
```diff
@@ -64,6 +64,26 @@ pub(crate) trait SideTraceInfo {
     fn as_any(self: Arc<Self>) -> Arc<dyn std::any::Any + Send + Sync + 'static>;
 }
 
+#[cfg(target_arch = "x86_64")]
+#[naked]
+#[no_mangle]
+unsafe extern "C" fn __exec_trace(
```
ltratt, line 70: Should this be `__yk_exec_trace`?

ykcapi/src/lib.rs
ptersilie: Yeah, stackmap ids are `u64`s. :-/

ykrt/src/mt.rs
ptersilie: Will fix.

ptersilie:
Both comments addressed in [c19f713](https://github.com/ykjit/yk/pull/1296/commits/c19f713b501dbc1e711b7a7b0d7b03a06ca0cb8c). Ok to squash?

ltratt:
Please squash.

ptersilie:
Squashed.

PR #1297, opened 2024-07-11
Fix submodule.

ptersilie:


ltratt:
I'll merge this but let's also raise another PR soon that checks this doesn't happen again. Thanks!

PR #1299, opened 2024-07-11
Check the submodule commit is from `main` (best we can).

vext01:
We check that the HEAD commit is in the last 100 commits of main (which is all the history we have due to CI's shallow commit).

(only tested a bit :) )


ltratt:
It's worth giving it a go -- thanks!

PR #1301, opened 2024-07-12
Test udiv in the udiv test.

ltratt:
At some point it seems that the udiv test was overwritten by the sdiv test. This commit makes the udiv test test udiv, not sdiv -- we have the sdiv test to test sdi.


ptersilie:
Huh that's odd. Seems I never added udiv tests in #1243. Even though I could have sworn I changed the test after copying it from the sdiv test. Maybe something went wrong during rebasing or maybe yk has finally driven me insane.

Anyway this looks good.

ltratt:
> maybe yk has finally driven me insane.

:p

PR #1307, opened 2024-07-19
Deserialise and display line-level debug info in the AOT IR.

vext01:
Requires https://github.com/ykjit/ykllvm/pull/183, so will need a ykllvm sync.

This is the first step towards having C source code in gdb when you are debugging a trace.

Unfortunately this requires some changes to `display()` for a few data structures. Hopefully these changes aren't too annoying for those using `display()` for debuging. At least the additional parameters are all `Option`, so you can pass `None` if you don't have the required information to hand.

I can already see the lines floating around a bit, but such is debug info. It has to be better than nothing.

I had to change the lang_tester comment character so as to match debug comments in the test. If you want a comment in the test, you now use `##`.


tests/c/truncate.newcg.c
```diff
@@ -1,6 +1,6 @@
-// # Truncation currently always truncates to i32, no matter the target type.
-// #ignore-if: true
-// # ignore-if: test $YK_JIT_COMPILER != "yk" -o "$YKB_TRACER" = "swt"
+// ## Truncation currently always truncates to i32, no matter the target type.
```
ltratt, line 1: Are the `##`s intentional?

tests/langtest_c.rs
```diff
@@ -52,7 +52,7 @@ fn run_suite(opt: &'static str) {
     };
 
     LangTester::new()
-        .comment_prefix("#")
+        .comment_prefix("##")
```
ltratt, line 55: Won't this break other tests than `truncate.newcg.c`? I really think this should be a PR in its own right.

ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -50,6 +53,29 @@ const LLVM_DEBUG_CALL_NAME: &str = "llvm.dbg.value";
...
 
+/// A struct identifying a line in a source file.
+#[derive(Debug)]
+#[deku_derive(DekuRead)]
+struct LineInfoLoc {
+    /// The path to the containing source file.
+    ///
+    /// The file existed at AOT compile time and there is no guarantee that it still exists now.
+    #[deku(until = "|v: &u8| *v == 0", map = "map_to_pathbuf")]
+    filename: PathBuf,
```
ltratt, line 64: I think this means that every `LineInfoLoc` has a pointer to a full C string on the heap? Isn't that going to be quite punishing? I imagine we can probably observe the slowdown that causes on deserialisation?

tests/c/truncate.newcg.c
vext01: Yep. See the last paragraph of the PR description.

tests/langtest_c.rs
vext01: Your test was the only one using a comment :)

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Yep. I've not optimised anything at this point. Really we need a table with indices.

tests/langtest_c.rs
ltratt: OK but let's at least break it out into a separate commit later? Someone (OK, maybe me!) might want to see this distinctly in the history.

ykrt/src/compile/jitc_yk/aot_ir.rs
ltratt: Right, but how much does this slow things down by? I'm *generally* fine with not optimising things, but I worry this could massively slow down deserialisation, but maybe I'm worrying over nothing.

tests/langtest_c.rs
vext01: Unfortunately I can't add the new debug info test without this change, so it's all dependent. It makes it possible to match the debuginfo lines that start with `#`.

ltratt: Yes, so when this comes to squashing, please make this a separate (first) commit.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Thinking about it, for a large program like yklua, this is probably going to be too expensive, so maybe I need to rethink this part.

ltratt: Please measure before making a decision.

vext01: do you mean measure master (no debuginfo) against this branch (some debuginfo)? that's about all I can do without actually implementing the other way first.

ltratt: Yes, master vs this branch: we just need to know if this is a performance cost that we can measure.

ltratt:
@vext01 I had to do a bit of hunting to remind myself of where I put this, but here's how Converge encodes the same sort of information as this (and the ykllvm) PR https://github.com/ltratt/converge/blob/master/compiler/Compiler/Targets/Sixty_Four_Bit.cv#L810:

```
//   bits 0-3  : num of instructions covered
//   bit  4    : if set, the next src info covers this instruction
//   bits 5-31 : src byte offset
//   bits 32-44: import num
//   bits 44-64: len
```

In essence bits 0-3 are the run-length encoding, but bit 4 allows you to deal with the "more than 3 bits of consecutive info" case. Bits 32-44 are, in essence, an index into file names, so you only have to heap allocate one string per src file. It's not super clever, but it was good enough: from memory I think it compressed things by 50% or more.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: Surprisingly the difference isn't measurable for yklua at millisecond granularity. About 34ms before and after.

I measured just the call to the deserialiser like this:
```diff
diff --git a/ykrt/src/compile/jitc_yk/aot_ir.rs b/ykrt/src/compile/jitc_yk/aot_ir.rs
index 06b56f89..843aba02 100644
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -257,7 +257,10 @@ impl std::fmt::Display for Module {

 /// Deserialise an AOT module from the slice `data`.
 pub(crate) fn deserialise_module(data: &[u8]) -> Result<Module, Box<dyn Error>> {
-    let ((_, _), modu) = Module::from_bytes((data, 0))?;
+    use std::hint::black_box;
+    let start = std::time::Instant::now();
+    let ((_, _), modu) = black_box(Module::from_bytes((data, 0)))?;
+    eprintln!("deser took: {}ms", start.elapsed().as_millis());
     Ok(modu)
 }
```

Do you see anything amiss, or do I believe my results?

ltratt: I don't know how this works, so I'm relying on you to tell me! But if in doubt, let's just do indexes.

vext01: It's not hard to implement, so me may as well.

ltratt: Works for me.

vext01:
https://github.com/ykjit/yk/pull/1307/commits/3d0024920e936aeeef8c1fbaec6811c737663681 deserialises path indices.

I had to jump through some hoops to get that through deku.

ltratt:
Hmm, I've only just realised what you meant with "changes to `display`". IMHO these are in the wrong place, and will make debugging (where we sometimes want to print these) a nightmare: an instruction shouldn't print its location, but `Module` can do so. Can we just shovel that stuff into `Module::display`?

ltratt:
Basically we can do that here:

```rust
impl fmt::Display for Module {
        ...
        for (iidx, inst) in self.iter_skipping_insts() {
            // XXX: Print out debug line information
            write!(f, "\n    {}", inst.display(iidx, self))?
        }
```

vext01:
I've spent too long on this. It's actually quite tricky to implement in a palatable manner.

If we want to lift the instruction formatting logic up into the `Module` level as above, then we'd have to **duplicate** (because we still want to be able to print things in isolation) the formatters of `Func`, `BBlock` up into `Module` too. Those routines are non-trivial so I'm not happy duplicating them.

https://github.com/ykjit/yk/pull/1307/commits/26e9e836e72a4b385e9d9a7032ae0bc57daba5d7 shows an alternative attempt (tests currently fail) where we lift the formatters into dedicated functions and call them differently in `Display for Module` and elsewhere. I don't like this either. It makes the code quite convoluted and hard to understand.


I wonder if all of this added complexity is worth the trouble just so that we can do:
```
inst.display(self.m)
```

instead of:
```
inst.display(self.m, None)
```

Is adding `, None` really so arduous?

ltratt:
I think you've done something much more complex than the dumb thing I was suggesting. What I was suggesting is *only* altering `Module::fmt` so that it prints out the comment lines: you don't want to add any information to `DisplayableInst` or whatnot. That should be very easy to do unless I'm missing out on something obvious. Something like should do the trick, right?

```rust
impl fmt::Display for Module {
        ...
        for (iidx, inst) in self.iter_skipping_insts() {
            if <there's a comment for inst> {
                write!(f, "; blah in blah.c")?;
            }
            write!(f, "\n    {}", inst.display(iidx, self))?
        }

        Ok(())
    }
}
```

vext01:
The issue is, the module doesn't have direct access to instructions. There are function and block intermediaries, so if you want to do what you suggest, we would have to duplicate their formatters into `Module` too.

I saw you proposed using `iter_skipping_insts()`. Just to be clear, this is the AOT IR.

ykrt/src/compile/jitc_yk/aot_ir.rs
vext01: https://github.com/ykjit/yk/pull/1307/commits/3d0024920e936aeeef8c1fbaec6811c737663681 implemented indices.

vext01:
We spoke offline and agreed that for now the best course of action is to roll with the initial design and we can adapt this later if it's too annoying.

In light of that I've reverted the last commit and I think we can now squash this?

vext01:
Squashed.

Lang tester change separate, as requested.

ykllvm update included.

PR #1309, opened 2024-07-19
Rename `GuardId` to `GuardIdx` and use it consistently.

ltratt:
Needs #1308 to be merged first.

We were previously inconsistent in (1) how we handled guard IDs (2) whether we used the `GuardId` type at all (3) whether `GuardId`s could be indexes into vectors or not. This PR sorts each thing out (hopefully).


ptersilie:
Please rebase once #1308 has merged.

ltratt:
Rebased.

PR #1310, opened 2024-07-22
More tidy ups in and around side-tracing

ltratt:
Hopefully each commit makes sense in and of itself.


ptersilie:
Some more merge conflicts for me, but otherwise looks fine.

PR #1311, opened 2024-07-23
Use fm's new matching features for x86 register matching.

ltratt:
Instead of using concrete register names, one can refer to a class of registers e.g. `r.8` is all 8-bit registers. To match a register name but ignore its value uses `r.8._`: to match a register name use `r.8.x`. Suffixes: must be unique (i.e. `r.8.x` and `r.8.y` must refer to different registers); and they refer to the "underlying" register (e.g. `r.8.x` and `r.64.x` might match `al` and `RAX` which is considered the same register, but will fail to match against `al` and `RBX`).

To make this concrete these `assert`s all succeed:

```rust
let fmm = fmatcher("r.8.x r.8.y r.64.x");
assert!(fmm.matches("al bl rax").is_ok());
assert!(fmm.matches("al al rax").is_err());
assert!(fmm.matches("al bl rbx").is_err());
assert!(fmm.matches("al bl eax").is_err());
```

This commit is unfinished in the sense that not all the tests are ported: this is an RFC in its current state. Needs https://github.com/softdevteam/fm/pull/50.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -1521,6 +1521,16 @@ impl<'a> AsmPrinter<'a> {
     }
 }
 
+/// x64 tests. These use an unusual form of pattern matching. Instead of using concrete register
+/// names, one can refer to a class of registers e.g. `r.8` is all 8-bit registers. To match a
+/// register name but ignore its value uses `r.8._`: to match a register name use `r.8.x`.
+///
+/// Suffixes: must be unique (i.e. `r.8.x` and `r.8.y` must refer to different registers); and they
+/// refer to the "underlying" register (e.g. `r.8.x` and `r.64.x` might match `al` and `RAX` which
+/// is considered the same register, but will fail to match against `al` and `RBX`).
+///
+/// Note that general purpose (`r.`) and floating point (`fp.`) registers occupy different suffix
+/// classes (i.e. `r.8.x` and `fp.128.x` do not match the same register "x").
 #[cfg(test)]
 mod tests {
     use super::X64CompiledTrace;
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -1521,42 +1521,151 @@ impl<'a> AsmPrinter<'a> {
...
-    /// Test helper to use `fm` to match a disassembled trace.
-    fn match_asm(cgo: Arc<X64CompiledTrace>, pattern: &str) {
-        let dis = cgo.disassemble().unwrap();
-
+    fn fmatcher<'a>(ptn: &'a str) -> FMatcher<'a> {
         // Use `{{name}}` to match non-literal strings in tests.
         let ptn_re = Regex::new(r"\{\{.+?\}\}").unwrap();
         let ptn_re_ignore = Regex::new(r"\{\{_}\}").unwrap();
         let text_re = Regex::new(r"[a-zA-Z0-9\._]+").unwrap();
-        // The dissamebler alternates between upper- and lowercase hex, making matching addresses
```
vext01, line 1548: Unrelated comment: the disassembler we use in tests is different from the one we use elsewhere (e.g. in PT). We used a different one because there was a sanitiser issue, which I believe is now fixed. We could probably use `iced` across the board if there is a hunger to minimise deps and maybe fix this casing issue.

ltratt: I have no idea. I definitely don't want to think about that in this PR :)

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -1521,42 +1521,151 @@ impl<'a> AsmPrinter<'a> {
...
-        // difficult. So just lowercase both pattern and text to avoid tests randomly breaking when
-        // addresses change.
-        let lowerpattern = pattern.to_lowercase();
-        let fmm = FMBuilder::new(&lowerpattern)
+        let mut fmb = FMBuilder::new(ptn)
             .unwrap()
+            .name_matching_processor(|names| {
```
vext01, line 1616: I was expecting this processor to have to look for the wildcard `_`  (in e.g. `r.8._`) at some point. How comes it's not needed?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
vext01: Of course. Hence flagging it unrelated.

vext01:
The user-observable functionality looks great, but I don't feel like I've understood `name_matching_processor()` sufficiently to do more than a superficial review (I've often struggled with understanding name matchers in general, but I usually manage to muddle through).

If you are OK with that, then we can proceed.

ptersilie:
> e.g. r.8.x and r.64.x might match al and RAX which is considered the same register, but will fail to match against al and RBX

Does this mean that `r.8.x` and `r.64.x` will only match if the register name is the same, even if the sizes are different. So the `x` part will bind to the `A` part of `AL, RAX`, etc?

vext01:
My understanding is that `r.8.x` would never match a `r.64.x` would never match. So `al` and `rax` are treated as distinct. I think that's the way we'd want it anyway.

ptersilie:
> My understanding is that `r.8.x` would never match a `r.64.x`. So `al` and `rax` are treated as distinct. I think that's the way we'd want it anyway.

Yes, they are distinct, but the `x` means that all other sizes need to be also from the RAX register. So `r.16.x` needs to be `ax`, and `r.32.x` needs to be `eax`. They can't suddenly be `ebx`.

ltratt:
>   Does this mean that r.8.x and r.64.x will only match if the register name is the same, even if the sizes are different. So the x part will bind to the A part of AL, RAX, etc?

Yes, though it's not done by text but by the underlying register: so if r.8.x is `al` then r.64.x can only match `rax` (and not rbx etc).

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ltratt: That's handled by an "ignore_name_matcher".

ltratt:
Ported all remaining tests (including floating point register support) in https://github.com/ykjit/yk/pull/1311/commits/8d4a343c91042c4b439fe264c7dbd3111fbc06bb. Hopefully ready for squashing!

vext01:
If @ptersilie is OK with this, I think I am.

ltratt:
@ptersilie OK to squash?

ptersilie:
Please squash.

ltratt:
Squashed.

PR #1312, opened 2024-07-23
Allow type_() to work on global operands.

vext01:
The type is always `Ty::Ptr` for globals.

This fixes a crash where when you try to `YKD_LOG_IR=-:aot` and you have instructions that defer the type decision to operands which may be globals, e.g. `phi` and `switch`. Test included.

(The test would have been better as a unit test, but manually constructing an AOT module is painful, so I used an integration test)


vext01:
Related (but not required): https://github.com/ykjit/ykllvm/pull/184

PR #1315, opened 2024-07-25
There's no need to nest these tests in a module.

ltratt:
We only have one register allocator now (and for the foreseeable future), and it's not called `spillalloc` anymore either, so remove the nested module.


vext01:
Agreed

PR #1316, opened 2024-07-25
Slightly rework errors

ltratt:
I've been confused in some situations where things have gone wrong and I haven't been notified. Long story short: we have at least some of the support we need to do a proper job, but I'm not 100% sure what "proper job" should be, other than "not exactly what we're doing right now". So this PR is probably a halfway house: if we're lucky it will help make us soon realise what we should do.

### DIFF
--- a/ykrt/src/compile/jitc_yk/trace_builder.rs
+++ b/ykrt/src/compile/jitc_yk/trace_builder.rs
@@ -168,31 +168,22 @@ impl TraceBuilder {
                         let aot_field_off = trace_input_struct_ty.field_byte_off(trace_input_idx);
                         let aot_field_ty = trace_input_struct_ty.field_tyidx(trace_input_idx);
                         // FIXME: we should check at compile-time that this will fit.
-                        match u32::try_from(aot_field_off) {
-                            Ok(u32_off) => {
-                                let input_tyidx =
-                                    self.handle_type(self.aot_mod.type_(aot_field_ty))?;
-                                let load_ti_inst =
-                                    jit_ir::LoadTraceInputInst::new(u32_off, input_tyidx).into();
-                                // We don't need to insert these loads when we are side-tracing,
-                                // since side-tracing handles inputs separately.
-                                if !is_sidetrace {
-                                    self.jit_mod.push(load_ti_inst)?;
-                                    // If this take fails, we didn't see a corresponding store and the
-                                    // IR is malformed.
-                                    self.local_map.insert(
-                                        last_store_ptr.take().unwrap().to_inst_id(),
-                                        jit_ir::Operand::Local(self.jit_mod.last_inst_idx()),
-                                    );
-                                }
-                                self.first_ti_idx = iidx;
-                            }
-                            _ => {
-                                return Err(CompilationError::InternalError(
-                                    "Offset {aot_field_off} doesn't fit".into(),
-                                ));
-                            }
+                        let aot_field_off = u32::try_from(aot_field_off).unwrap();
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -168,31 +168,22 @@ impl TraceBuilder {
                         let aot_field_off = trace_input_struct_ty.field_byte_off(trace_input_idx);
                         let aot_field_ty = trace_input_struct_ty.field_tyidx(trace_input_idx);
                         // FIXME: we should check at compile-time that this will fit.
-                        match u32::try_from(aot_field_off) {
```
vext01, line 171: Quite hard to see the extent of the change from the diff. Is it just the outermost match that now has an unwrap?

I'd have made the `return Err` a `panic!`, but this is fine too.

ltratt: > Is it just the outermost match that now has an unwrap?

Yes, which matches other cases where we say "ykllvm should/does check this property".

tests/c/setlongjmp.newcg.c
```diff
@@ -7,7 +7,7 @@
 //     jitstate: start-tracing
 //     we jumped
 //     jitstate: stop-tracing
-//     jitstate: trace-compilation-aborted: longjmp encountered
```
vext01, line 10: I'm not sure about this. The jit-state prints (when enabled) are incomplete with this change.

I'm also not certain what the best cause of action is though.

Is `ykd` expensive at runtime?

vext01: Perhaps we should split the jit-state prints into two categories (a log level, if you will):
 - those only printed when `ykd` is on.
 - those you always want to print (like this one).

But I think the messages should be formatted the same for both, with a `jit-state: ` prefix.

vext01: Sorry, we raced.

OK, so how about this, we print the error as you have, but also emit a `jit-state` line as before. That way the `jit-state` info is complete, and someone grepping for `jit-state` won't be confused?

ltratt: I disagree: these fundamentally aren't "jit-state" things IMHO and shouldn't be presented as such. The JIT hasn't changed state as a result of these errors: the JIT state changed earlier (from tracing -> not tracing etc); subsequently an error has happened.

FWIW, I agree that there are probably more kinds of error/warning that need to be thought of than we currently have. For example "trace too long" is kind-of an expected thing and not something we want to print out all the time. But currently I have very little sense of how often it happens as I rarely have JIT state logging on.

ptersilie: I agree that it would still be useful to know that an error has lead the jit to abort a trace. `yk warning: longjmp encountered` doesn't really tell us that this lead to the trace being aborted. But that might be crucial information for an end-user to figure out why their program isn't faster.

ptersilie: Though I also agree that this technically isn't a jitstate. How about we write something like:
`yk warning: longjmp encountered (trace aborted)`.

vext01:
I think the first commit is fine.

An offline discussion led us to think harder about logging diagnostics, and I think a future PR will do something better than the second commit. Right @ltratt ?

ltratt:
Hopefully! OK to squash the second commit away?

vext01:
Go ahead. Thanks for your patience.

PR #1317, opened 2024-07-25
Type unimplemented AOT instructions.

vext01:
Requires https://github.com/ykjit/ykllvm/pull/186

When trying to debug something, I realised that the AOT module printer can crash under certain circumstances involving unimplemented instructions.

Some instructions decide their type by asking for the type if their operands. If that operand is an unimplemented instruction then `def_type()` returns `None` and this can crash the IR printer.

This change makes unimplemented instructions define a value of the correct type, thus making the module at least "type consistent".


vext01:
I should mention, this allows us to print most of the yklua AOT IR. It's still crashing on something to do with an unimplemented poison constant, but I'll handle that in my next PR.

All of this is just so I can debug an issue in yklua :)

PR #1318, opened 2024-07-25
Pass live variables into trace via registers and stack.

ptersilie:
This change gets rids of our reliance on the big trace inputs structs in order to pass live variables into a trace. Instead we place the trace frame right next to the main interpreter frame and load live variables directly from registers or the parent frame's stack.

There's is still some unnecessary moving of live variables at the beginning of the trace, since we currently can't set the live variable locations directly in the register allocator (e.g. we would like to say to the register allocator: this live variable lives in the parent stack frame at offset X). But this shouldn't be hard to fix in a future PR.

We also had to temporarily disable side-tracing (by setting the default threshold really high) as fixing this would have made this change even bigger than it already is. We will rectify this immediately in a follow-up PR.

Some crude experiments show this gives us a decent 25% speed-up. This will likely be higher once we remove the (now no longer required) trace inputs struct construction from ykllvm (also in a follow-up PR).

We can also reduce the amount of spills (and spill reloads) before (and after) each control point call, by naturally returning from the control point and jumping into the trace by overwriting the control-points return address (yet another follow-up PR).

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
@@ -4,6 +4,8 @@
 //!  - describes the generic interface to register allocators.
 //!  - contains concrete implementations of register allocators.
 
+use dynasmrt::x64::{Rq, Rx};
+
 /// Where is an SSA variable stored?
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub(crate) enum VarLocation {
@@ -14,13 +16,33 @@ pub(crate) enum VarLocation {
         /// Size in bytes of the allocation.
         size: usize,
     },
+    /// The SSA variable is a stack pointer with the value `RBP-frame_off`.
+    Direct {
+        /// The offset from the base of the trace's function frame.
+        frame_off: i32,
+        /// Size in bytes of the allocation.
+        size: usize,
+    },
+    /// The SSA variable is behind a pointer that's stored on the stack: `[RBP-frame_off]`.
+    Indirect {
+        /// The offset from the base of the trace's function frame.
+        frame_off: i32,
+        /// Size in bytes of the allocation.
+        size: usize,
+    },
     /// The SSA variable is in a register.
-    Register,
+    Register(Register),
     /// A constant integer `bits` wide (see [jit_ir::Const::ConstInt] for the constraints on the
     /// bit width) and with value `v`.
     ConstInt { bits: u32, v: u64 },
     /// A constant float.
     ConstFloat(f64),
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub(crate) enum Register {
+    GP(Rq), // general purpose
+    FP(Rx), // floating point
 }
 
 /// Indicates the direction of stack growth.
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -13,7 +13,7 @@ use super::{
         jit_ir::{self, BinOp, FloatTy, Inst, InstIdx, Module, Operand, Ty},
         CompilationError,
     },
-    reg_alloc::{StackDirection, VarLocation},
+    reg_alloc::{self, StackDirection, VarLocation},
     CodeGen,
 };
 #[cfg(any(debug_assertions, test))]
@@ -81,6 +81,17 @@ static ARG_FP_REGS: [Rx; 8] = [
     Rx::XMM7,
 ];
 
+/// Registers used by stackmaps to store live variables.
+static STACKMAP_GP_REGS: [Rq; 7] = [
+    Rq::RBX,
+    Rq::R12,
+    Rq::R13,
+    Rq::R14,
+    Rq::R15,
+    Rq::RSI,
+    Rq::RDI,
+];
+
 /// The argument index of the live variables struct argument in the JITted code function.
 static JITFUNC_LIVEVARS_ARGIDX: usize = 0;
 
@@ -138,6 +149,7 @@ struct Assemble<'a> {
     asm: dynasmrt::x64::Assembler,
     /// Deopt info, with one entry per guard, in the order that the guards appear in the trace.
     deoptinfo: Vec<DeoptInfo>,
+    ///
     /// Maps assembly offsets to comments.
     ///
     /// Comments used by the trace printer for debugging and testing only.
@@ -332,7 +344,6 @@ impl<'a> Assemble<'a> {
             jit_ir::Inst::IndirectCall(i) => self.cg_indirectcall(iidx, i)?,
             jit_ir::Inst::ICmp(i) => self.cg_icmp(iidx, i),
             jit_ir::Inst::Guard(i) => self.cg_guard(iidx, i),
-            jit_ir::Inst::Arg(i) => self.cg_arg(iidx, *i),
             jit_ir::Inst::TraceLoopStart => self.cg_traceloopstart(),
             jit_ir::Inst::SExt(i) => self.cg_sext(iidx, i),
             jit_ir::Inst::ZeroExtend(i) => self.cg_zeroextend(iidx, i),
@@ -367,13 +378,12 @@ impl<'a> Assemble<'a> {
 
         // Start a frame for the JITted code.
         dynasm!(self.asm
-            ; push rbp
-            ; mov rbp, rsp
             // Save pointers to ctrlp_vars and frameaddr for later use.
             ; push rdi
             ; push rsi
-            // Reset the basepointer so the spill allocator doesn't overwrite the two values we
-            // just pushed.
+            // Save base pointer which we use to access the parent stack.
+            ; push rbp
+            // Reset base pointer.
             ; mov rbp, rsp
         );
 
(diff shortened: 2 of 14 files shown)


ptersilie:
Depends on some ykllvm changes that I'm pushing soon.

ptersilie:
Actually, the speed up in `release-with-debug` mode is more between 1.08x (db.lua) and 1.18x (bigloop.lua). I'll have to update that in the commit message before merging.

ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
```diff
@@ -14,15 +16,33 @@ pub(crate) enum VarLocation {
         /// Size in bytes of the allocation.
         size: usize,
     },
+    Direct {
```
ltratt, line 19: Can we document `Direct` and `Indirect`? At face value they seem identical?

ltratt:
>we would like to say to the register allocator: this live variable lives in the parent stack frame at offset X

There isn't an API for that right now, but it would be trivial to do so: all it would require is something like this (untested!):

```rust
diff --git ykrt/src/compile/jitc_yk/codegen/x86_64/lsregalloc.rs ykrt/src/compile/jitc_yk/codegen/x86_64/lsregalloc.rs
index 1547c2fb..76916c67 100644
--- ykrt/src/compile/jitc_yk/codegen/x86_64/lsregalloc.rs
+++ ykrt/src/compile/jitc_yk/codegen/x86_64/lsregalloc.rs
@@ -164,6 +164,12 @@ impl<'a> LSRegAlloc<'a> {
         }
     }
 
+    pub(crate) fn set_spill_loc(&mut self, iidx: InstIdx, off: u32) {
+        let off = usize::try_from(off).unwrap();
+        debug_assert!(!off == USIZE::MAX);
+        self.spilled_insts[iidx] = off;
+    }
+
     /// Before generating code for the instruction at `iidx`, see which registers are no longer
     /// needed and mark them as [RegState::Empty]. Calling this allows the register allocator to
     /// use the set of available registers more efficiently.
```

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -737,44 +736,86 @@ impl<'a> Assemble<'a> {
 
     fn cg_loadtraceinput(&mut self, iidx: jit_ir::InstIdx, inst: &jit_ir::LoadTraceInputInst) {
         // Find the argument register containing the pointer to the live variables struct.
-        let base_reg = ARG_GP_REGS[JITFUNC_LIVEVARS_ARGIDX];
         match self.m.type_(inst.tyidx()) {
             Ty::Integer(_) | Ty::Ptr => {
+                let avoid = RegSet::from_vec(vec![
```
ltratt, line 741: Why are we avoiding these particular registers? Should they be a `const` with a doc string at the top of the file?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -737,44 +736,86 @@ impl<'a> Assemble<'a> {
...
+                match self.m.tilocs()[usize::try_from(inst.locidx()).unwrap()] {
+                    VarLocation::Register(reg_alloc::Register::GP(reg)) => {
+                        // FIXME: Map this iidx to `reg` without this `mov`. Requires way to
+                        // initialise register allocator with existing values.
+                        dynasm!(self.asm; mov Rq(tgt_reg.code()), Rq(reg.code()));
                     }
-                    2 => {
-                        dynasm!(self.asm ; movzx Rq(tgt_reg.code()), WORD [Rq(base_reg.code()) + off])
+                    VarLocation::Register(reg_alloc::Register::FP(_)) => {
+                        panic!()
```
ltratt, line 766: I guess this is a todo rather than "something has gone wrong"?

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -504,6 +513,10 @@ impl Module {
     ) -> Result<GuardInfoIdx, CompilationError> {
         GuardInfoIdx::new(self.guard_info.len()).inspect(|_| self.guard_info.push(info))
     }
+
+    pub(crate) fn tilocs(&self) -> &Vec<VarLocation> {
```
ltratt, line 517: I expect Clippy will ask you to do `-> &[VarLocation]` ?

ltratt:
This broadly looks good and my comments above aren't very deep. A few slightly higher-level, if random, comments:

1. We don't have a test for any of the machine code output AFAICS? Or maybe `cg_call_with_different_args` suffices? Given this stuff is quite a delicate dance, I wondered if a specific test called something like `traceinput_unpack` might be worthwhile?
2. I think this has always been the case but it's now very clear that `load_trace_input`s must (a) come at the start of the trace (b) come before all other instructions. Maybe a quick check in `well_formed.rs` that `LoadTraceInput` doesn't appear later in the trace would be good?
3. I was wondering what other invariants/assumptions you might have uncovered (or added) as part of this PR? I wonder if there's anything else we can usefully document/assert/well-formedness-check etc?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ptersilie: These are the caller-saved registers, which are the only registers that are safe to use for live variables before a function call (and thus the only registers referenced by stackmaps).

Yeah those could be a `const` at the top. Will fix.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ptersilie: We know from the outermost `match` that the value is either an integer or a pointer, so I don't expect those to be stored in floating point registers. The same applies for float values which I don't expect to be stored in a normal register (see below).

ptersilie:
> > we would like to say to the register allocator: this live variable lives in the parent stack frame at offset X
> 
> There isn't an API for that right now, but it would be trivial to do so: all it would require is something like this (untested!):
> 

It's slightly more complicated than that. Ideally, I'd like to execute the trace in the same frame as the main interpreter frame. But that would require to tell the register allocator to initialise on an existing stack frame. But it might be as simple as saying `reg_alloc.rsp = parent_frame.size` or similar.

ptersilie:
> This broadly looks good and my comments above aren't very deep. A few slightly higher-level, if random, comments:
> 
>     1. We don't have a test for any of the machine code output AFAICS? Or maybe `cg_call_with_different_args` suffices? Given this stuff is quite a delicate dance, I wondered if a specific test called something like `traceinput_unpack` might be worthwhile?

There are some trace input tests like `cg_loadtraceinput_i8`. But until we have extended the parser to allow us to create different `load_ti` instructions, we can't really test this properly.

>     2. I think this has always been the case but it's now very clear that `load_trace_input`s must (a) come at the start of the trace (b) come before all other instructions. Maybe a quick check in `well_formed.rs` that `LoadTraceInput` doesn't appear later in the trace would be good?

I will look into this.

>     3. I was wondering what other invariants/assumptions you might have uncovered (or added) as part of this PR? I wonder if there's anything else we can usefully document/assert/well-formedness-check etc?

Nothing from the top of my head, but will try and think a bit about this.

vext01:
> 2. I think this has always been the case but it's now very clear that load_trace_inputs must (a) come at the start of the trace

Hah, good point! Nice spot.

ltratt:
> It's slightly more complicated than that. Ideally, I'd like to execute the trace in the same frame as the main interpreter frame. But that would require to tell the register allocator to initialise on an existing stack frame. But it 
might be as simple as saying reg_alloc.rsp = parent_frame.size or similar.

The register allocator doesn't do anything to `rsp`: all it knows is "instruction X's value is at `rbp-<off>`. I appreciate that might not be enough, though! Still hopefully it shows that in this regard the register allocator is pretty simple, and *hopefully* will fit well with whatever you need to do.

>But until we have extended the parser to allow us to create different load_ti instructions, we can't really test this properly.

OK, I'll look into that when this PR merges.

ptersilie:
> The register allocator doesn't do anything to `rsp`: all it knows is "instruction X's value is at `rbp-<off>`. I appreciate that might not be enough, though! Still hopefully it shows that in this regard the register allocator is pretty simple, and _hopefully_ will fit well with whatever you need to do.

What I meant was that we need to be able to tell the register allocator there's already a bunch of space reserved from the parent frame. So instead of putting the first spill value at [rbp - 8], it needs to take into account that the parent frame already has 80 bytes used, so the first value would go in [rbp - 88].

ykrt/src/compile/jitc_yk/codegen/reg_alloc/mod.rs
ptersilie: [9e8f08d](https://github.com/ykjit/yk/pull/1318/commits/9e8f08db1489fa33578063838ef651345903fcf1)

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ptersilie: [4d4dd1f](https://github.com/ykjit/yk/pull/1318/commits/4d4dd1f925705c7da45b7a69ad4f72d61a350141)

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ptersilie: It didn't but I changed it anyway: [0bbdc78](https://github.com/ykjit/yk/pull/1318/commits/0bbdc78673741da4ccafd26ad5737d207b1c0d87)

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ltratt: Does "expect" mean "I don't think can happen" or "I know for sure can't happen"? I think the latter. If so, is a comment worthwhile? It might be obvious enough that it's not worth doing, so please do what you see fit.

ltratt:
> What I meant was that we need to be able to tell the register allocator there's already a bunch of space reserved from the parent frame. So instead of putting the first spill value at [rbp - 8], it needs to take into account that the parent frame already has 80 bytes used, so the first value would go in [rbp - 88].

Right, so that would require changing `LSRegalloc::new` and `Stack::new` (and removing the `Default` implementation for `Stack`) to pass through the "initial offset" (which is currently `0`). Hopefully not too difficult, but not for this PR obviously!

ptersilie:
Added a well-formedness check for `load_ti`: [e33b85d](https://github.com/ykjit/yk/pull/1318/commits/e33b85dcf6c03df09018a2fdde490a1e6afff9c6)

ptersilie:
I believe this is ready for a re-review now.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ltratt: This is the only remaining comment.

ptersilie: My knowledge of x64 is still basic, so I don't know "for sure". But from what I understand the compiler shouldn't store normal integer values inside a floating point register. Added a comment in [bc99e72](https://github.com/ykjit/yk/pull/1318/commits/bc99e72e89a0b77d1d26888b4c86a8b9c266ce0f), similar to the ones I had already added for the reverse case.

vext01:
All fine by me.

ltratt:
Please squash.

ptersilie:
Squashed.

ptersilie:
Oh forgot to update the submodule...

ptersilie:
I think it's now just a matter of waiting for https://github.com/ykjit/yk/pull/1321 to go in as that updates the submodule already.

ptersilie:
I've tried updating the ykllvm submodule, but it's already up to date upstream so there was nothing to do for me here.

ptersilie:
It's this error again "the `` attribute is incompatible with `#[naked]`"

ltratt:
@ptersilie You'll need to remove the doc comment in the PR at https://github.com/ykjit/yk/pull/1318/files#diff-e609c9c537d75440ea3db6f14fa49100e4e511dab07bfb704cc4e8199fb1fae3R52 or turn it into `//` comment. No, I don't know why rustc complains about this! Please squash directly.

ptersilie:
Done.

ltratt:
@ptersilie Please force push a rebase.

ptersilie:
Ready

PR #1320, opened 2024-07-25
Fix two more AOT IR printer crashes.

vext01:
Requires: https://github.com/ykjit/ykllvm/pull/188

With these changes I can finally print the AOT IR for yklua in its entirety.


vext01:
Looks like an unrelated bug causing yklua to crash.

free(): invalid size

(We don't print AOT IR in the yklua tests)

We could retry, but eventually we should try to track these crashes down...

PR #1323, opened 2024-07-26
Rethink normal logging.

ltratt:
Previously we lumped a lot of logging under `YKD_LOG_JITSTATE` including JIT state transitions, location transitions, warnings, and genuine errors. That meant that you either saw everything or (most often) nothing.

This commit removes `YKD_LOG_JITSTATE` and adds
`YK_LOG=[<path|->:]<level>`. It defaults to logging to stderr at level 1 which is "log errors". I think everything up to level 3 is useful. I mostly don't find level 4 useful, but I'm keeping it for backwards compatibility.

Most of the C tests just change "jitstate" to "yk-jit-event" with the exception of `setlongjmp.newcg.c` which has this diff:

```
-//     jitstate: stop-tracing
-//     jitstate: trace-compilation-aborted: longjmp encountered
+//     yk-jit-event: stop-tracing
+//     yk-warning: trace-compilation-aborted: longjmp encountered
```

### DIFF
--- a/docs/src/dev/runtime_config.md
+++ b/docs/src/dev/runtime_config.md
@@ -10,6 +10,19 @@ Variables prefixed with `YK_` allow the user to control aspects of yk's executio
 
 The following environment variables are available:
 
+* `YK_LOG=[<path>:]<level>` specifies where, and how much, general information
+  yk will log during execution.
+
+  If `<path>:` (i.e. a path followed by ":") is specified then output is sent
+  to that path. The special value `-` (i.e. a single dash) can be used for
+  `<path>` to indicate stderr. If not specified, logs to stderr.
+
+  `<level>` specifies the level of logging: level 0 turns off all yk logging;
+  level 1 shows major errors only; level 2 warnings; and level 3 and above
+  being increasingly granular information about transitions within yk. Some
+  information may or may not be displayed based on compile-time options.
+  Defaults to 1.
+
 * `YK_HOT_THRESHOLD`: an integer from 0..4294967295 (both inclusive) that
   determines how many executions of a hot loop are needed before it is traced.
   Defaults to 50.
@@ -26,5 +39,4 @@ The following environment variables are available (some only in certain configur
 
 * [`YKD_LOG_IR`](understanding_traces.html#ykd_log_ir) [with the `ykd` feature]
 * [`YKD_TRACE_DEBUGINFO`](understanding_traces.html#ykd_trace_debuginfo) [with the `ykd` feature]
-* [`YKD_LOG_JITSTATE`](understanding_traces.html#ykd_log_jitstate) [with the `ykd` feature]
 * [`YKD_LOG_STATS`](profiling.html#jit-statistics)
--- a/tests/c/dyn_ptradd_mixed.newcg.c
+++ b/tests/c/dyn_ptradd_mixed.newcg.c
@@ -2,11 +2,11 @@
 // Run-time:
 //   env-var: YKD_LOG_IR=-:aot,jit-pre-opt
 //   env-var: YKD_SERIALISE_COMPILATION=1
-//   env-var: YKD_LOG_JITSTATE=-
+//   env-var: YK_LOG=4
 //   stderr:
-//     jitstate: start-tracing
+//     yk-jit-event: start-tracing
 //     i=4, y=7
-//     jitstate: stop-tracing
+//     yk-jit-event: stop-tracing
 //     --- Begin aot ---
 //     ...
 //     %{{9_4}}: ptr = ptr_add @line, 4 + (%{{9_3}} * 8)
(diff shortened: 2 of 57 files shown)


vext01:
Is the long-term plan to default to level 0? I suppose once the JIT is a bit more stable, we shouldn't print internal retryable JIT problems to users?

vext01:
Are there any follow up PRs to come after this? We discussed printing the address of the location for state transitions, and printing `from-state -> to-state` for state transitions.

docs/src/dev/runtime_config.md
```diff
@@ -10,6 +10,19 @@ Variables prefixed with `YK_` allow the user to control aspects of yk's executio
 
 The following environment variables are available:
 
+* `YK_LOG=[<path>:]<level>` specifies where, and how much, general information
+  yk will log during execution.
+
+  If `<path>:` (i.e. a path followed by ":") is specified then output is sent
+  to that path. The special value `-` (i.e. a single dash) can be used for
+  `<path>` to indicate stderr. If not specified, logs to stderr.
```
vext01, line 18: One for another day: maybe we should make paths in other env vars (e.g. YKD_LOG_IR) optional too for consistency?

tests/c/dyn_ptradd_mixed.newcg.c
```diff
@@ -19,10 +19,10 @@
 //     ...
 //     --- End jit-pre-opt ---
 //     i=3, y=6
-//     jitstate: enter-jit-code
+//     yk-jit-event: enter-jit-code
```
vext01, line 22: Dumb question: whats is the definition of a `jit-event` under this new model? I was expecting `enter-jit-code` to be printed differently from state transitions, since entering jit-code doesn't actually cause a state change.

tests/c/setlongjmp.newcg.c
```diff
@@ -2,12 +2,12 @@
...
-//   env-var: YKD_LOG_JITSTATE=-
+//   env-var: YK_LOG=4
 //   stderr:
-//     jitstate: start-tracing
+//     yk-jit-event: start-tracing
 //     we jumped
-//     jitstate: stop-tracing
-//     jitstate: trace-compilation-aborted: longjmp encountered
+//     yk-jit-event: stop-tracing
+//     yk-warning: trace-compilation-aborted: longjmp encountered
```
vext01, line 10: should there have been a `yk-jit-event` line informing us of a state transition (separate from the warning?)

vext01:
In hindsight, I wish I'd proposed the ability to turn on/off individual classes of things, so for example if I wanted to print only state transitions I could do `YKD_LOG=-:loc-state-trans` or similar.

That ship has sailed I fear.

tests/c/setlongjmp.newcg.c
ltratt: ?

docs/src/dev/runtime_config.md
ltratt: Wouldn't hurt but agreed: another PR.

ltratt:
> In hindsight, I wish I'd proposed the ability to turn on/off individual classes of things, so for example if I wanted to print only state transitions I could do YKD_LOG=-:loc-state-trans or similar.

Another PR could do that. I am not sure it's very useful, though, so I personally wouldn't consider it high priority.

tests/c/dyn_ptradd_mixed.newcg.c
vext01: Exactly! I was expecting those two things to be formatted differently.

ltratt: `enter-jit-code` is a JIT state change, not a location state change. "is-compiling" (or whatnot) would be a location state change. [FWIW, this PR only partially fixes this sort of thing. A full job requires a complete audit of all of ykrt, because we haven't got the right hooks in the right places yet. It's not a small job, and I don't want to do it right now.]

vext01: Agreed.

I was expecting classes like:
 - warning (e.g. trace too long)
 - loc-state-change: (e.g. location 0x123 went from counting to tracing)
 - jit-event: (e.g. enter jit code)

But I'm now pretty sure I've expected too much in one go, so I won't persist now.

tests/c/setlongjmp.newcg.c
ltratt: There's a real tension with "flexibility" and "usability". I don't know what the right trade-off is: my main concern (addressed in this PR) was that we were in a real mess with warnings/errors to the extent that we didn't understand how the system was running. I've sorted that out in this PR, but I don't have further concerns myself. If someone wants to do more, they are welcome to do so, but personally I'm done for now (famous last words).

vext01:
We've agreed offline to roll with this.

PR #1327, opened 2024-07-29
Remove last `WR[0-1]` usages

ltratt:
The old spill allocator used fixed work registers `WR0`, `WR1`, and `WR2`. The new register allocator remove `WR2`, and should have removed `WR1` (https://github.com/ykjit/yk/commit/4d762197eba048f60d4344379ee8811e3c657842). This PR removes `WR0`, allowing us to use `r12` as a general purpose register allocator. This doesn't make much of a benchmarking difference, because other inefficiencies mean that we don't gain much from having one more register to use, but at some point those other inefficiencies will go away, and then each register will be useful! In that sense, this PR is a small but useful step on that journey.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/lsregalloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/lsregalloc.rs
@@ -995,23 +995,23 @@ impl<'a> LSRegAlloc<'a> {
 /// What constraints are there on registers for an instruction?
 #[derive(Debug)]
 pub(crate) enum RegConstraint<R: Register> {
-    // Make sure `Operand` is loaded into a register *R* on entry; its value must be unchanged
-    // after the instruction is executed.
+    /// Make sure `Operand` is loaded into a register *R* on entry; its value must be unchanged
+    /// after the instruction is executed.
     Input(Operand),
-    // Make sure `Operand` is loaded into register `R` on entry; its value must be unchanged
-    // after the instruction is executed.
+    /// Make sure `Operand` is loaded into register `R` on entry; its value must be unchanged
+    /// after the instruction is executed.
     InputIntoReg(Operand, R),
-    // Make sure `Operand` is loaded into register `R` on entry and considered clobbered on exit.
+    /// Make sure `Operand` is loaded into register `R` on entry and considered clobbered on exit.
     InputIntoRegAndClobber(Operand, R),
-    // Make sure `Operand` is loaded into a register *x* on entry and considered clobbered on exit.
-    // The result of this instruction will be stored in register *x*.
+    /// Make sure `Operand` is loaded into a register *x* on entry and considered clobbered on exit.
+    /// The result of this instruction will be stored in register *x*.
     InputOutput(Operand),
-    // Make sure `Operand` is loaded into register `R` on entry and considered clobbered on exit.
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/codegen/x86_64/lsregalloc.rs
```diff
@@ -1013,6 +1035,9 @@ pub(crate) enum RegConstraint<R: Register> {
     Output,
     /// The result of this instruction will be stored in register `R`.
     OutputFromReg(R),
+    /// A temporary register *x*: it will be clobbered by the instruction and left in an
```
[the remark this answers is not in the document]
ltratt: A new register. It's in the same format as the other comments (for better or worse).

ykrt/src/compile/jitc_yk/codegen/x86_64/lsregalloc.rs
```diff
@@ -1013,6 +1035,9 @@ pub(crate) enum RegConstraint<R: Register> {
...
+    /// indeterminate state on exit.
```
vext01, line 1035: Does this mean a temporary is immediately "unallocated" before the next instruction?

ykrt/src/compile/jitc_yk/codegen/x86_64/lsregalloc.rs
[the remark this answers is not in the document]
vext01: OK, I thought I was missing another reference to `x` elsewhere.

ykrt/src/compile/jitc_yk/codegen/x86_64/lsregalloc.rs
vext01: If it spills, where is the original value restored, or does it not work like that?

vext01: Let me ask more precisely: If requesting a temporary register causes a spill, don't we need to restore the spilled value afterwards? If so, where does this happen?

vext01:
Looks good. Just some (probably naive) questions.

ykrt/src/compile/jitc_yk/codegen/x86_64/lsregalloc.rs
ltratt: The register allocator handles spilling: it returns to the caller a register they can use, and in a sense that's all they have to worry about. `Temporary` is not dissimilar to `Output`: it spills (if necessary) anything already in that register; and then you can write to that register to your heart's content.

vext01: Ah, so once something spills, it's spilled for good?

ltratt: Yes. It means we spill each SSA value at most once.

vext01: Interesting. That's a design I hadn't considered.

Does this extend to calls? Do we evict potentially frequently-used variables to the stack for caller save?

ltratt: Yes, we clobber according to the SysV ABI.

PR #1287, opened 2024-07-04
Implement side-tracing.

ptersilie:
Depends on https://github.com/ykjit/ykllvm/pull/180 and requires an update to the ykllvm submodule before merging.

Implement a basic version of side-tracing. The code can still be simplified a bit and there's some corner cases we don't handle yet.

PR #1328, opened 2024-07-29
Remove the LLVM JIT compiler

ltratt:
This PR removes most (but probably not yet all) vestiges of the LLVM JIT compiler, pushing further what was started in https://github.com/ykjit/yk/pull/1287. Each commit hopefully stands alone, and I've tried to alternate between "change code" and "delete lots of files" so the latter doesn't drown out diffs in the former.

By the end of the PR, `YKD_NEW_CODEGEN` is no longer a thing. Only the "new codegen" is a thing.

The main remaining job is to port what are now `old.c` tests to the new code gen. Some work as-is, and some are probably easily ported, but we should soon examine each and do what we can to reinstate those tests. Doing that will probably be a bit-by-bit job, and since we can't run them now anyway, this PR does improve the situation (I resurrected several tests), even if we need and want to push it further soon.


ptersilie:
This looks good to me. If @vext01 has nothing to add I think we can merge this.

vext01:
> The main remaining job is to port what are now old.c tests to the new code gen. Some work as-is, and some are probably easily ported, but we should soon examine each and do what we can to reinstate those tests.

There are probably also duplicates, in which case the old test can just be deleted. But yes, this will need to be done manually.

ltratt:
> There are probably also duplicates, in which case the old test can just be deleted.

I have dealt with the obvious duplicates (where the test file name showed it was a duplicate) but there are probably "harder" remaining cases.

ltratt:
Concretely, (hopefully) the commit message in https://github.com/ykjit/yk/pull/1328/commits/08bb97d13c2ad471eb90670003a06f5a405fbc81 shows the deduplication this PR does.

vext01:
OK, let's do this, and I'll have a grep around later to see if I can find any llvm detritus left-over.

If @ptersilie is OK with this, please go ahead.

ltratt:
You have both said you're OK so now one of you needs to press merge: I can't do it :p

PR #1330, opened 2024-07-30
Don't preemptively spill variables before a guard.

ltratt:
Previously the only way to pass an instruction's value to a guard was via spill -- so if a value was referenced in a guard and not already spilled, we had to spill it just before the guard, even if there was no other reason to do so.

This commit removes this penalty from the fast path: guards and deopt (sort of...) can take values from registers rather than spills.

For example here's a guard from a test before this commit:

```
 ; guard true, %304, [<lots of live variables>]
 : mov [rbp-0x930], r10
 : mov [rbp-0x938], r9
 : mov [rbp-0x940], r8
 : mov [rbp-0x948], rdx
 : mov [rbp-0x950], rcx
 : mov [rbp-0x958], rax
 : mov [rbp-0x960], r11
 : cmp r15b, 0x01
 : jnz 0xCB3
 ; %306: ptr = load %6
```

and after:

```
 ; guard true, %304, [<lots of live variables>]
 : cmp r15b, 0x01
 : jnz 0xC4A
 ; %306: ptr = load %6
```

This is a fairly extreme example (most guards tend to spill 1 or 2 variables), but it shows 7 spills being removed entirely.

Now to some extent, what this commit does is shuffle the problem around a bit: we still have to pass lots of register values to deopt, and there's not enough room to do so. We could use the SysV ABI in a clever way, but instead we push all registers (including reserved registers... which is silly... but easy) in a failing guard to the stack, and then fish values out in deopt.

What this means is that the happy path (i.e. a guard that succeeds) doesn't encounter any spills, but the unhappy path encounters lots. Since the latter case is slow anyway, we don't really notice the costs: for benchmarks which have lots of failing guards, this commit has no real impact. For benchmarks which stay in JITed code we have a noticable benefit (e.g. big_loop.lua speeds up by a bit over 7%).

There's more we could do to make this more sophisticated and complete (e.g. we'll hit a `TODO` if/when we hit a guard which references a floating point variable --- none of our tests does so), but this is a step in the right direction.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -249,10 +249,14 @@ impl<'a> Assemble<'a> {
             for (deoptid, fail_label) in fail_labels.into_iter().enumerate() {
                 #[cfg(any(debug_assertions, test))]
                 self.comment(self.asm.offset(), format!("Deopt ID for guard {deoptid}"));
+                // FIXME: Why are `deoptid`s 64 bit? We're not going to have that many guards!
+                let deoptid = i32::try_from(deoptid).unwrap();
                 dynasm!(self.asm
                     ;=> fail_label
-                    ; mov rsi, QWORD deoptid as i64
-                    ; jmp =>deopt_label
+                    ; push rsi // FIXME: We push RSI now so we can fish it back out in
+                               // `deopt_label`.
+                    ; mov rsi, deoptid
+                    ; jmp => deopt_label
                 );
             }
 
@@ -262,11 +266,25 @@ impl<'a> Assemble<'a> {
             // is x86 only, we don't need to worry about portability.
             #[allow(clippy::fn_to_numeric_cast)]
             {
+                dynasm!(self.asm; => deopt_label);
+                // Push all the general purpose registers to the stack.
+                for (i, reg) in lsregalloc::GP_REGS.iter().rev().enumerate() {
+                    if *reg == Rq::RSI {
+                        // RSI is handled differently in `fail_label`: RSI now contains the deopt
+                        // ID, so we have to fish the actual value out of the stack. See the FIXME
+                        // in `fail_label`.
+                        let off = i32::try_from(i * 8).unwrap();
+                        dynasm!(self.asm; push QWORD [rsp + off]);
+                    } else {
+                        dynasm!(self.asm; push Rq(reg.code()));
+                    }
+                }
                 dynasm!(self.asm
-                    ;=> deopt_label
                     ; mov rdi, [rbp]
                     ; mov rdx, rbp
+                    ; mov rcx, rsp
                     ; mov rax, QWORD __yk_deopt as i64
+                    ; sub rsp, 8 // Align the stack
                     ; call rax
                 );
             }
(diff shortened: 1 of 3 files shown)


ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -249,10 +249,14 @@ impl<'a> Assemble<'a> {
             for (deoptid, fail_label) in fail_labels.into_iter().enumerate() {
                 #[cfg(any(debug_assertions, test))]
                 self.comment(self.asm.offset(), format!("Deopt ID for guard {deoptid}"));
+                // FIXME: Why are `deoptid`s 64 bit? We're not going to have that many guards!
```
ptersilie, line 252: Because we just pass them through from LLVM's stackmap where they are `i64`.

ptersilie: Actually, that's not true. They are not the same as stackmap IDs. So we can turn them into `i32`s if you like.

ltratt: I *think* these are internal to yk, so they're per-trace? If so, we could probably get away with `u16`s! I guess we leave things as-is for this PR and fix this later?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -262,11 +266,25 @@ impl<'a> Assemble<'a> {
...
+                        dynasm!(self.asm; push QWORD [rsp + off]);
+                    } else {
+                        dynasm!(self.asm; push Rq(reg.code()));
+                    }
+                }
                 dynasm!(self.asm
-                    ;=> deopt_label
                     ; mov rdi, [rbp]
                     ; mov rdx, rbp
+                    ; mov rcx, rsp
```
ptersilie, line 285: Now that is something I had not thought of doing.

ltratt: I am not sure if I detect admiration or disgust :p

ptersilie: I was undecided at first, but I think I have settled on admiration. I would probably have used a `naked` function to pull the registers back out from the stack, but turning `rsp` into `[u64; 16]` is very clever.

ltratt: I just did the dumbest thing I could think of. I'm pretty sure you can come up with a better / faster approach (e.g. passing some values in registers rather than spilling), but I figured that we (probably you, because you're better at this than me) could do this in a later PR.

ptersilie:
Okay, I get it now. I was looking for direct register accesses, which I think was the plan when you started this. But instead you just moved the spills into the guard path. I agree that the performance gain from accessing registers directly is probably limited anyway. We can look into that in the future to tickle out the last few percentages.

I think what you've also done here by accident is solve part of my side-tracing problem. I believe I can now take the information from here to tell the side-trace from where to pick up the live variables.

ltratt:
> I think what you've also done here by accident is solve part of my side-tracing problem.

If so, I'm happy!

PR #1333, opened 2024-07-31
Move from `<index type>::new` to `<index type>::try_from`.

ltratt:
This commit makes use of the standard `TryFrom` trait to make clear that `InstIdx` and friends can return a `Result`. This is slightly wordier than `new`, but doing this will allow us to add another method in the future to make clear when we're overriding the possibility of an `Err` being returned.


ltratt:
@vext01 Notice that I've pushed a second commit here (and updated the PR description to boot).

vext01:
LGTM. Any squashing required?

ltratt:
Nope!

Issue #1334, opened 2024-08-02
What do all the things in yksmp mean?


ltratt:
yksmp has very few docstring, which meant that when I look at e.g. [Location](https://github.com/ykjit/yk/blob/master/yksmp/src/lib.rs#L36) I am left confused:

```rust
pub enum Location {
    Register(u16, u16, i32, u16),
    Direct(u16, i32, u16),
    Indirect(u16, i32, u16),
    Constant(u32),
    LargeConstant(u64),
}
```

Some of these are somewhat obvious (`Constant`) but can we at least briefly document these sorts of things? If nothing else, maybe we can link to the relevant part(s) of the LLVM spec?

ltratt:
So here's something that's really surprised me: `Loaction::Direct` can return both positive and negative stack offsets! I assumed the `i32` was always `>= 0` but a quick test run on yk's lang_tests shows values such as [-56, 0, 4]. I'm not quite sure how to interpret this: does a positive value mean that we reach into a parent function frame? If so, how does that interact with trace frames?

vext01:
The document you are looking for is:
https://llvm.org/docs/StackMaps.html#stack-map-format

![image](https://github.com/user-attachments/assets/0074a5a0-53a6-45dd-a64b-47e1459365ef)

> does a positive value mean that we reach into a parent function frame?

It depends what register the offset is relative to.

Notice how a direct record is a `Reg + Offset` pair.

> Loaction::Direct can return both positive and negative stack offsets!

An example, the same stack location can be addressed (on x64) as a positive offset from the stack pointer, or a negative index from the base pointer.

The register involved may not even be a stack-related register. It could be any old register (e.g. `rax`) holding any old address that (after the offset is applied) points into the stack somewhere.

ltratt:
Fixed in https://github.com/ykjit/yk/pull/1355.

PR #1335, opened 2024-08-02
De x64-ise

ltratt:
The main part of this PR is https://github.com/ykjit/yk/commit/ec0d82ec2068a4055c1f53f396d3e22b30143333. Hopefully the comment therein explains what's going on.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
@@ -46,6 +46,7 @@ use std::sync::{Arc, Weak};
 #[cfg(any(debug_assertions, test))]
 use std::{cell::Cell, slice};
 use ykaddr::addr::symbol_to_ptr;
+use yksmp;
 
 mod deopt;
 mod lsregalloc;
@@ -753,7 +754,40 @@ impl<'a> Assemble<'a> {
     }
 
     fn cg_loadtraceinput(&mut self, iidx: jit_ir::InstIdx, inst: &jit_ir::LoadTraceInputInst) {
-        // Find the argument register containing the pointer to the live variables struct.
+        let m = match &self.m.tilocs()[usize::try_from(inst.locidx()).unwrap()] {
+            yksmp::Location::Register(3, ..) => {
+                VarLocation::Register(reg_alloc::Register::GP(Rq::RBX))
+            }
+            yksmp::Location::Register(4, ..) => {
+                VarLocation::Register(reg_alloc::Register::GP(Rq::RSI))
+            }
+            yksmp::Location::Register(5, ..) => {
+                VarLocation::Register(reg_alloc::Register::GP(Rq::RDI))
+            }
+            yksmp::Location::Register(12, ..) => {
+                VarLocation::Register(reg_alloc::Register::GP(Rq::R12))
+            }
+            yksmp::Location::Register(13, ..) => {
+                VarLocation::Register(reg_alloc::Register::GP(Rq::R13))
+            }
+            yksmp::Location::Register(14, ..) => {
+                VarLocation::Register(reg_alloc::Register::GP(Rq::R14))
+            }
+            yksmp::Location::Register(15, ..) => {
+                VarLocation::Register(reg_alloc::Register::GP(Rq::R15))
+            }
+            yksmp::Location::Direct(6, off, size) => VarLocation::Direct {
+                frame_off: *off,
+                size: usize::from(*size),
+            },
+            yksmp::Location::Indirect(6, off, size) => VarLocation::Indirect {
+                frame_off: *off,
+                size: usize::from(*size),
+            },
+            e => {
+                todo!("{:?}", e);
+            }
+        };
         match self.m.type_(inst.tyidx()) {
             Ty::Integer(_) | Ty::Ptr => {
                 let [tgt_reg] = self.ra.get_gp_regs_avoiding(
@@ -765,7 +799,7 @@ impl<'a> Assemble<'a> {
                 let size = self.m.inst_no_proxies(iidx).def_byte_size(self.m);
                 debug_assert!(size <= REG64_SIZE);
 
-                match self.m.tilocs()[usize::try_from(inst.locidx()).unwrap()] {
+                match m {
                     VarLocation::Register(reg_alloc::Register::GP(reg)) => {
                         // FIXME: Map this iidx to `reg` without this `mov`. Requires way to
                         // initialise register allocator with existing values.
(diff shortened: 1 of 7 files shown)


vext01:
> yksmp::Location::Register(3, ..)
>
> What is the `3`?

To answer this part, yksmp is likely mirroring the register numbering scheme used in the [LLVM stackmap format](https://llvm.org/docs/StackMaps.html#stack-map-format), which is the DWARF convention. Registers are given abstract numbers, and each platform (or ABI?) has a mapping from abstract DWARF register number to hardware register.

The x86_64 DWARF mapping can be found in the [x86_64 SysV ABI spec](https://cs61.seas.harvard.edu/site/pdf/x86-64-abi-20210928.pdf):

image

vext01:
I didn't understand this part of the commit message:

> This creates a particular challenge for the JIT IR parser. Previously we assigned every instruction to either `RBX` or `XMM0` but that isn't going to work for very long, because optimisation passes could work out they're aliases and optimise things away. This commit uses distinct (general purpose? depends on yksmp's semantics!) registers for most types but puts floating point types on the heap

At the cost of looking dumb, what is the problem we are trying to solve here? Allowing the JIT IR syntax to express trace input register allocation decisions in a platform independent way? Would the DWARF numbers suffice?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -753,7 +754,40 @@ impl<'a> Assemble<'a> {
     }
 
     fn cg_loadtraceinput(&mut self, iidx: jit_ir::InstIdx, inst: &jit_ir::LoadTraceInputInst) {
-        // Find the argument register containing the pointer to the live variables struct.
+        let m = match &self.m.tilocs()[usize::try_from(inst.locidx()).unwrap()] {
+            yksmp::Location::Register(3, ..) => {
```
vext01, line 758: Hey look! This is just (a subset of) the DWARF SysV register mapping!

I'm now about 99% percent sure `yksmp` is using dwarf register numbers. Maybe (later) we can add a comment in `VarLocation::Register` saying so. Then these are no longer magic numbers.

ltratt: It would be good to do that, preferably someone who isn't me, and actually understands this :p

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
```diff
@@ -775,17 +809,14 @@ impl<'a> Assemble<'a> {
...
                         // stack size and don't create a new frame in the trace epilogue we can
                         // reference values directly using the current RBP value.
-                        dynasm!(self.asm; mov Rq(tgt_reg.code()), QWORD [Rq(Rq::RBP.code())]);
-                        match size {
-                            8 => {
-                                dynasm!(self.asm; lea Rq(tgt_reg.code()), [Rq(tgt_reg.code()) + frame_off])
-                            }
-                            _ => todo!(),
-                        }
+                        dynasm!(self.asm
```
vext01, line 816: How comes we no longer need to inspect the size here?

ltratt: I don't understand this code. It seems to work, but I don't really understand why the `lea` exists.

ykrt/src/compile/jitc_yk/jit_ir/parser.rs
```diff
@@ -128,6 +127,13 @@ impl<'lexer, 'input: 'lexer> JITIRParser<'lexer, 'input, '_> {
         self.process_func_types(func_types)?;
         self.process_func_decls(func_decls)?;
 
+        // We try and put trace inputs into registers, but place floating point values on the stack
+        // as yksmp currently doesn't seem able to differentiate general purpose from floating
+        // point.
+        #[cfg(target_arch = "x86_64")]
+        let mut reg_off: u16 = 3; // FIXME: Why do need to start from register 3? I have no idea.
```
vext01, line 134: FWIW I have no idea either.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
vext01: I think we need to ask Lukas when he's back. Perhaps we could raise an issue and assign him, as this is probably important enough that we don't want it to get lost?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ltratt: Oh, now I get it! These trace inputs become `ptr`s e.g. `%0: ptr = load_ti Direct(6, -32, 8)`.

ltratt: So I misunderstood how this worked earlier, but now I get it. Restored/fixed in 6f7157c28bfd46bfa4de6927a8c78b6675af0fe8.

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
ltratt: Hopefully that will shake out naturally from #1334.

ltratt:
@vext01 I *think* this is ready for squashing?

ykrt/src/compile/jitc_yk/codegen/x86_64/mod.rs
vext01: Right. `lea` is "load effective address". From memory it was designed to do certain pointer arithmetic operations to be done in one instruction (without explicit `add` and `mul`).

Nowadays people abuse it to do all kind of things :P

ykrt/src/compile/jitc_yk/jit_ir/parser.rs
```diff
@@ -128,6 +127,13 @@ impl<'lexer, 'input: 'lexer> JITIRParser<'lexer, 'input, '_> {
         self.process_func_types(func_types)?;
         self.process_func_decls(func_decls)?;
 
+        // We try and put trace inputs into registers, but place floating point values on the stack
+        // as yksmp currently doesn't seem able to differentiate general purpose from floating
+        // point.
```
vext01, line 132: If we are indeed using DWARF reg numbers, a floating point reg is in the range `[17, 32]` inclusive, since we use the `xmm` vector registers.

vext01: (that's x86_64 specific though!)

ykrt/src/compile/jitc_yk/jit_ir/parser.rs
```diff
@@ -305,18 +311,38 @@ impl<'lexer, 'input: 'lexer> JITIRParser<'lexer, 'input, '_> {
...
+                                    0,
+                                ));
                             }
-                            ASTType::Float(_) | ASTType::Double(_) => {
-                                self.m
-                                    .push_tiloc(VarLocation::Register(Register::FP(Rx::XMM0)));
+                            Ty::Float(_) => {
+                                let size = u16::try_from(size).unwrap();
+                                ti_off = ti_off.next_multiple_of(u32::from(size));
+                                // FIXME: Why 6?!
```
vext01, line 334: 6 is the base pointer.

ltratt: I'd like this stuff to be fixed in another PR: I've (mostly) just copied this over.

vext01:
I posted one more small comment after you suggested to squash. If  the knowledge that `6` == `rbp` doesn't change anything, please squash.

ltratt:
Squashed.

[I think everything surrounding yksmp registers needs more work, but this PR mostly makes the situation less bad. And it needs a more expert eye than I to fix this stuff. I've asked Lukas to have a look when he's back as he'll probably find this stuff easy to improve.]

vext01:
Thanks. And sorry I didn't have all the answers. I tried my best.

PR #1336, opened 2024-08-02
Move reg_alloc directory to module.

ltratt:
Since it's the sole entry in that directory, renaming the file has no change on any other part the code.


vext01:
I agree. This is a throwback to when we thought we could have different generic register allocators.

PR #1337, opened 2024-08-02
Rename "X86_64" to "x86"

ltratt:
The history of the naming convention is a bit more interesting than I realised. Here's [Wikipedia on this](https://en.wikipedia.org/wiki/X86-64#Industry_naming_conventions):

> Since AMD64 and Intel 64 are substantially similar, many software and hardware products use one vendor-neutral term to indicate their compatibility with both implementations. AMD's original designation for this processor architecture, "x86-64", is still used for this purpose,[[2]](https://en.wikipedia.org/wiki/X86-64#cite_note-apple-x86-64-2) as is the variant "x86_64".[[3]](https://en.wikipedia.org/wiki/X86-64#cite_note-apple-x86_64-arch-manpage-3)[[4]](https://en.wikipedia.org/wiki/X86-64#cite_note-apple-x86_64-email-4) Other companies, such as [Microsoft](https://en.wikipedia.org/wiki/Microsoft)[[6]](https://en.wikipedia.org/wiki/X86-64#cite_note-ms-x64-6) and [Sun Microsystems](https://en.wikipedia.org/wiki/Sun_Microsystems)/[Oracle Corporation](https://en.wikipedia.org/wiki/Oracle_Corporation),[[5]](https://en.wikipedia.org/wiki/X86-64#cite_note-solaris-x64-5) use the contraction "x64" in marketing material.

Since `x64` is shorter and universally understood, I proposed renaming our codegen dir to the shorter version. I'm not wedded to this, so if you think it's not worthwhile, please shout!


vext01:
I'm surprised there are so few!

Issue #1338, opened 2024-08-05
What are the implications of #1318 (removing the trace input struct)?


ltratt:
I've been digging into the x64 codegen and register allocator and I will admit that I have tied myself in knots trying to understand how #1318 has changed the system. That doesn't mean that #1318 isn't fully correct -- I may just not yet understand what it's doing properly! There seem to be 3 things that I can't adequately explain to myself.

First is that I don't understand why we sometimes calculate addresses and not values (in the [current code](https://github.com/ykjit/yk/blob/master/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs#L812) or [the original PR](https://github.com/ykjit/yk/pull/1318/files#diff-150b793f9518b19505892d1496e8641dc1de27061360bfb755cb1c85159f3eaeR778)):

```rust
match self.m.type_(inst.tyidx()) {
  Ty::Integer(_) | Ty::Ptr => {
    let [tgt_reg] = self.ra.get_gp_regs_avoiding(
      &mut self.asm,
      iidx,
      [RegConstraint::Output],
      RegSet::from_vec(&STACKMAP_GP_REGS),
    );
    ...
    match m {
      VarLocation::Direct { frame_off, size } => {
        // FIXME If we prime the register allocator with the interpreter frames
        // stack size and don't create a new frame in the trace epilogue we can
        // reference values directly using the current RBP value.
        match size {
            8 => dynasm!(self.asm
                 ; mov Rq(tgt_reg.code()), QWORD [Rq(Rq::RBP.code())]
                 ; lea Rq(tgt_reg.code()), [Rq(tgt_reg.code()) + frame_off]
            ),
            _ => panic!(),
        }
```

The `lea` towards the bottom of that code sample really surprises me: if we've got an object of `Ty::Integer` (for example), how is it correct to end up with an address to the stack? I would expect code more along the lines of:

```rust
match size {
  8 => dynasm!(self.asm
  ; mov Rq(tgt_reg.code()), QWORD [Rq(Rq::RBP.code())]
  ; mov Rq(tgt_reg.code()), [Rq(tgt_reg.code())]
  ),
```

i.e. loading the value out of the stack.

The second thing is that `VarLocation` has (amongst others) now has these three variants:

```rust
pub(crate) enum VarLocation {
    /// The SSA variable is on the stack.
    Stack {
        /// The offset from the base of the trace's function frame.
        frame_off: u32,
        /// Size in bytes of the allocation.
        size: usize,
    },
[...]

PR #1339, opened 2024-08-05
Refactor spilling

ltratt:
This PR refactors some of the code to do with spilling. I had hoped to make this a stepping stone to making the register allocator take advantage of the changes from #1318 but (as I outlined in #1338) I don't really understand how to do things properly right now. Whatever happens, I think this PR puts the system in a better state for future changes to this aspect.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
@@ -119,7 +119,7 @@ pub(crate) struct LSRegAlloc<'a> {
     inst_vals_alive_until: Vec<InstIdx>,
     /// Where on the stack is an instruction's value spilled? Set to `usize::MAX` if that offset is
     /// currently unknown.
-    spills: Vec<usize>,
+    spills: Vec<SpillState>,
     /// The abstract stack: shared between general purpose and floating point registers.
     stack: AbstractStack,
 }
@@ -156,7 +156,7 @@ impl<'a> LSRegAlloc<'a> {
             fp_regset: RegSet::with_fp_reserved(),
             fp_reg_states,
             inst_vals_alive_until,
-            spills: vec![usize::MAX; m.insts_len()],
+            spills: vec![SpillState::Empty; m.insts_len()],
             stack: Default::default(),
         }
     }
@@ -458,7 +458,7 @@ impl<'a> LSRegAlloc<'a> {
         match self.gp_reg_states[usize::from(reg.code())] {
             RegState::Reserved | RegState::Empty | RegState::FromConst(_) => (),
             RegState::FromInst(iidx) => {
-                if self.spills[usize::from(iidx)] == usize::MAX {
+                if self.spills[usize::from(iidx)] == SpillState::Empty {
                     let inst = self.m.inst_no_proxies(iidx);
                     let size = inst.def_byte_size(self.m);
                     self.stack.align(size); // FIXME
@@ -471,7 +471,7 @@ impl<'a> LSRegAlloc<'a> {
                         8 => dynasm!(asm; mov QWORD [rbp - off], Rq(reg.code())),
                         _ => unreachable!(),
                     }
-                    self.spills[usize::from(iidx)] = frame_off;
+                    self.spills[usize::from(iidx)] = SpillState::Stack(off);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


ltratt:
I am now pretty sure `SpillState::Direct` is wrong and it should be `SpillState::Stack`. Will fix and push a new commit.

ltratt:
a237bd54 fixes that. Will need squashing.

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
```diff
@@ -1169,8 +1169,8 @@ impl From<Rx> for RegSet<Rx> {
 enum SpillState {
     /// This variable has not yet been spilt, or has been spilt and will not be used again.
     Empty,
-    /// This variable is spilt to the stack with the same semantics as [VarLocation::Direct].
-    Direct(i32),
+    /// This variable is spilt to the stack with the same semantics as [VarLocation::Stack].
```
vext01, line 1172: I think we have to be careful here. Stackmaps have two ways of expressing stack storage: `Direct` and `Indirect`.

https://llvm.org/docs/StackMaps.html#stack-map-format:
> Values spilled to the stack are encoded as Indirect locations. The runtime must load those values from a stack address, typically in the form [BP + Offset]. If an alloca value is passed directly to a stack map intrinsic, then LLVM may fold the frame index into the stack map as an optimization to avoid allocating a register or stack slot. These frame indices will be encoded as Direct locations in the form BP + Offset.

Perhaps `Direct` is more precise after all?

ltratt: As things stand right now, it seems `Stack`, `Direct`, and `Indirect` are all different. *Why* exactly that's the case is one of the things I raised in the recent issue. It feels to me like there should only be 2 states, but maybe we really do need 3. I really don't know. What I do know is that this PR currently only handles `VarLocation::Stack` and so `Spillstate::Stack` is the right name (for now...).

vext01: I think you will need to ask Lukas if this is correct when he's back. We can merge it for now.

I don't know what `Stack` is though. That's not mentioned in the stackmap docs.

The valid locations are:
```
Register | Direct | Indirect | Constant | ConstantIndex
```

ltratt: `Stack` is the "traditional" yk notion of "the SSA value is on the trace function's frame stack at [rbp - frame_off]".

ltratt:
@vext01 OK to squash?

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
vext01: Right, and that is expressed in stackmaps as one of `Direct` or `Indirect`, so a third stack stack shouldn't be needed AFAIK.

vext01:
Oops, yes please.

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
ltratt: *Maybe* there is a good reasons for 3. `Stack` is "the yk reg alloc spilled the value" and `Direct` and `Indirect` are "LLVM put this on a stack somewhere". So *maybe* it's just the names that suck. I suspect we shouldn't follow LLVM's terminology here, and we should also rename `Stack` to something else. But this is something that's probably best handled by @ptersilie.

ltratt:
Squashed.

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
ptersilie: Right, I see what the issue is. It's correct that LLVM has 2 ways to describe a value on the stack:
1) Indirect: This is a value that was spilled to the stack, and the pointer is also stored in the stack. So to get the value we need to do an extra indirection: `val = *(rbp - 8)`
2) Direct: This is stackmaps realising the live value is itself a pointer to the stack. So instead of spilling the pointer, stackmaps just records the stack offset: `val = rbp - 8`

However, at the moment the interpreter frame and the compiled-trace frame are separate. And that's why, I believe, Laurie needed to add another `Stack` variant, to describe something that lives on the JIT stack.

This was one of the changes I had planned to do next: to execute the trace on the same stack as the parent frame. Then we don't need that extra variant anymore. Maybe this can be done in this PR, I'm not sure. It "should" be as simple as initialising the register allocator with the size of the main interpreter frame. In other words, we tell the register allocator: here's an existing stack, so the first 10 bytes are already used, anything else you need to push, you push after that. Happy to do that in a follow up PR though.

ptersilie: Oh, just saw this PR is already merged. So ignore my last point. :laughing:

PR #1342, opened 2024-08-05
Enabled `YKD_LOG_IR=jit-asm` in non-debug mode.

ltratt:
For ages I've been slightly frustrated that I can only check jit-asm output if yk is compiled in debug mode. Since we typically tend to tie release builds of interpreters together with a release mode of yk, that means in practise that `YKD_LOG_IR=-:jit-asm` hasn't worked with release build of yklua.

This commit de-`debug_assertions` the code that generates comments when building a trace, which is all we need to produce jit-asm output. This does cause to do a bit more heap allocation at run-time, but it's going to be pretty minor compared to our other costs: even with synchronised compilation on, I can't measure any performance difference with/without this commit.


vext01:
Fair enough.

Can you also update the docs here:
https://ykjit.github.io/yk/dev/understanding_traces.html#ykd_log_ir

ltratt:
Squashed.

PR #1346, opened 2024-08-06
Join worker threads and percolate panics outwards.

ltratt:
Currently if a worker thread (i.e. a compiler thread) panics, an error is printed to the terminal, but the main process charges on as if nothing happened. This makes it very easy for errors to be missed.

This commit makes `MT`, when it is closing down, check whether worker threads have panicked or not.

In one way this works fine, but it's also made it clear to me that relying on `drop` with an `Arc` is not a great solution here: we clone `MT` in various cases, so worker threads can stop `drop` ever being called in certain cases. Fixing that is a job for another day: this commit does work in enough cases already to be useful.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -13,7 +13,7 @@ use std::{
         atomic::{AtomicU16, AtomicU32, AtomicU64, AtomicUsize, Ordering},
         Arc,
     },
-    thread,
+    thread::{self, JoinHandle},
 };
 
 use parking_lot::{Condvar, Mutex, MutexGuard};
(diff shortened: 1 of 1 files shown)


ykrt/src/mt.rs
```diff
@@ -71,9 +70,11 @@ pub struct MT {
...
     /// The hard cap on the number of worker threads.
     max_worker_threads: AtomicUsize,
-    /// How many worker threads are currently running. Note that this may temporarily be `>`
-    /// [`max_worker_threads`].
-    active_worker_threads: AtomicUsize,
+    /// [JoinHandle]s to each worker thread so that when an [MT] value is dropped, we can try
+    /// joining each worker thread and see if it caused an error or not. If it did,  we can
+    /// percolate the error upwards, making it more likely that the main thread exits with an
+    /// error. In other words, this [Vec] makes it harder for errors to be missed.
+    active_worker_threads: Mutex<Vec<JoinHandle<()>>>,
```
vext01, line 77: Looks good, but one of the concerns I mentioned in the issue was the size of this vector being unbounded and consuming a lot of memory. Is it a real concern, do you think?

ltratt: No: it's bounded by `max_worker_threads`.

vext01: Jolly good. Thanks.

PR #1352, opened 2024-08-08
Explicit shutdown

ltratt:
This PR moves `MT` to an "explicit shutdown" model which solves the "`Arc` hasn't decremented to zero because it's being used in a thread / guard /etc" problem.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -11,16 +11,17 @@ TRACERS="hwt"
 #  - `yk-config` must be in PATH.
 #  - YK_BUILD_TYPE must be set.
 test_yklua() {
-    if [ ! -e "yklua" ]; then
-        git clone https://github.com/ykjit/yklua
-    fi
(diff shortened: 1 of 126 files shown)


ltratt:
Hmm, so now we have a bootstrapping problem: we can't merge yklua until it uses `yk_mt_shutdown` *but* it can't do that until this PR is merged.

My inclination is to briefly comment out yklua in this repo, merge this, go back to yklua, update it to use `yk_mt_shutdown`, then raise a PR to reenable yklua in this repo. Sound like a plan?

vext01:
Works for me.

ltratt:
Fixed in 489323e5.

vext01:
Happy to leave that commit separate?

vext01:
Left it separate, so you can just do `git revert`

ltratt:
Good idea.

PR #1356, opened 2024-08-09
Migrate most of the remaining tests.

vext01:
The only tests still suffixed `.old.c` are side-tracing tests, which I haven't even looked at, since side-tracing is broken for now.

This is probably unreviewable, sorry.


ltratt:
Can we change `YK_LOG=255` to `YK_LOG=5` please? If we use the former, we'll encourage everyone to use that, even though its behaviour isn't defined. I should probably make the code reject too-big values...

vext01:
The problem is I didn't know what value to use. The docs are vague:

> <level> specifies the level of logging: level 0 turns off all yk logging; level 1 shows major errors only; level 2 warnings; and level 3 and above being increasingly granular information about transitions within yk. Some information may or may not be displayed based on compile-time options. Defaults to 1.

255 is valid according to this.

256 is also valid according to the docs, but will crash the system.

If 5 is the max, we should document that and probably what all the other undocumented numbers do too?

vext01:
> I should probably make the code reject too-big values...

Agreed.

Since we've already let some `YK_LOG=255` in in my last PR, I suggest we merge this as-is, and a follow up PR can reject too large values, document the values and replace all `=255` with `=5`.

ltratt:
Actually 4 is the maximum value! I will raise a PR to document and enforce that, but let's do `YK_LOG=4` in this PR in the meantime.

vext01:
> but let's do YK_LOG=4 in this PR in the meantime.

I'm going to need a follow up PR to replace the ones already in-tree, so may as well do them all in one go? I can do that.

ltratt:
OK I will do them in my PR.

vext01:
Cheers

PR #1360, opened 2024-08-10
Enable aot_debuginfo test for both swt and hwt.

Pavel-Durov:
Add extra `...` after `aot_debuginfo` to match SWT tracing function. This works for SWT when the function is present and HWT when there are no IR tracing functions.

### DIFF
--- a/tests/c/aot_debuginfo.c
+++ b/tests/c/aot_debuginfo.c
@@ -1,10 +1,10 @@
-// ignore-if: test $YK_CARGO_PROFILE != "debug" -o "$YKB_TRACER" = "swt"
 // Run-time:
 //   env-var: YKD_LOG_IR=-:aot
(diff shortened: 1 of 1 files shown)


ltratt:
My guess is that:

```
ignore-if: test $YK_CARGO_PROFILE != "debug"
```

was there for a reason -- notice that the failure was in `release` mode.

Pavel-Durov:
> My guess is that:
> 
> ```
> ignore-if: test $YK_CARGO_PROFILE != "debug"
> ```
> 
> was there for a reason -- notice that the failure was in `release` mode.

That's probably it, I removed it by mistake.

Pavel-Durov:
Closing in favour of https://github.com/ykjit/yk/pull/1362 
I noticed that I opened this PR directly and not through my fork.

PR #1362, opened 2024-08-11
 Enable aot_debuginfo test for both swt and hwt.

Pavel-Durov:
BoldAdd extra `...` after `aot_debuginfo` to match SWT tracing function. This works for SWT when the function is present and HWT when there are no IR tracing functions.


ltratt:
Please squash.

Pavel-Durov:
> Please squash.

done :point_right: 979854e3273ad6ec3009987c010190423833f472

PR #1364, opened 2024-08-12
Tidy up guard failure

ltratt:
This PR (is probably the first to) tidy/tidies up how guard failure. The most important commits are https://github.com/ykjit/yk/commit/4e1a3de19dd03fbad5e58d0b87df7bc41bd754c7, which solves a long-standing problem where we kept recording side-traces until one of them compiled and https://github.com/ykjit/yk/commit/7a91ce12034c1963e3e190848e01058a9d71a78b, which makes guard failure a bit more like location transitions.

If this all looks good, we probably need to do some code shuffling in future PRs, before solving a not-really-necessary performance overhead introduced in 4e1a3d, where we introduce a `Mutex`. We can probably do some "atomic tricks" similar to `Location`, but one thing at a time!


ltratt:
Merge as-is is fine!

PR #1365, opened 2024-08-12
Minor cosmetic tidy-ups.

ltratt:
I happened to notice these while studying the file: I might as well change them while I remember them.


ltratt:
Force pushed a two-line change (adding backquotes) to stop rustdoc complaints. Should hopefully be mergable now.

PR #1370, opened 2024-08-13
Unbreak gdb_c_test.

ltratt:
I removed this line:

```
gdb.arg(&binpath).env("YKD_TRACE_DEBUGINFO", "1");
```

in 811ca274 when I should only have removed the `.env(...)` part. Doing so caused gdb_c_test to compile but not to work. This commit restores the `.arg(...)` part, and allows gdb_c_test to work again.


vext01:
My bad too for not catching this in review!

PR #1374, opened 2024-08-14
Promote more than just `usize`

ltratt:
This PR alters yk so that we can promote things that aren't `usize`d. As an example (and because I need it for yklua) I've added support for promoting `i32`s (https://github.com/ykjit/yk/commit/916484c229d5d2f526120e323d63b8f0991e22eb).

### DIFF
--- a/ykrt/src/compile/jitc_yk/trace_builder.rs
+++ b/ykrt/src/compile/jitc_yk/trace_builder.rs
@@ -54,23 +54,23 @@ pub(crate) struct TraceBuilder {
     jit_mod: jit_ir::Module,
     /// Maps an AOT instruction to a jit instruction via their index-based IDs.
     local_map: HashMap<aot_ir::InstID, jit_ir::Operand>,
-    // BBlock containing the current control point (i.e. the control point that started this trace).
+    /// BBlock containing the current control point (i.e. the control point that started this trace).
     cp_block: Option<aot_ir::BBlockId>,
-    // Index of the first traceinput instruction.
+    /// Index of the first traceinput instruction.
     first_ti_idx: usize,
-    // Inlined calls.
+    /// Inlined calls.
     frames: Vec<Frame>,
-    // The block at which to stop outlining.
+    /// The block at which to stop outlining.
     outline_target_blk: Option<BBlockId>,
-    // Current count of recursive calls to the function in which outlining was started. Will be 0
-    // if `outline_target_blk` is None.
+    /// Current count of recursive calls to the function in which outlining was started. Will be 0
+    /// if `outline_target_blk` is None.
     recursion_count: usize,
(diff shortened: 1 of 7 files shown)


ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -1025,13 +1027,26 @@ impl TraceBuilder {
...
-        let pval = self.promotions[self.promote_idx];
-        self.promote_idx += 1;
-        let c = Const::Int(
-            self.handle_type(self.aot_mod.type_(*tyidx))?,
-            u64::try_from(pval).unwrap(),
-        );
+        let pval = match self.jit_mod.type_(ty) {
+            jit_ir::Ty::Void => unreachable!(),
+            jit_ir::Ty::Integer(width_bits) => {
+                let width_bytes = usize::try_from(*width_bits).unwrap() / 8;
```
vext01, line 1035: Do I see it right that the unwrap is safe for systems with a pointer size >=32 bits?

We are unlikely to support anything else, right?

ltratt: I'm not sure these ones are useful to document TBH in the sense that they're kinda obvious. I know we used to say "document all `unwraps`" but this one is really "super weird portability edge cases" where getting things wrong is unlikely; the ones around `from_ne_bytes` are slightly irritating in that they can't actually fail but we're forced to add them anyway.

vext01: OK. I've been operating in "document all unwraps" mode, but thinking about it, it doesn't really matter. They will crash and we can investigate.

ltratt: I think in the general case it's a useful discipline because it forces us to think about whether we can actually encounter a panic. But I agree that a blanket commenting rule isn't quite right: I think we should document the ones where we have internal-yk invariants which mean we can't encounter a panic. For example a classic one is "we checked X elsewhere so we know it will fit in this type and the unwrap can't fail" -- that's really worth documenting because no future reader will know otherwise! But, say, i32 -> i64 type conversions are really a not-very-interesting artefact of Rust's integer type hierarchy.

ykrt/src/promote.rs
```diff
@@ -1,13 +1,24 @@
...
-/// Promote a value.
-///
-/// When encountered during trace recording, the returned value will be considered constant in the
-/// resulting compiled trace.
-///
-/// The user sees this as `yk_promote` via a macro.
+/// Promote a `usize` during trace recording.
+#[no_mangle]
+pub extern "C" fn __yk_promote_c_int(val: c_int) -> c_int {
+    println!("promote_c");
```
jacob-hughes, line 12: Debugging code left in?

ltratt: Oops....

PR #1379, opened 2024-08-19
Promote mustn't directly propagate constants.

ltratt:
Promotion as currently implemented both inserts a guard and propagates constants. Here's an example of a trace resulting from the `promote.c` test:

```
%15: i1 = eq %13, 100i64
guard true, %15, [%0, %1, %2, %3, %4, %5, %13]
*%3 = 100i64
%18: i64 = load %4
%19: i64 = add %18, 100i64
```

Let's look at just the first three instructions (`eq`, `guard`, and `*=`) in that trace:

```
%15: i1 = eq %13, 100i64
guard true, %15, [%0, %1, %2, %3, %4, %5, %13]
*%3 = 100i64
```

The problem is that the assignment in the third instruction is subtly incorrect -- at least in the JIT IR semantics that I believe we have, though we've never formally defined this! -- because it is decoupled from the guard.

I claim that this three line sequence is indistinguishable from just `*3 = 100i64`. In other words, a reasonable dead-code analysis would "realise" that the `eq` and `*=` are pointless, because the constant in the assign doesn't rely on them.

Another way of looking at this is that the trace fragment above could be created by either this pseudo-code:

```
if (x == 100) *y = 100;
```

or this pseudo-code:

```
x = 100;
...
*y = yk_promote(x);
```

Those two are semantically different, so conflating them into one in the trace is bound to cause optimisation problems at some point.

The reason for this subtle flaw is that `handle_promote` emits a `ProxyConst` when it should emit a `ProxyInst`. This commit changes that, so that the three instructions above become (written in long-hand without deproxying):

```
%15: i1 = eq %13, 100i64
guard true, %15, [%0, %1, %2, %3, %4, %5, %13]
*%3 = %15
```

That might look disappointing: have we actually promoted anything? We have! The `eq` and `guard` are inserted by `handle_promote`, but because we're using `ProxyInst` the `*=` now clearly relies on the `guard` if we want to propagate the constant. In other words, a dead-code analysis would realise that any optimisations that rely on propagating the constant also clearly rely on the guard for correctness.

Currently we don't have the optimisation passes which propagate the result of a guard (I have a now fairly old version in a branch, which I probably have to redo from scratch), but when we do, we will be able to use the information from the guard to to turn `*%3 = %15` into machine code that's really just `*%3 = 100`, but without deleting the guard necessary for correctness.


ptersilie:
This sounds reasonable to me.

PR #1381, opened 2024-08-19
Use labels to keep track of live variables at loop edges.

ptersilie:
When looping back to the top of a trace we need make sure that live
values are put back into the place where the beginning of the trace
expects them. This is because the register allocator may have moved them
somewhere else, e.g. via spilling.

The way we did this until now was by remembering the live values and
their locations at the beginning of the trace, and then restoring them
before jumping at the end of the trace. This isn't entirely correct, but
works as long as all live values live on the stack, because changes to a
value are instantly stored back on the stack.

Once we start using mem2reg, which moves live variables from the stack
into registers, this will no longer work. Changes to variables inside
the loop will no longer be stored on the stack, but instead use PHI
nodes. For example, a loop iterator may look like this:

```
%0: i32: load_ti Register(3, 8, 0, 0)
tloop_start:
%2 = phi [%0, bb0], [%3, bb1]
%3 = add %2, 1i32
jmp tloop_start
```

The increment to the iterator `%0` is stored in a new variable `%3`, and
then the the phi node `%2` picks up the new value on each iteration. If
we only remember the live variables at the beginning of the trace, and
restore them at the end, we would be restoring `%0`, which restores the
old value and causes an inifinte loop.

To fix this we need to track the live variables at the beginning **and**
the end of the trace using labels. Then before we do the backedge jump
we compare them. For any variable where the two locations don't match,
we need to copy the value from the end location to the start location.
For example, if we look at the IR for `simple.c` we now get this:

```
%0: ptr = load_ti Register(3)
%1: ptr = load_ti Direct(6, -40, 8)
%2: ptr = load_ti Direct(6, -32, 8)
%3: ptr = load_ti Register(5)
%4: ptr = load_ti Register(15)
tloop_start: [%0, %1, %2, %3, %4]
...
tloop_jump: [%0, %1, %2, %3, %17]
```

We can already see that the due to SSA the value stored in `%4` is now
stored in `%17`. If we check their locations we might get something
like:

```
%4 => Register(15)
%17 => Register(12)
```

This means we now need to copy the value from R12 into R15, which
ensures that if `%17`'s value is different, this change is propagated to
the beginning of the loop. (Note that since we don't have enabled
mem2reg yet, these changes don't make any difference now, since both
`%4` and `%17` are stack pointers and are already equivalent).


ptersilie:
I tried to measure performance with yklua but couldn't see any significant difference.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -137,6 +137,7 @@ struct Assemble<'a> {
     ra: LSRegAlloc<'a>,
     tloop_gp_reg_states: Option<[lsregalloc::RegState; 16]>,
```
ltratt, line 138: This field (and the one below it) is presumably unused now?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -137,6 +137,7 @@ struct Assemble<'a> {
...
     tloop_fp_reg_states: Option<[lsregalloc::RegState; 16]>,
+    loop_start_locs: Vec<VarLocation>,
```
ltratt, line 140: Let's add a doc comment to fields by default unless they're blindingly obvious.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1173,10 +1144,120 @@ impl<'a> Assemble<'a> {
         }
     }
 
+    fn cg_traceloopjump(&mut self) {
+        // Loop the JITted code if the `tloop_start` label is present.
```
ltratt, line 1148: Why wouldn't it be present?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1173,10 +1144,120 @@ impl<'a> Assemble<'a> {
...
+            }
+            Err(e) => {
+                // Any other error suggests something has gone quite wrong. Just crash.
+                panic!("{}", e.to_string())
+            }
+        }
+    }
+
     fn cg_traceloopstart(&mut self) {
+        assert_eq!(self.loop_start_locs.len(), 0);
```
ltratt, line 1250: `debug_assert_eq`

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -133,6 +133,9 @@ pub(crate) struct Module {
     guard_info: Vec<GuardInfo>,
     /// Indirect calls.
     indirect_calls: Vec<IndirectCallInst>,
+    // FIXME: no pub
+    pub loop_start_vars: Vec<Operand>,
```
ltratt, line 137: Let's fix this now, otherwise it'll be forgotten.

ltratt:
This mostly looks good: my comments are relatively minor.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: This comment was already there, I just copied it. But an example for a trace without a `tloop_start` label are side-traces.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ptersilie: [515e055](https://github.com/ykjit/yk/pull/1381/commits/515e055862595b4deeafccedb97181ac9f7ce829)

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: [d2d096b](https://github.com/ykjit/yk/pull/1381/commits/d2d096bce7557909684224b057652be9c3e84f70)

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: [7170a3a](https://github.com/ykjit/yk/pull/1381/commits/7170a3a8cdbbdc740e24647e4f5194bc0ff29c04)

ptersilie:
This should address all comments.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -526,6 +527,22 @@ impl Module {
     ) -> Result<GuardInfoIdx, CompilationError> {
         GuardInfoIdx::try_from(self.guard_info.len()).inspect(|_| self.guard_info.push(info))
     }
+
+    pub(crate) fn loop_start_vars(&self) -> &Vec<Operand> {
```
ltratt, line 531: We shouldn't return `&Vec<T>` except in very unusual circumstances: `&[T]` is the standard Rust idiom.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ltratt: I assume some methods that used these fields are now unused and are dead code?

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ptersilie: [9fa086d](https://github.com/ykjit/yk/pull/1381/commits/9fa086d73cfef632c2105203f5d7017b9c6eb7a7)

ltratt:
> I tried to measure performance with yklua but couldn't see any significant difference.

I forgot to say: that makes sense.

ltratt:
Please squash.

ptersilie:
Squashed.

PR #1382, opened 2024-08-22
Fix bug in trace loop jumping.

ptersilie:
Since `frame_off` is a negative value we need to add it to the base pointer. Substracting it writes values into the wrong frame leading to segfaults/UB.


vext01:
Nice find @ptersilie

PR #1384, opened 2024-08-22
Remove `inst_all` and clarify deproxying/deconsting.

ltratt:
There's more to be done in this regard, but this is a start.


vext01:
The first commit could have a functional change, right? Were those uses of `inst_all` incorrect?

ltratt:
> Were those uses of inst_all incorrect?

"Incorrect" isn't the right word here: until the recent introduction of `inst_deproxy` there was no alternative to `inst_all`. What I would say is that they didn't need the (dangerous) "power" of `inst_all`.

PR #1388, opened 2024-08-28
Enable mem2reg

vext01:
This is a draft because a) the commit structure is still preliminary, and b) the general approach is experimental, and c) I need to test mem2reg on more things (yklua).

At this point I'm most interested in knowing if https://github.com/ykjit/yk/commit/1b34d23c158db184c2b37a6e634ec20474983fef is tolerable.

Thoughts?


vext01:
side note: this removes the ability to specify a custom pre/post-link pipeline as it was clashing with what I was doing. We can re-add it later if need be.

bin/yk-config
```diff
@@ -10,22 +10,59 @@ if [ ! -d "${DIR}/../ykrt" ]; then
...
+set_aot_pipeline 0
 
 OUTPUT=""
 
 usage() {
     echo "Generate C compiler flags for building against the yk JIT.\n"
     echo "Usage:"
-    echo "    yk-config <profile> [--prelink-pipeline <passes>] [--postlink-pipeline <passes>] <--cc|--cxx|--ar|--ranlib|--cppflags|--cflags|--ldflags>\n"
-    echo "    Where <profile> is a Rust cargo profile starting with either 'debug' or 'release'."
+    echo "    yk-config <profile> [--ao <0|1>] \\"
```
vext01, line 56: I wonder if there's a better way to do this usage message formatting...

ltratt:
While I look at https://github.com/ykjit/yk/commit/1b34d23c158db184c2b37a6e634ec20474983fef, could we get some very quick performance numbers in here?

bin/yk-config
```diff
@@ -10,22 +10,59 @@ if [ ! -d "${DIR}/../ykrt" ]; then
...
 
-# The post-link time optimisations that we can tolerate. At this moment this is
-# only the instcombine pass in the below configuration (copied over from the
-# default pipeline). Using `instcombine` on its own breaks, possibly due to the
-# missing `no-verify-fixpoint`.
-POSTLINK_PASSES_STR="instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>"
+set_aot_pipeline() {
+    level=$1
+    # env var always takes precident.
+    if [ ! -z "${YKB_AOT_OPTLEVEL}" ]; then
```
vext01, line 16: I also wasn't 100% certain that this should be a `YKB`. Yes, it's for building stuff, but for building the **interpreter**, as opposed to yk components.

vext01:
> While I look at https://github.com/ykjit/yk/commit/1b34d23c158db184c2b37a6e634ec20474983fef, could we get some very quick performance numbers in here?

We need to check that the benchmarks (and probably the yklua tests too) even work with mem2reg first. I'll start on that today.

ltratt:
I admit that I'm not a big fan of https://github.com/ykjit/yk/commit/1b34d23c158db184c2b37a6e634ec20474983fef *but* it's tolerable for this PR, and I'd rather focus on mem2reg than fixing it right now.

Basically I think I'd prefer something where `yk-config` takes separate explicit flags for the Rust/yk and C/interpreter sides of things. And, probably, that would be handled by a wrapper script which takes in lang_test's actual envs, rather than the `##` hack. But, like I said, this is as much about aesthetics and long-term maintainability: what we have in this PR is good enough, and we don't want to distract ourselves.

vext01:
Yes, I know it isn't great. I considered something like that, but it would be a large change and probably a bit of a rabbit hole too. I just wanted something relatively small/quick to get us going.

It sounds like you are OK with this for now, so I'll carry on testing yklua with it.

vext01:
> yk-config takes separate explicit flags for the Rust/yk

^ Didn't understand this part though. `yk-config` isn't responsible for configuring the building of any rust code to my knowledge.

ltratt:
> It sounds like you are OK with this for now

I am!

> yk-config isn't responsible for configuring the building of any rust code to my knowledge.

It's responsible for picking up a specific build of yk.

vext01:
I'm going to close this draft and re-raise a new more reviewable PR.

PR #1391, opened 2024-09-02
Fix deproxying.

ltratt:
This was subtly broken (my fault!) before in two ways: `map_packed_operand_local`s is necessary for (our current) dead code elimination, but we also (incorrectly) used it elsewhere; and we didn't use the updated `iidx`s after calling `inst_deproxy`.

This commit fixes both aspects. However, I'm not sure this is the best long-term approach, and that perhaps the better thing to do is to continually update (i.e. deproxy) operands as we go along. That can wait for a later day once I've chewed this over sufficiently, however.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
@@ -125,7 +125,7 @@ pub(crate) struct LSRegAlloc<'a> {
 }
 
 impl<'a> LSRegAlloc<'a> {
-    pub(crate) fn new(m: &'a Module, inst_vals_alive_until: Vec<InstIdx>) -> Self {
+    pub(crate) fn new(m: &'a Module) -> Self {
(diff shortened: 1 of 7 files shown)


vext01:
I'm OK with this.

I take it it's hard to write a test for?

ltratt:
There is a new test that covers the first commit, though I haven't yet worked out one for the second commit. I *might* have an idea for that now, though.

vext01:
OK, I'll wait a moment and see if you come up with anything

ltratt:
OK I added a test in https://github.com/ykjit/yk/pull/1391/commits/0a11641815bcbb0d9b3f3ee224d94aabfe3d625e which not only covers the second commit but uncovered another (subtle) case I hadn't handled correctly.

vext01:
I can't vouch for most of the logic in here, but since it comes with tests, I'd be inclined to put it in. Do you agree?

vext01:
Please squash

ltratt:
Squashed.

PR #1392, opened 2024-09-03
Support functions that return floats/doubles.

ltratt:


ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
```diff
@@ -706,12 +706,11 @@ impl<'a> LSRegAlloc<'a> {
         for cnstr in &constraints {
             match cnstr {
                 RegConstraint::InputIntoReg(_, reg)
-                | RegConstraint::InputIntoRegAndClobber(_, reg) => avoid.set(*reg),
+                | RegConstraint::InputIntoRegAndClobber(_, reg)
```
vext01, line 709: The only part of the PR I'm uncertain of is this part. So the actions for the constraints were just wrong altogether before and this part isn't much to do with floats?

vext01:
Please squash.

ltratt:
This one has only 1 commit, so I can't squash it further :)

PR #1394, opened 2024-09-10
Enable mem2reg in the AOT pipeline.

ptersilie:
mem2reg replaces allocas with SSA variables so that they can be allocated to registers where possible:

https://llvm.org/docs/Passes.html#mem2reg-promote-memory-to-register

This commit contains about a week's worth of debugging work and now the yklua test suite runs to completion.

To make testing possible we also needed to add the ability for tests to specify an AOT opt level. This doesn't fit well with the current test harness, so it's a bit of a hack.

The test file needs a line at the top like:
// ## yk-config-env: YKB_AOT_OPTLEVEL=1

### DIFF
--- a/bin/yk-config
+++ b/bin/yk-config
@@ -10,21 +10,58 @@ if [ ! -d "${DIR}/../ykrt" ]; then
     exit 1
 fi
 
-# The post-link time optimisations that we can tolerate. At this moment this is
-# only the instcombine pass in the below configuration (copied over from the
-# default pipeline). Using `instcombine` on its own breaks, possibly due to the
-# missing `no-verify-fixpoint`.
-POSTLINK_PASSES_STR="instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>"
+set_aot_pipeline() {
+    level=$1
+    # env var always takes precident.
+    if [ ! -z "${YKB_AOT_OPTLEVEL}" ]; then
+        level=${YKB_AOT_OPTLEVEL}
+    fi
+    case ${level} in
+        0)
+            POSTLINK_PASSES=${POSTLINK_PASSES_AO0}
+            PRELINK_PASSES=${PRELINK_PASSES_AO0}
+            ;;
+        1)
+            POSTLINK_PASSES=${POSTLINK_PASSES_AO1}
+            PRELINK_PASSES=${PRELINK_PASSES_AO1}
+            ;;
+        *)
+            echo "error: invalid yk AOT optimisation level '${level}'" >&2
+            exit 1
+            ;;
+    esac
+}
+
+# Canned pipelines.
+#
+# Ideally we'd detect the optimisation level from the CFLAGS env var, but it's
+# not easy to reliably parse that info out of CFLAGS. So for now we allow `--ao
+# <n>` to yk-config (or set YKB_AOT_OPTLEVEL in the env) to specify the AOT
+# optimisation level.
+#
+# Note that clang is still passed -O0 regardless. We just add our own pipeline
+# specification on top.
+PRELINK_PASSES_AO0=""
+POSTLINK_PASSES_AO0="instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>"
+PRELINK_PASSES_AO1=""
+POSTLINK_PASSES_AO1="instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>,mem2reg,instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>"
+# Initialise the AOT pipeline to level 0 unless YKB_AOT_OPTLEVEL is set.
+set_aot_pipeline 0
 
 OUTPUT=""
 
 usage() {
     echo "Generate C compiler flags for building against the yk JIT.\n"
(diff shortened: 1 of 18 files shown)


ptersilie:
Needs to be rebase after https://github.com/ykjit/ykllvm/pull/198 is merged.

tests/benches/collect_and_decode.rs
```diff
@@ -27,7 +27,7 @@ fn compile_runner(tempdir: &TempDir) -> PathBuf {
     exe.push(tempdir);
     exe.push(src.file_stem().unwrap());
 
-    let mut compiler = mk_compiler(&ykllvm_bin("clang"), &exe, &src, &[], false);
+    let mut compiler = mk_compiler(&ykllvm_bin("clang"), &exe, &src, &[], false, None);
```
ltratt, line 30: I might be wrong but this might need `cargo fmt`.

ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
```diff
@@ -14,6 +14,17 @@ use yksmp::Location as SMLocation;
...
+/// Registers (in DWARF notation) that we want to restore during deopt. Excludes `rsp` (7) and
+/// `return register` (16), which we do not care about.
+const RECOVER_REG: [usize; 31] = [
+    0, 1, 2, 3, 4, 5, 6, 8, 9, 10, 11, 12, 13, 14, 15, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27,
+    28, 29, 30, 31, 32,
+];
+
+/// Number of normal + float registers including the ones missing above. Used to initialise
+/// the register array storing live register values during deopt.
+const REGISTER_NUM: usize = RECOVER_REG.len() + 2;
```
ltratt, line 26: Remind me why this ends up being 33?

tests/langtest_c.rs
```diff
@@ -11,6 +14,30 @@ use tests::{mk_compiler, EXTRA_LINK};
...
+    for line in rdr.lines() {
+        let line = line?;
+        if !line.starts_with(COMMENT) {
+            break; // won't find any lower down the file.
+        }
+        if line.starts_with(&env_prefix) {
+            let sfx = &line[(&env_prefix).len()..];
+            let mut elems = sfx.split("=");
+            extra.insert(
+                elems.next().ok_or("bad env spec")?.trim().into(),
```
ltratt, line 34: `bad env spec` is a bit obtuse: I'd prefer either an `unwrap` (so someone will look at this line of code and then suss out the fix) or a descriptive error message.

bin/yk-config
```diff
@@ -10,22 +10,59 @@ if [ ! -d "${DIR}/../ykrt" ]; then
...
+        0)
+            POSTLINK_PASSES=${POSTLINK_PASSES_AO0}
+            PRELINK_PASSES=${PRELINK_PASSES_AO0}
+            ;;
+        1)
+            POSTLINK_PASSES=${POSTLINK_PASSES_AO1}
+            PRELINK_PASSES=${PRELINK_PASSES_AO1}
+            ;;
+        *)
+            echo "error: invalid yk aot opt level '${level}'" >&2
```
ltratt, line 29: s/aot/AOT/ and s/opt/optimisation/ ?

ltratt:
Those changes (if you agree with them) can go in a force push / rebase.

tests/benches/collect_and_decode.rs
vext01: Just tried locally. Seems OK. Line is 87 chars long.

ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
vext01: 33 (dwarf) registers covers all of the GP and XMM registers.

image

The +2 is because `RECOVER_REG` omits two registers that we shouldn't interfere with during deopt (but we still wish to use the numbers as array indices later).

Hard to explain. Hope it makes sense?

ltratt:
Do we have a rough sense of the performance change from this?

ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
ltratt: Let's at least briefly document that: it's weird and unexpected.

vext01: I think that's what the docstring says?

ltratt: No, it doesn't explain why it's 33 (which is a very weird number in this context!), and it doesn't mention DWARF or RECOVER_REG or whatnot.

ltratt:
Just a rough sense from 1 or 2 benchmarks is enough: we don't need to go overboard, but performance-related PRs should come with a vague sense of how they affect things.

vext01:
I did some quick runs of Richards and Deltablue. We get about a 5-6% speedup.

We appear to be about 3-5% slower than regular Lua.

(all very rough measurements)

vext01:
(I force pushed the fixes and ykllvm sync, as requested btw)

PR #1397, opened 2024-09-13
Fix stack overflow when asking for an AOT operand's type.

vext01:
Requires https://github.com/ykjit/ykllvm/pull/200

When adding more passes to the AOT pipeline we found that the mere act of printing the AOT module can cause stack overflows.

If you have IR like:

```
bb0:
    %1 = ...
bb1:
    %2 = phi bb1 -> %3, bb0 -> %1
    %3 = %2 + 1
    goto bb1
```

Asking the type of %3 causes infinite recursion.

This is because an operation often asks its operands for their type to determine their own type.

Here, %3 asks %2, who asks %3, who asks %2, who asks... forever. Each "ask" is a function call, so this blows the stack.

This change breaks the cycle by putting a PHI's type into the instruction explicitly, so it doesn't have to search further afield to know it's type.

We haven't seen this before because usually the first operand to a PHI is from earlier (not later) in the module, so there is no cycle.

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -917,6 +917,7 @@ pub(crate) enum Inst {
     },
     #[deku(id = "14")]
     Phi {
+        tyidx: TyIdx,
         #[deku(temp)]
         num_incoming: usize,
         #[deku(count = "num_incoming")]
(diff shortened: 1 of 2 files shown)


vext01:
I've just noticed a small typo in the commit message. If we could pause before merging so I can fix it.

ltratt:
Please force push a fix.

vext01:
Fixed.

PR #1398, opened 2024-09-18
Remove "proxy" naming

ltratt:
The use of "Proxy" in instructions has become gradually more unfortunate. First, `ProxyConst` is *very* different to `ProxyInst`, so suggesting there's a link is misleading. Second, "proxy" isn't the standard SSA terminology. This PR fixes both things.


vext01:
It's unfortunate that SSA calls it a copy, because it isn't a copy. Or have I misunderstood?

ltratt:
`Copy` is the standard terminology.

PR #1399, opened 2024-09-19
Adjust to ykllvm now tracking arbitrary numbers of locations.

ptersilie:
We are now tracking arbitrarily many locations for each variable in ykllvm. This requires some changes in the stackmap parser and deopt to deal with this.

Depends on: https://github.com/ykjit/ykllvm/pull/204

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
@@ -227,33 +227,30 @@ pub(crate) extern "C" fn __yk_deopt(
                 todo!("Deal with multi register locations");
             };
             match aotloc {
-                SMLocation::Register(reg, size, off, extra) => {
+                SMLocation::Register(reg, size, off, extras) => {
                     registers[usize::from(*reg)] = jitval;
-                    if *extra != 0 {
-                        // The stackmap has recorded an additional register we need to write
-                        // this value to.
-                        registers[usize::from(*extra - 1)] = jitval;
-                    }
-                    // Check if there's an additional spill location for this value. Negative
-                    // values indicate stack offsets, positive values are registers. Lastly, 0
-                    // indicates that there's no additional location. Note, that this means
-                    // that in order to encode register locations (where RAX = 0), all register
-                    // values have been offset by 1.
-                    if *off < 0 {
-                        let temp = if i == 0 {
-                            unsafe { frameaddr.offset(isize::try_from(*off).unwrap()) }
-                        } else {
-                            unsafe { rbp.offset(isize::try_from(*off).unwrap()) }
-                        };
-                        debug_assert!(*off < i32::try_from(rec.size).unwrap());
-                        match size {
-                            // FIXME: Check that 16-byte writes are for float registers only.
-                            16 | 8 => unsafe { ptr::write::<u64>(temp as *mut u64, jitval) },
-                            4 => unsafe { ptr::write::<u32>(temp as *mut u32, jitval as u32) },
-                            _ => todo!("{}", size),
+                    for extra in extras {
+                        // Write any additional locations that were tracked for this variable.
+                        // Numbers greater or equal to zero are registers in Dwarf notation.
+                        // Negative numbers are offsets relative to RBP.
+                        if *extra >= 0 {
+                            registers[usize::try_from(*extra).unwrap()] = jitval;
+                        } else if *extra < 0 {
+                            let temp = if i == 0 {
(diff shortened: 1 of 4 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
```diff
@@ -227,33 +227,29 @@ pub(crate) extern "C" fn __yk_deopt(
...
-                        match size {
-                            // FIXME: Check that 16-byte writes are for float registers only.
-                            16 | 8 => unsafe { ptr::write::<u64>(temp as *mut u64, jitval) },
-                            4 => unsafe { ptr::write::<u32>(temp as *mut u32, jitval as u32) },
-                            _ => todo!("{}", size),
+                    for extra in extras {
+                        debug_assert_ne!(*extra, 0);
+                        // Write any additional locations that were tracked for this variable.
+                        if *extra > 0 {
+                            registers[usize::try_from(*extra).unwrap()] = jitval;
```
vext01, line 236: How comes the ` - 1` went away again? We still use zero as a special value, don't we?

ptersilie: But that means this need to be `>= 0`. Fixing.

vext01: Ah, OK. That's a good thing. I can cross that off my list.

ptersilie: Fixed in [06b9746](https://github.com/ykjit/yk/pull/1399/commits/06b97469b13269224be4b6a8ac8f0e1273a7b146). The fact that things were still working must mean we no lives are in `rax`. Which would make sense since it's used as a return and thus can't be used for tracking lives.

ltratt: "we lives are in rax"?

ptersilie: Although, on second thought. That's only try for calls. But we have also stackmaps at branches were rax is perfectly fine for storing live values.

yksmp/src/lib.rs
```diff
@@ -247,12 +247,16 @@ impl StackMapParser<'_> {
             self.read_u8();
             let size = self.read_u16();
             let dwreg = self.read_u16();
-            let extrareg = self.read_u16();
+            self.read_u16();
+            let mut extras = Vec::new();
```
ltratt, line 251: I think you can do `let extras = 0..self.read_u16().map(|_| self.read_i16()).collect::<Vec<_>>()` or similar if you'd prefer.

ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
ltratt: On the basis that I'm confused, I wonder if this needs a comment?

ltratt:
I've changed the PR title: at squashing time, it might be worth changing the commit(s) titles to something similar.

yksmp/src/lib.rs
ptersilie: Personally I find that harder to understand on a glance.

ltratt:
I'm fine with another title, but the first one didn't parse (and I then parsed it incorrectly at first before realising I'd got things the wrong way around).

yksmp/src/lib.rs
ltratt: Up to you!

ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
ptersilie: Before this PR, `0` meant there is no extra location. But now `0` just means `rax`. I just forgot to update the condition.

ltratt: Can we document what the numbers mean (e.g. "dwarf numbering" or whatnot)?

ptersilie: [5566c70](https://github.com/ykjit/yk/pull/1399/commits/5566c70b03729033eeb8314093fb22799d075eb0)

ptersilie:
Ok to squash?

ltratt:
Please squash.

ptersilie:
Squashed.

ltratt:
@ptersilie Please can we update the commit title and also rebase against master to fix the conflict?

ptersilie:
Here we go.

PR #1400, opened 2024-09-20
O1 fixes.

vext01:
Requires a compiler change that will come soon.

(Pre-apologies that this isn't all nicely bundled up into self-contained commits -- it's been quite a journey and picking it apart now doesn't seem worthwhile)

This change is a collection of fixes required for us to enable the -O1 AOT post-link pipeline in Yk. This is slightly different from just passing -O1, but it's a big step in the right direction.

Probably the most significant change is the transition from llvm.experimental.stackamp to llvm.experimental.patchpoint for the control point. This is needed because we were tracking live values at a slightly incorrect point in time. We needed the register state immediately before the call to _ykrt_control_point, but *including* any register shuffling/spills required due to the call ABI.

The lua_array_len.c test was removed because we now run enough optimisations for LLVM to statically detect UB in it. Hardly surprising since this was derived using cvise. We can't handle poison values in the JIT and even if we could, I doubt the test is a good one any more.

Gives between about 12% and 18% speedup on the 4 benchmarks I tried.


vext01:
> 12% to 18%

Raw numbers below.

(The columns don't line up. I know. That's been fixed in a newer rebench [here](https://github.com/smarr/ReBench/issues/244)]

## Before

```
----------------------------------------------------------------------------------------------
  Benchmark   Executor   Suite   Extra   Core      Size   Var   #Samples   Mean (ms)
----------------------------------------------------------------------------------------------
  Json        YkLua      awfy      100   default                                  10    4969
  Richards    YkLua      awfy      100   default                                  10   22313
  DeltaBlue   YkLua      awfy    12000   default                                  10    3213
  CD          YkLua      awfy      250   default                                  10   12209
----------------------------------------------------------------------------------------------
```

## After

```
----------------------------------------------------------------------------------------------
  Benchmark   Executor   Suite   Extra   Core      Size   Var   #Samples   Mean (ms)
----------------------------------------------------------------------------------------------
  Json        YkLua      awfy      100   default                                  10    4152
  Richards    YkLua      awfy      100   default                                  10   18608
  DeltaBlue   YkLua      awfy    12000   default                                  10    2804
  CD          YkLua      awfy      250   default                                  10   10036
----------------------------------------------------------------------------------------------
```

bin/yk-config
```diff
@@ -44,7 +44,9 @@ set_aot_pipeline() {
...
...globalopt,globaldce,elim-avail-extern,rpo-function-attrs,recompute-globalsaa,function<eager-inv>(float2int,lower-constant-intrinsics,loop(loop-rotate<header-duplication;no-prepare-for-lto>,loop-deletion),loop-distribute,inject-tli-mappings,loop-vectorize<no-interleave-forced-only;vectorize-forced-only;>,infer-alignment,loop-load-elim,instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>,simplifycfg<bonus-inst-threshold=1;forward-switch-cond;switch-range-to-icmp;switch-to-lookup;no-keep-loops;hoist-common-insts;sink-common-insts;speculate-blocks;simplify-cond-branch>,vector-combine,instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>,loop-unroll<O1>,transform-warning,sroa<preserve-cfg>,infer-alignment,instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>,loop-mssa(licm<allowspeculation>),alignment-from-assumptions,loop-sink,instsimplify,div-rem-pairs,tailcallelim,simplifycfg<bonus-inst-threshold=1;no-forward-switch-cond;switch-range-to-icmp;no-switch-to-lookup;keep-loops;no-hoist-common-insts;no-sink-common-insts;speculate-blocks;simplify-cond-branch>),globaldce,constmerge,cg-profile,rel-lookup-table-converter,function(annotation-remarks),verify"
```
ltratt, line 49: Is there a reason that we do this rather than just pass `-O1` through to the underlying compiler?

vext01: I believe it's (almost) equivalent, but I'm not quite ready to tear out all of this `-ao` stuff.

I say almost, because we still use an empty pre-link pipeline, which isn't quite the same as passing -O1.

If you don't mind, let's do this incrementally. I'm aiming for allowing the user to pass arbitrary `-ON`, but it needs to happen in steps.

vext01: The line above is pretty explicit with:

```
PRELINK_PASSES_AO1=""
```

I could add a comment, but it'd be like a story you once told me:
```
// set i to one :)
i = 1
```

ltratt: OK, so I'm a bit confused why we need to spell one out and not the other, but maybe this isn't worth worrying about.

vext01: Ah, because our `-ao 1` needn't necessarily be upstream's `-O1` (as is the case for our `-ao 0` and upstream's).

With any luck all of this `-ao` stuff can be deleted soon anyway.

ltratt:
@vext01 Please squash and rebase against the ykllvm master commit.

vext01:
llvm submodule synced/rebased in.

ltratt:
Please force push an update referencing the new ykllvm commit.

PR #1402, opened 2024-09-24
Make `inst_raw` not return a reference.

ltratt:
I recently adjusted the other "get an instruction" functions to do this, but forgot about this one. This commit rectifies that omission.


ltratt:
@ptersilie  This looks like it's probably a bug in nightly that's unconnected to this PR: 74ad9a7 pins us to yesterday's nightly. Maybe give it a kick and see if that fixes things?

PR #1404, opened 2024-09-25
Execute trace on the main interpreter loop frame.

ptersilie:
This simplifies and speeds up the way we access live variables stored on the main interpreter loop frame, as we no longer need to juggle around two base pointers.

We get a decent speed-up of 1.33x on `bigloop.lua`. Since we don't have side-tracing currently, there's not much speed-up to be found with the benchmark suite.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
@@ -222,6 +222,10 @@ impl<'a> LSRegAlloc<'a> {
         self.stack.size()
     }
 
+    pub(crate) fn init_stack(&mut self, size: usize) {
+        self.stack.grow(size);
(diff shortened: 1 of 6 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
```diff
@@ -222,6 +222,10 @@ impl<'a> LSRegAlloc<'a> {
         self.stack.size()
     }
 
+    pub(crate) fn init_stack(&mut self, size: usize) {
```
ltratt, line 225: A memo to me: this should probably be `grow`.

ltratt:
This is worth it for https://github.com/ykjit/yk/pull/1404/commits/0176e1ec0069e52c9884a95be2ac9cc06edb299e alone!

PR #1410, opened 2024-10-03
Rewrite the optimiser to be a one pass optimiser.

ltratt:
[Well, OK the first lie is in the first line. Because we have to do backwards code generation, we actually have to do a second pass. But that's a boring detail, and something that predates this commit.]

This commit moves the optimiser into a model that's closer to RPython: rather than multiple forward stages each doing one optimisation, we have a single forward stage that does all optimisations in one go. This is quicker (fairly obviously!) but also simpler, because it means that we can make strong assumptions at the point of the wavefront: no-one can change our knowledge of earlier parts of the trace.

The crucial difference between this commit and what came before is that as well as mutating the IR as we go along (e.g. constant folding) we maintain an analysis of what values an instruction can produce, which we gradually refine. Crucially, mutation can only happen at the current instruction, but analysis refinement can happen on previous instructions.

Consider a trace along these lines:

```
%0: i8 = load_ti ... ; at this point we know nothing about %0
%1: i8 = add %0, 1i8 ; we still know nothing about %0
%2: i1 = eq %0, 0i8  ; we know that henceforth %0 must be 0i8
guard true %2
%4: I8 = mul %0, %0  ; we know this must equal the constant 0i8
```

Notice that it's only at the point of the guard that our analysis allows us to start using the knowledge "%0 = 0i8" for optimisations.

The code is currently woefully incomplete (but if I fill in anything else, I trip up on the "removing guards tends to kill things" bug that I believe is fixed but not yet merged), I haven't yet bothered porting across one old optimisation (`mul_chain`), and I haven't really got a good API for sharing analysis and mutations (which is currently done inconsistently).

All that said, this nudges the optimiser forward just about enough that I think it's worth considering for merging now, and then it can be improved in-tree.


vext01:
Neat.

About this comment in the commit message.

> ; we know that henceforth %0 must be 0i8` at the point of the comparison.

Would be better as something like:

> ; If %2 is used as a guard condition, then it must assume the value 0i8 immediately after the guard.

But there's a hitch, that's only true if it's a `guard true`. In the case of a `guard false` we'd know that `%2` is an element of [1, 255] (i.e. that it's not zero).

Perhaps a tweak to clarify?

ykrt/src/compile/jitc_yk/opt/analyse.rs
```diff
@@ -0,0 +1,82 @@
...
+                            if pred == Predicate::Equal && m.const_(lhs_cidx) == m.const_(rhs_cidx)
+                            {
+                                self.values[usize::from(iidx)] = Value::Const(m.true_constidx());
+                                return Operand::Const(m.true_constidx());
+                            }
+                        }
+                    }
+                    op
+                }
+                Value::Const(cidx) => Operand::Const(cidx),
```
vext01, line 48: Potential future work out of scope of this PR.

Instead of tracking a single constant value, you could plug other domain(s) so that you could track richer information and use the analysis more widely for other optimisations besides just constant propagation.

For example an interval domain to express ranges, or a relational domain to express that `%2 = %1 + 1`, or `%2 < %1`.

ptersilie:
Looks good.

ykrt/src/compile/jitc_yk/opt/analyse.rs
ltratt: Yes, that's what RPython does, but one thing at a time.

ltratt:
>But there's a hitch, that's only true if it's a guard true. In the case of a guard false we'd know that %2 is an element of [1, 255] (i.e. that it's not zero).
>
>Perhaps a tweak to clarify?

That would make a very long line, and it seems fairly obvious to me from context?

vext01:
Well it wasn't obvious to me, but let's not hold this up.

PR #1411, opened 2024-10-03
Get rid of `ykconfig -ao`.

vext01:
Requires: https://github.com/ykjit/ykllvm/pull/209 (so will need a push to re-sync ykllvm)

Not so long ago we added `yk-config -ao <level>` to allow us to run our pre-canned franken-O1 pipeline for the entire test suite. This was only ever designed as an interim hack, and we've wanted Yk to simply work with standard optimisation flags like -O1, -O2 etc.

This change is a step in that direction.

We remove `-ao` and our non-standard defaults and add the ability for individual tests to specify extra compiler flags. This gives us the ability to specify optimisation parameters on a per-test basis.

I then ported as many tests as were practical to use -O1. The remaining tests (that were either too involved for me to convert right now, or actually rely on an IR shape generated by our old default pipeline) I've left using our franken-O1 by passing:

-O0 -Xclang -disable-O0-optnone -Xlinker \
    --lto-newpm-passes=instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>

We do have to disable yklua in CI for now, as yklua still uses `yk-config -ao` and we have to break a circular dependency.


bin/yk-config
```diff
@@ -77,32 +34,21 @@ handle_arg() {
...
-            #
-            # These are run on the individual pre-LTO-merged LLVM modules.
-            if [ ! -z "${PRELINK_PASSES}" ]; then
-                OUTPUT="${OUTPUT} -mllvm --newpm-passes=${PRELINK_PASSES}"
-            fi
-            case $profile in
-                debug*) OUTPUT="$OUTPUT -g" ;;
-            esac
+
+            # Disable backend optimisations that Yk can't handle by adding the
```
vext01, line 48: Just noticed a mistake here. We have a linker flag in the --cflags section!

Will fix and retest.

vext01: Fixed in [d874f9c](https://github.com/ykjit/yk/pull/1411/commits/d874f9cd2a4d0c9b67317ff4accf5e8991ee81a7)

tests/c/aot_debuginfo.c
```diff
@@ -1,4 +1,5 @@
-// ignore-if: test $YK_CARGO_PROFILE != "debug"
+// Compiler:
+//   env-var: YKT_EXTRA_CC_FLAGS=-O1 -g
```
ltratt, line 2: What does `YKT` mean?

ltratt: I *think* this should be `YKB_EXTRA_...`?

vext01: `YKT` was traditionally used for testing only, but I see it's been renamed/removed at some point.

You can't use this variable outside of tests since it requires running the compiler via the compiler wrapper. Is that OK still?

ltratt: Should we just rename it to `YKB` since it's a build-time option (even if only "build testing")? If you think that's the way to go, you can rename + squash in one go.

vext01: OK.

I must admit I get confused at times with our variables. I've seen so many reincarnations of them that I'm just constantly lost!

ltratt: I know the feeling! FWIW, I *think* the prefixes are now documented in the book tolerable well?

vext01: Yeah mostly. There's a couple of things that often still catch me out:

 - variables with no `D` or `B`, like `YK_LOG`. I always type `YKD_LOG`.
 - the mandatory `<path>:` prefix of `YKD_LOG_IR`, because I was spoiled by the optional prefix of `YK_LOG` :)

ltratt: > the mandatory <path>: prefix of YKD_LOG_IR, because I was spoiled by the optional prefix of YK_LOG :)

Hmm, you're right, that shouldn't be mandatory! PR welcome!

vext01: I know. It's just such a small fish right now!

vext01: Anyway YKT -> YKB in [a04c8f8](https://github.com/ykjit/yk/pull/1411/commits/a04c8f8a0efdc5c2ed2fc5f20e9fa0ab36560fdf)

ltratt:
Please squash.

vext01:
I've also sneaked in [3eda80d](https://github.com/ykjit/yk/pull/1411/commits/3eda80d547f8d58f3710d922289fe614091ccf6f) which will be needed for the yklua tests when we re-enable them.

Its a flag we already have enabled for link time, but is also needed for pre-link time, since that pas can (and does) run at both times.

ltratt:
Yes I saw it. Please squash.

vext01:
squashed.

vext01:
My apologies. I force pushed a change to sync in ykllvm too. Ready.

vext01:
Just a test failing due to SWT recorder calls.

Fixed in [7187176](https://github.com/ykjit/yk/pull/1411/commits/7187176d13b1444011a093f5a6e1933ef7492302)

OK to squash?

ltratt:
Please squash.

vext01:
squashed

PR #1418, opened 2024-10-10
Optimise codegen for 64-bit addition with a constant that fits in 32-bits. 

vext01:
I started out trying to cover too many cases any getting confused, so
this optimises only one very specific case:

    %2: i64 = add %1, <const>

Where the numeric value of `<const>` is expressible in 32 bits.

For this we use the instruction encoding:
```
  add r64, imm32
```

This builds on a sketch @ltratt started and handed to me to look at. His
work is the first commit. It contains some semi-related promote
improvements also.

I had hoped to make a series of self-contained commits, but the work wen
through so much flux that it didn't work out, and would only force the
reviewer to review stale code.

Here's some miscellaneous notes on the changes:

 - Fix a bugs in `Truncate` and `SignExtend`.
 - Unit test and document `Truncate` and `SignExtend` too.
 - Fix incorrect bit size computation (byte_size * 8 isn't what we want).
 - Unit test the optimisation.
 - Add C test to check runtime semantics of the optimisation.
 - size -> byte_size, bits -> bit_size (for improved comprehension)

All lua tests and benchmarks work.

bigloop.lua uses this optimisation in the trace it generates.

I think @ltratt should review my changes, but since this PR is technically co-authored, also assigned @ptersilie .

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -10,6 +10,7 @@
 
 use super::{
     super::{
+        int_signs::{SignExtend, Truncate},
         jit_ir::{self, BinOp, FloatTy, InstIdx, Module, Operand, Ty},
         CompilationError,
     },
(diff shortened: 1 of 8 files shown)


tests/c/x64_imm_add.c
```diff
@@ -0,0 +1,52 @@
+// ignore-if: test $(uname -m) != x86_64
```
ltratt, line 1: I don't think this should be a C test: it should be a unit test. C tests are slow and this one doesn't do anything a unit test couldn't AFAICS.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -419,13 +420,47 @@ impl<'a> Assemble<'a> {
...
+                        // expressed in 32-bits...
+                        //
+                        // We are only optimising (exactly) 64-bit add operations for now, but
+                        // there is certainly oppertunity to optimised other cases later.
+                        //
+                        // We could have used Rust `as` casts to truncate and sign-extend here,
+                        // since we are currently only dealing with i32s and i64s, but if/when we
+                        // want to cover operations on other "odd-bit-size" integers we will need
+                        // these custom implementations.
+                        if *bit_size == 64 && v.truncate(32).sign_extend(32, 64) == *v {
```
ltratt, line 444: We could, of course, easily handle `i32` here too!

ltratt:
At first glance (I've only done a quick read) this looks OK, modulo the C test, which I think serves no purpose, and just slows the test suite down.

vext01:
> modulo the C test, which I think serves no purpose

It's testing the runtime behaviour of the optimisation, but I realise that I forgot to add the expected stdout.

As I understand, the unit tests only test the static assembler emitted.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
vext01: Yes, in theory we could handle many different cases using this one instruction encoding, but I wanted to do the MVP in this first PR.

vext01: Ah, I  just realised that you mean we could optimise using `add r32, imm32`?

Yep, as another PR if that's ok?

tests/c/x64_imm_add.c
vext01: We have unit tests in addition to this. If you are happy for other (less targeted) tests to check the runtime semantics of the optimisation, we can delete this one though.

Entirely up to you.

ltratt: What I'm saying is I *think* everything this C test tests for *should* be covered in unit tests in ykrt *and* those tests will run much, much, quicker. If I'm right about those things, I'd prefer to delete this test.

vext01:
OK to squash?

ltratt:
Fine with me but I think @ptersilie should lead from now on.

ptersilie:
Please squash.

vext01:
splat

PR #1423, opened 2024-10-16
Avoid use of 8 and 16-bit registers in x64 codegen. 

vext01:
The codegen routines now try to avoid the use of operations which set 8
or 16-bit sub-registers. This is for two reasons: a) to simplify the
code, but also b) to avoid partial register stalls in the CPU pipeline.
Generally speaking we prefer to use 64-bit operations, but sometimes we
special-case 32-bit operations since they are so common.

If an object occupies only part of a 64-bit register, then there are no
guarantees about the unused higher-order bits. Codegen routines must be
mindful of this and, depending upon the operation, may be required to
(for example) "mask off" or "sign extend into" the undefined bits in
order to compute the correct result.


vext01:
(i'll want to update the commit message if we decide to merge this)

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -390,6 +391,57 @@ impl<'a> Assemble<'a> {
         alloc_off
     }
 
+    /// Sign extend the `from_bits`-sized integer stored in `reg` up to the full size of the 64-bit
+    /// register.
+    ///
+    /// `from_bits` must be between 1 and 64.
+    fn sign_extend_reg64(&mut self, reg: Rq, from_bits: u8) {
```
ltratt, line 398: Should probably be `sign_extend_to_reg64`?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -489,15 +543,17 @@ impl<'a> Assemble<'a> {
                     [RegConstraint::InputOutput(lhs), RegConstraint::Input(rhs)],
                 );
                 match size {
-                    1 => dynasm!(self.asm; and Rb(lhs_reg.code()), Rb(rhs_reg.code())),
-                    2 => dynasm!(self.asm; and Rw(lhs_reg.code()), Rw(rhs_reg.code())),
-                    4 => dynasm!(self.asm; and Rd(lhs_reg.code()), Rd(rhs_reg.code())),
-                    8 => dynasm!(self.asm; and Rq(lhs_reg.code()), Rq(rhs_reg.code())),
+                    1..=8 => {
```
ltratt, line 546: Why until `8`?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -526,16 +585,17 @@ impl<'a> Assemble<'a> {
...
                 );
                 debug_assert_eq!(_rhs_reg, Rq::RCX);
-                match size {
-                    1 => dynasm!(self.asm; shr Rb(lhs_reg.code()), cl),
-                    2 => dynasm!(self.asm; shr Rw(lhs_reg.code()), cl),
-                    4 => dynasm!(self.asm; shr Rd(lhs_reg.code()), cl),
-                    8 => dynasm!(self.asm; shr Rq(lhs_reg.code()), cl),
+                match bit_size {
+                    1..=64 => {
+                        // Ensure we shift in zeros at the most-significant bits.
```
ltratt, line 590: I eventually worked out what this means (I think), but it confused me for a while. I think it means "For the `shr` to work correctly, we need to ensure there are no undefined bits, as they would end up appearing in the defined part of the register".

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -567,102 +627,104 @@ impl<'a> Assemble<'a> {
...
-                // use `cdq`/`cqo` to double the size via sign extension and store the result in
-                // DX:AX, EDX:EAX, RDX:RAX.
-                match size {
-                    // There's no `cwd` equivalent for byte-sized values, so we use `movsx`
-                    // (sign-extend) instead.
-                    1 => dynasm!(self.asm; movsx ax, al; idiv Rb(rhs_reg.code())),
-                    2 => dynasm!(self.asm; cwd; idiv Rw(rhs_reg.code())),
-                    4 => dynasm!(self.asm; cdq; idiv Rd(rhs_reg.code())),
-                    8 => dynasm!(self.asm; cqo; idiv Rq(rhs_reg.code())),
+                match bit_size {
```
ltratt, line 672: A lot of these `match`es are also `if/else` in disguise (but with more indentation).

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -674,38 +736,37 @@ impl<'a> Assemble<'a> {
...
-                    2 => dynasm!(self.asm; div Rw(rhs_reg.code())),
-                    4 => dynasm!(self.asm; div Rd(rhs_reg.code())),
-                    8 => dynasm!(self.asm; div Rq(rhs_reg.code())),
+                debug_assert_eq!(lhs_reg, Rq::RAX);
+                match bit_size {
+                    1..=64 => {
+                        self.zero_extend_reg64(lhs_reg, u8::try_from(*bit_size).unwrap());
+                        self.zero_extend_reg64(rhs_reg, u8::try_from(*bit_size).unwrap());
+                        dynasm!(self.asm
+                            ; xor rdx, rdx // Zero extend RAX into RDX:RAX.
```
ltratt, line 766: This means "put 0 in rdx" I think?

vext01:
Investigating the source of that slowdown has been quite enlightening.

First, this branch doesn't make the bigloop.lua asm trace much longer. Before it was 409 instructions long, now it is 414.

The extra instructions are zero extends from 32 to 64 bits. This is required by operations like 32-bit `shr` where the semantics require the top 32-bits to be zeroed if we are to use a 64-bit shift operation (because you want zeros to come into the top bits, not undefined, potentially set bits).

The way we zero extend a 32-bit register up to 64-bits are special-cased like `mov eax, eax`. I wrote a little asm program that simply counts up to `u32::MAX` a few times. Adding a zero extend of this kind into the loop slowed the little program down by about 2x! Obviously the cost is inflated here because the loop is so tight.

But this looks like the source of the extra cost we are seeing in benchmarks. If we want to use 64-bit registers wherever possible, this is a cost we have to shoulder.

Any thoughts? Do you think the simplified code-gen is worth the performance hit?

Also found a bug that I'll push a fix for shortly.

vext01:
> For example, if we say that all "unused" bits are always zero

Right, but if you do that lots of operations will have to do zero-extension to meet that requirement, even if no later operation would actually take advantage of the requirement!

For example, if we have a tight loop that only ever adds 32-bit numbers (an operation which can pollute the top 32-bits), each add will be required to mask off the top 32-bits just in case an instruction like `shr` consumes the result -- which for this contrived example, it doesn't.

> For example, if we say that all "unused" bits are always zero, that changes things. What do RPython/Cranelift/etc do in this regard?

For right shifts, cranelift masks like this branch does (but there is not a focus on performance). Bear with me and I may be able to figure out what the others do.

ltratt:
Another thought: IIUC our theory is that 8 and 16 bit registers have bad performance properties; but 32 and 64 bit registers have the same performance properties. If so, we can special case 32-bit sizes and generate instructions for them that don't bother clearing the top 32 bits?

vext01:
> If so, we can special case 32-bit sizes and generate instructions for them that don't bother clearing the top 32 bits?

Let me see if I understand your proposal.

Taking logical shift right as an example. If the value to shift is **exactly** 32 bits wide, then use `shl r32, imm8`, otherwise zero off the unused top bits and use `shl r64, imm8`. Is this what you mean?

Yes, we could do that and it seems like it could be a decent compromise.

Note that in this scenario, we'd still need to zero off top bits if the value fits in 32 bits, but has < 32 defined bits. E.g. if you were to shift right a 30-bit integer, you still have undefined bits at the top regardless of whether you use a 32 or 64 bit shift instruction.

Sounds good?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
vext01: Because once you go beyond 8 bytes, you can't sore the value in a register any more and you'd have to do something different.

(Would it have helped if the `size` were called `byte_size`?)

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
vext01: Yes, but I was trying to keep the comments snappy as it's a recurring theme and we don't want to duplicate a lengthy explanation every time the theme reoccurs.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
vext01: I used match to take advantage of the `1..=64` notation, but I will convert them to ifs.

ltratt:
>  Taking logical shift right as an example. If the value to shift is exactly 32 bits wide, then use shl r32, imm8, otherwise zero off the unused top bits and use shl r64, imm8. Is this what you mean?

Yes.

> Note that in this scenario, we'd still need to zero off top bits if the value fits in 32 bits, but has < 32 defined bits

But then it's not *exactly* 32 bits, right? As you point out above, my suggestion is for the *exactly* 32 bit case only.

vext01:
> But then it's not exactly 32 bits, right? As you point out above, my suggestion is for the exactly 32 bit case only.

Yes, indeed. We are on the same page.

I'll try this strategy after fixing the handful of comments you added.

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
```diff
@@ -528,20 +526,15 @@ impl LSRegAlloc<'_> {
...
-                // The `as`s are all safe because the IR guarantees that no more than `width` bits
-                // are set in the integer i.e. we are only ever truncating zeros.
-                match width {
-                    1 | 8 => dynasm!(asm; mov Rb(reg.code()), BYTE *x as i8),
-                    16 => dynasm!(asm; mov Rw(reg.code()), WORD *x as i16),
-                    32 => dynasm!(asm; mov Rd(reg.code()), DWORD *x as i32),
-                    64 => dynasm!(asm; mov Rq(reg.code()), QWORD *x as i64),
+                // `unwrap` cannot fail, integers are sized.
+                match self.m.type_(*tyidx).byte_size().unwrap() {
+                    s if s <= REG64_BYTESIZE => dynasm!(asm; mov Rq(reg.code()), QWORD *x as i64),
```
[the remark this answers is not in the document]
vext01: Fixed in 3d566df36e0a1bb2f4abf705fd82c5e50c9392b9

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
vext01: Fixed in 53f3c3aa4f0ff0ad1a4cc524a5430646683a894a

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
vext01: Good catch. That's a stale comment. Fixed in f59d229a137f6069551581e65655da620e0f15f2

vext01:
Comments addressed.

In terms of PR discussions, I think that only leaves https://github.com/ykjit/yk/pull/1423#discussion_r1804733156 and https://github.com/ykjit/yk/pull/1423#discussion_r1804714427 that require your attention.

Please also take a look at 874812f6b4a252db7c3053749c40f92cd8ef77d8 and 9cbbd31117201894ce7179b3d85d24a7ae751bbc which are also new.

vext01:
I also noticed that there are other places in the codegen besides binops that require the same treatment:
 - cg_icmp
 - cg_sext
 - cg_zext
 - cg_trunc
 - cg_select

I'd like to handle those in a follow up PR if that's OK with you?

ltratt:
For some reason I can't comment on those commits.

Rather than comments like:

> // Note that it's UB to shift by >= the bit width of the first operand.

why not put in an assert? Or is that not possible? [Warning: I haven't fully thought this through.]

vext01:
> why not put in an assert?

Since (generally) the amount to shift by is a dynamic value (in a register), you can't really add an assertion into the code generator. If you wanted to crash the system, you'd have to emit code to do the check.

(Once we special case shifts by constants, we could in that specific case, but read on...)

But I don't think we need to. This is an LLVM IR constraint that we've inherited. LLVM stipulates that a poison value is generated, and in that case we are free to generate any code we wish, including pretending it doesn't happen.

I wonder then, if the comment actually does more harm that good? What do you think?

ltratt:
> I wonder then, if the comment actually does more harm that good? What do you think?

I think the comment is almost right: it should say something like "In LLVM IR, and thus our JIT IR, it's UB to shift by more than XYZ. Fortunately our generated code naturally handles al the defined behaviour cases and we don't have to worry about the UB."

vext01:
Fixed in fa2719253d37.

It's not something that can easily be explained in a one-liner unfortunately.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
vext01: The more I work on this system, the more I think variables and methods talking about a size should mention the units.

I'll admit I've confused bits for bytes in past. It's easily done.

ltratt: > Would it have helped if the size were called byte_size?

Yes it would!

vext01: Fixed in 70b5ab8e53dd41d1c07a0482f49c640420c7020d and 3b79aa5ae485ec050fd658a6277101b9e03479ff.

I'll leave the latter as a separate commit, since the rest of this PR is about integer binops.

vext01:
Should I add tests for all the exactly 32-bit code-gen paths?

I'm torn. One part of me wants to, another thinks we will go mad if we try to test every path through the codegen...

ltratt:
> Should I add tests for all the exactly 32-bit code-gen paths?

Yes, but they don't have to be new test *functions*. In these cases, I happily whack multiple test cases in a single function. I find that means it only takes me seconds to cover additional cases.

The reason we want to do this is because (have we measured this yet?) we believe this is both a performance and a correctness property, right?

vext01:
> have we measured this yet?

good point! I'll do that first.

vext01:
Even with serialised compilation enabled, I'm now measuring less than a 1% slowdown with bigloop, and similar figures for the other benchmarks.

So I'll now focus on some more tests and then this should be good to go in.

ltratt:
> So I'll now focus on some more tests and then this should be good to go in.

Agreed!

vext01:
https://github.com/ykjit/yk/pull/1423/commits/40fc3ddb9d48415440c6bf436f23dec0cdee2ce7 works around some quirks in the JIT IR parser required for better testing.

https://github.com/ykjit/yk/pull/1423/commits/62b865be0b5a7283b35da9cbeae1f9ab73fe7e94 tests everything I've touched here.

OK to squash/rebase?

ltratt:
Re: 40fc3dd can we find/add an issue that lumps all these yksmp-ish things together? We are going to have to deal with this properly soon-ish I suspect.

ltratt:
Please squash and rebase.

vext01:
Squashed + rebased + reworded.

ltratt:
Please mark as not draft.

vext01:
done

ltratt:
I had misunderstood this PR (totally my fault): I thought this fixed the entire generator but @vext01 pointed out that it only does binops. I *think* that means that various things now only work by accident within this PR as it currently stands.

@vext01 Is my intuition that this leaves us in an incorrect in-between state correct? If so, I think we should extend this PR until the change is complete.

vext01:
> Is my intuition that this leaves us in an incorrect in-between state correct? 

Technically yes, but all the yk tests, yklua tests and benchmarks apparently don't hit these cases.

We can add to this PR, yes.

On my list of things to audit next are:
 - cg_icmp
 - cg_trunc
 - cg_sext
 - cg_zext
 - cg_select

ltratt:
> all the yk tests, yklua tests and benchmarks apparently don't hit these cases.

Right. So that means we should add a new test now that we've identified the gap. I'm fine with a unit test (which also should be much easier to write than a C test for this).

> We can add to this PR, yes.

Thanks -- and sorry!

vext01:
> I'm fine with a unit test

OK, but it's likely that the codegen routines themselves will require a small amount of fiddling too.

Leave it with me. Cheers.

vext01:
I believe that last round of commits fixes all other places.

So to be clear, this change:
 - avoids the use of 8 and 16-bit registers wherever possible.
 - considers unused bits at the top of a register undefined.

I wonder if we should document this somewhere.

ltratt:
https://github.com/ykjit/yk/commit/129356581692082d95f14df87b747c41d5964675 is probably the wrong way around? I think you can actually just say something like "For a type of *n* bits, bits *64-n* in a register are undefined. If their value is important to an operation, those bits must be dealt with as if they could have any possible value".

vext01:
Thanks. I've squashed, leaving that comment as-is for now.

(I've started a branch that does some related cleanup, including removing all of this "finding the bit size" duplication that's crept in)

vext01:
https://github.com/ykjit/yk/pull/1423/commits/d85dadf8ea988695b8d470723cbe50389735ee90 fixes the CI failure. I guess we leave that commit separate.

ltratt:
Given other PRs have merged, I wonder if this is a weird transient error. Retrying.

vext01:
I think i see what has happened. The second commit changes ykllvm, but it shouldn't.

vext01:
https://github.com/ykjit/yk/pull/1423/commits/0d64d2e2c774a8448637dafb81edfbeb9ee718ae puts the ykllvm module back to what it should be.

If that's OK with you, it should be squashed into the second commit.

ltratt:
Please squash.

vext01:
squashed and rebased.

PR #1429, opened 2024-10-23
Adjust the root jump target for side-traces.

ptersilie:
When side-traces jumped back to the root trace, they previously always jumped to the entry address of the root trace. Since we want to implement loop peeling soon, we need to be able to jump to an offset inside the root trace. For now this is simply the place after the trace's prologue. But later we want to change this to the place after the preamble and before the peeled loop.

This change required some additional changes to the way we calculate and reset the base pointer when jumping to the root trace.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -159,8 +159,9 @@ impl CodeGen for X64CodeGen {
         mt: Arc<MT>,
         hl: Arc<Mutex<HotLocation>>,
         sp_offset: Option<usize>,
+        root_offset: Option<usize>,
     ) -> Result<Arc<dyn CompiledTrace>, CompilationError> {
-        Assemble::new(&m, sp_offset)?.codegen(mt, hl, sp_offset.is_some())
+        Assemble::new(&m, sp_offset, root_offset)?.codegen(mt, hl, sp_offset.is_some())
     }
 }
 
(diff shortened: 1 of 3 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -236,6 +242,11 @@ impl<'a> Assemble<'a> {
 
         let alloc_off = self.emit_prologue();
 
+        // The instruction offset after we've emitted the prologue (i.e. updated the stack
+        // pointer). We will later adjust this offset to also include one iteration of the trace
+        // (the preamble) so we can jump directly to the peeled loop.
+        let prologue_offset = self.asm.offset();
```
ltratt, line 248: Does RPython call this "preamble"? FWIW, I think I prefer "prologue" as the terminology.

ptersilie: Yeah that's an RPython term. We can stick to prologue if you like. I'll fix it.

ltratt: Let's stick with prologue.

ptersilie: [baf68c8](https://github.com/ykjit/yk/pull/1429/commits/baf68c8fb211101a86a6785dc45925de6fb989f7)

ltratt: Ah, I see why the change -- thanks!

ltratt:
Please squash.

ptersilie:
Squashed.

ptersilie:
Resolved conflicts.

ptersilie:
Rebased.

PR #1432, opened 2024-10-24
Implement common subexpression elimination.

ltratt:
This is, deliberately, simplistic at the moment, with terrible big-O performance: we search backwards for every instruction. In practise, this probably isn't a horrible bottleneck, at least for the length of traces we currently see. It can be easily optimised in the future.


vext01:
Looks good to me.

Please resolve conflicts and we can merge.

PR #1438, opened 2024-10-28
Generate more immediates

ltratt:
This PR gradually winds its way towards generating more code with immediates, but it does so in several steps (some of which are a bit tangential as I worked out what I was doing). The overall effect is a 10% improvement in big_loop. The biggest single improvement is in https://github.com/ykjit/yk/commit/f66bb5ec1335d4d53faa4f40b04dcd9ecc2142cb, but a couple of the other commits improve things by a percentage point or two.


ltratt:
@ptersilie Those 3 new commits bring the total speed boost of this PR to 19%.

vext01:
LGTM

PR #1439, opened 2024-10-28
Optimise `DynPtrAdd`.

ltratt:
When possible use `lea`; if not, aim for `shl; add`; and fall back on the general case of `imul; add` if necessary. Speeds big_loop up by a bit over 2%.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -1165,15 +1165,33 @@ impl<'a> Assemble<'a> {
         // LLVM semantics dictate that the element size and number of elements should be
         // sign-extended/truncated up/down to the size of the LLVM pointer index type. For address
         // space zero on x86_64, truncation can't happen, and when an immediate third operand is
-        // used for x86_64 `mul`, it is implicitly sign extended.
-        dynasm!(self.asm
-            // multiply the element size by the number of elements.
-            ; imul Rq(num_elems_reg.code()), Rq(num_elems_reg.code()), i32::from(inst.elem_size())
-            // add the result to the pointer. We make use of addition's commutative property to
-            // reverse the "obvious" ordering of registers: doing so allows us not to overwrite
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1165,15 +1165,33 @@ impl<'a> Assemble<'a> {
...
-        // used for x86_64 `mul`, it is implicitly sign extended.
-        dynasm!(self.asm
-            // multiply the element size by the number of elements.
-            ; imul Rq(num_elems_reg.code()), Rq(num_elems_reg.code()), i32::from(inst.elem_size())
-            // add the result to the pointer. We make use of addition's commutative property to
-            // reverse the "obvious" ordering of registers: doing so allows us not to overwrite
-            // ptr_reg.
-            ; add Rq(num_elems_reg.code()), Rq(ptr_reg.code())
-        );
+        // used, that also isn't a worry.
```
vext01, line 1168: I'm curious as to what this change in the comment means.

We are still required to sigh extend the 32-bit operand up to 64-bits, which still occurs implicitly with x64 `mul`. Or have I missed something?

ltratt: I don't understand the question.

vext01: The comment was changed from:

> when an immediate third operand is used for x86_64 `mul`, it is implicitly sign extended

to:

>  and when an immediate third operand is that also isn't a worry.

But I'm not sure why. I wonder if I missed something important.

ltratt: We're not just `imul` after this commit (we're using `lea` and `shl` too). I *believe* that the comment now correctly covers those cases, but please see if you agree!

vext01: OK, I see what you are saying.

PR #1441, opened 2024-10-29
More trace and x64 optimisations

ltratt:
This PR bundles up a number of optimisations that benefit big_loop.lua: mostly trace optimisations, but also one x64 optimisation (and one bug fix). With a suitable `yk_promote` in yklua, this PR speeds big_loop up by about 43%. The main reason for this is that we are able to constant fold a lot of goop around instruction decoding, so a lot of stuff in `trace-pre-opt` disappears entirely by `trace-post-opt`.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -1745,7 +1745,10 @@ impl<'a> Assemble<'a> {
         let tgt_vars = tgt_vars.unwrap_or(self.loop_start_locs.as_slice());
         for (i, op) in self.m.loop_jump_vars().iter().enumerate() {
             let (iidx, src) = match op {
-                Operand::Var(iidx) => (*iidx, self.ra.var_location(*iidx)),
+                Operand::Var(iidx) => {
+                    let iidx = self.m.inst_decopy(*iidx).0;
+                    (iidx, self.ra.var_location(iidx))
+                }
                 _ => panic!(),
             };
             let dst = tgt_vars[i];
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -860,20 +860,37 @@ impl<'a> Assemble<'a> {
                 let Ty::Integer(bit_size) = self.m.type_(lhs.tyidx(self.m)) else {
                     unreachable!()
                 };
-                let [lhs_reg, rhs_reg] = self.ra.assign_gp_regs(
-                    &mut self.asm,
-                    iidx,
-                    [RegConstraint::InputOutput(lhs), RegConstraint::Input(rhs)],
-                );
-                match bit_size {
-                    0 => unreachable!(),
-                    32 => dynasm!(self.asm; sub Rd(lhs_reg.code()), Rd(rhs_reg.code())),
-                    1..=64 => {
-                        self.sign_extend_to_reg64(lhs_reg, u8::try_from(*bit_size).unwrap());
-                        self.sign_extend_to_reg64(rhs_reg, u8::try_from(*bit_size).unwrap());
-                        dynasm!(self.asm; sub Rq(lhs_reg.code()), Rq(rhs_reg.code()));
+                if let Some(0) = self.op_to_i32(&lhs) {
+                    let [rhs_reg] = self.ra.assign_gp_regs(
+                        &mut self.asm,
+                        iidx,
+                        [RegConstraint::InputOutput(rhs)],
+                    );
+                    match bit_size {
+                        0 => unreachable!(),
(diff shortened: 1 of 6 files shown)


ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -268,6 +269,22 @@ impl Opt {
                     },
                     _ => (),
                 },
+                Inst::DynPtrAdd(x) => {
+                    if let Operand::Const(cidx) = self.an.op_map(&self.m, x.num_elems(&self.m)) {
+                        let Const::Int(_, v) = self.m.const_(cidx) else {
+                            panic!()
+                        };
+                        // Exceeding `u16::MAX` is undefined behaviour: we choose to saturate
```
vext01, line 277: I wasn't sure where the `u16::MAX` constraint comes from.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1745,7 +1745,10 @@ impl<'a> Assemble<'a> {
         let tgt_vars = tgt_vars.unwrap_or(self.loop_start_locs.as_slice());
         for (i, op) in self.m.loop_jump_vars().iter().enumerate() {
             let (iidx, src) = match op {
-                Operand::Var(iidx) => (*iidx, self.ra.var_location(*iidx)),
+                Operand::Var(iidx) => {
```
vext01, line 1748: Is this testable?

vext01:
Looks good. Just a couple of comments.

ykrt/src/compile/jitc_yk/opt/mod.rs
ltratt: It's totally arbitrary: anything above the bit width of the type is UB.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ltratt: It will get tested in an upcoming PR, but for boring reasons it's a bit awkward to do here: it's just a flat out bug that happens to cause problems in yklua (but you need several things to come together for it to explode).

ykrt/src/compile/jitc_yk/opt/mod.rs
vext01: > anything above the bit width of the type is UB.

If I'm understanding correctly, LLVM allows you to compute pointers to anywhere in memory regardless of the type you are operating on.

[The docs](https://llvm.org/docs/LangRef.html#id238) say:

> The result value of the getelementptr may be outside the object pointed to by the base pointer. 

Hopefully I haven't gotten the wrong end of the stick.

ltratt: Oops, sorry I misread where you were commenting on!

The docs don't seem to specify a maximum offset, but there obviously must be a maximum. I assumed, based on some other things, that a `u16` offset might be enough, but I admit that I didn't follow up on this. Any thoughts as toe the maximum offset?

vext01: I *think*  the maximum offset is determined by what LLVM calls the "pointer index type" for the address space in question.

It's documented in that same section of the docs (the same passage that I was fretting over in MM yesterday):

> The indices are first converted to offsets in the pointer’s index type. If the currently indexed type is a struct type, the struct offset corresponding to the index is sign-extended or truncated to the pointer index type. Otherwise, the index itself is sign-extended or truncated, and then multiplied by the type allocation size (that is, the size rounded up to the ABI alignment) of the currently indexed type.
> 
> The offsets are then added to the low bits of the base address up to the index type width, with silently-wrapping two’s complement arithmetic. If the pointer size is larger than the index size, this means that the bits outside the index type width will not be affected.

For us, it's important to know:
 - We only allow GEP on address space zero, and our [IR serialiser will reject all other GEPs](https://github.com/ykjit/ykllvm/blob/3ff37cbf086511357547776925f9a8b983533b82/llvm/lib/YkIR/YkIRWriter.cpp#L950)
 - We [assume and assert](https://github.com/ykjit/ykllvm/blob/3ff37cbf086511357547776925f9a8b983533b82/llvm/lib/YkIR/YkIRWriter.cpp#L1003) that the pointer index type is the same size as a pointer for address space zero.

Under these assumptions, I think the maximum offset is expressed by a signed 64-bit integer, not an unsigned 16-bit one. I think we are also expected to sign-extend the index if it's smaller. But please check my reasoning.

ltratt: OK, so this is messy because *we* have an `i32` limit, so we hit "badness" before LLVM says we should hit UB. 7d7e73 is my attempt to split the difference...

vext01: Hrm. Good point.

Let's roll with this for now.

vext01:
Please squash.

ltratt:
Squashed.

PR #1442, opened 2024-10-29
Remove now-unnecessary hack that keeps variables alive.

ltratt:
Before we had loop jumping, we didn't really know how long to keep variables alive for: we thus added a temporary hack (ha!) to keep alive all trace inputs until the end of the trace. After we added loop jumping, this became unnecessary -- but, of course, we all (including me!) had forgotten the temporary hack ever existed!

This commit removes the hack, which I accidentally stumbled across when I found a test I couldn't express because of it.

Removing this has a much more positive effect than I expected:

  1. We remove quite a bit of pressure from the register allocator (always a good thing)
  2. We don't have to restore anywhere near as many variables at the end of a trace (some of which we had to pop from the stack, which is always slow).
  3. In tests, we no longer add as many mystery "do something with the stack" instructions, making tests easier to understand.

Most of this commit just adjusts the test suite because of (3). In a couple of cases, the tests now become a bit more verbose, because the old tests accidentally relied on the register allocator keeping registers alive too long and thus making registers accidentally "unique" because it couldn't reuse them.


ltratt:
I have squeezed in a seemingly unrelated commit (9029c40) whose absence otherwise ended up confusing me in this PR.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1979,6 +1979,10 @@ impl<'a> Assemble<'a> {
         let dest_type = self.m.type_(i.dest_tyidx());
         let dest_bitsize = dest_type.bit_size().unwrap();
 
+        if src_bitsize == 32 && dest_bitsize == 64 {
```
vext01, line 1982: Why is this safe?

Don't we need to zero the upper 32-bits?

ltratt: Hmm, yes, if we had a left shift before, I suppose we need to.

I suggest I remove this commit during a rebase.

vext01: It's more about what follows. If the `zext` instruction is followed by for example, and `add`, then the undefined bits could cause an incorrect result (if any bits are set up there).

> I suggest I remove this commit during a rebase.

OK

vext01:
Please squash and rebase.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ltratt: > It's more about what follows.

Yes, we agree.

ltratt:
Squashed.

PR #1443, opened 2024-10-29
Optimise `PtrAdd` followed by `Load`.

ltratt:
We commonly see this sort of pattern:

```
%148: ptr = ptr_add %147, 1416
%149: ptr = load %148
```

which we currently translate to something like `add; mov`.

This commit recognises that in x64 we can combine these two instructions iff the result of the `ptr_add` isn't used again, generating a single `mov` with an offset:

```
; %148: ptr = ptr_add %147, 1416
; %149: ptr = load %148
mov rdx, [rdx+0x588]
```

Needs https://github.com/ykjit/yk/pull/1442 to be merged first.


ltratt:
@vext01 This should now be mergeable.

PR #1447, opened 2024-10-31
Optimise `PtrAdd` followed by `Store`

ltratt:
This PR optimises `PtrAdd`s followed by `Store`s in the same way we do for `Load`s. Internally to the CPU these *probably* end up being converted to the same thing as before, but they allow us to avoid clobbering a register, which can be beneficial in some cases.


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1215,6 +1228,30 @@ impl<'a> Assemble<'a> {
         }
     }
 
+    /// Optimise `PtrAdd` followed by `Load`. This has the following preconditions:
```
vext01, line 1218: a few instances of `load` about that need to be updated to `store`.

vext01:
Thanks. Please squash.

ltratt:
Squashed.

ltratt:
Force pushed Clippy fixes.

PR #1451, opened 2024-11-02
Add support for executing GDB commands.

Pavel-Durov:
Added `gdb_args` to capture all arguments provided after `--`, allowing multiple GDB-specific arguments.

Usage example:
```shell
./bin/gdb_c_test -s -n simple.c -- --command=comms.gdb
```

### DIFF
--- a/tests/src/bin/gdb_c_test.rs
+++ b/tests/src/bin/gdb_c_test.rs
@@ -32,6 +32,10 @@ struct Args {
     /// Don't immediately run the program.
     #[arg(short = 'n', long)]
     wait_at_prompt: bool,
+
+    /// Pass all arguments after `--` directly to GDB.
+    #[arg(last = true, required = false)]
+    gdb_args: Vec<String>,
 }
 
 fn main() {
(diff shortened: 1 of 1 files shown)


ltratt:
Should this perhaps be `--args` to reflect the same thing in gdb itself?

Pavel-Durov:
> Should this perhaps be `--args` to reflect the same thing in gdb itself?

Not sure I understand. Do you mean to use  `--args` in `gdb_c_test` as a container for any arguments passed to the `gdb`?

To clarify:
--command: Used to execute a series of GDB commands from a specified file
--args: Intended for passing arguments directly to the program being debugged

ltratt:
Ah, I misunderstood `--command`. I guess what I'm suggesting is `gdb_c_test --args "blah blah"` where "blah blah" could include `--command`. Or -- maybe even better? -- pass all args after `--` to gdb. So `gdb_c_test <gdb_c_test_args> -- <gdb_args>`.

vext01:
> pass all args after -- to gdb

This makes a lot of sense to me. This way we wouldn't need to special case each gdb switch that we wish to support.

What do you think @Pavel-Durov ?

Pavel-Durov:
> > pass all args after -- to gdb
> 
> This makes a lot of sense to me. This way we wouldn't need to special case each gdb switch that we wish to support.
> 
> What do you think @Pavel-Durov ?

yeah sounds good :)

Pavel-Durov:
Updated with `--` :point_right: https://github.com/ykjit/yk/pull/1451/commits/b5ec224b0c6c43dfc6f47676b835356e9cf30f00

ltratt:
Please squash.

Pavel-Durov:
Added

ltratt:
Good idea! Please squash.

PR #1452, opened 2024-11-04
Assert that decopying isn't needed here.

ltratt:
This is belt and braces, because I'm about 99.9% sure that this condition can never be violated, but were it to be, all sorts of hard-to-debug things would occur.


vext01:
I'm happy with this. I just need to debug what's currently blocking CI.

ltratt:
FWIW, yk's CI isn't blocked, so you can merge this PR if you want.

Issue #1440, opened 2024-10-29
Queens benchmark fails


vext01:
Today's master (e9db448) causes the Queens benchmark to fail like this:

```
Executing run:Queens (YkLua, awfy, 1000) default None None None
    cmd: /bm/yklua/src/yklua  harness.lua Queens 10  1000
    cwd: /bm/suites/awfy/Lua
    Run failed. Return code: 134
    Output:

        Starting Queens benchmark ...
        thread '<unnamed>' panicked at ykrt/src/compile/jitc_yk/codegen/x64/mod.rs:2219:18:
        Unexpected HotLocationKind
        note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

ltratt:
Fixed by https://github.com/ykjit/yk/pull/1461.

PR #1454, opened 2024-11-04
Assert that all promotion data has been consumed.

ltratt:
One thing that I realised is that if we didn't consume all promotion data -- e.g. because we forgot to consume it when we're outlining -- weird things would happen. This assert always succeeds on my tests, but I think it's worth having in.

### DIFF
--- a/ykrt/src/compile/jitc_yk/trace_builder.rs
+++ b/ykrt/src/compile/jitc_yk/trace_builder.rs
@@ -138,6 +138,28 @@ impl TraceBuilder {
         Ok(())
     }
 
+    /// Process promotions inside an otherwise outlined block.
+    fn process_promotions_only(&mut self, bid: &aot_ir::BBlockId) -> Result<(), CompilationError> {
+        let blk = self.aot_mod.bblock(bid);
+
+        for inst in blk.insts.iter() {
+            if let aot_ir::Inst::Promote {
+                val: aot_ir::Operand::LocalVariable(_),
+                tyidx,
+                ..
+            } = inst
+            {
+                let width_bits = match self.aot_mod.type_(*tyidx) {
+                    aot_ir::Ty::Integer(it) => it.num_bits(),
+                    _ => unreachable!(),
+                };
+                let width_bytes = usize::try_from(width_bits.div_ceil(8)).unwrap();
+                self.promote_idx += width_bytes;
+            }
+        }
+        Ok(())
+    }
+
     /// Walk over a traced AOT block, translating the constituent instructions into the JIT module.
     fn process_block(
         &mut self,
@@ -1242,7 +1264,28 @@ impl TraceBuilder {
                             // blocks normally.
                             self.outline_target_blk = None;
                         } else {
-                            // We are outlining so just skip this block.
+                            // We are outlining so just skip this block. However, we still need to
+                            // process promoted values to make sure we've processed all promotion
+                            // data and haven't messed up the mapping.
+                            #[cfg(tracer_hwt)]
+                            {
+                                last_blk_is_return = self.aot_mod.bblock(&bid).is_return();
+                                // Due to hardware tracing we see the same block twice whenever
+                                // there is a call. We only need to process one of them. We can
+                                // skip the block if:
+                                //  a) The previous block had a return.
+                                //  b) The previous block is unmappable and the current block isn't
+                                //  an entry block.
+                                if last_blk_is_return {
+                                    prev_bid = Some(bid);
+                                    continue;
+                                }
+                                if prev_bid.is_none() && !bid.is_entry() {
+                                    prev_bid = Some(bid);
+                                    continue;
+                                }
+                            }
+                            self.process_promotions_only(&bid)?;
                             prev_bid = Some(bid);
                             continue;
                         }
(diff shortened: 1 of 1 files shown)


vext01:
Good idea!

ltratt:
This is interesting:

```
16:49:40 .thread '<unnamed>' panicked at ykrt/src/compile/jitc_yk/trace_builder.rs:1299:9:
16:49:40 assertion `left == right` failed
16:49:40 left: 48
16:49:40 right: 60
16:49:40 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
16:49:40 thread '<unnamed>' panicked at ykrt/src/mt.rs:337:46:
```

That suggests we already have a test case where we aren't consuming all promotions as we would expect to.

@ptersilie I think this is one for you when you're back.

ptersilie:
I might have an idea what's going on here, though I haven't confirmed this yet. From what I understand, we collect a list of all constants promoted during tracing and then this check makes sure we use all promotions when constructing the trace. What I believe this is forgetting is that we don't process everything we see during tracing. For example if there's a function containing a loop that also contains promotions, we will later outline that function and thus not process those promotions during trace building.

I will try constructing a test using this information to confirm.

ltratt:
> For example if there's a function containing a loop that also contains promotions, we will later outline that function and thus not process those promotions during trace building.

Right, this is my guess too: but *notice* that we should still be advancing `promotion_idx` in this case!

So I think the assert should always be true, because what it says is "we got to the end of the promotion data" (which, I now realise, isn't quite the same as "we consumed all the promotion data" implies).

ltratt:
I must admit that I'm slightly surprised we do get a `yk_promote` in an outline function: it might be worth checking *how* this happens. A first thought is that this *might* be because in lua we get a function call which calls an outlined function (which one?) that then recalls the main interpreter loop. If that *is* the case, it means we're not inlining Lua-level functions, which would be less than ideal in yklua.

So one way of thinking of this in the short term is to say "if we haven't got to the end of the promotion data, the user should change their interpreter to make sure `yk_promote` isn't (directly or indirectly) contained in an outlinable function".

ltratt:
This LGTM but I think @vext01 needs to take over reviewing now.

ltratt:
And I suggest before merge we squash the two commits into one, so we don't leave a "this breaks" commit in history. Lukas can do a `co-authored-by`!

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -138,6 +138,29 @@ impl TraceBuilder {
...
+
+        for inst in blk.insts.iter() {
+            match inst {
+                aot_ir::Inst::Promote { val, tyidx, .. } => match val {
+                    aot_ir::Operand::LocalVariable(_) => {
+                        let width_bits = match self.aot_mod.type_(*tyidx) {
+                            aot_ir::Ty::Integer(it) => it.num_bits(),
+                            _ => unreachable!(),
+                        };
+                        let width_bytes = usize::try_from(width_bits).unwrap() / 8;
```
vext01, line 153: Oopsy. An `i1` would report a byte size of 0! I know we can't promote an `i1`, but still.

See this [jit ir routine](https://github.com/ykjit/yk/blob/3179eb557cb41201ba3b8a3ed12b9afc48fe5c5d/ykrt/src/compile/jitc_yk/jit_ir/mod.rs#L1001) for inspiration.

Every time we are tempted to divide a bit size by 8, an alarm bell should ring :)

vext01:
Seems good, modulo that one comment.

ykrt/src/compile/jitc_yk/trace_builder.rs
ptersilie: I just copied this from `handle_operand`. We should fix it there too, but I'm not sure this PR is the right place.

ltratt:
Fixes https://github.com/ykjit/yk/issues/1440.

ykrt/src/compile/jitc_yk/trace_builder.rs
vext01: OK, please follow this PR up with the fix. It'll surely bite us later (again).

ptersilie:
I need to squash first.

ptersilie:
Squashed.

vext01:
go ahead.

ptersilie:
Squashed.

vext01:
You can force push the clippy fix.

ptersilie:
Ready.

ptersilie:
> Unexpected HotLocationKind

Huh. I thought I fixed that.

ltratt:
@ptersilie I thought we'd got rid of this code:

```rust
        let root_ctr = match &hl.kind {
            HotLocationKind::Compiled(root_ctr) => Arc::clone(root_ctr)
                .as_any()
                .downcast::<X64CompiledTrace>()
                .unwrap(),
            HotLocationKind::SideTracing { root_ctr, .. } => Arc::clone(root_ctr)
                .as_any()
                .downcast::<X64CompiledTrace>()
                .unwrap(),
            _ => panic!("Unexpected HotLocationKind"),
        };
```

but obviously we didn't! Do we even need it? There's no way of making this code correct AFAICS: it's broken.

ltratt:
After offline discussion we realised this is a bit of code duplication.

ltratt:
@vext01 If/when https://github.com/ykjit/yk/pull/1464 merges, this PR can hopefully be merged.

PR #1456, opened 2024-11-05
More careful handling of register handling around calls.

ltratt:
Register allocation around calls was -- entirely my fault! -- rather sloppy. This was partly a failure of imagination on my part: I couldn't really work out how to jam the complex requirements of x64 and the SysV ABI into the register allocation API. Instead, I relied on some implicit properties of the allocator and crossed my fingers.

This commit rewrites `emit_call` to do a much better job of register allocation. The basic insight is that we always have to deal with a fixed number of registers, even though "fixed number" is "quite a lot". That means that we can use the existing allocator API -- we just have to be slightly creative about it!

First we introduce `RegConstraint::Clobber` which specifies that an operation clobbers a given register. I don't know why I didn't add this earlier: doing so allows us to clear up some other messiness in the API and means we can get rid of the `hack` and `avoiding` methods entirely. For example `BinOp::Mul` goes from:

```rust
self.ra
    .clobber_gp_regs_hack(&mut self.asm, iidx, &[Rq::RDX]);
let [_lhs_reg, rhs_reg] = self.ra.assign_gp_regs_avoiding(
    &mut self.asm,
    iidx,
    [
        RegConstraint::InputOutputIntoReg(lhs, Rq::RAX),
        RegConstraint::Input(rhs),
    ],
    RegSet::from(Rq::RDX),
);
```

to:

```rust
let [_lhs_reg, rhs_reg, _] = self.ra.assign_gp_regs(
    &mut self.asm,
    iidx,
    [
        RegConstraint::InputOutputIntoReg(lhs, Rq::RAX),
        RegConstraint::Input(rhs),
        RegConstraint::Clobber(Rq::RDX),
    ],
);
```

For calls, we basically start all GP and FP registers in the `Clobber` state then gradually refine them as we realise more about the specific call we have to generate code for. We use the "specific register" constraints to express this. Thus we always assign 10 GP registers (9 for the ABI and 1 for the call operand) and 16 FP registers.

This commit at the very least makes everything much tidier, but it also fixes some bugs. For example: we could overwrite input GP registers before; we didn't clobber XMM registers before; and probably some others. That doesn't mean that this commit is perfect: FP allocation is not as well done as GP allocation (i.e. spills/moves too often, but at least isn't obviously buggy to my eyes); we can certainly optimise away, in many cases, the 10th GP register; and probably more. But it is, in my opinion, a definite improvement on the status quo.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -67,12 +67,12 @@ mod deopt;
 mod lsregalloc;
 
 use deopt::{__yk_deopt, __yk_guardcheck, __yk_reenter_jit};
-use lsregalloc::{LSRegAlloc, RegConstraint, RegSet};
+use lsregalloc::{LSRegAlloc, RegConstraint};
 
 /// General purpose argument registers as defined by the x64 SysV ABI.
 static ARG_GP_REGS: [Rq; 6] = [Rq::RDI, Rq::RSI, Rq::RDX, Rq::RCX, Rq::R8, Rq::R9];
 
-/// The registers clobbered by a function call in the x64 SysV ABI.
+/// The GP registers clobbered by a function call in the x64 SysV ABI.
 static CALLER_CLOBBER_REGS: [Rq; 9] = [
     Rq::RAX,
     Rq::RCX,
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1424,97 +1438,105 @@ impl<'a> Assemble<'a> {
...
+                    self.ra
+                        .assign_gp_regs(&mut self.asm, iidx, gp_cnstrs.try_into().unwrap());
+
+                if fty.is_vararg() {
+                    dynasm!(self.asm; mov rax, num_float_args); // SysV x64 ABI
+                }
                 dynasm!(self.asm
-                    ; mov Rq(reg.code()), QWORD p as i64
-                    ; call Rq(reg.code())
+                    ; mov Rq(tmp_reg.code()), QWORD p as i64
```
vext01, line 1526: Side note: it's been on my todo list to optimise the way we do the actual call.

It should be possible to use one of the encodings which doesn't require using a register, but last time I tried (ages ago) it wasn't as simple as I thought.

vext01:
A lot to review, but conceptually I agree that this is the way to go.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ltratt: Agreed: I want this too. Also, when we can't do so, we often (but not always!) have a spare register such that we don't have to clobber another random register.

PR #1461, opened 2024-11-11
Don't require updating the running trace during side-tracing.

ptersilie:
In order to find the correct information during a guard failure we had
to switch the running trace whenever we jumped from a root trace into a
side-trace and back. Not only does this slow us down, it occasionally
lead to bugs when trying to find the root trace.

This commit no longer requires updating the running trace and instead
passes down a list of all the guards that have failed leading up to the
current guard failure. This list can then be used to find the correct
running trace and guard failure information by traversing down from the
root trace.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -66,7 +66,7 @@ use yksmp;
 mod deopt;
 mod lsregalloc;
 
-use deopt::{__yk_deopt, __yk_guardcheck, __yk_reenter_jit};
+use deopt::{__yk_deopt, __yk_guardcheck};
 use lsregalloc::{LSRegAlloc, RegConstraint};
 
 /// General purpose argument registers as defined by the x64 SysV ABI.
@@ -160,8 +160,9 @@ impl CodeGen for X64CodeGen {
         hl: Arc<Mutex<HotLocation>>,
         sp_offset: Option<usize>,
         root_offset: Option<usize>,
+        prevguards: Option<Vec<GuardIdx>>,
     ) -> Result<Arc<dyn CompiledTrace>, CompilationError> {
-        Assemble::new(&m, sp_offset, root_offset)?.codegen(mt, hl, sp_offset.is_some())
+        Assemble::new(&m, sp_offset, root_offset)?.codegen(mt, hl, prevguards)
     }
 }
 
@@ -257,9 +258,9 @@ impl<'a> Assemble<'a> {
         mut self: Box<Self>,
         mt: Arc<MT>,
         hl: Arc<Mutex<HotLocation>>,
-        issidetrace: bool,
+        prevguards: Option<Vec<GuardIdx>>,
     ) -> Result<Arc<dyn CompiledTrace>, CompilationError> {
-        if issidetrace {
+        if prevguards.is_some() {
             // Recover registers.
             for reg in lsregalloc::FP_REGS.iter() {
                 dynasm!(self.asm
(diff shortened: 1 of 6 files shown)


ptersilie:
Surprisingly, I couldn't measure any significant speedup/slowdown with this change. But since this fixes a bug, it's worth merging anyway.

ptersilie:
Ok to squash?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -347,14 +348,45 @@ impl<'a> Assemble<'a> {
                 }
                 dynasm!(self.asm; mov rcx, rsp);
 
+                // Push [GuardIdx]'s of previous guard failures.
+                if let Some(ref guards) = prevguards {
```
ltratt, line 352: I might have misunderstood what's going on here, but isn't this producing at run-time an ever-growing list of guard IDs? Looked at another way, does this mean we get a gradual slow-down at run-time as you go deeper into the side-trace tree?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -347,14 +348,45 @@ impl<'a> Assemble<'a> {
...
+                // Save the pointer to this list.
+                dynasm!(self.asm
+                    ; mov r8, rsp
+                );
+
+                let len = i64::try_from(prevguards.as_ref().map_or(0, |v| v.len())).unwrap();
+                // Total alignment caused by pushing the parent guards.
+                let mut totalalign = len * 8;
+
+                // Pushing RSI above mis-aligned the stack. Unless we accidentally re-aligned it by
```
ltratt, line 369: How is it misaligned? Aren't the guard IDs 64 bits?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: Sort-of yes. But this only happens during guard failures, so I'm not sure how much a few extra pushes hurt here. How deeply nested to we expect side-traces to become?

ltratt: We can deal with it for now but this will become a signifcant overhead at some point, even for a single level of nesting, and we can regularly expect to have 3-5 levels deep at least I would expect.

What we'll need to do -- not in this PR! -- is work out how to make this a JIT-compile-time cost that has zero impact on the running trace.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: We need to be 16 byte aligned before calls and pushing RDI further above misaligns the stack.

ltratt: Ah! Can we clarify that it's 16 bytes please (I incorrectly assumed 8).

ykrt/src/compile/guard.rs
```diff
@@ -80,3 +80,9 @@ impl From<GuardIdx> for usize {
         v.0
     }
 }
+
+impl From<GuardIdx> for i64 {
+    fn from(v: GuardIdx) -> Self {
+        i64::try_from(v.0).unwrap()
```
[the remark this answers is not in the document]
ptersilie: [7a5b78a](https://github.com/ykjit/yk/pull/1461/commits/7a5b78adf69b2af3466fc60d1583b4d084c9e1ff)

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: [ffda13e](https://github.com/ykjit/yk/pull/1461/commits/ffda13e4e44b0333b166d6da8a1e0e7afa6d12a1)

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -351,8 +351,9 @@ impl<'a> Assemble<'a> {
                 // Push [GuardIdx]'s of previous guard failures.
                 if let Some(ref guards) = prevguards {
                     for gidx in guards.iter().rev() {
+                        let g = i64::try_from(<GuardIdx as Into<usize>>::into(*gidx)).unwrap();
```
ltratt, line 354: Is this `i64::try_from(usize::from(gidx)).unwrap()` in disguise? If so, please force push an update.

ptersilie: It was. Don't ask me how this happened. :laughing:

ptersilie:
Squashed.

ptersilie:
Every time I run clippy it has nothing to say, and the one time I forget to run it it does. Sorry. Fixed and force-pushed. Please try merging again.

ptersilie:
Ah no, hold on. Some tests need fixing too.

ltratt:
Please force push an update (notice also you'll need to rebase away the ykllvm change) when you're ready.

ptersilie:
Saw you comment too late. Rebasing now.

ptersilie:
All done.

PR #1462, opened 2024-11-11
Implement the fneg instruction.

vext01:
Required for LuaJIT tests.

Please carefully check the register allocator changes. I had to implement `RegConstraint::Temporary` in `assign_fp_regs()`.

Requires: https://github.com/ykjit/ykllvm/pull/214


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -4329,4 +4367,28 @@ mod tests {
...
+            "
+                ...
+                ; %2: float = fneg %0
+                {{_}} {{_}}: mov r.32.x, 0x80000000
+                {{_}} {{_}}: movd fp.128.y, r.32.x
+                {{_}} {{_}}: xorps fp.128.z, fp.128.y
+                ; %3: double = fneg %1
+                {{_}} {{_}}: mov r.64.x, 0x8000000000000000
+                {{_}} {{_}}: movq fp.128.y, r.64.x
+                {{_}} {{_}}: xorpd fp.128.a, fp.128.y
```
vext01, line 4390: I was curious that `%3` was allocated a fresh register (`fp.128.a`), instead of reusing the register previously assigned to the now-dead `%2` (`fp.128.z`). Could indicate an error in my register allocator changes?

ltratt:
LGTM.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ltratt: We currently prefer to allocate "higher" registers, which is pointless in FP-land, but in GP-land means we're less likely to get twonked by x64's fondness for RAX and the sysv abi's preference for "lower" registers in CALLs. Does that *guarantee* what you're seeing is intended? Nope. But it might give you a clue to what look for.

vext01:
LLVM synced. OK to squash?

vext01:
squashed.

vext01:
Once https://github.com/ykjit/yk/pull/1463 is in, this should merge. It's the ambiguous methods warning again.

vext01:
Not yet :)

PR #1467, opened 2024-11-18
Optimise code gen of `Shl`.

ltratt:
This shaves another couple of percentage points off big_loop.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -755,9 +755,11 @@ impl<'a> Assemble<'a> {
                 }
             }
             BinOp::AShr | BinOp::LShr | BinOp::Shl => {
-                // We inherit from LLVM the following semantics: a poison value is computed if you
-                // shift by >= the bit width of the first operand. We can ignore this, since we are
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -794,6 +794,19 @@ impl<'a> Assemble<'a> {
                             }
                             _ => todo!(),
                         },
+                        BinOp::Shl => match bit_size {
+                            0 => unreachable!(),
+                            32 => dynasm!(self.asm; shl Rd(lhs_reg.code()), v),
+                            1..=64 => {
+                                // Ensure we shift in zeros at the most-significant bits.
```
vext01, line 803: Unless I'm mistaken, I don't think you need any sign or zero extend stuff here. `shl` will always shift in zeros at the LSB and the undefined bits towards the MSB are irrelevant, right?

ltratt: Oops, you're right -- I wasn't thinking! Fixed in ec22a96.

vext01:
squash away.

ltratt:
Squashed.

PR #1468, opened 2024-11-18
Skip intermediate `PtrAdd`s.

ltratt:
A common pattern we see is:

```
%7: ptr = ptr_add %0, 2
%8: ptr = ptr_add %7, 3
```

This can be annoying because `%7` sometimes isn't really used other than as a vehicle for `%8`. This commit rewrites `%8` to:

```
%8: ptr = ptr_add %0, 5
```

This can reduce the pressure on the register allocator a little bit. Has a small (0.5%), but measurable effect on big_loop.


ltratt:
I think this one has failed for spurious reasons, but I may have had a cunning plan to extend it, so let's not merge it quite yet.

ltratt:
Closing in favour of an upcoming PR.

ltratt:
Oops, yes!

PR #1470, opened 2024-11-19
Support calls which take and return floats/double.

ltratt:


vext01:
Thanks for this.

ltratt:
Force pushed a formatting fix (mea culpa).

ltratt:
This is failing for the same reason @vext01 saw the nightly fail. We can try merging again and we might get lucky or we can wait until Edd fixes things.

PR #1471, opened 2024-11-21
Make the path specification of YKD_LOG_IR optional.

vext01:
Log to stderr if not specified, the same as YK_LOG.

### DIFF
--- a/ykrt/src/log/mod.rs
+++ b/ykrt/src/log/mod.rs
@@ -122,27 +122,25 @@ mod internals {
     use super::IRPhase;
     use std::{collections::HashSet, env, error::Error, fs::File, io::Write, sync::LazyLock};
 
-    // YKD_LOG_IR
-
     static LOG_IR: LazyLock<Option<(String, HashSet<IRPhase>)>> = LazyLock::new(|| {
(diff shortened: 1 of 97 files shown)


ykrt/src/log/mod.rs
```diff
@@ -122,27 +122,25 @@ mod internals {
...
-                        // appends to the log aren't appending to a previous log run.
-                        File::create(p).ok();
-                    }
-                    Some((p.to_string(), log_phases))
-                }
+            let (path, phases) = match x.split(':').collect::<Vec<_>>().as_slice() {
+                [path, phases] => (*path, *phases),
+                [phases] => ("-", *phases),
                 _ => panic!(
                     "YKD_LOG_IR must be of the format '<path>:<irstage_1>[,...,<irstage_n>]'"
```
ltratt, line 132: I think it should be `[<path>]:...`? If so, please force push an update.

vext01: Like YK_LOG the colon is also optional, so I think this is correct.

But I did notice that the panic needs updating, so force push a fix for that.

vext01: Yep. The force push fixed it.

PR #1474, opened 2024-11-22
Add offsets to loads and stores

ltratt:
The main commit here is https://github.com/ykjit/yk/commit/04c879d5c6537886917187ce5f79935f10b89952 which (hopefully) extensively documents the motivation and consequences of this PR. I do consider this PR an RFC of sorts, though I definitely come down on the side of "the pros outweigh the cons", though I'm open to arguments the other way!

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -460,47 +460,17 @@ impl<'a> Assemble<'a> {
 
             match inst {
                 #[cfg(test)]
-                jit_ir::Inst::BlackBox(_) => unreachable!(),
+                jit_ir::Inst::BlackBox(_) => (),
                 jit_ir::Inst::Const(_) | jit_ir::Inst::Copy(_) | jit_ir::Inst::Tombstone => {
                     unreachable!();
                 }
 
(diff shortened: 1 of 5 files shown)


vext01:
This all seems sensible to me.

If we are to allow instructions to be larger than a word, I wonder if we'd still strive to pack operands? I recall you saying that the indices are likely to be too small for long-term use, and if they were full-on usizes then we could get rid of a load of `try_from`s.

Note a typo in the PR title.

ltratt:
>  If we are to allow instructions to be larger than a word, I wonder if we'd still strive to pack operands

An (accidental but nice) advantage of `PackedOperand` is that we de-copy it automatically when converting it to `Operand` i.e. we make it harder to misuse `Operand`. But, yes, I would probably like to make `PackedOperand` a `u24` shortly.

> Note a typo in the PR title.

Ooops, fixed!

ptersilie:
Looks good. Are there now other instructions that now could be simplified. `DirectCall` comes to mind, but I think due to the arbitrary arguments it's unlikely to fit into two words. But there may be others that now can?

ltratt:
Please don't merge this: I now think we can do the same thing, but better, with reverse code generation. I need to think about it a bit more.

ltratt:
This is now superseded by https://github.com/ykjit/yk/pull/1485. Closing.

PR #1475, opened 2024-11-25
Fix a side-tracing + hardware tracing bug.

vext01:
When we are constructing a side trace we will have first deopted back to the conditional branch that caused the guard to fail and side-tracing to start. Because the conditional branch is re-executed, this means that when we are doing hardware tracing, we will see a spurious block at the beginning of the trace that isn't actually part of the trace. We therefore skip it.

Fixes crashes like this, where we get a local_map miss for local variables relating to the conditional branch:

thread '<unnamed>' panicked at ykrt/src/compile/jitc_yk/trace_builder.rs:404:69: no entry found for key

### DIFF
--- a/ykrt/src/compile/jitc_yk/trace_builder.rs
+++ b/ykrt/src/compile/jitc_yk/trace_builder.rs
@@ -1202,29 +1202,53 @@ impl TraceBuilder {
                 })
                 .collect::<Vec<_>>();
 
-            #[cfg(tracer_swt)]
+            // When side-tracing a switch guard failure, we need to reprocess the switch statement
+            // (and only the switch statement) in order to emit a guard at the beginning of the
+            // side-trace to check that the case requiring execution is that case the trace
+            // captures.
+            //
+            // Note that it is not necessary to emit such a guard into side-traces stemming from
+            // regular conditionals, since a conditional has only two sucessors. The parent trace
+            // captures one, so by construction the side trace must capture the other.
+            let prevbb = self.aot_mod.bblock(prev_bid.as_ref().unwrap());
+            if let aot_ir::Inst::Switch {
+                test_val,
+                default_dest,
+                case_values,
+                case_dests,
+                safepoint,
+            } = &prevbb.insts.last().unwrap()
             {
-                // FIXME: This is a hack! When we side-trace the guard failure of a switch
-                // statement, the software-tracer doesn't report the switch-statement block as the
-                // first block in the trace. This means we don't generate a guard at the top of the
-                // side-trace checking that the switch-case is correct when executing the trace. We
-                // work around this force processing the previous block that's passed in via
-                // `SideTraceInfo` if it's last instruction is a switch statement. Technically,
-                // this is the correct fix for hardware-tracing too, since we shouldn't re-process
-                // the switch-statement block, as it was already executed in the parent trace,
-                // which is problematic if the block has side-effects.
-                let prevbb = self.aot_mod.bblock(prev_bid.as_ref().unwrap());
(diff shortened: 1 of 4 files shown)


vext01:
There is still one other non-deterministic bug lurking, but it's quite rare, and I hope this is enough to clear the PR backlog.

Once all the PRs are in I can try to repro the other bug. If I'm lucky, something fixes it already ! :hand_with_index_finger_and_thumb_crossed:

ltratt:
Please force push a fix.

vext01:
Force pushed the fix. It's not pretty.

vext01:
Turns out we didn't understand the problem entirely in our last attempt. [9c3613d](https://github.com/ykjit/yk/pull/1475/commits/9c3613df773b9a27058046d66f0a4b83e75a663d) is (we believe) a correct fix.

I've cherry picked the commit from https://github.com/ykjit/yk/pull/1473 into this branch, because I'm seeing that problem crop up more and more and it's unlikely that my our changes will merge without it.

I've done a fairly large benchmark run, run unit tests and lua tests and haven't had any issues. Fingers crossed.

If you are OK with this, then I'd like to squash, reorder the commits and reword the commit message before merge.

vext01:
@ptersilie you may want to double-check the comments I've revised since you last saw this.

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -1202,29 +1202,55 @@ impl TraceBuilder {
...
+                    case_dests,
+                )?;
             }
         }
 
+        // When `prev_bid` is not None then we are side tracing from a guard in the block
+        // identified by `prev_bid`. The way that we deopt back to (and re-execute) the conditional
+        // that caused the deopt, combined with the way that hardware tracing works, means that we
+        // can see duplicated `prev_bid` entries at the start of the trace. When we side-trace, we
+        // don't want those duplicates in the trace, hence we strip them.
```
ptersilie, line 1243: Suggestion:
```
The variable `prev_bid` contains the block of the guard that initiated side-tracing (for normal traces this is set to `None`). When hardware tracing, we capture this block again as part of the side-trace. However, since we've already processed this block in the parent trace, we must not process it again in the side-trace.

Typically, the mapper would strip this block for us, but for codegen related reasons, e.g. a switch statement codegenning to many machine blocks, it's possible for multiple duplicates of this same block to show up here, which we all need to skip.
```

vext01: Fixed in 7a19aba21340

ykrt/src/trace/hwt/mapper.rs
```diff
@@ -169,7 +169,18 @@ impl Iterator for HWTTraceIterator {
...
+            //
+            // If we are collecting a top-level trace, this removes the remainder of the block
+            // containing the control point.
+            //
+            // If we are side-tracing then this attempts to remove the re-executed conditional that
+            // caused the guard failure and in turn the side trace to be collected. Note though,
+            // that some conditionals (e.g. switches) span multiple machine blocks, so this may
+            // remove only part of the conditional. Unfortunately we don't have enough information
+            // at this level to kill all of the conditional, so there's an additional purge loop in
+            // the trace builder.
```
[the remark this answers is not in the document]
vext01: Fixed in ad61c8e1b1e3

ltratt:
I'm fine with this: as soon as Lukas is happy with the updated comments, please squash.

vext01:
squashed and rebased.

vext01:
Can we retry this? That's a different bug.

PR #1477, opened 2024-11-25
swt clone module function references test

Pavel-Durov:
Add test to validate https://github.com/ykjit/ykllvm/pull/216.

### DIFF
--- a/tests/c/swt_module_clone.c
+++ b/tests/c/swt_module_clone.c
@@ -1,0 +1,83 @@
+// ignore-if: test "$YKB_TRACER" != "swt"
+// Run-time:
+//   env-var: YKD_LOG_IR=aot,jit-post-opt,jit-pre-opt
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YK_LOG=4
+//   stderr:
+//     4
+//     --- Begin aot ---
+//     ...
+//     #[yk_outline]
+//     func add(%arg0: i32, %arg1: i32) -> i32;
+//     ...
+//     #[yk_outline]
+//     func dec(%arg0: i32) -> i32;
+//     ...
+//     #[yk_outline]
+//     func main(%arg0: i32, %arg1: ptr) -> i32 {
+//     ...
+//     *%{{dec_addr}} = dec
+//     ...
+//     %{{dec_ptr}}: ptr = load %{{dec_addr}}
+//     %{{i}}: i32 = load %{{_}}
+//     %{{_}}: i32 = icall %{{dec_ptr}}(%{{i}})
+//     ...
+//     --- End aot ---
+//     --- Begin jit-pre-opt ---
+//     ...
+//     %{{_}}: i32 = icall %{{_}}(%{{_}})
+//     ...
+//     --- End jit-pre-opt ---
+//     --- Begin jit-post-opt ---
+//     ...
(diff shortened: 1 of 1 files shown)


Pavel-Durov:
updated ykllvm submodule with the latest ykllvm version since https://github.com/ykjit/ykllvm/pull/216 was merged.

tests/c/swt_module_clone.c
```diff
@@ -0,0 +1,49 @@
...
+// be visible in the aot ir since they are not serialised.
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+int add(int i, int j) { return i + j; }
```
ltratt, line 20: These should probably be `__attribute__((yk_outline))`: at the moment I would expect that these are inlined, so the test will pass "by accident" (TM). Also, the `stderr` lang_test should check that the IR actually looks like we expect. Specifically, we should remove `jit-pre-opt` from `YKD_LOG_IR` but we should actually check that the functions, with the names we expect, are present in the AOT IR.

Pavel-Durov: OK, will update.
But there's a problem with adding IR checks here because we don't serialize the cloned/optimized functions at all.

ltratt: That's fine: if the AOT IR should only refer to the unoptimised functions, then that's our test!

Pavel-Durov: Updated :point_right: 6bf7c04fdd10dd811dfcda75e03acd1bcc1cbc1e and :point_right: 0dedb052da113a0e39ac097888b4de25acd14afb

ltratt: Should there be a `call dec` in the trace IR too?

Pavel-Durov: How do I get the trace IR?

ltratt: It depends which you want: `aot`, `jit-pre-opt`, or `jit-post-opt`.

Pavel-Durov: ok, got it.
I got confused cause originally we said to remove `jit-pre-opt` from YKD_LOG_IR

Pavel-Durov: > Should there be a call dec in the trace IR too?

Added trace checks :point_right: d4b95bfebbf30426952bc5db2e75be949748c92e

ltratt: Can we check the call is to `dec`? I'm not totally sure off-hand if that is possible or not.

ltratt:
@Pavel-Durov Is this PR still relevant? I'm a bit unsure of its status.

Pavel-Durov:
> @Pavel-Durov Is this PR still relevant? I'm a bit unsure of its status.

Oops, completely forgot about this one.
Test updated :point_right: d105c8337dd1c3c5bc415bbfd04af253195834d9

tests/c/swt_module_clone.c
Pavel-Durov: Updated :point_right: 77c14ccdb0949e8e79dade44a34c027747f4a360

ltratt:
Please squash.

Pavel-Durov:
Squashed :point_right: 370f4d915430c7e5ef063686992fe1b51e33a019

PR #1478, opened 2024-11-26
Somewhat fix loop jump operands.

ltratt:
Previously we called these "loop jump variables", but they're not -- we can pass `Operand::Const`s!

This PR moves us further in the direction of "store `PackedOperand`s, use `Operand`s". [There is still Module::loop_start_vars which is `Vec<Operand>`, but I haven't fully thought that through yet.]

This PR fixes a bug in the way we represented big_loop.lua: we optimised away a variable (%81) since it was a constant (0) but erroneously printed out the variable in `tloop_jump` instead of the constant.

Unfortunately as part of this I've realised that `write_jump_vars` is very broken. It only happens to work right now by accident and I'm pretty sure one can create situations where it fails in confusing ways. Fixing that isn't going to be trivial, and the change in this PR is enough of an improvement on the status quo that I think it's worth merging, even though we know more work like this needs to be done.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/reg_alloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/reg_alloc.rs
@@ -31,6 +31,8 @@ pub(crate) enum VarLocation {
     ConstInt { bits: u32, v: u64 },
     /// A constant float.
     ConstFloat(f64),
+    /// A constant pointer.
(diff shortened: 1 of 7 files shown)


ltratt:
Force pushed a rebase against master.

ltratt:
I *think* this is one of our unrelated bugs: maybe retry?

PR #1479, opened 2024-11-29
Split root and sidetrace compilation

ltratt:
Previously we muddled together the compilation of root traces and sidetraces, in a way that allowed other bugs to creep in because of all the if/elseing. This PR gradually teases those two things apart (chiefly https://github.com/ykjit/yk/commit/ae805f8ca32f7746fdadf434604023fefc5c5461) and fixes some bugs that become apparent as a result (chiefly https://github.com/ykjit/yk/commit/b45deec7f78eaf8663e963ea7370c0066a6521a3).

There is more we could do here, but this is, I think, already a meaningful improvement.

### DIFF
--- a/ykrt/src/compile/jitc_yk/mod.rs
+++ b/ykrt/src/compile/jitc_yk/mod.rs
@@ -102,9 +102,8 @@ impl JITCYk {
             codegen: codegen::default_codegen()?,
         }))
     }
-}
 
-impl Compiler for JITCYk {
+    // FIXME: This should probably be split into separate root / sidetrace functions.
     fn compile(
         &self,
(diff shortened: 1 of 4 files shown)


ptersilie:
> This commit fixes that, but does highlight the fact that we have no way of tracking how often a guard has failed to compile a trace.

The easiest fix for this would be to add a counter to the `Guard` struct, I imagine.

ltratt:
> The easiest fix for this would be to add a counter to the Guard struct, I imagine.

That sound plausible. I will look at that in a future PR perhaps.

ptersilie:
This looks all good to me. Should I just merge?

PR #1480, opened 2024-11-29
Fix cargo-deny.

ltratt:
We no longer encounter this license.


ltratt:
Force pushed a fix (I needed to `cargo update` to see the same failure as BB).

PR #1481, opened 2024-11-29
Control point control flow

ltratt:
This *DRAFT* PR shows two problems we have right now:

1. The interpreter loop calling itself (i.e. for language-level calls).
2. The interpreter loop starting tracing and finishing before tracing completes.

Both tests bork on master.

I have put in what looks like a fix `yk_mt_early_return`, but it leads to a segfault in Rust code in `early_return.c` (run the test under gdb and you'll see it crashes at a completely innocuous looking place in Rust). This *might* be a minimal test case of the problems that have recently plagued yklua: certainly, these two tests are things that yklua can do, and that we don't handle properly.

Thoughts welcome!


ltratt:
Here's something I can't immediately explain. Look at this trace (from `recursive.c`):

```
%0: ptr = load_ti Register(13, 8, 0, [])
%1: ptr = load_ti Register(14, 8, 0, [])
%2: ptr = load_ti Register(12, 8, 0, [])
%3: ptr = load_ti Register(3, 8, 0, [])
%4: ptr = load_ti Register(5, 8, 0, [])
%5: ptr = load_ti Register(4, 8, 0, [])
tloop_start [%0, %1, %2, %3, %4, %5]:
%7: i32 = load %2
%8: i1 = sgt %7, 2i32
guard false, %8, [0:%0_3: %0, 0:%0_4: %1, 0:%0_5: %2, 0:%0_6: %3, 0:%6_1: 1i1]
%10: ptr = lookup_global @stderr
%11: ptr = load %10
%12: i32 = load %2
%13: ptr = lookup_global @.str
%14: i32 = call @fprintf(%11, %13, %12)
%15: i32 = load %2
%16: i32 = add %15, 4294967295i32
*%2 = %16
%18: i32 = load %2
%19: i1 = sgt %18, 0i32
guard true, %19, [0:%0_3: %0, 0:%0_4: %1, 0:%0_5: %2, 0:%0_6: %3, 0:%4_1: 0i1]
%21: ptr = load %0
%22: ptr = load %1
tloop_jump [%0, %1, %2, %3, %21, %22]:
```

Notice that `%4` and `%5` are inputs, but they are not used in the trace or a guard. Even weirder at the loop point we pass *different* values (`%21` and `%22`) for them, which are `load`s, but again those can have no effect.

Obvious question: should `%4` and `%5` be optimised away? Or are they doing something important and we've got a bug somewhere which isn't capturing that thing properly?

ptersilie:
The trace inputs are taken verbatim from the AOT stackmap. Are you sure `%4` and `%5` aren't optimised away by our JIT optimiser. Is this output `pre-opt` or `post-opt`?

ltratt:
From memory, you can see this in `pre-opt` (and `post-opt`, but that doesn't matter).

ykrt/src/mt.rs
```diff
@@ -578,16 +591,31 @@ impl MT {
...
-                                        // tracing the loop.
-                                        lk.kind = HotLocationKind::Compiling;
-                                        TransitionControlPoint::StopTracing
+                                        // ...and it's this location...
+                                        if frameaddr == *tracing_frameaddr {
+                                            lk.kind = HotLocationKind::Compiling;
+                                            TransitionControlPoint::StopTracing
+                                        } else {
+                                            // ...but in a different frame.
+                                            #[cfg(target_arch = "x86_64")]
```
vext01, line 610: Nitpick: Here we could check for `StackDirection::GrowsDown` (defined in `regalloc.rs`).

ykrt/src/mt.rs
```diff
@@ -748,13 +781,48 @@ impl MT {
...
+            }
+
+            if abort {
+                match mtt.tstate.replace(MTThreadState::Interpreting) {
+                    MTThreadState::Tracing {
+                        thread_tracer, hl, ..
+                    } => {
+                        // We don't care if the thread tracer went wrong: we're not going to use
+                        // its result anyway.
+                        thread_tracer.stop().unwrap();
```
vext01, line 813: Following from the comment above, I guess the `unwrap()` should be an `ok()`?

vext01:
OK, as far as recursive interpreter loops go, I think I understand the strategy: we never permit tracing the exit of an interpreter loop. Correct?

I added a test for a scenario that I was curious about in https://github.com/ykjit/yk/pull/1481/commits/a1e927f390f300c6ebe2acc21b9037684a15c572. Does this work as expected @ltratt -- I think so.

ltratt:
I think so.

ptersilie:
I think this is mostly ready now to be turned into a proper PR. Are you taking back over @ltratt or should I be in charge of this PR now?

ltratt:
I think you or Edd should take over now.

vext01:
Bear with me, there, this needs a little more work.

ltratt:
Closing in favour of #1513.

PR #1482, opened 2024-12-01
Derive debugs in various places.

ltratt:
When hunting a bug in `mt.rs`, I found that I needed some missing `derive(Debug)`s. I suspect we will need those again, so this commit adds them in for whoever might need them in the future.


vext01:
Lgtm. Did you mean to raise it as a draft?

ltratt:
Oops, no! Fixed.

ltratt:
Force pushed a fix for the swt case.

PR #1483, opened 2024-12-01
"Traceinput" to "Parameter"

ltratt:
While looking at https://github.com/ykjit/yk/pull/1481, and then looking inside yk, I was reminded anew how misleading "loadtraceinput" is: in particular, we don't load anything! In particular I think this has misled has several times when passing arguments between traces.

This PR is purely cosmetic and, bit by bit, renames "loadtraceinput" (and its variations) to "parameter" (or -- once I realised that's a long word that I tend to misspell -- `Param`). I believe this better represents what we're actually dealing with: it will make it clearer to us what are parameters (things going into a trace) and arguments (which we pass from a parent trace to a side trace).


vext01:
I'm neutral on this. I think "parameter" and "argument" are fine, since the trace is a function (which is literally called). I think "input" (without the load) is also fine.

ltratt:
The problem with "input" is that the way we tend to use it conflates the two concepts: the syntatic thing you write (parameter) and the run-time value (argument). "Input" also doesn't work that well for languages with pointers, because an "input" (e.g. `*uint8_t`) is often used for outputs.

ptersilie:
I'm not bothered either way. Let's keep it.

> since the trace is a function (which is literally called)

I wouldn't say that the trace is a function though. It's neither called (we jump into it from different places) nor does it ever return.

vext01:
> I wouldn't say that the trace is a function though. It's neither called (we jump into it from different places) nor does it ever return.

Well maybe that's problematic, because "argument" and "parameter" are function terms, aren't they?

> nor does it ever return.

Functions don't necessarily converge, so that's nothing to worry about.

(We are in danger of bikeshedding here)

ltratt:
A "function call" at run-time means "put arguments where the parameters tell us, do some stuff with the stack (i.e. call frame), then jump". A trace call is basically the same thing. We don't quite do stuff with the stack in the same way (because we reuse the parent stack) but otherwise it's conceptually the same: we put arguments where the parameters tell us to then jump.

PR #1484, opened 2024-12-02
Make clear that parameter indexes are `InstIdx`s.

ltratt:
Previously we used a `u32` (why such a big type? dunno!), requiring `ParamInst` to be `packed`. By definition a parameter is stored in an SSA variable which, in our context, means it's an `InstIdx`.

This commit makes this explicit:

  1. Parameters are now indexed by `InstIdx`.
  2. The vague `locidx` is now `paramidx`.
  3. The unnecessary largesse of the `params()` function returning a slice is hidden by a `param()` function which returns a single parameter.

This change did spot a bug in a test, so it is useful.

### DIFF
--- a/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
@@ -94,6 +94,7 @@ use super::{aot_ir, codegen::reg_alloc::VarLocation};
 use crate::compile::CompilationError;
 use indexmap::IndexSet;
 use std::{
+    assert_matches::debug_assert_matches,
     ffi::{c_void, CString},
     fmt,
     hash::Hash,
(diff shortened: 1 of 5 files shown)


ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -448,9 +449,10 @@ impl Module {
         self.params.push(loc);
     }
 
-    /// Return a slice over all trace parameters.
-    pub(crate) fn params(&self) -> &[yksmp::Location] {
-        &self.params
+    /// Return the parameter at a given [InstIdx].
+    pub(crate) fn param(&self, iidx: InstIdx) -> &yksmp::Location {
```
ptersilie, line 453: I'm not sure about making this `InstIdx`. My understanding is that an `InstIdx` is only used to index into the `insts` vector of the module. Should this not be a new type `ParamIdx`, as it's indexing into the `params` vector?

ltratt: Interestingly, I originally made it `ParamIdx`, got almost as far as a PR then.... I realised that the if we have *n* parameters, the first *n* instructions must be parameters. It's baked into our system that `%0`...`%n` are `ParameterInst`s, so introducing `ParamIdx` meant I just ended up with (infallible) conversions to/from `InstIdx` without really doing much useful.

ptersilie: Ok.

ltratt:
Oops, I didn't actually *force* push. Now I have. Hopefully it'll merge now!

ptersilie:
Third time's a charm.

ltratt:
Fourth time, maybe... force pushed a formatting fix exposed by the Clippy change. Oops.

PR #1488, opened 2024-12-04
Remove the "retry PT" logic.

ltratt:
This can lead to seemingly infinite hangs (50000 * 0.1s = 83 minutes), but AFAICS if it doesn't succeed at first, it doesn't succeed later.


vext01:
Do tests reliably pass with this diff on? It would surprise me if that were the case!

ltratt:
The test suite works for me, at least at small scale.

vext01:
I am surprised. And clicking merge.

PR #1489, opened 2024-12-04
Hack in "keep running if perf says it's busy".

ltratt:
This is a horrible way of doing things, but means that "perf can't open an FD" errors no longer stop the system. We can do better, but this is a reasonable temporary measure.


vext01:
Brutal.

ltratt:
This looks like one of our not-related-to-the-PR bugs. Retry?

PR #1490, opened 2024-12-05
Rejig where we calculate which instructions are used.

ltratt:
I didn't quite get this right before, so over-approximated the live variable lifetimes.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/rev_analyse.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/rev_analyse.rs
@@ -27,14 +27,15 @@ pub(super) fn rev_analyse(
     for iidx in m.iter_all_inst_idxs().rev() {
         let inst = m.inst_raw(iidx);
 
-        // "Inline" `PtrAdd`s into loads/stores, and don't mark the `PtrAdd` as used. This
-        // means that some (though not all) `PtrAdd`s will not lead to actual code being
-        // generated.
         if used_insts.get(usize::from(iidx)).unwrap()
             || inst.has_store_effect(m)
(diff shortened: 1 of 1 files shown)


ltratt:
I should have said: this small tweak gains us another 4% on big_loop. Such is the power of reducing pressure on the register allocator!

ykrt/src/compile/jitc_yk/codegen/x64/rev_analyse.rs
```diff
@@ -129,7 +128,7 @@ mod test {
         let alives = rev_analyse(&m).unwrap().0;
         assert_eq!(
             alives,
-            vec![3, 0, 5, 0, 0, 0]
+            vec![2, 0, 5, 0, 0, 0]
```
ptersilie, line 131: What do these numbers mean and why did it change from 3 to 2?

ltratt: This tells us "what is the last instruction that uses the value produced by this instruction?" Because of ptradd inlining, the 3rd instruction is dead code, so the last instruction that uses this value is actually the 2nd instruction.

ptersilie: Aha. Thank you.

ltratt:
This has got stuck in BB on the `nested_sidetrace.c` test. This might just be a weird nondeterministic error, but it might indicate a problem in the PR. I will investigate a bit.

ltratt:
@ptersilie I haven't been able to replicate this, so I guess it was a weird nondeterministic error, and we should retry the PR.

PR #1493, opened 2024-12-06
Use a pinned version of yklua in `.buildbot.sh`.

ltratt:
Previously we have got ourselves into trouble when yklua has changed, and it causes later yk builds to break. There's no guaranteed way around this problem, but we can avoid most of the pain by forcing ourselves to use a specific version of yklua in tests.

This commit does that for `.buildbot.sh` (and, while I'm here, I've done a two-stage-clone-checkout to minimise network traffic). We could perhaps use a submodule for this, but since we don't use yklua anywhere else in the test suite, that feels a bit intrusive.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -2,6 +2,9 @@
 
 set -eu
 
+# What git commit hash of yklua will we test in buildbot?
+YKLUA_COMMIT="7ced271cf4f36a78127d0cc745906cc21409ae0d"
+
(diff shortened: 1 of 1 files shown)


ltratt:
Retry?

ltratt:
Oh, hang on, bb suggests I must have made a mistake here. Investigating.

ltratt:
Force pushed what I hope is the fix.

PR #1494, opened 2024-12-06
Don't clobber a register when CALLing if we can avoid it.

ltratt:
If we're calling a non-varargs function, `rax` is clobbered, but not used, so we can `MOV` the function pointer to it. This saves us clobbering an additional register.


vext01:
Aha. I see.

My only comment would be, in the commit message:

> If we're calling a non-varargs function, `rax` is clobbered, but not used

could mention that it's not *used for arguments*. A later comment says this, so no big deal, but I was starting to question why the return value (stored in `rdx:rax`) isn't important here.

ltratt:
Force pushed a fix to the commit message.

PR #1499, opened 2024-12-09
Add a test for shadow stack allocation before an external call.

vext01:
An external call can "call-back" to interpreter code, so we still need to allocate shadow space.

Requires https://github.com/ykjit/ykllvm/pull/217


vext01:
hold off on this for this and the companion for a moment actually.

tests/src/lib.rs
```diff
@@ -21,6 +21,7 @@ pub static EXTRA_LINK: LazyLock<HashMap<&'static str, Vec<ExtraLinkage>>> = Lazy
         "loopy_funcs_not_inlined_by_default.c",
         "not_loopy_funcs_inlined_by_default.c",
         "reentrant.c",
+        "shadow_reentrant.c",
```
ltratt, line 24: I don't immediately see why this test needs extra linking?

vext01: It calls to an external function `call_callback()` which is compiled externally.

ltratt: Ah!

vext01:
Force pushed test fixes that I didn't anticipate. Go ahead.

vext01:
Oh this will need a ykllvm sync once the compiler PR is merged.

ltratt:
Please force push when that's ready.

vext01:
synced and ready.

(I note that the ykllvm sync also pulled in some SWT-related stuff)

ltratt:
Yes the SWT parts are expected.

vext01:
Ready when you are then.

ltratt:
```
11:58:56 ***** FILE 'events.lua'*****
11:58:56 testing metatables
11:58:56 ..Segmentation fault (core dumped)
11:58:59 + status=139
11:58:59 + [ 1 -eq 1 ]
```

Do we think that's a non-deterministic failure, or perhaps evidence of a problem caused the ykllvm change?

vext01:
Hrm, hard to say. I have a try_repeat going to see if I can repro it locally.

Maybe retry it?

vext01:
My run of 100 serialised debug runs just finished without hitting this seg :thinking:

ltratt:
OK, well let's retry, and see what happens.

ltratt:
OK, it's crashed in the same place, so I think we can assume we've hit an actual problem.

vext01:
I've pushed an assertion that may save us some debugging effort in the future.

With https://github.com/ykjit/ykllvm/pull/219 I think this should merge (if the benchmarks turn out to be ok).

vext01:
This will need an llvm sync after https://github.com/ykjit/ykllvm/pull/219 is in.

vext01:
synced llvm.

PR #1500, opened 2024-12-09
Don't execute a trace while tracing.

vext01:
This is a fix @ltratt supplied to me last week. I assume we still want this?

I updated the test naively. Please check it's correct.


ltratt:
I think the system is incorrect without this, because it can start executing a trace while tracing another!

PR #1503, opened 2024-12-10
Rework common subexpression elimination

ltratt:
This PR has been in the back of my mind now for ages -- as an optimisation. I've now realised it allows us to further simplify our internal APIs and remove a footgun or two.

In essence, we port over a datastructure from LuaJIT that allows us to efficiently iterate over all instructions of type T: this allows us to recast CSE in these terms (https://github.com/ykjit/yk/commit/ba5acf2d984a350d32dced2d8e37a795d88d3818). That then makes it clearer that we had a sort-of-second form of CSE, just for guards, which we can then fold into the main CSE (https://github.com/ykjit/yk/commit/7a292abaefe154ebc76d694b71395f47c2d876c2).

The performance difference of this PR is too small for me to able to measure, but that's less the point: basically we get to move some subtle code out of the `Opt` keeping the latter simpler, and also removing another "globalish" function from `Module`.

### DIFF
--- a/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
@@ -99,6 +99,7 @@ use std::{
     hash::Hash,
     mem,
 };
+use strum::{EnumCount, EnumDiscriminants};
 #[cfg(not(test))]
 use ykaddr::addr::symbol_to_ptr;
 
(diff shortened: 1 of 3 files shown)


ptersilie:
This is neat. I was hoping I could use this for loop peeling and query it to figure out if an instruction is a CSE candidate. But it seems I still need to potentially iterate over multiple other candidates, so it's going to be slower than using a `Vob`.

ptersilie:
Looks like one of the bugs we have. Let's try again.

ltratt:
> I was hoping I could use this for loop peeling and query it to figure out if an instruction is a CSE candidate.

Right, you *could* use this, but it has a worse case O(n) whereas for your particular use case you should (I think) be able to use a data-structure (e.g. `Vob`) that has a worse case O(1).

PR #1505, opened 2024-12-10
Minor tidy ups

ltratt:
A couple of minor tweaks to #1503 that I noticed on rereading the code.


ltratt:
Force pushed an update to fix the unused import (mea culpa).

PR #1506, opened 2024-12-11
Propagate register information backwards

ltratt:
This PR is the start -- but not the end! -- of implementing our poor man's reverse code generator. We take a leaf out of RPython's book, and in the reverse analysis phase that happens before code gen, we propagate information about what register a value should end up in backwards. We do our best to propagate this across multiple instructions. Overall this gets another 5% or so improvement on big_loop.

There is more we can do here: most obviously, we could propagate information about where values should be spilled. And I suspect that if I think hard about things I can probably find more ways of using the new `OutputCanBeSameAsInput` constraint (https://github.com/ykjit/yk/commit/e3743e39278f62cd4face2977efeac5066007248). However, this PR is already quite big and complex so it seems best to raise it now before it gets out of hand.

@ptersilie I wonder if https://github.com/ykjit/yk/commit/c219baef65a444dbc08f32df49a0634bb054e8df is useful for your ongoing PR? If so, you could cherry pick it if you want.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
@@ -342,6 +342,7 @@ impl LSRegAlloc<'_> {
                     | RegConstraint::InputIntoRegAndClobber(_, _) => false,
                     RegConstraint::InputOutputIntoReg(_, _)
                     | RegConstraint::Output
+                    | RegConstraint::OutputCanBeSameAsInput(_)
                     | RegConstraint::OutputFromReg(_)
                     | RegConstraint::InputOutput(_) => true,
                     RegConstraint::Clobber(_) | RegConstraint::Temporary => false,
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/rev_analyse.rs
```diff
@@ -57,6 +75,64 @@ impl<'a> RevAnalyse<'a> {
         }
     }
 
+    fn an_binop(&mut self, iidx: InstIdx, binst: BinOpInst) {
+        match binst.binop() {
+            BinOp::Add | BinOp::And | BinOp::Or | BinOp::Xor => {
```
vext01, line 80: Have you considered doing one for `Call` where the ABI constrains the argument/return registers?

ltratt: Yes, but big_loop doesn't exercise that part, so I'll do it in the futuer.

PR #1510, opened 2024-12-13
Implement loop peeling.

ptersilie:
This implements a basic version of loop peeling. Note, that additional optimisations based on loop peeling, like common subexpression eliminiation, are not included in this PR. Those will be implemented in subsequent PRs.


ltratt:
On a fresh clone (and even if I merge in `master`), I see a different error:

```
thread '<unnamed>' panicked at ykrt/src/compile/jitc_yk/jit_ir/mod.rs:333:19:
index out of bounds: the len is 67 but the index is 69
stack backtrace:
   0: rust_begin_unwind
             at /rustc/5e1440ae514d98ddfcbf1607acb64d41e07ef616/library/std/src/panicking.rs:681:5
   1: core::panicking::panic_fmt
             at /rustc/5e1440ae514d98ddfcbf1607acb64d41e07ef616/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_bounds_check
             at /rustc/5e1440ae514d98ddfcbf1607acb64d41e07ef616/library/core/src/panicking.rs:273:5
   3: <usize as core::slice::index::SliceIndex<[T]>>::index
             at /rustc/5e1440ae514d98ddfcbf1607acb64d41e07ef616/library/core/src/slice/index.rs:274:10
   4: core::slice::index::<impl core::ops::index::Index<I> for [T]>::index
             at /rustc/5e1440ae514d98ddfcbf1607acb64d41e07ef616/library/core/src/slice/index.rs:16:9
   5: <alloc::vec::Vec<T,A> as core::ops::index::Index<I>>::index
             at /rustc/5e1440ae514d98ddfcbf1607acb64d41e07ef616/library/alloc/src/vec/mod.rs:3346:9
   6: ykrt::compile::jitc_yk::jit_ir::Module::inst_raw
             at /vol/extra/ltratt/yk/ykrt/src/compile/jitc_yk/jit_ir/mod.rs:333:19
   7: ykrt::compile::jitc_yk::jit_ir::PackedOperand::unpack
             at /vol/extra/ltratt/yk/ykrt/src/compile/jitc_yk/jit_ir/mod.rs:1104:23
   8: ykrt::compile::jitc_yk::jit_ir::Inst::map_operand_locals
             at /vol/extra/ltratt/yk/ykrt/src/compile/jitc_yk/jit_ir/mod.rs:1588:21
   9: ykrt::compile::jitc_yk::jit_ir::dead_code::<impl ykrt::compile::jitc_yk::jit_ir::Module>::dead_code_elimination
             at /vol/extra/ltratt/yk/ykrt/src/compile/jitc_yk/jit_ir/dead_code.rs:20:17
  10: ykrt::compile::jitc_yk::opt::Opt::opt
             at /vol/extra/ltratt/yk/ykrt/src/compile/jitc_yk/opt/mod.rs:426:9
  11: ykrt::compile::jitc_yk::opt::opt
             at /vol/extra/ltratt/yk/ykrt/src/compile/jitc_yk/opt/mod.rs:567:5
```

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1844,14 +1853,18 @@ impl<'a> Assemble<'a> {
                         reg_alloc::Register::GP(r) => {
                             let [_] = self.ra.assign_gp_regs(
                                 &mut self.asm,
-                                iidx,
+                                // This tells the register allocator to assign this register
+                                // to the instruction at `iidx` so we can look it up later. Since
```
ltratt, line 1857: The "since we are at the end" bit is dangerous: the register allocator does (must!) take into account `iidx` when making decisions about what to spill etc. By passing in `0` here, I think weird things will happen. Is there any reason not to pass `iidx` here (and below)?

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -423,8 +429,10 @@ impl Module {
     }
 
     /// Push the location of a trace parameter.
-    pub(crate) fn push_param(&mut self, loc: yksmp::Location) {
+    pub(crate) fn push_param(&mut self, loc: yksmp::Location) -> usize {
```
ltratt, line 432: This should return a `ParamIdx`.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -1582,6 +1617,183 @@ impl Inst {
...
+                let args = (0..(dc.num_args()))
+                    .map(|i| op_mapper(&dc.operand(m, i), m))
+                    .collect::<Vec<_>>();
+                let dc = DirectCallInst::new(m, dc.target, args)?;
+                Inst::Call(dc)
+            }
+            Inst::IndirectCall(iidx) => {
+                let ic = m.indirect_call(*iidx);
+                // Clone and map arguments.
+                let args = (0..(ic.num_args()))
```
ltratt, line 1658: Ditto `IndirectCallInst::iter_args_idx`.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -1582,6 +1617,183 @@ impl Inst {
...
+    /// Duplicate this [Inst] while applying function `f` to each operand.
+    pub(crate) fn dup_and_remap_locals<F>(
+        &self,
+        m: &mut Module,
+        f: &F,
+    ) -> Result<Self, CompilationError>
+    where
+        F: Fn(InstIdx, &Module) -> Operand,
+    {
+        let mapper = |x: &PackedOperand, m: &Module| match x.unpack(m) {
```
ltratt, line 1629: We always make `m` the first argument (it's a useful convention IMHO).

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -1582,6 +1617,183 @@ impl Inst {
...
+                Inst::BlackBox(BlackBoxInst { op: mapper(op, m) })
+            }
+            Inst::BinOp(BinOpInst { lhs, binop, rhs }) => Inst::BinOp(BinOpInst {
+                lhs: mapper(lhs, m),
+                binop: *binop,
+                rhs: mapper(rhs, m),
+            }),
+            Inst::Call(dc) => {
+                // Clone and map arguments.
+                let args = (0..(dc.num_args()))
```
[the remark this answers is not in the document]
ptersilie: I don't follow how this would help make things easier here.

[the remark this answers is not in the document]
ltratt: I think you can simplify this to something like:

```rust
let args = dc.iter_args_idx().map(|x| op_mapper(m, x)).collect::<Vec<_>>()
```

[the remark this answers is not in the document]
ltratt: I'm wrong. And the `operand` method is horrible and should be removed, but that's a different matter!

So I think this should be:

```rust
let args = dc.iter_args_idx().map(|x| op_mapper(m, m.arg(x))).collect::<Vec<_>>()
```

Preferably we should kill the `operand` method, though we don't have to do it in this PR.

tests/c/nested_sidetrace.c
```diff
@@ -32,16 +32,17 @@
 //     18
 //     yk-jit-event: enter-jit-code
 //     yk-jit-event: deoptimise
-//     yk-jit-event: start-side-tracing
```
ltratt, line 35: Why has the PR changed when side tracing starts?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -507,9 +514,11 @@ impl<'a> Assemble<'a> {
...
                 }
                 jit_ir::Inst::Guard(i) => self.cg_guard(iidx, i),
-                jit_ir::Inst::TraceLoopStart => self.cg_traceloopstart(),
-                jit_ir::Inst::TraceLoopJump => self.cg_traceloopjump(),
-                jit_ir::Inst::RootJump => self.cg_rootjump(self.m.root_jump_addr()),
+                jit_ir::Inst::TraceHeaderStart => self.cg_header_start(),
+                jit_ir::Inst::TraceHeaderEnd => self.cg_header_end(),
+                jit_ir::Inst::TraceBodyStart => self.cg_body_start(),
+                jit_ir::Inst::TraceBodyEnd => self.cg_body_end(iidx),
+                jit_ir::Inst::RootJump => self.cg_rootjump(iidx, self.m.root_jump_addr()),
```
ltratt, line 521: For consistency should this be `SideTraceEnd` or similar?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1776,20 +1788,28 @@ impl<'a> Assemble<'a> {
     /// # Arguments
     ///
     /// * `tgt_vars` - The target locations. If `None` use `self.loop_start_locs` instead.
-    fn write_jump_vars(&mut self, tgt_vars: Option<&[VarLocation]>) {
+    fn write_jump_vars(&mut self, iidx: InstIdx, is_sidetrace: bool) {
+        let (tgt_vars, src_ops) = if is_sidetrace {
+            // Side-traces don't have a body.
```
ltratt, line 1793: This briefly confused me, but I then realised what it's saying. I think we should document *somewhere* this property, because it's kind of fundamental to the things we can do with a side-trace.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1776,20 +1788,28 @@ impl<'a> Assemble<'a> {
...
+            (self.m.root_entry_vars(), self.m.trace_header_end())
+        } else {
+            (self.body_start_locs.as_slice(), self.m.trace_body_end())
+        };
         // If we pass in `None` use `self.loop_start_locs` instead. We need to do this since we
         // can't pass in `&self.loop_start_locs` directly due to borrowing restrictions.
-        let tgt_vars = tgt_vars.unwrap_or(self.loop_start_locs.as_slice());
-        for (i, op) in self.m.loop_jump_operands().iter().enumerate() {
+        //let tgt_vars = tgt_vars.unwrap_or(self.loop_start_locs.as_slice());
```
ltratt, line 1800: Commented code.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ptersilie: [d61a6cf](https://github.com/ykjit/yk/pull/1510/commits/d61a6cf41a5990dd88cc141223d232e925ab951a)

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ptersilie: [1975eae](https://github.com/ykjit/yk/pull/1510/commits/1975eae011eb4e365fb6e417f3165a709b4cd89d)

tests/c/nested_sidetrace.c
ptersilie: Because the guard in the header and the guard in the body are two different guards. So we typically fail one iteration later than before (in order to hit 5 guard failures we need to fail 6 times since the first failure is in the body). I tried to minimise the changes by reducing the default threshold to 4, but that didn't do the trick for all tests (likely because not all tests have the header guard fail immediately).

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: Could do.

ltratt:
We will need at least one (and preferably more) C tests for this: they don't necessarily have to be super-super complex, but we should at least check that `jit-post-opt` has peeled the loop!

tests/c/nested_sidetrace.c
ltratt: Ah, right, that makes sense. Let's put the default threshold back to 5: it's better to just bite the bullet and adjust the tests now. If you want -- but I wouldn't necessarily do this -- you can set the threshold to 4 in individual tests.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ltratt: We fixed this in a commit (https://github.com/ykjit/yk/pull/1510/commits/860f69c87b4009b32a680b9cfeb8ecf7a4341f1b).

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -2293,9 +2376,10 @@ impl CompiledTrace for X64CompiledTrace {
             .collect();
         let callframes = deoptinfo.inlined_frames.clone();
 
-        // Calculate the address inside the root trace we want side-traces to jump to. Currently
-        // this is directly after the prologue. Later we will change this to jump to after the
-        // preamble and before the peeled loop.
+        // Calculate the address inside the root trace we want side-traces to jump. Since the
+        // side-trace finishes at the control point we need to re-enter via the trace header and
+        // cannot jump back directly into the trace body.
+        // FIXME: Check if RPython has found a solution to this (if there is any).
```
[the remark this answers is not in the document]
ltratt: I mean: a guard that fails in the body should, I think, create a side trace which jumps back to the beginning of the body (not the beginning of the head).

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: [3beabc9](https://github.com/ykjit/yk/pull/1510/commits/3beabc991ca6360e255a180faf69870b740896d1)

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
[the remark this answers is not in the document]
ptersilie: But that's not how side-tracing works. Both side-traces execute until the get back to the control point. If you now jump into the body you've missed the execution of the head and then miss values that the head computes and passes into the body. Where else are you going to get those values from?

tests/c/nested_sidetrace.c
ptersilie: Ok, but note this will implicitly increase the side-trace threshold to 6, at least as far as an external observer is concerned.

ltratt: Hmm, yeah, we probably need to fix this more thoroughly. OK, let's do the `4` hack for now.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: [e74e498](https://github.com/ykjit/yk/pull/1510/commits/e74e4983fd39ab895860b92ee3541244233e45ea)

tests/c/nested_sidetrace.c
ptersilie: [0ae874b](https://github.com/ykjit/yk/pull/1510/commits/0ae874b880bd85495f277e63bde4c15e43e245bb)

I've reset it now. :-/ Do you want me to revert it again?

I've also left the manual adjustment in the `side-trace.c` test in as it was easier this way.

ltratt: Do whichever is easiest for now.

ptersilie: > Hmm, yeah, we probably need to fix this more thoroughly. OK, let's do the `4` hack for now.

Yes, also since a side-trace itself doesn't have a header and thus setting a threshold of 5 will behave as expected.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: [6b79d8e](https://github.com/ykjit/yk/pull/1510/commits/6b79d8effdac0f3d4051d7b7ea2f12f98c43c664)

tests/c/nested_sidetrace.c
ptersilie: Ok, let's leave it how it is now.

ltratt:
Right, so just the missing C tests left I think.

ptersilie:
I've just added one simple test. I'm currently trying to make a C tests where we can see constant forwarding into the body, but I haven't had any luck so far.

ltratt:
>I'm currently trying to make a C tests where we can see constant forwarding into the body, but I haven't had any luck so far.

I think `yk_promote` will do what you want.

ptersilie:
> I think `yk_promote` will do what you want.

I tried all the promote tests. But it seems because we are using NOOPT_VAL they are always loaded from the stack and thus not passed into the body.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -1587,6 +1628,193 @@ impl Inst {
...
     }
 
+    /// Duplicate this [Inst] while applying function `f` to each operand.
+    pub(crate) fn dup_and_remap_locals<F>(
+        &self,
+        m: &mut Module,
+        f: &F,
+    ) -> Result<Self, CompilationError>
+    where
+        F: Fn(InstIdx, &Module) -> Operand,
```
ltratt, line 1638: Just noticed: this should be `(&Module, InstIdx)`.

ltratt:
Please squash.

ptersilie:
Squashed and rebased.

PR #1512, opened 2024-12-13
Turn hard errors in trace decoding into a recoverable failure.

ltratt:
We sometimes -- not often, but sometimes -- encounter either the `assert` or `unreachable` in the code below. This commit turns these cases into recoverable errors. Is that the right thing to do? I'm not entirely sure, and welcome Edd's comments!


vext01:
Ideally we'd investigate and figure out why this happens, but perhaps we have bigger fish to fry?

ltratt:
>Ideally we'd investigate and figure out why this happens, but perhaps we have bigger fish to fry?

This is your parish, so you decide.

vext01:
(if you mark it ready, i will merge)

ltratt:
Marked as ready.

PR #1513, opened 2024-12-14
Detect and abort tracing if "early return" occurs.

ltratt:
If we start tracing a loop and the function-with-a-control-point returns whilst tracing is ongoing, we are in dangerous territory. For quite a while we assumed that "dangerous territory" meant (gongs of doom!) "UB". Actually, we just hadn't created very good examples: when I did, it turns out that an `unwrap` in trace_builder fails if early return happens. We *could*, therefore, detect early returns there, but that means that we've done a lot of work before realising this.

This commit instead has `mt.rs` detect early returns, and abort tracing, as soon as we reasonably can. [Note, however, that if a user annotates `return`s from the function, they could stop things even earlier. Is this worth it? It is quite complicated to explain and, as we'll see in a bit, it's not a very common case]

All that said, it turns out early returns aren't that common: you have to start tracing on the same iteration that you "early return" from. This is definitely possible -- indeed the two C tests in the PR do that, one "normally" and one when side-tracing -- but it seems unlikely that it will happen very often. So although we could force the user to help us detect this even sooner, I'm not sure it's worth burdening them or us with that responsibility right now: the cases where it would save serious resources are borderline pathological. Indeed, the only such case I can think of is: if you start tracing from the "root" control point call frame, and then return to the main program (i.e. you will never go back to the interpreter), then tracing is never turned off. One day we can fix this, but I don't think we will encounter it very soon.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -28,6 +28,7 @@ use crate::{
         stats::{Stats, TimingState},
         Log, Verbosity,
     },
+    stack::{StackDirection, STACK_DIRECTION},
     trace::{default_tracer, AOTTraceIterator, TraceRecorder, Tracer},
 };
 
@@ -401,8 +402,19 @@ impl MT {
 
     #[allow(clippy::not_unsafe_ptr_arg_deref)]
     pub fn control_point(self: &Arc<Self>, loc: &Location, frameaddr: *mut c_void, smid: u64) {
-        match self.transition_control_point(loc) {
+        match self.transition_control_point(loc, frameaddr) {
             TransitionControlPoint::NoAction => (),
+            TransitionControlPoint::AbortTracing => {
+                let thread_tracer =
+                    MTThread::with(
+                        |mtt| match mtt.tstate.replace(MTThreadState::Interpreting) {
+                            MTThreadState::Tracing { thread_tracer, .. } => thread_tracer,
+                            _ => unreachable!(),
+                        },
+                    );
+                thread_tracer.stop().ok();
+                self.log.log(Verbosity::JITEvent, "tracing-aborted");
+            }
             TransitionControlPoint::Execute(ctr) => {
                 self.log.log(Verbosity::JITEvent, "enter-jit-code");
                 self.stats.trace_executed();
@@ -438,6 +450,7 @@ impl MT {
                             hl,
                             thread_tracer: tt,
                             promotions: Vec::new(),
+                            frameaddr,
                         };
                     }),
                     Err(e) => {
@@ -474,7 +487,11 @@ impl MT {
                                 hl,
                                 thread_tracer,
                                 promotions,
-                            } => (hl, thread_tracer, promotions),
+                                frameaddr: tracing_frameaddr,
+                            } => {
+                                assert_eq!(frameaddr, tracing_frameaddr);
+                                (hl, thread_tracer, promotions)
+                            }
                             _ => unreachable!(),
                         },
                     );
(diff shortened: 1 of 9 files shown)


ykrt/src/mt.rs
```diff
@@ -599,16 +624,49 @@ impl MT {
...
                                     } else {
-                                        // ...and it's this location: we have therefore finished
-                                        // tracing the loop.
-                                        lk.kind = HotLocationKind::Compiling;
-                                        TransitionControlPoint::StopTracing
+                                        // ...and it's this location...
+                                        match STACK_DIRECTION {
+                                            StackDirection::GrowsToHigherAddress => todo!(),
+                                            StackDirection::GrowsToLowerAddress => {
+                                                if frameaddr <= *tracing_frameaddr {
```
vext01, line 641: So it's OK to stop tracing in deeper recursive calls?

I'm not sure. Doesn't that mean the trace looping will loop across frames, but without including the "bubbling up", which may be important for program sematics?

e.g.

```
f() {
  loop {
   control_point()
    if somthing {
      f();
    }
  }
  important()
}
```

So here, if you start tracing at frame depth 0, and stop and frame depth (say) 10, does it matter that the trace doesn't include `important()`? I'm not sure.

ptersilie:
Looks good to me.

ykrt/src/mt.rs
```diff
@@ -1039,20 +1143,25 @@ mod tests {
...
                 promotions: Vec::new(),
+                frameaddr: 0 as *mut c_void,
             };
         });
     }
 
     fn expect_stop_tracing(mt: &Arc<MT>, loc: &Location) {
-        let TransitionControlPoint::StopTracing = mt.transition_control_point(loc) else {
+        let TransitionControlPoint::StopTracing =
+            mt.transition_control_point(loc, 0 as *mut c_void)
```
vext01, line 1163: Clippy will probably say to use `ptr::null_mut()`.

ykrt/src/stack.rs
```diff
@@ -0,0 +1,12 @@
+/// Which direction does the CPU stack grow when more room is required?
+pub(crate) enum StackDirection {
+    /// The CPU stack grows "downwards" i.e. growth changes it to a lower address.
+    #[allow(dead_code)]
+    GrowsToHigherAddress,
+    /// The CPU stack grows "upwards" i.e. growth changes it to a higher address.
+    #[allow(dead_code)]
```
vext01, line 7: Is this second variant dead?

ykrt/src/mt.rs
vext01: As long as the trace looper is happy with this, I am.

vext01:
Do I understand correctly that this will superceed #1481? Because we would no longer require the user to do manual annotations for early returns?

ltratt:
>Do I understand correctly that this will superceed https://github.com/ykjit/yk/pull/1481? Because we would no longer require the user to do manual annotations for early returns?

Yes and yes.

ykrt/src/mt.rs
ltratt: There are no warnings from Clippy. Note this is a `c_void`.

vext01: It should work since `null_mut<T>()` parametrised. Fewer magic numbers, right?

vext01:
Surprised the casts were needed. Doesn't it just coerce?

ltratt:
You're right. Force pushed the simplification.

vext01:
Thanks!

PR #1514, opened 2024-12-16
`InlinedFrame` should store `PackedOperand` not `Operand`.

ltratt:
This exposes (so far rare) additional opportunities for constants to be identified, so we have to remove the `panic` branch. I admit that I haven't been able to work out a small test for this, though it happens in the yklua test suite.


vext01:
> I admit that I haven't been able to work out a small test for this, though it happens in the yklua test suite.

Maybe reduce to try to get clues as to what causes this?

ltratt:
I don't have a setup or time for doing so right now, but I wouldn't object to someone doing so.

PR #1517, opened 2024-12-16
Percolate dladdr() errors upwards.

ltratt:
The fundamental change here is to turn an `unwrap` in `vaddr_to_off` into an `Err` but this part of hwtracer is somewhat messy, so a seemingly simple change (`unwrap` -> `Err(...)`) requires lots of boilerplate.


vext01:
Yikes! Thanks for this. At least we will get a descriptive error message.

PR #1518, opened 2024-12-19
Minor tidy ups

ltratt:


ltratt:
That looks like `Queens` failed, but probably for nondeterministic reasons unrelated to this PR. Retry?

PR #1519, opened 2024-12-19
Peel the peeler

ltratt:
This PR gets more optimisations from peeling, in essence by propagating more constants from the header into the body. The easiest way to see this is that we can now write a trivial test for peeling, showing that a guard has been removed in the body (https://github.com/ykjit/yk/commit/83c9f166da2bafb7c6a33475f9df7f60bf273d6c).

The main enabling commit is https://github.com/ykjit/yk/commit/4858809e79b6aefd00ee4455dc93b08129f7ab03, which rewrites all of an instruction's operands to pick up everything we have learnt from the analyser.

However, to get to the point that we can do that safely, we have to make a number of changes. The most fundamental change is https://github.com/ykjit/yk/commit/ab09557f32b88d42b753a26920b688272659b081, which in essence moves instruction-changing from the `dup_and_remap` method to the optimiser.

Not everything about this PR is perfect: I've left a couple of `pub(crate)` fields, I've introduced a `RefCell` (in a way that offends me to my core), and I've not worried about making the optimiser work very efficiently. However, overall, I think this nudges us further in the right direction, which is reflected in 4% boost to big_loop. More will come in the future, with luck.


ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -77,13 +78,15 @@ impl Opt {
...
         let skipping = self.m.iter_skipping_insts().collect::<Vec<_>>();
         for (iidx, inst) in skipping.into_iter() {
-            let c = inst.dup_and_remap_vars(&mut self.m, |op_iidx: InstIdx| {
-                Operand::Var(iidx_map[usize::from(op_iidx)])
-            })?;
-            let copy_iidx = self.m.push(c)?;
-            iidx_map[usize::from(iidx)] = copy_iidx;
             match inst {
                 Inst::TraceHeaderStart => {
+                    self.m.trace_body_start = self.m.trace_header_start().to_vec();
```
ptersilie, line 78: This has the disadvantage that we have to iterate over these values twice. Once when cloning them and again during remapping. Why don't we just `assert` here that we can only see `TraceHeaderStart` once per `Module`?

ltratt: I agree, we could probably fix that. The performance difference will be tiny and I'm not quite sure how to do it without adding more code.

ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -118,7 +114,26 @@ impl Opt {
...
+        }
+
+        // Create a fresh `instll`. Normal CSE in the body (a) can't possibly reference the header
+        // (b) the number of instructions in the `instll`-for-the-header is wrong as a result of
+        // peeling. So create a fresh `instll`.
+        let mut instll = InstLinkedList::new(&self.m);
+        let skipping = self
+            .m
+            .iter_skipping_insts()
+            .skip_while(|(_, inst)| !matches!(inst, Inst::TraceBodyStart))
```
ptersilie, line 127: Didn't know about `skip_while`, but that's very neat.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -1800,6 +1795,16 @@ impl Inst {
                 m.trace_body_end = m.trace_body_end.iter().map(|op| mapper(m, op)).collect();
                 Inst::TraceBodyEnd
             }
+            Inst::TraceHeaderEnd => {
+                // Copy the header label into the body while remapping the operands.
+                m.trace_header_end = m.trace_header_end.iter().map(|op| mapper(m, op)).collect();
```
ptersilie, line 1800: Now you've lost me. Why are we remapping the header? Shouldn't the header stay as it was before we peeled?

ltratt: This is where we discover constants from the analyser. [Well, not here, but `opt/mod.rs` calls this, and finds out constants this way.]

ptersilie: Oh, this makes more sense after seeing the following commit. Just to check I got this right: you added this because you later use this function for other things than just peeling off a loop (e.g. to propagate analyser info). When peeling the loop, this function is never called with `TraceHeaderEnd` so we never remap this at that time.

ltratt: Yep!

ptersilie:
Good stuff.

ltratt:
The failure seems to be ASAN putting in extra code that makes our peeling test bork. AFAICS there is no way to write the (quite tight) test we want if ASAN does this, so I've just `ignore-if`ed that test under ASAN in the force push.

ltratt:
Groan, so swt generates different IR than hwt, so I've had to duplicate the test for each tracer with two lines differing both in `......` (i.e. "a line needs to be here but I don't care what"). Force pushed the update.

PR #1522, opened 2024-12-20
Reorder display() arguments in one place.

vext01:
Across both IRs the display() methods always accept a module as the first argument... apart from this one place which catches me out every time.

Let's make it consistent across the board.


ltratt:
Agreed!

PR #1525, opened 2024-12-22
Use `Module::tracekind()` to examine the kind of trace we have.

ltratt:
This is more idiomatic, and also allows us to catch one possible way that we might do things wrong.


ptersilie:
Is there a file missing here or have you introduced TraceKind in another PR that I've missed?

ltratt:
It was introduced in https://github.com/ykjit/yk/pull/1523.

ltratt:
I think this is another of our favourite nondeterministic errors. Retry?

PR #1527, opened 2024-12-22
Constant fold subtractions.

ltratt:
Note we have to be careful about this, because subtraction, unlike addition, is not commutative.

This shaves another 2.5% off big_loop.


vext01:
Needs clippy :)

PR #1529, opened 2024-12-24
Forward register hints from header to body.

ltratt:
When we implemented peeling, we forgot -- it didn't even occur to me! -- to forward register hints from the header to the body. In essence, we told the body where to put things, put not the header, so the register allocator started the body in a different state to the one we wanted to end up in.

This all worked, of course, because hints are just hints. However, the lack of hints hugely increased the chances of needing to do lots of register swaps at the end of the body.

Propagating the hints gets us another 12% or so on big_loop.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/rev_analyse.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/rev_analyse.rs
@@ -54,6 +54,9 @@ impl<'a> RevAnalyse<'a> {
                             continue;
                         }
                     }
+                    Inst::TraceHeaderEnd => {
+                        self.an_header_end();
+                    }
                     Inst::TraceBodyEnd => {
                         self.an_body_end();
                     }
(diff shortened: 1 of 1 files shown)


ptersilie:
Huh, I'm surprised this results in such a big speed up. I assumed that the body start and body end already have the same locations. The header end only matters when we jump into the body, which is only once on big loop. Or is it that we copy the params from the header_ens into the body_start and that's why they are different?

Anyway, a speedup is a speedup so let's get this in.

ltratt:
In essence, because we didn't give hints to the allocator as where to place things at `header_end`, it placed them in any old position; but the `body_end` had the "correct" hints; so `header_end` and `body_end` would (in 99.99999% of cases) end up with different allocations.

As to why it's such a big speed-up, there are two not-entirely-obvious factors:

1. Every time we shave a CPU cycle off the code we generate, the next CPU cycle we shave off constitutes a relatively bigger % win.
2. All the shadow stack stores we had present until recently constituted a big fixed portion of the running code and in a sense were somewhat masking actual improvements we made.

In a sense, (2) is a special case of (1), but it's such a big thing that it's worth pulling out separately IMHO.

PR #1531, opened 2024-12-28
Change and extend the reverse analyser

ltratt:
This PR gradually gets the reverse analyser in better shape, ultimately speeding big_loop up by another 5%. There are really 3 things going on here:

1. We're only ever propagating register hints, not (the more general) `VarLocation`s (https://github.com/ykjit/yk/commit/277a52bcd789faaf74a7cfedebcc69f2ffa5380f and https://github.com/ykjit/yk/commit/c3680da61d1e3e0641e6ef123147954813428b51).
2. Previously we forced the trace header *and* body to have the same hints, but this misses an opportunity: the header can leave additional things in registers for the body. This requires a number of refactorings, as we now have to run reverse analysis in two phases: once for the header, then we run the code generator, then we reverse analyse the body (https://github.com/ykjit/yk/commit/8f3aab78555ac925314301f3c3c846e9477d9993 and https://github.com/ykjit/yk/commit/4a9454b86e723ecee0fb19e0aa55cf6c3f198364).
3. We were generating inappropriate hints for `OutputCanBeSameAsInput` instructions (https://github.com/ykjit/yk/commit/6ceac885bb5e63de1c91ee2e866634c28061cbac).

[FWIW, I experimented with a couple of designs for (2), but came to the conclusion that there's not much point in wholesale change here. It *might* be marginally better to treat header and body as separate `Module`s which we `jmp` between (I think that would be closer in spirit to RPython), but we are now fairly well set-up to deal with the two combined. I don't think changing now would gain us much, and it would cause a lot of churn.]

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
@@ -169,7 +169,7 @@ impl<'a> LSRegAlloc<'a> {
 
     /// Reset the register allocator. We use this when moving from the trace header into the trace
     /// body.
-    pub(crate) fn reset(&mut self) {
+    pub(crate) fn reset(&mut self, header_end_vlocs: &[VarLocation]) {
         for rs in self.gp_reg_states.iter_mut() {
             *rs = RegState::Empty;
         }
@@ -186,7 +186,7 @@ impl<'a> LSRegAlloc<'a> {
         }
         self.fp_regset = RegSet::with_fp_reserved();
 
-        self.rev_an.analyse_body();
+        self.rev_an.analyse_body(header_end_vlocs);
     }
 
     /// Before generating code for the instruction at `iidx`, see which registers are no longer
--- a/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
@@ -124,7 +124,7 @@ pub(crate) struct LSRegAlloc<'a> {
     /// The abstract stack: shared between general purpose and floating point registers.
     stack: AbstractStack,
     /// What [VarLocation] should an instruction aim to put its output to?
-    vloc_hints: Vec<Option<VarLocation>>,
+    reg_hints: Vec<Option<VarLocation>>,
 }
 
 impl<'a> LSRegAlloc<'a> {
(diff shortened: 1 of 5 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/rev_analyse.rs
```diff
@@ -1,142 +1,265 @@
...
+
+                while let Some((iidx, inst)) = next {
+                    self.analyse(iidx, inst);
+                    next = iter.next();
+                }
+            }
+        }
+    }
+
+    /// Analyse a trace body. This must be called iff both of the following are true:
```
vext01, line 121: would it be worth adding some debug_asserts for these preconditions then?

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
```diff
@@ -262,24 +256,36 @@ impl<'a> LSRegAlloc<'a> {
         self.stack.size()
     }
 
+    /// Is the instruction at [iidx] dead?
+    pub(crate) fn is_inst_dead(&self, iidx: InstIdx) -> bool {
```
vext01, line 260: Does "dead" mean "was alive and now is not", or could it also mean "has not been live yet, but will be"?

I think the former, since iirc this replaces something with a name like "is_still_alive_after()"?

ykrt/src/compile/jitc_yk/codegen/x64/rev_analyse.rs
ltratt: I did try but the only way of doing so is via global state: I ended up removing it as it just felt like overkill (probably 10LoC, spread over the file, for something that you'd have to be fairly silly to do). I'm not sure I can think of a better way TBH...

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
ltratt: Good question. It means "has DCE identified it as equivalent to a tombstone?" "Dead" was my quick attempt to capture that, but I can see that it might not be entirely clear. I'm open to suggestions!

vext01: Does `has_inst_died()` capture it perhaps? The past tense is an attempt to anchor the event in time.

ltratt: It's not exactly time based: it really means "has DCE identified this as dead". That *could* even happen conceptually during trace creation. So "is" is probably right. I think "died" might be the contentious bit: it could be that we should call is `is_inst_tombstone` or similar?

ltratt: Changed to `is_inst_tombstone` in the `fixup!` commits below. [Note: I accidentally pushed another change which I didn't intend, so force pushed that away.]

vext01:
please squash.

ltratt:
Squashed.

PR #1532, opened 2024-12-29
Clippy fixes

ltratt:
Various fixes suggested by the new version of Clippy.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -52,8 +52,7 @@ fn clang_format(check_only: bool) {
                                             Some(format!("returned exit code {c}:\n\n{stderr}"))
                                     }
                                     None => {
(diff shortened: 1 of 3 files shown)


xtask/src/main.rs
```diff
@@ -52,8 +52,7 @@ fn clang_format(check_only: bool) {
                                             Some(format!("returned exit code {c}:\n\n{stderr}"))
                                     }
                                     None => {
-                                        err_msg =
-                                            Some("terminated by signal:\n\n{stderr}".to_owned())
+                                        err_msg = Some(format!("terminated by signal:\n\n{stderr}"))
```
vext01, line 55: Hah! I bet there was a variable called `stderr` at one time.

PR #1537, opened 2025-01-07
Teach the reverse analyser about calls

ltratt:
The main commit ensures that operands tend to end up in the right register for calls; the second commit is the bare minimum necessary to take advantage of this for FP arguments.


ptersilie:
Just a clippy error it seems.

ltratt:
Oops! Force pushed a Clippy fix.

ptersilie:
One of the nondeterministic bugs bit us. Let's try again.

ltratt:
I can't replicate this on b16 after `try_repeat 1000`. So maybe we should retry it again, but perhaps also bear in mind the possibility that it's highlighting a bug that b13 makes particularly apparent.

PR #1538, opened 2025-01-08
Better document Direct/Indirect locations.

ptersilie:
While we are here add an assertion that's been on my mind and is loosely related to direct locations.

@ltratt I'm not sure if you wanted more documentation here, but my mind is blank. So please let me know if you think there's more that can be added here.

### DIFF
--- a/yksmp/src/lib.rs
+++ b/yksmp/src/lib.rs
@@ -45,7 +45,7 @@ impl Record {
 }
 
 /// Describes where live variables are stored at specific times during execution.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub enum Location {
     /// The live variable is stored in a register. Note, that LLVM's stackmap only stores one
     /// location per live variable, which is enough for reading them out. For deoptimisation
@@ -64,7 +64,7 @@ pub enum Location {
     /// pointer. To get the value, we first need to compute the pointer via `rbp - offset` and then
     /// dereference it. For example, the following C-code is likely going to produce an indirect
     /// variable:
-    /// ```
+    /// ```ignore
     /// int x = 0;
     /// control_point()   // Live vars: [x]
     /// printf("%p", &x); // Forces x to be stack allocated.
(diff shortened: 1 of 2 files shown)


yksmp/src/lib.rs
```diff
@@ -59,17 +59,21 @@ pub enum Location {
     /// FIXME: We may need more additional locations in the future, which however will require
     /// rewriting the stackmap format (until now we managed to get by with two extra locations).
     Register(u16, u16, Vec<i16>),
-    /// The live variable is a pointer into the stack. To avoid unnecessary spilling and
-    /// dereferencing LLVM just records the value as an (offset, register) pair where the register
-    /// is typically the base pointer:
-    /// * `u16`: Dwarf register number
+    /// The live variable's value is a pointer that point into the stack. Pointer values may be
```
ltratt, line 62: I think the first sentence is sort-of correct, but it's also what has led both of us up the garden path.

I think the first thing to do is to put `Indirect` *before* `Direct` in the enum, since `Indirect` is the general case. Then `Direct` can say something like "`Direct` is an optimisation for those `Indirect` values that are themselves pointers to things on the stack. In other words, if an `Indirect` value is a pointer, then we can skip a dereference, and turn that `Indirect` into a `Direct`". [That's not very good, but hopefully you get the idea!]

I also think that one more more examples -- even brief sketches -- will help. The problem here is that it's really easy to misread what "pointer of pointer" and "value" and whatnot are. Thanks to your explanation yesterday, I (think I) now understand, but without that, I don't think the text here would quite be enough to kick my brain in the right direction.

yksmp/src/lib.rs
```diff
@@ -59,17 +59,21 @@ pub enum Location {
...
+    /// * `u16`: Dwarf register number. This appears to be always RBP.
     /// * `i32`: offset
     /// * `u16`: size of the value
     Direct(u16, i32, u16),
-    /// The live variable lives on the stack. Similar to the `Direct` location, it's recorded as an
-    /// (offset, register) pair where the register is typically the base pointer. However, to read
-    /// out the value the address described by the (offset, register) pair needs to be dereferenced
-    /// first.
+    /// The live variable lives on the stack. It may have been put there directly via an `alloca`
+    /// or has been spilled. Similar to `Direct`, the location is encoded as an offset relative to
```
ltratt, line 74: I would drop "similar to `Direct`" because `Indirect` is the general case.

ltratt:
Please force push updates if you think it sensible based on my questions / comments.

ptersilie:
Ok to squash?

ptersilie:
Squashed.

PR #1547, opened 2025-01-15
Better aim spills

ltratt:
This PR gradually nudges spilling to a better place: we avoid spillings unnecessarily (https://github.com/ykjit/yk/commit/38419d3498d19f538a7e02ae031994015d9785c0) and we spill things we don't need on the fast (https://github.com/ykjit/yk/commit/d8790939e8732cd07c1967354842e1137cf4681a). Both of these are intended to help with medium to long traces, with the latter potentially significantly reducing pressure on the register allocator. The various other commits are various helpers/cleanups needed to make the main commits possible.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
@@ -556,10 +556,7 @@ impl LSRegAlloc<'_> {
                                 if furthest.is_none() {
                                     furthest = Some((reg, from_iidx));
                                 } else if let Some((_, furthest_iidx)) = furthest {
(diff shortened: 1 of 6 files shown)


ltratt:
This is `Bounce` failing so *probably* this just needs to be retried (unless we've actually broken `Bounce` for good in this PR!).

ltratt:
`early_return2.c` has failed but (somehow) buildbot has mangled stdout/stderr so I can't see why. The parts of the test that *are* output all look correct. Did it segfault or similar? Dunno. I have `try_repeat`ed this test on b16 and it passed successfully over 100 times so it might be something as simple as PT temporarily bombing out and the test not being marked as `rerun-if-`. Let's retry and see what happens?

Issue #1544, opened 2025-01-13
Nondeterministic segfault in bounce.lua


ptersilie:
I've been investigating the nondeterministic segfault we are getting in `bounce.lua` which currently keeps us from merging pull requests. Here's where I got so far.

Here's the diff that relatively quickly tickles out the segfault after only a few runs:
```diff
diff --git a/suites/awfy/Lua/bounce.lua b/suites/awfy/Lua/bounce.lua
index 2817432..80a5930 100644
--- a/suites/awfy/Lua/bounce.lua
+++ b/suites/awfy/Lua/bounce.lua
@@ -79,7 +79,7 @@ function bounce:benchmark ()
         balls[i] = Ball.new(random)
     end

-    for _ = 1, 50 do
+    for _ = 1, 200 do
         for i = 1, #balls do
             local ball = balls[i]
             if ball:bounce() then
@@ -96,4 +96,8 @@ end

 end -- object bounce

+for _ = 1, 10 do
+    bounce:benchmark()
+end
+
 return bounce
 ```

Run this with `YKD_SERIALISE_COMPILATION=0 /home/lukasd/research/yklua_new/src/lua bounce.lua`. If it runs for longer then 200ms you won't see the segfault. It always shows up after about 100ms.

Since this is nondeterministic it's impossible to debug, so the my first goal was to make this deterministic. For this I tried to force the serialised compilation to execute traces and side-traces at exactly the same time as the non-deterministic segfault run would do. For this I first added counters for traces and sidetraces and printed them whenever a trace/side-trace is executed:
```diff
diff --git a/ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs b/ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
index 592d4beeaa8e..41c310a5356a 100644
--- a/ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
@@ -11,7 +11,7 @@ use dynasmrt::Register as _;
 use libc::c_void;
 #[cfg(debug_assertions)]
 use std::collections::HashMap;
-use std::{ptr, slice, sync::Arc};
+use std::{ptr, slice, sync::{atomic::{AtomicU32, Ordering}, Arc}};
 use yksmp::Location as SMLocation;
 
 use super::{X64CompiledTrace, RBP_DWARF_NUM, REG64_BYTESIZE};
[...]

PR #1550, opened 2025-01-16
Fix bounce

vext01:
We've been experiencing non-deterministic segfaults, assertion failures
and Lua-level errors in bounce.lua for a while:
https://github.com/ykjit/yk/issues/1544

After a couple of days of systematically narrowing down the cause, it
appears to be something to do with the re-use of the PT resources in
collect.c.

In theory reusing the PT stuff for subsequent tracing sessions on the
same thread *should* be just fine -- I've even found examples on the
internet where people do this -- however, something is wrong with our
implementation, so I'm disabling the caching for now.

This commit does several things related to this:

 - Fixes a bug in the aux buffer read code in the case that the ring
   buffer wraps around. We were incorrectly bumping the length of the
   recorded trace, which would cause trace corruption in the form of
   uninitialised reads when we come to decode the trace:

   ```diff
   -    trace->len += size + head;
   +    trace->len += head;
   ```

   Later changes (not re-using the aux buffer) will actually mean that
   the wraparound case can't happen (since the aux buffer is the same
   size as the allocated copyout buffer), but I'm leaving it in in case
   we ever decide to re-use perf stuff again.

   Sadly this fix alone isn't enough to fix bounce.lua.

 - Removes caching of perf-related mmap buffers and the perf file
   descriptor itself. This basically reverts https://github.com/ykjit/yk/commit/bedafd9259361840d17eb665547b91b4a993b166. This means each
   tracing session now allocates fresh memory and opens a fresh perf fd.
   As part of this change we re-introduce `PacketParserState::Init` and
   add a couple of other permissible packets that I've seen come up when
   using try_repeat on the test suite.

 - Ensures that the collector's resources are freed by calling
   `hwt_perf_free_collector()`, even if collecting a trace is aborted by
   (e.g.) "trace too long". We were failing to do this, meaning that
   resources would potentially leak. Since we no longer cache the perf
   fd, and this is closed in `hwt_perf_free_collector()`, if we don't
   call that, then the thread is effectively blocked from ever tracing
   again: trying to open perf agaiin would give "resource busy".

This appears to fix the original `bounce.lua` and the reduced one that me an Lukas devised.

I did 3 try_ci runs of this and all suceeded! Fingers crossed.

### DIFF
--- a/.buildbot_dockerfile_debian
+++ b/.buildbot_dockerfile_debian
@@ -19,4 +19,4 @@ RUN useradd -m -u ${CI_UID} ci && chown ${CI_UID}:${CI_UID} .
 ARG CI_RUNNER
 ENV CI_RUNNER=${CI_RUNNER}
 COPY --chown=${CI_UID}:${CI_UID} . .
-CMD sh -x .buildbot.sh
+CMD ["sh", "-x", ".buildbot.sh"]
--- a/hwtracer/src/perf/collect.c
+++ b/hwtracer/src/perf/collect.c
@@ -41,19 +41,6 @@
 // The bit in the IA32_RTIT_CTL MSR that disables compressed returns.
 #define IA32_RTIT_CTL_DISRETC 1 << 11
 
-/*
- * The thread's perf file descriptor and its associated underlying `mmap(2)`
- * regions. The file descriptor is re-used for subsequent trace collections for
- * the same thread.
- *
- * FIXME: These leak when a thread dies.
- */
-static thread_local void *cache_base_buf = NULL;
-static thread_local size_t cache_base_buf_size = 0;
-static thread_local void *cache_aux_buf = NULL;
-static thread_local size_t cache_aux_buf_size = 0;
-static thread_local int cache_perf_fd = -1;
-
 enum hwt_cerror_kind {
   hwt_cerror_unused = 0,
   hwt_cerror_unknown = 1,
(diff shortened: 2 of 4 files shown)


hwtracer/src/perf/collect.c
```diff
@@ -765,6 +742,11 @@ bool hwt_perf_free_collector(struct hwt_perf_ctx *tr_ctx,
   if (tr_ctx->stop_fds[0] != -1) {
     close(tr_ctx->stop_fds[0]);
   }
+  if (tr_ctx->perf_fd >= 0) {
+    /* fprintf(stderr, "close perf\n"); */
```
vext01, line 746: gah.

.buildbot_dockerfile_debian
```diff
@@ -19,4 +19,4 @@ RUN useradd -m -u ${CI_UID} ci && chown ${CI_UID}:${CI_UID} .
 ARG CI_RUNNER
 ENV CI_RUNNER=${CI_RUNNER}
 COPY --chown=${CI_UID}:${CI_UID} . .
-CMD sh -x .buildbot.sh
+CMD ["sh", "-x", ".buildbot.sh"]
```
ltratt, line 22: This doesn't seem necessary?

ltratt:
Please force push an update.

.buildbot_dockerfile_debian
ltratt: Ah, I see why now. Ignore me!

vext01: It fixes a docker warning and is a separate commit.

vext01:
Force pushed the fix. Sorry about that!

PR #1552, opened 2025-01-16
More trace optimisations

ltratt:
This PR implements a number of new trace optimisations (most of which I realised are necessary by tracing some Lua programs with yklua), and tidies up some of the trace optimiser along the way. Each commit should be self-contained.

### DIFF
--- a/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
@@ -415,6 +415,16 @@ impl Module {
         self.insts[usize::from(iidx)] = inst;
     }
 
+    /// Replace the instruction in `iidx` with an instruction that will generate `op`. In other
+    /// words, `Operand::Var(...)` will become `Inst::Copy` and `Operand::Const` will become
+    /// `Inst::Const`. This is a convenience function over [Self::replace].
+    pub(crate) fn replace_with_op(&mut self, iidx: InstIdx, op: Operand) {
+        match op {
+            Operand::Var(op_iidx) => self.replace(iidx, Inst::Copy(op_iidx)),
+            Operand::Const(cidx) => self.replace(iidx, Inst::Const(cidx)),
+        }
+    }
+
     /// Push an instruction to the end of the [Module] and create a local variable [Operand] out of
     /// the value that the instruction defines.
     ///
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -1116,7 +1159,7 @@ mod test {
             "
           entry:
             %0: i8 = 2i8
-            %1: i8 = 1i8
+            %1: i8 = 3i8
```
ptersilie, line 1194: How did this test pass before?

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -415,6 +415,16 @@ impl Module {
         self.insts[usize::from(iidx)] = inst;
     }
 
+    /// Replace the instruction it `iidx` with an instruction that will generate `op`. In other
```
ptersilie, line 418: s/it/in

ykrt/src/compile/jitc_yk/opt/mod.rs
ptersilie: Oh, I confused this with `||`. Nevermind.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ltratt: Fixed in 762413f260011101f75734f14ca64702da0bc210.

ptersilie:
Please squash.

ltratt:
Squashed.

ltratt:
Hmm, with `tryci` only Bounce (surprise, surprise) fails. Maybe try again?

ltratt:
Oof, it seems we are probably hitting a codegen problem with `xor` and `i1`. I think this reflects a deep flaw in our IR where we use byte size instead of bit size. I can fix this, but it won't be quick. My suggestion is that I remove the `Xor` commit (which will cause some conflicts with the next commit), force push, and then I can fix the deeper problem.

ltratt:
OK, so this is actually because we (well... the perpendicular pronoun would be more accurate) didn't consider signed/zero extension in constants in the codegen. I have a partial fix but I need to audit the whole module for other instances. That's a tomorrow job.

ltratt:
https://github.com/ykjit/yk/pull/1553 fixes a problem that happens to manifest in this PR (but isn't caused by this PR) with havlak. When #1553 merges, we should be able to retry this PR, but `Bounce` may still cause us problems.

ltratt:
@ptersilie I think this should now merge.

PR #1553, opened 2025-01-18
Fix generation of immediates for and/or/xor.

ltratt:
We were incorrectly sign extending immediates in these cases, which was particularly obvious in the `i1` case where things like `xor %0, 1i1` would become (in x64) `xor r11d, 0xffffffff`. This commit splits apart zero and sign extension cases more carefully so that `xor r11d, 0x01` is generated.


ltratt:
No -- it's a fix!

vext01:
We spoke about this offline and, although the old generated code harder for a human to read, it appears to be correct.

The new generated code is also correct, so I'm merging this on the grounds that it makes the generated code easier for a human to read.

That it fixes something suggests that some other operation is failing to zero off undefined bits before doing its computation, but @ltratt says he's cooking something up to make it harder for that to happen in the future.

PR #1554, opened 2025-01-20
Deal properly with undefined bits

ltratt:
This PR fixes two cases where we checked whole bytes, instead of a single bit, in generated code.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -1803,7 +1803,7 @@ impl<'a> Assemble<'a> {
         match bit_size {
             0 => unreachable!(),
             32 => dynasm!(self.asm; cmp Rd(lhs_reg.code()), rhs),
-            1..=64 => {
+            8 | 16 | 64 => {
                 if pred.signed() {
                     self.sign_extend_to_reg64(lhs_reg, u8::try_from(bit_size).unwrap());
                 } else {
@@ -1819,7 +1819,7 @@ impl<'a> Assemble<'a> {
         match bit_size {
             0 => unreachable!(),
             32 => dynasm!(self.asm; cmp Rd(lhs_reg.code()), Rd(rhs_reg.code())),
-            1..=64 => {
+            8 | 16 | 64 => {
                 if pred.signed() {
                     self.sign_extend_to_reg64(lhs_reg, u8::try_from(bit_size).unwrap());
                     self.sign_extend_to_reg64(rhs_reg, u8::try_from(bit_size).unwrap());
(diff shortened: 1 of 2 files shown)


vext01:
About the first commit:

> we compared bytes instead of bits, breaking the "the upper bits are undefined" invariant.

Suppose we have two `i1`s, `%lhs` and `%rhs` and we want to compare for equality.

And let's assume that these values are stored in registers and the registers contain `0x1` and `0xffffffffffffffff`. Since only the bottom bit is defined, these numbers both encode `1i1` (when the number is interpreted unsigned).

What the current codegen will do is sign or zero extend (depending on the signedness of the operation) the n-bit values (here n=1) up to a 64-bit value. This is done with `self.{sign,zero}_extend_to_reg64(reg, n)`. Here `n` tells the routine how many bits to consider undefined so that it can replace them with either zeros or ones (again, depending on whether the comparison is signed or not).

In this example `eq` is an unsigned comparison, so `%lhs` is zero extended and remains `0x1` and `%rhs` is zero extended to `0x1`. Then it is the zero extended values that are compared. Here both values are the same `0x1` and thus the comparison is computed correctly to be true.

All this to say: I disagree that the current code breaks the "the upper bits are undefined" invariant. It works by ensuring the undefined bits are defined before doing the 64-bit comparison. Or at least that was the intention when I wrote this code. Have I missed something?

I do agree that special casing `i1` to a bit check is a good optimisation though.

ltratt:
I claim that it is likely that the bug fixed in #1553 was caused by the flaws in this PR. Remember that #1553 fixed a literal crash.

ltratt:
Actually, notice that `cg_guard` did not sign/zero extend/truncate! So it literally is broken before this commit, because it compared a whole byte, and did not do anything with the upper bits.

vext01:
Hang on, we raced.

vext01:
OK, now we are getting somewhere. It looks like `cg_guard` is indeed incorrect. I only verified the general `cmp` case (which I still believe to be correct).

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -2492,8 +2492,16 @@ impl<'a> Assemble<'a> {
                 RegConstraint::Input(inst.falseval(self.m)),
             ],
         );
-        dynasm!(self.asm ; cmp Rb(cond_reg.code()), 0);
```
vext01, line 2495: I agree that this one is comparing undefined bits too!

vext01:
In summary:
 - cg_guard was reading undefined bits
 - cg_select was too
 - cg_icmp was correct, but we've optimised it a bit.

PR #1555, opened 2025-01-20
Be somewhat clearer about upper bits in the codegen.

ltratt:
This is a bit of a hodgepodge of changes, but there are three major things:

  1. Get rid of all the "weird" sized ints, none of which we see in practise, and which we thus have low confidence are correct.
  2. Simplify some of the zero extension stuff.
  3. Zero extend all values to 64 bits before a call. We do this crudely for now.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -11,14 +11,8 @@
 //!     partial register stalls in the CPU pipeline. Generally speaking we prefer to use 64-bit
 //!     operations, but sometimes we special-case 32-bit operations since they are so common.
 //!
-//!   * If an object occupies only part of a 64-bit register, then there are no guarantees about
-//!     the unused higher-order bits. Codegen routines must be mindful of this and, depending upon
-//!     the operation, may be required to (for example) "mask off" or "sign extend into" the
-//!     undefined bits in order to compute the correct result.
-//!
-//! FIXME: the code generator clobbers registers willy-nilly because at the time of writing we have
-//! a register allocator that doesn't actually use any registers. Later we will have to audit the
-//! backend and insert register save/restore for clobbered registers.
+//!   * When a value is in a register, we make no guarantees about what the upper bits are set to.
+//!     You must sign or zero extend at all points that these values are important.
 
 use super::{
     super::{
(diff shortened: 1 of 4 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -11,14 +11,8 @@
...
 //!
-//!   * If an object occupies only part of a 64-bit register, then there are no guarantees about
-//!     the unused higher-order bits. Codegen routines must be mindful of this and, depending upon
-//!     the operation, may be required to (for example) "mask off" or "sign extend into" the
-//!     undefined bits in order to compute the correct result.
-//!
-//! FIXME: the code generator clobbers registers willy-nilly because at the time of writing we have
-//! a register allocator that doesn't actually use any registers. Later we will have to audit the
-//! backend and insert register save/restore for clobbered registers.
+//!   * When a value is in a register, we make no guarantees about what the upper bits are set to.
```
vext01, line 14: I know what you mean here, but we should be clear that the upper undefined bits are *unused* by the value, and that the value is encoded in the lower order bits.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -1844,7 +1856,7 @@ impl Inst {
         Ok(inst)
     }
 
-    /// Returns the size of the local variable that this instruction defines (if any).
+    /// Returns the size in bytes of the local variable that this instruction defines (if any).
```
vext01, line 1859: Should the function be called `def_bytew` for consistency?

vext01:
LGTM. Just a few doc/naming suggestions.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
vext01: I wonder if we should also give an example of a typical scenario where zero/sign extend is necessary. That would help us check we are on the same page and help newcomers to understand what we mean.

My understaning is that:
 - inputs to operations (where the result would be incorrect if we didn't) must define the undefined bits with sign/zero extend (or equivalent).
 - outputs of operations may leave the undefined bits undefined: they may assume that any operation consuming the value will define the upper bits appropriately for the operation in question.

ltratt: The phrasing I'm using is a fairly standard phrasing from specifications. I would prefer to follow that lead rather than introduce our own phrasing that people will not be familiar with.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1662,6 +1649,26 @@ impl<'a> Assemble<'a> {
         Ok(())
     }
 
+    /// When we're about to make a call, we need to clear upper bits in registers.
```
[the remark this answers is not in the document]
ltratt: Good point.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ltratt: I would actually prefer it to be just `bytew` but I have several renaming PRs to do :)

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
vext01: OK.  I would have liked a bit more explanation, but i suppose we have bigger fish to fry.

vext01:
Please squash.

ltratt:
Squashed.

ltratt:
Force pushed a Clippy update.

ltratt:
Force pushed a bit of extra code needed for havlak. Note: the new code correctly masks off upper bits when we spill, which we didn't do before, so this is another case where we now handle `i1` correctly. [Note: bounce doesn't hit this case, so it won't help there. But still...]

vext01:
Nice catch.

My initial thought was "would it not be the responsibility of the operation using the unspilled value to ensure the undefined bits are sanitised?", but I'm not really sure what the rules are.

ltratt:
deopt, for example, doesn't consider undefined bits, so we want to deal with them here. At least for now.

vext01:
> deopt, for example, doesn't consider undefined bits, so we want to deal with them here. At least for now.

That's an interesting observation that I've not considered before: We don't know the assumptions the AOT code generator makes regarding undefined bits. I wonder if our assumptions are compatible with those assumptions.

I'll merge this for now, but I think we may have to think hard about this at some point.

PR #1556, opened 2025-01-22
Minor AOT renamings for consistency with the JIT IR

ltratt:
This is (mostly) a cosmetic PR (other than https://github.com/ykjit/yk/commit/23691c5772a044baa3e9708747364ccba0c35dd9), which we'll need for some other upcoming changes, but which sits nicely on its own. It brings some more parts of the AOT IR into line with the JIT IR. There is of course much more we could do, but every little thing helps!

### DIFF
--- a/ykrt/src/compile/jitc_yk/aot_ir.rs
+++ b/ykrt/src/compile/jitc_yk/aot_ir.rs
@@ -1613,13 +1613,7 @@ impl IntegerTy {
     /// Padding for alignment is not included.
     #[cfg(test)]
     pub(crate) fn byte_size(&self) -> usize {
-        let bits = self.bitw();
-        let mut ret = bits / 8;
-        // If it wasn't an exactly byte-sized thing, round up to the next byte.
-        if bits % 8 != 0 {
-            ret += 1;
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/aot_ir.rs
```diff
@@ -1591,29 +1591,29 @@ pub(crate) fn const_int_bytes_to_string(num_bits: u32, bytes: &[u8]) -> String {
 #[deku_derive(DekuRead)]
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub(crate) struct IntegerTy {
-    pub(crate) num_bits: u32,
+    bitw: u32,
```
vext01, line 1594: May as well add a doc string while we are here.

vext01:
LGTM with one suggestion.

ykrt/src/compile/jitc_yk/aot_ir.rs
ltratt: You're right, but I'd rather not do that here. Fish, frying.

PR #1557, opened 2025-01-23
Change how we store and sign/zero extend integers.

ltratt:
Previously we stored raw `u64`s and expected the user to remember that they needed to zero/sign extend the underlying integer whenever they wanted to do anything with that. We did not always do this, and we did it incorrectly in a couple of places!

This commit introduces a new struct `ArbBitInt` which is basically a pair `(bit_width: u32, value: u64)` which hides the underlying value. To get a raw Rust-level integer you have to call a method, and those methods have names with `sign_extend` and `zero_extend` in them. While one can, of course, call the wrong one, it is now impossible not to sign/zero extend. This struct is currently rather simple, but the API is flexible enough to extend to beyond-64-bit ints transparently. The (fairly extensive) test suite is a bit overkill right now, but is intended to help give us confidence if/when we support more than 64 bit ints in the future.

This commit also necessarily requires a full audit of everything to do with ints-in-traces. That means a lot of code churn, but it's absolutely necessary, and (a) makes much clearer where we should sign/zero extend (b) catches some places where we didn't do this but should.

This commit isn't perfect. In particular, I'm not very happy that `Const::Int` has both a `TyIdx` that contains a bit width *and* an `ArbBitInt` that separately records a bit width. That feels icky, but doing something neater will require at least some ickiness elsewhere. I'll worry about that another day.

### DIFF
--- a/ykrt/src/compile/jitc_yk/arbbitint.rs
+++ b/ykrt/src/compile/jitc_yk/arbbitint.rs
@@ -1,0 +1,840 @@
+//! An integer of an arbitrary, dynamic, bit width.
+//!
+//! This hides away the underlying representation, and forces the user to consider whether they
+//! want to zero or sign extend the underlying the integer whenever they want access to a
+//! Rust-level integer.
+//!
+//! Currently only up to 64 bits are supported, though the API is flexible enough to transparently
+//! support greater bit widths in the future.
+
+use super::int_signs::{SignExtend, Truncate};
+use std::{
+    hash::Hash,
+    ops::{BitAnd, BitOr, BitXor},
+};
+
+/// An integer of an arbitrary, dynamic, bit width.
+///
+/// Currently can only represent a max of 64 bits: this could be extended in the future.
+#[derive(Clone, Debug)]
+pub(crate) struct ArbBitInt {
+    bitw: u32,
+    /// The underlying value. Any bits above `self.bitw` have an undefined value: they may be set
+    /// or unset.
+    ///
+    /// Currently we can only store ints that can fit in 64 bits: in the future we could use
+    /// another scheme to e.g `Box` bigger integers.
+    val: u64,
+}
+
+impl ArbBitInt {
+    /// Create a new `ArbBitInt` that is `width` bits wide and has a value `val`. Any bits above
+    /// `width` bits are ignored (i.e. it is safe for those bits to be set or unset when calling
+    /// this function).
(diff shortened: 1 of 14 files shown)


vext01:
We are talking only about constant integers here, right? I think we store local variables in as many bytes as is required (not always a u64).

ltratt:
Yes, more-or-less trace-level `Const::Int`s, though there are some variations on this that make me a bit reluctant to just say "constants".

ykrt/src/compile/jitc_yk/opt/analyse.rs
```diff
@@ -56,16 +56,18 @@ impl Analyse {
...
+                            (&lhs, &rhs)
                         {
-                            if pred == Predicate::Equal && m.const_(lhs_cidx) == m.const_(rhs_cidx)
-                            {
-                                self.values.borrow_mut()[usize::from(iidx)] =
-                                    Value::Const(m.true_constidx());
-                                return Operand::Const(m.true_constidx());
-                            }
+                            todo!();
+                            // let pred = inst.predicate();
```
ptersilie, line 64: What's are these comments about? Do you want to keep them?

ykrt/src/compile/jitc_yk/arbbitint.rs
```diff
@@ -0,0 +1,840 @@
...
+impl Hash for ArbBitInt {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.bitw.hash(state);
+        self.val.truncate(self.bitw).hash(state);
+    }
+}
+
+impl PartialEq for ArbBitInt {
+    fn eq(&self, other: &Self) -> bool {
+        self.bitw == other.bitw && self.val.truncate(self.bitw) == other.val.truncate(self.bitw)
```
vext01, line 249: I wonder if comparing different bit sizes should panic here?

ltratt: I think because we can hash `Const`s that wouldn't be the right thing to do here: we will end up comparing different bit sized things.

vext01: Ah, right.

vext01:
LGTM.

This API is basically what LLVM does with `.getZExtValue()` and `.getSExtValue()` :)

vext01:
Please squash.

PR #1558, opened 2025-01-24
Break the optimiser into separate functions.

ltratt:
The optimiser has now exceeded even my (high!) tolerance for long functions. Previously it was structured as:

```rust
match inst {
  Inst::A => {
    ...
  }
  Inst::B => {
    ...
  }
}
```

where the `...`s could be arbitrarily long.

This commit -- which has no functional change, despite its size! -- changes the structure to:

```rust
match inst {
  Inst::A => self.opt_a(...),
  Inst::B => self.opt_b(...),
}
```

So now each instruction has its "own" function. `opt_binop` is still fearsomely large, but this commit moves us in a more sustainable direction.

### DIFF
--- a/ykrt/src/compile/jitc_yk/opt/mod.rs
+++ b/ykrt/src/compile/jitc_yk/opt/mod.rs
@@ -9,8 +9,9 @@
 use super::{
     arbbitint::ArbBitInt,
     jit_ir::{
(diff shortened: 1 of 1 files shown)


vext01:
Bravo!

ltratt:
Bounce failed. Unlikely to be related to this PR! Can we retry please?

PR #1561, opened 2025-01-24
Don't allow signed pointer comparison.

ltratt:
The LLVM manual is unclear as to whether signed pointer comparisons are allowed. I can't see that they make sense, and I haven't yet encountered such things, so I prefer to `panic` if we do see them: this means that I no longer need to worry about zero/sign extending pointers in the optimiser.

[Note: `ptrdiff_t`-esque things clearly can be signed. But I'm not sure that pointers per se make sense as being signed.]


vext01:
I agree! :)

PR #1562, opened 2025-01-24
Abort PT trace decoding if the "taken" bug arises.

vext01:
I'm working on a proper fix, but in the meantime this ensures we don't construct nonsense traces in release mode builds.


ptersilie:
I think we might want to also increase the file descriptor size, because I had rebench fail occasionally without it. `2048` worked for me.

PR #1565, opened 2025-01-26
Start letting through some more LLVM intrinsics disguised as calls

ltratt:
Some of these are no-ops (https://github.com/ykjit/yk/commit/c42bc99683cdaf964bf9e9e45523cf210471a528), some require codegen (https://github.com/ykjit/yk/commit/e3f549d7a8118a17203bc8fab3a7a4a36c5dfe67).


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1499,9 +1499,15 @@ impl<'a> Assemble<'a> {
             .collect::<Vec<_>>();
 
         // unwrap safe on account of linker symbol names not containing internal NULL bytes.
-        let va = symbol_to_ptr(self.m.func_decl(func_decl_idx).name())
-            .map_err(|e| CompilationError::General(e.to_string()))?;
-        self.emit_call(iidx, fty, Some(va), None, &args)
+        match self.m.func_decl(func_decl_idx).name() {
```
vext01, line 1502: Why don't we just filter these away at serialisation time?

ltratt: We *might* be able to make use of the information -- it's not yet clear to me.

In the interim, their presence in a trace prevents that trace being compiled. So this is a quick fix for that; and, perhaps, we can do something more appropriate later.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1503,6 +1503,11 @@ impl<'a> Assemble<'a> {
             "llvm.assume" => Ok(()),
             "llvm.lifetime.start.p0" => Ok(()),
             "llvm.lifetime.end.p0" => Ok(()),
+            x if x.starts_with("llvm.smax") => {
```
vext01, line 1506: Do we know for sure that `llvm.smax` is never inlined at the MIR level?

ltratt: I see no evidence of that, but really it's beyond my purview.

vext01: The problem is, if an intrinsic is inlined at the codegen level, then we trace the inlined version and then call it a second time.

That's probably OK (if a little iffy) for an operation like smax, but you certainly don't want to do that for operations that are expensive (in terms of performance, since you would incur the cost twice) or have side effects (which would break program semantics).

I'd like to hear what @ptersilie thinks, but my gut feeling is to not merge this until we have a better way of knowing whether an intrinsic has been inlined or not.

vext01:
We had an offline discussion and think this is safe.

ltratt:
I currently cannot replicate this. Shall we retry?

ltratt:
Closing this as so much has changed elsewhere that this PR needed redoing from scratch.

PR #1567, opened 2025-01-27
Don't give up on a location the first time compiling fails.

ltratt:
This comment in the code was partly wrong:

```
// FIXME: We will endless retry tracing this
// [HotLocation]. There should be a `Counting`
// state.
```

In fact, we wouldn't endlessly retry: the `HotLocation` would get stuck in the `Compiling` state. Still, the end effect was the same as the comment suggested!

This commit allows `HotLocation`s to transition back to the `Counting` state so that we can retry. That requires allowing `HotLocations` to also be in the `Counting` state: in other words, we can "count" both with and without a `HotLocation`.

### DIFF
--- a/ykrt/src/location.rs
+++ b/ykrt/src/location.rs
@@ -42,48 +42,71 @@ pub struct Location {
     /// A Location is a state machine which operates as follows (where Counting is the start state):
     ///
     /// ```text
-    ///              ┌──────────────┐
-    ///              │              │─────────────┐
-    ///   reprofile  │   Counting   │             │
-    ///  ┌──────────▶│              │◀────────────┘
-    ///  │           └──────────────┘    increment
-    ///  │             │                 count
-    ///  │             │ start tracing
-    ///  │             ▼
-    ///  │           ┌──────────────┐
-    ///  │           │              │ incomplete  ┌─────────────┐
-    ///  │           │   Tracing    │────────────▶│  DontTrace  │
-    ///  │           │              │             └─────────────┘
-    ///  │           └──────────────┘
-    ///  │             │ start compiling trace
-    ///  │             │ in thread
-    ///  │             ▼
-    ///  │           ┌──────────────┐             ┌───────────┐
-    ///  │           │  Compiling   │────────────▶│  Dropped  │
-    ///  │           └──────────────┘             └───────────┘
-    ///  │             │
-    ///  │             │ trace compiled
-    ///  │             ▼
-    ///  │           ┌──────────────┐
-    ///  └───────────│   Compiled   │◀────────────┐
-    ///              └──────────────┘             │
-    ///                │                          │
-    ///                │ guard failed             │
-    ///                ▼                          │
-    ///              ┌──────────────┐             │
-    ///              │  SideTracing │─────────────┘
-    ///              └──────────────┘
+    ///                                           │
+    ///                                           │
+    ///                                           ▼
+    ///                                         ┌─────────────────────────────────────────────────────────────────────────────────────┐   increment count
+    ///                                         │                                                                                     │ ──────────────────┐
+    ///                                         │                                      Counting                                       │                   │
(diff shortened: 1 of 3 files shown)


ykrt/src/location.rs
```diff
@@ -243,6 +269,7 @@ pub(crate) enum HotLocationKind {
     /// A trace for this HotLocation is being compiled in another trace. When compilation is
     /// complete, the compiling thread will update the state of this HotLocation.
     Compiling,
+    Counting(HotThreshold),
```
vext01, line 272: Is the thing inside the threshold at which to give up, or the number of times we've failed to compile this location? I think it's the latter despite the type name?

ltratt: There are two "sort of counting" things:

1. "How many times have we encountered this location" is the `Counting` state
2. "How many times have we failed to trace/compile this location" is `HotLocation::tracecompilation_errors`.

In this case, while we failed fewer than "trace compilation errors threshold" we reset `Counting` to 0 and let it naturally be retraced/compiled.

In that sense, one could argue that `Counting` is somewhat misleading: it really means "how many times have we encountered this location since the count was last set to 0". But I think it's quite a nice, and very simple, back-off heuristic.

vext01: Right, so having read a bit:
 - 1. is before the location became hot
 - 2. is a separate counter inside the hot location.

The way I'm reading this field is `CountingHavingFailed`, is that right?

I got a bit confused because I wasn't sure if this new `Counting` is the same as the other counting. I guess the answer is "kind of".

Would it be OK to add a quick docstring?

ykrt/src/mt.rs
```diff
@@ -629,6 +631,21 @@ impl MT {
                             }
                         }
                         HotLocationKind::Compiling => TransitionControlPoint::NoAction,
+                        HotLocationKind::Counting(c) => {
+                            if is_tracing {
+                                // This thread is tracing something, so bail out as quickly as possible
+                                TransitionControlPoint::NoAction
+                            } else {
+                                if c < self.hot_threshold() {
```
vext01, line 639: So we retry compiling a trace as many times as the hot threshold that triggered the first compilation? I wonder if there should be a separate threshold at some point?

ltratt: > So we retry compiling a trace as many times as the hot threshold that triggered the first compilation?

No, we retry `TraceFailed` times. Each of those is spaced out by `HotThreshold` times.

vext01: > we retry TraceFailed times

At the risk of sounding like dummy: I'm not sure what you mean.

`TraceFailed` is a type:
```
pub(crate) enum TraceFailed {
    KeepTrying,
    DontTrace,
}
```

ltratt: Oops, I meant `TraceCompilationErrorThreshold.

ykrt/src/location.rs
ltratt: Oops, there should have been a docstring, you're quite right! Fixed in 2a32832a.

ykrt/src/mt.rs
vext01: OK, I see it. That logic was already in place, but I think we are saying that due to a bug (getting stuck in `Compiling`) it wasn't used properly.

ltratt: It was used in one place but not in another (there are lots of ways that things can go wrong :)). This PR fixes the latter case.

ykrt/src/location.rs
vext01: Thanks!

ltratt:
OK to squash?

vext01:
Oh yeah, sorry. Please go ahead.

ltratt:
Squashed.

ltratt:
Force pushed Clippy fixes (oops).

PR #1568, opened 2025-01-27
Fix long-standing PT "taken" bug.

vext01:
For a while now we've noticed fishy behaviour in the PT decoder, where a compressed return is detected, but the following TNT bit is not one, as it should be.

We also know that carrying on after seeing this causes nonsense traces. For this reason we recently added code to detect the bug and abort tracing.

The reason for the bug is that `seek_tnt_or_tip()`, whose job it is to read packets until the next control flow event, would return any packet that carries a TIP (target instruction pointer) payload, even if the payload was "out of context" (PT terminology for "unknown TIP").

To follow the CFG properly, some (not all) call-sites to `seek_tnt_or_tip()` will be uninterested in out of context TIP updates and will want to ignore them.

Critically `is_compressed_return()` *must* skip out of context TIPs, since they have no bearing on whether the return is compressed or not. This was the underlying cause of the taken bug.

Rather than forcing call-sites to loop if they get an out of context TIP when they want to skip them, this change adds a `skip_ooc` flag to `seek_tnt_or_tip()` which determines whether it will skip OOC TIPs or not.

Adding the flag also forces us to think about whether we should be skipping such packets or not.

[Since compressed returns are currently disabled at tracing time, we should in fact never decode a compressed return. Due to this I've replaced a chunk of compressed return logic with an unreachable!()]

### DIFF
--- a/hwtracer/src/pt/ykpt/mod.rs
+++ b/hwtracer/src/pt/ykpt/mod.rs
@@ -436,27 +436,17 @@ impl YkPTBlockIterator<'_> {
             // *compressed* return.
             true
         } else {
-            // This *may* be a compressed return. If the next event packet carries a TIP update
-            // then this was an uncompressed return, otherwise it was compressed.
-            let pkt = self.seek_tnt_or_tip()?;
+            // This *may* be a compressed return. If the next (non-out-of-context) event packet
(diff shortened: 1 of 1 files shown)


ltratt:
Excellent!

PR #1569, opened 2025-01-28
Allow longer traces

ltratt:
This PR is as much a RFC as "let's merge this". Right now, we reject a lot of traces as being "too long". There are two current causes for this:

1. We exceeded the PT buffer.
2. We exceeded yk's willingness to compile the trace.

It's easy to think these two cases are the same, but they're not. With `YK_LOG=3` the first case will be detected as stop-tracing time; the second case at compilation time.

This PR only tackles the second case: it allows yk to compile longer traces. First, I think we should give up on the 16/15-bit `InstIdx`s, at least for now: it's a very small type, and probably too restrictive. https://github.com/ykjit/yk/commit/04d3e8ae22fae3f03917f9da4ad562929088b989 changes this to 32/31 bits. Note: `Inst`s now become two machine words big. I think this is fine for now.

That then allows us to increase the size of traces we'll compile. How high is too high? We don't yet know, but it seems worth it to me to substantially increase the size 20,000 is still, I suspect, a fairly conservative limit https://github.com/ykjit/yk/commit/efad15801d691cb6eb90285947ab0dd8109178ba. But we're going to have to experiment with longer traces in order to know "how long is too long", so perhaps we should go higher, at least temporarily.

### DIFF
--- a/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
@@ -1,5 +1,6 @@
 //! The Intermediate Representation (IR) of a trace.
 //!
+//!
 //! ## General SSA properties
 //!
 //! The most important part of the IR is the
@@ -730,10 +731,10 @@ impl GlobalDecl {
 ///  * `*_SHIFT`: the number of bits required to left shift a field's value into position (from the
 ///    LSB).
 ///
-const OPERAND_IDX_MASK: u16 = 0x7fff;
+const OPERAND_IDX_MASK: u32 = (1 << 23) - 1;
 
-/// The largest operand index we can express in 15 bits.
-const MAX_OPERAND_IDX: u16 = (1 << 15) - 1;
+/// The largest operand index we can express in 23 bits.
+const MAX_OPERAND_IDX: u32 = (1 << 23) - 1;
 
 /// The symbol name of the global variable pointers array.
 #[cfg(not(test))]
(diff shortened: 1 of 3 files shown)


ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -3143,7 +3201,7 @@ mod tests {
     /// Ensure that any given instruction fits in 64-bits.
     #[test]
     fn inst_size() {
-        assert!(mem::size_of::<Inst>() <= mem::size_of::<u64>());
```
vext01, line 3146: Docstring above needs updating.

vext01:
I don't object to this because we do indeed reject a lot of traces.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
vext01: Me may as well say our instructions are `u128` sized rather than "two 64-bit sized"

ptersilie:
Works for me too.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ltratt: Fixed in the force push.

ltratt:
OK so it seems after some local testing that the only way to get this to explode is to make swt take a *very* long time. I am inclined to `ignore` the test. OK to push on that basis?

ltratt:
Ah, hang on, it's ignored on swt, so my measuring that was pointless :man_facepalming: Let me see if I can find a "big enough" value that's not *too* slow...

ltratt:
I finally worked out what I was doing wrong: the `trace_too_long` test needed two constants updating. Fixed in the force push.

PR #1574, opened 2025-01-30
Fix all clippy lints and update deku.

vext01:
No functional change.

Many of the lints came from deku macros, so to do a proper job of fixing the lints I had to upgrade deku and the API had changed quite a bit.

Namely certain parts of the API are now byte- (not bit)-centric.

Newer dekus also claim to be more performant:
https://github.com/sharksforarms/deku/blob/master/CHANGELOG.md


ltratt:
Ouch! Still, it's a job better done than not.

Issue #1571, opened 2025-01-29
meta_framegap.lua deopt crash


vext01:
With today's yklua (e99ff2) and yk (3dc428) an OpenResty test causes a deopt crash:

```
$ YK_OPT=0 YKD_SERIALISE_COMPILATION=1 ~/research/yklua/src/lua luajit2-test-suite/test/misc/meta_framegap.lua
thread '<unnamed>' panicked at ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs:40:30:
no entry found for key
```

To get the test file, you can use the branch from [this PR](https://github.com/ykjit/yklua-tests/pull/1)

(Once it's merged, you can just use `main`)

vext01:
It's very deterministic, so I have high hopes of reducing it and tracking down the cause.

vext01:
More info on this crash.

It is 100% deterministic with `YK_OPT=0 YKD_SERIALISE_COMPILATION=1` which helps.

There are no side-traces involved. There are two traces executed, call them T1 and T2.

T1 has 83 deopt exits. T2 has only 11.

T1 is executed once at the beginning, then only T2 is executed after that (many times).

Here is a reduced test case:
```lua
A = setmetatable({}, {
	__add = function(A, B)
		if B > 200 then
			for B = 1, B do
			end
		end
	end,
})
function C(A, B)
	do
		return A + B
	end
end
for B = 1, 300 do
	C(A, B)
end
```

Putting prints in `__yk_guardcheck()` the crash appears to be caused when we look up guard index 56 in a trace (T1) with guard indices 0..10:
```
guard index 56
ctr has these guard indices: [4, 0, 7, 10, 1, 2, 8, 9, 3]
```

Looking at the jit-asm confirms 56 is well out of bounds, but there is another compiled trace for which that'd be a valid index (i.e. T2). This made me question whether the thread's idea of the "currently running trace". Is it using a guardindex for T2 when running T2?

Putting more prints around and pumping up the logging I can see something else fishy. A common pattern in the output is:
```
...
yk-jit-event: enter-jit-code
set a running trace: 0x7fa1851d9490
root is None
__yk_guardcheck()
ctr addr: 0x7fa1851d9490
guard index 8
ctr has these guard indices: [4, 0, 7, 10, 1, 2, 8, 9, 3]
yk-jit-event: deoptimise
replace stack
control point
yk-jit-event: enter-jit-code
...
```

That all makes sense to me.

[...]

PR #1575, opened 2025-01-30
Add third party yklua tests to CI.

vext01:
This currently fails due to [this deopt bug](https://github.com/ykjit/yk/issues/1571), hence this is a draft.

A similar change should be applied to `yklua` so that these tests are run in CI there too.

I'll start investigating that deopt bug tomorrow.


ltratt:
@vext01 I think we can now consider merging this one?

vext01:
I think so :)

ltratt:
OK, please undraft it?

ltratt:
I've removed these from the queue because these have so far spent an hour running in CI, which is clearly vastly too long for us to want to run for each PR.

vext01:
Re-visiting this, just noticed that we only test yklua with serialised compilation. I followed precedent in this PR, but is that what we want?

ltratt:
Serialised compilation is a choice with pros and cons. In general it forces more things to be tested, but "in general" is the key qualifier.

vext01:
Oops. I was wrong. We test both modes.

If you are happy with that, please merge.

ltratt:
The reason I cancelled the merge is that these tests took over an hour to run. We can't merge them until we know they run in reasonable time. 

It might be that these long-running tests should be run overnight, or after a PR's merged, or something like that?

vext01:
I must admit I don't know how long the tests take to run since your changes last week. Will measure.

vext01:
I've merged master and added a commit that runs the extra tests only in release mode in the interest of saving time. That should be an OK compromise.

One of the tests hits the ">32-bit constant" todo that is fixed as part of https://github.com/ykjit/yk/pull/1606, so we will need to wait for this first.

OK to squash in the meantime?

vext01:
Currently the extra tests are run only in release mode. With your most recent change, do we want to run in the `release-with-assert` profile in addition?

ltratt:
There is no `release-with-assert` profile here. It's probably best to run this in `debug` mode. I suggest doing that but measuring the effect on `tryci` on b16 (which is currently ~9 minutes): how much does it slow down by?

vext01:
Currently these additional tests just take far too long to add to CI, so I'm closing this for now. At some point I will make an overnight job for these tests instead.

PR #1577, opened 2025-02-03
Remove `root` attribute from `Executing`.

ltratt:
I am not sure if/when this was ever used, but in the current code the attribute is only ever to `None`, and thus has no meaningful effect on behaviour.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -434,7 +434,7 @@ impl MT {
                 }
                 let trace_addr = ctr.entry();
                 MTThread::with(|mtt| {
-                    mtt.set_running_trace(Some(ctr), None);
+                    mtt.set_running_trace(ctr);
                 });
                 self.stats.timing_state(TimingState::JitExecuting);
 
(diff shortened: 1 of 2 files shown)


vext01:
Me an Lukas were discussing removing this as well. It sure confused me last week!

ykrt/src/mt.rs
```diff
@@ -1020,31 +1018,16 @@ impl MTThread {
...
-    pub(crate) fn set_running_trace(
-        &self,
-        ctr: Option<Arc<dyn CompiledTrace>>,
-        root: Option<Arc<dyn CompiledTrace>>,
-    ) {
-        *self.tstate.borrow_mut() = match ctr {
-            Some(ctr) => MTThreadState::Executing { ctr, root },
-            None => MTThreadState::Interpreting,
-        };
+    pub(crate) fn set_running_trace(&self, ctr: Arc<dyn CompiledTrace>) {
```
vext01, line 1029: We've removed the ability to set no running trace (by passing `None`) with this function, but the docstring implies that's still possible.

Either the docsting needs updating, or the parameter should accept `Option<...>`.

Is there still just one call site to this code? Should we inline it?

ltratt: Removed the second clause of the sentence in the force push.

PR #1578, opened 2025-02-04
Allow tracing recursion.

ltratt:
This test previously hit the (incorrect) `assert` that this commit removes.


ltratt:
Force pushed `cargo fmt` fix (oops)..

PR #1581, opened 2025-02-04
Abort traces finishing in frames other than the ones they started in.

### LINKED ISSUE #1579: Calling an interpreter loop twice can lead to infinite loops?
Consider this simple C test:

```c
#include <assert.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <yk.h>
#include <yk_testing.h>

void loop(YkMT *, YkLocation *, int);

void loop(YkMT *mt, YkLocation *loc, int i) {
  NOOPT_VAL(i);
  while (i > 0) {
    yk_mt_control_point(mt, loc);
    fprintf(stderr, "%d\n", i);
    if (i == 5)
      loop(mt, loc, i - 1);
    i--;
  }
  fprintf(stderr, "return\n");
  return;
}

  YkMT *mt = yk_mt_new(NULL);
  yk_mt_hot_threshold_set(mt, 1);
  yk_mt_sidetrace_threshold_set(mt, 2);

  loop(mt, &loc, 6);
  loop(mt, &loc, 6);
  fprintf(stderr, "exit\n");
  yk_location_drop(loc);
  yk_mt_shutdown(mt);
  return (EXIT_SUCCESS);
}
```

It loops infinitely. If I remove the second call to `loop` in `main` (line 71ish), things work as expected.

Is there a requirement for us to not "re-call" `yk_mt_control_point` or is this a flat our bug?

ptersilie:
Fixes #1579

### DIFF
--- a/tests/c/tracing_recursion1.c
+++ b/tests/c/tracing_recursion1.c
@@ -6,38 +6,11 @@
 //     6
 //     yk-jit-event: start-tracing
 //     5
-//     yk-jit-event: stop-tracing
-//     --- Begin jit-pre-opt ---
+//     yk-jit-event: tracing-aborted
 //     ...
-//     --- End jit-pre-opt ---
-//     4
-//     yk-jit-event: enter-jit-code
-//     3
-//     yk-jit-event: deoptimise
-//     yk-jit-event: enter-jit-code
-//     2
-//     yk-jit-event: deoptimise
-//     yk-jit-event: start-side-tracing
-//     yk-jit-event: stop-tracing
-//     --- Begin jit-pre-opt ---
-//     ...
-//     --- End jit-pre-opt ---
-//     1
-//     return
-//     yk-jit-event: enter-jit-code
-//     4
-//     yk-jit-event: execute-side-trace
-//     3
-//     yk-jit-event: execute-side-trace
-//     2
-//     yk-jit-event: execute-side-trace
-//     1
-//     yk-jit-event: execute-side-trace
-//     yk-jit-event: deoptimise
-//     return
-//     exit
 
-// Check that recursive tracing doesn't hit an error.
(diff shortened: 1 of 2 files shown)


ptersilie:
Ah, it fails the test from https://github.com/ykjit/yk/pull/1578. I need to fix the test and make sure it still tests what it's supposed to.

ptersilie:
Hm, I think this PR makes https://github.com/ykjit/yk/pull/1578 redundant, because we now ensure that the stack frames are the same _before_ we hit the StopTracing part from the other PR here: https://github.com/ykjit/yk/pull/1578/files#diff-9fa85f3ab0ab08c2bb4337dd17d47605f520c2780158e50e66eff7cd9dc1b29bR496

ltratt:
Do you mean #1578 makes this PR redundant or ... ?

ptersilie:
No, I mean this PR makes both the fix and test in https://github.com/ykjit/yk/pull/1578 redunant as I don't believe we reach that point in the code with different frame pointers anymore, since in this PR we abort before we transition to `StopTracing` if the two frameaddr's don't match.

ltratt:
It can't make the test from #1578 redundant, surely? Though it could change the outcome of the test. But if you believe this PR is superior, you could (in essence) revert #1578 and then do what you think is best (whilst retaining, and modifying as appropriate, the test from #1578).

ltratt:
Isn't it better to let it through rather than abort it?

ptersilie:
We've all agreed yesterday that for the time being we abort traces that finish in different frames than the one they started in. Unless you changed your mind about this.

ltratt:
Point taken. Let's do it.

ptersilie:
Ok to squash?

ltratt:
Please squash.

ptersilie:
Squashed.

PR #1582, opened 2025-02-04
Track nested meta-tracing states.

### LINKED ISSUE #1571: meta_framegap.lua deopt crash

ltratt:
Previously we -- well, this one is entirely my fault, so "I"! -- tracked per-thread meta-tracing state as a single `MTThreadState` that we updates as necessary. This doesn't work when we have nested execution, tracing and the like, as the bug in #1571 highlighted.

This commit moves us to a stack of `MTThreadState`s. The basic idea is that the stack always has at least one element: `Interpreting`. As we go through other states, we push / pop as appropriate. The test below (from Edd and Lukas) fails on `master` but is fixed by this commit.

The implementation is a bit more awkward than one might hope as naive implementations either:

  1. Spread lots of knowledge about the stack around the code. That's a disaster waiting to happen.
  2. Run into borrow checker problems.

This commit gets around this in two phases:

  1. We pass closures to `MTThread` which can peek at the current `MTThreadState` (which isn't `Copy`!).
  2. Those closures return the "things" we need to update outside that context.

This is a bit awkward, and perhaps there's a better API, but this one is at least safe.

Fixes #1571.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -1,7 +1,7 @@
 //! The main end-user interface to the meta-tracing system.
 
 use std::{
-    assert_matches::debug_assert_matches,
+    assert_matches::{assert_matches, debug_assert_matches},
     cell::RefCell,
     cmp,
     collections::VecDeque,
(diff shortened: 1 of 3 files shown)


ykrt/src/mt.rs
```diff
@@ -867,6 +877,12 @@ impl MT {
         }
     }
 
+    /// Inform this `MT` instance that `deopt` has occurred: this updates the stack of
+    /// [MTThreadState]s.
+    pub(crate) fn deopt(self: &Arc<Self>) {
+        MTThread::with(|mtt| mtt.pop_tstate());
```
vext01, line 883: shall we assert that the top of the stack is `Executing` here for sanity?

vext01:
I'm OK with this, but @ptersilie should have a look before we merge.

ykrt/src/mt.rs
ltratt: Fixed in 116ba0371f9c1f5a03b5b83ee986907ca4065a0b.

vext01: Thanks!

ptersilie:
Looks good to me.

vext01:
If you are ready @ltratt, please squash.

ltratt:
Squashed.

PR #1583, opened 2025-02-05
Catch another case of problematic control flow for tracing.

vext01:
This change catches the case where we exit the function that started tracing, but re-enter it later and stop tracing. This is problematic for tracing, as evidenced by trace builder crashing with variable mapping problems (it doesn't know about variables in the parent function).

We already have a check in place that was designed to detect "exiting from the function that started tracing", but it's not sufficient for the above scenario: it checks that the frame addresses at the time we start and stop tracing are the same. Here they are the same.

To fix this, while building the trace, we check if a `Inst::Ret` would cause the trace builder's frame stack to become empty.

It's still worth keeping the old check around, since they detect a subset of the "bad returns" earlier than this new check.

### DIFF
--- a/tests/c/exit_and_reenter_interploop.c
+++ b/tests/c/exit_and_reenter_interploop.c
@@ -1,0 +1,50 @@
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YK_LOG=4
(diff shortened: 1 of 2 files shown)


ptersilie:
Looks good to me.

PR #1585, opened 2025-02-06
Intrinsics as calls

ltratt:
This PR ignores / codegens several LLVM intrinsics that are disguised as `call`s. These crop up frequently in traces, and cause us to abort compiling.


ltratt:
I have seen the `ykd_opt_off` fail once before and I haven't been able to replicate it since. Nothing in this PR effects it (unless I've missed something very obvious!) so I think we should retry, but we should keep an eye on `ykd_opt_off` for other reasons.

PR #1588, opened 2025-02-06
Explicitly flush `stdout`.

ltratt:
Once in a blue moon this test fails because nothing appears on `stdout`. This is technically allowed, I believe, because unless we flush `stdout` explicitly, non-interactive `stdout` is not guaranteed to be flushed, even on (normal or abnormal) program exit.

If the test fails after this change, it suggests there's a "real" problem.


ltratt:
That failure is unrelated to the PR. Please retry?

PR #1591, opened 2025-02-06
Implement side-trace patching.

ptersilie:
This commit implements patching side-traces directly into their parent trace. This gets rid of a good chunk of overhead that was required to find and execute the correct side-trace. With this change, executing a side-trace is all but a simple jump.

Some notable performance results (running with `YKD_SERIALISE_COMPILATION=1` to make sure the differences aren't down to variance of which guards are being compiled):

Richards, Queens, Bounce: 1.3x speedup
Sieve: 1.5x speedup.


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -425,24 +414,57 @@ impl<'a> Assemble<'a> {
...
                     self.asm.offset(),
-                    format!("Deopt ID for guard {:?}", deoptid),
+                    format!("Deopt ID and patch point for guard {:?}", deoptid),
                 );
                 // FIXME: Why are `deoptid`s 64 bit? We're not going to have that many guards!
+
+                // Word-align this location so we can patch this later in a single atomic
+                // operation.
+                let off = self.asm.offset();
+                let align = 8 - (off.0 % 8);
```
vext01, line 429: If we are already aligned, doesn't this push 8 superfluous nops?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -2878,6 +2874,58 @@ impl CompiledTrace for X64CompiledTrace {
...
+        // FIXME: This might not work on other architectures or operating systems. However, the
+        // solution is very involved and requires changing `mt.rs` to make sure no other thread is
+        // accessing the trace (which includes executing the trace itself or any of it's
+        // side-traces, or side-tracig a guard failure).
+        let trace_addr = self.buf.ptr(AssemblyOffset(0)) as *mut libc::c_void;
+        let res = unsafe {
+            libc::mprotect(
+                trace_addr,
+                self.buf.len() as libc::size_t,
+                libc::PROT_WRITE | libc::PROT_EXEC,
```
vext01, line 2891: Asides from the concurrency issues mentioned above, we also shouldn't allocate pages W+X. Can we put a FIXME to remind us?

vext01: Also, do we ever reset the page permissions?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ltratt: Since we're on nightly, we can use the `checked_next_multiple_of` function instead of calculating alignment ourselves. [As Edd points out, it's surprisingly difficult to do this correctly!]

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ltratt: It's fine for us to *temporarily* mark pages as such: we just don't want to *leave* them in this state. 

However the particular way we're doing it here is a race condition. If thread1 `mprotect`s loc X to WRITE + EXEC *and* thread2 does the same, then thread1 can `mprotect` X back to EXEC before thread2 has had the chance to do its update. We need to use a lock *somewhere* to prevent this race condition.

Fortunately, because we're calling this from a parallel compile job in `mt`, we can do something pretty dumb at first: a global `MPROTECT` lock which we hold just before the first `mprotect` call and which we release immediately after the second `mprotect` call.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: Oh, you're right. Gonna look into `checked_next_multiple_of`.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: Doesn't this fixme already cover this? `FIXME: This might not work on other architectures or operating systems.`

ltratt: The race condition is not architecture dependent.

ptersilie: > However the particular way we're doing it here is a race condition.

Ah, I see it now too. Just a simple spin lock is fine, right?

ltratt: We need to hold e.g. a `Mutex`. Whether it's a spin lock or not shouldn't concern us.

ptersilie: Yeah, I meant a lock that waits until it gets access.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: [096e8be](https://github.com/ykjit/yk/pull/1591/commits/096e8be03a2fc213ff50e2e19c0a08d7497ecd7c)

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: I'm not sure if [3897c38](https://github.com/ykjit/yk/pull/1591/commits/3897c38c05bfe7aae1765d93ec4ade1aba7eddbe) is what you had in mind.

ltratt: More-or-less. I would do `Mutex::new(());` i.e. use the unit type rather than an integer. Conventionally in yk we also spell "lock" variables `lk`. Those are very minor, cosmetic changes, so we might as well do 'em, but they're a 2 second job.

One thing that I am chewing on right now is whether we should `panic` if either `mprotect` call fails. *Mostly* these can only fail if *we* do something stupid *but* if user code calls `mlockall` then `mprotect` would fail. Ideally we would throw a warning to `stderr` with `Verbosity::Error` or maybe `Verbosity::Warning` but that would mean we'd need a fallback mechanism for when patching fails. I suspect this commit removes the fallback mechanism, at which point the `panic` is OK until/if we ever see this happen in the wild.

vext01: > It's fine for us to temporarily mark pages as such

Well, it depends. There are configurations, even on linux, where a call to mprotect with W+X would fail.

Granted, that's off by default on linux, so we are unlikely to encounter it...

ltratt: This is a problem for JITs in general and we're not going to solve it in this PR, nor do I think we need worry about it.

ptersilie: > I would do Mutex::new(()); i.e. use the unit type rather than an integer.

I'm using the parking_lot mutex because we already import that. And it needs a u32. Want me to witch to `std::sync` in instead?

ptersilie: Addressed comments in [f5934a6](https://github.com/ykjit/yk/pull/1591/commits/f5934a6e8de19c96000a48aa20c1dbc4bd96519f) [872f241](https://github.com/ykjit/yk/pull/1591/commits/872f241bfed964aa0a005d4d8cdbe96a6c078351) [07f445d](https://github.com/ykjit/yk/pull/1591/commits/07f445d465ba3f6df81bdf44dc0fde4243388c31)

ptersilie:
Addressed all comments. But now need to look into an error in the awfy suite I'm suddenly getting. I wasn't getting this when I measured the performance. I can't remember if I rebased inbetween, though I might have. Investigating.

ptersilie:
Ready for re-review.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -2884,24 +2885,32 @@ impl CompiledTrace for X64CompiledTrace {
...
+        // to protect the entire page anyway.
+        let mprot_size = 8;
+
         // Make the trace writable while keeping it executable.
         // FIXME: This might not work on other architectures or operating systems. However, the
         // solution is very involved and requires changing `mt.rs` to make sure no other thread is
         // accessing the trace (which includes executing the trace itself or any of it's
         // side-traces, or side-tracig a guard failure).
-        let trace_addr = self.buf.ptr(AssemblyOffset(0)) as *mut libc::c_void;
+        //let trace_addr = self.buf.ptr(AssemblyOffset(0)) as *mut libc::c_void;
```
vext01, line 2905: commented code here

ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
```diff
@@ -69,6 +69,7 @@ pub(crate) extern "C" fn __yk_deopt(
     glen: usize,
 ) -> *const libc::c_void {
     let v = unsafe { slice::from_raw_parts(gptr, glen) };
+    // dbg!(v);
```
ltratt, line 72: Debugging.

ykrt/src/mt.rs
```diff
@@ -366,7 +366,13 @@ impl MT {
                 trace_iter.2,
             ) {
                 Ok(ct) => {
+                    let addr = ct.entry();
+                    // The order of the following two calls is important to avoid a race condition.
+                    // If we were to patch the parent trace first, there is a small window where
+                    // another thread takes the patched jump and deopts before we had a chance to
+                    // call `set_ctr` which sets information required by deopt.
```
ltratt, line 373: I think there is still a race here:

1. Thread A creates a side trace for guard G, calls `set_ctr` and is then interrupted.
2. Thread B also creates a side trace for G, calls `set_ctr` and `patch_guard`.
3. Thread A wakes up and calls `patch_guard.

Since `set_ctr` grabs a lock on G, why not use that to also call `patch_guard`? AFAICS we never call `set_ctr` without also calling `patch_guard`.

ptersilie: Is this what you had in mind? [03d3eb6](https://github.com/ykjit/yk/pull/1591/commits/03d3eb6250734d9a1b84a7f7af2e6f1b14690bbe)

ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
ptersilie: [01d5ff4](https://github.com/ykjit/yk/pull/1591/commits/01d5ff4cdd1841f34f745bed09b4d5fb8c1b149d)

ykrt/src/mt.rs
ltratt: Yes. We should update the doc string too, but otherwise I think this fixes the probable race.

ptersilie: Like so [7341847](https://github.com/ykjit/yk/pull/1591/commits/734184788a4609dbda62713fcada126471c0ad23)?

ltratt: That's very weird formatting which might cause `cargo doc` or `clippy` to bork (they're quite fussy). I'd suggest either making those part of the paragraph *or* using `*` bullet points. Please squash and make whichever update you see fit.

ptersilie: Neither cargo doc nor clippy complained. But as my lord wishes: [c05f68e](https://github.com/ykjit/yk/pull/1591/commits/c05f68e9d5d085cb02010584f96aed104e7aef2b) [0f74ef3](https://github.com/ykjit/yk/pull/1591/commits/0f74ef3b5fc092ba90e6bf24ec9351eda5eb29cb)

ltratt:
Please squash and rebase.

ptersilie:
Squashed and rebase. Note, there's a new commit which implements 64-bit jumps which we didn't support before. Be kind, it was hard work.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -139,10 +139,17 @@ pub(crate) type VarLocation = super::reg_alloc::VarLocation<Register>;
...
 static LK_PATCH: Mutex<()> = Mutex::new(());
-/// x86 JMP opcode.
+/// x86 JMP opcode for a 32bit offset.
 const JMP_OPCODE: u8 = 0xE9;
 /// x86 NOP opcode.
-const NOP_OPCODE: u8 = 0x80;
+const NOP_OPCODE: u8 = 0x90;
+/// x64 opcode to jump to an address stored in [RSP-8].
+const JMP_RSP_8: u32 = 0xFF6424F8;
+/// x64 opcode to move an imm32 into [rsp-8].
```
ltratt, line 148: I assume this is meant to be `imm64`?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -440,30 +447,33 @@ impl<'a> Assemble<'a> {
...
-                    // worth of nops (we could leave less nops if we were to overwrite the
+                    // After compiling a side-trace for this location, we want to patch in a jump
+                    // to its address here later, so we can jump to side-traces directly. Depending
+                    // on the distance of the side-trace to the parent trace we can either emit a
+                    // 32-bit jump (5 bytes) or need to patch a more involved 64-bit jump (22
+                    // bytes). We will leave enough space for either. The jump is required so that
+                    // we can patch the space following it with instructions to do the 64-bit jump
+                    // without having other threads pick up partial changes. Only in the last
+                    // atomic operation is the jump removed and we fall through to the 64-bit jump.
+                    // FIXME: we currently leave 36 bytes of patch space, though we likely need
```
ltratt, line 463: 36 seems a weird number? Shouldn't we leave at most `x - 1` (where `x` is a multiple of 8) bytes?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -440,30 +447,33 @@ impl<'a> Assemble<'a> {
...
                     // another time.
-                    ; nop
-                    ; nop
-                    ; nop
-                    ; nop
-                    ; nop
-                    ; nop
-                    ; nop
-                    ; nop
+                    ; jmp DWORD 27;            nop; nop; nop
```
ltratt, line 473: Does the `JMP` affect performance relative to just leaving it a `nop`?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -2916,25 +2928,45 @@ impl CompiledTrace for X64CompiledTrace {
                 libc::PROT_WRITE | libc::PROT_EXEC,
             )
         };
+        // Check that the target address is 8-byte aligned. This is required so we can patch
+        // this atomically.
+        let patch_addr = patch_addr as *mut u64;
+        debug_assert!(patch_addr.addr() % 8 == 0);
```
ltratt, line 2934: `assert`

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -2916,25 +2928,45 @@ impl CompiledTrace for X64CompiledTrace {
...
+                pcode.extend(&addr_in_bytes[0..4]);
+                // Emit `mov qword [rsp-4], ...`
+                pcode.extend(&MOV_RSP_4);
+                pcode.extend(&addr_in_bytes[4..8]);
+                // Emit `jmp [rsp-8]`
+                pcode.extend(&JMP_RSP_8.to_be_bytes());
+                // Write the jump code after the old jump that skips the NOPs.
+                let patch_addr_64 = unsafe { (patch_addr as *mut u8).offset(8) };
+                unsafe {
+                    std::ptr::copy_nonoverlapping(pcode.as_ptr(), patch_addr_64, pcode.len());
```
ltratt, line 2964: This is unsafe: at the very least it needs to be a sequence of atomic writes rather than a non-atomic copy.

However, that's not enough: it is possible that you write 8 bytes atomically, and then another thread executes this code. So you need to make sure that, after each atomic write, the "partially written" code can't do bad things. That will require a *very* careful comment justifying this.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -2916,25 +2928,45 @@ impl CompiledTrace for X64CompiledTrace {
...
+            let offset = i32::try_from(staddr as i64 - (patch_addr as i64 + JMP_32_SIZE));
+            if let Ok(offset) = offset {
+                // Emit a jump to a 32-bit offset.
+                let mut bytes: [u8; 8] = [NOP_OPCODE; 8]; // Fill with NOPs
+                bytes[0] = JMP_OPCODE;
+                bytes[1..5].copy_from_slice(&offset.to_ne_bytes());
+                let pval = u64::from_ne_bytes(bytes);
+                unsafe {
+                    // Now write the new instruction in a single atomic operation.
+                    *patch_addr = pval;
```
ltratt, line 2947: This isn't an atomic operation.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: No, we need to split the 64-bit address up into two 32-bit parts and write one to `rsp-8` and the other to `rsp-4`.

ltratt: OK can we document that these are two halfs of a 64 bit address and we have to write them in two parts?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: You are right and I'm bad at maths. It's supposed to be 32.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -139,10 +139,17 @@ pub(crate) type VarLocation = super::reg_alloc::VarLocation<Register>;
...
 const JMP_OPCODE: u8 = 0xE9;
 /// x86 NOP opcode.
-const NOP_OPCODE: u8 = 0x80;
+const NOP_OPCODE: u8 = 0x90;
+/// x64 opcode to jump to an address stored in [RSP-8].
+const JMP_RSP_8: u32 = 0xFF6424F8;
+/// x64 opcode to move an imm32 into [rsp-8].
+const MOV_RSP_8: [u8; 5] = [0x48, 0xc7, 0x44, 0x24, 0xf8];
+/// x64 opcode to move an imm32 into [rsp-4].
+const MOV_RSP_4: [u8; 5] = [0x48, 0xc7, 0x44, 0x24, 0xfc];
```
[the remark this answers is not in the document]
ltratt: You've implicitly explained why we have this in https://github.com/ykjit/yk/pull/1591#discussion_r1952267475. I just didn't get it when I wrote the comment.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: It might have a small performance impact. But unfortunately it's necessary. This ties in with your comment on the `copy_overlapping`. Essentially, this jump here prevents another thread from executing the patched code until we are done patching it. Because we are inserting our jump patch just below this jump, another thread executing this trace will jump over the patched code, while the patch is being written. Only when we have fully patched in the 64-bit jump we remove this jump here by overwriting it with NOPs (https://github.com/ykjit/yk/pull/1591/files/88498b3239afba436370c110a9557872731420b7..0d1df0671843c34300935bbf6fbf09d618665dcf#diff-b7513e8c31bfe56ee2b41fa041e8632d81f4bafc2e759b22b597f3867359269eR2969). This then allows any future executions of this trace to fall through into the 64-bit jump.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: See this comment: https://github.com/ykjit/yk/pull/1591#discussion_r1952285268

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: Will look into `AtomicPtr` as discussed offline.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -437,12 +437,15 @@ impl<'a> Assemble<'a> {
                     .get_ready_for_deopt(&mut self.asm, &self.deoptinfo[&deoptid].0);
                 // FIXME: Why are `deoptid`s 64 bit? We're not going to have that many guards!
 
-                // Word-align this location so we can patch this later in a single atomic
-                // operation.
+                // Align this location to the cache-line, which since a cache-line is a multiple of
+                // 8 bytes, will also word-align it too. This allows us to safely patch this place
+                //   later in a single atomic operation.
                 let off = self.asm.offset().0;
+                let cachesize =
```
ltratt, line 444: I would prefer to call this `clsize` or something that doesn't suggest it's the *cache's* size. [I know, I know, bikeshedding, but this is easily misunderstood.]

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -319,6 +319,7 @@ struct Assemble<'a> {
     asm: dynasmrt::x64::Assembler,
     /// Deopt info, with one entry per guard, in the order that the guards appear in the trace.
     deoptinfo: HashMap<usize, (GuardSnapshot, DeoptInfo)>,
+    free_guard_reg: HashMap<usize, Rq>,
```
ltratt, line 322: This will definitely need documenting: I'm not really sure what it's doing.

Also, do we need to think about FP registers here?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -2927,10 +2927,11 @@ impl CompiledTrace for X64CompiledTrace {
         // dynasmrt's is hard-coded to 4096!
         let mprot_addr = patch_addr - (patch_addr % page_size::get());
         // We need to mark everything from the beginning of the memory page up to the `patch_addr`
-        // plus the actual bytes we want to patch. The number of bytes needs to match the patch
-        // space (jump + NOPs) we left after `fail_label`. `mprotect` will work out how many pages
-        // it needs to change.
-        let mprot_size = (patch_addr - mprot_addr) + 36; // XXX this is now smaller
+        // plus the actual bytes we want to patch.
+        // FIXME: If we know that the page-size is a multiple of 8 and the patch address is 8-byte
```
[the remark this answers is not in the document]
ptersilie: Because `mprotect` can only mark entire pages, we want to mark the minimal amount of pages we have to in order to make the bit we want to patch writeable. I think if we 8-byte align the address we want to patch, then by definition it will never cross the boundary of a page, so we don't need to do any maths here.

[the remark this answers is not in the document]
ltratt: Yes, we can safely assume that pages are a multiple of 8.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: Yeah, I was rushing to make these changes, so used whatever name first came to mind. I'm not sold on this name either.

No, we don't need to worry about FP registers, since we only need one GP register to store the 64-bit jump address.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: Oh, I've forgotten about this too.

ptersilie: [5ffe5fb](https://github.com/ykjit/yk/pull/1591/commits/5ffe5fbf8bae53e974f65f16d96cdfcec7198231)

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: [9046424](https://github.com/ykjit/yk/pull/1591/commits/9046424d13ed52469418f28e9ce8b3d2d764c04d)

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: See [d0172f0](https://github.com/ykjit/yk/pull/1591/commits/d0172f0ceafc3e70129eb0a7b7c8060fefb4116e).

To add more context: this is the register that the register allocator frees up for us before we jump to the `fail_label`. We use this register to store the address of the side-trace in order to do 64-bit jumps.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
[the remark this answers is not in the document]
ptersilie: Simplified here: [c334277](https://github.com/ykjit/yk/pull/1591/commits/c334277db9af49252c35c2b26629b1c78035e290)

ptersilie:
Pushed some commits that address comments, and some new ones addressing FIXMEs or updating comments/docs.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: [10c8975](https://github.com/ykjit/yk/pull/1591/commits/10c8975f836dff4c27290657bd86ce63eebd79fa)

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: This is no longer how we do things.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -3017,12 +3017,12 @@ impl CompiledTrace for X64CompiledTrace {
...
         // this atomically.
-        let patch_addr = patch_addr as *mut u64;
-        debug_assert!(patch_addr.addr() % 8 == 0);
+        debug_assert!(patch_addr % 8 == 0);
+        let patch_addr = std::sync::atomic::AtomicPtr::new(patch_addr as *mut u64);
         if res == 0 {
             // Patch the new 64-bit address into the mov instruction.
-            let addr_in_bytes = staddr as u64;
-            unsafe { *patch_addr = addr_in_bytes };
+            let val = &mut (staddr as u64);
```
ltratt, line 3008: Does `&mut` actually work here? I would have expected `unsafe` to be needed.

I'm also not entirely sure this does what we think it should. I *think* (but I might be wrong) that what is this doing is the equivalent of:

```c
uint_ptr_t *patch_addr = ...;
patch_addr = val;
```

i.e. I *think* you're expecting this to be `*patch_addr = val` but I don't think it's doing that additional level of indirection.

ptersilie: The `AtomicPtr::store` instruction expects a pointer to the new value. That's why I had to turn `staddr` into a pointer. I wouldn't expect the code to run if this doesn't write the `staddr` into the `patch_addr` properly.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: This is also gone now.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ltratt: Don't flatter my ego :p

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -425,26 +417,64 @@ impl<'a> Assemble<'a> {
...
                 self.ra
                     .get_ready_for_deopt(&mut self.asm, &self.deoptinfo[&deoptid].0);
                 // FIXME: Why are `deoptid`s 64 bit? We're not going to have that many guards!
+
+                // Align this location to the cache-line, which since a cache-line is a multiple of
+                // 8 bytes, will also word-align it too. This allows us to safely patch this place
+                // later in a single atomic operation.
+                let clsize =
+                    cache_size::cache_line_size(1, cache_size::CacheType::Instruction).unwrap();
+                // Unwrap can only fail if there's an overflow.
```
ltratt, line 436: I don't think there's an `unwrap` that this refers to?

Oh, hang on, is it referring to the line *above`? But I don't see overflow happening anywhere?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -425,26 +417,64 @@ impl<'a> Assemble<'a> {
...
+                    // FIXME: If the side-trace offset to its parent-trace is < 32-bit we can emit
+                    // a relative jump here that doesn't require a register and is likely faster.
+                    // FIXME: Ideally, instead of patching this place, we could patch the guards
+                    // directly which gets rid of an extra jump. But since `cg_icmp_guard` makes
+                    // use of various types of jumps (e.g. jne, jl, etc), this is an optimisation
+                    // for another time.
+                    //
+                    // Since the mov instruction's non-operand part is 2 bytes, push 6 bytes here
+                    // so the address operand is 8-byte aligned.
+                    ; nop; nop; nop; nop; nop; nop
```
ltratt, line 462: I might have misunderstood this, but I *think* this code is assuming that at the point of the `dynasm` call we are already aligned to 8 bytes? If so, is that correct? I don't think any given `dynasm` has to end on an 8 byte alignment. Don't we need to add a variable number of `noop`s here?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -425,26 +417,64 @@ impl<'a> Assemble<'a> {
...
+                    ; mov Rq(jumpreg.code()), QWORD 0x0
```
ltratt, line 463: Maybe I'm a dummy but I *though* we were going to do something like:

```
MOV Rq(jumpreg.code()), QWORD deopt
JMP Rq(jumpreg.code())
```

without the `cmp/je` dance?

I thought that the problem you had with this is that global labels don't work with `MOV` or similar in dynasm? [Did we confirm this?] If so, we could patch this address before codegen finishes. [It would be nicer to use labels, but maybe that doesn't work.]

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: Think that's an old comment.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: You are correct. At this point we assume we are 8 byte aligned since we are aligning to the cache line size with `self.asm.align(clsize, NOP_OPCODE);`. This assumes that a cache line size is a multiple of 8 bytes.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: Yeah, we can't use labels as they emit a different mov instruction that we can't patch later with an absolute address. So I will do the patching.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: We also have an assert here to make sure we never try to patch an unaligned location: https://github.com/ykjit/yk/pull/1591/files#diff-b7513e8c31bfe56ee2b41fa041e8632d81f4bafc2e759b22b597f3867359269eR2993

ltratt: Why don't we just calculate the minimal number of `nop`s necessary -- it's a simple calculation. Right now, we can insert up to 7 unnecessary `nop`s per guard!

ptersilie: Maybe I'm missing something obvious here, but I don't see the simple calculation you are referring to (especially after you previously said how surprisingly difficult alignment can be).

We need to ensure two things here:
1. The entire mov instruction (10 bytes) needs to fit within a cache line
2. The operand of the mov instruction needs to be aligned to 8 bytes.

For simplicity let's assume a cache line is 64 bytes. If the current address before emitting the mov operation is 0, then the minimum amount of `nop`s we need to insert is 6 to align it to 6 bytes. If the current address is 12, we need to insert 2 bytes to align to 14 bytes. And so on. Other addresses we can align to are 22, 30, 38, 46, 54. Now if the current address is 55, the mov instruction won't fit into the cache line, so we first need to align to 64+6=70.

ltratt: I don't see anything here that isn't basic arithmetic?

ptersilie: Lol, yes, I didn't say it was impossible, just that the working out the formula isn't what I would call simple. Unless, you meant simple for a compiler to calculate.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ltratt: Maybe I'm wrong but isn't it something like:

```rust
let cur_addr = asm.len_bytes();
let mut align = (cur_addr+2).next_multiple_of(8);
if align.next_multiple_of(page_size) == align { align += 8 }
let num_nops = align - cur_addr;
```

?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ltratt: Maybe I'm wrong but isn't it something like:

```rust
let cur_addr = asm.len_bytes();
let mut align = (cur_addr+2).next_multiple_of(8);
if align.next_multiple_of(page_size) == align { align += 8 }
let num_nops = align - cur_addr;
```

?

ptersilie: Huh, that seems to do the trick with one small change. It needs to be `let num_nops = align - cur_addr - 2;`. I don't know why, but I just couldn't work this out so quickly.

ltratt: Oops, you're totally right about the `-2`!

ptersilie: [199f341](https://github.com/ykjit/yk/pull/1591/commits/199f341e1104c19f4d3ef1907f67ee6e28e94668)

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: [10b0e34](https://github.com/ykjit/yk/pull/1591/commits/10b0e34a22b294dac02b12ec6959bb8005e195ec)

ltratt:
Please squash.

ptersilie:
Merged in master and fixed some tests.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -4721,11 +4727,14 @@ mod tests {
                 cmp r.32.x, 0x03
                 setz r.8._
                 ; guard true, %1, [] ; ...
-                bt r.32._, 0x00
+                and r.32.y, 0x01
+                mov [rbp-0x01], r.8.y
```
ptersilie, line 4731: The reason for this extra `mov` here is that we are now spilling to free up a register needed for the new deopt jump. I'm not super sure about the extra `and`. Is this because the value we are spilling is an `i8` and we need to make sure we are not writing any additional bytes?

ltratt: The `and` is because it's an `i1` with `Undefined` upper bits: it must be zero extended before spilling to the stack.

ltratt:
Please squash.

ptersilie:
Squashed.

ptersilie:
> 19:03:42 error[rejected]: failed to satisfy license requirements

That's one I haven't seen before.

ptersilie:
Fixed the license problem. Did a tryci run too to make sure. All good. Ok to squash?

ltratt:
Please squash.

ptersilie:
Squashed.

PR #1592, opened 2025-02-07
Abort tracing as soon as we notice we're in a different call frame

ltratt:
This PR first refactors a lot of code, which is mostly boring, but without which we couldn't get to the meat, which is https://github.com/ykjit/yk/commit/bbcb1c4fcdff262788c2912c6523ac80a1874c55.

That commit aborts tracing as soon as the control point notices we're in a different frame. Previously we detected "we're in a different frame" when we wanted to stop tracing. This commit changes this so that the control point checks this at all points a thread is tracing. This means that we can spot problematic traces as they're ongoing, and abort them early. Entertainingly, this allows us to simplify some surprisingly hairy code.

@ptersilie I would very much welcome your thoughts on the changes to the `early_return2` test in https://github.com/ykjit/yk/commit/bbcb1c4fcdff262788c2912c6523ac80a1874c55: does this do what you would expect?


ykrt/src/mt.rs
```diff
@@ -573,6 +572,27 @@ impl MT {
...
                 Some(hl) => {
+                    if is_tracing {
+                        if let MTThreadState::Tracing {
+                            frameaddr: tracing_frameaddr,
+                            hl: tracing_hl,
+                            ..
+                        } = mtt.peek_tstate()
+                        {
+                            if frameaddr != *tracing_frameaddr {
+                                // ...but no longer in the frame we started in, so we need
```
ptersilie, line 583: The `...` suggests this is a continuation of a previous comment, but I can't seem to find that previous comment.

ltratt: Oops! Fixed in f9dbd81aacecdce6052d5ccb73a645cbe804da0f.

ltratt:
I think this is ready for review.

ltratt:
Does dcd08c29 match what you were expecting?

ptersilie:
Yes, that's what I had in mind. Now it's clear why the `trace-aborted` happened at that point. Please squash.

ltratt:
Squashed.

PR #1593, opened 2025-02-07
Don't overwrite registers that we may still need to move.

ltratt:
Fixes https://github.com/ykjit/yk/issues/1571.


ltratt:
Havlak as per normal. Retry?

PR #1596, opened 2025-02-10
Temporarily disable Havlak in CI.

ltratt:
We know this test fails, and rather than retry PRs, we might as well acknowledge that testing this in CI is currently not the right approach. As soon as we've fixed the underlying bug, we can remove the `sed` line in this commit and easily reenable Havlak.


ltratt:
A test has (randomly and not connected to this PR) failed, but I can't work out from the BB log which! Either way, we should probably just retry this PR.

vext01:
It was a disassembly error in funcptrarg_noir.c. We've seen it before, but rarely.

ltratt:
I'm an idiot: I copy and pasted the file name `rebench.conf` in without thinking. The force push fixes that.

PR #1599, opened 2025-02-10
More floating point codegen

ltratt:
We previously didn't handle `select` with floating point values -- which I've now seen in the wild! Please check the codegen properly for this: it made my head hurt.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -2639,6 +2639,16 @@ impl<'a> Assemble<'a> {
     fn cg_select(&mut self, iidx: jit_ir::InstIdx, inst: &jit_ir::SelectInst) {
         // First load the true case. We then immediately follow this up with a conditional move,
         // overwriting the value with the false case, if the condition was false.
+        match self.m.type_(inst.trueval(self.m).tyidx(self.m)) {
+            Ty::Void => todo!(),
+            Ty::Integer(_) | Ty::Ptr => self.cg_select_int_ptr(iidx, inst),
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -2675,6 +2685,61 @@ impl<'a> Assemble<'a> {
...
+        );
+
+        assert_eq!(
+            inst.trueval(self.m).bitw(self.m),
+            inst.falseval(self.m).bitw(self.m)
+        );
+        match inst.trueval(self.m).bitw(self.m) {
+            32 => {
+                dynasm!(self.asm
+                    ;   bt Rd(cond_reg.code()), 0
```
vext01, line 2719: Would the `fcmov` instruction help here? That would bring this in parity with the integer version.

ltratt: You tell me. I ended up baffled by this; the documentation on this stuff is much sparser -- and frankly, much worse -- than for general purpose registers. Personally I'd be happy to get something we're confident is *correct* in and optimise it later.

vext01: Now I look at it, `fcmov` uses the old fashioned floating point stack and not the xmm registers, so we can't (and shouldn't) use it.

Your codegen looks correct to me. AI tells me there is no xmm equivalent to `fcmov` and gave me a technique similar to what you have done here.

ltratt: Sounds like we can merge this then?

PR #1603, opened 2025-02-13
Stop tracing when we start unrolling.

ltratt:
Though it may not be obvious that this is what it's doing, this commit stops us tracing undesirable things, notably when we start unrolling inner loops.

Consider this example program:

```python
for x in range(...):
  if x > ...:
    for y in range(...):
```

The outer loop can get hot before the inner loop, so we start tracing the outer loop -- which is fine. But when we trace, we may go around the inner loop an unbounded number of times, endlessly unrolling the inner loop. This can lead to gigantic, and "unstable" traces -- because we've unrolled a specific number of iterations, there's a very high chance that we'll see a different number of iterations next time and encounter a guard on the first "iteration" of the gigantic-loop. This then leads to us creating large numbers of essentially identical side-traces, hugely bloating the system. Using havlak as an example this (combined with a couple of other things) leads to ~10x more traces being created than we would expect.

To avoid this happening, we have to track what `Location`s we've seen while tracing. There are some potentially cunning things we could do here if/when performance becomes an issue, but for now we force every `Location` we see when tracing to become a `HotTracing`. That gives us a stable ID (the `HotLocation`'s `Mutex`) which we can use as a reliable proxy for "have we seen this `Location` before?"

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
@@ -267,9 +267,18 @@ pub(crate) extern "C" fn __yk_deopt(
                                 unsafe { rbp.offset(isize::from(*extra)) }
                             };
                             match size {
-                                8 => unsafe { ptr::write::<u64>(temp as *mut u64, jitval) },
-                                4 => unsafe { ptr::write::<u32>(temp as *mut u32, jitval as u32) },
+                                1 => unsafe { ptr::write::<u16>(temp as *mut u16, jitval as u16) },
                                 2 => unsafe { ptr::write::<u16>(temp as *mut u16, jitval as u16) },
+                                4 => unsafe { ptr::write::<u32>(temp as *mut u32, jitval as u32) },
+                                8 => unsafe { ptr::write::<u64>(temp as *mut u64, jitval) },
+                                16 => {
+                                    // FIXME: This case is clearly not safe in general: it just so
+                                    // happens to work because it the moment the biggest value we
+                                    // handle is 64 bits (be that a pointer, u64, or double). If
+                                    // and when we support values bigger than 64 bits, this line
+                                    // will lead to weird problems.
+                                    unsafe { ptr::write::<u64>(temp as *mut u64, jitval) };
+                                }
                                 _ => todo!("{}", size),
                             }
                         }
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -4,7 +4,7 @@ use std::{
     assert_matches::{assert_matches, debug_assert_matches},
     cell::RefCell,
     cmp,
-    collections::VecDeque,
+    collections::{HashSet, VecDeque},
     env,
     error::Error,
     ffi::c_void,
@@ -454,6 +454,7 @@ impl MT {
                             promotions: Vec::new(),
                             debug_strs: Vec::new(),
                             frameaddr,
+                            seen_hls: HashSet::new(),
                         });
                     }),
                     Err(e) => {
@@ -491,6 +492,7 @@ impl MT {
                             promotions,
                             debug_strs,
                             frameaddr: tracing_frameaddr,
+                            seen_hls: _,
                         } => {
                             // If this assert fails then the code in `transition_control_point`,
                             // which rejects traces that end in another frame, didn't work.
@@ -531,6 +533,7 @@ impl MT {
                             promotions,
                             debug_strs,
                             frameaddr: tracing_frameaddr,
+                            seen_hls: _,
                         } => {
                             assert_eq!(frameaddr, tracing_frameaddr);
                             (hl, thread_tracer, promotions, debug_strs)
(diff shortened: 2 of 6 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
```diff
@@ -267,9 +267,18 @@ pub(crate) extern "C" fn __yk_deopt(
...
                             };
                             match size {
-                                8 => unsafe { ptr::write::<u64>(temp as *mut u64, jitval) },
-                                4 => unsafe { ptr::write::<u32>(temp as *mut u32, jitval as u32) },
+                                1 => unsafe { ptr::write::<u16>(temp as *mut u16, jitval as u16) },
                                 2 => unsafe { ptr::write::<u16>(temp as *mut u16, jitval as u16) },
+                                4 => unsafe { ptr::write::<u32>(temp as *mut u32, jitval as u32) },
+                                8 => unsafe { ptr::write::<u64>(temp as *mut u64, jitval) },
+                                16 => {
+                                    // FIXME: This case is clearly not safe in general: it just so
```
vext01, line 275: I know this isn't really what we are doing here in this PR, but a casual reader will think "if the largest object we can handle is 64-bits, then why is this arm reachable". Did we get to the bottom of why this happens?

ltratt: We didn't -- but, regrettably, the different paths we now trace hit this case. It sucks, and we need work out why we're being told to sometimes use a 16 byte register.

Suggestion: we let this go for now (as much as I dislike it) and I raise an issue as soon as the PR is merged to remind us to work out what's going on here.

vext01: I tend to agree. Perhaps raise an issue?

vext01:
Trying to understand the high-level goal here. So you keep track of which locations you've already seen while tracing and if you see a one a second time, but which isn't the location that started tracing, we have unrolled an inner loop, and we abort? Is that the gist of this?

ltratt:
Correct!

ptersilie:
Do we not outline functions with loops in it already? Does this mean we observe this only on the main interpreter loop?

ltratt:
"Don't outline functions with loops" is an LLVM-level thing (i.e. "don't outline C functions with loops in"). This PR is working at the upper-language (e.g. Lua) level.

ptersilie:
Oh right. Got you.

ykrt/src/mt.rs
```diff
@@ -741,7 +753,36 @@ impl MT {
...
+                        if let Some(hl_ptr) = hl_ptr {
+                            let MTThreadState::Tracing {
+                                frameaddr: tracing_frameaddr,
+                                seen_hls,
+                                ..
+                            } = mtt.peek_mut_tstate()
+                            else {
+                                panic!()
+                            };
+                            if frameaddr != *tracing_frameaddr {
```
vext01, line 776: Is this bit necessary, or something unrelated we are sneaking in?

ykrt/src/mt.rs
```diff
@@ -741,7 +753,36 @@ impl MT {
...
+                                // We're tracing but no longer in the frame we started in, so we
+                                // need to stop tracing and report the original [HotLocation] as
+                                // having failed to trace properly.
+                                return TransitionControlPoint::AbortTracing;
+                            }
+                            if !seen_hls.insert(hl_ptr) {
```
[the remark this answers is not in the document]
ltratt: It's due to the two representations of `Counting`: one without a `HotLocation`, one with.

ykrt/src/mt.rs
ltratt: [i.e. it should always have been present]

vext01: Do you want to keep it?

ltratt: It should always have been in, and it directly affects this code, so yes.

ykrt/src/mt.rs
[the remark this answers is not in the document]
vext01: OK, I've forgotten how that part of the system works, so I'll have to trust you on this one!

ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
ptersilie: What's extra weird is that the jit value reports a size of 64-bit. I wonder how that mismatch happened.

ltratt: I think it's something we should look at pretty pronto: it only crops up in a small number of traces, but it does indeed crop up.

vext01:
This is at the edge of my understanding of mt.rs, but seems ok to me.

Perhaps add some more info to the commit message+PR description about why we are tracking what locations we've seen, as that wasn't immediately obvious to me (it allows us to detect back edges that aren't for the loop we are tracing).

ykrt/src/mt.rs
vext01: sorry, we raced. All ok.

vext01:
I leave it up to you.

ltratt:
I think it's good to go myself. It could always be improved, I suppose, but I feel it accurately describes the problem and the fix. And this is the first of (probably) a number of fixes we'll be making in this regard, so now we've all got some context as to the general problem, I think future PRs will be easier to grok.

PR #1606, opened 2025-02-14
Introduce "null" `YkLocation`s.

ltratt:
Previously we allowed users to pass `NULL` to `yk_mt_control_point` to signify "there's nothing to trace at this point". However that meant that most interpreters would need to maintain two structures for locations: one to say "is there anything to trace at this point?" and another for "this is the YkLocation for this point, if one is needed". This was messy, and easy to misunderstand.

This commit no longer allows passing `NULL` to `yk_mt_control_point`, and instead introduces "null"-equivalent `YkLocation`s which are created by `Location::null()` or `yk_location_null()`. In essence, these tell `yk_mt_control_point` that there's nothing to do at this point in the high-level program, allowing us to compress the two structures we needed previously into one.

Is this a perfect API? Probably not. We might be able to do better again, but this is a useful improvement already, and has allowed me to meaningfully improve yklua.

Note: because this is a breaking change, I have temporarily had to point `.buildbot.sh` at my yklua fork. As soon as this merges, we can update yklua, and then have `.buildbot.sh` point back to the main yklua repository. [The yklua PR is at https://github.com/ykjit/yklua/pull/101].

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -186,11 +186,16 @@ impl VarLocation {
             },
             yksmp::Location::Constant(v) => {
                 let bitw = m.inst(iidx).def_bitw(m);
-                assert!(bitw <= 64);
+                assert!(bitw <= 32);
                 VarLocation::ConstInt {
                     bits: bitw,
                     v: u64::from(*v),
                 }
+            }
+            yksmp::Location::LargeConstant(v) => {
+                let bitw = m.inst(iidx).def_bitw(m);
+                assert!(bitw <= 64);
+                VarLocation::ConstInt { bits: bitw, v: *v }
             }
             e => {
                 todo!("{:?}", e);
@@ -259,6 +264,8 @@ impl From<&VarLocation> for yksmp::Location {
             VarLocation::ConstInt { bits, v } => {
                 if *bits <= 32 {
                     yksmp::Location::Constant(u32::try_from(*v).unwrap())
+                } else if *bits <= 64 {
+                    yksmp::Location::LargeConstant(*v)
                 } else {
                     todo!(">32 bit constant")
                 }
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -2,8 +2,11 @@
 
 set -eu
 
-# What git commit hash of yklua will we test in buildbot?
-YKLUA_COMMIT="7ced271cf4f36a78127d0cc745906cc21409ae0d"
+# What git commit hash of yklua & ykcbf will we test in buildbot?
+YKLUA_REPO="https://github.com/ltratt/yklua.git"
+YKLUA_COMMIT="305bef80816b84e0848a6694c87e715c4cd263f5"
+YKCBF_REPO="https://github.com/ltratt/ykcbf.git"
+YKCBF_COMMIT="a7eb0e0dbeef38191c4fe823c5aed50e5768df5c"
 
 TRACERS="hwt swt"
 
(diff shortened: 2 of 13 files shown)


vext01:
> interpreters would need to maintain two structures for locations: one to say "is there anything to trace at this point?" and another for "this is the YkLocation for this point, if one is needed"

> This commit no longer allows passing NULL to yk_mt_control_point, and instead introduces "null"-equivalent YkLocations which are created by Location::null() or yk_location_null(). In essence, these tell yk_mt_control_point that there's nothing to do at this point in the high-level program, allowing us to compress the two structures we needed previously into one.

I don't understand the semantic difference between the new and old design.

The old design already captured "nothing to do" and "here's the location if there is something to do" in one structure (a pointer).

Looking at the diff, it looks like we are trying to avoid conditionally allocating and conditionally fetching, the locations(?) If that's the goal, then why do we need `Location:null()`? The representation of the "null" location seems separate?

vext01:
After discussing offline I think I now understand.

The critical detail is, the optimal starting point for a loop isn't necessarily determined by the kind of opcode.

For example in [this PR](https://github.com/ykjit/yklua/pull/101) Laurie figured out that the best place to start a trace is not at the FORLOOP opcode (which is the end of a loop), but at the first opcode in the loop body (which could be any bytecode).

This change allows interpreters to break the assumption that the start of a loop is characterised by a specific kind of bytecode, by pushing the "null" location down one level, into the location itself.

vext01:
Two questions I had:

 - Are we worried with the memory consumption of storing an `YkLocation` for every bytecode in the program? (Before we only stored one for a "potential trace start bytecode"). Rather than ask you to answer that I did a quick calculation. We'd add about 14KiB memory consumption of a program like deltablue. Which I think is OK.

 - Is the name "null location" going to falsely imply that the interpreter author can pass a "null pointer" instead?

ltratt:
>    Are we worried with the memory consumption of storing an YkLocation for every bytecode in the program? 

That is not, and must not be, yk's concern. Here's a tool: you decide how to use it, for good or ill.

> Is the name "null location" going to falsely imply that the interpreter author can pass a "null pointer" instead?

I don't think so.

vext01:
> That is not, and must not be, yk's concern. Here's a tool: you decide how to use it, for good or ill.

The design definitely encourages storing one location per-bytecode. You could have just one null location that you pass for each bytecode that you don't want to trace from, but that would re-introduce the conditional checks that this PR aims to remove.

Regardless, the memory consumption is small, so I'm going to go ahead.

ltratt:
Just to be clear: you don't have to use `yk_location_null`. For example, I investigated "is there a spare bit in Lua opcodes that would allow yklua to tag opcodes" but there is no such spare bit. If there was, I just wouldn't use `yk_location_null` there.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -186,12 +186,17 @@ impl VarLocation {
...
             yksmp::Location::Constant(v) => {
                 let bitw = m.inst(iidx).def_bitw(m);
-                assert!(bitw <= 64);
+                assert!(bitw <= 32);
                 VarLocation::ConstInt {
                     bits: bitw,
                     v: u64::from(*v),
                 }
             }
+            yksmp::Location::LargeConstant(v) => {
```
ptersilie, line 195: Are these changes relevant to this PR?

ltratt: Yes and no: the relevant changes to yklua trigger new traces that encounter these cases.

ptersilie: Ah, that's fair!

ltratt:
This one hits a horrible performance problem that I have a separate fix for.

ltratt:
We should be able to retry this if/when https://github.com/ykjit/yk/pull/1610 merges.

PR #1608, opened 2025-02-15
Codegen more `llvm.call` items

ltratt:
These are calls that I've seen in traces (and that currently prevent those traces being compiled).

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -1629,6 +1629,11 @@ impl<'a> Assemble<'a> {
                 self.cg_ctpop(iidx, op);
                 Ok(())
             }
+            x if x.starts_with("llvm.floor") => {
+                let [op] = args.try_into().unwrap();
+                self.cg_floor(iidx, op);
+                Ok(())
+            }
             x if x.starts_with("llvm.smax") => {
                 let [lhs_op, rhs_op] = args.try_into().unwrap();
                 self.cg_smax(iidx, lhs_op, rhs_op);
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -3189,6 +3194,12 @@ mod tests {
         let fmm = fmatcher("fp.128.x fp.128.y fp.128.x");
         assert!(fmm.matches("xmm0 xmm1 xmm0").is_ok());
         assert!(fmm.matches("xmm0 xmm0 xmm0").is_err());
+
+        let fmm = fmatcher("fp.128.x fp.128.y fp.128._");
```
vext01, line 3267: Does this mean we've seen f128 in the wild?

ltratt: The `xmm` registers are 128 bit, so we classify them as such, even if we only use them for doubles/floats.

vext01: Ah, I see, it's just the register names. OK, cool.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -2163,17 +2191,18 @@ impl<'a> Assemble<'a> {
             jit_ir::FloatPredicate::OrderedGreater | jit_ir::FloatPredicate::UnorderedGreater => {
                 dynasm!(self.asm; seta Rb(tgt_reg.code()))
             }
-            jit_ir::FloatPredicate::OrderedGreaterEqual => {
+            jit_ir::FloatPredicate::OrderedGreaterEqual
+            | jit_ir::FloatPredicate::UnorderedGreaterEqual => {
```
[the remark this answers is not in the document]
ltratt: I am out of my depth on these things. I know that you understand these better so I'm hoping for pointers from you!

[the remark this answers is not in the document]
vext01: I'm not sure our fcmp routine is right, but it's not your fault.

If this change is working in this instance, I suggest we should merge this while I try to make sense of this separately.

PR #1610, opened 2025-02-17
Properly reset the `HotLocation` when tracing outside the original frame

ltratt:
When we fall (into or out of) the original frame for the interpreter loop, we need to decide what to do with the `HotLocation` that started the trace off. Previously we could leave it in "perma-tracing" state that then got caught by a branch that also deals with side tracing. That works, but is incredibly inefficient, because it kept retrying to trace from scratch. That could mean we tried (unsuccessfully!) recording hundreds of thousands of traces!

This PR changes that: when we detect the "fall out of frame" case, we deal with it more sensibly. In particular, if we were side-tracing, we restore the original `CompiledTrace`, moving us back to the (at least somewhat) happy path.

### DIFF
--- a/tests/c/early_return2.c
+++ b/tests/c/early_return2.c
@@ -21,35 +21,43 @@
 //     yk-jit-event: tracing-aborted
 //     b3
 //     8
-//     b3
-//     7
-//     yk-jit-event: start-tracing
-//     b3
-//     6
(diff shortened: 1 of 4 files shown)


vext01:
Looks decent to me. @ptersilie ?

PR #1612, opened 2025-02-18
Optimise arithmetic shift right.

ltratt:
Doing this means that arbbitint.rs needs to be more clear about its naming scheme: ash vs. lshr is rather important!

[Until this commit, I had not realised that `shr` in Rust is lshr for unsigned integers and ashr for signed integers!]


vext01:
LGTM.

PR #1614, opened 2025-02-19
Prefer `Temporary` over `clobber_reg`.

ltratt:
`Temporary` will often have a spare register to hand out, and not need to spill anything; `clobber_reg` will more often have to spill.

Now, we could do better here, because the register we're asking for only needs to be unused *after* the inputs are processed. But, still, this is better than nothing.


ltratt:
It turns out that this PR tickled an existing two-line bug in codegen for guards which we'd got lucky with before. I've fixed that in the force push (the first force push accidentally rebased; the second force push fixes that, but you'll need to fix the base commit if you want to see that).

ltratt:
@vext01 We can readd this to the queue.

ltratt:
Let's hold off on this one for a bit. Something is going on that I don't yet fully understand.

ltratt:
@vext01 I believe #1629 has fixed the bug that this PR accidentally happened to bump into. I believe we can retry as-is.

PR #1615, opened 2025-02-20
Better jump var writing

ltratt:
This PR optimises the writing of jump vars, in essence getting rid of the (repeated) `push rax; move a spill; pop rax` dance we had before. This is particularly useful for long traces, where we could do a *lot* of this at the end.


vext01:
Based on superficial knowledge of this part of the system, LGTM.

PR #1622, opened 2025-02-21
Reduce register pressure for float comparisons.

vext01:
Before we'd always request a temporary register, even for comparisons that didn't need one. This change fixes this.


ltratt:
By definition this is a functional change, right?

vext01:
Hah, I was thinking about this just after I raised it. It doesn't affect any semantics of interpreters using yk, so no "functional change" on that level, but yes, a functional change one level down. I'll force push to fix the message, because it's confusing me too now!

vext01:
pushed a fix.

PR #1625, opened 2025-02-25
Function flags

vext01:
(First commit is a new rust nightly thing)

Requires https://github.com/ykjit/ykllvm/pull/230 (and thus a ykllvm sync).


vext01:
I'm surprised that merged without the ykllvm sync.

PR #1627, opened 2025-02-26
`AND` with a constant implicitly zero extends.

ltratt:
By definition, if we `AND` a register with a constant, the result must be zero extended (all bits to the left of the constant will be `AND`ed away).


vext01:
If it continues to be an issue, we can disable List momentarily.

PR #1628, opened 2025-02-26
Always sort by deopt ID.

ltratt:
Diffing traces is massively easier if the deopt routines are sorted: the tiny performance hit we take in release mode is well worth it.


vext01:
I agree!

PR #1631, opened 2025-02-27
Continue splitting the register allocator up into more testable chunks.

ltratt:
This commit splits the register allocator into two major parts:

  1. A non-mutating part which decides what registers to allocate.
  2. A mutating part which generates code.

Previously these were intermingled together in a way that was hard to understand and impossible to test. Happily the second part is now tiny: all of the hard stuff is in the first part.

This commit also takes a section of that first part -- the bit about which I was least confident -- and makes it entirely stateless in `reg_copies_to_actions` (i.e. it doesn't even read state), making it much more easily testable in a unit test setting.

In essence (and the comments in the code explain this, I hope, in more detail) the register allocator now works in terms of "actions": that is, we gradually build up a sequence of things we will do in the future (i.e. code we will generate), but we don't do them yet.

Happily this commit also makes the allocator simpler overall, despite generating code that is at least as good as we did before.

There is more we need to do here, but this is a meaningful step in the right direction.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
@@ -397,16 +397,18 @@ impl LSRegAlloc<'_> {
         &mut self,
         asm: &mut Assembler,
         iidx: InstIdx,
-        mut constraints: [GPConstraint; N],
+        cnstrs: [GPConstraint; N],
     ) -> [Rq; N] {
-        // Register assignment is split into two stages:
-        //   1. Find a register for each `GPConstraint`. During this phase, no changes to state in
-        //      `self` must be made.
-        //   2. Move values out of, and into, registers and update `self`.
+        // Register assignment is split into three stages:
+        //   1. Find a register for each `GPConstraint`. No state changes to `self` are made.
+        //   2. Decide how we shuffle the existing registers around. No state changes to `self` are
+        //      made.
+        //   3. Generate code to shuffle the existing registers around, and update the output
+        //      state. State changes to `self` are made during this stage.
 
         // No constraint operands should be float-typed.
         assert_eq!(
-            constraints
+            cnstrs
                 .iter()
                 .filter(|x| match x {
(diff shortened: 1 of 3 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
```diff
@@ -435,60 +437,72 @@ impl LSRegAlloc<'_> {
...
+                }
+                RegAction::Spill => {
+                    self.spill_gp_if_not_already(asm, iidx, *reg);
+                }
+            }
+        }
 
-        // Put values that aren't in registers into them -- spilling values if we need them later.
-        for (cnstr, reg) in constraints.iter().zip(cnstr_regs.into_iter()) {
+        // Stage 3.2: For any input constraints that don't yet contain a value, unspill the value.
```
ptersilie, line 485: I don't fully understand this step. Does this unspill values that we have just spilled in 3.1?

ltratt: No: the previous steps have got all values that were into registers into the right registers. Any registers that don't yet have the right value must have been spilled by a previous instruction.

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
```diff
@@ -1814,6 +1764,93 @@ pub(crate) enum GPConstraint {
...
+                    actions.insert(i, (to_reg, RegAction::CopyFrom(from_reg)));
+                }
+                None => {
+                    actions.push((to_reg, RegAction::CopyFrom(from_reg)));
+                }
+            }
+        }
+    }
+
+    // Step 2: Deal with cycles
```
ptersilie, line 1807: Could the `xchg` instruction be useful here? Maybe not now, but in the future?

ltratt: `xchg` is very slow because it is also (weirdly IMHO) a memory barrier. Using a temporary register is much faster.

PR #1633, opened 2025-02-27
Implement idempotent function optimisations.

vext01:
Calls to functions marked `yk_idempotent` are replaced by the return value observed during tracing iff the function's arguments can be proven to be constant.

This is what RPython calls "elidable" functions.

*This is a draft because this is currently missing the necessary trace optimiser supprort*

The optimiser needs to check if the arguments to the idempotent function are constant and if so replace the call with the value stored inside the `idem_const` field of the call instruction.

This will need more tests once we have optimiser support. I'm happy to do those.


ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -776,8 +777,12 @@ impl<Register: Send + Sync + 'static> TraceBuilder<Register> {
         // Check if this is a recursive call by scanning the call stack for the callee.
         let is_recursive = self.frames.iter().any(|f| f.funcidx == Some(*callee));
 
+        // FIXME: currently all functions marked idempotent are outlined, however we should be able
```
ltratt, line 780: I'm not sure this would be a good idea, because at this point in things we don't know what's actually constant or not -- we know *some* constant things, but the optimiser will find out more. Either option feels supoptimal, but inlining things is hard, so I'm inclined not to consider it. [And it would be an `OPT` not a `FIXME`, even if we did want to do it.]

vext01:
Pushed a few more commits.

If you are happy with them, I suggest we squash this down into one co-authored commit for Lukas to give the once over?

tests/c/idempotent.c
```diff
@@ -52,24 +54,23 @@ uintptr_t add(uintptr_t x, uintptr_t y) {
...
 int main(int argc, char **argv) {
   YkMT *mt = yk_mt_new(NULL);
   yk_mt_hot_threshold_set(mt, 0);
   YkLocation loc = yk_location_new();
 
   size_t i = 4;
+  uintptr_t j = 37;
   NOOPT_VAL(loc);
   NOOPT_VAL(i);
+  NOOPT_VAL(j);
```
ltratt, line 66: When I tried doing this, `j` was inlined in `jit-pre-opt` (whether by LLVM or us I wasn't sure), which is why I used `invis_int` to be sure the constant couldn't appear. Are we sure that hasn't happened here?

vext01: Here's the jit-pre-opt:

```
--- Begin jit-pre-opt ---
; compiled trace ID #0

func_decl add (i64, i64) -> i64
func_decl fprintf (ptr, ptr, ...) -> i32
global_decl @stderr
global_decl @.str

entry:
    %0: ptr = param Direct(6, -48, 8)
    %1: ptr = param Register(3, 8, [])
    %2: ptr = param Register(14, 8, [])
    %3: ptr = param Register(12, 8, [])
    header_start [%0, %1, %2, %3]
    %5: i64 = load %2
    %7: i64 = call @add(%5, 2i64) <idem_const 39i64>
    %10: i1 = eq %5, 37i64
    guard true, %10, [2:%0_5: %0, 2:%0_6: %1, 2:%0_7: %2, 2:%1_0: %3, 2:%13_0: %5, 2:%14_0: %7, 2:%15_0: %5] ; trace_gidx 0 safepoint_id 3
    %13: i64 = call @add(%5, 2i64) <idem_const 39i64>
    %15: ptr = lookup_global @stderr
    %16: ptr = load %15
    %17: i64 = load %1
    %18: ptr = lookup_global @.str
    %19: i32 = call @fprintf(%16, %18, %17, %7, %13)
    %20: i64 = load %1
    %21: i64 = add %20, 18446744073709551615i64
    *%1 = %21
    %23: i1 = eq %21, 0i64
    guard false, %23, [2:%0_5: %0, 2:%0_6: %1, 2:%0_7: %2, 2:%1_0: %3, 2:%19_3: 1i1] ; trace_gidx 1 safepoint_id 5
    header_end [%0, %1, %2, %3]
--- End jit-pre-opt ---
```

I think this is OK? In both cases `j` is passed to `add()` as a variable.

ltratt:
Please squash.

vext01:
Squashed. Also removed that comment you pointed out above. We can let the code talk for itself IMHO.

vext01:
@ptersilie mind having a quick look at this? (Both Laurie and I have contributed to this PR).

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -2539,13 +2545,17 @@ pub struct DirectCallInst {
     args_idx: ArgsIdx,
     /// How many arguments in [Module::args] is this call passing?
     num_args: u16,
+    /// If this is a call to an idempotent function, then the index of the constant containing the
```
ptersilie, line 2548: This comment doesn't quite parse. I think it's trying to say:
> The index of the constant we can replace this function with, iff this function is idempotent and the call's arguments are proven to be constant.

ltratt: I think that the comment is trying to do too much and we should get rid of the second clause. So something like "If this is a call to an idempotent function, then this will contain a [ConstIdx] referencing the return value" might be better?

vext01: Force pushed a fix

ptersilie:
Looks good.

vext01:
This needs a ykllvm sync once the compiler change has merged.

vext01:
ykllvm synced. Ready to merge.

vext01:
OK to squash? The second commit I just pushed (adding barrier tests) should stay separate though.

vext01:
Squashed.

PR #1635, opened 2025-03-02
We always comment at `asm.offset` so remove the parameter.

ltratt:
This parameter serves no purpose, and probably never will.


vext01:
Took me a moment to understand what you meant.

We only ever add comments for the current offset, so `self.comment` can access that directly rather than having the callee pass an offset.

PR #1636, opened 2025-03-03
Squash `mprotect` calls into one

ltratt:
This PR changes `patch_address` to `patch_addresses` so that we don't call `mprotect` once per guard: on large traces, we have *lots* of guards (sometimes low hundreds!), so we only want to ask the kernel and MMU to do this once if possible.


ltratt:
I should have said: on at least one benchmark (havlak) I measure ~5% improvement with this. [There is a lot of noise in such benchmarks right now, so I'd consider this an approximate figure, but it does suggest that the PR is genuinely useful in practise.]

PR #1641, opened 2025-03-06
Don't preemptively spill values.

ltratt:
We once needed this hack when:

  1. our register allocator was more rubbish than it is today;
  2. we didn't have side traces (or, soon to come, bridge/link traces).

The situation now is a bit different:

  a. Preemptively spilling (I think...) doesn't meaningfully help the
     register allocator (or, if it does, that's something I should fix). b. For side traces, we end up unspilling everything we spill, even if there were spare registers we could have kept values in.


vext01:
one more for luck

ltratt:
This *probably* hits a bigger bug I'm working on. I'm going to close it, because it's probably best tackled as part of that wider effort.

PR #1642, opened 2025-03-06
Implement stuff required to elide instruction loading in yklua.

vext01:
This includes:

 - Promote support for pointers.
 - Implementing various todo! relating to constant pointers.
 - Optimisations for ptr_add upon const pointers.
 - SpillState support for const pointers.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
@@ -370,10 +370,16 @@ impl LSRegAlloc<'_> {
         self.spills[usize::from(iidx)] = SpillState::Stack(frame_off);
     }
 
-    /// Forcibly assign a constant to an instruction. This typically only happens when traces pass
-    /// live variables that have been optimised to constants into side-traces.
-    pub(crate) fn assign_const(&mut self, iidx: InstIdx, bits: u32, v: u64) {
+    /// Forcibly assign a constant integer to an instruction. This typically only happens when
+    /// traces pass live variables that have been optimised to constants into side-traces.
+    pub(crate) fn assign_const_int(&mut self, iidx: InstIdx, bits: u32, v: u64) {
         self.spills[usize::from(iidx)] = SpillState::ConstInt { bits, v };
+    }
+
+    /// Forcibly assign a constant pointer to an instruction. This typically only happens when
+    /// traces pass live variables that have been optimised to constants into side-traces.
+    pub(crate) fn assign_const_ptr(&mut self, iidx: InstIdx, v: usize) {
+        self.spills[usize::from(iidx)] = SpillState::ConstPtr(v);
     }
 
     /// Return a currently unused general purpose register, if one exists.
@@ -710,7 +716,7 @@ impl LSRegAlloc<'_> {
                                             }
                                         }
                                     },
-                                    SpillState::ConstInt { .. } => {
+                                    SpillState::ConstInt { .. } | SpillState::ConstPtr(_) => {
                                         // Should we encounter multiple constants in registers
                                         // (which isn't very likely...), we want to spill the one
                                         // in the lowest register, since that's more likely to be
@@ -1143,6 +1149,12 @@ impl LSRegAlloc<'_> {
                 }
                 _ => todo!("{bits}"),
             },
+            SpillState::ConstPtr(v) => {
+                // unwrap cannot fail since pointers are sized.
+                let bitw = self.m.type_(self.m.ptr_tyidx()).bitw().unwrap();
+                assert_eq!(bitw, 64);
+                dynasm!(asm; mov Rq(reg.code()), QWORD v as i64);
+            }
         }
         self.gp_regset.set(reg);
         self.gp_reg_states[usize::from(reg.code())] =
@@ -1220,6 +1232,7 @@ impl LSRegAlloc<'_> {
                         size,
                     },
                     SpillState::ConstInt { bits, v } => VarLocation::ConstInt { bits, v },
+                    SpillState::ConstPtr(v) => VarLocation::ConstPtr(v),
                 },
             }
         }
@@ -1636,8 +1649,8 @@ impl LSRegAlloc<'_> {
                 self.fp_regset.set(reg);
             }
             SpillState::Direct(_off) => todo!(),
-            SpillState::ConstInt { bits: _bits, v: _v } => {
-                todo!()
+            SpillState::ConstInt { .. } | SpillState::ConstPtr(_) => {
+                panic!(); // would indicate some kind of type confusion.
             }
         }
     }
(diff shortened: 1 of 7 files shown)


vext01:
@ltratt you could help me with a test for the `heapvalues.rs` change?

I had a brief look, but I don't know this part of the system at all.

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
```diff
@@ -1143,6 +1149,15 @@ impl LSRegAlloc<'_> {
                 }
                 _ => todo!("{bits}"),
             },
+            SpillState::ConstPtr(v) => {
+                // unwrap cannot fail since pointers are sized.
+                let bitw = self.m.type_(self.m.ptr_tyidx()).bitw().unwrap();
+
+                match bitw {
```
ltratt, line 1156: This is an `assert` masquerading as a `todo` I think?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -2609,7 +2610,11 @@ impl<'a> Assemble<'a> {
                             );
                         }
                         VarLocation::ConstInt { bits, v } => {
-                            self.ra.assign_const(iidx, bits, v);
+                            self.ra.assign_const_int(iidx, bits, v);
+                        }
+                        VarLocation::ConstPtr(v) => {
+                            // unwrap can't fail because pointers are sized.
```
ltratt, line 2616: There is no `unwrap` here.

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
vext01: Yeah, just the result of copying code from elsewhere :) Will fix.

ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -782,7 +782,20 @@ impl Opt {
...
+                    let Const::Ptr(cptr) = self.m.const_(cidx) else {
+                        panic!();
+                    };
+                    // Important details:
+                    //  - offsets can be negative
+                    //  - LLVM semantics require the offset to be sign extended or truncated to
+                    //    the "pointer index type" before computing the offset.
+                    let cidx = self
+                        .m
+                        .insert_const(Const::Ptr((*cptr as isize + off as isize) as usize))?;
```
ltratt, line 795: We avoid `as` if at all possible. These should probably be `isize::try_from(...).unwrap()` (etc.) and you will need to decide if you want `saturating_add` or `wrapped_add` etc. to deal with poison cases. `+` will bork in debug mode when it detects overflow, which is not what we want, particularly as these situations do happen because `select` later gets rid of the poisoned value.

ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -1795,6 +1815,24 @@ mod test {
             black_box %0
         ",
         );
+
+        // constant pointer optimisations.
```
ltratt, line 1819: This needs to test all the edge cases too e.g. adding to the max value etc.

ltratt:
This looks good! I do have several comments, mostly because the optimiser has to be kept in tip-top condition. But this is already a solid start!

ykrt/src/compile/jitc_yk/opt/mod.rs
vext01: By `as` casting the offset to isize we truncate or sign extend, as is (I believe) required by LLVM.

This section of the [llvm manual](https://llvm.org/docs/LangRef.html#id238):

> The indices are first converted to offsets in the pointer’s index type. If the currently indexed type is a struct type, the struct offset corresponding to the index is sign-extended or truncated to the pointer index type. Otherwise, the index itself is sign-extended or truncated, and then multiplied by the type allocation size (that is, the size rounded up to the ABI alignment) of the currently indexed type.

I must admit, I find this part of the documentation confusing. I hope I've interpreted it correctly.

ltratt: The problem with `as` is that if you change the types later, it still "works" but does weird things. It's better to use the full type names, as we do elsewhere. Only if you *really* have to should you use `as` (regrettably sometimes you do have to; I don't *think* this is one of those cases, but I might be wrong).

ltratt: BTW, in terms of the `+` stuff, the `BinOp::Add` optimisation probably serves as a useful template for this.

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
vext01: fixed in f9da2df36725f8e28714bb64acd9e52604b9c73f

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
vext01: Fixed in f9da2df36725f8e28714bb64acd9e52604b9c73f

ykrt/src/compile/jitc_yk/opt/mod.rs
vext01: Fixed in 9055875dbb7aca0c38f37f40201e87ec2f050ebd with the caveat that in the JIT IR a constant pointer is assumed to be the same size as `usize`, so we have a similar problem as in the last PR.

https://github.com/ykjit/yk/blob/13c330cccbdc750e05d4a032c84a91ed17ee172d/ykrt/src/compile/jitc_yk/jit_ir/mod.rs#L1298

I'm not sure if I should try to tackle this in this PR or not though.

ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -1762,6 +1776,12 @@ mod test {
 
     #[test]
     fn opt_ptradd() {
+        // FIXME: The offset operand of ptr_add should allow specifying a type (e.g. `ptr_add %0,
```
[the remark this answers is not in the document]
vext01: Decided that I was talking nonsense in be9345628c86eadce254d0b0d2cf85e96a68a4e1

ykrt/src/compile/jitc_yk/opt/mod.rs
ltratt: Don't tackle `Ptr(usize)` in this PR: there's an element of original sin in the JIT IR. Fixing it won't be a small job.

In terms of the 9055875 (which -- thanks GitHub! -- I can't add comments to), `unsafe transmute` is worse than `as` :p

Looking at the code after the change, the good news is that we *do* have to use `as` but it's for the safe (in our case...) idiom of "bitcast from X to Y where X and Y are the same bitwidth". So we probably just need:

```
.insert_const(Const::Ptr((*ptr as isize).wrapping_add(off)))?
```

ltratt:
Only one comment left!

vext01:
OK to squash?

vext01:
squashed.

vext01:
force pushed formatting fix.

PR #1644, opened 2025-03-09
Update ykllvm - additional locations in patchpoint instructions.

Pavel-Durov:
Previously, patchpoint instructions assumed a 1:1 mapping between a register live variable and its location, meaning that each register variable was tracked only in single location, the register itself. This change generalises this to a 1:n mapping, allowing a register live variable to have multiple tracked locations, such as another register or a location on a stack.
This is essential for enabling SWT control point transition between function clones while preserving live variables.


ltratt:
Somehow I missed this one but I think it's implicitly happened with #1645.

PR #1645, opened 2025-03-10
Audit ptr_add and dyn_ptr_add

vext01:
Requires https://github.com/ykjit/ykllvm/pull/234 (and an llvm sync)

### DIFF
--- a/ykrt/src/compile/jitc_yk/opt/mod.rs
+++ b/ykrt/src/compile/jitc_yk/opt/mod.rs
@@ -572,6 +572,7 @@ impl Opt {
         Ok(())
     }
 
+    #[cfg(target_pointer_width = "64")]
     fn opt_dynptradd(
         &mut self,
         iidx: InstIdx,
@@ -581,20 +582,18 @@ impl Opt {
             let Const::Int(_, v) = self.m.const_(cidx) else {
                 panic!()
             };
-            // DynPtrAdd indices are signed, so we have to be careful to interpret the
-            // constant as such.
+            // LLVM IR semantics are such that GEP indices are sign-extended or truncated to the
+            // "pointer index size" (which for address space zero is a pointer-sized integer).
+            // First make sure we will be operating on that type.
             let v = v.to_sign_ext_i64().unwrap();
-            // LLVM IR allows `off` to be an `i64` but our IR currently allows only an
-            // `i32`. On that basis, we can hit our limits before the program has
-            // itself hit UB, at which point we can't go any further.
-            let off = i32::try_from(v)
-                .map_err(|_| ())
-                .and_then(|v| v.checked_mul(i32::from(inst.elem_size())).ok_or(()))
-                .map_err(|_| {
-                    CompilationError::LimitExceeded(
-                        "`DynPtrAdd` offset exceeded `i32` bounds".into(),
-                    )
-                })?;
+            // Now multiply by the element size.
+            //
+            // In LLVM slient two's compliment wrapping is permitted, but in Rust an
+            // `unchecked_mul()` that wraps is UB. Currently the overflow case can't happen because
+            // `inst.elem_size` can only range from [u16::MIN, ui16::MAX].
+            let off = v.checked_mul(i64::from(inst.elem_size())).unwrap();
+            let off = i32::try_from(off).unwrap();
+            // Proceed to optimise.
             if off == 0 {
                 match self.an.op_map(&self.m, inst.ptr(&self.m)) {
                     Operand::Var(op_iidx) => self.m.replace(iidx, Inst::Copy(op_iidx)),
@@ -773,14 +772,22 @@ impl Opt {
         Ok(())
     }
 
+    #[cfg(target_pointer_width = "64")]
     fn opt_ptradd(
         &mut self,
         iidx: InstIdx,
         mut pa_inst: PtrAddInst,
     ) -> Result<(), CompilationError> {
-        let mut off = 0;
+        // LLVM semantics require pointer arithmetic to wrap as though they were "pointer index
+        // typed" (a pointer-sized integer, for addrspace 0, the only address space we support
+        // right now).
+        let mut off: i64 = 0;
+
         loop {
-            off += pa_inst.off();
+            // FIXME: LLVM semantics permit a silent wrap here. We don't support it yet, but to
+            // implement it it (when it arises, and we find a good way to test it) replace
+            // `checked_add` with `wrapping_add`.
+            off = off.checked_add(i64::from(pa_inst.off())).unwrap();
             match self.an.op_map(&self.m, pa_inst.ptr(&self.m)) {
                 Operand::Const(_) => todo!(),
                 Operand::Var(op_iidx) => {
@@ -790,6 +797,8 @@ impl Opt {
                         if off == 0 {
                             self.m.replace(iidx, Inst::Copy(op_iidx));
                         } else {
+                            // FIXME: will panic if the folded offset doesn't fit.
+                            let off = i32::try_from(off).unwrap();
                             self.m.replace(
                                 iidx,
                                 Inst::PtrAdd(PtrAddInst::new(Operand::Var(op_iidx), off)),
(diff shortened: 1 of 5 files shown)


ykrt/src/compile/jitc_yk/arbbitint.rs
```diff
@@ -123,6 +123,11 @@ impl ArbBitInt {
         Some(self.val.sign_extend(self.bitw, 64) as i64)
     }
 
+    /// Sign extend the underlying value and, if it is representable as an `isize`, return it.
+    pub(crate) fn to_sign_ext_isize(&self) -> Option<isize> {
+        Some(self.val.sign_extend(self.bitw, 64) as isize)
```
ltratt, line 128: This isn't safe, because we have no guarantees about `isize`s size. We will have to think *very* carefully about what this function should do, and I don't immediately know if the function can even exist safely.

ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -581,20 +581,27 @@ impl Opt {
...
+            let v = v.to_sign_ext_isize().unwrap();
+            // Now multiply by the element size.
+            //
+            // In LLVM slient two's compliment wrapping is permitted, but in Rust an
+            // `unchecked_mul()` that wraps is UB.
+            //
+            // The unwrap cannot fail on platforms we care about.
+            let Some(off) = v.checked_mul(isize::try_from(inst.elem_size()).unwrap()) else {
+                // Currently the overflow case can't happen (on any platform we care about) because
+                // elem_size can only range from [u16::MIN, ui16::MAX].
```
ltratt, line 596: If this is the case can we not convert the `isize` earlier to a fixed sized integer type?

ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -778,9 +785,24 @@ impl Opt {
...
+        // Although we store ptr_add offsets as an i32, LLVM semantics require the offset to wrap
+        // as though they were "pointer index typed" (a pointer-sized integer, for addrspace 0, the
+        // only address space we support right now).
+        let mut off: isize = 0;
+
         loop {
-            off += pa_inst.off();
+            // LLVM semantics permit silent wrapping.
+            // The unwrap() can't fail on platforms we care about.
+            off = match off.checked_add(isize::try_from(pa_inst.off()).unwrap()) {
```
ltratt, line 796: This would be clearer as just `off.checked_add(...).unwrap()` without the `match` IMHO. The comment is fine *but* I wouldn't mind if it was a couple of lines shorter.

[You'll notice that I'm very protective of the optimiser: it's so dangerous if we get things wrong that code quality *and* code size are constraints here that we don't face elsewhere.]

ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -790,10 +812,13 @@ impl Opt {
                         if off == 0 {
                             self.m.replace(iidx, Inst::Copy(op_iidx));
                         } else {
-                            self.m.replace(
-                                iidx,
-                                Inst::PtrAdd(PtrAddInst::new(Operand::Var(op_iidx), off)),
-                            );
+                            // We can only optimise if the offset can be expressed by an i32.
+                            if let Ok(off) = i32::try_from(off) {
```
ltratt, line 816: When we (eventually) replace `i32`, this will slip by the type checker. I'd prefer the "can panic" variant where we just have `Inst::PtrAdd(..., (i32::try_from(off).unwrap())` without the `if let`.

vext01:
sneaked in a last-second comment tweak.

ykrt/src/compile/jitc_yk/arbbitint.rs
ltratt: The easiest (and perhaps best) thing we could do -- if it's possible! -- is to get rid of this function. Based on the code elsewhere (see a later comment) I'm not even sure if we actually need a "to isize" function yet.

If we do need it, we'll have to add checks (maybe using `static_assert`?) that `isize` and `type_of(self.val)` are the same size.

ykrt/src/compile/jitc_yk/opt/mod.rs
vext01: I'm trying to keep the code as faithful to how LLVM describes things should work. In this case it's clear that we have a very specific type that offset computations should work on:

> The indices are first converted to offsets in the pointer’s index type

And that type is pointer-sized, and thus platform specific in itself, which is reason 1 I'd not like to use a fixed-width type.

The overflow can't happen in this particular case because the largest offset we can compute with a dyn_ptr_add with constant "num elems" and "elem size" would be from an instruction like:

```
dyn_ptr_add %0, isize::MAX, u16::MAX
```

This would fold the latter two operands into a number smaller than isize::MAX:
```
((2**31) - 1) + ((2**16) -1) < ((2 **64) -1)
```

But if at some point we can support the full arbitrarily sized indices for dyn_ptr_add, this wouldn't be the case. That's reason 2 I wouldn't like to use a fixed-width type.

Hope that makes sense?

ykrt/src/compile/jitc_yk/opt/mod.rs
vext01: Did you mean `wrapping_add`?

ltratt: No. What I meant is : the `match` is just a bloated way of saying "we can't handle the None case", which is what `unwrap` is meant to do.

ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -581,20 +581,27 @@ impl Opt {
...
+            // LLVM IR semantics are such that GEP indices are sign-extended or truncated to the
+            // "pointer index size" (which for address space zero is a pointer-sized integer).
+            // First make sure we will be operating on that type.
+            let v = v.to_sign_ext_isize().unwrap();
+            // Now multiply by the element size.
+            //
+            // In LLVM slient two's compliment wrapping is permitted, but in Rust an
+            // `unchecked_mul()` that wraps is UB.
+            //
+            // The unwrap cannot fail on platforms we care about.
```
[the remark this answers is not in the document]
vext01: We have to comment all non-testing unwraps though, don't we?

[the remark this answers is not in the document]
ltratt: No, we stopped doing that ages ago, because we realised most cases were noise. In this case -- a good example! -- the comment line doesn't tell me as a reader anything useful. When we can usefully say "this can't happen because X", it's still worth adding in, but it's not worth killing ourselves and/or padding the code.

ykrt/src/compile/jitc_yk/opt/mod.rs
ltratt: A problem is that `isize` in Rust is broken in the sense that it conflates multiple notions into one: (some bore wrote a blog post on this ages ago... that bore was me). Pointers can -- and e.g. on CHERI are -- smaller than `isize` and that's not fun to deal with. Overall, I would prefer that we don't use `isize` if possible. If we do have to use it, we have to do so very carefully.

ykrt/src/compile/jitc_yk/opt/mod.rs
vext01: Having reviewed this chunk, I don't think this would slip through because constructing the optimised `PtrAddInst` would fail if we had updated the `off` field with a different type: we'd be assigning a `i32` to a different type, so this chunk of code wouldn't compile, right?

We appear not to encounter this case (apart from in the test I devised for it, which I will mark `should_panic` if we really don't want to cover that case for now).

vext01:
I believe that last push addresses all of your comments and everything we discussed offline.

ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -584,23 +585,15 @@ impl Opt {
...
-                // elem_size can only range from [u16::MIN, ui16::MAX].
-                unreachable!();
-            };
-            // Now check we can actually represent the folded offset in an i32.
-            let Ok(off) = i32::try_from(off) else {
-                // Don't do the optimisation.
-                return Ok(());
-            };
+            // `unchecked_mul()` that wraps is UB. Currently the overflow case can't happen (on any
+            // platform we care about) because elem_size can only range from [u16::MIN, ui16::MAX].
```
ltratt, line 593: Let's remove the "any platform we care about" part -- it's just noise.

ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -584,23 +585,15 @@ impl Opt {
...
+            let off = v.checked_mul(i64::from(inst.elem_size())).unwrap();
+            // FIXME: this will panic if we can't represent the folded offset in an i32.
```
ltratt, line 595: This comment is redundant and can be removed.

ltratt:
I have two typo-ish comments about comments. If you're OK with them, you can squash and fix in one go.

vext01:
squashed an ykllvm synced.

PR #1647, opened 2025-03-12
Implement support for outlining over a promoted pointer.

vext01:
This fixes the problem we are seeing in https://github.com/ykjit/yklua/pull/104

### DIFF
--- a/ykrt/src/compile/jitc_yk/trace_builder.rs
+++ b/ykrt/src/compile/jitc_yk/trace_builder.rs
@@ -173,12 +173,8 @@ impl<Register: Send + Sync + 'static> TraceBuilder<Register> {
                 aot_ir::Inst::Promote { tyidx, .. }
                 | aot_ir::Inst::IdempotentPromote { tyidx, .. } => {
                     // Consume the correct number of bytes from the promoted values array.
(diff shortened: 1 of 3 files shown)


ltratt:
Maybe I'm a dummy but I don't see what this PR changes to fix that bug?

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -173,12 +173,8 @@ impl<Register: Send + Sync + 'static> TraceBuilder<Register> {
                 aot_ir::Inst::Promote { tyidx, .. }
                 | aot_ir::Inst::IdempotentPromote { tyidx, .. } => {
                     // Consume the correct number of bytes from the promoted values array.
-                    let width_bits = match self.aot_mod.type_(*tyidx) {
-                        aot_ir::Ty::Integer(x) => x.bitw(),
-                        _ => unreachable!(),
```
vext01, line 178: Here we would encounter a `Ty::Pointer` and panic.

ltratt: Got it. It was hidden by the (not necessarily bad, but not exactly key) refactoring.

PR #1648, opened 2025-03-12
Remove register alloctor hacks that impact correctness

ltratt:
The ultimate aim of this PR is to remove the "dealiasing hack" https://github.com/ykjit/yk/commit/695fda3ff54e36d382597d25a6853480cbffe3c1 (originally introduced in ad9f4254dc64895e2fd3cf9b0f51cedaa6df1c76). I was always a bit queasy about that hack, but I couldn't immediately break it. However it turns out that it causes (in-rare-but-they-can-happen-situations) stack aliasing when jumping back to traces.

Before we can get there, however, we need to make some important changes to the register allocator: in particular, we need to allow a single register to track multiple variables. So the first part of the PR gets the register allocator in shape to remove the dealiasing hack.

While I'm here I also removed the purely performance orientated "meaningful register" hack https://github.com/ykjit/yk/commit/04ae7a60bf8cb783689fd452e4dce7c148e2becc. Ironically, despite this being the same theoretical *cause* of the same incorrectness as the daliasing hack, in practise it also somewhat protected us from it -- by chance!

The exact effects of this PR on performance are a bit hard to predict. https://github.com/ykjit/yk/commit/919b97bc8ceb71f027ec1eddbf8bbea590ed1d27 will (mostly) increase performance but https://github.com/ykjit/yk/commit/04ae7a60bf8cb783689fd452e4dce7c148e2becc will sometimes decrease it (particularly for side traces). My guess is that, overall, any differences will be relatively minor, though individual benchmarks might gain/lose more than that suggests.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
@@ -190,7 +190,7 @@ impl<'a> LSRegAlloc<'a> {
     /// use the set of available registers more efficiently.
     pub(crate) fn expire_regs(&mut self, iidx: InstIdx) {
         for reg in GP_REGS {
-            match self.gp_reg_states[usize::from(reg.code())] {
+            match &mut self.gp_reg_states[usize::from(reg.code())] {
                 RegState::Reserved => {
                     debug_assert!(self.gp_regset.is_set(reg));
                 }
@@ -203,9 +203,10 @@ impl<'a> LSRegAlloc<'a> {
                     self.gp_regset.unset(reg);
                     self.gp_reg_states[usize::from(reg.code())] = RegState::Empty;
                 }
-                RegState::FromInst(reg_iidx, _) => {
+                RegState::FromInst(iidxs, _) => {
                     assert!(self.gp_regset.is_set(reg));
-                    if !self.rev_an.is_inst_var_still_used_at(iidx, reg_iidx) {
+                    iidxs.retain(|x| self.rev_an.is_inst_var_still_used_at(iidx, *x));
+                    if iidxs.is_empty() {
                         self.gp_regset.unset(reg);
                         *self.gp_reg_states.get_mut(usize::from(reg.code())).unwrap() =
                             RegState::Empty;
@@ -215,7 +216,7 @@ impl<'a> LSRegAlloc<'a> {
         }
 
         for reg in FP_REGS {
-            match self.fp_reg_states[usize::from(reg.code())] {
+            match &mut self.fp_reg_states[usize::from(reg.code())] {
                 RegState::Reserved => {
                     debug_assert!(self.fp_regset.is_set(reg));
                 }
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
```diff
@@ -1074,38 +1100,46 @@ impl LSRegAlloc<'_> {
...
+                        self.stack.align(bytew);
+                        let frame_off = self.stack.grow(bytew);
+                        let off = i32::try_from(frame_off).unwrap();
+                        match bitw {
+                            1 => {
+                                if *ext == RegExtension::ZeroExtended {
+                                    dynasm!(asm; mov BYTE [rbp - off], Rb(reg.code()));
+                                } else {
+                                    dynasm!(asm
+                                        ; bt Rd(reg.code()), 0
```
vext01, line 1249: As you mentioned in the commit message, this changes EFLAGS.

Are there situations where instr N+1 relies on the flag assignment of instr N? If so, are you worried that the flag clobbering could introduce hard to track down bugs?

I've been skimming the intel manual again. There are the `pushf*` instructions which save the flags to the stack, but it'd be expensive to do that conservatively (just in case the following instruction requires the flags), and I'm not sure if the architecture of the allocator/codegen allows it to peek ahead and ask "does someone need that flag assignment later?".

Tricky.

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
```diff
@@ -1807,6 +1830,9 @@ pub(crate) enum GPConstraint {
         force_reg: Option<Rq>,
         can_be_same_as_input: bool,
     },
+    /// Align `op`'s upper bits to `out_ext`: the instruction must not use the resulting register
```
vext01, line 1929: I'm not sure what it means to align an operand's upper bits. Can you elaborate?

vext01:
I've done a superficial pass over this. A couple of high-level questions.

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
ltratt: IMHO no-one should rely on CPU flags being held across `assign_regs` calls, and I've adjusted the once instance of that I know of.

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
ltratt: "Align extensions" is the terminology we use elsewhere. So this should be "Align op's extension to out_ext" or similar if that's OK?

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
vext01: OK, cool.

If the docstring for `assign_regs` doesn't already mention that you can't rely on flag assignments, I think that'd be a good addition.

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
vext01: I've just read the implementation of `align_extensions()` and now see the confusion.

Here "align" isn't about memory alignment, it's about **satisfying** extension requirements.

I think naming like `unify_extensions()` would make more sense, but that's one for another day.

I'm happy to go with whatever you want in light of this epiphany :)

ltratt:
While hunting for a bug that may have nothing to do with this PR, I have made a number of simplifications and generalisations in 219ad35.

ltratt:
I was right: the bug had nothing to do with this PR! https://github.com/ykjit/yk/pull/1654 fixes that bug. Once that's merged, I believe this PR is ready for squashing.

ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
ltratt: Yes, another term might be a good idea for a future PR.

vext01:
please squash.

ltratt:
https://github.com/ykjit/yk/pull/1654 will need merging before this PR can merge.

ltratt:
[Squashed and force pushed in advance of that.]

ltratt:
@vext01 We should now be able to merge this one.

PR #1652, opened 2025-03-13
Fix trace disassembler.

vext01:
Laurie noticed two jumps in deopt routines printing as:

```
jmp 0x0000000000001660
...
jmp 0x0000000000001660
```

Looking at the raw instruction bytes, the jumps are RIP-relative jumps that are supposed to jump to the same address. Well, they do, but the address printed is nonsense.

We were driving the disassembler wrong. This fixes that, and now the instructions print like:

```
jmp 0x00007F6DBC8AB120
...
jmp 0x00007F6DBC8AB120
```

i.e. with proper virtual addresses.


ltratt:
Thanks!

vext01:
Realised that this cast isn't needed. Force pushed.

ltratt:
One last try, because it would be useful to have this in master.

PR #1656, opened 2025-03-17
Deal with tracing starting in nested interpreter frames.

ltratt:
The "main" test here is `trace_while_executing1.c`, which models a failure we see occasionally in yklua: a JIT frame ends up calling an interpreter frame, which starts tracing, and tracing doesn't falls back into the JIT frame, which then happens to deopt. It took me quite a while to convince myself that's what happening, but it really is!

`trace_while_executing2.c` is a sort-of variant test of something similar that we happened to already deal with correctly, but which no test captured.

### DIFF
--- a/tests/c/trace_while_executing1.c
+++ b/tests/c/trace_while_executing1.c
@@ -1,0 +1,72 @@
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_LOG=4
+//   env-var: YKD_LOG_IR=jit-pre-opt
+//   stderr:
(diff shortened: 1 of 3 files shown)


ltratt:
Yes. Though that doesn't *directly* cause problems, because we later throw the trace out for having unmappable code. But it confuses the life out of `mt.rs` and means we can do a *lot* of pointless work that we know for sure we'll have to throw away later.

vext01:
Right, so we'd assume that we'd only ever deopt when the thread is in the Executing phase, but we've found a scenario where that's not the case.

Makes sense to me.

PR #1658, opened 2025-03-18
Don't clobber `rax` with `icall` unless doing a varargs call.

ltratt:
This commit refines indirect calls so that in the non-varargs case (the common case!), we use `rax` for the function pointer. This prevents us having to clobber another register, reducing register pressure (which is always high around calls on x64).


ltratt:
List timed out, so we should probably just retry.

PR #1661, opened 2025-03-19
Start understanding aborted traces

ltratt:
This PR nudges things to the point that we can actually start understanding *why* we're seeing `trace-aborted`.

[I've already used the information in this PR to hack "fixes" for some scenarios, though they're not yet general enough for me to want to raise them as PRs.]

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -422,13 +422,14 @@ impl MT {
     pub fn control_point(self: &Arc<Self>, loc: &Location, frameaddr: *mut c_void, smid: u64) {
         match self.transition_control_point(loc, frameaddr) {
             TransitionControlPoint::NoAction => (),
-            TransitionControlPoint::AbortTracing => {
+            TransitionControlPoint::AbortTracing(ak) => {
                 let thread_tracer = MTThread::with_borrow_mut(|mtt| match mtt.pop_tstate() {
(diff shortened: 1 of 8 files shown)


ykrt/src/mt.rs
```diff
@@ -1279,6 +1283,30 @@ enum TransitionControlPoint {
...
+/// Why did we abort tracing?
+#[derive(Debug)]
+enum AbortKind {
+    /// While tracing we fell back from an interpreter to a JIT frame.
+    BackIntoExecution,
+    /// While tracing, we encountered a compiled trace.
+    EncounteredCompiledTrace,
+    /// Tracing continued while the interpreter frame address changed.
+    OutOfFrame,
+    /// We unrolled a loop (i.e. we traced a [Location] more than once).
```
vext01, line 1295: I had to think about this one a bit.

 - If we encounter the same location twice and it's the place we started tracing we stop tracing and compile.
 - If we enounter the same location twice and it's *not* the place we started tracing, this is indicative of an inner loop being unrolled, which is what the `Unrolled` abort kind is for.
 
 Is that right?

ltratt: You're right.

[Just to be clear: this PR does not change behaviour, it just makes it more obvious what happened.]

ltratt:
Oops, force pushed `cargo fmt` fix.

PR #1664, opened 2025-03-20
Implement support for threads.

ptersilie:
There wasn't support for threads in yk as it requires the shadow stack to be a thread local. This was previously not possible due to us using LLVM to compile our traces. Now that we use our own JIT compiler, there isn't much standing in the way of thread support as this commit shows.

The first (and easier) step is to change the shadow stack to a thread local in ykllvm. We then intercept the creation of threads so we can allocate a new shadowstack for each thread. At JIT compile time, we first calculate the offset of the thread local in relation to the segment register `fs`. At JIT runtime (when the segment register has been updated by a new thread) we can then use this offset to calculate the thread local's pointer.


tests/c/pthread_create.c
```diff
@@ -1,27 +1,86 @@
...
+
+void overwrite() {
+  int a = 99;
+  int b = 99;
+  int c = 99;
+  int d = 99;
+  int e = 99;
+  int f = 99;
+  int g = 99;
+  int h = 99;
```
vext01, line 33: Is this test x64-specific? I assume a..h are to fill the right number of registers so that spills happen?

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -716,6 +716,12 @@ impl GlobalDecl {
         self.is_threadlocal
     }
 
+    /// Return whether the declaration is a thread local.
```
vext01, line 719: Is this docstring right?

ykrt/src/thread_intercept.rs
```diff
@@ -1,11 +1,59 @@
...
+
+extern "C" fn wrap_thread_routine(tgt: *mut c_void) -> *mut c_void {
+    let str = CString::new("shadowstack_0").unwrap();
+    let tgt = unsafe { Box::from_raw(tgt as *mut Target) };
+    // Obtain address of a shadowstack_0 symbol
+    let shadowstack_symbol_addr = unsafe { dlsym(null_mut(), str.as_ptr()) };
+    if shadowstack_symbol_addr.is_null() {
+        panic!("Unable to find shadowstack address")
+    }
+    let newsstack = unsafe { malloc(10000) };
```
vext01, line 20: shall we define the new thread's shadow stack size with a const somewhere?

tests/c/pthread_create.c
```diff
@@ -1,27 +1,86 @@
...
-//    not yet implemented: Allocate and set thread-local ShadowStack instance. No support for threads yet!
+//   stderr:
 //    ...
+//    ret1 = 0xdead
+//    ...
+//    oof2: {{_}} 20
+//    ...
+//    ret2 = 0xbaff1ed
+
+// This test checks that new threads create their own shadow stack. For this is
```
vext01, line 11: nit: s/is/it/ on the end of the line

vext01:
Looks good Lukas! Glad you managed to codegen the thread local access directly.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ptersilie: Huh. I swear I fixed it. Must have gotten lost during my various rebases. Will fix.

ykrt/src/thread_intercept.rs
ptersilie: Ah yes, and we should probably make it the same size as the shadowstack in main.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ptersilie: [ca9db08](https://github.com/ykjit/yk/pull/1664/commits/ca9db08f029e2607ce42ebd8a7e99078e391f6a7)

ykrt/src/thread_intercept.rs
ptersilie: [489f989](https://github.com/ykjit/yk/pull/1664/commits/489f9892308ea8b0114ebd1264ea19d3dc086d00)

tests/c/pthread_create.c
ptersilie: [4115f6f](https://github.com/ykjit/yk/pull/1664/commits/4115f6f15d858560e7c7a756896f49d234672e13)

vext01:
Please squash.

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -143,6 +143,27 @@ static LK_PATCH: Mutex<()> = Mutex::new(());
 /// x86 NOP opcode.
 const NOP_OPCODE: u8 = 0x90;
 
+/// Returns the offset of the given thread local in relation to the segment register `fs`. At JIT
+/// runtime we use this offset to calculate the absolute address of the thread local for each
+/// thread.
+fn get_tls_offset(name: &std::ffi::CString) -> i32 {
```
ltratt, line 149: Is there a reason for the signedness here? Presumably it's always a -ve offset?

ltratt: Let's document that and add an `assert` guaranteeing that. [Or change it to a `u32`, but I take your point about dynasm.]

ltratt: [Feel free to squash those changes straight down.]

ptersilie: Fixed.

PR #1665, opened 2025-03-24
Add the ability to compile inner loops discovered during tracing an outer loop.

vext01:
Please critique this approach to compiling inner loops found when tracing an outer loop.

It works by counting how many times we encounter the control point since starting tracing, and chopping a prefix (of that many control points) off the mapped trace before the trace builder starts proper.

All yk tests pass. Not tried on Lua tests or benchmarks yet.

Showing this now only to get an idea as to whether this is worth continuing. This is woefully untested and unpolished. Don't review the code too carefully, just the overarching approach for now.

### DIFF
--- a/tests/c/compile_inner_loop_simple.c
+++ b/tests/c/compile_inner_loop_simple.c
@@ -1,0 +1,70 @@
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_LOG=4
+//   stderr:
+//     ...
+//     yk-jit-event: start-tracing
+//     i=0, loc0
+//     i=1, loc1
+//     yk-jit-event: stop-tracing (inner loop detected)
+//     i=2, loc1
+//     yk-jit-event: enter-jit-code
+//     i=3, loc1
+//     i=4, loc1
+//     yk-jit-event: deoptimise
+//     i=5, loc0
+//     ...
+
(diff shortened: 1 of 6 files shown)


ltratt:
Can you benchmark this on (at least) permute and see if it reduces performance non-determinism? You should easily see this with something like `multitime -n 10 -s 0 ...`.

ltratt:
OK so the crucial part of this is:

```rust
    // If this thread is tracing, count how many times we've seen the control point since starting.
    //
    // This has to happen before the `loc.is_null()` check below because we have to include calls
    // to the control point with null locations.
    use ykrt::MTThread;
    MTThread::inc_cp_idx();
```

This will clearly work *but* it imposes a penalty -- one we'll definitely notice -- on the control point.

We really would prefer to avoid this if possible and have the relevant detection isolated to the mapper / trace_builder. That said, to understand how much of an effect "compiler inner loops" has, we can temporarily keep this in, even if we don't want to merge it (or replace it soon afterwards).

[I don't immediately know the best way to avoid doing this, but if we think carefully about it, something might occr to us!]

vext01:
I've had an idea for avoiding that cost, but I'll do the measurements on what we have now first.

ltratt:
I think we should close this one: I still would like to investigate this idea, but it's fallen down the TODO list quite a bit, as other changes have partly mitigated the problem it's trying to solve, and it's also (inevitably) fallen a long way behind `master`.

PR #1667, opened 2025-03-27
Give `icmp_guard` a better way of assigning registers.

ltratt:
Previously we allocated registers for both the `icmp` and the `guard` in one go, but that causes unspilling issues (and also caused us to generate slightly worse code than necessary, because we needed 3 registers in a combined icmp/guard rather than than 2 and then 1 registers separately). To work around that, however, seems tricky, because `assign_reg`s can set CPU flags.

This commit offers an alternative mechanism (which, currently, we don't need, but we probably will soon so it's `todo`ed) where in the worst case we will need to `push`/`pop` if we encounter a non-zero-extended `i1` that we need to spill.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
@@ -414,11 +414,35 @@ impl LSRegAlloc<'_> {
 
     /// Assign general purpose registers for the instruction at position `iidx`.
     ///
-    /// This is a convenience function for [Self::assign_regs] when there are no FP registers.
+    /// Note: this can change CPU flags. It is therefore undefined behaviour to check CPU flags
+    /// after this function has been called.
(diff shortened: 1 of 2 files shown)


vext01:
The only nit I have is that "cpu flags" is a bit vague. There's probably hundreds of flags in an intel chip, but I think you are specifically speaking about the `EFLAGS` register here? This is the name of the register containing the flags used for branching decisions etc.

ltratt:
In an ideal world, the register allocator would be architecture independent, so I would prefer not to be *too* specific.

ltratt:
https://github.com/ykjit/yk/pull/1667/commits/5c0e6d774c269e43568c55040fe08be6e2b262b8 should fix the bug we hit. [I'm finding it hard to `tryci` these days as I almost always end up with "trace too long" on some random test.] Will need squashing.

ltratt:
Actually, let's hold off on this. I might have a better idea that's also simpler.

PR #1668, opened 2025-03-27
Give `icmp_guard` a better way of assigning registers.

ltratt:
Previously we allocated registers for both the `icmp` and the `guard` in one go, but that:

  1. causes unspilling issues
  2. also caused us to generate slightly worse code than necessary, because we needed 3 registers in a combined icmp/guard rather than than 2 and then 1 registers separately)

It turns out that I really undercooked this on my first attempt! `icmp_guard` is actually more special than it first appears: we want to get a temporary register *but* unlike other instructions we don't want to put all possible inputs into registers (if nothing else, in general there are far more live variables than there are registers!). So our previous "give me a `GPConstraint::Temporary`" went wrong when all the registers contained values the guard needed: we hadn't told `assign_gp_regs` what we really wanted.

This commit is a pragmatic approach: we create a special function for `icmp_guard` that finds (or forces) an empty register, but doesn't remove information about values in registers (though it might spill one register). It also does that in a way that doesn't touch CPU flags.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs
@@ -412,6 +412,52 @@ impl LSRegAlloc<'_> {
         None
     }
 
+    /// Find -- or make! -- an empty register suitable for the guard portion of an `icmp_guard`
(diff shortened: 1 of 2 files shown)


ltratt:
I'm probably going to close this one: I woke up with a better idea.

vext01:
third time lucky :)

PR #1670, opened 2025-04-01
Tweak the null location in yk.h.

vext01:
I need the ability to (from within the C code of an interpreter) say whether a location is a null location.

I tried using the (unused, it seems) `YKLOCATION_NULL` macro in `yk.h`, but it's not valid C. You can't cast an integer to a struct.

This change replaces it with an inline function that does a null location check by comparing the state field to zero. I've done it this way to avoid calling into ykcapi.so, which would come at a higher cost.

This is required for an upcoming change to fix placement of YkLocations in yklua.

### DIFF
--- a/ykcapi/yk.h
+++ b/ykcapi/yk.h
@@ -1,6 +1,7 @@
 #ifndef YK_H
 #define YK_H
 
+#include <stdbool.h>
 #include <stdint.h>
 #include <sys/types.h>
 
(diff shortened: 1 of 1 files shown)


ltratt:
It's interesting you've done this as I've been thinking of doing this myself in the last couple of days, though with a different use case! In essence, I *think* we should pull out the "is yklocation null" check *from* the control point, and force the user to do it. This would solve the problem in https://github.com/ykjit/yk/pull/1665 where we can't equate "number of control_point" calls with "number of non-null-yklocations". Now you've give us this API, this becomes more concrete!

PR #1671, opened 2025-04-01
Fix codegen for indirect calls returning void.

ltratt:
This was a nasty one in the sense that it tended to lead to a segfault: we didn't pass the operand containing the indirect pointer to the register allocator, assuming (incorrectly!) the value was in rax. Much unhilarious comedy ensued.

Fixes a bug Edd encountered.


ltratt:
Force pushed `cargo fmt` (mea culpa!).

PR #1674, opened 2025-04-02
Fix spilling

ltratt:
It turns out I made a somewhat horrible mistake (fixed in https://github.com/ykjit/yk/commit/8714c92e5cd34e20d9803d6daee4cc954fad7151) which then misled me into some other silly changes (fixed in https://github.com/ykjit/yk/commit/e9d125ed20a82d51085699a19cea4e09692134f8).


ltratt:
I'm an idiot. Force pushed a `cargo fmt`.

ltratt:
`list` timed out... So probably worth kicking.

PR #1676, opened 2025-04-03
Add the ability to attach debug strings to hot locations.

vext01:
The api call yk_location_set_debug_str() allows the interpreter author to attach debug strings to hot locations. These strings are then printed in log messages (where possible).

This functionality is guarded behind the (pre-existing) "ykd" feature to avoid the costs for "production" builds (when you do `cargo build -p ykcapi`).

The logging facility is now behind a macro so that hot location debugging is zero-cost if not compiled in.


ltratt:
Works for me!

vext01:
Force pushed clippy fixes.

PR #1678, opened 2025-04-04
Split `transition_control_point` into two.

ltratt:
This function is, to put it mildly, unwieldy. It also handles two very different cases: what we do when we're not tracing (the common case) and when we're tracing (the uncommon case). These two things were interleaved in a way that has become gradually unfathomable, at least to my pea brain.

This commit in essence :

  1. Duplicates `transition_control_point` into `transition_control_point_not_tracing` and `transition_control_point_tracing` and

  2. Removes all the irrelevant code from each function.

There is no meaningful functional change as a result of this commit.


vext01:
Yes, thank you! I agree!

vext01:
Richards crashed. Retrying.

PR #1680, opened 2025-04-07
Add codegen support for `memset`.

ltratt:


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1647,6 +1647,11 @@ impl<'a> Assemble<'a> {
                 self.cg_memcpy(iidx, dst, src, len, is_volatile);
                 Ok(())
             }
+            x if x.starts_with("llvm.memset") => {
```
vext01, line 1650: The "startswith" is making me nervous.

Do we mean to cover:
 - `llvm.memset.inline`
 - `llvm.memset.element.unordered.atomic`
 - `llvm.memset.p0.i32`
 
 I see only tests for `llvm.memset.p0.i64`.

ltratt: Point taken! Force pushed an update which only works for `llvm.memset.p0.i64`.

ltratt:
Force pushed a Clippy fix.

PR #1685, opened 2025-04-10
Adjustments required now that yk_outline omits AOT IR.

vext01:
Requires https://github.com/ykjit/ykllvm/pull/239

Note that we cannot delete the "skipping over promotes/yk_debug_str" mode in the trace builder, because you can still outline over functions not marked yk_outline. For example, for indirect calls and if a yk_outline function calls a non-yk_outline function.


tests/c/idempotent_outline.c
```diff
@@ -24,12 +24,12 @@
...
 //     --- Begin jit-pre-opt ---
 //     ...
-//     %{{_}}: i{{size}} = call @idem1(%{{_}}, %{{_}}) <idem_const 10i{{size}}>
+//     %{{_}}: i{{size}} = call @idem1(%{{_}}, %{{_}}) <idem_const 28i{{size}}>
 //     ...
 //     --- End jit-pre-opt ---
 //     --- Begin jit-post-opt ---
 //     ...
-//     %31: i32 = call @fprintf(%{{_}}, %{{_}}, 4i{{size}}, 10i{{size}})
+//     %31: i32 = call @fprintf(%{{_}}, %{{_}}, 4i{{size}}, 28i{{size}})
```
vext01, line 32: By the way, the expected output in this test looked wrong to me. I'll investigate this after, as there might be a (separate) bug in our idempotent promotes.

vext01:
This will need a ykllvm sync.

tests/c/idempotent_outline.c
ltratt: I am not sure I understand how this PR can change the idem_const. That feels like this PR (or the ykllvm PR) *might* itself introduce a new bug?

vext01: I actually think the opposite. In "untickles" an existing bug.

Manual inspection of the test, I think `28` is the correct value to promote. Do you agree?

PR #1686, opened 2025-04-10
Fix a bug where we asked the wrong thing for its size.

ltratt:
We previously nearly always checked that stores didn't overlap by 8 bytes, rather than the actual bytewidth of the value in question. Since we don't support more than 64 bit values currently, this wasn't a potential crashing bug, but it made the analysis unnecessarily pessimistic.

### DIFF
--- a/ykrt/src/compile/jitc_yk/opt/heapvalues.rs
+++ b/ykrt/src/compile/jitc_yk/opt/heapvalues.rs
@@ -95,12 +95,10 @@ impl HeapValues {
                 // we're writing 8 bytes and we're storing to `%3 + 8` and `%3 + 24` we can be
                 // entirely sure the stores don't overlap: in any other situation, we assume
                 // overlap is possible. This can be relaxed in the future.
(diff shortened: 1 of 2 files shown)


jacob-hughes:
This looks sensible but I don't think I understand the bigger picture. Is the problem that you were previously checking the pointer size rather than the pointee size?

So IIUC this is a bug where you had a big trap door just waiting for someone to fall through, but it never happened because so far we only work with ptr-sized values which just so happened to be the size we were checking?

jacob-hughes:
Ah cool! Does that mean the comment above the match is still relevant?

ltratt:
The comment is still correct. The new test makes sure the problem can't creep back in.

jacob-hughes:
Eagerly awaiting your new test-in-ci-driven-development blog post :)

Please squash

ltratt:
If I wrote such a post, recent experience suggests that I'd probably forget to proof read it!

Squashed.

PR #1690, opened 2025-04-11
Fix idempotent promotes.

vext01:
We were immediately consuming the return value from the promote buffer upon encountering a call to an idempotent function. If the callee calls functions that then promote their own values, then this means we consume the promote buffer in the wrong order.

This change ensures we only consume the return value from the promote buffer *after* we have returned from the idempotent function, and thus all inner promotions will already have been handled.

We achieve this by introducing a "pending" state for idempotent constants. When the trace builder encounters a call to an idempotent function it sets the `idem_const` field to `IdemConst::Pending`. When the trace builder has finishes outlining, it checks if it just emitted a call instruction with a pending `idem_const`, and if so it replaces it with `IdemConst::Const(cidx)`.

### DIFF
--- a/ykrt/src/compile/jitc_yk/trace_builder.rs
+++ b/ykrt/src/compile/jitc_yk/trace_builder.rs
@@ -799,23 +799,13 @@ impl<Register: Send + Sync + 'static> TraceBuilder<Register> {
         } else {
             // This call can't be inlined. It is either unmappable (a declaration or an indirect
             // call) or the compiler annotated it with `yk_outline`.
-            let idem_const = if func.is_idempotent() {
-                let aot_ir::Ty::Func(fty) = self.aot_mod.type_(func.tyidx()) else {
-                    panic!();
-                };
-                let ret_tyidx = self.handle_type(self.aot_mod.type_(fty.ret_ty()))?;
-                Some(self.promote_bytes_to_const(ret_tyidx)?)
-            } else {
-                None
-            };
-
             self.outline_until_after_call(bid, nextinst);
             let jit_func_decl_idx = self.handle_func(*callee)?;
             let inst = jit_ir::DirectCallInst::new(
                 &mut self.jit_mod,
                 jit_func_decl_idx,
                 jit_args,
-                idem_const,
+                None, // Note: the `idem_const` field may be populated later.
             )?
             .into();
             self.copy_inst(inst, bid, aot_inst_idx)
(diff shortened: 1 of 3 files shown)


vext01:
Force pushed a fix.

ltratt:
In retrospect I might have rushed my review of this.

First random thought: should we bork as soon as we detect a promote in an idempotent? IMHO if anyone's got an idempotent function that complicated they've almost certainly done something wrong!

Second thought: IIUC the code, we insert a "Call(Pending)" (forgive the syntax!) and later change that to "Call(ConstIdx)"? I wonder if we've done that in a more complex way than necessary, in the sense that JIT IR now has to know about something that one might hope is restricted to trace_builder. I wonder if instead we can just insert "Call" (no `Pending` or whatnot) and later change it with a call with a `idemcost`?

vext01:
> should we bork as soon as we detect a promote in an idempotent?

We will bork because:
 - ykllvm will bork if there's a promote in a yk_outline function (see [here](https://github.com/ykjit/ykllvm/pull/239/files#diff-9a98c82d803f659ad8f727223c544d73eab8f1967e132949117784b434d64381R1654) and [here](https://github.com/ykjit/ykllvm/pull/239/files#diff-e839fc88e7a34cfbd35a3c0630f59ac418a7d6a4938fc269189947b4440242aaR8)), and
 - `yk_idempotent` implies `yk_outline` (see [here](https://github.com/ykjit/ykllvm/blob/74aa4970a44a78ada3224dae7063b60ffe5f627c/clang/lib/CodeGen/CodeGenModule.cpp#L2422))

We can add a ykllvm test that explicitly checks `yk_idempotent` triggers the same error for peace of mind.

> I wonder if we've done that in a more complex way than necessary, in the sense that JIT IR now has to know about something that one might hope is restricted to trace_builder. I wonder if instead we can just insert "Call" (no Pending or whatnot) and later change it with a call with a idemcost?

I considered this design as well. There's not much in it, but I went with the "pending" design because it feels more explicit and less error prone (because `NotRequired` cannot be conflated with "required but we somehow didn't fill in the required value").

I don't feel all that strongly about this.

ltratt:
> ykllvm will bork if there's a promote in a yk_outline function (see [here](https://github.com/ykjit/ykllvm/pull/239/files#diff-9a98c82d803f659ad8f727223c544d73eab8f1967e132949117784b434d64381R1654) and [here](https://github.com/ykjit/ykllvm/pull/239/files#diff-e839fc88e7a34cfbd35a3c0630f59ac418a7d6a4938fc269189947b4440242aaR8))

How does this work with indirect function calls at runtime? i.e. I'm in `yk_outline` then I dynamically call a function with a `promote`.

>>    I wonder if we've done that in a more complex way than necessary, in the sense that JIT IR now has to know about something that one might hope is restricted to trace_builder. I wonder if instead we can just insert "Call" (no Pending or whatnot) and later change it with a call with a idemcost?
>I considered this design as well. There's not much in it

The more I reflect on it, the less I like polluting the JIT IR for something that no other part of the system needs to know about. I think we should try and keep the JIT IR as clean as possible.

ltratt:
> We'd have to add a dynamic check the the outlining mode of the trace builder.

Agreed. Which makes me wonder if we even need to bother with the static checks?

vext01:
> Agreed. Which makes me wonder if we even need to bother with the static checks?

If we want to go ahead with https://github.com/ykjit/ykllvm/pull/239 then they will be essential, since we won't have IR for `yk_outline` functions, and thus wouldn't be able to say if they contain `yk_promote` dynamically.

ltratt:
Point taken.

vext01:
(we can still opt not omit IR for `yk_outline` functions, but: a) we will need to change the representation of safepoints in the AOT IR (to make them optional), and b) I was hoping that by minimising the amount of IR we emit, we can speed up linking).

ltratt:
The JIT IR is more important than linking time IMHO.

vext01:
AFAICS, the JIT IR shouldn't be affected, since safepoints are an AOT IR concept.

vext01:
@ltratt I've pushed a couple of commits. One makes the JIT IR as it was before, whilst also fixing idempotent promotes, the second addresses a question in the discussion above.

The third is just a merge that will be squashed away.

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -1426,38 +1441,6 @@ impl<Register: Send + Sync + 'static> TraceBuilder<Register> {
...
-                            // Because we haven't emitted any instructions (only outlined) since
-                            // emitting the call to the idempotent function, the call we need to
-                            // modify is still the last instruction in the JIT module.
-                            let last_iidx = self.jit_mod.last_inst_idx();
-                            // It's essential we use `inst_raw` here, since a call to an idempotent function is
-                            // followed by an call to `__yk_idempotent_promote_*()` which the trace
-                            // builder will convert into a [Inst::Copy] to the first call. If
-                            // we were to decopy the instruction we'd consume from the promote
-                            // buffer twice!
-                            let inst = self.jit_mod.inst_raw(last_iidx);
```
ltratt: I am so happy the `inst_raw` call has gone! FWIW, if you need this again, you *probably* really want `inst_nocopy`.

ltratt:
Thanks! Please squash.

Issue #1689, opened 2025-04-11
Optimiser can encounter impossible guard with yklua


vext01:
yk d303fc5177

Input lua program:
```lua
do
        local a
        for A = 0, 132 do
                a = a == 0 or 0
        end
end
```

Can crash like this:
```
$ YKD_SERIALISE_COMPILATION=1 ~/research/yklua/src/lua boom.lua
thread '<unnamed>' panicked at ykrt/src/compile/jitc_yk/opt/mod.rs:628:13:
assertion failed: (inst.expect() && v.to_zero_ext_u8().unwrap() == 1) ||
    (!inst.expect() && v.to_zero_ext_u8().unwrap() == 0)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

But not with the optimiser off:
```
$ YKD_OPT=0 YKD_SERIALISE_COMPILATION=1 ~/research/yklua/src/lua boom.lua
$
```

Peeling bug?

ltratt:
Initial thoughts: the optimiser appears to be doing the right thing. What *seems* to be happening is that we're tracing the wrong thing.

I applied this diff to yklua:

```diff
--- src/lvm.c
+++ src/lvm.c
@@ -1185,6 +1185,7 @@ Instruction load_inst(uint64_t pv, const Instruction *pc) {
   pc = (Instruction *) yk_promote((void *) pc); \
   uint64_t pv = yk_promote(cl->p->proto_version); \
   i = load_inst(pv, pc); \
+  printf("%lu %d\n", pc, i);\
   pc++; \
 }
 #else
```

and ran it with `YKD_LOG=4` and got lots of output the end of which is:

```
...
103708816 2147450881
103708820 -2147483464
103708832 229577
103708808 8355901
103708828 7
103708832 229577
103708808 8355901
103708816 2147450881
103708820 -2147483464
103708832 229577
103708808 8355901
103708828 7
103708832 229577
103708808 8355901
103708816 2147450881
103708820 -2147483464
103708832 229577
103708808 8355901
yk-jit-event: start-tracing
103708828 7
103708832 229577
103708808 8355901
yk-jit-event: stop-tracing
103708816 2147450881
103708820 -2147483464
103708832 229577
103708836 16842822
<program finishes>
```

[...]

PR #1691, opened 2025-04-14
Deal correctly when peeling discovers guards that always fail.

ltratt:
When we're peeling we might discover that the peeled iteration can never execute the same as the first iteration. Consider a loop such as:

```
for x in 0..100 {
  if x % 2 == 0 { ... }
  else { ... }
}
```

In other words, successive iterations of the loop flip between the true/else branches. Thus, whatever iteration we trace, the peeled iteration will never be able to execute to its end: it will always fail at a guard before that point.

When we're optimising the peeled iteration, our analyses can thus detect guards that will always fail. As soon as we discover one of those, there's no point trying to optimise the guard or anything after it in the trace: the guard will always fail at run-time.

This commit thus fixes a panic first discovered in https://github.com/ykjit/yk/issues/1689.

### DIFF
--- a/ykrt/src/compile/jitc_yk/opt/mod.rs
+++ b/ykrt/src/compile/jitc_yk/opt/mod.rs
@@ -151,6 +151,46 @@ impl Opt {
             match inst {
                 Inst::TraceHeaderStart | Inst::TraceHeaderEnd(_) => panic!(),
                 Inst::TraceBodyStart => (),
+                Inst::Guard(ginst) => {
+                    // When we're peeling we might discover that the peeled iteration can never
+                    // execute the same as the first iteration. Consider a loop such as:
+                    //
+                    // ```
+                    // for x in 0..100 {
+                    //   if x % 2 == 0 { ... }
+                    //   else { ... }
+                    // }
+                    // ```
+                    //
+                    // In other words, successive iterations of the loop flip between the true/else
+                    // branches. Thus, whatever iteration we trace, the peeled iteration will never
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/jitc_yk/opt/mod.rs
```diff
@@ -151,6 +151,46 @@ impl Opt {
...
+                        if (ginst.expect() && v.to_zero_ext_u8().unwrap() == 0)
+                            || (!ginst.expect() && v.to_zero_ext_u8().unwrap() == 1)
+                        {
+                            // This guard will always fail. We could be more clever than just
+                            // stopping optimising at this point. For example, we could introduce a
+                            // new type of "always fail" trace; or Tombstone all the subsequent
+                            // instructions to avoid the code generator doing pointless work. But
+                            // we're lazy, and simply stopping optimising the peeled loop at this
+                            // point is correct.
+                            break;
```
vext01, line 188: If I understand correctly, we will emit no further instructions into the trace, right? An "impossible" guard basically terminates the trace.

ltratt: No: we stop *optimising* at that point. All the instructions after the guard are left as-is. They won't ever be exeucted, so it's sort-of wasted effort, but being clever here is somewhat complicated (and probably not worth the effort).

ltratt:
Well, that's embarrassing -- I managed to mangle a string in an unconnected function and not even spot it in the diff! Fixed in the force push.

PR #1693, opened 2025-04-16
Remove the well-formedness check for "impossible" guards.

ltratt:
I once thought that "0 == 1"-esque guards were impossible, but peeling makes them possible. This check is thus, in general, wrong. [Of course, it *might* sometimes pick up genuine bugs, but we have no way of distinguishing those from cases we actually want to allow.]


ltratt:
Ouch, I hadn't realised some of those imports are needed in *some* configurations. Force pushed a fix.

PR #1695, opened 2025-04-16
Remove ineffectual flags from yk-config.

vext01:
After a bit of a wild goose chase it looks like these switches aren't required.

The (standard) llvm passes in question are already skipped if a function is marked optnone, which (currently) all functions are after the last high-level IR pass has finished.

Basically, any pass that uses `skipFunction()` will not be run on a function that is marked optnone.

(There are some other skip conditions too. See `skipFunction()` in ykllvm/llvm/lib/IR/Pass.cpp for details).

### DIFF
--- a/bin/yk-config
+++ b/bin/yk-config
@@ -65,23 +65,6 @@ handle_arg() {
             # Add wrapper for `pthread_exit`.
             OUTPUT="${OUTPUT} -Wl,--wrap=pthread_exit"
 
-            # Disable machine passes that would interfere with block mapping.
-            #
-            # If you are trying to figure out which pass is breaking the
-            # mapping, you can add "-Wl,--mllvm=--print-before-all" and/or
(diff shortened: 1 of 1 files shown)


ltratt:
Dumb question: do we `optnone` e.g. `yk_outline` functions?

vext01:
As it happens, i've just tried a change that:
 - doesn't mark anything optnone after the high level optimisations
 - modifies `skipFunction()` to skip things not marked `yk_outline`

This has an equivalent effect to what you've just suggested and would enable a slew of backend optimisations on anything that would never be traced.

Sadly it makes ykllvm seg fault when dealing with some stackmap related stuff. This is likely to be something to do with our stackmap customisations. I think it's going to be worth a look though, because it could really help interpreter performance.

[There's some irony here too -- outlined functions will never need stackmaps, but we know from last week's experiment that removing them there upsets the PT decoder somehow]

PR #1696, opened 2025-04-17
Add a Lua part to the test suite.

ltratt:
*This is a request for comments*

It's become gradually harder and harder to test the "tracing statemachine" aspects of yk as they've become more sophisticated. Unit tests in yk require contorting the code, sometimes making it difficult to read. C tests become baroque, and in some cases I can't really work out how to express them.

For a little while I've been toying with writing a little interpreter which would allow us to write "normal program" tests that tickle all the bits we want. It has now occurred to me that yklua can serve this purpose (I'm slow)! I've included one test as an example of what this looks like.

Right now this is a draft because -- if we think this a good idea! -- I will at least need to tidy up `build.rs` so that it rebuilds when ykllvm changes (otherwise we're going to really confuse ourselves).

### DIFF
--- a/.buildbot_dockerfile_debian
+++ b/.buildbot_dockerfile_debian
@@ -5,7 +5,7 @@ RUN --mount=target=/var/lib/apt/lists,type=cache,sharing=locked \
   rm -f /etc/apt/apt.conf.d/docker-clean && \
   apt-get update && \
   apt-get -y install clang-15 make curl procps file git cmake python3 \
-    libtinfo-dev libzip-dev mold ninja-build gdb pipx && \
+    libtinfo-dev libzip-dev mold ninja-build gdb pipx rsync && \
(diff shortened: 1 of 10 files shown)


vext01:
I don't object.

It's more direct than adding tests elsewhere (e.g. yklua-tests, which is only run overnight, or in yklua PRs) and it also means we don't have to port all the langtester stuff elsewhere too.

ltratt:
[Just force pushed a Clippy warning. In fairness, it's right and I'm wrong, in this regard at least!]

ltratt:
That turned out to be not much fun to fix at all. https://github.com/ykjit/yk/pull/1696/commits/c06c7ce32f2467de92ae1da3d94b0d0948804f84 fixes it, but it's a hack -- and may be the best we can do. It does at least pass `tryci`!

Will need squashing.

vext01:
please squash.

ltratt:
Squashed.

PR #1703, opened 2025-04-22
Naked functions are now stable.

ltratt:
Necessary to unbreak building with the latest rust nightly.


vext01:
Hah. I was just preparing the same PR!

PR #1707, opened 2025-04-23
Don't bork on non-UTF-8 debug strings.

ltratt:
yklua can produce non-UTF-8 debug strings (Lua source files are not required to be UTF-8). We could handle this (albeit awkwardly) in yklua, but it's a general problem: C strings aren't required to be UTF-8. So rather than forcing the interpreter to deal with this -- which, probably, would mostly involve it doing a UTF-8 check and then saying "no UTF-8 string for this debug string" -- when we encounter a non-UTF-8 string put a (hopefully clear!) placeholder in its stead.


ltratt:
Good idea! Addressed in f39dcfb8.

vext01:
please squash.

ltratt:
Squashed.

PR #1708, opened 2025-04-23
Better debug string tests

ltratt:
The heart of this PR is 03eb753879117c3882c7a48207ad3aecc8d0078c, which allows us to much more accurately test what we're tracing.

There *may* be an issue with this PR: for me `cargo test` fairly frequently fails on these tests on b16 with "Trace too long". I have a small hope that this is just a weird quirk of b16 or my setup there, but it may well be that yklua in debug mode really is creating traces that (at least at the moment!) are too long because we're tracing quite a bit of Rust code. Possible remedies include:

* Only run these tests in release mode (i.e. `cargo test --release`).
* Wait for "exclude stuff from PT traces so we only trace stuff with IR".

### DIFF
--- a/tests/langtest_lua.rs
+++ b/tests/langtest_lua.rs
@@ -91,7 +91,11 @@ fn build() {
     // there's nothing to do, it'll finish quicker than we can notice!
     let mut make_cmd = Command::new("make");
     make_cmd
-        .args(["-j".into(), num_cpus::get().to_string()])
+        .args([
+            "-j".into(),
+            num_cpus::get().to_string(),
+            "MYCFLAGS=-DYKLUA_DEBUG_STRS=2".into(),
+        ])
         .env(
             "PATH",
             format!(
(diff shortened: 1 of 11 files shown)


vext01:
FWIW, i see "trace too long" lots on b16 with debug builds.

BTW, did you mean to assign someone to this?

ltratt:
> FWIW, i see "trace too long" lots on b16 with debug builds.

I *suspect* this will happen on other machines too, but *perhaps* if we like the PR, we can let it be tried on b13 in merge? Or should I just not run the tests in debug `cargo test`? That does suck, but it is a pragmatic thing to do until/unless we solve "trace too long" another way.

vext01:
More tests are always good. And these look worthwhile.

I don't mind. We can either wait to fix "trace too long" or disable some tests. I leave the decision up to you.

ltratt:
https://github.com/ykjit/yk/pull/1708/commits/fbeca582dbbcbe9889a1d0151b0d2933e46dd339 makes the Lua tests only run in release mode but does so in a way that adjusting that should be fairly easy if/when we want to do so.

vext01:
please squash (if you need to)

ltratt:
Squashed.

PR #1709, opened 2025-04-24
Fix pthread_create test segfaulting.

ptersilie:
This test would occasionally segfault as the while-loops would sometimes create extremely large traces that the hwt collector couldn't handle. We can work around this by inserting small delays into the loops. Additionally we increase the main loop iteration count so we a complete trace before the program terminates.

### DIFF
--- a/tests/c/pthread_create.c
+++ b/tests/c/pthread_create.c
@@ -7,6 +7,7 @@
 //    oof2: {{_}} 20
 //    ...
 //    ret2 = 0xbaff1ed
+//    ...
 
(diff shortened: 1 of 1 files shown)


ptersilie:
Ready for re-review.

ltratt:
Wouldn't it be better to `ignore-if` this on swt?

ptersilie:
Either would fix this, but I'd rather not trace anything alltogether as that's not the goal of this test.

ptersilie:
Squashed.

PR #1710, opened 2025-04-24
Reenable the `OutsideYk` timing state.

ptersilie:
When using `YKD_STATS` I noticed that the `OutsideYk` time wasn't reporting correctly. This was because this state was only set after deopting, but not after tracing, compiling, etc. Setting this state at the end of the controlpoint, which is just before we return to the interpreter, will measure the time we spent outside of yk correctly again.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -481,6 +481,7 @@ impl MT {
                     &format!("tracing-aborted: {ak}"),
                     loc.hot_location()
                 );
+                self.stats.timing_state(TimingState::OutsideYk);
             }
             TransitionControlPoint::Execute(ctr) => {
                 yklog!(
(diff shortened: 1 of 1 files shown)


ltratt:
This is too crude and will cost us, since now we'll pay the price of that call on every control point call (and we want control point calls to be as fast as possible). What we need to do is fine those places where we aren't going back to "outside" properly. Since compiling (etc.) happens in a thread, there's no need to go back to "outside" at that point. I can believe we might not handle the "we've finished tracing" cases properly: we need to find each one and put the timing call in there.

ptersilie:
You're right! I've moved it into stop-tracing, stop-side-tracing, and tracing-aborted. See [d1153ae](https://github.com/ykjit/yk/pull/1710/commits/d1153ae0419f61149476021ddf6cb1e4d66ef03b)

ltratt:
This looks good to me. Please squash.

ptersilie:
Squashed.

PR #1711, opened 2025-04-24
Various clippy fixes.

ptersilie:
With the recent rust update clippy is complaining about a lot of code. There's two main fixes I can see:

1) Inlining variables into the format string.
2) Combining the conditions of nested `if`s via `&&`.

I can agree with number 1), however I'm not fully sold on 2) yet. I don't really mind having nested `if`s, and fixing all of them changes our code **a lot**, because it also results in reformatting of those if statements. Have a look at the diff and see what you think @vext01 @ltratt.

### DIFF
--- a/hwtracer/src/block.rs
+++ b/hwtracer/src/block.rs
@@ -31,7 +31,7 @@ impl fmt::Debug for Block {
                 first_inst,
                 last_inst,
             } => {
-                write!(f, "Block({:x}..={:x})", first_inst, last_inst)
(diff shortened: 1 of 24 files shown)


ltratt:
This is an improvement IMHO: I've long been waiting for `if let` to be allowed as a general clause! Unless @vext01 feels differently, I think this can go in.

vext01:
> Combining the conditions of nested ifs via &&

Does it insist on this if there's a comment between the conditionals? Often I split conditionals so I can add a comment and help comprehension.

PR #1712, opened 2025-04-24
Avoid intermediate spills

ltratt:
This PR optimises the (relatively frequent) case that a variable *will* be spilled at the end of a trace but we spill it earlier to a different place. We then end up shuffling spill locations around before jumping, which is slow and pointless.

The main part of this PR is https://github.com/ykjit/yk/commit/9460d699710bf332049a929a25dfcdafe11466e8, but that enables a very simple second optimisation in https://github.com/ykjit/yk/commit/a35326569979e6b3d591003a665f248a5baa5a24.

This is a small speedup, on a few examples I tried perhaps of the order of 1%. Better than nothing, though.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/rev_analyse.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/rev_analyse.rs
@@ -50,6 +50,11 @@ pub(crate) struct RevAnalyse<'a> {
     /// instruction %1, we would like %0 to already be in rax; when generating code for instruction
     /// %2, we would like %) to already be in rdi".
     reg_hints: Vec<Vec<(InstIdx, Register)>>,
+    /// For each instruction, record whether we know whether it should spill to in advance.
+    /// Currently this only happens for instructions referenced in the header/body end, where we
+    /// have to put the value in a specific place on the stack before we loop/jump. `u32::MAX` is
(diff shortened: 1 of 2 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/rev_analyse.rs
```diff
@@ -109,7 +133,7 @@ impl<'a> RevAnalyse<'a> {
                     .downcast::<X64CompiledTrace>()
                     .unwrap();
                 let vlocs = ctr.entry_vars();
-                // Side-traces don't have a trace body since we don't apply loop peeling and thus use
+                // Connector traces don't have a trace body since we don't apply loop peeling and thus use
```
ptersilie, line 136: When did we rename side-traces to connector traces? We are still using `TraceKind::Sidetrace` though it seems?

ltratt: This is the fix :)

ptersilie: Oohh, right! :-)

ltratt:
I got done by the new Clippy! Force pushed the simple fix necessary.

ltratt:
And I forgot to rustfmt. Force pushed...

PR #1713, opened 2025-04-25
Align jump targets inside traces.

ptersilie:
This change aligns the jump targets of the trace loop and side traces jumping back into a parent trace to 16 bytes in order to improve or at the very least stabilise performance.

Before this, changes as small as a single extra instruction inside the trace header, had the potential to significantly alter the performance of the trace, making it difficult to test changes in for example yklua.

There's some more easy to reach fruits hanging around:
- Experiment with aligning jump targets to a cache line
- Replace `nop` slides with multibyte `nop`s as described in https://stackoverflow.com/questions/18113995/performance-optimisations-of-x86-64-assembly-alignment-and-branch-prediction/18279617#18279617
- Maaaaaybe experiment with aligning deopt jump targets

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -2692,6 +2692,12 @@ impl<'a> Assemble<'a> {
             };
             self.header_start_locs.push(loc);
         }
+        // 16-byte align this jump target to improve performance.
+        let off = self.asm.offset().0;
+        let align = off.next_multiple_of(16);
+        for _ in 0..(align - off) {
+            self.asm.push(NOP_OPCODE);
+        }
         match self.m.tracekind() {
             TraceKind::HeaderOnly => {
                 dynasm!(self.asm; ->tloop_start:);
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -2692,6 +2692,12 @@ impl<'a> Assemble<'a> {
             };
             self.header_start_locs.push(loc);
         }
+        // 16-byte align this jump target to improve performance.
+        let off = self.asm.offset().0;
+        let align = off.next_multiple_of(16);
```
vext01, line 2697: If you are already aligned, do we insert 16 nops?

vext01:
Asides that one comment, is it work making that logic a function for easy experimentation later?

ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
ptersilie: Good question. Let me check that.

ptersilie: It doesn't. There's even an example in the docs: https://doc.rust-lang.org/std/primitive.u32.html#method.next_multiple_of

ltratt: Yep, `next_multiple_of` does exactly what we want!

PR #1714, opened 2025-04-25
PT trace only the executable segment of the main object.

vext01:
deltablue (benchmark that spends a lot of time mapping):
  - time spent mapping reduced in the order of ~10x
  - overall performance improved about 15%

This change means that we can no longer spot longjmp in code outside of
the traced range. Whereas before, we'd spot calls to longjmp when the PT
decoder was disassembling foreign code, now what we do is do some basic
"control flow integrity" checking in the trace builder, checking that
each mappable block is either: the entry block of a function, a return
block from a function, or a static successor of the last mappable block
we processed.

(Annoyingly this means we have to keep track of the previous block ID
(which may be unmappable) *and* the previous mappable block ID
separately -- although Lukas and I think we could probably do away with
unmappable blocks and thus also the former, but that's another story)

Although to detect longjmp this way we would only need to check this
property when we return from a function, it turns out to be simpler to
implement for every mappable block (return or not) and arguably it's a
really good sanity check that we should probably have had in place from
the start.

An added bonus is that this approach is tracer-agnostic, so the longjmp
tests work for software tracing too.

I had hoped to detect signal handlers this way too, but unfortunately I
can't see an universally unambiguous way to distinguish them from
regular calls. Since signal handlers weren't really the aim of this PR,
I won't let this hold us up. For now I've added an ignored test showing
the issue: currently signal handlers that have IR get inlined.

### DIFF
--- a/hwtracer/src/perf/collect.c
+++ b/hwtracer/src/perf/collect.c
@@ -38,6 +38,8 @@
 // The bit in the IA32_RTIT_CTL MSR that disables compressed returns.
 #define IA32_RTIT_CTL_DISRETC 1 << 11
 
+void get_tracing_extent(char **, size_t *, size_t *);
+
 enum hwt_cerror_kind {
   hwt_cerror_unused = 0,
   hwt_cerror_unknown = 1,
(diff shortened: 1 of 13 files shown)


ltratt:
Please force push.

vext01:
Force pushed and updated the PR description.

Requires: https://github.com/ykjit/ykllvm/pull/263

ptersilie:
`trace_builder` part looks ok.

PR #1715, opened 2025-04-26
Introduce a `tracing` timing state.

ltratt:
For quite a while the `deopt` timing state has shown very large values, but when I investigated the code carefully there was nothing which could possibly account for this. It turns out that most of the time we've thought is "deopting" is when `mt.guard_failure` calls `start_recorder`, which is nothing to do with deopting per se.

This commit thus introduces a new timing state `tracing` (used for both "normal" and "side" tracing) which gives us a much better sense of what costs we're incurring where. Deopting itself turns out to be a minor performance overhead (one that it would be nice to improve, but it wouldn't be a massive factor); tracing (which is "start tracing, actually trace a loop, stop tracing") is an order of magnitude more expensive.


ltratt:
Ah, right, I'm going to need to get the clock stuff in yklua sorted.

PR #1716, opened 2025-04-27
Allow writing of 64-bit constant ints.

ltratt:
A recent yklua change means that we do now see this case in the wild.


ptersilie:
Looks like this also depends on https://github.com/ykjit/yklua/pull/120? Now that that is merged I will try again. Or do we need to wait for https://github.com/ykjit/yk-benchmarks/pull/21 first?

ltratt:
This bit hadn't occurred to me: this PR needs the new yklua for buildbot.sh. Fixed in the force push.

PR #1720, opened 2025-04-29
Cache perf fd and mmapped data structures.

vext01:
This is similar to a revert of 3345db18, but I did it manually to force me to think about what I was doing. I also put all the cached bits into a struct.

(Note that I also didn't revert the bugfix hidden in 3345db18)

Improves the time spent tracing (as reported by YKD_LOG_STATS) and overall performance a bit too.

# Deltablue before

```
1: lua harness.lua deltablue 5 12000
real        9.625       0.388       8.963       9.837       10.017
user        17.971      1.021       16.339      18.067      19.258
sys         19.923      2.485       15.222      20.640      22.107
```

# Deltablue after

```
1: lua harness.lua deltablue 5 12000
real        8.043       0.653       6.817       8.158       8.762
user        15.019      1.727       11.609      15.702      16.391
sys         17.363      4.113       10.052      20.088      20.619
```

# Richards before

```
1: lua harness.lua richards 5 100
real        29.138      0.216       28.959      29.076      29.549
user        33.300      0.592       32.619      33.152      34.096
sys         8.249       0.966       6.370       8.631       9.119
```

# Richards after

```
1: lua harness.lua richards 5 100
real        28.939      0.238       28.598      28.942      29.274
user        32.733      0.570       32.084      32.682      33.408
sys         6.776       1.321       5.266       6.307       8.970
```

### DIFF
--- a/hwtracer/src/perf/collect.c
+++ b/hwtracer/src/perf/collect.c
@@ -76,6 +76,21 @@ struct hwt_perf_ctx {
 };
 
 /*
+ * Setting up perf is slow, so we cache a number of things.
+ *
+ * Since we attach perf to a thread, all of this is per-thread too.
+ */
+struct hwt_perf_thread_cached {
+  int perf_fd;
+  void *aux_buf;
+  size_t aux_bufsize;
+  void *base_buf;
+  size_t base_bufsize;
+};
+
+thread_local struct hwt_perf_thread_cached thread_cached = { -1, NULL, 0, NULL, 0 };
+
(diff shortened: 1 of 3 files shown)


ltratt:
A question (not to be done in this PR!): do we initialise the cached perf stuff in a thread? If not, it might be worth doing so: though perhaps this doesn't work if each thread has to make its own perf connection? Or perhaps one thread can initialise a collector and "pass" it to another thread?

PR #1721, opened 2025-04-30
Replace `cmp R, 0` with `test R, R`.

ltratt:
To my surprise, this is slightly, but measurably, faster (~1% of bigloop).


vext01:
Nice.

`cmp` is basically `sub` in disguise, which I guess it more expensive that `test` which is a `and` in disguise.

PR #1723, opened 2025-05-01
Optimise load/binop/store code generation.

ltratt:
For JIT IR along the lines of:

```
%0: ptr = param reg
%1: i8 = load %0
%2: i8 = add %1, 1i8
*%0 = %2
```

We previously generated code such as:

```
mov r11, [r15+0x7abcdef]
add r11, 0x1
mov [r15+0x7abcdef], r11
```

i.e. the `load`, `add`, and `store` all lead to one x86 instruction. This commit is able in some situations to boil this down to one instruction:

```
add ptr [r15+0x7abcdef], 0x1
```

This reduces register pressure and seems to make the CPU much happier. This idiom crops up semi-frequently in yklua (e.g. around `for` loops). It makes bigloop 40% faster -- which perhaps also shows that benchmark is becoming less useful to us! -- and also speeds up some other benchmarks, though in the order of a small number of %.

[Another way of looking at this commit is that it shows us that our backends will need their own IR if we want to make this code more generalisable and tidier.]

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -595,7 +595,7 @@ impl<'a> Assemble<'a> {
 
     /// Codegen an instruction.
     fn cg_insts(&mut self) -> Result<(), CompilationError> {
-        let mut iter = self.m.iter_skipping_insts();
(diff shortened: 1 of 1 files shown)


vext01:
Nice!

ptersilie:
> `add ptr [r15+0x7abcdef], 0x1`

I did not know you could do that!

PR #1724, opened 2025-05-01
Robust gmake check

jacob-hughes:


jacob-hughes:
I'm not sure. I doubt it, but if you symlinked it to BSD make then this would definitely break. I added this check because I don't have a `gmake` binary so I check `make` as a fallback. I suppose then there's no need for the `make --version` check to be in the else branch.

ltratt:
Yes, I think it's fine to say "start with `gmake` and then fallback to `make` if it doesn't exist". It is possible (not likely, at the moment, I admit) that someone makes a GNU-compatible make that doesn't report `GNU Make` with `--version`.

jacob-hughes:
Ah I see what you mean now. In that case maybe I should leave the version check in the else branch in-case a GNU compatible make doesn't have GNU in the version?

ltratt:
TBH, version string checks are always a bit dodgy because they over-specify. I'd say "try gmake; if that doesn't work use make" and if the user doesn't have a make that works for us... Well, they'll soon find out ;)

jacob-hughes:
Something like this (39c1edcef12f2dce15e4cb0b1c4cae3eb201624f)?

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #1725, opened 2025-05-01
Canonicalise `const predicate val` to `val predicate const` for `ICmp`.

ltratt:
This makes codegen simpler (at the moment we only handle `val predicate const` so canonicalising here saves us having to implement the other case).


ptersilie:
Huh, a segfault on `closure.lua`. Could be random. Want me to try again?

ltratt:
It's because I'm an idiot and got some of the cases wrong. This time I've actually checked! I also realised from the yklua test suite that this PR unlocks a new case for `smax` which we hadn't encountered before, so that's in the force push too.

PR #1726, opened 2025-05-02
Fix PT decoding over block terminated by synthetic conditional branches.

vext01:
...and some other bits, see commits.

Requires https://github.com/ykjit/ykllvm/pull/248, so will need a sync.

### DIFF
--- a/hwtracer/src/perf/collect.c
+++ b/hwtracer/src/perf/collect.c
@@ -4,6 +4,7 @@
 // buffer. It would probably be better to either never handle it (for
 // simplicity) or fully handle it.
 
+#include <assert.h>
 #include <err.h>
 #include <errno.h>
 #include <fcntl.h>
@@ -231,7 +232,7 @@ static bool handle_sample(void *aux_buf, struct perf_event_mmap_page *hdr,
                         memory_order_relaxed);
 
   void *next_sample = data_tmp;
-  while (next_sample != data_tmp_end) {
+  while (next_sample < data_tmp_end) {
     struct perf_event_header *sample_hdr = next_sample;
     struct perf_record_aux_sample *rec_aux_sample;
     switch (sample_hdr->type) {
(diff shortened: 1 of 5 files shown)


vext01:
Just noticed the second commit message has a typo:
"Commit assumption checking in many places" -- `s/Commit/Strengthen/g`

Will fix when i force push the sync.

ptersilie:
Since you're already force-pushing typos, might as well do these:

If a block is terminated by two conditional branches th**e**n we will have
to consume between 1 and 2 TNT bits, depending upon whether the fi**r**st
branch was taken or not.

This fixes a long-standing PT bug where the PT decoder would crash
expect**ing** a TIP packet from an uncompressed return. We were consuming one
TNT where we needed to consume two, meaning that the otherwise
unconsumed TNT was being used to decide~~d~~ if a return was compressed or
not.

hwtracer/src/perf/collect.c
```diff
@@ -257,9 +257,15 @@ static bool handle_sample(void *aux_buf, struct perf_event_mmap_page *hdr,
       // Shouldn't happen with PT.
       errx(EXIT_FAILURE, "Unexpected PERF_RECORD_LOST_SAMPLES sample");
       break;
+    case PERF_RECORD_AUX_OUTPUT_HW_ID:
+      errx(EXIT_FAILURE, "Unimplemented PERF_RECORD_AUX_OUTPUT_HW_ID sample");
+      break;
     }
     next_sample += sample_hdr->size;
   }
+  if (next_sample > data_tmp_end) {
```
ltratt, line 266: Is this an `assert` in disguise or do we expect to recover from this situation?

ltratt:
My one comment (if it's correct!) can be addressed in a force push.

vext01:
Thanks, force pushed fixes.

ltratt:
Please force push a ykllvm sync when that PR merges.

ltratt:
@vext01 You can force push a ykllvm sync now.

vext01:
Force pushed.

vext01:
Force pushed another ykllvm sync.

PR #1727, opened 2025-05-02
Only optimise load/bin/store for consts.

ltratt:
We *could* do non-constants too, but the register allocator currently isn't really set up for that. Since this is probably a small win, just don't try and optimise it.

This stops us hitting a `todo` that is stopping yklua (and friends) PRs from merging.


jacob-hughes:
LGTM. Does this one need merging first? https://github.com/ykjit/yklua/pull/123

ltratt:
On the contrary, what's in this PR is necessary for the yklua PRs :)

PR #1731, opened 2025-05-06
Fix trace test to use wildcard for shadow stack load

jacob-hughes:
In [1] we add a debug variable which inserts an extra load when the test is compiled in debug mode. This means that the variable match for the `ptr_add` instruction does not necessarily take an operand whose variable name increments by 1.

[1]: https://github.com/ykjit/ykllvm/pull/253


jacob-hughes:
This can be merged whenever, but it's probably best to wait for https://github.com/ykjit/ykllvm/pull/253 to check that it fixes the bug we saw before.

ltratt:
Please force push a ykllvm sync (so that we test ykllvm after 253 has merged there).

PR #1732, opened 2025-05-06
Sync ykllvm

jacob-hughes:


jacob-hughes:
This can be merged once https://github.com/ykjit/ykllvm/pull/249/files lands.

jacob-hughes:
This failed with a `Trace too long` error. I think this is the unrelated non-determinism we were discussing yesterday?

ltratt:
It's really unusual to see this on b13, so if it fails again, it might indicate a deeper issue.

PR #1733, opened 2025-05-07
Fix ring buffer handling

vext01:
See commits.


vext01:
I've just pushed a commit to turn on PT caching again. This appears to work reliably with the ring buffer fix.

(no squashing required)

ltratt:
That failure has nothing to do with this PR AFAICS.

PR #1735, opened 2025-05-07
Fix tests.

ltratt:
This is a subtle mistake teased out by proptest when we had a `usize` whose where bit n was set when converting to an n-bit sized integer. In essence, we were comparing something slightly different with Rust's `try_from` vs `ArbBitInt`'s `to_sign_ext` function.


vext01:
Nice catch!

Typo in commit message "whose where bit". Fix if you think it's worth it :)

ltratt:
Force pushed a fix.

PR #1740, opened 2025-05-08
Fix benchmarks

ptersilie:


ptersilie:
Ah whoops. Left the other commit in there. Can I force push?

.buildbot.sh
```diff
@@ -238,12 +238,16 @@ for tracer in $TRACERS; do
 
         # Do a quick run of the benchmark suite as a smoke test.
         pipx install rebench
-        git clone https://github.com/ykjit/yk-benchmarks
+        git clone https://github.com/ykjit/yk-benchmarks --recurse-submodules
```
ltratt, line 241: Since we're here, we might as well add `--depth=1` to reduce network traffic.

ltratt:
Please force push away the ykllvm sync + add the `--depth=1` at the same time.

.buildbot.sh
ptersilie: Right!

ltratt: I didn't know that was a thing, but I think I'm "yes"!

ptersilie:
Done.

.buildbot.sh
ptersilie: Ah, didn't see this. One sec.

.buildbot.sh
```diff
@@ -238,12 +238,16 @@ for tracer in $TRACERS; do
 
         # Do a quick run of the benchmark suite as a smoke test.
         pipx install rebench
-        git clone https://github.com/ykjit/yk-benchmarks
+        git clone https://github.com/ykjit/yk-benchmarks --recurse-submodules --depth 1
```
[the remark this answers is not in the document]
ptersilie: Think it weirds both ways but changed it just in case.

ptersilie:
Okay, that should be all.

PR #1745, opened 2025-05-13
Lessen overhead

ltratt:
These are two slightly separate commits: the first tries to reduce the quantity of non-IR code that PT will have to disassemble; and the second tries to speed up functions like `yk_promote`. The latter is probably the more important of the two (though it depends on the first).


ltratt:
Force pushed a Clippy fix.

ltratt:
Force pushed a fix.

PR #1748, opened 2025-05-15
Refine the job queue

ltratt:
This PR refines the job queue somewhat:

1. https://github.com/ykjit/yk/commit/37dc61b5e98950553351548589c7e9cd25ccdb29 lazily creates worker threads only when no thread is available to process a job.
2. https://github.com/ykjit/yk/commit/0c2bb32f55f5a489e0faab49bdf1ff91f9c410d3 only wakes up other threads after a root compile job has finished if there are other jobs waiting on it.

On b16, this PR has a small but measurable effect on some benchmarks (e.g. ~2% on CD). I don't know if it'll make much difference on machines with very few cores, but I doubt it will hurt there.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -104,6 +104,8 @@ pub struct MT {
     )>,
     /// The hard cap on the number of worker threads.
     max_worker_threads: AtomicUsize,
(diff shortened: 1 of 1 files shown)


ltratt:
Force pushed Clippy fix.

PR #1750, opened 2025-05-16
Break the job queue out into its own module.

ltratt:
`mt.rs` is quite big enough already, and this is a relatively discrete chunk of code.


ptersilie:
Good idea.

PR #1751, opened 2025-05-16
Add codegen support for abs.

ltratt:


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -1949,6 +1954,34 @@ impl<'a> Assemble<'a> {
...
+                    in_ext: RegExtension::ZeroExtended,
+                    force_reg: None,
+                    clobber_reg: false,
+                },
+                GPConstraint::Temporary,
+            ],
+        );
+        match bitw {
+            64 => {
+                dynasm!(self.asm
```
vext01, line 1974: I can't believe there's no `abs()` in x86!

I found [this implementation](https://stackoverflow.com/a/11927940/109414) which I find easier to understand:
```
mov ebx, eax
neg eax
cmovl eax, ebx ;if eax is now negative, restore its saved value
```

What do you think? People in the comments claiming this is what clang does and is fast.

ltratt: See what godbolt says clang does now? `cmovl` isn't obviously fast FWIW, so I'm mildly sceptical (but only mildly).

vext01:
Merging this into my branch, json now gives:
```
thread '<unnamed>' panicked at ykrt/src/compile/jitc_yk/codegen/x64/lsregalloc.rs:1386:34:
internal error: entered unreachable code
```

(Could be unrelated)

vext01:
pushed a test that generates that same panic

ltratt:
OK that should make things work now (TM).

vext01:
json no longer crashes now :)

tests/c/abs.c
```diff
@@ -0,0 +1,38 @@
+// Run-time:
+//   env-var: YKD_LOG_IR=aot,jit-pre-opt
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_LOG=4
+//   stderr:
+//     ...
```
vext01, line 6: Can you fill in the expected output?

PR #1752, opened 2025-05-16
Retry errors.

ltratt:
This commit does several related things (which is how they all got jumbled together):

  1. It gives queue jobs their own `struct`. Jobs now have a "main" function and a "connector_failed" function.
  2. If a job is waiting on a connector trace and that fails, the "connector_failed" function is called and the job is removed from the queue.
  3. When a trace fails to compile, we now mark that as an error; when the threshold (for `Location`s and `Guard`s) is exceeded, that place will no longer try to trace again.
  4. Guards now track how often tracing/compiling have led to errors.

Put another way: when tracing / compilation goes wrong, locations and guards can no longer get "stuck". Before this commit, guards could get stuck in a "we tried tracing once and it didn't work" state forever and locations could get stuck in a "we're waiting for a connector trace, but it's failed and will never be available" state forever.

Tiny benchmarks tend not to encounter these things, but larger benchmarks are more likely to. LuLPeg and Storage both benefit by 8%, Towers by 3%, and DeltaBlue by 2%. I was expecting CD to slow down (long story), but it doesn't seem to be meaningfully effected.

### DIFF
--- a/ykrt/src/compile/guard.rs
+++ b/ykrt/src/compile/guard.rs
@@ -2,33 +2,45 @@
 
 use crate::{
     compile::CompiledTrace,
-    mt::{HotThreshold, MT},
+    mt::{AtomicTraceCompilationErrorThreshold, HotThreshold, MT},
 };
 use parking_lot::Mutex;
-use std::sync::Arc;
+use std::sync::{atomic::Ordering, Arc};
 
 /// Responsible for tracking how often a guard in a `CompiledTrace` fails. A hotness counter is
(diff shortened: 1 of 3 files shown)


ltratt:
A lang_test failed with `free(): corrupted unsorted chunks`. I see similar errors intermittently on master and suspect they're the bug(s) that @ptersilie is looking at right now. I suggest we retry the merge; if that error reoccurs, then I will look in detail.

ltratt:
Force pushed a rebase against master.

PR #1753, opened 2025-05-17
Teach the reverse analyser about more instructions.

ltratt:
This is particularly effect for int_to_ptr and ptr_to_int since in the common case neither needs to generate any actual code, so passing through the register leads to noticeably fewer spills.


ltratt:
Fixed conditional in the force commit (I didn't check x64/mod.rs carefully enough at first!).

PR #1757, opened 2025-05-19
Split `YKD_LOG=4` into two levels.

ltratt:
I often want to know when tracing has started and stopped but, outside of tests, I almost never want to know when a trace has executed. Before this commit `YKD_LOG=4` bundled those two classes of things into one.

This commit splits them into two. `YKD_LOG=4` now captures start/stop tracing and `YKD_LOG=5` captures trace execution / deopt.

@vext01 Separately, in the past you were -- I think! -- adamant that we needed all these different `yk-` prefixes even though the RHS part of the message always differentiates things. I *think* we should just have a `yk_log:` prefix (or `ykd_log:`?) for consistency. If you agree, that would be a sensible follow-up PR IMHO.

### DIFF
--- a/tests/c/arg_mapping_callee.c
+++ b/tests/c/arg_mapping_callee.c
@@ -1,18 +1,18 @@
 // Run-time:
 //   env-var: YKD_LOG_IR=jit-pre-opt
 //   env-var: YKD_SERIALISE_COMPILATION=1
-//   env-var: YKD_LOG=4
+//   env-var: YKD_LOG=5
 //   stderr:
-//     yk-jit-event: start-tracing
+//     yk-tracing: start-tracing
 //     3:5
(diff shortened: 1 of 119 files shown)


vext01:
I agree with this, and have considered it myself too. The problem with entering jitted code is that, as we get better at jitting, it drowns out the other events more and more.

About the naming: that preference was at a time when I though we could parse the stderr output to draw a trace tree (using dot or similar), but I don't think that will be realistic any more, especially since the events required can't all be reliably logged (there was something to do with side traces).

So I'd say go ahead with your follow up.

ltratt:
Should that prefix be `yk-log:` or `ykd-log:`?

vext01:
The latter seems better to me.

PR #1760, opened 2025-05-20
Implement uitofp for 64-bit values on x64.

vext01:
Requires https://github.com/ykjit/ykllvm/pull/256

This is required for `heightmap` on the indirect call inlining branch.

This is the minimum viable implementation. Raising as a draft, because I'm not sure if we want to polish this now or later.

It could do with optimising for code size, since at the moment it emits duplicated magic constants each time we encounter a ui_to_fp bytecode.

For example, we could emit the constants once at the beginning or end of the trace, or can we maybe reference the memory of a rust array somehow?


ltratt:
This is fine for now, at least. Please undraft so we can merge.

vext01:
undrafted this and the ykllvm PR too.

ltratt:
Please force push a sync commit when appropriate.

vext01:
synced ykllvm

PR #1761, opened 2025-05-20
Teach the reverse analyser about LLVM intrinsics disguised as calls.

ltratt:
While I'm here use a more consistent "check the name" schemes in `x64/mod.rs`.


vext01:
Any speedup?

ltratt:
I can't tell with all the noise and the mapping overhead TBH. I can *see* better code being generated, and that's good enough.

PR #1766, opened 2025-05-22
Terminate guard traces at the first `Location` encountered

ltratt:
The main aim of this PR is best described in https://github.com/ykjit/yk/commit/8764a1074a28e0e6fed1266c3eea806920bc1b73: terminate guard traces at the first non-null-`Location` encountered. To get there, we have to do a lot of preparation, but there are various rewards along the way: overall, the code ends up cleaner. Notably, all of this work eventually lets us remove `HotLocation::SideTracing` entirely.

### DIFF
--- a/tests/lua/nested_loops.lua
+++ b/tests/lua/nested_loops.lua
@@ -4,34 +4,38 @@
 --   env-var: YKD_LOG_IR=debugstrs
 --   env-var: YKD_SERIALISE_COMPILATION=1
 --   stderr:
---     yk-tracing: start-tracing: nested_loops.lua:42: ADDI
---     yk-tracing: stop-tracing: nested_loops.lua:42: ADDI
---     --- Begin debugstrs: header: nested_loops.lua:42: ADDI ---
---       nested_loops.lua:41: FORLOOP
---       nested_loops.lua:42: ADDI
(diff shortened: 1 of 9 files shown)


ptersilie:
Can I suggest changing this paragraph in [8764a10](https://github.com/ykjit/yk/pull/1766/commits/8764a1074a28e0e6fed1266c3eea806920bc1b73)

```
This commit changes that: guard traces now always terminate at the first
non-null-`Location` they encounter. If that is a `Counting` `Location`,
then we immediately start tracing that `Location`. That means that we
put the guard trace in the job queue before we've even started tracing
the new `Location`! That leads to the seeming-surprise that the
`Counting` location is compiled first.
```

to

```
This commit changes that: guard traces now always terminate at the first
non-null-`Location` they encounter. If that is a `Counting` `Location`, then we
immediately start tracing that `Location`, and add it as a depency for the
guard trace. That means that even though we put the guard trace in the job
queue before we've even started tracing the new `Location`, the `Counting`
location is compiled first.
```

ltratt:
See if https://github.com/ykjit/yk/pull/1766/commits/19efd8db50395e5e48158fe8dad01cc173585871 works for you.

ptersilie:
Looks good.

PR #1769, opened 2025-05-23
Fix code generation with imm64s.

ltratt:
In most cases, x64 sign extends `imm64` even though it's represented as an `i32`. This commit fixes the mistakes that I made when I didn't account for this.

There is at least one further optimisation we could make to the code below but this *probably* still generates "optimal" code in the vast majority of cases we see, so the extra complexity is not yet obviously worth it.

@vext01 Please check this works with your branch before considering this one for merging.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -652,7 +652,7 @@ impl<'a> Assemble<'a> {
                     //   *%1 = %2
                     if let Some((b_iidx, Inst::BinOp(b_inst))) = next
                         && b_inst.binop() == BinOp::Add
-                        && self.op_to_zero_ext_i32(&b_inst.rhs(self.m)).is_some()
+                        && self.op_to_sign_ext_i32(&b_inst.rhs(self.m)).is_some()
                         && let Some((s_iidx, Inst::Store(s_inst))) = iter.peek()
                     {
                         // Now we have to check that -- taking into account ptr offsets -- the two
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -4647,23 +4719,34 @@ mod tests {
...
+                ; %4: i8 = and %1, 3i8
                 and r.32.b, 0x03
-                ; %4: i1 = eq %3, 4i8
+                ; %5: i1 = eq %4, 4i8
                 cmp r.32.b, 0x04
                 setz ...
+                ; %6: i64 = and %2, 2147483647i64
+                mov r.64.x, r.64.y
+                and r.64.y, 0x7fffffff
+                ; %7: i64 = and %2, 2147483648i64
```
vext01, line 4747: I think this is what I wanted to see. The moment the top bit is set, we no longer use `and r64, imm32` because the sign extend would give incorrect semantics.

Looks like you checked one value either side of the threshold. Did I understand correctly?

ltratt: > Did I understand correctly?

Yep.

vext01:
I can confirm that this fixes the test case that was crashing on my branch!

If I were to nitpick, I'd change this (in the commit message and PR descr):

> In most cases, x64 sign extends imm64 even though it's represented as an i32.

To this, which I think is more precise:

> In most x64 instructions where the operands are `r/m64, imm32`, the immediate operand is implicitly sign extended up to an `imm64`.

Up to you if you want to fix it though.

ltratt:
I think that one falls below our "message worth fixing" threshold.

PR #1773, opened 2025-05-25
Guard tidyups

ltratt:
This PR somewhat tidies up guards (what use to be called `DeoptInfo`). After first converting a `HashMap` to a `Vec`, https://github.com/ykjit/yk/commit/947c664717bdc27fc32b58c416f6b59ac3fda82e is really the main part of this PR: it recognises that we need to store different info while we're compiling a trace than after we've compiled a trace. Overall, there's only a mild functional change here, but this PR is a stepping stone towards making these parts of the system easier to work on.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
@@ -52,7 +52,7 @@ pub(crate) extern "C" fn __yk_deopt(
         .unwrap();
     let gidx = GuardIdx::from(usize::try_from(gidx).unwrap());
     let aot_smaps = AOT_STACKMAPS.as_ref().unwrap();
-    let info = &ctr.deoptinfo[&usize::from(gidx)];
+    let info = &ctr.deoptinfo[usize::from(gidx)];
     let mt = Arc::clone(&ctr.mt);
 
     mt.deopt();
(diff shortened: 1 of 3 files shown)


vext01:
LGTM, but since deopt machinery is @ptersilie's baby, I think he should have a look too.

ptersilie:
About [968ce1a](https://github.com/ykjit/yk/pull/1773/commits/968ce1a86868014ea060740fa766b6c01d6ffce7): This used to be a `Vec`. The reason we changed it to a `HashMap` was that the indexes got skewed when the optimiser removed guards (see https://github.com/ykjit/yk/commit/050d6bd748cfb69fe51f3e1f2fbbb7b3986d659b). We should make doubly sure this isn't the case anymore.

ltratt:
> The reason we changed it to a HashMap was that the indexes got skewed when the optimiser removed guards

Yes -- but by the time codegen sees this, the number of guards is stable (i.e. doesn't change) so a `Vec` is fine.

ptersilie:
That does make sense, though I'm not sure why this was a problem back then.

ptersilie:
Okay, I see why this was a problem: we use the `GuardIdx` stored inside the `GuardInst` to lookup the info. So when we optimise away one of the guards, we don't update the index of the following guards. If we then use that index to lookup stuff in a `vec` it won't always work (or at least it didn't in the past). Example:

```
GuardInst 0
GuardInst 1 // optimise out
GuardInst 2
```

Now `deoptinfo` has two elements. But when we try to lookup the deoptinfo for `GuardInst 2` we would get an index error.

ltratt:
Yes, we now "recast" guard indexes in the codegen so that they're contiguous (because, from the codegen's POV, they really are).

ptersilie:
Ok, I had to re-read some code that has changed since the last time I looked at it. It looks like we no longer pass through the `GuardIdx` directly from the `GuardInst`, so I agree this is fine now. I just wanted to double check.

PR #1774, opened 2025-05-29
Don't spill at a guard

ltratt:
The main commit here is https://github.com/ykjit/yk/commit/7d02382948677096333430df6d56848c194c24bf. The performance effect of this PR is probably minor. when I measured it, there was enough noise that I couldn't detect a clear pattern: knowing the current quirks of our current register allocator, it'll probably make some things a bit faster and a smaller number a bit slower.


ltratt:
This should (with luck) merge when #1776 goes through. [Edd is right: we should institute a "sync yklua" process just as with "sync ykllvm". I was testing new yklua, where there are no problems, vs. a fairly ancient ykua in buildbot.sh and our tests.]

PR #1779, opened 2025-06-04
Remove incorrect "trace too long" check.

ltratt:
We inherited this general idea from RPython, but the check is clearly wrong. The limit in `mt.rs` was (vaguely) intended to reference "how many JIT IR instructions are there?". However, what we actually check is "how many hwt/swt numbers of a random metric do we get". In other words, we're not checking what we thought, and we're rejecting traces for reasons we don't understand.

Right now, since we don't really know what "too long" looks like in our setting, the best thing to do is to remove this check entirely, and come back to it later. Overall, doing so is a wash: some benchmarks get faster, some slower. Those most affected are:

```
Json/YkLua/100            4.55% faster
List/YkLua/1500           3.36% faster
LuLPeg/YkLua/default      1.64% faster
CD/YkLua/250              1.19% faster

DeltaBlue/YkLua/12000     1.68% slower
NBody/YkLua/250000        1.95% slower
Havlak/YkLua/1500         5.37% slower
```


ltratt:
This failed because of a network error. Retry?

ltratt:
Replaced an incorrect, existing `unwrap` with `?` in the force push. Should hopefully fix the panic we saw.

ltratt:
@vext01 If you are OK with me doing so, I think we should remove the "trace too long" tests. They were always very fragile and now they're mostly wrong (of course on *my* account they nondeterministically pass...).

PR #1783, opened 2025-06-05
Move instructions into guard bodies.

ltratt:
We frequently encounter values that are only used when a guard fails. Here's a simple example:

```
%6: i32 = ...
%7: i8 = ...
%8: i1 = ...
guard true, %8, [%7]
jmp beginning_of_trace [%6]
```

In other words, `%7` is not used by the trace in the "happy" case: we only calculate its value in case the unhappy (guard failing!) case occurs.

This commit exploits what I'm now calling "guard bodies" (a nameless version of which was introduced in
7d02382948677096333430df6d56848c194c24bf). During code generation, we identify "guard failing only" side-effect-free values such as `%7` and push them into the guard body. So the above is compiled to something along the lines of:

```
%6: i32 = ...
%8: i1 = ...
guard true, %8, [%7]
jmp beginning_of_trace [%6]
; deopt for guard
%7: i8 = ...
```

In other words, we only pay the cost of evaluating `%7` when we really need its values.

This commit has the "obvious" virtue of not evaluating `%7` in the happy case: probably (but it's hard to measure the effect of this directly) the more significant factor is that it decreases register pressure. When I manually diff code, I can see noticeably fewer spills as a result of this commit.

This commit is horrible in many ways:

  1. It only identifies values that are immediately unused: it doesn't recursively check for this property. That's a missed opportunity.

  2. Inside guard bodies, the register allocator does a *terrible* job, often putting subsequent values in the same register and immediately spilling them. This is because it assumes there are lots of intermediate instructions, which there aren't.

  3. It brutalises several parts of the code, chiefly `x64/mod.rs`. It makes it significantly harder to understand and much more fragile. It's like an invading army that strips everything bare, pulls down buildings, and does unspeakable acts in barns. It's hard to overstate this: it really does put this part of the code close to breaking point, because we're stretching it in directions we never imagined it could or should go.

Still, it seems its worth it. Here's rough benchmarking numbers from b16:

```
BigLoop/YkLua/1000000000  29.55% faster
Towers/YkLua/600          16.15% faster
Heightmap/YkLua/2000      14.95% faster
Permute/YkLua/1000        13.23% faster
Sieve/YkLua/3000          12.14% faster
LuLPeg/YkLua/default      10.45% faster
Queens/YkLua/1000         9.28% faster
Richards/YkLua/100        5.61% faster
Bounce/YkLua/1500         4.73% faster
NBody/YkLua/250000        4.16% faster
HashIds/YkLua/6000        2.83% faster
Json/YkLua/100            2.52% faster
CD/YkLua/250              1.99% faster
Havlak/YkLua/1500         1.97% faster
Storage/YkLua/1000        1.95% faster
List/YkLua/1500           1.93% faster
Mandelbrot/YkLua/500      0.22% slower
DeltaBlue/YkLua/12000     3.39% slower
```

I don't really understand the deltablue slowdown: it looks like it might just be that we now compile some traces at a slightly different time and get a little unluckier, but I'm not entirely sure. Still, since everything else gets faster (the Mandelbrot figure should really be 0% -- it runs too fast to tell!), I think this is a clear win.


ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
```diff
@@ -771,6 +854,8 @@ impl<'a> Assemble<'a> {
                 inlined_frames: gd.inlined_frames,
                 guard: Guard::new(),
             });
+
+            stack_sizes.push(self.ra.stack_size());
```
ptersilie, line 858: It probably doesn't make a difference, but if we didn't want to use a vec here we could just move the `max_stack_size` calculation in here, so do it after every guard, and then just return an integer here.

ltratt: Fixed in 4efbf3d9.

ptersilie:
Neat. Please squash.

PR #1803, opened 2025-07-01
Adjust promotion assertion in TraceBuilder.

Pavel-Durov:
This might be completely wrong :)  
Its a behavior I observed while working on the multi-version SWT which occurred only in `release` mode.

My assumption is that sidetraces are created when guards fail, meaning they follow alternative execution paths than the original trace. These alternative paths may not encounter all the same `yk_promote()` calls.

Tested this change on latest` yklua` in `debug` and `release` modes.

### DIFF
--- a/ykrt/src/compile/jitc_yk/trace_builder.rs
+++ b/ykrt/src/compile/jitc_yk/trace_builder.rs
@@ -1613,8 +1613,25 @@ impl TraceBuilder {
                 }
             }
         }
-
-        assert_eq!(self.promote_idx, self.promotions.len());
+        match self.jit_mod.tracekind() {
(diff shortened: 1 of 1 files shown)


ltratt:
The promotion data is recorded per-trace, so each trace necessarily contains the "correct" amount of promotion data. The relationship of a "loop" vs. a "guard" trace isn't relevant in that regard. So I think the `assert` here is incorrect: a trace (loop or guard) can't consume more promotion data than it has.

*However* outlined functions make this hard to see: it means that a trace has recorded promotion data that it needs to skip (or throw away, depending on how you look at it). I wonder if that's what you're seeing, indirectly?

PR #1806, opened 2025-07-03
Update Rust edition.

ptersilie:
This updates the rust edition to 2024 so we can continue using let-chains without getting errors. Unfortunately, this brings with it a bunch of other changes:

- Slightly different formatting of imports
- Some other formatting changes
- Change `#[no_mangle]` to `#[unsafe(no_mangle)]`
- Change `extern` to `unsafe extern`
- Changes in the match ergonomics (see https://doc.rust-lang.org/nightly/edition-guide/rust-2024/match-ergonomics.html)

### DIFF
--- a/hwtracer/Cargo.toml
+++ b/hwtracer/Cargo.toml
@@ -2,7 +2,7 @@
 name = "hwtracer"
 version = "0.1.0"
 authors = ["The Yk Developers"]
-edition = "2021"
+edition = "2024"
 license = "Apache-2.0 OR MIT"
 
 [dependencies]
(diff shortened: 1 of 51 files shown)


ptersilie:
I've gone through the env vars and I think they are okay as is. I've added some comments. Thoughts?

ptersilie:
Squashed.

ptersilie:
Force pushed a missing `no_mangle` in the software tracer.

ptersilie:
Force pushed some missed `set_var`s in the yklua test suite. Sorry. Hopefully that was it now.

PR #1807, opened 2025-07-04
Disable Pavel's SWT module cloning work for now (and update tests).

vext01:
Set YKB_SWTMODCLONE=1 to turn on Pavel's work.

Many tests can be unignored now module cloning is disabled by default. Three tests have new issues, and have to be disabled for now.

(Heads up @Pavel-Durov)


vext01:
Good idea! I hadn't thought of that.

vext01:
Force pushed what I believe to be a fix.

(note the two force pushes -- sorry about that, I realised that the comment was a bit ambiguous and revised it)

ltratt:
Multiple force pushes are fine! It's only if there's a rebase that I get upset.

PR #1808, opened 2025-07-04
Support traces that end in a different frame than they started in.

ptersilie:
We currently quite often abort traces where the trace stopped tracing in a different frame than it started in. This happens when the interpreter calls itself recursively.

There's two cases for this:

1) Stop tracing after entering a recursive interpreter call. To compile
   such traces we outline anything following the recursive call, emit a
   call instruction, and deopt immediately upon returning from that
   call.

2) Stop tracing after returning from a recursive interpreter call. To
   compile these traces, we ignore any blocks following the return and
   emit a return instruction at the end of the trace. After restoring
   callee-saved registers and informing the meta-tracer that we exited
   the compiled trace, this naturally returns back into the interpeter.

There's some more things we could have done here, but there appear to be many corner cases and trying to implement these quickly exceeded the scope of this change:

1) By default recursive interpreter calls are outlined. In some traces
   we can get multiple recursive interpreter calls, which happens when
   the trace doesn't hit a non-null location within the recursion.
   Since, by definition, this means this call doesn't contain loops and
   is likely fairly small, it could be beneficial to inline these calls.

2) In order to detect a recursive interpreter call in trace-builder we
   need to be able to see the actual call in the IR. It can happen,
   however, that the call is hidden within an unmappable or outlined
   block. Not only does this make them difficult to detect, it also
   means these calls have no safepoints/stackmaps attached to them,
   which are required to emit the deopt at the end of the trace.

In terms of performance I couldn't measure any significant improvements or degradations and we still abort many traces. Hopefully, once we solve some of the other aborted traces, we will get some return of investment on this.

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
@@ -42,6 +42,20 @@ const RECOVER_REG: [usize; 31] = [
 /// from zero). This is used to allocate arrays whose indices need to be the DWARF register number.
 const REGISTER_NUM: usize = RECOVER_REG.len() + 2;
 
+#[unsafe(no_mangle)]
+pub(crate) extern "C" fn __yk_ret_from_trace(ctrid: u64) {
+    let ctr = MTThread::with_borrow(|mtt| mtt.compiled_trace(TraceId::from_u64(ctrid)))
+        .as_any()
+        .downcast::<X64CompiledTrace>()
+        .unwrap();
+    let mt = Arc::clone(&ctr.mt);
+    mt.deopt();
+    mt.stats
+        .timing_state(crate::log::stats::TimingState::OutsideYk);
+    mt.log
+        .log(Verbosity::Execution, &format!("return {:?}", ctr.ctrid()));
+}
+
 /// Deoptimise back to the interpreter. This function is called from a failing guard (see
 /// [super::Assemble::codegen]).
 ///
--- a/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
+++ b/ykrt/src/compile/jitc_yk/jit_ir/mod.rs
@@ -117,6 +117,17 @@ use ykaddr::addr::symbol_to_ptr;
 // This is simple and can be shared across both IRs.
 pub(crate) use super::aot_ir::{BinOp, FloatPredicate, FloatTy, Predicate};
 
+/// Did this trace end in a different frame than it started in?
+#[derive(Debug)]
+pub(crate) enum TraceEndFrame {
+    /// The trace stopped in the same frame that it started in.
+    Same,
+    /// The trace stopped after entering a new frame.
+    Entered,
+    /// The trace stopped after leaving the starting frame.
+    Left,
+}
+
 /// What kind of trace does this module represent?
 #[derive(Clone)]
 pub(crate) enum TraceKind {
@@ -129,6 +140,8 @@ pub(crate) enum TraceKind {
     /// A sidetrace: this will start at the point of a guard and will jump to
     /// [SideTraceInfo::target_ctr].
     Sidetrace(Arc<YkSideTraceInfo<Register>>),
+    /// A trace where we stopped tracing in a different frame than we started in.
+    DifferentFrames,
 }
 
 impl std::fmt::Debug for TraceKind {
@@ -138,6 +151,7 @@ impl std::fmt::Debug for TraceKind {
             TraceKind::HeaderAndBody => write!(f, "HeaderAndBody"),
             TraceKind::Connector(_) => write!(f, "Connector"),
             TraceKind::Sidetrace(_) => write!(f, "Sidetrace"),
+            TraceKind::DifferentFrames => write!(f, "DifferentFrames"),
         }
     }
 }
@@ -244,6 +258,7 @@ impl Module {
     pub(crate) fn set_tracekind(&mut self, tracekind: TraceKind) {
         match (&self.tracekind, &tracekind) {
             (&TraceKind::HeaderOnly, &TraceKind::HeaderAndBody) => (),
+            (&TraceKind::HeaderOnly, &TraceKind::DifferentFrames) => (),
             (from, to) => panic!("Can't transition from a {from:?} trace to a {to:?} trace"),
         }
         self.tracekind = tracekind;
@@ -643,6 +658,10 @@ impl Module {
         info: GuardInfo,
     ) -> Result<GuardInfoIdx, CompilationError> {
         GuardInfoIdx::try_from(self.guard_info.len()).inspect(|_| self.guard_info.push(info))
+    }
+
+    pub(crate) fn guard_info(&self, gidx: GuardInfoIdx) -> &GuardInfo {
+        &self.guard_info[usize::from(gidx)]
     }
 
     pub(crate) fn trace_header_start(&self) -> &[PackedOperand] {
@@ -1555,6 +1574,8 @@ pub(crate) enum Inst {
     TraceBodyStart,
     TraceBodyEnd,
     SidetraceEnd,
+    Deopt(GuardInfoIdx),
+    Return(u64),
 
     SExt(SExtInst),
     ZExt(ZExtInst),
@@ -1621,6 +1642,8 @@ impl Inst {
             Self::TraceBodyStart => m.void_tyidx(),
             Self::TraceBodyEnd => m.void_tyidx(),
             Self::SidetraceEnd => m.void_tyidx(),
+            Self::Deopt(_) => m.void_tyidx(),
+            Self::Return(_) => m.void_tyidx(),
             Self::SExt(si) => si.dest_tyidx(),
             Self::ZExt(si) => si.dest_tyidx(),
             Self::BitCast(i) => i.dest_tyidx(),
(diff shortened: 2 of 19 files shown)


ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -19,6 +21,14 @@ use crate::{
 use std::{collections::HashMap, ffi::CString, sync::Arc};
 use ykaddr::addr::symbol_to_ptr;
 
+/// The mode for processing blocks.
+enum ProcessMode {
+    /// Process blocks normally.
+    Normal,
+    /// Don't process any remaining blocks (except in order to count promotions/debugstrs etc).
+    Skip,
```
vext01, line 29: Is this to do with outlining? We already have a mechanism for skipping blocks during outlining, and I wonder if we are duplicating that here?

vext01:
LGTM. Just once query.

ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
```diff
@@ -42,6 +42,20 @@ const RECOVER_REG: [usize; 31] = [
...
+pub(crate) extern "C" fn __yk_ret_from_trace(ctrid: u64) {
+    let ctr = MTThread::with_borrow(|mtt| mtt.compiled_trace(TraceId::from_u64(ctrid)))
+        .as_any()
+        .downcast::<X64CompiledTrace>()
+        .unwrap();
+    let mt = Arc::clone(&ctr.mt);
+    mt.deopt();
+    mt.stats
+        .timing_state(crate::log::stats::TimingState::OutsideYk);
+    mt.log
```
ltratt, line 55: This reminds me: we really need to fix logging, because at the moment we often generate trsing with `format` then `log` says "no logging needed" and throws them away. That must cost us.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
```diff
@@ -3187,6 +3271,21 @@ impl GuardInst {
     }
 }
 
+#[derive(Debug)]
+pub(crate) enum HasGuardInfo {
```
ltratt, line 3275: This needs a quick doc string (and, ideally, a different name, but that's more subjective).

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -340,7 +360,9 @@ impl TraceBuilder {
                     let nextinst = blk.insts.last().unwrap();
                     self.handle_idempotent_promote(bid, iidx, val, nextinst)
                 }
-                _ => todo!("{:?}", inst),
+                _ => Err(CompilationError::General(format!(
```
ltratt, line 363: I think this should still be `todo`.

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -688,10 +716,25 @@ impl TraceBuilder {
...
-            Err(CompilationError::General(
-                "returned from function that started tracing".into(),
-            ))
+            if let TraceKind::Sidetrace(_) = self.jit_mod.tracekind() {
+                // Even though we currently catch side-traces that leave the main interpreter loop
+                // in `mt.rs`, it can happen that a trace re-enters the main interpreter loop after
+                // leaving it briefly without encountering a non-null location. As we only check
+                // the frame addresses at non-null locations, this special case slips through and
+                // we have to handle it here.
+                // FIXME: Abort side-traces the same way we do normal traces by emitting a return
```
ltratt, line 725: What does this FIXME mean? Why would we abort side-traces with a `return`?

ykrt/src/compile/jitc_yk/trace_builder.rs
ptersilie: We discussed this but you may have forgotten. If I make this a `todo` some benchmarks won't run anymore because we are now tracing more code. There's two cases that come up here:
1. `align 1`
2. `no return`
I'm not sure yet what the solution for these are, but I don't think it should go in this PR. But I should look at that sooner rather than later.

ykrt/src/compile/jitc_yk/trace_builder.rs
ptersilie: Sorry, that shouldn't say "abort". I meant that we should do the same that we now do for normal traces that return from the interpreter: emit a return. Will fix the comment.

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ptersilie: Looking at this I realise that other things also need renaming now. For example the `guard_to_deopt` function in the codegen should probably we called something like `push_deopt` and `self.guards` should become `self.deopts`. It might be better to do that in a quick follow up PR once this goes in. As for renaming `HasGuardInfo` my mind is pulling a blank at the moment. The best I could come up with is `DeoptProvider`.

To be honest this would have been better as a trait, but I couldn't get it through the borrow checker, so an enum was much simpler.

ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs
```diff
@@ -42,6 +42,20 @@ const RECOVER_REG: [usize; 31] = [
 /// from zero). This is used to allocate arrays whose indices need to be the DWARF register number.
 const REGISTER_NUM: usize = RECOVER_REG.len() + 2;
 
+#[unsafe(no_mangle)]
+pub(crate) extern "C" fn __yk_ret_from_trace(ctrid: u64) {
+    let ctr = MTThread::with_borrow(|mtt| mtt.compiled_trace(TraceId::from_u64(ctrid)))
+        .as_any()
+        .downcast::<X64CompiledTrace>()
+        .unwrap();
+    let mt = Arc::clone(&ctr.mt);
```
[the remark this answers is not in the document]
ptersilie: [3c1c4b4](https://github.com/ykjit/yk/pull/1808/commits/3c1c4b4b3510f09e1d2e0d8805ddc3a95b79b334)

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -19,6 +21,14 @@ use crate::{
 use std::{collections::HashMap, ffi::CString, sync::Arc};
 use ykaddr::addr::symbol_to_ptr;
 
+/// The mode for processing blocks.
+enum ProcessMode {
```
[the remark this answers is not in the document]
ptersilie: [4ac9d8a](https://github.com/ykjit/yk/pull/1808/commits/4ac9d8a437cb5e0c80afc607424208f5c73a6f70)

ykrt/src/compile/jitc_yk/trace_builder.rs
ptersilie: [7b198c7](https://github.com/ykjit/yk/pull/1808/commits/7b198c741db495bdec93a4833eca49cb6e935a76)

ykrt/src/mt.rs
```diff
@@ -636,8 +644,14 @@ impl MT {
                 } => {
                     // If this assert fails then the code in `transition_control_point`,
                     // which rejects traces that end in another frame, didn't work.
-                    assert_eq!(frameaddr, tracing_frameaddr);
-                    (hl, thread_tracer, promotions, debug_strs)
+                    let endframe = if frameaddr < tracing_frameaddr {
```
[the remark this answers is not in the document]
ptersilie: [9e36b63](https://github.com/ykjit/yk/pull/1808/commits/9e36b63861b81b7ceb3c28f7ce034b7bb1e6b9bb)

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -57,7 +57,9 @@ pub(crate) struct TraceBuilder {
     inferred_consts: HashMap<jit_ir::InstIdx, jit_ir::ConstIdx>,
     /// Did this trace end in another frame?
     endframe: TraceEndFrame,
-    /// Whether or not we want blocks to be processed.
+    /// Whether or not we want blocks to be processed. Set to `true` by default and changed to
```
ltratt, line 60: Dumb question: why do we need to keep processing blocks at this point? Can't we just immediately stop?

ptersilie: We still have to process blocks to count the promotions, otherwise we could stop processing immediately. We could disable that check, but then we might miss it when things with promotions go wrong.

ptersilie: Your wish is my command!

ptersilie: [4b31ae4](https://github.com/ykjit/yk/pull/1808/commits/4b31ae413bca2a0494ab837ee568247b59f525ea)

ykrt/src/compile/jitc_yk/jit_ir/mod.rs
ltratt: I'm pretty sure this commit makes things worse I'm afraid :/ Most of these really should be "guard", not "deopt", IMHO.

ltratt: In retrospect, I suggest we revert https://github.com/ykjit/yk/pull/1808/commits/7c56427d256f09ff710713b6577fce3bcf59aab9. We can merge this PR without it IMHO.

ltratt:
I think we can revert https://github.com/ykjit/yk/pull/1808/commits/7c56427d256f09ff710713b6577fce3bcf59aab9 and squash in one go, if you agree.

ptersilie:
Agreed. I took the liberty to at least rename that one `ginst`. If ok I will squash again.

ltratt:
Please squash.

ltratt:
Got it. Please squash.

PR #1813, opened 2025-07-15
Reduce the control point's overhead.

ptersilie:
Every time we call the control point (which is on each interpreter loop iteration) we currently need to push and pop a lot of registers. This is quite expensive, especially when the location is null and we return from the control point call immediately.

The reason we do this because we execute compiled traces in the interpreter frame and the compiled trace expects live values to live in certain registers as specified by the stackmap accompanying the control point. This makes it necessary to reset the interpreter (i.e. live values in registers) to the state just before the control point was called.

To avoid having to do these pushes and pops, we need to work around two register types:

1) Callee-saved registers. Currently, when executing a trace we jump
   into the trace from within the control point, immediately after
   hard-resetting the stack. This means callee-saved registers are not
   restored by the control points epilogue. We can fix this, but
   overwriting the control points return address with the address of the
   trace we want to execute. As a bonus this solves long-standing worry
   we had about not properly returning from Rust functions which may
   have repercussions due to drop methods not being executed.

2) Caller-saved registers. Even though we are now returning normally from the control
   point, this does not help for caller-saved registers which are
   restored upon returning to the interpreter. This does not happen when
   we instead jump to a trace. However, we know that any live value that
   is stored in a caller-saved register needs to have been spilled. We
   can use this to replace all caller-saved registers in the trace's
   input params with their spill location, thus no longer depending on
   these registers having the correct value.

Performance-wise this gives us a juice 7% on Richards, Bounce, and LuLPeg, and 5% on Queens. There's some 4% improvements but they may or may not be noise.

On the other end we get a 5% slowdown on Bigloop. I could not find a satisfying explanation for this. We seem to be spending an extra 0.5s running traces (7.7s vs 8.2s), though apart from an extra `nop` in the loop body I can't see a good reason for the slowdown (maybe alignment?). There's also some extra moves and `nop`s in the header and deopt routines, but those shouldn't have an observable effect as they are only executed once.

### DIFF
--- a/ykcapi/src/lib.rs
+++ b/ykcapi/src/lib.rs
@@ -67,51 +67,14 @@ pub extern "C" fn __ykrt_control_point(
     // Stackmap id for the control point.
     smid: u64,
 ) {
-    // FIXME: We can possibly avoid the below (and this entire function) by patching the return
-    // address of the control point on the stack to point to the compiled trace. This means we
-    // still run the epilogue of the control point function call, which automatically restores the
-    // callee-saved registers for us (so we don't have to do it here).
+    // FIXME: We could get rid of this entire function if we pass the frame's base pointer into the
+    // control point from the interpreter.
     std::arch::naked_asm!(
-        // Push all registers to the stack as these may contain trace inputs (live
-        // variables) referenced by the control point's stackmap.
-        //
-        // We don't need to push and restore `rdx` since `smid` can never be a live value and
-        // thus won't be tracked by the stackmap.
-        //
-        // FIXME: In the future we want the control point to return naturally into the compiled
-        // trace (at the moment we just rip out the control point's stack), which means we then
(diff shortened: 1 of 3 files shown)


ptersilie:
We also discussed a different approach to avoid the push/pops in the control
point: not calling the control point in the first place if the location is
null. As a quick hack I did the following in yklua:
```c
    if (ykloc->state != 0) {
      yk_mt_control_point(G(L)->yk_mt, ykloc);
    }
```
which lead to a 15% slowdown on Richards! Looking at the `jit-asm` there's a
big chunk of extra stuff in the trace (see below image). I would have expected
a few extra instructions (e.g. a `mov`, `cmp`, and `jmp`) but not this much. I
didn't dig any further. But this is potentially something we could revisit when
Pavel's multi-interpreter work has gone in as we can then remove this check
from the tracing version.

image

ltratt:
I am a man of few words.

PR #1821, opened 2025-07-20
SWT multi module dynamic transition

Pavel-Durov:
Dynamic transition between unopt and opt control points based on the JIT state.

- Added a new configuration flag for `SWT MODCLONE` to handle multi-version IR execution.
- Added regex pattern in lang_tester to support `__yk_unopt_ prefix`.
- Implemented asm code gen for transitions between `optimised` and `unoptimised` code.

Example of transitionfor `bf.O0.c` in `opt` to `unopt` transition and trace execution.

Live variables:
| # | Source | Destination |
|---|--------|-------------|
| 1 | Register(8, 8, [-48]) | Register(8, 8, [-48]) |
| 2 | Register(8, 8, [-48]) | Register(8, 8, [-48]) |
| 3 | Register(0, 8, [-72]) | Register(0, 8, [-72]) |
| 4 | Indirect(6, -64, 8) | Indirect(6, -64, 8) |
| 5 | Indirect(6, -56, 8) | Indirect(6, -56, 8) |
| 6 | Register(12, 8, [-80]) | Register(12, 8, [-80]) |
| 7 | Register(3, 8, [-112]) | Register(3, 8, [-112]) |
| 8 | Register(14, 8, [-120]) | Register(14, 8, [-120]) |
| 9 | Indirect(6, -104, 8) | Indirect(6, -104, 8) |
| 10 | Register(15, 8, [-96]) | Register(15, 8, [-96]) |
| 11 | Register(13, 8, [-88]) | Register(13, 8, [-88]) |

ASM:
```asm
   movabs rbp, 0x7fffa3ddf340
   movabs rsp, 0x7fffa3ddf340
   sub rsp, 0xb0
   movabs rax, 0x323464e0
   mov rcx, qword ptr [rbp - 0x40]
   mov qword ptr [rax + riz], rcx
   mov rcx, qword ptr [rbp - 0x38]
   mov qword ptr [rax + riz + 8], rcx
   mov rcx, qword ptr [rbp - 0x68]
   mov qword ptr [rax + riz + 0x10], rcx
   mov rax, qword ptr [rbp - 0xe8]
   mov qword ptr [rbp - 0x30], rax
   mov r8, qword ptr [rbp - 0xe8]
   mov rax, qword ptr [rbp - 0xe8]
   mov qword ptr [rbp - 0x30], rax
   mov r8, qword ptr [rbp - 0xe8]
   movabs rax, 0x323464e0
   mov rcx, qword ptr [rax + riz]
   mov qword ptr [rbp - 0x40], rcx
   movabs rax, 0x323464e0
   mov rcx, qword ptr [rax + riz + 8]
   mov qword ptr [rbp - 0x38], rcx
   mov rax, qword ptr [rbp - 0x108]
   mov qword ptr [rbp - 0x50], rax
   mov r12, qword ptr [rbp - 0x108]
   mov rax, qword ptr [rbp - 0xd0]
   mov qword ptr [rbp - 0x70], rax
   mov rbx, qword ptr [rbp - 0xd0]
   mov rax, qword ptr [rbp - 0x118]
   mov qword ptr [rbp - 0x78], rax
   mov r14, qword ptr [rbp - 0x118]
   movabs rax, 0x323464e0
   mov rcx, qword ptr [rax + riz + 0x10]
   mov qword ptr [rbp - 0x68], rcx
   mov rax, qword ptr [rbp - 0x120]
   mov qword ptr [rbp - 0x60], rax
   mov r15, qword ptr [rbp - 0x120]
   mov rax, qword ptr [rbp - 0x110]
   mov qword ptr [rbp - 0x58], rax
   mov r13, qword ptr [rbp - 0x110]
   mov rax, qword ptr [rbp - 0xc0]
   mov qword ptr [rbp - 0x48], rax
   mov rax, qword ptr [rbp - 0xc0]
   mov rcx, qword ptr [rbp - 0xc8]
   mov rdi, qword ptr [rbp - 0xd8]
   mov rsi, qword ptr [rbp - 0xe0]
   mov r9, qword ptr [rbp - 0xf0]
   mov r10, qword ptr [rbp - 0xf8]
   mov r11, qword ptr [rbp - 0x100]
   movabs rdx, 0x7fb6ebe62000
   jmp rdx

```

Followup from ykllvm PR :point_right: https://github.com/ykjit/ykllvm/pull/267


tests/c/many_threads_many_locs.c
```diff
@@ -1,3 +1,5 @@
+// ## FIXME: SWT can't handle control points not in main.
+// ignore-if: test "$YKB_TRACER" = "swt"
```
ptersilie, line 2: Same here. If we can't use SWT outside of main then yklua wouldn't work.

ykcapi/src/lib.rs
```diff
@@ -67,18 +67,59 @@ pub extern "C" fn __ykrt_control_point(
...
+            "sub rsp, 8",   // Alignment
+            "mov rcx, rbp", // Pass interpreter frame's base pointer via 4th argument register.
+            "call __ykrt_control_point_real",
+            "add rsp, 8",
+            "ret",
+        );
+    }
+    #[cfg(swt_modclone)]
+    {
+        // FIXME: Adapt multi version swt control point transition.
```
ptersilie, line 85: I remember you asking if we can use the same trick here as in https://github.com/ykjit/yk/pull/1813. I changed my mind about it. We might not be able to get rid of the callee-saved registers here, but we should be able to remove the caller-saved registers by reading them from the extra locations instead. We don't need to do this here, but it might be worth investigating in a follow up PR. Maybe the callee-saved registers can also be removed by returning into the transitioning routine, as we do with traces we want to execute, but I'm not sure yet if/how that would work.

ykrt/src/mt.rs
```diff
@@ -431,7 +438,7 @@ impl MT {
     #[allow(clippy::not_unsafe_ptr_arg_deref)]
     pub fn control_point(self: &Arc<Self>, loc: &Location, frameaddr: *mut c_void, smid: u64) {
         match self.transition_control_point(loc, frameaddr) {
-            TransitionControlPoint::NoAction => (),
+            TransitionControlPoint::NoAction => {}
```
ptersilie, line 441: ?

ykrt/src/trace/swt/asm.rs
```diff
@@ -0,0 +1,111 @@
+use capstone::prelude::*;
```
ptersilie, line 1: Is everything in this file only used for testing?

ykrt/src/trace/swt/cp.rs
```diff
@@ -0,0 +1,623 @@
...
+/// control point to the unoptimised control point. It copies the live variables and
+/// restores the registers.
+///
+/// # Arguments
+///
+/// * `frameaddr` - Pointer to the current frame address
+/// * `stats` - Statistics tracker for recording the transition
+pub(crate) unsafe fn cp_transition_to_unopt(frameaddr: *const c_void, stats: &Stats) {
+    // TODO: add cache for the asm code generation
+    let buffer = generate_transition_asm(CPTransition {
```
ptersilie, line 121: What is the reason for using dynasm here instead of writing this in Rust or inline asm, e.g. like we do in `deopt.rs`?

ykrt/src/trace/swt/cp.rs
```diff
@@ -0,0 +1,623 @@
...
+fn restore_register(asm: &mut Assembler, dwarf_reg_num: u16, rbp_offset_reg_store: i32) {
+    let reg_offset = REG_OFFSETS.get(&dwarf_reg_num).unwrap();
+    let reg_val_rbp_offset = i32::try_from(rbp_offset_reg_store - reg_offset).unwrap();
+    let dynasm_reg = dwarf_to_dynasm_reg(dwarf_reg_num.try_into().unwrap());
+    dynasm!(asm
+        ; mov Rq(dynasm_reg), QWORD [rbp - reg_val_rbp_offset]
+    );
+}
+
+// Calculates the offset of the call instruction after the control point.
```
ptersilie, line 352: What call instruction after the control point are we looking for here?

Pavel-Durov:
> There's quite a lot to review here. I must admit I struggle a bit understanding what's going on in `cp.rs` and `live_vars.rs`. Could you explain some more what we are doing here and why it's being done this way? In particular why couldn't we use the same approach we use in `deopt.rs`, i.e. write the new stack to some location in memory and then copy it over in one go?

### `cp.rs` - Transition Orchestration 
- Generates assembly code for each transition type:
     - Opt → UnOpt (when starting tracing)
     - UnOpt → Opt (when stopping tracing)
     - Opt → UnOpt + Execute Trace (transition and jump to compiled
       trace)
- Stack Frame Adjustment 
- Register State Management
   - Maintains mapping of saved registers (`REG_OFFSETS`)
   - Restores registers

### `live_vars.rs` - Live variables and additional locations transition copy

- Two-Phase Variable Transfer
   ```rust
   // Phase 1: Copy indirect/direct variables to temporary buffer
   let buffer = copy_live_vars_to_temp_buffer(src_stackmap);
   // Phase 2: Transfer from buffer to destination locations
   set_destination_live_vars(buffer, dst_stackmap);
   ```
- Locations Mappings:
   - **Register → Register**: Direct transfer via temporary if needed
   - **Memory → Register**: Load from stack to register
   - **Register → Memory**: Store register to stack
   - **Memory → Memory**: Transfer through temporary register

tests/c/trace_while_executing2.c
```diff
@@ -1,3 +1,4 @@
+// ignore-if: test $CP_PATCHPOINT=1
```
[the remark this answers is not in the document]
Pavel-Durov: removed :point_right: 8b1340a30d2388c845c20f8765cf62f2d3a87694

ykrt/src/trace/swt/cp.rs
Pavel-Durov: Added comment for clarity  :point_right: 94673a94fc2134499265afdc2b89b6c954115125

ykrt/src/mt.rs
ltratt: My local version of `cargo fmt` isn't changing this, and I'd be kinda surprised if it did. I wonder if you changed this manually, by accident, and then ascribed it to `cargo fmt`?

ykrt/src/trace/swt/asm.rs
ltratt: That needs documenting with `//!` at the top of this module.

ykrt/src/trace/swt/buffer.rs
```diff
@@ -0,0 +1,205 @@
...
+// to the same buffer would be unsafe. Each buffer should be used by only one
+// thread at a time, or protected by explicit synchronization.
+
+thread_local! {
+    static OPT_BUFFER: std::cell::RefCell<Option<AlignedBuffer>> = std::cell::RefCell::new(None);
+    static UNOPT_BUFFER: std::cell::RefCell<Option<AlignedBuffer>> = std::cell::RefCell::new(None);
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct LiveVarsBuffer {
```
ltratt, line 72: I can only roughly guess what this struct is for. A doc string would help me a lot!

ykrt/src/mt.rs
Pavel-Durov: Interesting, I tested it yesterday and it did remove, will check again.

ykrt/src/trace/swt/asm.rs
Pavel-Durov: will add

ykrt/src/log/stats.rs
```diff
@@ -47,6 +47,10 @@ struct StatsInner {
     /// How many times have traces been executed? Note that the same trace can count arbitrarily
     /// many times to this.
     trace_executions: u64,
+    /// How many times have SWT transitions from unopt to opt been executed?
+    swt_transition_unopt_to_opt: u64,
```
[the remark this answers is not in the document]
Pavel-Durov: Yes, they will be the same unless something goes completely wrong.

ykrt/src/trace/swt/cp.rs
```diff
@@ -0,0 +1,713 @@
...
+    // Example: r15 address = rbp - rbp_offset_reg_store
+    let rbp_offset_reg_store = src_frame_size as i64 + (14 * REG64_BYTESIZE) as i64;
+
+    let temp_live_vars_buffer = copy_live_vars_to_temp_buffer(&mut asm, src_rec, transition.smid);
+    if *YKD_SWT_VERBOSE {
+        eprintln!(
+            "Transition from {:?} to {:?}, trace: {:?}",
+            src_smid, dst_smid, transition.trace_addr
+        );
+        eprintln!(
```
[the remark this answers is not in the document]
Pavel-Durov: Yes, I've implemented several env vars for debugging the control point transition. 
These could potentially be converted to compile-time configuration:

`YKD_SWT_VERBOSE` - for verbose logs
`YKD_SWT_VERBOSE_ASM` - generated asm logs
`YKD_SWT_CP_BREAK` - Inserts an `int3` (breakpoint) instruction at the beginning of generated assembly code

Pavel-Durov:
Docs added:
cp.rs - cc6be4fad273acd0bde6108aa358c5616d18ce0d
asm.rs - 582d5e382c717c415f23914e578e96f97d3708a6
buffer.rs - f4f88d2742f62e8093c1351b00848ff2b5801d6c
live_vars.rs - 9a7e5dbf9ed0de34a754f282cb2af82fc28a70e4
langtest - efece01c84557d56eed31a054b30fd89fbe694ac

ykrt/src/trace/swt/buffer.rs
```diff
@@ -0,0 +1,205 @@
...
+#[derive(Debug)]
+pub(crate) struct AlignedBuffer {
+    ptr: *mut u8,
+    layout: Layout,
+}
+
+impl AlignedBuffer {
+    /// Create a new aligned buffer with the specified size
+    pub(crate) unsafe fn new(size: usize) -> Self {
+        let layout = Layout::from_size_align(size, 16)
```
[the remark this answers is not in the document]
Pavel-Durov: I think 8-byte alignment makes more sense here because we currently only copy 1, 2, 4, and 8-byte values
updated :point_right: afa285d0684589fd76dc71c11612fd50543c2537

ltratt:
Let's wait for benchmark results before working out what we do next: they could change a great deal about what we think the right path is!

Pavel-Durov:
> Let's wait for benchmark results before working out what we do next: they could change a great deal about what we think the right path is!

Ok. But currently I can only run subset of these benchmarks cause of that bug with trace_builder missing trace instructions with longjumps.

ltratt:
> that bug with trace_builder missing trace instructions with longjumps.

Does merging https://github.com/ykjit/yk/pull/1834 into your branch help here?

Pavel-Durov:
> Let's wait for benchmark results before working out what we do next: they could change a great deal about what we think the right path is!

With awfy benchmarks that runs I get this:

`Datum2` - yk swt
`Datum3` - yk swt + multimod

| Benchmark             | Datum2 (ms, yk swt) | Datum3 (ms, yk swt + multimod) | Ratio | Summary        |
|-----------------------|---------------------|----------------------------------|-------|----------------|
| Queens/YkLua/1000     | 487                 | 470                              | 0.96  | 3.59% faster   |
| Permute/YkLua/1000    | 846                 | 819                              | 0.97  | 3.28% faster   |
| Towers/YkLua/600      | 894                 | 895                              | 1.00  | 0.10% slower   |
| Mandelbrot/YkLua/500  | 129                 | 133                              | 1.03  | 3.17% slower   |
| NBody/YkLua/250000    | 459                 | 474                              | 1.03  | 3.33% slower   |
| Sieve/YkLua/3000      | 448                 | 675                              | 1.51  | 50.74% slower  |

ykrt/src/mt.rs
```diff
@@ -480,6 +487,16 @@ impl MT {
...
                 self.stats.timing_state(TimingState::JitExecuting);
+                #[cfg(all(swt_modclone, target_arch = "x86_64"))]
+                unsafe {
+                    // Do the transition to unopt only if we are in opt mode.
+                    if smid == ControlPointStackMapId::Opt.into() {
+                        cp_transition_to_unopt_and_exec_trace(frameaddr, trace_addr, &self.stats);
+                    }
+                }
+
+                // FIXME: Calling this function overwrites the current (Rust) function frame,
```
Pavel-Durov, line 498: Remove this comment

PR #1823, opened 2025-07-21
Allow through some overlapping memory deopts.

vext01:
ack.lua is giving us:
```
thread '<unnamed>' panicked at ykrt/src/compile/jitc_yk/codegen/x64/deopt.rs:128:25:
Overlapping memory deopt! range: -56..-48, val: 0x0000000000000003 vs. range -56..-52, val: 0x0000000000000003
```

Here the second write is actually idempotent, because on a little endian X64 machine (where the 0x3 byte will come first in memory), the second write is ineffectual.


ltratt:
This is a known artefact of LLVM (and jitc_yk does the same thing, for the same reasons!). If you `zext` a value, you can store its smaller and lower values (they're the same!) in one location/register. [This is true, but with some additional qualifications, for `sext`.]

PR #1834, opened 2025-08-06
Reliably detecting longjmp (again).

vext01:
This is a draft PR that requires discussion. The code is not yet polished.

The other day I devised a [test](ef2ec80813a894591dc892965faad78a833cc872) where it's difficult to say whether a longjmp occurred or not using the information currently available to the system. The test also demonstrates that miscompilation currently occurs because the trace builder fails to recognise the longjmp.

I've been racking my brain trying to find a way to solve this and I've convinced myself that it can't be done using only the information currently available to the JIT (i.e. control-flow-integrity and frame depths).

This PR shows the only other way I can think of detecting longjmp reliably: using `ld --wrap` (similarly to how we do for `pthread_create`) to wrap `longjmp`. In our wrapper, if the system is tracing, we invalidate the trace (and stop pushing into the trace vector). When we stop tracing, we then discard invalid traces.

This catches not only the problematic case that I devised, but also all other cases already tested (as can be seen by the change of error message in those tests).

Pros:
 - We can catch all longjumps in traces in one place via the same method.
 - We can catch them sooner: at stop_tracing time, rather than in the trace compiler.

Cons:
 - Calling longjmp with Rust frames on the stack is a [grey area](https://github.com/rust-lang/rfcs/issues/2625). Then again, if we are careful, I think we might be able to do this safely. See the comment in the code for my attempt at a justification of safety.  It appears to be working in practice.

Preliminary measurements:
```
Datum0: before-longjmp-detect
Datum1: after-longjmp-detect

 Benchmark                  Datum0 (ms)  Datum1 (ms)  Ratio  Summary
 Permute/YkLua/1000         815          778          0.95   4.52% faster
 revcomp/YkLua/default      1838         1758         0.96   4.36% faster
 DeltaBlue/YkLua/12000      2002         1917         0.96   4.27% faster
 Towers/YkLua/600           925          890          0.96   3.84% faster
 LuLPeg/YkLua/default       3244         3122         0.96   3.77% faster
 Storage/YkLua/1000         9816         9477         0.97   3.45% faster
 HashIds/YkLua/6000         2842         2747         0.97   3.32% faster
 binarytrees/YkLua/15       3977         3861         0.97   2.92% faster
 Heightmap/YkLua/2000       774          753          0.97   2.72% faster
 CD/YkLua/250               9004         8822         0.98   2.02% faster
 fasta/YkLua/500000         841          825          0.98   1.91% faster
 Json/YkLua/100             2469         2435         0.99   1.38% faster
 Sieve/YkLua/3000           471          466          0.99   1.07% faster
 knucleotide/YkLua/default  1615         1604         0.99   0.71% faster
 NBody/YkLua/250000         511          508          0.99   0.63% faster
 Richards/YkLua/100         4075         4069         1.00   0.15% faster
 spectralnorm/YkLua/1000    902          901          1.00   0.12% faster
 Mandelbrot/YkLua/500       129          129          1.00   0.08% slower
 Bounce/YkLua/1500          1138         1140         1.00   0.23% slower
 BigLoop/YkLua/1000000000   1614         1625         1.01   0.66% slower
 Havlak/YkLua/1500          19408        19591        1.01   0.94% slower
 Queens/YkLua/1000          492          498          1.01   1.38% slower
 fannkuchredux/YkLua/10     1219         1238         1.02   1.57% slower
 List/YkLua/1500            840          888          1.06   5.75% slower
```

I manually checked the benchmarks with the biggest apparent slowdown (List) and biggest apparent speedup (permute) using multitime. I can't say I see an "obviously significant performance difference". There's some variation at the process execution level and sometimes different traces are collected, but it looks like nothing more than non-determinism. Neither of those benchmarks report longjmps before or after this change.

I'm curious to see what others think.

TODO:
 - detect siglongjmp by the same method.
 - tidy up.

### DIFF
--- a/bin/yk-config
+++ b/bin/yk-config
@@ -108,6 +108,8 @@ handle_arg() {
             else
                 # SWT-specific flags.
                 OUTPUT="${OUTPUT} -Wl,--mllvm=--yk-basicblock-tracer"
+                # Wrap longjmp so we can detect it.
+                OUTPUT="${OUTPUT} -Wl,--wrap=longjmp"
                 # Experimental module cloning support.
                 if [ "${YKB_SWTMODCLONE}" = "1" ]; then
                     OUTPUT="${OUTPUT} -Wl,--mllvm=--yk-module-clone"
--- a/tests/c/longjmp_from_foreign_into_trace1.c
+++ b/tests/c/longjmp_from_foreign_into_trace1.c
@@ -5,8 +5,7 @@
 //   stderr:
 //     yk-tracing: start-tracing
 //     we jumped
-//     yk-tracing: stop-tracing
-//     yk-warning: trace-compilation-aborted: irregular control flow detected
+//     yk-warning: stop-tracing-aborted: traced longjmp
 //     ...
 
 // Tests that we can deal with setjmp/longjmp when we jump from foreign code
--- a/tests/c/longjmp_from_foreign_into_trace2.c
+++ b/tests/c/longjmp_from_foreign_into_trace2.c
@@ -5,8 +5,7 @@
 //   stderr:
 //     yk-tracing: start-tracing
 //     we jumped
-//     yk-tracing: stop-tracing
-//     yk-warning: trace-compilation-aborted: irregular control flow detected
+//     yk-warning: stop-tracing-aborted: traced longjmp
 //     we jumped
 //     exit
 //     ...
--- a/tests/c/longjmp_problem.c
+++ b/tests/c/longjmp_problem.c
@@ -1,0 +1,107 @@
+// Run-time:
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_LOG=4
+//   env-var: YKD_LOG_IR=jit-pre-opt
+
+#include <assert.h>
+#include <setjmp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+#define ITERS 5
+
+#define G_LONGJMP     1
+#define G_CALL_DEEPER 2
+
+jmp_buf buf;
+
+void g(int x) {
+  if (x == G_LONGJMP) {
+    fprintf(stderr, "I'm in g via a longjump\n");
+  } else {
+    fprintf(stderr, "I'm in g via calling deeper\n");
+  }
+}
+
+// This function is opaque to the JIT.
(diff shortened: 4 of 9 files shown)


ltratt:
Initial reaction: I think that this PR might be a reasonable quick fix *but* I think we can perhaps also see the outlines of a "perhaps better / more complete" solution. Warning: I might be very wrong.

Let me start with an observation: in a sense, the problem seems to boil down to "what happens when we encounter outlined function calls?" I *think* this PR might be assuming that we have to look for the next traced block? But I would say "we're looking for the next successor block the outlined function call" which is basically a variant of our "look for the end of a recursive function call" stack. In `trace_builder` it's hard to see what's going on in this case, but the j2 equivalent hopefully makes it plain:

```rust
let mut recurse = 0;
loop {
    let ta = {
        let Some(Ok(ta)) = self.ta_iter.peek() else {
            panic!()
        };
        ta.to_owned()
    };
    let cnd_bid = self.ta_to_bid(&ta).unwrap();
    if cnd_bid.funcidx() == bid.funcidx() {
        if cnd_bid.is_entry() {
            recurse += 1;
        } else if self.am.bblock(&cnd_bid).is_return() {
            assert!(recurse > 0);
            recurse -= 1;
        }

        if recurse == 0 && cnd_bid == next_bid {
            break;
        }
    }
    self.ta_iter.next();
}
```

In essence we keep looking at the trace for the successor block `next_bid` -- *but* we're also aware that one way we can see that is if we keep recursing. So we keep track of how many recursive frames deep we are with the `recurse` variable: only when it's at 0 do we actually match `cnd_bid == next_bid`.

So when we hit an opaque call -- whether that's an outline call *or* (to keep it simple) a non-promoted indirect function call -- I claim we can pull off the same trick. Let's look at the core code:

```c
while (i > 0) {
  maybe_hidden_longjmp(i);
  i--;
}
```

`maybe_hidden_longjmp` is outlined so at that point my claim is that we can use something similar to the j2 code above: we keep searching for the `i--` block at the appropriate frame, and if we can't find it, the trace is ill-formed.

Let's try and think of a nasty case, which I think we can adapt from this PR: what happens if `maybe_hidden_longjmp` jumps *over* the interpreter frame? At that point, the "you can't trace [on x64] a frame at a higher address than where you started" check should kick in and we wouldn't even go around the `loop` above. [Right now it may end up creating one of Lukas's new trace return types; we could probably fix that in this case.]

Can this be broken? Dunno!

PR #1839, opened 2025-08-18
longjmp() soundness.

vext01:
- Reject a trace if it outlines over the block that started outlining in the first place (i.e. before we see the successor to the block that started outlining). This could be a `longjmp()`.

 - Reject a trace if the trace builder finishes in outlining mode and the call being outlined is *not* `yk_mt_control_point()`. This could also be a longjmp() (e.g. the `opaque_longjmp.c` test).

This means we reject some traces that would actually be OK (see disabled tests), but we are at least (hopefully) sound now.

In terms of performance, I don't think I can see any real effect, at least not for our current set of benchmarks. I measured List manually with `multitime` and saw no discernible effect.

```
Datum0: master
Datum1: longjmpv3

 Benchmark                  Datum0 (ms)  Datum1 (ms)  Ratio  Summary
 NBody/YkLua/250000         467          437          0.94   6.46% faster
 revcomp/YkLua/default      1795         1724         0.96   3.92% faster
 HashIds/YkLua/6000         2728         2634         0.97   3.44% faster
 Storage/YkLua/1000         9521         9232         0.97   3.04% faster
 CD/YkLua/250               8797         8550         0.97   2.81% faster
 binarytrees/YkLua/15       3896         3788         0.97   2.77% faster
 LuLPeg/YkLua/default       3104         3035         0.98   2.23% faster
 DeltaBlue/YkLua/12000      1930         1887         0.98   2.19% faster
 Havlak/YkLua/1500          18687        18292        0.98   2.11% faster
 Json/YkLua/100             2324         2276         0.98   2.08% faster
 Towers/YkLua/600           885          873          0.99   1.35% faster
 fannkuchredux/YkLua/10     1243         1227         0.99   1.32% faster
 Heightmap/YkLua/2000       757          748          0.99   1.22% faster
 Sieve/YkLua/3000           489          484          0.99   1.00% faster
 knucleotide/YkLua/default  1546         1536         0.99   0.62% faster
 fasta/YkLua/500000         800          797          1.00   0.36% faster
 Mandelbrot/YkLua/500       128          128          1.00   0.18% faster
 spectralnorm/YkLua/1000    898          896          1.00   0.13% faster
 Richards/YkLua/100         3993         4000         1.00   0.18% slower
 BigLoop/YkLua/1000000000   2378         2384         1.00   0.24% slower
 Bounce/YkLua/1500          1121         1125         1.00   0.34% slower
 Queens/YkLua/1000          484          499          1.03   3.15% slower
 Permute/YkLua/1000         758          783          1.03   3.36% slower
 List/YkLua/1500            826          856          1.04   3.63% slower
 ```

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -10,7 +10,7 @@ YKLUA_COMMIT="8a288908efc4611892993a32f8ffadd08df8e8c6"
 YKCBF_REPO="https://github.com/ykjit/ykcbf.git"
 YKCBF_COMMIT="431b92593180e1e376d08ecf383c4a1ab8473b3d"
 
-TRACERS="hwt swt"
+TRACERS="swt"
 
 # Build yklua and run the test suite.
 #
@@ -266,7 +266,7 @@ done
 # benchmark individually.
 #
 # Note: collect_and_decode is hwt-specific.
-YKB_TRACER=hwt cargo bench --bench collect_and_decode -- --profile-time 1
+#YKB_TRACER=hwt cargo bench --bench collect_and_decode -- --profile-time 1
 YKB_TRACER=swt cargo bench --bench promote -- --profile-time 1
 
 # Test some BF programs.


vext01:
That test is what was originally called `longjmp_problem.c`. It's a previously untested scenario where, without our changes, there's ambiguity between call-deeper or longjmp. Without our current change, we miscompile the trace horribly.

I'd like to keep it, if that's ok.

ltratt:
I'm still confused: does it test anything different than the new test in https://github.com/ykjit/yk/commit/1264c5ae23a0e1fac4c11ef78f5b33229511923d ?

vext01:
`opaque_longjmp.c` jumps to an opaque parent, the other one jumps to a non-opaque parent interpreter loop.

vext01:
Force pushed and rebased.

ltratt:
We don't. I'm fine with disabling this check on hwt (or whatever is the easiest thing to do there).

vext01:
Disabled testing hwt altogether.

PR #1840, opened 2025-08-18
Fix guard index when displaying guard instructions.

ptersilie:
The JIT IR optimisation can remove guard instructions, which invalidates the `gidx` stored in later guard instructions. The codegen, for this reason, ignores the stored `gidx` entirely and uses the position in the vector as the guard index. In order to not confuse ourselves when displaying JIT IR, we should do the same there.

This is a bit of a hack and not the fastest but printing IR is slow anyway so this doesn't matter too much.


ptersilie:
Squashed.

PR #1867, opened 2025-10-05
Determine the trace optimisation level in MT.

ltratt:
Rather than having each JIT backend interpret `YKD_OPT` (which, perhaps, should be `YK_OPT` but that's a decision for another day), record it at the `MT` level. We could be very clever with what the levels mean (should they be an `Option`? an `enum`? or ...?), but ultimately each backend optimiser can interpret them differently: really only "0" and "not 0" make sense at a global level. Hence defaulting to "1" is probably good enough for now and quite a way into the future.

### DIFF
--- a/ykrt/src/mt.rs
+++ b/ykrt/src/mt.rs
@@ -10,7 +10,7 @@ use std::{
     marker::PhantomData,
     sync::{
         Arc,
-        atomic::{AtomicBool, AtomicU16, AtomicU32, AtomicU64, Ordering},
+        atomic::{AtomicBool, AtomicU8, AtomicU16, AtomicU32, AtomicU64, Ordering},
     },
 };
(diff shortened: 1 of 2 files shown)


ykrt/src/mt.rs
```diff
@@ -94,6 +94,9 @@ pub struct MT {
     hot_threshold: AtomicHotThreshold,
     sidetrace_threshold: AtomicHotThreshold,
     trace_failure_threshold: AtomicTraceCompilationErrorThreshold,
+    /// The requested optimisation level, where 0 is the lowest possible level1. How these levels
```
vext01, line 97: > 0 is the lowest possible level1

I assume that's a typo?

vext01:
No objections.

PR #1871, opened 2025-10-15
Implement the bare minimum to get the float.c lang_test working.

ltratt:
More-or-less that means implementing `fpext`, `sitofp`, and some parts of `call`.

I've removed the `floats.c` test as it is a weaker version of `float.c`.


ltratt:
The benchmark doesn't do any floating point comparisons. But I'll get round to fcmp and friends soon.

PR #1872, opened 2025-10-16
Arrange register fills only once

ltratt:
The "aim" of this PR is the final commit (https://github.com/ykjit/yk/commit/ee26b58bdc850554a3258c78cd260a58e5ddc270) which reduces a silly inefficiency in the code we generate.

To get there, it made sense to rationalise the way the register allocator goes about testing. Long story short: we can get around most of fm's seeming-shortcomings by allowing multiple test patterns (https://github.com/ykjit/yk/commit/b54d5f96090b547659fad77110cb8e99511f02ab). The for the register allocator specifically, we want to log details that are not relevant to all tests so we then provide a way for tests to focus on what they want (https://github.com/ykjit/yk/commit/436c6a1438191de1fed816664a9ffd9c23eda826). In retrospect, these two techniques are so obvious that I can't believe it's taken me until now to realise that they could exist!


vext01:
I think I get it!

PR #1877, opened 2025-10-19
Floating point constants

ltratt:
This PR is (mostly) about supporting floating point constants in traces. There's some awkwardness in this, because x64 doesn't allow floating point immediates to be put directly in `xmm` registers: they either need to be loaded from RAM or bitcast from a general purpose register.

The first part of this PR (chiefly https://github.com/ykjit/yk/commit/3ee7cda65a2f5bf3eb84aae0a5374fc01b5227e4, https://github.com/ykjit/yk/commit/2bcdd45582cdf9480fceff4053d641c8100cf071, and https://github.com/ykjit/yk/commit/1919167eaccea974c4c0ffc3516bfd77336174be) gets the system ready for the concept of "putting constants in a register might need a temporary register" before actually doing so in https://github.com/ykjit/yk/commit/822ffc529a2765492547d50fda81c501360766b2. Interestingly, https://github.com/ykjit/yk/commit/2bcdd45582cdf9480fceff4053d641c8100cf071 gets rid of a hack that had previously troubled me but I couldn't see a way around, so that's a useful improvement full stop!

To actually get the lang_tests that use this doing anything, one then needs to support calling functions with floating point arguments / return types (https://github.com/ykjit/yk/commit/e3c9d4d63fd0097a7ed078c9ae26b05ab7815b22). With this, 3 floating point lang_tests now pass (the remaining ones are due to unimplemented operations such as `fmul`; but I think this PR is quite big enough already!).


ltratt:
Hmm `lang_tests::many_threads_many_locs.c` again. We need to look into that -- but it's not caused by this PR.

PR #1885, opened 2025-10-23
Implement `fcmp`

ltratt:
The main part of this PR is https://github.com/ykjit/yk/pull/1885/commits/7b7733fbfed4d34d56495a3dbe411f1988764537, but I realised I hadn't fully considered how to differentiate icmp/fcmp before. I've chosen to follow LLVM's syntactic lead over jitc_yk's, because there didn't seem a compelling difference, and so LLVM's popularity was the tie-breaker.


ltratt:
Yep, they all work with this PR!

PR #1889, opened 2025-10-24
Implement `fptosi`.

ltratt:
This gets the last floating point lang_tests passing.


ltratt:
`many_threads_one_loc` again...

PR #1891, opened 2025-10-24
Arrange destination register fills after copying.

ltratt:
This one was staring me in the face: reverse code generation was doing copying and arranging fills in the wrong order.


vext01:
This is a soundness fix, right?

ltratt:
Correct.

PR #1900, opened 2025-11-03
Logical shift right must not shift in set bits.

ltratt:
For `lshr N` to work as expected, the most significant N bits must be set to zero i.e. the input value must be zero extended. This was previously not done correctly for `lshr x, const` for `x != 32`.


vext01:
Good catch!

PR #1901, opened 2025-11-03
Introduce register casting

ltratt:
The main aim of this PR is to introduce `RegCnstr::Cast` (https://github.com/ykjit/yk/commit/189300e756f846e5d38fc42913fbb22fffcd8116). I _think_ -- but we will need to implement a fair few more things before I can be 100% sure -- that this fixes a weakness in the previous API.

In essence, `InutputOutput` constraints hide cases where we can merge multiple instructions' values into a single register. Right now, this PR allows _some_ merging to happen, but until we teach the register allocator to move, and not spill, values, there are only a limited number of cases where this can happen. Still, getting the API (hopefully) right now has clarified quite a few things in my mind.


vext01:
This all sounds sensible to me

PR #1908, opened 2025-11-03
Get early return working.

ltratt:
This requires implementing `return` and some bits in `aot_to_hir.rs`. With this the `early_return1` lang_test passes.


ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -758,7 +771,13 @@ impl<Reg: RegT + 'static> AotToHir<Reg> {
                 Inst::Phi { .. } => self.p_phi(iid, prev_bid.unwrap().bbidx(), bid, inst)?,
                 Inst::Promote { .. } => self.p_promote(iid, bid, inst)?,
                 Inst::PtrAdd { .. } => self.p_ptradd(iid, inst)?,
-                Inst::Ret { .. } => self.p_ret(iid, inst)?,
+                Inst::Ret { .. } => {
+                    if self.p_ret(iid, inst)? {
+                        println!("woo");
```
vext01: woo! :)

PR #1916, opened 2025-11-05
Use a single `dlsym` cache.

ltratt:
Previously we cached some lookups, but not others. Given the relative cost of `dlsym`, a cache will sometimes be a noticeable win, so make sure we use it whenever possible.


vext01:
Are we duplicating the cache from ykaddr?

PR #1926, opened 2025-11-13
Support deopting of floating point registers.

ltratt:


ltratt:
Well, that was silly of me: fixed in the force push.

PR #1928, opened 2025-11-14
Detect irregular control flow

ltratt:
This PR makes `aot_to_hir` detect irregular control flow. _Mostly_ this is similar to `trace_builder` _but_ there are two differences:

1. `aot_to_hir` doesn't have to deal with hwt's quirks, so there's quite a bit of simplification possible because of that.
2. There is one part of `trace_builder` that doesn't make sense to me and that I haven't ported.

For (2) the bit that I'm confused by is at line 1701:

```rust
if outbid == &bid {
    // a longjmp (or similar) may have occurred.
    return Err(CompilationError::General(
        "irregular control flow detected (outline recursion)".into(),
    ));
}
```

AFAICS this means that outlining will stop if it detects the starting block ID again: in other words, this seems like a special check for recursion? But just below that code is a more general check which captures a more accurate version of this case. In other words, I _think_ the code above incorrectly fails on a recursion case that we can safely handle. The best evidence I have for this is that the `outline_recursion.c` test is ignored in jitc_yk and, if you unignore it, it hits the "irregular control flow detected (outline recursion)" case above: j2, in contrast, happily lets it through. Is this definitely correct in j2? I'm not 100% confident but I _think_ it is.

I'm also not very happy with the field added in https://github.com/ykjit/yk/commit/ca2a631b968ac6ce00b5a7fd5c52537e460aa11a, but I haven't yet thought of a significantly nicer design. I think it's better to move forward and for me to come back to that soon, hopefully having had some sort of minor brain wave.

### DIFF
--- a/ykrt/src/compile/j2/aot_to_hir.rs
+++ b/ykrt/src/compile/j2/aot_to_hir.rs
@@ -52,6 +52,11 @@ pub(super) struct AotToHir<Reg: RegT> {
     am: &'static Module,
     hl: Arc<Mutex<HotLocation>>,
     ta_iter: Peekable<TraceActionIterator>,
+    /// What was the previous [BBlockId] fully processed by [TraceActionIterator]? Note: this is a
+    /// bit more subtle than "the value before the most recent `next`". It really means "the last
+    /// value before `p_block` or equivalent fully ran". As that suggests, this is rather fragile:
+    /// there must be, and we should aim to find soon, a better way to do this.
+    prev_bid: Option<BBlockId>,
     trid: TraceId,
     bkind: BuildKind,
     /// The data passed to successive calls to `yk_promote`. Note: some of this may have been
@@ -108,6 +113,7 @@ impl<Reg: RegT + 'static> AotToHir<Reg> {
             am,
             hl,
             ta_iter: TraceActionIterator::new(ta_iter).peekable(),
+            prev_bid: None,
             trid,
             bkind,
             promotions,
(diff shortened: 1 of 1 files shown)


vext01:
Having reviewed the code, I think the intention of that snippet of code is:

"If you arrive at the block that contained the call that started outlining before you see the block that you'd land at if you were to return from that call, then we could have longjumped to before we started outlining"

This is confirmed by the commit message of 7bfac18b579bb60544b0695247276de2b374a8da:

```
     - Reject a trace if it outlines over the block that started outlining
       in the first place (i.e. before we see the successor to the block
       that started outlining). This could be a `longjmp()`.
```

You are right that it's indistinguishable from outline recursion though, which I guess is what this part of the commit message alludes to:

```
This means we reject some traces that would actually be OK (see disabled
tests), but we are at least (hopefully) sound now.
```

I'm not sure if this changes anything? Do the longjmp tests all pass? If so, I wonder if the test coverage includes this scenario.

ltratt:
jitc_yk cannot pass the `outline_recursion.c` test but j2 can pass the `outline_recursion.j2.c` test. Which I guess is the "see disabled tests" bit.

> If you arrive at the block that contained the call that started outlining before you see the block that you'd land at if you were to return from that call, then we could have longjumped to before we started outlining

I think this is true _but_ I think that if this does happen then it will cause our "check recursion when outlining" check to fail. Indeed, there is one yklua test where I see j2 failing with "outline didn't find a successor" where jitc_yk says "outline recursion detected".

I am am not yet 100% sure -- I'll definitely be thinking about this over the coming days -- but my I _think_ j2's check is sound, and lets through more good traces. I'm inclined to think we should go with it, and we can revisit this if we find a discrepancy later. The good news at the moment is that delta debugging between jitc_yk and j2 is easy, so when j2 is more complete we'll be able to look at a wide variety of (say) yklua traces and see which traces jitc_yk / j2 let through / reject, and better understand them.

ltratt:
`outline_recursion.j2.c` is that test (it was added in a previous PR and this PR doesn't break it).

ltratt:
Forced push Clippy fixes.

PR #1934, opened 2025-11-16
Implement various 64 bit cases we see in yklua tests

ltratt:
These are mostly fairly mechanical, though I've made one "mild optimisation" since I noticed it in this process (https://github.com/ykjit/yk/commit/51e89e8f09e475799edc3fb042ac2ead44a3a48c).

### DIFF
--- a/ykrt/src/compile/j2/x64/x64hir_to_asm.rs
+++ b/ykrt/src/compile/j2/x64/x64hir_to_asm.rs
@@ -3554,6 +3554,7 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
         )?;
         self.asm.push_inst(match bitw {
             1..=32 => IcedInst::with2(Code::Xor_rm32_r32, lhsr.to_reg32(), rhsr.to_reg32()),
+            64 => IcedInst::with2(Code::Xor_rm64_r64, lhsr.to_reg64(), rhsr.to_reg64()),
             x => todo!("{x}"),
         });
 
@@ -7236,6 +7237,22 @@ mod test {
               ...
               ; %2: i32 = xor %0, %1
               xor r.32._, r.32._
+              ...
+            "],
+        );
+
+        // i64
+        codegen_and_test(
+            "
+              %0: i64 = arg [reg]
+              %1: i64 = arg [reg]
+              %2: i64 = xor %0, %1
+              exit [%0, %2]
+            ",
+            &["
+              ...
+              ; %2: i64 = xor %0, %1
+              xor r.64._, r.64._
               ...
             "],
         );


ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -5777,6 +5782,22 @@ mod test {
...
+            "
+              %0: i8 = arg [reg]
+              %1: i8 = arg [reg]
+              %2: i8 = lshr %0, %1
+              exit [%0, %2]
+            ",
+            &["
+              ...
+              ; %2: i8 = lshr %0, %1
+              shr r.32.x, cl
```
Fidget-Spinner: I forgot that this meant a wildcard 32-bit register.. Just curious: why does it have to be 32-bit though?

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -3052,8 +3055,8 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
                 [
                     RegCnstr::InputOutput {
                         in_iidx: *lhs,
-                        in_fill: RegCnstrFill::Zeroed,
-                        out_fill: RegCnstrFill::Zeroed,
+                        in_fill: RegCnstrFill::Undefined,
```
Fidget-Spinner: I read your other commit, but I'm still slightly confused why this works? Could you explain please?

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Fidget-Spinner: Right. I was thinking if it were 64-bit instead. However, I think that probably causes more register pressure? So maybe we don't want that.

Fidget-Spinner (approved):
LGTM. Will wait for a few hours if Edd wants to look at this.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
ltratt: The advantage of using 32-bit registers is that, for 32 bit values, we implicitly zero extend them to 64-bits (i.e. we get some zero extension for free), which helps a little in some cases.

ltratt:
@Fidget-Spinner You're the sole reviewer here, so unless you _need_ a second opinion, this is yours to merge.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Fidget-Spinner: Cool. Thank you for the explanation.

Fidget-Spinner:
@ltratt can you please rebase or squash the existing commits into how ever you feel is appropiate? I think the current commit history is a little messy. Thanks!

PR #1935, opened 2025-11-16
`smax` and `smin` can handle all bit widths.

ltratt:
Because the inputs are `Signed`, it's correct -- if maybe not always efficient! -- to always use 64-bit comparisons. That means both these functions can handle all bit-widths as-is.


ltratt:
@Fidget-Spinner In our model, only the reviewer gets to press the merge button -- that's you :)

PR #1944, opened 2025-11-21
Hack in a fix required for a block-splitting optimisation.

vext01:
Requires: https://github.com/ykjit/ykllvm/pull/288

`prev_bid` doesn't take into account that we could have just returned from a function, meaning that it's possible to have a `prev_bid` from another frame (and therefore function), but the intention is that `prev_bid` is the previous BBlockId for the *current* frame.

This is a horrible hack. We expect trace_builder to be removed in favour of j2 in the nearish future, so we will tolerate it for now.

Initial benchmarking suggests a few % speedup for a couple of benchmarks.

### DIFF
--- a/ykrt/src/compile/jitc_yk/trace_builder.rs
+++ b/ykrt/src/compile/jitc_yk/trace_builder.rs
@@ -247,12 +247,19 @@ impl TraceBuilder {
     }
 
     /// Walk over a traced AOT block, translating the constituent instructions into the JIT module.
+    ///
+    /// When all is well:
+    ///
+    ///  - Returns `Ok(Some(new_prevbb))` if the block was terminated by a `ret` and tracebuilder's
+    ///    `prevbid` needs to be updated as a result.
+    ///
+    ///  - Returns `Ok(None)` for all other block terminators.
     fn process_block(
         &mut self,
         bid: &aot_ir::BBlockId,
         prevbb: &Option<aot_ir::BBlockId>,
         nextbb: Option<aot_ir::BBlockId>,
-    ) -> Result<(), CompilationError> {
+    ) -> Result<Option<Option<BBlockId>>, CompilationError> {
         // We have already checked this (and aborted building the trace) at the time we encountered
         // an `Inst::Ret` that would make the frame stack empty.
         assert!(!self.frames.is_empty());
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -263,12 +270,15 @@ impl TraceBuilder {
...
-                aot_ir::Inst::Br { .. } => Ok(()),
+                aot_ir::Inst::Br { .. } => Ok(None::<Option<BBlockId>>),
                 aot_ir::Inst::Load {
                     ptr,
                     tyidx,
                     volatile,
-                } => self.handle_load(bid, iidx, ptr, tyidx, *volatile),
+                } => {
+                    self.handle_load(bid, iidx, ptr, tyidx, *volatile)?;
+                    Ok(None)
```
ltratt: Would I be an idiot if I said that these `Ok(None)`s are not assigned to a variable and can thus be removed?

ykrt/src/compile/jitc_yk/trace_builder.rs
```diff
@@ -301,42 +314,76 @@ impl TraceBuilder {
...
+                    // This is nasty.
+                    let ret_prevbb: Option<Option<BBlockId>> =
+                        if let Some(ref ci) = retframe.callinst {
+                            Some(Some(BBlockId::new(ci.funcidx(), ci.bbidx())))
+                        } else {
+                            Some(None)
+                        };
+                    // Early return OK because `ret` is a terminator and there can be no further
+                    // instructions in the block.
+                    return Ok(ret_prevbb);
```
ltratt: AFAICS this is the only thing we need to add: the `return` short-circuits the implict `Ok(None)` at the end of the function?

vext01:
Force pushed a cleanup.

But I notice that the ykllvm submodule change is bigger than expected. It seems the submodule has been lagging behind a bit (probably as a result of backing out shims). I'll need to figure out how to handle this on Monday.

ltratt:
I don't _think_ the submodule is a problem here: you're right that you're just seeing the history of backing out shims. When the ykllvm PR merges I'll try this PR: I do not anticipate it causing us issues because of the "seemingly additional" (if that makes sense!) history.

PR #1949, opened 2025-11-25
Try to `mmap` memory near the interpreter and existing traces.

ltratt:
If we leave `mmap` to its own devices, it will often (but, at least on Linux, not always!) put traces very
far in virtual memory from the interpreter. That means we end up having to do inefficient far calls from traces back to the interpreter.

This commit takes inspiration from LuaJIT: we try to force `mmap` to allocate traces near to the interpreter. We use `mmap`'s "hint" feature: if you give it an address, it will "try" to allocate memory nearby in virtual memory. However, if nothing nearby is available, it will give up, and allocate somewhere completely different. What we do is to look at the address we've got back and see if, on our current platform, the address is nearby; if not, we `munmap` the memory, and try a different hint address.

This works quite well for yklua. My guess is that for interpreters with a much greater quantity of code (e.g. CPython / CRuby), we might have to fiddle with the "try another hint address" heuristic, but that's a bridge we can cross when we get to it.

### DIFF
--- a/ykrt/src/compile/j2/codebuf.rs
+++ b/ykrt/src/compile/j2/codebuf.rs
@@ -7,10 +7,9 @@
 //! the quantity of memory it contains will not change.
 
 use crate::compile::j2::SyncSafePtr;
-use libc::{
-    __errno_location, MAP_ANON, MAP_FAILED, MAP_PRIVATE, PROT_EXEC, PROT_READ, PROT_WRITE, mmap,
-    mprotect, munmap,
-};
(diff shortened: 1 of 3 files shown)


ltratt:
> For example if the trace has lots and lots of calls to f() and some to g(), then you could try hinting the address of f and then g.

Ideally we're hoping that all traces end up "near" to each other (which is what LuaJIT enforces!) so in that sense it doesn't make _too_ much difference which function we pick. That said, I'm pretty sure we'll need to revisit the "find a nearby address when `mmap` gave us a distant page" heuristic with bigger interpreters, so maybe ideas like this will become relevant then.

PR #1950, opened 2025-11-26
Rejig hir_to_asm

ltratt:
This is, mostly, a refactoring PR: most of it has no functional change at all (with a small exception in https://github.com/ykjit/yk/commit/0e668594b26e12724e512989e3239ef74f45a5d3). In essence, it renames and reorders the functions relating to the start and end of trace code generation in `hir_to_asm` and then makes sure that `x64hir_to_asm` (and the test suite in `regalloc`) follow suit. This will stop me, at least, getting confused when I move between files! As a small bonus, I've also better documented the general layout of traces, and tweaked some names of things to be a bit more future proof.


ltratt:
Force pushed a fmt fix.

PR #1954, opened 2025-11-26
Fix incorrect extension in jitc_yk cg_udiv.

vext01:
This has been bugging me for a while.

I'm pretty sure the input operands for a udiv should be zero extended, yet the right hand side is sign-extended.

Looking to j2 for a second opinion, we have:
```
        let [_lhsr, rhsr, _] = ra.alloc(
            self,
            iidx,
            [
                RegCnstr::InputOutput {
                    in_iidx: *lhs,
                    in_fill: RegCnstrFill::Zeroed,
                    out_fill: RegCnstrFill::Zeroed,
                    regs: &[Reg::RAX],
                },
                RegCnstr::Input {
                    in_iidx: *rhs,
                    in_fill: RegCnstrFill::Zeroed,
                    regs: &NORMAL_GP_REGS,
                    clobber: false,
                },
                RegCnstr::Clobber { reg: Reg::RDX },
            ],
        )?;
```

i.e. inputs are zero extended there too.

Should be zero extended, right?

### DIFF
--- a/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
+++ b/ykrt/src/compile/jitc_yk/codegen/x64/mod.rs
@@ -1390,7 +1390,7 @@ impl<'a> Assemble<'a> {
                         },
                         GPConstraint::Input {
                             op: rhs,
-                            in_ext: RegExtension::SignExtended,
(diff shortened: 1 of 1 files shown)


ltratt:
Oops, I agree!

PR #1957, opened 2025-12-03
Light(er)weight tracing.

vext01:
Requires: https://github.com/ykjit/ykllvm/pull/291

Decide if we the current thread is tracing inline, in the interpreter, rather than calling off to Rust in every block.

The Rust side now defines the thread local shared with LLVM IR generated by our software tracing pass.

Also many fixes to make this work.


ykrt/src/mt.rs
```diff
@@ -1319,17 +1324,18 @@ impl MTThread {
...
     /// Is this thread currently tracing something?
     pub(crate) fn is_tracing() -> bool {
-        THREAD_IS_TRACING.with(|x| x.load(Ordering::Relaxed) != IsTracing::None)
+        unsafe { __yk_thread_tracing_state != IsTracing::None }
     }
 
     /// What kind of tracing (if any!) is this thread undertaking?
     fn tracing_kind() -> IsTracing {
-        THREAD_IS_TRACING.with(|x| x.load(Ordering::Relaxed))
+        let raw = &raw mut __yk_thread_tracing_state;
```
vext01: I will try again (in a follow up PR) to use an atomic here, thus avoiding the need for &raw, but as you can see, this PR is already big enough.

vext01:
Not sure what happened there. Re-synced ykllvm anyway.

PR #1976, opened 2025-12-10
Optimise `mul` and `udiv`

ltratt:


ltratt:
OK that I wasn't expecting: it optimised something we relied on in a test! Fixed in the force push.

[This also highlights one thing about j2 I haven't resolved in my mind yet: it doesn't really have a concept of "pre" and "post" opt IR. I _could_ fix this by keeping two things around, but perhaps we will ultimately just say "if you want pre opt, use `YKD_OPT=0`". I'll punt on it for a while further I hope!]

vext01:
> it doesn't really have a concept of "pre" and "post" opt IR

Good point. The only benefit I can think of is that in non-unit tests you can match the pre-opt IR and the post-opt IR, if you are interested in checking a certain optimisation kicks in, but arguably that's better checked by unit tests.

PR #1979, opened 2025-12-11
Strength reduce abs in J2

Fidget-Spinner:
I will squash/cleanup the commit history before merging.

### DIFF
--- a/ykrt/src/compile/j2/opt/strength_fold.rs
+++ b/ykrt/src/compile/j2/opt/strength_fold.rs
@@ -13,6 +13,7 @@ use crate::compile::{
 
 pub(super) fn strength_fold(opt: &mut Opt, inst: Inst) -> OptOutcome {
     match inst {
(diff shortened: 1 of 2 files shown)


Fidget-Spinner:
Squashed.

ltratt:
@Fidget-Spinner Ah, right, that one actually can be true or false because it comes to us via an LLVM call. I forgot that!

We now need to look carefully at LLVM's definition in https://llvm.org/docs/LangRef.html#id2281 and make sure that the optimisation follows that spec to the letter. To do that we're going to have to expand the IR parser so that it's `abs %0, true` (etc) or perhaps we could even spell it out with `abs %0, int_min_poison` (I don't immediately have a strong opinion, but I _think_ I prefer the latter) where `int_min_poison` means "true" and the absence of it means "false".

Sound like a plan?

Fidget-Spinner:
Ready for re-review/merge queue after linting.

PR #1982, opened 2025-12-12
`int_min_poison`

ltratt:
#1979 showed that I hadn't thought about `int_min_poison` in `abs` properly. This commit adds support to the HIR parser to allow us to specify this on/off (https://github.com/ykjit/yk/commit/d2b53d9c42769fc030385833be4c9526721e2427). The first commit renames this to `int_min_poison` as I had previously misled myself with the "is_" prefix.

@Fidget-Spinner This should allow you to check whether the opt_abs tests are (a) complete given `int_min_poison: true` (we currently `assert` that's the case, so we don't yet need tests for the false case) (b) if not, raise a later PR extending things.


ykrt/src/compile/j2/opt/strength_fold.rs
```diff
@@ -895,7 +896,7 @@ mod test {
         test_sf(
             "
           %0: i8 = 2
-          %1: i8 = abs %0
+          %1: i8 = abs %0, int_min_poison
```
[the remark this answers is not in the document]
ltratt: Other way around: C sets `int_min_poison` to true, and currently we only care about C code, so the tests are "do what C does". We _could_ add "no int_min_poison" tests if we want, but we'd have to think about it!

ykrt/src/compile/j2/hir.y
```diff
@@ -59,8 +59,8 @@ Inst -> Result<AstInst, Box<dyn Error>>:
   | "LOCAL" ":" Ty "=" "CALL" "ID" "LOCAL" "(" Locals ")" {
       Ok(AstInst::Call { local: Some($1?.span()), ty: Some($3?), extern_: $6?.span(), tgt: $7?.span(), args: $9? })
     }
-  | "LOCAL" ":" Ty "=" "ABS" "LOCAL" {
-      Ok(AstInst::Abs { local: $1?.span(), ty: $3?, val: $6?.span() })
+  | "LOCAL" ":" Ty "=" "ABS" "LOCAL" IntMinPoison {
+      Ok(AstInst::Abs { local: $1?.span(), ty: $3?, val: $6?.span(), int_min_poison: $7? })
```
Fidget-Spinner: I tried searching the GRM book for what `.span()` does, but couldn't find it. What does it do?

ltratt: https://docs.rs/lrpar/0.14.0/lrpar/trait.Lexeme.html#tymethod.span

PR #1988, opened 2025-12-16
Constant fold FSub in J2

Fidget-Spinner:


Fidget-Spinner:
> FWIW I think this is "constant fold" not "strength reduce" (but these are slippery terms, I admit). Do you want to force push a commit / PR title change?

Oh yes, in this case it is a constant fold. I will force push a changed commit title.

PR #1989, opened 2025-12-16
Remove a `&mut self`

ltratt:
The main commit here is [cc72d](https://github.com/ykjit/yk/pull/1989/commits/cc72d31345785c6c1b6f2ab2a8b893380e131945): I always knew that having `equiv_iidx` take `&mut self` might cause borrow checker problems and, in a branch, I've finally encountered one. Oh well. We don't strictly _have_ to have `&mut self` for correctness: it's an optimisation.

[I will probably come back to this and use some sort of horror like `UnsafeCell` in the
future when I more precisely understand the performance impact. For now, I'm just happy to be able to use this without borrow checking errors.]

### DIFF
--- a/ykrt/src/compile/j2/aot_to_hir.rs
+++ b/ykrt/src/compile/j2/aot_to_hir.rs
@@ -239,7 +239,7 @@ impl<Reg: RegT + 'static> AotToHir<Reg> {
             let exit_vars = exit_safepoint
                 .lives
                 .iter()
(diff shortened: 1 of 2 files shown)


Fidget-Spinner (approved):
Yeah correctness is better. You explained this to me earlier today, so I know what it's doing at least.

ltratt:
Oops! Clippy errors fixed in the force push.

PR #1993, opened 2025-12-17
Common subexpression elimintation

ltratt:
This PR implements CSE: it is, I believe, correct, but there are also some rough edges. However, making everything perfect in one go doesn't seem the right trade-off: the rough edges can be smoothed off in subsequent PRs.

This benefits some benchmarks more than other: nbody speeds up by over 8%, but that's the most. Another 3 benchmarks definitely speed up, but by less. Most benchmarks see small/no improvements, at least at the granularity I can measure.

### DIFF
--- a/ykrt/src/compile/j2/opt/cse.rs
+++ b/ykrt/src/compile/j2/opt/cse.rs
@@ -1,0 +1,370 @@
+//! Common subexpression elimination.
+//!
+//! ## Efficiency
+//!
+//! CSE is simple but if implemented naively it has O(n^2) costs. We use two tricks to reduce the
+//! cost.
+//!
+//! First, we take a lesson from LuaJIT and maintain a linked list of instruction kinds: if we're
+//! checking CSE for, say [PtrAdd]s, that means we only check `PtrAdd`s (and not, say, [And] etc).
+//!
+//! Second, we can use an instruction's operand indices to work out if we've gone implausibly
+//! far back. For example, if we're checking `add %102, %105` for CSE, there is no chance of a
+//! match at an [InstIdx] below 102 and we can stop checking.
+//!
+//! This allows one to iterate backwards over a trace starting at instruction X and efficiently
+//! view all previous instructions of the same kind as X, typically bailing out early if there is
+//! no match.
+
+use crate::compile::j2::{
+    hir::{Inst, InstDiscriminants, InstIdx, InstT},
+    opt::OptT,
+};
+use strum::EnumCount;
+
+/// Common supexpression elimination.
+///
+/// Internally this uses a linked list for each [Inst] encoded in `heads` and `predecessors`.
+pub(super) struct CSE {
+    /// The head of the linked list with one entry for each [Inst] discriminant. This points to the
+    /// most recent entry in [Self::predecessors]. `InstIDX::MAX_INDEX` is used to represent "no
+    /// previous entry".
+    heads: [InstIdx; Inst::COUNT],
+    /// The linked list with one entry for each instruction in a trace. `InstIDX::MAX_INDEX` is
+    /// used to represent "no previous entry".
+    predecessors: Vec<InstIdx>,
+}
+
+impl CSE {
+    /// Create an empty CSE object. As new instructions are committed to the eventual [Module],
+    /// they must also be `push`ed here so that CSE can be in-sync with the [Module].
+    pub(super) fn new() -> Self {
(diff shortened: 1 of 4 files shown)


Fidget-Spinner (commented):
I have one quibble.

ykrt/src/compile/j2/opt/cse.rs
```diff
@@ -0,0 +1,345 @@
...
+            assert!(InstIdx::MAX_INDEX > Inst::COUNT);
+        }
+        Self {
+            heads: [InstIdx::from_usize(InstIdx::MAX_INDEX); Inst::COUNT],
+            predecessors: Vec::new(),
+        }
+    }
+
+    pub(super) fn is_equiv(&self, opt: &dyn OptT, inst: &Inst) -> Option<InstIdx> {
+        // FIXME: This is a hack for "does this instruction have side effects".
```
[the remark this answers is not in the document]
ltratt: Both options suck: what we need to do is get the effect system sorted (that is: the mechanism that allows us to tell if two instructions side effects overlap). That's on my list, but I haven't got to it yet.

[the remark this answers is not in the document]
ltratt: JSC has a really nice model which I think we can easily adapt. I hope to do that somewhat soon, as we need it for things like e.g. heap analysis.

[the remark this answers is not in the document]
ltratt: `ThreadLocal` always returns the same value, so is OK.

ykrt/src/compile/j2/opt/cse.rs
```diff
@@ -0,0 +1,345 @@
...
+    };
+    use std::{cell::RefCell, rc::Rc};
+
+    fn test_cse(mod_s: &str, ptn: &str) {
+        let cse = Rc::new(RefCell::new(CSE::new()));
+        opt_and_test(
+            mod_s,
+            |opt, mut inst| {
+                if let Inst::Guard(_) = inst {
+                    match strength_fold(opt, inst) {
```
Fidget-Spinner: I'm a little torn whether we want strength reduction to be in the CSE test. That would make it possible for strength reduction to hide a CSE bug. Ideally, a unit test should just test the CSE itself?

We can have a larger integration test that tests the whole optimization suite later.

Fidget-Spinner: A naiive question: why can't we optimise across that assuming same SSA variables and the operations are "pure". E.g.
```
          %0: i8 = arg [reg]
          %1: i8 = arg [reg]
          %4: i8 = add %0, %0
          %2: i1 = icmp eq %0, %1
          guard true, %2, []
          %5: i8 = add %1, %1
```
We should be able to CSE `%5` with `%4`?

ltratt: IIUC, that's what the test starting at line 169 shows.

Fidget-Spinner: Ah ok, that makes more sense now! You need the equivalences to be built before the CSE pass IIUC?

Fidget-Spinner: Ah I misunderstood the test then. Thank you. This makes sense. Basically the guard just says `%1 == %0` after the `icmp` and CSE can operate on that. Neat.

ykrt/src/compile/j2/opt/cse.rs
```diff
@@ -0,0 +1,345 @@
...
+            "
+          %0: i8 = arg
+          %1: i8 = arg
+          %2: i8 = add %0, %0
+          %3: i1 = icmp eq %0, %1
+          guard true, %3, []
+          blackbox %2
+          blackbox %2
+        ",
+        );
```
Fidget-Spinner: Can you please add one more test where CSE is done over a side-effecting instruction as well?
Maybe
```
          %0: ptr = arg [reg]
          %1: ptr = arg [reg]
          %2: i64 = arg [reg]
          %3: i64 = add %0, %1
          memcpy %0, %1, %2, true
          %4: i64 = add %0, %1 // This should be CSE-d as it is equivalent to %3 trivially.
          blackbox ...
```

ltratt: Test added in 8be0308e.

Fidget-Spinner:
Please squash. Thanks!

PR #1994, opened 2025-12-18
Call optimised functions if they're available.

ltratt:


ltratt:
I thought I'd run the lang_tests but I obviously didn't... mea culpa!

ltratt:
Fixed in the force push.

PR #2000, opened 2025-12-22
Rework the internal optimiser API.

ltratt:
The previous API was a rush job, and not a good one at that: it had no explicit notions of "passes", and had to treat `cse` differently than `strength_fold`; and it presented the same API to `aot_to_hir` as to "passes".

The main thing this commit does is to split the optimiser such that it has two different APIs: an outward-facing API for `aot_to_hir` and an internal-facing API (`PassOpt`) for optimisation passes. Both `strength_fold` and `cse` are now "proper" passes (implementing a trait `PassT`) on an equal footing.

The way this is achieved is by having the outward and internal facing objects share a `OptInternal` struct. The main advantage of this is that it allows us to assuage the borrow-checker. However, it also indirectly means that we can expose different functionality in the outward and inward-facing APIs. That does mean some changes in `hir`, in particular the introduction of a new trait `EquivIIdxT`, which allows us to pass much "smaller" objects around (i.e. not full optimisers!).

That allows this commit to deal with one of the true horrors of the previous situation: `strength_fold` could insert instructions into the trace. These are now "preinstructions" and the API makes it much harder for them to be used incorrectly (as well as documenting why/where one should think carefully about them).

Overall I think this commit shoves us a long way in the right direction. There is still work to do though -- notably `OptOutcome` as it currently stands doesn't allow a pass to hand-over knowledge it has about an instruction. However, this commit is big enough as-is, and I haven't yet fully worked out what the next step would be. So it feels better to consider this now, rather than wait ever longer for perfection.

### DIFF
--- a/ykrt/src/compile/j2/hir.rs
+++ b/ykrt/src/compile/j2/hir.rs
@@ -133,7 +133,7 @@ use crate::{
     compile::{
         j2::{
             compiled_trace::J2CompiledTrace,
-            opt::OptT,
+            opt::EquivIIdxT,
             regalloc::{RegT, VarLocs},
         },
         jitc_yk::{
@@ -558,7 +558,7 @@ pub(super) trait InstT: std::fmt::Debug {
 
     /// Canonicalise this instruction. This rewrites operands to their most recent equivalent
     /// [InstIdx]s and performs other, basic, canonicalisation on a per-instruction kind basis.
-    fn canonicalise(&mut self, _opt: &dyn OptT);
+    fn canonicalise<T: BlockLikeT + EquivIIdxT>(&mut self, _be: &T);
 
     /// For the purposes of common subexpression elimination is `other` equivalent to `self`?
     /// `other` must be canonicalised for this comparison to return true in all cases where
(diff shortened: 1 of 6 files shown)


ltratt:
> Why is it useful to have instructions hovering around that aren't accessible by the current pass?

Do you mean preinstructions? Right now I have only needed to insert constants, but it was as easy to consider the general case (i.e. all instructions), so I did that. Consider this trace:

```
%0: i8 = arg [reg]
%1: i8 = 8
%2: i8 = mul %0, %1
```

We want to convert this to:

```
%0: i8 = arg [reg]
%1: i8 = 8
%2: i8 = 2
%3: i8 = shl %0, %2
```

In other words we need to magic up the `Const` instruction `2` which does not exist in the trace. That's what preinstructions are used for right now.

In one sense, preinstructions are a horrible hack. The more general solution is to have the optimiser work on sequences of instructions. It is possible we will find we need to do that _but_ it's not fun to work with. If we're lucky, we will be able to avoid doing so, and keep the code much easier to understand because of it. For example, the CSE pass is now under 100LoC!

> when do we emit OptOutcome::Rewritten(Constant)`` rather than push_pre_inst(Constant)` or vice versa?

Using strength_fold as an example, constant folding always leads to `Rewritten`; strength reduction _sometimes_ needs preinstructions.

If we had:

```
%0: i8 = 2
%1: i8 = 3
%2: i8 = add %0, %1
```

we would want to turn this into just `%2: i8 = 5` i.e. `Rewritten` the `add` to a `Const`.

> I can sense a whopper of a merge conflict with my own branch

FWIW, migrating the two current passes was easy. That said, if I was you, I'd probably start a new branch and copy your code over and edit it in the process. But it could be done as a normal merge too!

Fidget-Spinner:
Thank you for the clarifications. This is a much cleaner API, so let's get this in.

PR #2001, opened 2025-12-24
Redundant load/store pass

ltratt:
This PR adds a new redundant `load_store` elimination pass (https://github.com/ykjit/yk/commit/2bdc42bc9233741f5ef2546b0b7ff75ef1b5d418): it's simple, and more-or-less a direct port from jitc_yk. The main reason I've done this now is to help me understand what's needed from the optimiser API: indeed, a change was necessary (https://github.com/ykjit/yk/commit/675444e991d9d9c98190317aecbd097442f0fb0d).

@Fidget-Spinner My guess is that the known_bits pass might need the extended `inst_committed` API too. And perhaps other bits I haven't considered yet.

### DIFF
--- a/tests/c/funcptrarg_noir.j2.c
+++ b/tests/c/funcptrarg_noir.j2.c
@@ -3,6 +3,7 @@
 //   env-var: YKD_SERIALISE_COMPILATION=1
 //   env-var: YKD_LOG_IR=jit-pre-opt
 //   env-var: YKD_LOG=4
+//   env-var: YKD_OPT=0
 //   stderr:
 //     yk-tracing: start-tracing
 //     z=3
--- a/tests/c/sext.c
+++ b/tests/c/sext.c
@@ -2,6 +2,7 @@
 //   env-var: YKD_LOG_IR=jit-pre-opt
 //   env-var: YKD_SERIALISE_COMPILATION=1
 //   env-var: YKD_LOG=4
+//   env-var: YKD_OPT=0
 //   stderr:
 //     yk-tracing: start-tracing
 //     -4 -4 -4 -4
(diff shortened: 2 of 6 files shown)


ltratt:
Oops, I should have added the tests in https://github.com/ykjit/yk/pull/2001/commits/410fa9ecaf18f37e07e25e01f4800432e8fab2c8 in the original PR. Will need squashing.

ltratt:
I do:

```
 fannkuchredux/yklua/10     13979 ± 103   13005 ± 143   0.93  6.97% faster
 nbody/yklua/250000          4694 ±  93    4458 ±  65   0.95  5.03% faster
 Heightmap/yklua/2000        8338 ±  76    7983 ±  12   0.96  4.26% faster
 sieve/yklua/3000            5066 ±  47    4860 ±  28   0.96  4.06% faster
 spectralnorm/yklua/1000    12564 ±  31   12340 ±   2   0.98  1.79% faster
 cd/yklua/250               33563 ± 140   33176 ± 161   0.99  1.15% faster
 mandelbrot/yklua/500        2266 ±   4    2292 ±   3   1.01  1.16% slower
```

Everything else is indistinguistable (apart from `list`, whose results bounce around too much to be useful). The `mandelbrot` result is odd but replicates at scale: there's nothing obvious causing it that I've yet understood. It may be as simple as slightly different code layout or similar...

Fidget-Spinner:
lang tests are failing

ltratt:
Ah, I'm being bitten by the lack of "proper" jit-pre-opt. I'll need to chew on the best solution to this.

ltratt:
After playing for a little bit, I don't see a nice way for j2 to do `jit-pre-opt` and `jit-post-opt`: it seems that I end up either giving misleading `jit-pre-opt` or messing up inlining edge cases (or add _lots_ of code to do both, which is messy). The new commit thus doesn't try to do this: it either makes the IR tests more general or sets `YKD_OPT=0`. Since the lang_tests are _mostly_ testing `aot_to_hir` I think the latter is OK.

While doing this, it turned out -- to my surprise! -- that we're also encountering the `const` case in j2 (which we didn't in jitc_yk, I think because of the worse inlining). So I've had to implement that too.

@Fidget-Spinner If you're OK with this, it will need squashing.

ltratt:
Squashed.

PR #2002, opened 2025-12-28
Slightly improve compilation performance

ltratt:
While looking at profiles of the j2's compiler, I noticed an easy win, which this PR addresses: `rstate_diff_to_action` was taking lots of time, mostly because it `malloc`ed chunks of memory in cases where it didn't need to. Fixing this makes a small (~5%) improvement to compilation time.


ltratt:
No thanks.

PR #2004, opened 2025-12-28
Implement known bits analysis

Fidget-Spinner:


ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,474 @@
...
+    pub(super) fn new() -> Self {
+        KnownBits {
+            iidx_to_known_bits: IndexVec::new(),
+            curr_value_pending_commit: None,
+        }
+    }
+
+    /// Returns what we known about the bits of `iidx`.
+    fn as_knownbits(&self, opt: &PassOpt, iidx: InstIdx) -> KnownBitValue {
+        if let Some(bits) = self.iidx_to_known_bits[iidx].clone() {
```
ltratt: I think this is `self.iidx_to_known_bits[iidx].unwrap_or_else(|| ...)` in disguise? Does clippy spot that?

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,474 @@
...
+        KnownBitValue {
+            ones: set_ones,
+            unknowns,
+        }
+    }
+}
+
+/// Known-bits analysis.
+pub(super) struct KnownBits {
+    iidx_to_known_bits: IndexVec<InstIdx, Option<KnownBitValue>>,
```
ltratt: These are very long names, and not hugely helpful :) This could just be something like `known_bits` or even shorter.

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,474 @@
...
+
+        // If we know the output's bits, emit that.
+        if res.all_known() {
+            return OptOutcome::Rewritten(Inst::Const(Const {
+                tyidx,
+                kind: ConstKind::Int(res.as_int()),
+            }));
+        }
+
+        // lhs = any and rhs = constant
```
ltratt: I'm not sure this comment adds anything?

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    pub fn unknown(bitw: u32) -> Self {
+        KnownBitValue {
+            ones: ArbBitInt::from_u64(bitw, 0),
+            unknowns: ArbBitInt::from_u64(bitw, !0u64),
+        }
+    }
+
+    fn as_int(&self) -> ArbBitInt {
```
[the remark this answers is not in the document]
Fidget-Spinner: Addressed in 8b922fb6df8aaac3ab076a130af8c7949d43f227

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,474 @@
...
+            .bitand(&set_ones.bitneg());
+        KnownBitValue {
+            ones: set_ones,
+            unknowns,
+        }
+    }
+}
+
+/// Known-bits analysis.
+pub(super) struct KnownBits {
```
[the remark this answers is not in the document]
Fidget-Spinner: Addressed in 038f2bd3fed439f3572b5b29880d1d0f1dc0affa

ykrt/src/compile/j2/opt/known_bits.rs
Fidget-Spinner: No it doesn't. I changed it in b69981a3cf7816458ce29fe6336e6b5a2a439157

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,474 @@
...
+        // If we know the output's bits, emit that.
+        if res.all_known() {
+            return OptOutcome::Rewritten(Inst::Const(Const {
+                tyidx,
+                kind: ConstKind::Int(res.as_int()),
+            }));
+        }
+
+        // lhs = any and rhs = constant
+        // If no new information (set zeroes) is gained, that means this
```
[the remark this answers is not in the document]
Fidget-Spinner: documented in 273af03d9ea7dfbb132198e40ee07dfd31a71c5f

ykrt/src/compile/j2/opt/known_bits.rs
Fidget-Spinner: Remove as part of documentation change in 273af03d9ea7dfbb132198e40ee07dfd31a71c5f

Fidget-Spinner:
> This is a quick partial review. Broadly speaking, I want to get this in, but I would like it to slightly better match existing conventions in the codebase (even trivial things like naming and comment width). I also think there are some trivial comments that are too long, and some surprising things that aren't documented (e.g. functions with surprising pre/post-conditions). Nothing major, but they'll just make the code easier to live with in the long term.

Yep I'm still getting a feel of the conventions. Please bear with me on this one, thanks!

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,474 @@
+//! Known bits analysis.
+//!
+//! Data-flow analysis to gain information about bits.
+//! We use similar ideas to PyPy:
+//! https://pypy.org/posts/2024/08/toy-knownbits.html
+//!
```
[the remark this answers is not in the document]
ltratt: I also meant (but wasn't clear... sorry!): module comments don't end with `//!` on a blank line on their own.

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,482 @@
...
+        },
+    },
+    jitc_yk::arbbitint::ArbBitInt,
+};
+use index_vec::IndexVec;
+
+/// Known-bits analysis.
+pub(super) struct KnownBits {
+    known_bits: IndexVec<InstIdx, Option<KnownBitValue>>,
+    pending_commit: Option<KnownBitValue>,
```
ltratt: This attribute needs documenting (I know what it does because I know its existence is because of an incomplete API... which is shameful on my part!).

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,482 @@
...
+            Inst::And(x) => self.opt_and(opt, x),
+            Inst::Const(x) => self.opt_const(x),
+            Inst::Or(x) => self.opt_or(opt, x),
+            _ => OptOutcome::Rewritten(inst),
+        }
+    }
+
+    fn inst_committed(&mut self, _opt: &CommitInstOpt, _iidx: InstIdx, _inst: &Inst) {
+        self.known_bits.push(self.pending_commit.clone());
+        assert_eq!(_iidx.index(), self.known_bits.len() - 1);
```
ltratt: `iidx` doesn't need the leading `_` if it's used in `assert`. Also, wouldn't this `assert` be clearer _before_ the push when the `- 1` wouldn't be needed?

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,482 @@
...
+            return OptOutcome::Rewritten(Inst::Const(Const {
+                tyidx,
+                kind: ConstKind::Int(res.as_int()),
+            }));
+        }
+
+        // If no new information (set zeroes) was gained, that means this
+        // op is useless.
+        if rhs_b.all_known()
+            && (rhs_b
```
ltratt: I'm surprised `cargo fmt` hasn't removed the brackets. If it doesn't, we should remove them, as "over bracketing" isn't really needed in Rust like it is in C.

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,482 @@
...
+    }
+}
+
+/// Known bits for a single value.
+///
+/// This uses the same bit representation found in
+/// https://pypy.org/posts/2024/08/toy-knownbits.html#the-knownbits-abstract-domain
+///
+/// In short:
+/// one     unknown     knownbit
```
ltratt: Probably best to either use Markdown formatting for the table (or, at a push, a code block).

ltratt: I would also probably make the first row the `?` state since that's the first. Then it becomes clear that one can go from row 1 to either rows 2 or 3, but no other transitions are legal (so we can condense the slightly long lattice comment perhaps?).

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,482 @@
...
+            ones: num.clone(),
+            unknowns: ArbBitInt::from_u64(bitw, 0),
+        }
+    }
+
+    /// Constructs an unknown KnownBitValue.
+    pub fn unknown(bitw: u32) -> Self {
+        KnownBitValue {
+            ones: ArbBitInt::from_u64(bitw, 0),
+            unknowns: ArbBitInt::from_u64(bitw, !0u64),
```
ltratt: `u6::MAX` may be clearer?

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,482 @@
...
+        }
+    }
+
+    /// If all bits are konnw, return the constant value.
```
ltratt: "konnw"

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,482 @@
...
+    fn as_int(&self) -> ArbBitInt {
```
ltratt: Should be called `as_arbbitint` I think.

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,482 @@
...
+          %3: i8 = 3
+          %4: i8 = or %0, %1
+          %5: i8 = or %4, %2
+          blackbox %5
+        ",
+        );
+    }
+
+    #[test]
+    fn opt_and_or() {
```
ltratt: I'm not convinced by this test because scaling it's going to be _really_ hard as we analyse more IR instructions! I'd probably remove it.

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,482 @@
...
+          %2: i8 = or %0, %1
+          %3: i8 = 1
+          blackbox %3
+        ",
+        );
+    }
+
+    #[test]
+    fn with_intermediate() {
+        // Test that other instructions stay around
```
ltratt: I'm not sure what these tests are testing.

ykrt/src/compile/j2/opt/known_bits.rs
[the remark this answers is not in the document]
Fidget-Spinner: Does the new documentation in 37aecec9bb207c056cc1fc17f3f310c593e77739 seem better?

ltratt:
Dumb question: with this PR do all the tests pass and benchmarks run?

ltratt:
I just realised that we probably need an additional test(s) where we prove that the knownbits pass can turn operations into constants. So I assume that we should be able to optimise this sort of thing:

```
%0: i8 = arg [reg]
%1: i8 = 3
%2: i8 = or %0, %1
%3: i8 = 1
%4: i8 = and %2, %3
```

into `%4: i8 = 1` (assuming I've got my set bits vaguely right)?

One other (very minor) comment: comments need formatting to 100 chars. [I'm somewhat neutral on comment width, but the current code is formatted to 100 chars so we should follow that _unless_ we make a consistent decision to reformat all comments.]

Fidget-Spinner:
100-width comments added in 332cee00ec13dcdf67fdb9c4ce1ae6f6b17bee0f
Constant tests added in 9938104b11c4bf9e542e9fa78122371d8da238b7

All tests pass on b16 and all of AWFY passes last I tried, I will try the full benchmark suite.

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,398 @@
...
+/// | 0   | 1       | ?        |
+/// | 0   | 0       | 0        |
+/// | 1   | 0       | 1        |
+/// | 1   | 1       | illegal  |
+///
+/// To ensure monotonicity,transitions from `?` to` 0` or `1` are valid, but not the other way
+/// around. `illegal` occurs when both `0` and `1` are set and known, which is impossible in a
+/// valid program. `illegal` indicates a likely bug in the optimizer/IR.
+#[derive(Clone)]
+pub struct KnownBitValue {
```
ltratt: Oops, just realised: why are these `pub`? They shouldn't be known outside this file AFAICS?

ltratt:
Two minor comments: we're almost there!

ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -0,0 +1,398 @@
...
+    }
+
+    /// Returns an integer with all the known ones.
+    fn known_ones(&self) -> ArbBitInt {
+        self.knowns().bitand(&self.ones)
+    }
+
+    fn bitand(&self, other: &KnownBitValue) -> KnownBitValue {
+        let set_ones = self.ones.bitand(&other.ones);
+        let set_zeroes = self.zeroes().bitor(&other.zeroes());
```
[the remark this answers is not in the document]
Fidget-Spinner: This is due to the bit representation: ones is not computed, while zeroes is.

Fidget-Spinner:
All tests pass. All benchmarks on `yk-benchmarks` pass.

Fidget-Spinner:
Oops, the clippy on the buildbot fails but not locally, guess the buildbot has a newer clippy? In any case I force pushed a formatting fix. Please retry, thanks!

ltratt:
buildbot always uses the latest nightly Clippy. I personally find it worth `rustup`ing my local rustc roughly monthly, as rustc and the other tools do introduce "won't merge" changes every so often.

Issue #1933, opened 2025-11-16
imm32 handling in the x64 backend can be improved


ltratt:
Currently `zero_ext_op_for_imm32` does not return constants such as `0xFFFFFFFFFFFFFFFF` as suitable `imm32` immediates -- though they are for the (vast majority) of x64 instructions that sign extend immediates. Fixing this will require carefully auditing each function will calls this.

PR #2005, opened 2025-12-29
Audit `zero_ext_op_for_imm32`.

ltratt:
In x64, instructions of the form `op r64, imm32` sign-extend `imm32` to 64 bits. I must admit that I failed to fully appreciate the implications of this and that the `zero_ext_op_for_imm32` function was too conservative.

This commit relaxes `zero_ext_op_for_imm32` so that 64-bit zero-extended immediates that happen to also be safely representable as a sign-extended `i32` are let through.

Fixes https://github.com/ykjit/yk/issues/1933. [Note also that https://github.com/ykjit/yk/pull/1938 claimed it encoded a case that needed fixing. I don't think that's true.]


ltratt:
Oops. Fixed in the force push.

PR #2006, opened 2025-12-30
Deduplicate constants in traces.

ltratt:
On average, about 1/3 of instructions in a trace are constants, of which under half are unique. In particular it is common to see traces which look like this:

```
%77: i8 = 0
%78: i8 = 1
...
%82: i8 = 1
%83: i8 = 0
```

and so on. Constants aren't particularly expensive in of themselves, but they impose second order effects, in particular for parts of the compiler tool chain (including, but not only, optimisation passes) that record information per instruction or which have to do scans (e.g. the x64 backend seeing if it can squeeze loads and comparisons into one instruction). These costs are high enough that they can be clearly observed in profiles of the compiler.

This commit deduplicates constants in traces where possible. Floating point numbers mean that this can't do everything one might initially expect, but our aim here is not to have a semantic effect, but a performance effect: any deduplication we can do is welcome.

To make this work, we have to do two things. First, `HashableConst` is a wrapper around `Const` solely for deduplication purposes: it treats floating point numbers as bit patterns, so it can never deduplicate floating point numbers that might be considered equal / not-equal in different circumstances. Second, we have to make sure we never deduplicate arguments, which are a mix of `arg` and `const` instructions.

Overall, this gives roughly a 10% speed-up to the compiler. That is, currently, largely lost in the noise of benchmarks, except for the wonderfully-stable mandelbrot, where it gives a consistent 0.5% speed-up!


Fidget-Spinner:
Change looks clean enough. I do wonder though: by keeping constants around longer, we extend their live range. Would this be worse for the register allocator?

ltratt:
Squashed.

PR #2007, opened 2026-01-02
From snapshots to `VarLocs`

ltratt:
This PR gradually moves us to a state where we no longer snapshot the entire register allocator: this is both slow and also an API dead-end.

The good news is that snapshotting was never really necessary: we could always have used `VarLoc`s (and `VarLocs`), but I was too stupid to notice this. In that sense, this PR is simplifying something that should have been present from j2's birth -- better late than never!

Ultimately I'm doing this now because I need it for peeling. But this PR also makes possible something I've been wanting to do for ages: guard bodies (i.e. where we shove instructions only relevant to guards into the guard's body). There is quite a bit of further work that will be necessary for this, but pretty much everything in this PR is a necessary step for that. Hopefully I can get to that soon.

I wasn't particularly expecting to see performance improvements from this, but this PR improves havlak and deltablue by 2.8% each -- but makes fannkuchredux 1.5% slower. I'm not sure the latter result is truly significant, and either way, overall performance isn't really the aim of this PR.

### DIFF
--- a/ykrt/src/compile/j2/hir_to_asm.rs
+++ b/ykrt/src/compile/j2/hir_to_asm.rs
@@ -73,7 +73,7 @@ use crate::{
             codebuf::ExeCodeBuf,
             compiled_trace::{DeoptFrame, J2CompiledGuard, J2CompiledTrace, J2CompiledTraceKind},
             hir::*,
-            regalloc::{RegAlloc, RegFill, RegT, SnapshotIdx, VarLocs},
+            regalloc::{RegAlloc, RegFill, RegT, VarLoc, VarLocs},
         },
         jitc_yk::aot_ir::{self},
     },
@@ -134,7 +134,7 @@ impl<'a, AB: HirToAsmBackend> HirToAsm<'a, AB> {
                 self.be.coupler_trace_end(tgt_ctr)?;
 
                 // Assemble the body
-                let (guards, body_ra) = self.p_block(
+                let (guards, entry_stack_off) = self.p_block(
                     entry,
                     base_stack_off,
                     &entry_vlocs,
@@ -142,11 +142,10 @@ impl<'a, AB: HirToAsmBackend> HirToAsm<'a, AB> {
                     true,
                     logging,
                 )?;
-                let entry_stack_off = body_ra.stack_off();
                 let instrs_label = self
                     .be
                     .coupler_trace_start(entry_stack_off - base_stack_off)?;
-                self.asm_guards(entry, guards, body_ra)?;
+                self.asm_guards(entry, guards)?;
                 let (buf, guards, log, label_offs) = self.be.build_exe(logging, &[instrs_label])?;
                 let [sidetrace_off] = &*label_offs else {
                     panic!()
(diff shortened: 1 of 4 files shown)


Fidget-Spinner (commented):
Just two comments, overall this looks okay. Though I trust the tests more than I trust myself for this PR :).

ykrt/src/compile/j2/hir_to_asm.rs
```diff
@@ -344,12 +345,22 @@ impl<'a, AB: HirToAsmBackend> HirToAsm<'a, AB> {
...
                                 .map(|(iidx, (aot_op, smap_loc))| {
-                                    let fromvlocs = ra.varlocs_for_deopt(*iidx);
-                                    let mut tovlocs = AB::smp_to_vloc(smap_loc);
+                                    let fromvlocs = entry_vlocs[asmgrestore
+                                        .entry_vars
+                                        .iter()
+                                        .position(|x| x == iidx)
+                                        .unwrap()]
+                                    .iter()
+                                    // FIXME: We don't need to spill everything before deopt /
```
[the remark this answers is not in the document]
ltratt: Correct, it's an optimisation -- my guess is a not insignificant one (but not something I want to tackle yet). At the end of traces, and going into guards, we currently do a _lot_ of spills. The advantage is that it makes a lot of code easier, but it's definitely inefficient, because in most cases we fish the values back out immediately!

ykrt/src/compile/j2/regalloc.rs
```diff
@@ -95,34 +94,31 @@ impl<'a, AB: HirToAsmBackend> RegAlloc<'a, AB> {
...
     /// Snapshot this register allocator in a way that is suitable for a
     /// [super::hir::GuardRestore].
-    pub(super) fn snapshot(&mut self, iidx: InstIdx) -> SnapshotIdx {
-        self.snapshots.push(Snapshot {
-            istates: self.istates[..iidx].to_vec(),
-            rstates: self.rstates.clone(),
-        })
+    pub(super) fn vlocs_from_iidxs(&mut self, iidxs: &[InstIdx]) -> Vec<VarLocs<AB::Reg>> {
+        let mut out = Vec::with_capacity(iidxs.len());
+        for iidx in iidxs {
```
Fidget-Spinner: Would it save some memory if you only did this for guards? Rather than doing it for everything at the moment.

Fidget-Spinner (commented):
One thing I find a little tricky/subtle is that because we are backwards codegening, the deopt state should be n'th instruction/guard instead of the usual n+1th?

Basically the question boils down to:
```
INST1
<----- STATE1
GUARD1
<----- STATE2
INST2
```
We need `STATE1`, but the backwards codegen tracks `STATE2` (if i understand it correctly). How do you reconcile that?

ykrt/src/compile/j2/regalloc.rs
ltratt: This function is only called for guards, so your wish is granted :)

Fidget-Spinner:
> > We need STATE1, but the backwards codegen tracks STATE2 (if i understand it correctly). How do you reconcile that?
> 
> We generate the guard's code so that we get STATE1. You can see that around line 604ish in hir_to_asm:

That is truly a really subtle one!

ltratt:
Squashed.

PR #2009, opened 2026-01-05
Return traces

ltratt:
This PR introduces "return traces": that is a trace where the main loop is terminated by a `return` statement. We previously encoded these (oddly!) as loop traces, even though they can't loop. This PR introduces a new trace kind: return traces. This is a useful simplification in its own right, but will prove particularly useful for things like peeling, where previously we had to guess whether a trace was a genuine loop (or not).

### DIFF
--- a/ykrt/src/compile/j2/hir_to_asm.rs
+++ b/ykrt/src/compile/j2/hir_to_asm.rs
@@ -244,7 +244,7 @@ impl<'a, AB: HirToAsmBackend> HirToAsm<'a, AB> {
                     .map(|smap| AB::smp_to_vloc(smap, RegFill::Undefined))
                     .collect::<Vec<_>>();
 
-                let post_stack_label = self.be.return_trace_end(exit_safepoint)?;
+                self.be.return_trace_end(exit_safepoint)?;
 
                 // Assemble the body
                 let (guards, entry_stack_off) = self.p_block(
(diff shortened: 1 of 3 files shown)


ykrt/src/compile/j2/regalloc.rs
```diff
@@ -2135,7 +2136,18 @@ pub(crate) mod test {
...
         }
 
-        fn loop_trace_start(&mut self, _iter0_label: Self::Label, _stack_off: u32) {}
+        fn loop_trace_start(&mut self, _post_stack_label: Self::Label, _stack_off: u32) {}
+
+        fn return_trace_end(
+            &mut self,
+            _exit_safepoint: &'static DeoptSafepoint,
+        ) -> Result<(), CompilationError> {
+            todo!()
```
Fidget-Spinner: I'm confused: why do we not need to implement this yet?

ltratt: This is test code: the register allocator provides a "dummy" backend that only implements the minimum amount of stuff necessary for its tests.

ykrt/src/compile/j2/hir_to_asm.rs
```diff
@@ -215,6 +215,61 @@ impl<'a, AB: HirToAsmBackend> HirToAsm<'a, AB> {
                 };
                 (buf, guards, log, modkind)
             }
+            ModKind::Return {
+                entry_safepoint,
+                entry,
+                exit_safepoint,
+            } => {
```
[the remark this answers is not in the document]
ltratt: Yes, this part is notably messy, because I didn't understand what I was doing when I started, and I'm only just wrapping my head around it now. That said, it might be that what we end up doing is packaging _some_ functionality up and sharing it, but probably most of it will be subtly different. The different trace kinds are becoming _more_ different over time (rather than the other way around), which is good!

[I would like to de-x64ify this soon. It's on my todo list, but there's quite a bit between here and there to do first.]

PR #2016, opened 2026-01-06
Implement known bits bitshifts

Fidget-Spinner:
Nothing surprising for this one, thankfully.

### DIFF
--- a/ykrt/src/compile/j2/opt/known_bits.rs
+++ b/ykrt/src/compile/j2/opt/known_bits.rs
@@ -30,11 +30,14 @@ impl PassT for KnownBits {
     fn feed(&mut self, opt: &mut PassOpt, inst: Inst) -> OptOutcome {
         self.pending_commit = None;
         match inst {
+            Inst::AShr(x) => self.opt_ashr(opt, x),
             Inst::And(x) => self.opt_and(opt, x),
             Inst::Const(x) => self.opt_const(x),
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/j2/opt/known_bits.rs
```diff
@@ -72,6 +75,19 @@ impl KnownBits {
         self.pending_commit = Some(bits);
     }
 
+    fn opt_ashr(&mut self, opt: &PassOpt, inst: AShr) -> OptOutcome {
+        let AShr { tyidx: _, lhs, rhs, exact: _ } = inst;
+        if let Some(lhs_b) = self.as_knownbits(opt, lhs)
+            && let Some(rhs_b) = self.as_knownbits(opt, rhs)
+            && rhs_b.all_known()
```
ltratt: Isn't this the same as saying "if RHS is a const"?

Fidget-Spinner: Yes it is. We don't have a convenience function to check both at the same time, and I don't feel like duplicating functionality as it's only used in two places.

Fidget-Spinner: Done in 6a5f653d3832357241ec606888d35d37c82c7f9a

Fidget-Spinner:
Done.

PR #2019, opened 2026-01-07
Unpin rustc.

ltratt:
The temporary error we experienced a few months back no longer happens, so we can go back to using nightly.

If/once this has merged, https://github.com/ykjit/yk/pull/2018 should then hopefully pass.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -210,12 +210,15 @@ for tracer in $TRACERS; do
     export YKB_TRACER="${tracer}"
     cargo build
     ASAN_SYMBOLIZER_PATH="${YKLLVM_BIN_DIR}/llvm-symbolizer" \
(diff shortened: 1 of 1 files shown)


Fidget-Spinner:
Hah it's now complaining it's an unnecessary unsafe block. I think you need to merge the other PR into this one as a separate commit to get things working.

ltratt:
OK, that's a very interesting possible-threading error, which I'll investigate. I don't think it's related to this PR, as such, so we should probably retry this, and I'll look at the error in parallel.

ltratt:
It looks like we might need https://github.com/ykjit/yk/pull/2008 to be merged before this one will work. [The error is a bit odd, but it seems that hwtracer's doc tests in some way are a problem? Still, since we're about to unlink that, I don't think it matters too much.]

PR #2024, opened 2026-01-11
If near-`mmap` is not possible, `mmap` something.

ltratt:
We try and `mmap` "nearby" to allow near jumps, but sometimes that doesn't work (e.g. it almost never works under valgrind). This commit implements a generic "mmap something" fallback. This is enough for `valgrind --tool=memcheck` to work.

### DIFF
--- a/ykrt/src/compile/j2/mod.rs
+++ b/ykrt/src/compile/j2/mod.rs
@@ -119,27 +119,25 @@ impl J2 {
                     0,
                 ) as *const c_void
             };
-            if buf == MAP_FAILED {
-                todo!();
-            }
+            if buf != MAP_FAILED {
+                let is_near = if !hint_ptr.is_null() {
+                    #[cfg(target_arch = "x86_64")]
+                    let is_near = unsafe { hint_ptr.offset_from(buf) }.abs() < 0x80000000;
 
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/j2/mod.rs
```diff
@@ -119,27 +119,25 @@ impl J2 {
                     0,
                 ) as *const c_void
             };
-            if buf == MAP_FAILED {
-                todo!();
-            }
+            if buf != MAP_FAILED {
+                let is_near = if !hint_ptr.is_null() {
+                    #[cfg(target_arch = "x86_64")]
+                    let is_near = unsafe { hint_ptr.offset_from(buf) }.abs() < 0x80000000;
```
vext01: I know it was pre-existing, but there's a magic number here that we could define as a constant for comprehension purposes.

vext01: Isn't it just the threshold after which we consider an address "far"? Something like `CLOSE_THRESHOLD`?

ltratt: Oh, that constant -- sorry! IMHO this should be shoved in a backend. Again, for the future, I think.

PR #2042, opened 2026-01-20
Make known bits analysis the last pass

Fidget-Spinner:
We've come to realise known bits benefits more from equivalences + canonicalisation of previous passes, such as in strength fold and CSE.

This PR just moves the known bits pass to the last one.


Fidget-Spinner:
> Does this make any difference?

Not (yet) for our benchmarks I suspect. I can always defer this decision to later.

ltratt:
Closing for now. I suspect we will want to do this, but I need to benchmark and think about a few other things before having a sense of what the best course of action is.

PR #2057, opened 2026-01-28
Guard bodies

ltratt:
This PR implements full-fat guard bodies: finally we can move code into guard bodies! There is quite a bit of warm-up work, because the approach this PR takes is quite different from our initial support: rather than having the optimiser create guard bodies, this PR has hir_to_asm do so. So while the main commit is https://github.com/ykjit/yk/commit/23a26346a3580cfc0e5ec7747e60bb8926c234cf, the preceding four commits are vital to get us to the point where that is feasible.

Note: this PR looks scary, and in a sense it is barely reviewable (as with all big PRs), but really the core "algorithm" that's worth looking at is the extensively documented modifications to hir_to_asm in https://github.com/ykjit/yk/commit/23a26346a3580cfc0e5ec7747e60bb8926c234cf. That PR is way over 50% testing code: the core algorithm is under 100LoC!

In terms of benchmarks, this speeds up about 40% of yklua benchmarks (and has no ill effect on any) in the range ~2-16%:

```
 bigloop/yklua/1000000000   31056 ± 65   25938 ±  6   0.84  16.48% faster
 queens/yklua/1000           5375 ±295    4586 ± 97   0.85  14.68% faster
 sieve/yklua/3000            4616 ± 52    4373 ±  7   0.95  5.27% faster
 fannkuchredux/yklua/10     12675 ± 18   12037 ± 17   0.95  5.03% faster
 LuLPeg/yklua/              13296 ± 53   12667 ± 49   0.95  4.74% faster
 havlak/yklua/1500          77208 ±440   73937 ±893   0.96  4.24% faster
 Heightmap/yklua/2000        7937 ± 39    7651 ±110   0.96  3.60% faster
 binarytrees/yklua/15       11660 ± 79   11303 ± 99   0.97  3.06% faster
 deltablue/yklua/12000       8822 ± 26    8611 ±130   0.98  2.40% faster
 mandelbrot/yklua/500        2193 ±  1    2162 ±  3   0.99  1.44% faster
```

From memory, when we did this in jitc_yk, the difference was more like 2-3%. My guess that two factors explain the relatively bigger speedup we see here: (mostly) jitc_yk could only copy one level of instructions, whereas this PR can recurse arbitrarily far into side-effect-free instructions; (but also) guard bodies don't slow the JIT compiler down as much as one might expect (roughly by about 10-20%) and, almost certainly, by less than in jitc_yk.

### DIFF
--- a/ykrt/src/compile/j2/aot_to_hir.rs
+++ b/ykrt/src/compile/j2/aot_to_hir.rs
@@ -34,6 +34,8 @@ use crate::{
     mt::{MT, TraceId},
     trace::TraceAction,
 };
+#[cfg(test)]
+use index_vec::IndexVec;
 use parking_lot::Mutex;
 use smallvec::{SmallVec, smallvec};
 use std::{assert_matches, collections::HashMap, iter::Peekable, marker::PhantomData, sync::Arc};
@@ -305,6 +307,8 @@ impl<Reg: RegT + 'static> AotToHir<Reg> {
             tyidx_ptr0,
             tyidx_void,
             addr_name_map: self.addr_name_map,
+            #[cfg(test)]
+            smaps: IndexVec::new(),
         };
 
         let ds = if let Some(x) = &self.hl.lock().debug_str {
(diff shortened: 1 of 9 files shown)


vext01:
Thanks. My superficial review of this says merge :)

PR #2064, opened 2026-02-04
Don't mix bits and bytes.

ltratt:
When porting load_store over from jitc_yk, I mangled the bits/bytes conversion, fortunately in a way which over-approximated soundness. In other words, load_store in j2 was less effective than it should be (e.g. the new test case did not have the final load removed).

This commit fixes that. Interestingly -- and I don't think this happened in jitc_yk -- we then see cases of type punning occuring in yklua. Fixing that properly is worthwhile, but a job for the future.


ltratt:
Fixed the test (which was, in retrospect, a bit weird before) in the force push: it got optimised a bit further (correctly) by this PR!

vext01:
Did your force push succeed? I don't see one.

ltratt:
This time I actually force pushed (mea culpa).

PR #2171, opened 2026-05-19
Pop a tstate when a trace `ret`s.

ltratt:
`tstate`s track the state of execution: if we're interpreting, executing etc. We maintain a stack, which, it turns out, we (I!) didn't maintain properly: if a trace ended in `ret`, we didn't pop the associated `Executing` state.


ltratt:
I should have added: this has no performance impact on yklua.

PR #2175, opened 2026-06-07
Add i1 support to x64 load and store codegen.

Pavel-Durov:
I noticed this unimplemented case while working on [yksompp](https://github.com/ykjit/yksompp/).

Example from logs:
```shell
thread '<unnamed>' (3133972) panicked at ykrt/src/compile/j2/x64/x64hir_to_asm.rs:873:22:
not yet implemented: 1

thread '<unnamed>' (3133970) panicked at ykrt/src/job_queue.rs:129:44:
called `Result::unwrap()` on an `Err` value: Any { .. }
fatal runtime error: failed to initiate panic, error 5, aborting
```

### DIFF
--- a/ykrt/src/compile/j2/x64/x64hir_to_asm.rs
+++ b/ykrt/src/compile/j2/x64/x64hir_to_asm.rs
@@ -668,7 +668,7 @@ impl<'a> X64HirToAsm<'a> {
 
         self.asm.push_inst(match self.m.ty(*tyidx) {
             Ty::Int(bitw) => match bitw {
-                8 => {
+                1 | 8 => {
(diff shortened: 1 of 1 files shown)


ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -6949,6 +6949,22 @@ mod test {
 
     #[test]
     fn cg_load_int() {
+        // i1
```
ltratt, line 6952: This is only testing 1 of the 4 `match`es above (and I'm not even sure which one).

PR #2182, opened 2026-06-15
Only log `aot` IR once.

ltratt:
In one way this is a hack, but having one `aot` entry per trace created is bonkers (and deeply confusing).


Pavel-Durov:
Agree, I find it very confusing as well!

PR #2219, opened 2026-07-29
Support struct-by-value returns from direct calls in j2

Pavel-Durov:
A small struct return is packed into RAX:RDX only across a real, non-inlined call - LLVM IR never encodes that packing itself. j2 now handles both representations:

1. register-packed - reading field sstraight from RAX/RDX for a real call,
2. memory-backed - `ptr_add`+`load` path for an inlined calls.

### DIFF
--- a/tests/c/extract_struct.c
+++ b/tests/c/extract_struct.c
@@ -1,4 +1,3 @@
-// ignore-if: true # not yet implemented in j2
 // Compiler:
 //   env-var: YKB_EXTRA_CC_FLAGS=-O0
 // Run-time:
@@ -7,8 +6,8 @@
 //   env-var: YKD_LOG=4
 //   stderr:
 //     yk-tracing: start-tracing
-//     2
-//     999
+//     s1.a: 4, s2.a: 5
+//     s1.b: 999, s2.b: 999
 //     yk-tracing: stop-tracing
 //     --- Begin aot ---
 //     ...
@@ -18,25 +17,33 @@
 //     ...
 //     %{{11_5}}: i64 = extractvalue %{{10_1}}, [1]
 //     ...
+//     %{{20_1}}: {0: i8, 64: i64} = call make_struct_inlined()...
+//     ...
+//     %{{21_2}}: i8 = extractvalue %{{20_1}}, [0]
+//     ...
+//     %{{21_5}}: i64 = extractvalue %{{20_1}}, [1]
+//     ...
 //     --- End aot ---
 //     --- Begin hir ---
 //     ...
 //     --- End hir ---
-//     2
-//     999
-//     yk-execution: enter-jit-code
-//     2
-//     999
-//     2
-//     999
-//     yk-execution: deoptimise ...
+//     s1.a: 3, s2.a: 4
+//     s1.b: 999, s2.b: 999
+//     yk-execution: enter-jit-code {"trid": "0"}
+//     s1.a: 2, s2.a: 3
+//     s1.b: 999, s2.b: 999
+//     s1.a: 1, s2.a: 2
+//     s1.b: 999, s2.b: 999
+//     yk-execution: deoptimise {"trid": "0", "gidx": "0"}
 //     exit
 
-// Test that trace builder handles loading structs by rewriting `extractvalue`
-// instructions into `ptr_add`s and `load`s. We only check that the AOT IR
-// contains a struct load and check that the trace does the correct thing.
-// Matching on the JIT IR is difficult since the `ptr_add`s and `load`s are
-// not easily distinguished from unrelated `ptr_add`s and `load`s.
+// Test that `extractvalue` on a call struct return works for both representations the trace 
+// builder can produce: 
+// 1. register-packed (RAX:RDX, from a real non-inlined `call`)
+// 2. memory-backed (`ptr_add` + `load`, from an inlined callee's own struct-typed `load`. 
+// We only check that the AOT IR contains the expected `call`/`extractvalue`
+// shape and that the trace does the correct thing at runtime; matching on the JIT IR itself is
+// difficult since its `ptr_add`s and `load`s aren't easily distinguished from unrelated ones.
 
 #include <assert.h>
 #include <stdint.h>
@@ -51,7 +58,17 @@ struct S {
   uint64_t b;
 };
 
+// `yk_outline` stops the trace-builder inlining this call, so it stays a real, non-inlined
+// `call` in the trace. That's a genuine machine-call boundary, so the SysV ABI packs this
+// struct's return into RAX:RDX.
+__attribute__((yk_outline))
 struct S make_struct() {
+  struct S ret = {1, 999};
+  return ret;
+}
+
+// This call is inlined, so struct return value is always set as real memory.
+struct S make_struct_inlined() {
   struct S ret = {1, 999};
   return ret;
 }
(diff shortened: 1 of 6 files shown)


ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -1766,6 +1768,27 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
         Ok(())
     }
 
+    fn i_alloca(
```
ltratt: `alloca` isn't safe in our context, at least on the normal stack. At the very least it would have to go on the shadow stack but, so far, we've just banned it entirely.

Pavel-Durov: got it

ltratt: A question that I can't quite answer is whether this is _only_ intended for struct returns? If so, that would be safe, but then it's an open question as to whether we want to expose a general purpose instruction (that isn't safe in the general case) when we only need it for a specific solution.

Pavel-Durov:
@ltratt Ready for another review. 
I force pushed my changes - removing allocas completely, strictly focusing on direct calls returning structs.
No ykllvm changes needed.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Resolving this comment - allocas are removed and the code changed completely.

tests/c/extract_struct.c
```diff
@@ -18,25 +17,33 @@
...
 //     exit
 
-// Test that trace builder handles loading structs by rewriting `extractvalue`
-// instructions into `ptr_add`s and `load`s. We only check that the AOT IR
-// contains a struct load and check that the trace does the correct thing.
-// Matching on the JIT IR is difficult since the `ptr_add`s and `load`s are
-// not easily distinguished from unrelated `ptr_add`s and `load`s.
+// Test that `extractvalue` on a call struct return works for both representations the trace 
+// builder can produce: 
+// 1. register-packed (RAX:RDX, from a real non-inlined `call`)
```
ltratt, line 42: We need to be clear that this test file _isn't_ x64/sysv specific, but on that combination it does happen to test certain things.

tests/c/extract_struct.c
```diff
@@ -51,11 +58,21 @@ struct S {
   uint64_t b;
 };
 
+// `yk_outline` stops the trace-builder inlining this call, so it stays a real, non-inlined
```
ltratt, line 61: This comment needs removing (I can see the LLM explaining itself!).

tests/c/extract_struct.c
```diff
@@ -51,11 +58,21 @@ struct S {
...
+// `call` in the trace. That's a genuine machine-call boundary, so the SysV ABI packs this
+// struct's return into RAX:RDX.
+__attribute__((yk_outline))
 struct S make_struct() {
   struct S ret = {1, 999};
   return ret;
 }
 
+// This call is inlined, so struct return value is always set as real memory.
+struct S make_struct_inlined() {
```
ltratt, line 71: This call is not guaranteed to be inlined, so the comment is wrong. You can strongly encourage it to be inlined with an `__attribute__` though.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -892,35 +891,39 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
-                vlocs.push(VarLoc::Reg(Reg::from_dwarf_reg(*dwarf_reg), reg_fill));
-                for x in extras {
-                    if *x >= 0 {
-                        vlocs.push(VarLoc::Reg(
-                            Reg::from_dwarf_reg(x.cast_unsigned()),
-                            reg_fill,
-                        ));
-                    } else {
-                        vlocs.push(VarLoc::Stack(u32::from(x.unsigned_abs())));
+        // A register-packed struct return spanning >1 register/stack slot can surface as
```
ltratt, line 894: Is this true? Can you give a concrete example from real code?

ltratt:
I have two over-arching comments before I dive into further detail:

1. I _think_ the PR is doing two things: `return`ing `struct`s; and the "inlined" `struct` case. The latter seems much easier to me (correct me if I'm wrong!) and could be done in a PR first. That would break this PR up nicely.
2. The "call low/high" thing is not as neat as I'd like :) It's also, I think, encoding x86/sysv-ness at the aot_to_hir level, which we mustn't do. Instead I suggest having an `extractvalue` instruction in HIR, which more-or-less mirrors what LLVM does. Then `x64hir_to_asm.rs` can identify these and handle them in whatever way it wants.

Please see whether these seem sensible or not: bear in mind that I might be wrong!

With regards to (2), the "trick" is related to backwards code gen: I think x64hir_to_asm can say something like this:

```rust
fn p_extractvalue(iidx: InstIdx...) {
  if iidx - 1 == Inst::Call && <this extractvalue refers to that call> {
    // don't do anything
  } else if iidx - 1 == Inst::ExtractValue && iidx - 2 == Inst:Call && <both extract values refer to that call> {
    let [rax, rdx] = ra.alloc(..); // handle the RAX and RDX extraction in one go
  } else {
    // some sequence of extractvalues/calls we don't handle
    todo!();
  }
}
```

This would avoid the "call high/low" thing entirely. Again, bear in mind that I might be very wrong about this! It normally takes me several attempts before I find a nice design for this sort of thing.

tests/c/extract_struct.c
Pavel-Durov: is there a yk_inline attribute for it?

Pavel-Durov:
> I have two over-arching comments before I dive into further detail:
> 
>     1. I _think_ the PR is doing two things: `return`ing `struct`s; and the "inlined" `struct` case. The latter seems much easier to me (correct me if I'm wrong!) and could be done in a PR first. That would break this PR up nicely.
> 
>     2. The "call low/high" thing is not as neat as I'd like :) It's also, I think, encoding x86/sysv-ness at the aot_to_hir level, which we mustn't do. Instead I suggest having an `extractvalue` instruction in HIR, which more-or-less mirrors what LLVM does. Then `x64hir_to_asm.rs` can identify these and handle them in whatever way it wants.
> 
> 
> Please see whether these seem sensible or not: bear in mind that I might be wrong!
> 
> With regards to (2), the "trick" is related to backwards code gen: I think x64hir_to_asm can say something like this:
> 
> ```rust
> fn p_extractvalue(iidx: InstIdx...) {
>   if iidx - 1 == Inst::Call && <this extractvalue refers to that call> {
>     // don't do anything
>   } else if iidx - 1 == Inst::ExtractValue && iidx - 2 == Inst:Call && <both extract values refer to that call> {
>     let [rax, rdx] = ra.alloc(..); // handle the RAX and RDX extraction in one go
>   } else {
>     // some sequence of extractvalues/calls we don't handle
>     todo!();
>   }
> }
> ```
> 
> This would avoid the "call high/low" thing entirely. Again, bear in mind that I might be very wrong about this! It normally takes me several attempts before I find a nice design for this sort of thing.

Submitted separate PR for 1 👉  https://github.com/ykjit/yk/pull/2221

tests/c/extract_struct.c
ltratt: No, just the normal `inline` attribute clang supports (but note this is only ever a hint, not a guarantee!).

PR #2221, opened 2026-07-31
Support extractvalue for memory-backed structs in j2.

Pavel-Durov:
Add handler for `extractvalue` in `j2` on a struct operand as `ptr_add`+`load` pair. 
This covers the memory-backed case, where the struct operand is a pointer to it.

First part of the bigger PR https://github.com/ykjit/yk/pull/2219

### DIFF
--- a/tests/c/extract_struct.c
+++ b/tests/c/extract_struct.c
@@ -1,4 +1,3 @@
-// ignore-if: true # not yet implemented in j2
 // Compiler:
 //   env-var: YKB_EXTRA_CC_FLAGS=-O0
 // Run-time:
@@ -7,36 +6,50 @@
 //   env-var: YKD_LOG=4
 //   stderr:
 //     yk-tracing: start-tracing
-//     2
+//     1
 //     999
 //     yk-tracing: stop-tracing
 //     --- Begin aot ---
 //     ...
-//     %{{10_1}}: {0: i8, 64: i64} = call make_struct()...
+//     %{{s}}: {0: i8, 64: i64} = load %{{_}}
 //     ...
-//     %{{11_2}}: i8 = extractvalue %{{10_1}}, [0]
+//     %{{_}}: i8 = extractvalue %{{s}}, [0]
 //     ...
-//     %{{11_5}}: i64 = extractvalue %{{10_1}}, [1]
+//     %{{_}}: i64 = extractvalue %{{s}}, [1]
 //     ...
 //     --- End aot ---
 //     --- Begin hir ---
 //     ...
+//     %{{a}}: i8 = load %{{ptr}}
+//     store %{{a}}, %{{s1}}
+//     %{{s1b}}: ptr = ptradd %{{s1}}, 8
+//     %{{padd}}: ptr = ptradd %{{ptr}}, 8
+//     %{{b}}: i64 = load %{{padd}}
+//     store %{{b}}, %{{s1b}}
+//     ...
+//     %{{a_ext}}: i32 = zext %{{a}}
+//     ...
+//     %{{_}}: i32 = call %{{_}}(%{{_}}, %{{_}}, %{{a_ext}}) ; @fprintf
+//     ...
+//     %{{b_load}}: i64 = load %{{s1b}}
+//     ...
+//     %{{_}}: i32 = call %{{_}}(%{{_}}, %{{_}}, %{{b_load}}) ; @fprintf
+//     ...
 //     --- End hir ---
-//     2
+//     1
 //     999
-//     yk-execution: enter-jit-code
-//     2
+//     yk-execution: enter-jit-code {"trid": "0"}
+//     1
 //     999
-//     2
+//     1
 //     999
 //     yk-execution: deoptimise ...
 //     exit
 
-// Test that trace builder handles loading structs by rewriting `extractvalue`
-// instructions into `ptr_add`s and `load`s. We only check that the AOT IR
-// contains a struct load and check that the trace does the correct thing.
-// Matching on the JIT IR is difficult since the `ptr_add`s and `load`s are
-// not easily distinguished from unrelated `ptr_add`s and `load`s.
+// Test that the trace builder handles loading structs by rewriting
+// `extractvalue` instructions into `ptradd`s and `load`s. Both fields are
+// followed from their `load` to the `fprintf` that prints them, so that we
+// know we haven't matched unrelated `ptradd`s/`load`s.
 
 #include <assert.h>
 #include <stdint.h>
@@ -51,6 +64,7 @@ struct S {
   uint64_t b;
 };
 
+__attribute__((always_inline))
 struct S make_struct() {
   struct S ret = {1, 999};
   return ret;
@@ -69,7 +83,6 @@ void interp(){
   while (i > 0) {
     yk_mt_control_point(mt, &loc);
     struct S s1 = make_struct();
-    s1.a = 2;
     fprintf(stderr, "%d\n", s1.a);
     fprintf(stderr, "%ld\n", s1.b);
     i--;
--- a/ykrt/src/compile/j2/aot_to_hir.rs
+++ b/ykrt/src/compile/j2/aot_to_hir.rs
@@ -25,7 +25,7 @@ use crate::{
             compiled_trace::{
                 CompiledGuardIdx, DeoptFrame, DeoptVar, J2CompiledTrace, J2TraceStart,
             },
-            hir,
+            hir::{self, InstT},
             opt::{OptT, fullopt::FullOpt, noopt::NoOpt},
             regalloc::{RegT, VarLoc, VarLocs},
         },
@@ -915,7 +915,7 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
                 Inst::Cast { .. } => self.p_cast(pc.clone(), inst)?,
                 Inst::CondBr { .. } => self.p_condbr(pc.clone(), bid, inst)?,
                 Inst::DebugStr { .. } => self.p_debugstr(pc.clone())?,
-                Inst::ExtractValue { .. } => todo!(),
+                Inst::ExtractValue { .. } => self.p_extractvalue(pc.clone(), inst)?,
                 Inst::FCmp { .. } => self.p_fcmp(pc.clone(), inst)?,
                 Inst::FNeg { .. } => self.p_fneg(pc.clone(), inst)?,
                 Inst::Freeze { .. } => self.p_freeze(pc.clone(), inst)?,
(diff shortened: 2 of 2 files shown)


tests/c/extract_struct.c
```diff
@@ -51,6 +50,7 @@ struct S {
   uint64_t b;
 };
 
+
```
ltratt: Extraneous blank line.

tests/c/extract_struct.c
ltratt, line 22: We need to check at least _something_ in the HIR: presumably a `load` and `ptradd`?

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1816,6 +1823,46 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
         .map(|_| ())
     }
 
+    /// Read a struct field via `extractvalue` supporting only memory-backed as `ptr_add`+`load`.
+    fn p_extractvalue(&mut self, iid: InstId, inst: &Inst) -> Result<(), CompilationError> {
+        let Inst::ExtractValue { tyidx, op, indices } = inst else {
+            panic!("extractvalue must have a type, an operand and indices.")
```
ltratt: We don't use strings in `panic`s like this in j2 (same the `panic` a few lines later).

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1816,6 +1823,46 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+        };
+
+        let Ty::Struct(struct_ty) = op.type_(self.am) else {
+            panic!("extractvalue's operand must be a struct.")
+        };
+
+        assert_eq!(indices.len(), 1, "extractvalue with nested indices");
+        let field_bit_off = struct_ty.field_bit_offs()[indices[0]];
+        let byte_off = field_bit_off / 8;
```
ltratt, line 1842: What happens if `field_bit_off` is not a multiple of 8? I can imagine this happening with bit fields.

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1816,6 +1823,46 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+
+        let mut ptr = self.p_operand(op)?;
+        if byte_off != 0 {
```
ltratt: `byte_off > 0` is clearer here IMHO. [Technically this optimisation will be discovered by strength_fold, but I agree that doing it here is easy enough that the (small) perf win is worth it.]

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1816,6 +1823,46 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+                }
+                .into(),
+            )?;
+        }
+
+        let res_tyidx = self.p_ty(self.am.type_(*tyidx))?;
+        self.push_inst_and_link_local(
+            iid,
+            hir::Load {
+                tyidx: res_tyidx,
```
ltratt, line 1866: Unless it's necessary for the borrow checker, inline `res_tyidx` here.

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1816,6 +1823,46 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
         .map(|_| ())
     }
 
+    /// Read a struct field via `extractvalue` supporting only memory-backed as `ptr_add`+`load`.
```
ltratt: I can't parse this sentence.

tests/c/extract_struct.c
Pavel-Durov: removed ce63f2d7593bd94e193ec4ba551ceb7ce7ea9b61

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: Removed 161fc36ba981e3d9034ae4508ea36edd68b66f75

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: I'll add assert

ltratt: I'd also like us to at least think about the question, even if the answer is "not sure, so an `assert` is safest".

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: updated 8fe04867d72c964b4f8b4a249b665aeb5f50357a

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: that's my broken english, will update

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: Required by the borrow checker, error I get:
```
error[E0499]: cannot borrow `*self` as mutable more than once at a time
    --> ykrt/src/compile/j2/aot_to_hir.rs:1861:24
     |
1858 |         self.push_inst_and_link_local(
     |         ---- ------------------------ first borrow later used by call
     |         |
     |         first mutable borrow occurs here
...
1861 |                 tyidx: self.p_ty(self.am.type_(*tyidx))?,
     |                        ^^^^ second mutable borrow occurs here
     |
help: try adding a local storing this argument...
    --> ykrt/src/compile/j2/aot_to_hir.rs:1861:24
     |
1861 |                 tyidx: self.p_ty(self.am.type_(*tyidx))?,
     |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: ...and then using that local as the argument to this call
    --> ykrt/src/compile/j2/aot_to_hir.rs:1858:9
     |
1858 | /         self.push_inst_and_link_local(
1859 | |             iid,
1860 | |             hir::Load {
1861 | |                 tyidx: self.p_ty(self.am.type_(*tyidx))?,
...    |
1864 | |             },
1865 | |         )
     | |_________^
```

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: Removed the comment cause not of the other p_* methods have comments.
e32cdb0eba1202f4b1da298be95b34ab121acc7c

tests/c/extract_struct.c
Pavel-Durov: Added f8bc65b421a0f1d016edd3e03a178028412f59eb

ltratt: The problem with that is that it's so generic that nearly any old HIR generated from C will match it. The typical way we solve this is to e.g. pass the output to `printf`, then we know we haven't matched a random ptradd/load.

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: Added assert 870f5be33fe56b2ab90d813eefc5e11334d1a20d

Pavel-Durov: `field_bit_offs` comes from [YkIRWriter.cpp:2058](https://github.com/ykjit/ykllvm/blob/main/llvm/lib/YkIR/YkIRWriter.cpp#L2058C33-L2058C55) 

Which calls `getElementOffsetInBits` which is:
```cpp
  TypeSize getElementOffsetInBits(unsigned Idx) const {
    return getElementOffset(Idx) * 8;
  }
```
in [DataLayout.h:760](https://github.com/ykjit/ykllvm/blob/main/llvm/include/llvm/IR/DataLayout.h#L760) 

So every entry is a multiple of 8.

ltratt: OK, I tend to agree.

tests/c/extract_struct.c
Pavel-Durov: Added ce28bb9ffcf5c7bd4e0a853d2319d51d944946e3

Pavel-Durov: Its a bit weird cause `s.a` is being assigned to const 2:
```
//     %{{a2}}: i32 = 2
//     ...
//     %{{_}}: i32 = call %{{_}}(%{{_}}, %{{_}}, %{{a2}}) ; @fprintf
```
but I think it shows the load and args to printf.

ltratt: I think this test might be better off without `s1.a = 2;`?

Pavel-Durov: ok. I didn't want to change the original test cause I thought it was there for a reason - to see that the struct fields are updated

ltratt: The test is outdated, so we should make it be the best test for _now_ (which might mean keeping it as-is or changing it). I'm not sure that the `=2` is helpful, because it means we can't check the old value. FWIW, I would store a weirded bit pattern than `1` in there, because that will be a better check against corruption / accidentally-seeing-the-right-value.

Pavel-Durov: Removed f038871a5d1f0f61df919b1abd275ffa424bd702

Pavel-Durov: sorry pushed the wrong thing

Pavel-Durov: updated eee59acdce4df41b494cb6b3820c78ba56f588fe

Pavel-Durov: Added allways_inline attribute 
5b3ec6152919a87ef282aea9ebcc26258b76ff01

tests/c/extract_struct.c
```diff
@@ -7,36 +6,50 @@
...
 //     yk-tracing: start-tracing
-//     2
+//     1
 //     999
 //     yk-tracing: stop-tracing
 //     --- Begin aot ---
 //     ...
-//     %{{10_1}}: {0: i8, 64: i64} = call make_struct()...
+//     %{{s}}: {0: i8, 64: i64} = load %{{_}}
 //     ...
```
ltratt, line 15: Out of curiosity, do the `extractvalue`s appear some distance after the `load`?

Pavel-Durov: There's `ptr_add` which is overriden with the result of `extractvalue`:
Here's the related basic block:
```
bb10:
    %10_0: {0: i8, 64: i64} = load %0_5
    # extract_struct.c:85: struct S s1 = make_struct();
    %10_1: ptr = ptr_add %0_8, 0
    %10_2: i8 = extractvalue %10_0, [0]
    *%10_1 = %10_2
    %10_4: ptr = ptr_add %0_8, 8
    %10_5: i64 = extractvalue %10_0, [1]
    *%10_4 = %10_5
    # extract_struct.c:86: fprintf(stderr, "%d\n", s1.a);
    %10_7: ptr = load @stderr
    %10_8: ptr = ptr_add %0_8, 0
    %10_9: i8 = load %10_8
    %10_10: i32 = zext %10_9, i32
    %10_11: i32 = call fprintf(%10_7, @.str, %10_10)
    # extract_struct.c:87: fprintf(stderr, "%ld\n", s1.b);
    br bb11
```

ltratt: That's a "no" then :)

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1816,6 +1823,47 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+            )?;
+        }
+
+        let res_tyidx = self.p_ty(self.am.type_(*tyidx))?;
+        self.push_inst_and_link_local(
+            iid,
+            hir::Load {
+                tyidx: res_tyidx,
+                ptr,
+                is_volatile: false,
```
ltratt: After a bit of looking into this, I think this line isn't quite right. Iff `ptr` is a volatile load, then this should be a volatile load too.

This embeds a deeper assumption: that `ptr` really is a pointer! I don't think that's true in general, so we probably need to add an assertion around line 1841 along the lines of `assert_eq!(ptr.ty(), Ty::Ptr(0))`.

Pavel-Durov: ok

Pavel-Durov: Added assert 0f3a16b49bec188a0db4a163c99bce2b13fb8090

ltratt: We still need the `volatile` fix as well.

Pavel-Durov: We could also do something like this:
```
let is_volatile = matches!(self.am.inst(op_iid), Inst::Load { volatile: true, .. });
```
But it assumes the instruction defining the operand is the struct load. 
I think that holds for the current test with the inline attribute:
```
%10_0: {0: i8, 64: i64} = load %0_5
%10_2: i8 = extractvalue %10_0, [0]
```
But wont hold for the function call:
```
%10_1: {0: i8, 64: i64} = call make_struct()
%11_2: i8 = extractvalue %10_1, [0]
```

ltratt: AIUI, that's the point: it's only a `volatile` load if it references a `volatile` load; any other instruction can become a non-volatile load.

ltratt:
Let's get this PR done before we think about doing another on top of this one.

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: Volatile updated 157d6016bae014ee8ed1ce3e3fc9fd731614e810

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1831,6 +1831,9 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
         let Ty::Struct(struct_ty) = op.type_(self.am) else {
             panic!()
         };
+        let Inst::Load { volatile, .. } = op.to_inst(self.am) else {
+            todo!("extractvalue from a non-load is not yet supported")
```
ltratt: Let's not burden ourselves with strings like this in `todo`s unless they add value (this one doesn't).

ltratt:
I have one minor comment, but I think this is otherwise ready for squashing? If so, that comment can be addressed at the same time without a separate commit.

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: ok

Pavel-Durov:
> I have one minor comment, but I think this is otherwise ready for squashing? If so, that comment can be addressed at the same time without a separate commit.

todo message removed and squashed 👉  244237f11267571575d69b7327fb9cc63601baaf

PR #2223, opened 2026-07-31
`OUT_DIR`s value has changed in nightly.

ltratt:
This was changed in https://github.com/rust-lang/cargo/pull/17272. This might change back for all I know, but at least for now this unbreaks things.

### DIFF
--- a/tests/src/lib.rs
+++ b/tests/src/lib.rs
@@ -111,7 +111,7 @@ pub fn full_cargo_profile() -> String {
     let out_dir = std::env::var("OUT_DIR").unwrap();
     Path::new(&out_dir)
         .components()
-        .nth_back(3)
+        .nth_back(4)
         .map(|x| x.as_os_str().to_str().unwrap())
         .unwrap()
         .to_owned()
--- a/ykbuild/src/lib.rs
+++ b/ykbuild/src/lib.rs
@@ -18,6 +18,8 @@ pub fn target_dir() -> PathBuf {
         .unwrap()
         .parent()
         .unwrap()
+        .parent()
+        .unwrap()
         .to_owned()
 }
 
--- a/ykbuild/build.rs
+++ b/ykbuild/build.rs
@@ -58,6 +58,8 @@ fn main() {
         .unwrap()
         .parent()
         .unwrap()
+        .parent()
+        .unwrap()
         .to_owned();
     ykllvm_build_dir.push("ykllvm");
     create_dir_all(&ykllvm_build_dir).unwrap();


tests/src/lib.rs
```diff
@@ -111,7 +111,7 @@ pub fn full_cargo_profile() -> String {
     let out_dir = std::env::var("OUT_DIR").unwrap();
     Path::new(&out_dir)
         .components()
-        .nth_back(3)
+        .nth_back(4)
```
Pavel-Durov, line 114: We could do something like:
```rust
pub fn full_cargo_profile() -> String {
    let target_dir = target_dir();
    let out_dir = PathBuf::from(env!("OUT_DIR"));
    out_dir
        .ancestors()
        .find(|p| p.parent() == Some(target_dir.as_path()))
        .and_then(|p| p.file_name())
        .and_then(|n| n.to_str())
}
```
Way more code, but then we don't need to rely on dir layout

ykbuild/src/lib.rs
```diff
@@ -18,6 +18,8 @@ pub fn target_dir() -> PathBuf {
         .unwrap()
         .parent()
         .unwrap()
+        .parent()
+        .unwrap()
         .to_owned()
 }
```
Pavel-Durov, line 24: We can use `cargo_metadata` for that:
```suggestion
pub fn target_dir() -> PathBuf {
    let cargo = std::env::var("CARGO").unwrap_or_else(|_| "cargo".to_owned());
    let metadata = cargo_metadata::MetadataCommand::new()
        .cargo_path(cargo)
        .no_deps()
        .exec()
        .expect("failed to run `cargo metadata`");
    metadata.target_directory.into_std_path_buf().clone()
}
```

tests/src/lib.rs
Pavel-Durov, line 118:
```suggestion
pub fn full_cargo_profile() -> String {
    let out_dir = PathBuf::from(std::env::var("OUT_DIR").unwrap());
    out_dir
        .strip_prefix(ykbuild::target_dir())
        .expect("OUT_DIR is not inside cargo's target directory")
        .components()
        .next()
        .map(|x| x.as_os_str().to_str().unwrap())
        .unwrap()
        .to_owned()
}
```

Pavel-Durov:
Alternative PR github.com/ykjit/yk/pull/2224

ltratt:
Right now, I don't think we can work around this problem other than by setting the `CARGO_UNSTABLE_BUILD_DIR_NEW_LAYOUT=false` variable, so I'm closing this PR.

PR #2224, opened 2026-08-02
Don't rely on `OUT_DIR`'s layout to find the target dir.

Pavel-Durov:
Cargo has changed the layout on nightly
(https://github.com/rust-lang/cargo/pull/17272), so counting path
components back from `OUT_DIR` no longer lands on the profile directory,
and may change again in the future.

In this change we ask `cargo metadata` where the target directory is:
the profile is then the first component of `OUT_DIR` beneath it.
Tested on (`cargo build` and `cargo test`)
```
$ rustc --version
rustc 1.99.0-nightly (73dc9167f 2026-08-01)
```

### DIFF
--- a/bin/yk-config
+++ b/bin/yk-config
@@ -12,6 +12,29 @@ fi
 
 OUTPUT=""
 
+# Find the directory containing `libykcapi.so` under `target/<profile>`.
+#
+# Cargo only guarantees the top-level location when `ykcapi` is built directly; if it's only a
+# dependency of whatever's being built (e.g. a plain `cargo test`), the `.so` can end up nested
+# under `build/ykcapi-<hash>/out/` instead.
+libykcapi_dir() {
+    if [ -n "${YKCAPI_DIR}" ]; then
+        echo "${YKCAPI_DIR}"
+        return
+    fi
+    profile_dir=`realpath ${DIR}/../target/${1}`
+    # Not ideal: a search can't tell us whether the match it finds was built with the features
+    # the caller actually needs, and it's the reason $YKCAPI_DIR exists as a way to skip this.
+    # But this script has no better option here. This is the last-resort fallback for
+    # callers that don't (or can't) supply $YKCAPI_DIR.
+    found=$(find "${profile_dir}/deps" "${profile_dir}/build/"ykcapi* -name libykcapi.so -printf '%T@ %p\n' 2>/dev/null | sort -rn | head -n1 | cut -d' ' -f2)
+    if [ -z "${found}" ]; then
+        1>&2 echo "error: couldn't find libykcapi.so under ${profile_dir}"
+        exit 1
+    fi
+    dirname "${found}"
+}
+
 usage() {
     echo "Generate C compiler flags for building against the yk JIT.\n"
     echo "Usage:"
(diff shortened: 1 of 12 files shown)


ykbuild/build.rs
```diff
@@ -50,15 +50,27 @@ fn main() {
...
+        .expect("OUT_DIR is not inside cargo's target directory")
+        .components()
+        .next()
         .unwrap()
+        .as_os_str()
         .to_owned();
+    let mut ykllvm_build_dir = target_dir.join(profile);
+    assert!(
+        ykllvm_build_dir.is_dir(),
+        "{ykllvm_build_dir:?} is not a directory"
```
ltratt: This string doesn't add anything (in general, assert strings are just noise... there are exceptions, mostly when they're TODOs in disguise).

ykbuild/src/lib.rs
```diff
@@ -11,14 +11,23 @@ pub mod completion_wrapper;
...
-        .parent()
+    let target_dir = cargo_metadata::MetadataCommand::new()
+        .no_deps()
+        .exec()
+        .expect("failed to run `cargo metadata`")
+        .target_directory
+        .into_std_path_buf();
+    let profile = Path::new(env!("OUT_DIR"))
+        .strip_prefix(&target_dir)
+        .expect("OUT_DIR is not inside cargo's target directory")
```
ranger-ross: This could be possible if [`build.build-dir`](https://doc.rust-lang.org/cargo/reference/config.html#buildbuild-dir) is set in `.cargo/config.toml`.

I think `MetadataCommand` also has a `build_directory` that can be used to add handling for that scenario, if its important to you

ltratt:
Unfortunately this PR doesn't work (for the same reasons mine doesn't): you need to `rm -rf target/` to see it though. When you do that you'll get:

```
ld.lld: error: unable to find library -lykcapi
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Basically `libykcapi.so` is no longer put in a predictable place. We'd need to solve that before anything else.

ykbuild/build.rs
Pavel-Durov: removed 9e3f49b6ed1c4825c033b9ac2a1747a00ec5ee3f

bin/yk-config
```diff
@@ -12,6 +12,29 @@ fi
...
+    if [ -n "${YKCAPI_DIR}" ]; then
+        echo "${YKCAPI_DIR}"
+        return
+    fi
+    profile_dir=`realpath ${DIR}/../target/${1}`
+    # Not ideal: a search can't tell us whether the match it finds was built with the features
+    # the caller actually needs, and it's the reason $YKCAPI_DIR exists as a way to skip this.
+    # But this script has no better option here. This is the last-resort fallback for
+    # callers that don't (or can't) supply $YKCAPI_DIR.
+    found=$(find "${profile_dir}" -name libykcapi.so -printf '%T@ %p\n' 2>/dev/null | sort -rn | head -n1 | cut -d' ' -f2-)
```
Pavel-Durov: This is pretty hacky 👀 a filesystem search can't guarantee it finds the right`.so`, it's just picking whatever's newest on disk.

The alternative is to not have `yk-config` guess at all - the interpreter's own build process (yklua's `Makefile`, in this case) could resolve and supply `$YKCAPI_DIR` itself, the same way `tests` now does via cargo's bindeps. But its also not ideal...

ltratt: I agree, this is the least terrible way I can think of too. I _think_ we can make this `find "${profile_dir}/deps" which on my machine (even allowing for FS caching!) speeds the search up from 0.11s to 0.003s.

Pavel-Durov: updated ed0ca0eb7caac3080f2ba5f26cada5ea56e91086

Pavel-Durov: And reverted: 489faf8900f861ca61c0e72267be137d086af511

Pavel-Durov: `find ${profile_dir}/deps -name libykcapi.so` 
doesn't sees libraries like `target/debug/build/ykcapi/$hash/out/libykcapi.so`

Pavel-Durov: This combo 810edfa7ca0709c732fdd541f7481e9540dccb62 seems to work.
tryci is happy for me :)

ltratt: I am not convinced by `debug/build/ykcapi`: I don't think there's any such directory name so that will bork at least sometimes. We could do `debug/build/ykcapi*` I _think_?

Pavel-Durov: added 9c690fb8f590d9fa7ad18a75e5ddc96a652ca921

ltratt:
Please squash.

Pavel-Durov:
Squashed fdaaf6e629f5400b91e29d4d2409e46c10c47652

Pavel-Durov:
Build failed cause I added `...ykcapi*"` to the find glob pattern.

But it failed cause `*` inside double quotes is not expandin, the fix was to move it out side of quotes:
`"${profile_dir}/build/"ykcapi*`.

Squashed again 06a3f256fd148033d92aee2285abcd431dfdb095

PR #2238, opened 2026-08-13
Support extractvalue on register-packed struct returns in j2.

Pavel-Durov:
A small (<=16 bytes) struct returned from functions is packed into `RAX:RDX` across non-inlined calls. 
This change introduces `extractvalue` hir instruction.


ykrt/src/compile/j2/hir.rs
```diff
@@ -1811,6 +1812,56 @@ impl InstT for DynPtrAdd {
     }
 }
 
+#[derive(Clone, Debug)]
+pub(super) struct ExtractValue {
+    pub aggregate: InstIdx,
```
ltratt: To fit in with the rest of HIR let's call these `val` and `off`, but document that the latter is the bit off relative to the start of the storage of `val`.

ykrt/src/compile/j2/hir.rs
```diff
@@ -1811,6 +1812,56 @@ impl InstT for DynPtrAdd {
     }
 }
 
+#[derive(Clone, Debug)]
+pub(super) struct ExtractValue {
```
ltratt: We're quite happy to shorten names in HIR so this could be `ExtractVal`.

ykrt/src/compile/j2/x64/x64regalloc.rs
```diff
@@ -459,6 +459,22 @@ pub(super) const NORMAL_GP_REGS: [Reg; 14] = [
     Reg::RAX,
 ];
 
+pub(super) const GP_REGS_NO_RDX: [Reg; 13] = [
```
ltratt: I think this needs to be `NORMAL_GP_REGS_MINUS_RDX` (or `NO_RDX` as you prefer). It's a mouthful, but that makes it clearer that this set doesn't include _all_ general purpose registers except RDX.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -2329,6 +2377,28 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+            field_bit_off,
+            ..
+        }: &ExtractValue,
+    ) -> Result<(), CompilationError> {
+        // We only support extracting from a register-packed struct return.
+        if !matches!(b.inst(*aggregate), Inst::Call(_)) {
+            panic!();
+        }
+        if *field_bit_off == 0 {
+            self.i_extractvalue_low(ra, iidx, *aggregate)
```
ltratt: Splitting these into subfunctions isn't very helpful IMHO.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -2329,6 +2377,28 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+            aggregate,
+            field_bit_off,
+            ..
+        }: &ExtractValue,
+    ) -> Result<(), CompilationError> {
+        // We only support extracting from a register-packed struct return.
+        if !matches!(b.inst(*aggregate), Inst::Call(_)) {
+            panic!();
+        }
+        if *field_bit_off == 0 {
```
ltratt: I _think_ the code below currently handles structs that are 65..=128 bits wide? And maybe it only handles the case where there are two things to extract, the first of which is 64 bits and the second 1..=64 bits? Or can it handle more general cases?

tests/c/extract_struct_x64.c
```diff
@@ -0,0 +1,128 @@
+// Compiler:
```
ltratt, line 2: I agree this lang_test is a good idea _but_ it's only going to be correct on a subset of platforms (imagine a platform with 128 bit registers: `extractvalue` wouldn't be needed). So we probably need an `ignore-if: !x64 && sysvabi` or similar, but that might be hard to encode: for now we could just document that as `// #` comment if it is too hard. I'd probably even name the file `extract_struct_x64.c` to make that explicit.

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -765,6 +765,73 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
         self.opt.push_ty(tyidx)
     }
 
+    fn p_call_func_ty<'b>(
```
ltratt: I don't fully understand what this function is doing, or why it's separate.

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -765,6 +765,73 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+        &mut self,
+        aot_fty: &'b FuncTy,
+    ) -> Result<(hir::TyIdx, Option<&'b StructTy>), CompilationError> {
+        let aot_rtn_ty = self.am.type_(aot_fty.ret_ty());
+        let (rtn_tyidx, struct_ty) = match aot_rtn_ty {
+            Ty::Struct(struct_ty) => {
+                let bit_size = struct_ty.bit_size();
+                assert!(
```
ltratt: This is an x64-sysv-abi assumption and mustn't appear in aot_to_hir: it should be checked in x64hir_to_asm.

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -765,6 +765,73 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+            .push_ty(hir::Ty::Int(u32::try_from(bit_size.min(64)).unwrap()))?;
+        self.opt.feed(
+            hir::ExtractValue {
+                aggregate: call_iidx,
+                field_bit_off: 0,
+                tyidx: lo_tyidx,
+            }
+            .into(),
+        )?;
+        if bit_size > 64 {
```
ltratt: Again, this is x64 stuff that mustn't appear in aot_to_hir. I also don't think it even needs hardcoding and can be done in a platform neutral way?

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -765,6 +765,73 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+
+    fn p_call_extractvalues(
+        &mut self,
+        call_iidx: hir::InstIdx,
+        struct_ty: Option<&StructTy>,
+    ) -> Result<(), CompilationError> {
+        let Some(struct_ty) = struct_ty else {
+            return Ok(());
+        };
+        let bit_size = struct_ty.bit_size();
```
ltratt: Inside j2 this is called `bitw` (I know the name is different to the AOT IR).

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1837,10 +1909,92 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
         let Ty::Struct(struct_ty) = op.type_(self.am) else {
             panic!()
         };
-        let Inst::Load { volatile, .. } = op.to_inst(self.am) else {
-            todo!()
+
+        if let Operand::Local(aot_iid) = op
```
ltratt: What's this doing?

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1837,10 +1909,92 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+                    struct_ty,
+                    indices[0],
+                    aggregate_iidx,
+                );
+            }
+        }
+        self.p_extractvalue_memory_backed(iid, *tyidx, op, struct_ty, indices)
+    }
+
+    fn p_extractvalue_register_packed(
```
ltratt: What's this doing?

ltratt:
High level comments:

* As with the previous PRs, we're muddling platform idiosyncrasies (specifically x64-sysv-abi) into parts that must be platform independent. aot_to_hir _must_ be platform neutral, unless we find some reason why it can't be (and I'm reasonably sure this particular feature doesn't tip over that threshold).
* Since there's no description of why the PR takes the approach it does, I had to guess, and I couldn't always do so (or did so incorrectly, probably!). My pea brain really does need to see "here's why this takes the approach it does" :) In particular, I _think_ the "memory backed" thing is trying to solve the general `extractval` case? If so, I suggest we _don't_ do that in this PR, at least not in the x64 backend. I would suggest doing the minimum necessary to get the "values returned in two register" case working, then a future PR could do "and store extra stuff on the stack". [Warning: I might be very wrong about this.]
* We need to unit test this stuff. Fortunately it's fairly easy: we minimally extend `hir.l` and `hir.y` and `hir_parser.rs` (you can see this happening in a lot of PRs, so there's lots of precedent). Then we can write x64 unit tests nice and easily.

ykrt/src/compile/j2/x64/x64regalloc.rs
Pavel-Durov: renamed 9b60db8ecda45fc28250b73702b3c3bfbfd3377f

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: ok, 13ceeba3220bf4d17218563eab43156d71c93533

tests/c/extract_struct_x64.c
Pavel-Durov: I think we can use the $YK_ARCH for this condition https://github.com/ykjit/yk/blob/master/tests/langtest_c.rs#L44

Pavel-Durov: 5d99d9a3ad6137b6968843a5719f6b12d64b4518

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: it is a precondition check for register packed extactvalue instruction

tests/c/extract_struct_x64.c
ltratt: Technically we need to test for the SysV ABI too, but this is good enough for now.

ykrt/src/compile/j2/aot_to_hir.rs
ltratt: What are the checks doing?

ykrt/src/compile/j2/hir.rs
Pavel-Durov: renamed ac96a91712d10a874b79b611ec539152ac001a3b

ykrt/src/compile/j2/hir.rs
Pavel-Durov: done 7f82c99f59c92b57f4ca3ba1fe54c58ec364c04e

Pavel-Durov: I used `aggregate` because LLVM extractvalue source operand is documented as "aggregate operand"

ltratt: Yes, I know, but in j2 we conventionally use `val`.

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: Can be inlined.

From my understanding, every instruction in HIR is only allowed to produce one value living in a single register.

So here we tell hir the call produces exactly one value - the first chunk of the register-packed struct.

After the call, p_call_extractvalues adds `ExtractVal` instructions to expose the rest: one for chunk 0 itself and only if the struct doesn't fit in one register a second `ExtractVal` for chunk 1.

ltratt: OK, so I think what you're saying is "this is the `struct` (or other wide value) return type case". That should probably be inlined, unless there's a good reason not to.

Pavel-Durov: inlined 4b9c5454981faf0a98f387fa4021c64bda19a612

ykrt/src/compile/j2/hir.rs
Pavel-Durov: makes sense

Pavel-Durov:
> High level comments:
> 
>     1. As with the previous PRs, we're muddling platform idiosyncrasies (specifically x64-sysv-abi) into parts that must be platform independent. aot_to_hir _must_ be platform neutral, unless we find some reason why it can't be (and I'm reasonably sure this particular feature doesn't tip over that threshold).
> 
>     2. Since there's no description of why the PR takes the approach it does, I had to guess, and I couldn't always do so (or did so incorrectly, probably!). My pea brain really does need to see "here's why this takes the approach it does" :) In particular, I _think_ the "memory backed" thing is trying to solve the general `extractval` case? If so, I suggest we _don't_ do that in this PR, at least not in the x64 backend. I would suggest doing the minimum necessary to get the "values returned in two register" case working, then a future PR could do "and store extra stuff on the stack". [Warning: I might be very wrong about this.]
> 
>     3. We need to unit test this stuff. Fortunately it's fairly easy: we minimally extend `hir.l` and `hir.y` and `hir_parser.rs` (you can see this happening in a lot of PRs, so there's lots of precedent). Then we can write x64 unit tests nice and easily.


1. I _think_ I removed all the x86 specifics from aot_to_hir in 589e53994f7399dc14435306f11191853bc7c3eb

2. I didn't introduce a general memory-backed struct handling here, If you are referring to `p_extractvalue_memory_backed` function - it's the pre-existing `p_extractvalue` body moved into its own function unchanged as in https://github.com/ykjit/yk/pull/2221.

3. on it!

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1246,20 +1273,47 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
             };
 
+            // Build this call's HIR function type.
+            let Ty::Func(aot_fty) = self.am.type_(func.tyidx()) else {
+                panic!()
+            };
+            let aot_rtn_ty = self.am.type_(aot_fty.ret_ty());
+            let (rtn_tyidx, struct_ty) = match aot_rtn_ty {
+                Ty::Struct(struct_ty) => {
+                    // HIR has no aggregate SSA value, so use the call's first register chunk as
```
ltratt: What is "first register chunk"?

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1847,10 +1901,60 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+            if let hir::Inst::Call(_) = self.opt.inst(aggregate_iidx) {
+                return self.p_extractvalue_register_packed(
+                    iid,
+                    *tyidx,
+                    indices[0],
+                    aggregate_iidx,
+                );
+            }
+        }
+        self.p_extractvalue_memory_backed(iid, *tyidx, op, struct_ty, indices)
```
ltratt: I still don't understand `memory_backed` vs `register_packed`. Maybe I'm an idiot!

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -2329,6 +2331,51 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+                self,
+                iidx,
+                [RegCnstr::Cast {
+                    in_iidx: *val,
+                    out_fill: RegCnstrFill::Undefined,
+                    regs: &NORMAL_GP_REGS_MINUS_RDX,
+                }],
+            )?;
+        } else {
+            if val.to_raw_index() + 2 != iidx.to_raw_index() {
```
ltratt: This code needs a comment (I have eventually realised what it's doing, but at first the `+2` seemed like magic to me!).

ltratt: [I'd also do `checked_add` here, which is an implicit `assert` that we don't wrap.]

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -2329,6 +2331,51 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+                todo!("")
+            }
+            let lo_iidx = InstIdx::from_raw_index(iidx.to_raw_index() - 1);
```
ltratt: `checked_sub` here would also be good (in case `iidx` is somehow 0).

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -2329,6 +2331,51 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+            if !matches!(b.inst(lo_iidx), Inst::ExtractVal(lo) if lo.val == *val && lo.off == 0) {
```
ltratt: Probably clearer as `if let Inst::ExtractVal(...) = b.inst(lo_iidx) && lo.val == ...`. [`matches!` is the older spelling.]

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: The code was changed, it handles every field offset now, based on the info from aot instead of hardcoded ranges.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -2329,6 +2331,51 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+            panic!();
+        }
+        if *off == 0 {
+            if val.to_raw_index() + 1 != iidx.to_raw_index() {
+                todo!("")
+            }
+            let [_] = ra.alloc(
+                self,
+                iidx,
+                [RegCnstr::Cast {
```
ltratt: Right, so I now think I can go back to my Cast/Output question. I claim both of these should be the same kind because what we're doing is saying "this value will be in register R". I don't think it matters if we do Cast/Output, but consistency is good.

More importantly, I also _think_ we should say `regs: &[RAX]` for the first extract value and `regs: &[RDX]` as appropriate because that's where the values will be (`i_extractvalue` doesn't move the value, so if it was in RAX, and the register allocator says you've got (say) RCX, bad things are going to happen).

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: `memory_backed` - the struct is the result of a load, its a pointer.

`register_packed` - a small struct (<=16 byte) returned from a function sitting in RAX/RDX registers

I can remove the `packed/backed` terminology

ltratt: Those are x64 & sysv ABI matters: they should be dealt with solely there, and not in aot_to_hir. Imagine we compiled on a 256 bit machine: aot_to_hir should stay unchanged but imaginary25bit_hir_to_asm should be updated to deal with the ABI requirements of function return values.

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: removed

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: removed

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: removed in f7126c9e8228bcaaf33afbfe06c44aa2aeb2af2d

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: index arithmetics removed in f7126c9e8228bcaaf33afbfe06c44aa2aeb2af2d

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: index arithmetics removed in f7126c9e8228bcaaf33afbfe06c44aa2aeb2af2d

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: updated in  f7126c9e8228bcaaf33afbfe06c44aa2aeb2af2d

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: removed in https://github.com/ykjit/yk/commit/f7126c9e8228bcaaf33afbfe06c44aa2aeb2af2d

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: updated 4d2ed17503b60977422cb90624de87aaace80db5

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: updated 3cb4b4bfee6209536364f4f17328823fe1a521ae

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: Two checks:
1. is the operand is a local value 
2. Was it returned from a function

If both hold, we know that we need to handle it via extractvalue_from_call.

tests/c/extract_struct_x64.c
```diff
@@ -0,0 +1,132 @@
...
+//     1
+//     999
+//     2
+//     77
+//     888
+//     yk-tracing: stop-tracing
+//     --- Begin aot ---
+//     ...
+//     %{{10_1}}: {0: i8, 64: i64} = call make_struct()...
+//     ...
```
ltratt, line 19: Dumb question: naively, I expected the pattern would be "CALL; extractvalue; extractvalue" but in this test there seem to be other instructions inbetween them?

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -7473,6 +7511,93 @@ mod test {
         );
     }
 
+    #[test]
+    fn cg_extractvalue() {
+        codegen_and_test(
+            "
+              extern abort() -> i64
```
ltratt: `abort` isn't a great function to use here, especially as its actual return type is `void` :) Right now it _happens_ to work, but one can imagine giving the wrong signature causing weird issues, at least for human readers! The easiest way is just to make up a function e.g. `extern f() -> i64`.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -7473,6 +7511,93 @@ mod test {
...
+
+              %0: ptr = @abort
+              %1: i64 = call abort %0()
+              %2: i64 = extractvalue %1 [0]
```
ltratt: I claim this example is wrong and should cause a validation error in HIR::extractvalue::assert_well_formed. It's saying "I want to extract 128 bits from the 64 bit value at %1", which is obviously not sensible. So as well as fixing this example, `extractvalue::assert_well_formed` needs checks to ensure the user can't express obviously wrong things (and a unit test or two in hir.rs for that). If you look at the bottom of that file you'll see a shedload of `[should_panic]` tests doing just this.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -7473,6 +7511,93 @@ mod test {
...
+              ; %2: i64 = extractvalue %1 [0]
+              ...
+              ; %3: i64 = extractvalue %1 [64]
+              ...
+            "#],
+        );
+    }
+
+    #[test]
+    fn cg_extractvalue_trunc() {
```
ltratt: In general, if you look elsewhere in this file, you'll see that we prefer to put all the tests for a given feature (e.g. `extratvalue`) in one function. There are pros and cons to this, but it will mean we're consistent.

tests/c/extract_struct_x64.c
Pavel-Durov: The blocks looks like this:
```
bb9:
    %9_0: {0: i8, 64: i64} = call make_struct() [statepoint: 2i64, (%0_2, %0_0, %0_1, %0_5, %0_6, %0_7, %0_8)]
    br bb10
  bb10:
    %10_0: ptr = ptr_add %0_7, 0
    %10_1: i8 = extractvalue %9_0, [0]
    *%10_0 = %10_1
    %10_3: ptr = ptr_add %0_7, 8
    %10_4: i64 = extractvalue %9_0, [1]
    *%10_3 = %10_4
    # extract_struct_x64.c:115: fprintf(stderr, "%d\n", s1.a);
```

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1246,17 +1245,41 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
                 FuncMemory::ReadWrite => hir::CallEffects::ReadWrite,
             };
 
+            // Build this call's HIR function type.
```
ltratt, line 1248: I don't fully understand this bit but I _think_ what it's doing is a deliberate hack to work around the fact that HIR doesn't have `struct`s and it's converting "struct S of X bits" to `iX`? If I've got that right, I don't mind that as a hack, but we should document _why_ we're doing this.

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1246,17 +1245,41 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+            let Ty::Func(aot_fty) = self.am.type_(func.tyidx()) else {
+                panic!()
+            };
+            let aot_rtn_ty = self.am.type_(aot_fty.ret_ty());
+            let rtn_tyidx = match aot_rtn_ty {
+                Ty::Struct(struct_ty) => {
+                    // HIR has no aggregate SSA value, so the call result type is just the first field.
+                    let offs = struct_ty.field_bit_offs();
+                    let end = offs
+                        .get(1)
```
ltratt, line 1261: I _think_ this assumes there are only two `extractvalue`s / elements in the struct? If so, we'd need `assert_eq!(offs.len(), 2)`. I also feel a bit queezy about this whole structure, particularly the `unwrap_or_else` where I can't work out why the `struct_ty.bit_size()` is safe (or not).

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1847,10 +1870,35 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
         let Ty::Struct(struct_ty) = op.type_(self.am) else {
             panic!()
         };
-        let Inst::Load { volatile, .. } = op.to_inst(self.am) else {
-            todo!()
-        };
 
+        if let Operand::Local(aot_iid) = op {
+            let aggregate_iidx = self.frames.last().unwrap().get_local(&*self.opt, aot_iid);
```
ltratt: I suspect `val_iidx` would be a better (shorter) name here?

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1847,10 +1870,35 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+            if let hir::Inst::Call(_) = self.opt.inst(aggregate_iidx) {
```
ltratt: What happens if this isn't true?

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1847,10 +1870,35 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+                        hir::ExtractVal {
+                            val: aggregate_iidx,
+                            off,
+                            tyidx: res_tyidx,
+                        },
+                    )
+                    .map(|_| ());
+            }
+        }
+        self.p_extractvalue_from_load(iid, *tyidx, op, struct_ty, indices)
```
ltratt, line 1895: I'm still unsure why sometimes we extract from a call and sometimes a load.

ykrt/src/compile/j2/hir.rs
```diff
@@ -1886,6 +1887,53 @@ impl InstT for DynPtrAdd {
...
+#[derive(Clone, Debug)]
+pub(super) struct ExtractVal {
+    pub val: InstIdx,
+    /// Offset relative to the start of `val`
+    pub off: u32,
+    pub tyidx: TyIdx,
+}
+
+impl InstT for ExtractVal {
+    fn assert_well_formed(&self, _m: &dyn ModLikeT, _b: &dyn BlockLikeT, _iidx: InstIdx) {}
```
ltratt: This needs filling in.

ykrt/src/compile/j2/hir.rs
```diff
@@ -1886,6 +1887,53 @@ impl InstT for DynPtrAdd {
...
+
+    fn rewrite_iidxs<F>(&mut self, _b: &mut dyn BlockLikeT, mut iidx_map: F)
+    where
+        F: FnMut(InstIdx) -> InstIdx,
+    {
+        self.val = iidx_map(self.val);
+    }
+
+    fn to_string<M: ModLikeT, B: BlockLikeT>(&self, _m: &M, _b: &B) -> String {
+        format!("extractvalue %{} [{}]", self.val.to_raw_index(), self.off)
```
ltratt: Should be `extractval` IMHO.

ykrt/src/compile/j2/hir.l
```diff
@@ -22,6 +22,7 @@ cttz "CTTZ"
 dynptradd "DYNPTRADD"
 eq "EQ"
 extern "EXTERN"
+extractvalue "EXTRACTVALUE"
```
ltratt: I think we can (in HIR) consistently use `extractval`. The extra letters are just noise (and we already diverge from LLVM's syntax, so this isn't a big deal).

tests/c/extract_struct_x64.c
ltratt: Hmm, I am a bit worried that the intermedia code between the `extratvalue`s could cause the register allocator to overwrite things. Let me think a bit more about this.

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: yes it is a hack. will add comment.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: ok, I think I just copied from other examples, will update.

Pavel-Durov: I used `random` here 18554f36159ed82dadfe640f58ee9391ce1968a6
Using made up function cause this error:
```
 panic message: "Line 4, column 25:\n  %0: ptr = @f\nSymbol 'f' not found"
```

ltratt: You say `random` but the error message is about `f`?

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: renamed 1b4b6ad14a17bb9b25fba52bc6f05d6c500e88bd

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: I don't see any `[should_panic]` in [x64hir_to_asm.rs](https://github.com/ykjit/yk/blob/master/ykrt/src/compile/j2/x64/x64hir_to_asm.rs) but I get the idea.

ltratt: `assert_well_formed` and the associated tests are in hir.rs.

Pavel-Durov: updated e37613ee333a7cbbf2ab8ee34c691b1585bb8e32

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: grouped e37613ee333a7cbbf2ab8ee34c691b1585bb8e32
I think `[should_panic]` tests needs to stay separate

ykrt/src/compile/j2/hir.rs
Pavel-Durov: added 1b4b6ad14a17bb9b25fba52bc6f05d6c500e88bd

ykrt/src/compile/j2/hir.rs
Pavel-Durov: updated 6e71057670f2d382b380a92a0547b3b340398318

ykrt/src/compile/j2/hir.l
Pavel-Durov: updated 6e71057670f2d382b380a92a0547b3b340398318

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: yes, that's the error I got when using:
```
extern f() -> i64
```

ltratt: I see. OK, so _ideally_ we'd find a standard POSiX function which has a 128 bit return type, but I'm not sure there are any. So let's at least use a function which has a non-void return type (e.g. `getc`), lie about its return type, and add a quick comment about why we're lying about its return type so we don't confuse our future selves.

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: comment updated 82d7d0022c60e13c243d4b60134aaa75deb4395b

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: It takes either the first part of the sturuct if it has multiple fields, or the the whole size if its a single field struct.

Alternatively we could do something like:
```rust
let end = u32::try_from(struct_ty.bit_size().min(64)).unwrap();
self.opt.push_ty(hir::Ty::Int(end))?
```

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: Then we try to parse stuct from load:
```
self.p_extractvalue_from_load(iid, *tyidx, op, struct_ty, indices)
```

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: It comes down to whether the struct returned from an inlined function or not.
This determines whether the struct is located in registers or in memory

Pavel-Durov: aot of inlined function return (see **load** instruction)
```
%10_0: {0: i8, 64: i64} = load %0_5
%10_1: ptr = ptr_add %0_8, 0
%10_2: i8 = extractvalue %10_0, [0]
*%10_1 = %10_2
%10_4: ptr = ptr_add %0_8, 8
%10_5: i64 = extractvalue %10_0, [1]
```
aot of outlined function return (no **load** instruction)
```
%12_0: i64 = call make_struct()
%13_0: ptr = ptr_add %0_8, 0
%13_1: i64 = extractvalue %12_0, [0]
*%13_0 = %13_1
%13_3: ptr = ptr_add %0_8, 8
%13_4: i64 = extractvalue %12_0, [1]
*%13_3 = %13_4
```

ykrt/src/compile/j2/aot_to_hir.rs
ltratt: Why don't we just inline that case? Really I _think_ we're dealing with:

```rust
match self.opt.inst(...) {
  Inst::Call => ...,
  Inst::Load => ...,
  _ => panic!()
}
```

?

ykrt/src/compile/j2/hir.rs
ltratt: Commit seems unrelated?

ykrt/src/compile/j2/aot_to_hir.rs
ltratt: Hang on, are you saying that a struct with one field has an empty sequence of aggregates?

tests/c/extract_struct_x64.c
Pavel-Durov: I think we could avoid this intermediate code if we let `aot_to_hir` handle the chunk-splitting and emit it as dedicated instructions for x64 backend.

ltratt:
I'm not quite sure where we are with this one but I _think_ that last week we agreed that there are two separate PRs conflated in this one:

1. The case where one has a `struct` defined _inside_ a function and one extracts a value from one of its fields.
2. Returning a `struct` from a function.

My proposal would be that this PR should only tackle (1) in part because I _think_ it doesn't interact in scary ways with assumptions in the register allocator.

tests/c/return_struct_x64.c
```diff
@@ -0,0 +1,132 @@
+// ignore-if: test ${YK_ARCH} != "x86_64"
```
ltratt, line 1: I think this test should _probably_ be called `return_struct_x64.c` because that's the challenge it imposes (on the register allocator)?

tests/c/return_struct_x64.c
```diff
@@ -0,0 +1,132 @@
...
+//     1
+//     999
+//     2
+//     77
+//     888
+//     yk-tracing: stop-tracing
+//     --- Begin aot ---
+//     ...
+//     %{{10_1}}: {0: i8, 64: i64} = call make_struct()...
+//     ...
```
ltratt: I think the easiest (ha!) fix is to ensure that `extractvalue`s go next to calls. So what we need aot_to_hir do is always generate:

```
%9: i128 = call ...
%10: i64 = extractval %9, 0
%11: i64 = extractval %9, 64
%12: i64 = load %10
```

Then (say) the x64 backend can easily generate code where we know that RAX/RDX (etc) aren't stomped on.

The obvious challenge with this is that at the point of the `call` we don't know if there will be `extractval`s that use its values. So let's cheat! When we see a `call` that returns a `struct`, always emit all the possible `extractval`s immediately after it. If we later see an actual `extractval` referencing the call, we point it at the already-emitted `extractval`. We can rely on the optimiser and codegen to get rid of any `extractval`s that don't end up getting used.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
ltratt: Yes, should_panic tests _have_ to stay separate! But they're the only ones (in general).

tests/c/return_struct_x64.c
Pavel-Durov: renamed 39403af45e09a885b3f86a866cc89687b0eb3418

ykrt/src/compile/j2/hir.rs
Pavel-Durov: yes, probably copy paste error.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: What's wrong with `extern random() -> i64`?

Pavel-Durov:
> I'm not quite sure where we are with this one but I _think_ that last week we agreed that there are two separate PRs conflated in this one:
> 
>     1. The case where one has a `struct` defined _inside_ a function and one extracts a value from one of its fields.
> 
>     2. Returning a `struct` from a function.
> 
> 
> My proposal would be that this PR should only tackle (1) in part because I _think_ it doesn't interact in scary ways with assumptions in the register allocator.

I don't think we have these two cases in this PR.
Previously we merged https://github.com/ykjit/yk/pull/2221 which dealt with returned structs as `ptr load`, this one deals with returned structs located in registers.

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: Correction for the above - its not about inlined/outlined functions its about the optimisations done by the clang compiler which might eliminate function calls.

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: yes, that's way better!
9ca06e17125f277a4fa2074d308222abac3c5e9e

tests/c/return_struct_x64.c
Pavel-Durov: Ah, that's great!
Updated 60d1c6171348bd74aa20c008c703254292bbccd0

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: No. 
A struct with a single field offset slice is `[0]`, not `[]`. In this case `offs.get(1)` simply returns `None` because there's no second field.

Pavel-Durov:
I think I addressed all the comments in this PR and [this](https://github.com/ykjit/yk/pull/2238#discussion_r3782530098) big blocker :)

ykrt/src/compile/j2/aot_to_hir.rs
ltratt: Dumb question: I don't understand why we'd hardcode `get(1)`, especially as it sometimes doesn't exist. Am I missing out on something?

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -2329,6 +2329,44 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+            let [_] = ra.alloc(
+                self,
+                iidx,
+                [RegCnstr::Output {
+                    out_fill: RegCnstrFill::Undefined,
+                    regs: &[Reg::RAX],
+                    can_be_same_as_input: false,
+                }],
+            )?;
+        } else {
```
ltratt: I _think_ this might be clearer as something like:

```rust
match *off {
  0 => // RAX case
  1 => // RDX case
  2 => todo!() // Bigger struct case
}
```

although now I've written that, I realise I'm not sure if in the sysv ABI a >128 bit struct places its first two elements in RAX/RDX or not!

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -7492,6 +7530,121 @@ mod test {
...
+              %2: i128 = extractval %1 [0]
+              blackbox %2
+              term []
+            ",
+            &[""],
+        );
+    }
+
+    #[test]
+    #[should_panic]
```
ltratt: This test should be aot_to_hir (there's nothing x64 specific about it).

ykrt/src/compile/j2/hir.rs
```diff
@@ -1886,6 +1887,55 @@ impl InstT for DynPtrAdd {
     }
 }
 
+#[derive(Clone, Debug)]
+pub(super) struct ExtractVal {
+    pub val: InstIdx,
+    /// Offset relative to the start of `val`
```
ltratt: "Offset in bytes" I think?

ykrt/src/compile/j2/hir.rs
```diff
@@ -1886,6 +1887,55 @@ impl InstT for DynPtrAdd {
...
+    pub off: u32,
+    pub tyidx: TyIdx,
+}
+
+impl InstT for ExtractVal {
+    fn assert_well_formed(&self, m: &dyn ModLikeT, b: &dyn BlockLikeT, _iidx: InstIdx) {
+        assert!(self.off + m.ty(self.tyidx).bitw() <= b.inst_bitw(m, self.val));
```
ltratt: This still needs expanding/filling-in. If you look at other `assert_well_formed`s you'll see they are of the form `assert!(..., "message in case it went wrong")` otherwise debugging is impossible!

In this case, I think there are at least a couple of well-formedness properties we want:

1. The check you've got above.
2. This `extractval` is preceded by either a `call`, a `load`, or an `extractval`. [Warning: I'm not totally sure about this!]

I _think_ the natural thing is then to extend `assert_well_formed` on hir::Call so that it says something like "if this call is followed by `extractval`s, those do not overlap with each other".

ykrt/src/compile/j2/hir.rs
```diff
@@ -5626,6 +5676,39 @@ mod test {
         );
     }
 
+    #[test]
+    fn extractval_in_bounds() {
```
ltratt, line 5707: We definitely want an example in here of a `call`.

ykrt/src/compile/j2/hir.rs
```diff
@@ -5626,6 +5676,39 @@ mod test {
...
+        str_to_mod::<DummyReg>(
+            "
+          %0: i128 = arg [reg]
+          %1: i64 = extractval %0 [0]
+          %2: i64 = extractval %0 [64]
+        ",
+        );
+    }
+
+    #[test]
```
ltratt, line 5719: `should_panic` tests in aot_to_hir always match on the string that an `assert` fails on (otherwise we might not be testing the property we think we're testing).

ltratt:
We're getting there! None of my new comments are big.

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: Yes good point, updated 2b2d6102e893dba79a6432b76cda82f07e682b3b

ykrt/src/compile/j2/hir.rs
Pavel-Durov: Updated 226a569b5a5184c83b5094777af142662911d68b

ykrt/src/compile/j2/hir.rs
Pavel-Durov: Its in bits. Comment updated 02d3c69ca4e145ab4a0974c235a5bd49ab006ce2

ykrt/src/compile/j2/hir.rs
Pavel-Durov: I think we already have calls in these tests.
https://github.com/ykjit/yk/pull/2238/changes#diff-74837b1c259dedbd5b9a1bddb6b2018f450b5d945abcc4c8b6bd980b7d5a5ed0R5712
```rust
fn extractval_in_bounds() {
        str_to_mod::<DummyReg>(
            "
          extern f() -> i128;
          %0: ptr = arg [reg]
          %1: i128 = call f %0() <------------------ Call
          %2: i64 = extractval %1 [0]
          %3: i64 = extractval %1 [64]
        ",
        );
    }
```

ykrt/src/compile/j2/hir.rs
Pavel-Durov: It's not what we have [here](https://github.com/ykjit/yk/pull/2238/changes#diff-74837b1c259dedbd5b9a1bddb6b2018f450b5d945abcc4c8b6bd980b7d5a5ed0R5720)?

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Removed d065ef320290cf61fba87e6ddbcbc54d897411b7.

Pavel-Durov: `aot_to_hir` has no test module setup as in `hir_to_asm`, and these test cases are not reachable through valid AOT so I didn't readded them.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Updated 02d3c69ca4e145ab4a0974c235a5bd49ab006ce2.

For the sysv ABI, a struct is only returned in registers RAX/RDX if it's ≤128 bits, otherwise it's returned as pointer in RAX.

ykrt/src/compile/j2/hir.rs
ltratt: Mea culpa!

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -7492,6 +7517,90 @@ mod test {
...
+              ; %2: i64 = extractval %1 [0]
+              ...
+              ; %3: i8 = trunc %2
+              ...
+            "#],
+        );
+    }
+
+    #[test]
+    #[should_panic]
```
ltratt: Oops, I attached the `should_panic` comment to the wrong `should_panic`s! Ideally these tests would have a string attached to them even if it's just `todo` or whatnot.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
ltratt: Wrong commit?

ykrt/src/compile/j2/hir.rs
ltratt: I'm glad I asked!

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: ah yes my bad.

Pavel-Durov: Correct commit https://github.com/ykjit/yk/pull/2238/changes/8aaa1302ad16e6bb7250b2f5615b716e39e1943f

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: ok, updated 009497acf3608f9c7637ee132b6fd256f2f72e0a

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -7492,6 +7517,90 @@ mod test {
...
+              %1: i128 = call random %0()
+              %2: i64 = extractval %1 [0]
+              %3: i64 = extractval %1 [64]
+              blackbox %2
+              blackbox %3
+              term []
+            ",
+            &[r#"
+              ...
+              ; %1: i128 = call %0()
```
ltratt, line 7548: I've just realised: this test doesn't really test anything! I think it needs to check that `%2` ends up in RAX and `%3` in RDX. One way to do that might be to have something like `%4: i64 = add %2, %3` and then `blackbox %4` and actually have the x64 code be on RAX and RDX.

This might prove a bit tricky because my guess is that the allocator will probably insert moves. You'll see lots of tests where we just deal with that and maybe we have to here. _But_ we might also be able to easily extend `about_to_process_block` to add precise register hints for `extractval`s (ideally in a way that only adds RAX/RDX as hints for the 128 bit case). Then we (probably) won't end up with moves and the code will be easier to test (and faster).

Pavel-Durov: Added hint + updated test
5d850458ceb9f926b0747cff28725690daf9033b
Followed by small fix, removing `...` in 6f63927ea509757a5e98d1393c0253f6c9edb620

ltratt:
I _think_ this is ready for squashing?

Pavel-Durov:
> I _think_ this is ready for squashing?

It would be my pleasure :)

Pavel-Durov:
Actually this https://github.com/ykjit/yk/pull/2238/changes/e83e28198119840904144021a42868a845a353a4 commit made test fail :)  
Checking

Pavel-Durov:
> Actually this [e83e281](https://github.com/ykjit/yk/commit/e83e28198119840904144021a42868a845a353a4) commit made test fail :) Checking

Fixed ab5869fc1afbb65e5ae134f654d926f207735322

The register-hintonly matched exact offsets `0` and `64`, so any register-unaligned offset (e.g. 32) hit its todo!(). 
Switched to ranges - any offset in `0..64` maps to `RAX` and any in `64..128` to `RDX`.

For example, this struct `{ int32_t a; int32_t b; }` will be packed into a single register RAX and its second field `b` would be extracted at offset `32`.

ltratt:
Please squash.

Pavel-Durov:
Squashed ﻿﻿c2fae0dc1e11c18412de3eff7d5ce16cdb3804e1

Pavel-Durov:
CI failed on clipy:
```
error: you seem to be trying to use `match` for destructuring a single pattern. Consider using `if let`
    --> ykrt/src/compile/j2/aot_to_hir.rs:1288:17
     |
1288 | /                 match aot_rtn_ty {
1289 | |                     Ty::Struct(struct_ty) => {
1290 | |                         let offs = struct_ty.field_bit_offs();
1291 | |                         let field_tyidxs = struct_ty.field_tyidxs();
...    |
1309 | |                     _ => {}
1310 | |                 }
     | |_________________^
```
fixed and squashed again 2f464c0cfd365c2cdd76207101bf8b5af0627527

ltratt:
I don't think that test failure is related (I _think_ I may have seen that test fail once in the past... we should keep an eye on it).

PR #2242, opened 2026-08-14
Don't fail if ykllvm's inst directory already exists.

Pavel-Durov:
`mkdir inst` aborts the script when the directory is already there, which happens if ykllvm is bind-mounted into the container (e.g. by tryci). Changing to `mkdir -p` lets the script to continue with the exeisting directory.

I was using [tryci](https://github.com/softdevteam/tryci) as following:
```shell
tryci -m /vol/.cache/tryci-ykllvm/build:/ci/ykllvm/build,/vol/.cache/tryci-ykllvm/inst:/ci/ykllvm/inst
```

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -117,7 +117,7 @@ cd ykllvm
 ykllvm_hash=$(git rev-parse HEAD)
 cached_ykllvm=0
 if [ -f /opt/ykllvm_cache/ykllvm-release-with-assertions-"${ykllvm_hash}".tgz ]; then
-    mkdir inst
+    mkdir -p inst
     cd inst
     tar xfz /opt/ykllvm_cache/ykllvm-release-with-assertions-"${ykllvm_hash}".tgz
     # Minimally check that we can at least run `clang --version`: if we can't,


ltratt:
I'm not quite sure how this happens, or whether this comes under the heading of "don't do that in the first place", and that we might be dealing with symptoms in this PR and not the real cause. I have wondered for a while whether `tryci` should respect `.gitignore` files and not transfer various things over. Any thoughts?

Pavel-Durov:
> I'm not quite sure how this happens, or whether this comes under the heading of "don't do that in the first place", and that we might be dealing with symptoms in this PR and not the real cause. I have wondered for a while whether `tryci` should respect `.gitignore` files and not transfer various things over. Any thoughts?

I'm using tryci's bind [mount](https://github.com/softdevteam/tryci#using-bindmounts-for-faster-build-times) (-m/--mount) to cache the ykllvm build across runs as:
```shell
tryci -m /vol/.cache/tryci-ykllvm/build:/ci/ykllvm/build,/vol/.cache/tryci-ykllvm/inst:/ci/ykllvm/inst
```
When ran with a mounts the `inst` dir will be already existing in the container - that's what this PR attempts to fix.
Not sure how `.gitignore` will help with this.

.buildbot.sh
```diff
@@ -93,7 +93,8 @@ cd ykllvm
 ykllvm_hash=$(git rev-parse HEAD)
 cached_ykllvm=0
 if [ -f /opt/ykllvm_cache/ykllvm-release-with-assertions-"${ykllvm_hash}".tgz ]; then
-    mkdir inst
+    # inst directory may already exist (e.g. ykllvm bind-mounted into the container with tryci).
```
ltratt: Let's remove this comment in a force push.

Pavel-Durov:
comment removed and squashed bf206ac7dbe1f02b12c0fdf5be910ef80caf5937

PR #2243, opened 2026-08-15
Add a yksompp tests.

Pavel-Durov:
Add yksompp tests to .buildbot.sh.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -12,6 +12,8 @@ YKLUA_REPO="https://github.com/ykjit/yklua.git"
 YKLUA_COMMIT=$(git -C tests/yklua rev-parse HEAD)
 YKCBF_REPO="https://github.com/ykjit/ykcbf.git"
 YKCBF_COMMIT="431b92593180e1e376d08ecf383c4a1ab8473b3d"
+YKSOMPP_REPO="https://github.com/ykjit/yksompp.git"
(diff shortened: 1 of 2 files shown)


ltratt:
I am not sure if this is the way we want to go, if nothing else because I'm not sure yksompp is stable enough yet? [If it isn't stable, having these tests creates a frustrating circular dependency between yk and yksompp that -- from experience with yklua -- is a pain.]

FWIW, what I thought we were going to do was purely in `.buildbot.sh` where we'd do something like `git clone SOM ; cd SOM ; make test`. That would be a much easier thing for us to deal with, even if yksompp isn't itself fully stable.

Pavel-Durov:
> I am not sure if this is the way we want to go, if nothing else because I'm not sure yksompp is stable enough yet? [If it isn't stable, having these tests creates a frustrating circular dependency between yk and yksompp that -- from experience with yklua -- is a pain.]
> 
> FWIW, what I thought we were going to do was purely in `.buildbot.sh` where we'd do something like `git clone SOM ; cd SOM ; make test`. That would be a much easier thing for us to deal with, even if yksompp isn't itself fully stable.

That depends by what do we mean by stable? I think it is stable - as in its not crushing.

I happy to remove it from `cargo test` if we want it to be only in the `.buildbot.sh`.

Pavel-Durov:
Forced pushed 31173a76662c1fc8963fc4653b72aae445df1a72 only .buildbot.sh changes.

ltratt:
Works for me!

PR #2252, opened 2026-08-20
Support `select` between two function operands.

Pavel-Durov:
A select between direct function references keeps the aot function type rather than ptr. 
This change normalises that type to a plain pointer, since a function value is just an address.

This is something came up while working on ykmruby.

### DIFF
--- a/tests/c/select_fnptr.c
+++ b/tests/c/select_fnptr.c
@@ -1,0 +1,66 @@
+// Run-time:
+//   env-var: YKD_LOG_IR=aot,hir
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_LOG=4
+//   stderr:
+//     yk-tracing: start-tracing
+//     5
+//     yk-tracing: stop-tracing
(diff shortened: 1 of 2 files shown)


tests/c/select_fnptr.c
```diff
@@ -0,0 +1,66 @@
...
+//     ...
+//     --- End hir ---
+//     5
+//     yk-execution: enter-jit-code {"trid": "0"}
+//     3
+//     3
+//     yk-execution: deoptimise {"trid": "0", "gidx": "0"}
+//     exit
+
+// Check that `select` between two function operands works.
```
ltratt: I _think_ this should be "between two function _pointers_ works"?

Pavel-Durov: Updated 227042c328ccf93276493e00b32a16b20477ce00

ltratt:
Please squash.

Pavel-Durov:
Squashed 3e947c97a697316df4d28c128837ef9b8499f5f4

PR #2254, opened 2026-08-23
 Add codegen support for llvm.fshl.i16/i32/i64.

Pavel-Durov:
Introduce new `Fshl` hir instruciton that is lowered to x64 `shld`.

This is something that came up while working on ykifying mruby:
https://github.com/mruby/mruby/blob/master/src/etc.c#L227


tests/c/fshl.c
```diff
@@ -0,0 +1,61 @@
...
+//     %{{_}}: i32 = call %{{_}}(%{{_}}, %{{_}}, %{{_}}, %{{hshift_res}}) ; @fprintf
+//     ...
+//     --- End hir ---
+//     4: 9223372036854775808
+//     3: 6917529027641081856
+//     2: 4611686018427387904
+//     1: 2305843009213693952
+//     exit
+
+// Check that the shift-or rotate is supported by the yk compiler.
```
ltratt: "shift-or rotate" doesn't make sense? This is a funnel shift AFAIK.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -3025,6 +3025,97 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+    ) -> Result<(), CompilationError> {
+        let bitw = b.inst_bitw(self.m, *lhs);
+        if let Some(imm) = self.zero_ext_op_for_imm8(b, *shift) {
+            let [lhsr, rhsr] = ra.alloc(
+                self,
+                iidx,
+                [
+                    RegCnstr::InputOutput {
+                        in_iidx: *lhs,
+                        in_fill: RegCnstrFill::Undefined,
```
ltratt: I doubt this is correct. If you have (say) an `i16` with undefined fill bits and funnel shift it using a 32-bit register, I think you'll end up with random bits on the RHS. That suggest `in_fill` should be `Zeroed`. Warning: I've never thought about this type of shift before!

tests/c/fshl.c
Pavel-Durov: removed c677392ac598faa0c49b87d735ca49c130a461cf

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Right. I was specifically relying on i32/i64. 
I didn't implement i16/i8 to keep the scope of this PR smaller, so it will hit `todo!()` at the moment.

ltratt: I don't like this because when someone later hits the `todo` they'll fix it and forget about the `in_fill` case. If you want to punt on this for now, an `assert` _before_ the `ra.alloc` call to test for 32|64 bits is safer.

Pavel-Durov: ok

Pavel-Durov: Implemented over multiple commits.
I tried to simplify the test but clang only emits `llvm.fshl.*` with the zero-guards otherwise it just produces plain shifts.

tests/c/fshl_x64.c
```diff
@@ -1,3 +1,4 @@
+// ignore-if: test ${YK_ARCH} != "x86_64"
```
ltratt: I would hope this test _isn't_ x64 specific. Are we sure it needs to be?

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -3038,14 +3038,16 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
         }: &Fshl,
     ) -> Result<(), CompilationError> {
         let bitw = b.inst_bitw(self.m, *lhs);
+        let needs_mask = bitw == 16;
         if let Some(imm) = self.zero_ext_op_for_imm8(b, *shift) {
+            let imm = if needs_mask { imm & 0xF } else { imm };
```
ltratt: I don't understand why we would mask a 16 bit int in this way: how could `imm` ever not fit within the mask?

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -3058,6 +3060,12 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
                 ],
             )?;
             self.asm.push_inst(match bitw {
+                16 => IcedInst::with3(
+                    Code::Shld_rm16_r16_imm8,
```
ltratt: Uff, I'm generally reluctant to use 16 bit registers. Do we have code which actually exercises this case? Still, it is probably correct, even if I find it a bit ugly!

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -3093,11 +3101,18 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
                         in_iidx: *shift,
                         in_fill: RegCnstrFill::Zeroed,
                         regs: &[Reg::RCX],
-                        clobber: false,
+                        clobber: needs_mask,
                     },
                 ],
             )?;
+            // Instructions are pushed in reverse execution order.
```
ltratt: Please let's stop adding in pointless comments like this in future commits.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -3112,6 +3127,10 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
                 ),
                 x => todo!("{x}"),
             });
+            if needs_mask {
```
ltratt: This look fishy: it seems to be clamping a 16 bit value to a 0xF mask?

ltratt: Ah, I get what it's doing. I think this is pointless: `cl` _is_ 8 bits, so masking like this doesn't achieve anything unless I'm missing something.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Removed 0badc60b3f4baee73ba76780348f6cbf30a9ee88

tests/c/fshl_x64.c
Pavel-Durov: Looking at it now again, probably not.

Pavel-Durov: 52077ca01600d07623bd1e9d860cf260c890019b

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: I was sure that I saw it in mruby... but I checked again and its not required.
Removed 1b1be713678d65ac7d767ab8fc549576a5db3910.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Removed https://github.com/ykjit/yk/commit/1b1be713678d65ac7d767ab8fc549576a5db3910.

ltratt: Just to be clear: I don't want you to remove it just because I asked the question :) Why was it added and why is it OK to remove it?

ltratt: Ah, I see we're removing that whole codepath. Fine, that does mean we can punt on this!

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: mm I thought it was needed cause shld 16-bit masks is mod 32 (not mod 16), I think I got confused here
Removed and it works fine, added test 7e2379bacc1229011ee7ba7fdc2190764595f541

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -3062,6 +3062,103 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+                        regs: &NORMAL_GP_REGS,
+                    },
+                    RegCnstr::Input {
+                        in_iidx: *rhs,
+                        in_fill: RegCnstrFill::Undefined,
+                        regs: &NORMAL_GP_REGS,
+                        clobber: false,
+                    },
+                    RegCnstr::Input {
+                        in_iidx: *shift,
```
ltratt, line 3135: OK, so the correctness question I have here (relative to unfilled bits) is: what is the type of `shift`?

Pavel-Durov: shift has the same type for `lhs` and `rhs` - asserted `assert_well_formed`. 
In practice that's always `i16/i32/i64`, since `i_fshl` panics on anything else.

ltratt: What happens if the type is an `i64` and the runtime value exceeds `cl`'s bounds?

Pavel-Durov: In that case, cl holds the low 8 bits of the runtime value (0–255), and x64 shld masks that mod 64 which is basically runtime value mod 64.

ltratt: That makes sense, but it's not obvious, so let's document it (and maybe try and make it a bit easier to understand, so an idiot like me doesn't have to read it 7 times to work out what's going on :)).

Pavel-Durov: Added 163d7e270a4215ee68aa3e5474c3752f7c750947

ltratt: Elsewhere in the file, these sorts of comments go inside a function, so let's do that.

I don't understand the phrasing "`fshl` wants its shift amount taken modulo the operand's bit width". Taken literally it would suggest that "an i64 with fshl x" would be "x % 64" i.e. always "== x" (except for x == 64 when it's  "== 0").

I think the comment we want looks more like "`fshl` is a funnel shift, which means that undefined bits are problematic. For now we solve this by only implementing types which round to equally-sized registers in x64 (16/32/64 bits) and where we don't have to worry about undefined bits"?

Pavel-Durov: updated 3f01d138a43cb2eeabc8366225c8b13e47f63afb

ltratt:
Please squash.

Pavel-Durov:
> Please squash.

Squashed 48ecd4779dffe8cc1429739ec2c14b9d2888e74c

PR #2262, opened 2026-08-28
Add support for llvm.uadd.with.overflow.

Pavel-Durov:
This change adds support for llvm.uadd.with.overflow.

llvm.uadd.with.overflow returns a struct of `{i32, i1}` shape. Since yk already lowers a
struct-returning call into extractvalue, we reuse that same mechanism here: `aot_to_hir` builds
one `UAddOverflow` instruction that produces the `{sum, flag}` value, and two `ExtractVal`
instructions (exactly as for a real struct-returning call) pull the fields back out - the sum at bit offset 0, the flag at bit offset 32.


ltratt:
Can you add a high-level explanation of why you've taken the approach you've taken? Remember that I will guess incorrectly otherwise!

Pavel-Durov:
> Can you add a high-level explanation of why you've taken the approach you've taken? Remember that I will guess incorrectly otherwise!

Pr description updated.

tests/c/uadd_overflow.c
```diff
@@ -0,0 +1,78 @@
...
+//     %{{uadd_r}}: i32 = extractvalue %{{uadd}}, [0]
+//     ...
+//     %{{uadd_ovz}}: i32 = zext %{{uadd_ov}}, i32
+//     %{{_}}: i32 = call fprintf(%{{_}}, @{{_}}, %{{_}}, %{{uadd_ovz}}, %{{uadd_r}})
+//     ...
+//     --- End aot ---
+//     --- Begin hir ---
+//     ...
+//     %{{h_uadd_r}}: i32 = add %{{h_lhs}}, %{{h_rhs}}
+//     %{{h_uadd_ov}}: i1 = uadd_overflow %{{h_lhs}}, %{{h_rhs}}
```
ltratt: This line confuses me: does `uadd` only return a `u1` or ... ?

Pavel-Durov: Yes, uadd_overflow onlyreturns the overflow flag; the actual sum comes from a separate add instruction right above. This is a workaround to the hir restriction of each instruction only returning a single value.

ltratt: Instead of this, could we try the same `extractval` trick we did for `call`s?

Pavel-Durov: This is a very good point 🤦 
Updated 6e496c3fff76319411830eeaf935deabe8e09c1e

ltratt: There are no yk changes in that commit?

Pavel-Durov: wrong commit, sorry
I meant this one https://github.com/ykjit/yk/pull/2262/changes/f2f39b3fdb386508f34546443bce2ed558ccdb4a
(followed by a few other commits mainly removing comments)

ltratt:
I have some comments first but this will need hir.l/hir.y updates and tests before we can merge.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -2347,30 +2347,52 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+                let [outr] = ra.alloc(
+                    self,
+                    iidx,
+                    [RegCnstr::Cast {
+                        in_iidx: *val,
+                        out_fill: RegCnstrFill::Undefined,
+                        regs: &NORMAL_GP_REGS,
+                    }],
+                )?;
+                if *off != 0 {
```
ltratt: [To make sure that it shows up in the timeline] My comment on this line is the most important: I'm not convinced that what we have here is correct. I _think_ users should never take the output of `uadd_overflow` directly and should only access it via `extractval`?

Pavel-Durov: Added assert 96e6c83ca8149ca98b33d69a764d91c609fceb9b.

Pavel-Durov:
Also found and fixed this bug 0c2be1215a8ef22cefb106bbadff03968184b418

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -2347,30 +2347,75 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+                    [RegCnstr::Output {
+                        out_fill: RegCnstrFill::Undefined,
+                        regs: &[reg],
+                        can_be_same_as_input: false,
+                    }],
+                )?;
+                Ok(())
+            }
+            Inst::UAddOverflow(_) => {
+                if *off == 0 {
```
ltratt: This looks pretty fishy to me: it seems to be saying "uadd_overflow's value will end up in some random register". I think we need to rethink this, for correctness and efficiency.

We have now constrained the IR so that it is (I think) of the shape:

```
%3: i128 = uadd_overflow ...
%4: i64 = extractval %3, [0]
%5: i1 = extractval %3, [64]
```

What would make most sense now is to generate code at the `extractval`s and never at the `uadd` itself. We're sort-of doing that now, but in a way that I'm not sure is safe (and probably doesn't work for 64 bit uadd_overflow?).

What we really want to do here is (assuming the i1 is used!) generate code of the form:

```
add rax, rbx
setc r8b
```

It is a little awkward to do this right now but I think we can. In essence I think we need to do something along the lines of the following where we generated code (in the above example) at %5 and suppress code generation at :

```rust
fn extractval(iidx) {
  if extractval_i1 && is_used(extractval_i1) && is_used(extractval_i64) {
   // The case where we want the integer add and the overflow in registers
    let [_lhsr, _rhsr, i1_outr] = ra.alloc(iidx, [RegCnstr::Input { uadd_overflow.lhs }, ..., RegCnstr::Output { .. });
    let [lhsr, rhsr, add_outr, i1_outr] = ra.alloc(iidx - 1, [..., ..., RegCnstr::Output { .. }, RegCnstr::Clobber { i1_outr });
    // generate the x64 code now
  }
}
```

And then `uadd_overflow` itself wouldn't generate any code.

Why this dance? I _think_ the register allocator needs to see `Output` for both of the `extractval`s (but we can get away without doing it for uadd_overflow), so we need to `ra.alloc` for both `iidx`s. We do it for the second `extractval` first (reverse codegen!), get the register for the `i1`, then allocate registers for the `i64` (or whatever) case, saying that we'll clobber the `i1` register. At that point we have (I think) safely worked out what registers to use and can generate code. As you can tell, this is not something you can blindly implement without _very_ careful checking and testing that it works in all cases one can reasonably think about.

Note: there are optimisations we can do (e.g. if code wants the `i1` but not the `i64`). I wouldn't do those at first but I would `assert`/`todo` for when we see them in the future.

Pavel-Durov: Updated 1571487959d4daf9fe13c9675bdcaca2b2a87774

I only implemented uadd_overflow for the case where both the sum and the carry are used. ykmruby always uses both, so that should be good for now.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -2431,30 +2431,108 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+                    [RegCnstr::Output {
+                        out_fill: RegCnstrFill::Undefined,
+                        regs: &[reg],
+                        can_be_same_as_input: false,
+                    }],
+                )?;
+                Ok(())
+            }
+            Inst::UAddOverflow(UAddOverflow { lhs, rhs, .. }) => {
+                // uadd_overflow sum and carry are two independent values. Codegen for both
```
ltratt, line 2462: Dumb question: will this generate the code _twice_ for each `UAddOverflow` because it generates code for each `extractval` at each `extractval`? [Warning: I might be very wrong here!]

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -7857,6 +7934,32 @@ mod test {
...
+              blackbox %3
+              blackbox %4
+              term [%0, %1]
+            ",
+            &[r#"
+              ...
+              ; %2: i64 = uadd_overflow %0, %1
+              ; %3: i32 = extractval %2 [0]
+              ; %4: i1 = extractval %2 [32]
+              ...
```
ltratt: The `...`s might be right here, but it means I can't work out if my earlier comment is correct or not. Just for my benefit can you paste the complete test output into a comment? It might help me understand if I was wrong above or not!

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Not duplicated. Extractvals are visited in ascendin order, so the sum (off=0) is always visited before the carry (off=32) and it has early return `ra.is_used(carry_iidx) => return Ok(())`.

Pavel-Durov: I can add a comment with short explanation.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov:
```
; l0
sub rsp, 0
; %0: i32 = arg [Reg("r15", Undefined)]
; %1: i32 = arg [Reg("r14", Undefined)]
mov rax, r15
; %2: i64 = uadd_overflow %0, %1
; %3: i32 = extractval %2 [0]
; %4: i1 = extractval %2 [32]
add r15d, r14d
setb r13b
mov r15, rax
; blackbox %3
; blackbox %4
; term [%0, %1]
```

ltratt: Right so the test should IMHO be:

```
              ; %4: i1 = extractval %2 [32]
              add r.32._, r.32._
```

i.e. without the `...` between `%4: i1 = ...` and `add r.32_, ...`.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
ltratt: > the carry (off=32) and it has early return 

Aha, this was what I missed -- sorry! A little comment to point this out, perhaps along the lines of "Codegen for both is emitted at `sum_iidx`" might be all the fix that's needed?

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -2431,30 +2431,108 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+                // is emitted here rather than at the uadd_overflow instruction, which has
+                // no single register that could hold both results.
+                let (sum_iidx, carry_iidx) = match *off {
+                    0 => (iidx, InstIdx::from_raw_index(iidx.to_raw_index() + 1)),
+                    32 => (InstIdx::from_raw_index(iidx.to_raw_index() - 1), iidx),
+                    _ => panic!("uadd_overflow extractval offset {off} must be 0 or 32"),
+                };
+                match *off {
+                    0 if ra.is_used(carry_iidx) => return Ok(()),
+                    0 => todo!("uadd_overflow: sum-only codegen is not supported yet"),
```
ltratt: Again, no need for strings here: `todo!()` is unambiguous. We only add strings when they tell us something that wouldn't be obvious from the code (same comment from the `todo!()` and `assert`s a few lines later)

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -2431,30 +2431,108 @@ impl HirToAsmBackend for X64HirToAsm<'_> {
...
+                    32 if !ra.is_used(sum_iidx) => {
+                        todo!("uadd_overflow: carry-only codegen is not supported yet")
+                    }
+                    _ => (),
+                }
+                assert_eq!(
+                    bitw,
+                    if *off == 0 { 32 } else { 1 },
+                    "uadd_overflow: only 32-bit operands are supported"
```
ltratt: [I think we should get rid of this string but...] isn't the string wrong: `i1` and `i32` is supported?

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -7857,6 +7934,32 @@ mod test {
         );
     }
 
+    #[test]
+    fn cg_uadd_overflow() {
```
ltratt, line 7936: I would like to see two additional tests: one that _only_ uses `sum` and one that _only_ uses `carry`. Just to check that they do the right thing.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Added comment 12075ca8550c188bde52bd5d34b6dd5a954f86c0

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Updated 0f787d3c124890e7de3e897d97b3346c107cfeae

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
ltratt: I meant this should probably go at the top of the block (with the existing multiline comment) so that readers know what they're going to see before they see it.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Removed 09aeb5f062c05c1a3c90dc741d7031edd685a535

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: yes its wrong - removed f6ae93e00512893292d73efb81ba32555abc7ac4

ltratt: No such commit?

Pavel-Durov: Its there, I was just faster then github
https://github.com/ykjit/yk/pull/2262/changes/f6ae93e00512893292d73efb81ba32555abc7ac4

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Added 90c5f13eb4ed64b2a546d4c4b13c9b783f437c4b

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Updated c22439895063ed3c05de1e65d1fe0eadeff358c9

ltratt:
Please format the comment from the last two commits properly and squash at the same time.

Pavel-Durov:
> Please format the comment from the last two commits properly and squash at the same time.

Formatted in https://github.com/ykjit/yk/pull/2262/commits/ed4720d13aba0d8f881300b552b62d4969d00190

ltratt:
I think there is no use in the line break in that comment, and it just wastes space. Please fix that and squash.

Pavel-Durov:
Squashed 5487e643ffdc0d8920cacda400a5ae73275db353

Pavel-Durov:
Merge failed on lua tests failed bercause `llvm.usub.with.overflow.i32` is not yet implemented.
Added `-yk-no-overflow-intrinsics` (until we have the intrinsics in) and squashed again.

PR #2274, opened 2026-09-03
Add support for llvm.usub.with.overflow.

Pavel-Durov:
Mirroring the uadd PR https://github.com/ykjit/yk/pull/2262.

Lower the intrinsic to a new usub_overflow HIR instruction which result may only be consumed via extractval and implemented x64 codegen using sub/setb.


tests/c/overflow_usub.c
```diff
@@ -0,0 +1,79 @@
...
+  yk_mt_hot_threshold_set(mt, 0);
+  YkLocation loc = yk_location_new();
+
+  int i = 5;
+  NOOPT_VAL(loc);
+  NOOPT_VAL(i);
+  while (i > 0) {
+    yk_mt_control_point(mt, &loc);
+
+    // Underflow test only on odd `i`.
```
ltratt: Pointless comment. I also think we can do `unsigned int ua = i % 2 ? 1 : 5` or is the separate variable important?

ykrt/src/compile/j2/hir.rs
```diff
@@ -3174,6 +3175,74 @@ impl InstT for UAddOverflow {
...
+impl InstT for USubOverflow {
+    fn assert_well_formed(&self, m: &dyn ModLikeT, b: &dyn BlockLikeT, iidx: InstIdx) {
+        assert_eq!(
+            b.inst(self.lhs).tyidx(m),
+            b.inst(self.rhs).tyidx(m),
+            "%{iidx:?}: inconsistent lhs / rhs types"
+        );
+    }
+
+    /// Canonicalise operands. Unlike `uadd_overflow`, `lhs` and `rhs` cannot be swapped: `sub`
```
ltratt: Looking at `sub` it doesn't have this comment: I think it's obvious and can be removed.

Pavel-Durov: Removed fcf4ddba769a88be52be05de7460d53cab17b5b7

tests/c/overflow_usub.c
ltratt: I think this is the only thing we need to do before squashing?

Pavel-Durov: Done a0c5ec72abe5a38aa75edb353a2e66f20378982b

ltratt:
Please squash.

Pavel-Durov:
Squashed 9a874559b606fc6d506ce0689c4ff97d21aebbb5

PR #2276, opened 2026-09-03
Add support for llvm.sadd.with.overflow.

Pavel-Durov:
Mirroring previous overflow intrinsic PR https://github.com/ykjit/yk/pull/2274

Lower the `llvm.sadd.with.overflow` intrinsic to a new `sadd_overflow` hir instruction whose result may only be consumed via `extractval`.


ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -267,13 +267,14 @@ impl<'a> X64HirToAsm<'a> {
...
         op_code: Code,
+        set_code: Code,
         name: &str,
     ) -> Result<(), CompilationError> {
         // The wrapped result and the overflow flag are two independent values. Codegen for
-        // both is emitted here rather than at the uadd_overflow/usub_overflow instruction,
-        // which has no single register that could hold both results. Codegen is emitted at
-        // the flag extractval, so the result extractval early-returns if it sees the flag is
-        // also used.
+        // both is emitted here rather than at the sadd_overflow/uadd_overflow/usub_overflow
```
ltratt: Rather than adding each instruction just say at the "add/sub_overflow" or similar maybe? I quite like set notation here so `{sadd, ssub, uadd, usub}_overflow` and that sort of thing, but there's an element of individual taste.

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1511,6 +1511,41 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
         let parts = name.split(".").collect::<Vec<&str>>();
         assert_eq!(parts[0], "llvm");
         match parts[1] {
+            "sadd" if parts[2] == "with" && parts[3] == "overflow" => {
```
ltratt: Since this is more-or-less the same as `uadd` etc can we combine those together here and just `match` on the part that creates `hir::SAddOverflow` etc?

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Updated 8bf7dd284b24bec2548f293b422bb0b05f1f9919

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: Good point. Updated 3d3eacbb06cfa20194e9ae0950f0eafa151af9af

ltratt:
Please squash.

Pavel-Durov:
Squashed 8c79ad53cc4b210e1e2e7f5f60c10ec3e255adcf

PR #2280, opened 2026-09-06
Lower InsertValue in aot_to_hir

Pavel-Durov:
Adding support for a struct return value via a chain of `InsertValue`. 
Structs are never map to a single HIR value, `p_insertvalue` doesn't emit a real HIR instruction - it just extends the field-index already used for struct-returning `Call`s so `p_extractvalue` and `p_store` can resolve fields in the same way.

### DIFF
--- a/tests/c/insertvalue_struct.c
+++ b/tests/c/insertvalue_struct.c
@@ -1,0 +1,101 @@
+// Compiler:
+//   env-var: YKB_EXTRA_CC_FLAGS=-O1
+// Run-time:
+//   env-var: YKD_LOG_IR=aot,hir
+//   env-var: YKD_SERIALISE_COMPILATION=1
+//   env-var: YKD_LOG=4
+//   stderr:
+//     yk-tracing: start-tracing
+//     1
+//     999
+//     yk-tracing: stop-tracing
+//     --- Begin aot ---
+//     ...
+//     func make_struct(%{{_}}: i8, %{{_}}: i64) -> {0: i8, 64: i64} {
+//     ...
+//     %{{parg_a}}: i8 = arg(0)
+//     %{{parg_b}}: i64 = arg(1)
+//     ...
+//     %{{s0}}: {0: i8, 64: i64} = insert_val poison<{0: i8, 64: i64}>, %{{parg_a}}
+//     %{{s}}: {0: i8, 64: i64} = insert_val %{{s0}}, %{{parg_b}}
+//     ret %{{s}}
+//     ...
+//     %{{call_ret}}: {0: i8, 64: i64} = call make_struct(%{{_}}, %{{_}})...
+//     ...
+//     %{{a}}: i8 = extractvalue %{{call_ret}}, [0]
+//     %{{b}}: i64 = extractvalue %{{call_ret}}, [1]
+//     ...
+//     --- End aot ---
+//     --- Begin hir ---
+//     ...
+//     %{{a2}}: i8 = load %{{_}}
+//     %{{b2}}: i64 = load %{{_}}
+//     ...
+//     %{{a2_ext}}: i32 = zext %{{a2}}
+//     ...
+//     %{{_}}: i32 = call %{{_}}(%{{_}}, %{{_}}, %{{a2_ext}}) ; @fprintf
+//     ...
+//     %{{_}}: i32 = call %{{_}}(%{{_}}, %{{_}}, %{{b2}}) ; @fprintf
+//     ...
+//     --- End hir ---
+//     1
+//     999
+//     yk-execution: enter-jit-code {"trid": "0"}
+//     1
+//     999
+//     1
+//     999
+//     yk-execution: deoptimise ...
+//     exit
+
+// Test that the trace builder handles a struct built via a chain of `insertvalue` instructions.
+
+#include <assert.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <yk.h>
+#include <yk_testing.h>
+
+struct S {
+  uint8_t a;
+  uint64_t b;
+};
+
+__attribute__((noinline)) struct S make_struct(uint8_t a, uint64_t b) {
+  struct S ret = {a, b};
+  return ret;
+}
+
+void interp() {
+  YkMT *mt = yk_mt_new(NULL);
+  yk_mt_hot_threshold_set(mt, 0);
+  YkLocation loc = yk_location_new();
+
+  int res = 9998;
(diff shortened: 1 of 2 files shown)


tests/c/insertvalue_struct.c
```diff
@@ -0,0 +1,101 @@
...
+//     999
+//     yk-tracing: stop-tracing
+//     --- Begin aot ---
+//     ...
+//     func make_struct(%{{_}}: i8, %{{_}}: i64) -> {0: i8, 64: i64} {
+//     ...
+//     %{{parg_a}}: i8 = arg(0)
+//     %{{parg_b}}: i64 = arg(1)
+//     ...
+//     %{{s0}}: {0: i8, 64: i64} = insert_val poison<{0: i8, 64: i64}>, %{{parg_a}}
```
ltratt, line 19: We currently don't really handle `posion` properly: it also suggests that the example code might have UB, which we probably don't want!

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -1967,6 +1967,40 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+                let Inst::InsertValue { .. } = agg.to_inst(self.am) else {
+                    panic!()
+                };
+                let prev_iid = agg.to_inst_id();
+                let inst_idx = self.frames.last().unwrap().get_local(&*self.opt, &prev_iid);
+                self.call_extractvals[&inst_idx].clone()
+            }
+            _ => panic!(),
+        };
+        assert!(field_iidxs.len() < num_fields);
```
ltratt: This assertion seems surprising to me: why `<`? Looking a bit later, I wonder if `field_iidx.len() + 1 == num_field` is the more precise assertion?

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -2211,6 +2245,41 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
             panic!()
         };
         let ptr = self.p_operand(tgt)?;
+        if let Ty::Struct(struct_ty) = val.type_(self.am) {
+            let aot_iid = val.to_inst_id();
+            let inst_idx = self.frames.last().unwrap().get_local(&*self.opt, &aot_iid);
+            let field_iidxs = self.call_extractvals[&inst_idx].clone();
```
ltratt: Why `clone`?

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -2211,6 +2245,41 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+            let offs = struct_ty.field_bit_offs();
+            assert_eq!(field_iidxs.len(), offs.len());
+            for (field_iidx, off) in field_iidxs.iter().zip(offs.iter()) {
```
ltratt: I might prefer `bitoff` to make clear that `off` isn't in bytes. [This is only a minor comment: do as you see fit.]

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -2211,6 +2245,41 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
+                assert_eq!(off % 8, 0);
+                let byte_off = u32::try_from(off / 8).unwrap();
+                let field_ptr = if byte_off > 0 {
```
ltratt: No need to do this optimisation here: the optimisation backend will (or at least should!) do it for you.

Pavel-Durov: 4d8e76e69a33ee8347ae75fc92a58272122c1495

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: c41595edc4e2f0a9ea2ee423b292af286218bd34

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: Removed, not needed f761db66ad46a06c6842e04a550159100cc0ca7c

tests/c/insertvalue_struct.c
Pavel-Durov: Just to clarify - the UB concern is about the C test source or j2?

My reading: poison here is just clang's usual way of seeding a struct built up field-by-field . In j2 we never read that poison value, we rebuilt only from the insert elements - so it can't get into the trace.

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: `field_iidxs` grows by one field per InsertValue. Before pushing, its length is just "how many fields inserted so far" - always somewhere between 0 and num_fields - 1. The assert here guarantees there's room for one more: len() < num_fields.

Maybe we don't this assert at all?

ltratt: I agree: replacing the `assert` with a comment (along the lines of what you've just written!) would be more valuable.

tests/c/insertvalue_struct.c
ltratt: > poison here is just clang's usual way of seeding a struct built up field-by-field

Can we double check that's true? If so, I'm fine with it.

ykrt/src/compile/j2/aot_to_hir.rs
Pavel-Durov: c1f8b462a0cd05b90b0e36efc5defd474601dcb1

tests/c/insertvalue_struct.c
Pavel-Durov: Clang emits poison for seeding a struct built up with `-O1`. At `-O0` there's no `insertvalue` at all (it uses alloca/store/load).
Based on https://llvm.org/docs/LangRef.html#poison-values, a poison value only triggers UB if it's actually *used* (branched on, dereferenced as a pointer, etc.). If we don't do anything with the poison value - then there is no UB.

ltratt: > Clang emits poison for seeding a struct built up with -O1. At -O0 there's no insertvalue at all (it uses alloca/store/load).

I did not know that!

ltratt:
Please squash.

Pavel-Durov:
Squashed 5e008dc4bca546097bacb18587cc97dc662c8063

PR #2284, opened 2026-09-12
Codegen single instruction for overflow intrinsics.

Pavel-Durov:
Summary

`llvm.{sadd,uadd,ssub,usub}.with.overflow` previously generated both the result and the overflow flag unconditionally, even when only one was actually used. This change updates `i_overflow` to emit a single instruction when only the result or only the overflow flag is needed.

The optimisation happens purely at x64 codegen, `aot_to_hir` always generates both ExtractVal hir instructions (result and overflow flag) regardless of which one is actually used, so both still appear in hir dumps. The dead code is  only aliminated at register-allocation time.

Mruby

The mruby interpreter's use of `__builtin_*_overflow` lowers to `llvm.{sadd,uadd,ssub,usub}.with.overflow`, which returns a struct holding the result and the overflow flag. This hit a "Multi-locations not yet supported" stackmap panic in yk. 
However, in practice the ykmruby only reads one of the two, never both, so this optimisation bypasses this yk limitation for ykmruby.


ltratt:
We need to find a way of combining these lang tests: they're now so numerous and repetitive that they're overwhelming. I suspect we need just one `.c` file that contains all the operators within it.

Pavel-Durov:
> We need to find a way of combining these lang tests: they're now so numerous and repetitive that they're overwhelming. I suspect we need just one `.c` file that contains all the operators within it.

Combined 58560a968ecc9df1a8de3a395556be07bcb7aef0

tests/c/overflow.c
```diff
@@ -0,0 +1,170 @@
...
+//     yk-execution: deoptimise {"trid": "0", "gidx": "0"}
+//     exit
+
+// Check that llvm.{sadd,uadd,ssub,usub}.with.overflow are supported by the yk when:
+// 1. Only the result is used
+// 2. Only the overflow flag is used
+// 2. Both the result and the overflow flag are used.
+//
+// Note: this only checks behaviour and the HIR (which always keeps both
+// extractvals). The x64 codegen's dead-code elimination of the unused
```
ltratt: This is a platform independent test so x64 shouldn't have anything to do with it.

Pavel-Durov: Removed 9feaabed34f7c49c3d3cfe337cba445788aee12a

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -8104,12 +8158,17 @@ mod test {
               blackbox %3
               term [%0, %1]
             ",
-            &[""],
+            &[r#"
+              ...
+              ; %2: i64 = sadd_overflow %0, %1
+              ; %3: i32 = extractval %2 [0]
+              add r.32._, r.32._
```
ltratt: Quite often we do need to use `_` as a placeholder, but it means weaker tests, and we need to think carefully "could I write a stronger test?" Here, for example, I believe that `add r.32.x, r.32.y` is a correct, meaningfully stronger test (it means that `add rax, rax` will not pass, which it shouldn't if `%0` and `%1` are distinct).

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -348,8 +348,63 @@ impl<'a> X64HirToAsm<'a> {
         };
         match off {
             0 if ra.is_used(flag_iidx) => return Ok(()),
-            0 => todo!(),
-            32 if !ra.is_used(res_iidx) => todo!(),
+            // Only the result is used - the overflow flag is dead.
```
ltratt: I'm struggling to follow the logic here. I _think_ this PR is trying to deal with three cases:

1. The sum and flag are used.
2. The sum is used and the flag is unused.
3. The sum is unused and the flag is used.

But I admit from this code I'm struggling to see how that works. I'm not sure the fixed-width bit offsets are helping with that (hardcoding it to 32 bits does feel pretty weird). Can we find a clearer rephrasing of this code, or is this as good as it gets?

Pavel-Durov: refactored: a3b0c9dd4f371d05e7c6fbdd994ccc75ad7179e8

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Updated b215828fcc41fe6b5363da69347f4f3449cde1f4

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -337,69 +337,131 @@ impl<'a> X64HirToAsm<'a> {
         set_code: Code,
         name: &str,
     ) -> Result<(), CompilationError> {
-        // The wrapped result and the overflow flag are two independent values. Codegen for
-        // both is emitted here rather than at the {sadd, uadd, usub, ssub}_overflow instruction, which
-        // has no single register that could hold both results. Codegen is emitted at the flag
-        // extractval, so the result extractval early-returns if it sees the flag is also used.
+        // {s,u}{add,sub}_overflow returns a {i32 sum, i1 flag} struct, so `off` (the bit
+        // offset the extractval is reading) is always 0 for the result and 32 for overflow flag.
```
ltratt: Again, I'm not comfortable with the hardcoding of `32` in so many places. I'm fine with only _supporting_ 32 bits, but hardcoding it like this -- even in comments -- makes future maintenance harder. We need to find a more neutral way of explaining/implementing this.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -337,69 +337,131 @@ impl<'a> X64HirToAsm<'a> {
...
             32 => (InstIdx::from_raw_index(iidx.to_raw_index() - 1), iidx),
             _ => panic!("{name} extractval offset {off} must be 0 or 32"),
         };
-        match off {
-            0 if ra.is_used(flag_iidx) => return Ok(()),
-            0 => todo!(),
-            32 if !ra.is_used(res_iidx) => todo!(),
-            _ => (),
+        // Identifying which of the values - result, overflow flag, or both - are used, and
+        // performing dead code elimination on whichever isn't, requires seeing both extractvals.
```
ltratt: "seeing both extractvals"?

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -337,69 +337,131 @@ impl<'a> X64HirToAsm<'a> {
         set_code: Code,
```
ltratt, line 346: I have now forgotten what `set_code` and friends are. A doc comment would help.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -337,69 +337,131 @@ impl<'a> X64HirToAsm<'a> {
...
+                        },
+                    ],
+                )?;
+                self.asm
+                    .push_inst(IcedInst::with2(op_code, lhsr.to_reg32(), rhsr.to_reg32()));
+                Ok(())
+            }
+            // Case 2: only the flag is used, the sum is dead. Do the add/sub purely to set the
+            // flag, discarding its result.
+            (32, false, _) => {
```
ltratt: For example, I think that any non-zero value (or perhaps even just `_`) might work here? Warning: I might be wrong.

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Reworded f9e8cd996132e218f46d5695e2e4f5d437ef540b

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: updated 125fbd8f90d9c4240b84f1a3c3245102a89cdb04

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: Added comment cf43a3f7b49eb358da5f9364f3f5efde289dc8b7

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
Pavel-Durov: updated ad3b3c16b83e132e95525647da3078429c0b0487

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -337,69 +340,129 @@ impl<'a> X64HirToAsm<'a> {
...
+                            in_fill: RegCnstrFill::Undefined,
+                            regs: &NORMAL_GP_REGS,
+                            clobber: false,
+                        },
+                        RegCnstr::Clobber { reg: i1_outr },
+                    ],
+                )?;
+                self.asm
+                    .push_inst(IcedInst::with1(set_code, i1_outr.to_reg8()));
+                self.asm
```
ltratt, line 467: OK so here's an interesting case: `to_reg32` implies we are using exactly a 32-bit value. The easiest way to solve this is to add an `assert` that checks that -- when it fails, at least the person who sees the failure will quickly realise what they need to fix.

Pavel-Durov: Added 58ee567e10f6e3bfacb1667434f7db627778d49e

Pavel-Durov: Sorry wrong commit.
This is the right one 6e9138a76fe302dd3c76013c004d2cb0df377a42

ltratt: The `assert` works for me, though I'd prefer it to be moved down to the first point where 32-bitness is assumed. You can do that in the squash though.

ltratt:
Please squash.

Pavel-Durov:
Squashed 9b07cd3b076cd9cba87799c0786acd35c1e16be6

ykrt/src/compile/j2/x64/x64hir_to_asm.rs
ltratt: This comment wasn't addressed in the squash?

Pavel-Durov: Updated https://github.com/ykjit/yk/pull/2284/changes/a0170f5bcd7c2d806744c29d035f082661fe5a82

ltratt:
Please squash.

Pavel-Durov:
Squashed 84199737482a9f7e13f7d478c10136963068ca33

PR #2310, opened 2026-09-18
Add 64-bit register support to overflow-intrinsic codegen.

Pavel-Durov:
i_overflow only emitted 32-bit add/sub for {sadd,uadd,ssub,usub}_overflow. 
This change generalises it to support both 32 and 64 bits.


ykrt/src/compile/j2/x64/x64hir_to_asm.rs
```diff
@@ -8350,6 +8378,74 @@ mod test {
         );
     }
 
+    #[test]
+    fn cg_sadd_overflow_64() {
```
ltratt: We don't split tests up into separate functions like this: we tend to group them (see elsewhere in the file for precedence of how to order these).

Pavel-Durov: I keep on forgetting about this.
Updated 347aa72faef438d92bf0d8b622e7ae7603350793

ltratt:
Please squash.

Pavel-Durov:
Squashed 12cd82dd48b33becc1144a1046ca3c4a1171a1f7

Pavel-Durov:
buildbot failed on unused `x` var in match arm - replaced them with `_` and squashed again.

Pavel-Durov:
buildbot faied on `many_threads_one_loc.c`, which is unrelated to this change. 
I think its a flaky test that needs to be fixing!

ltratt:
Yes, that test needs more looking at: so far I don't know why it sometimes fails.

PR #2313, opened 2026-09-20
Give poison constants a concrete value when lowering 

Pavel-Durov:
Five of the AWFY `ykmruby` benchmarks (Bounce, Havlak, Json, Richards and Storage) crash on `Const::Poison(_) => todo!()`.

Nearly every function in mruby takes an `mrb_state *` as its first argument, whether or not it uses it. 
Under `-O3 -flto`, LLVM can see across translation units. When it proves that a callee never reads a parameter, it stops computing that argument at the call site and passes `poison` instead:

```llvm
tail call void @mrb_free(ptr poison, ptr %p)               ; gc.c, 17 sites
tail call double @mrb_bint_as_float(ptr poison, i64 %x)    ; bigint.c
tail call void @kh_del_set_val(ptr poison, ptr %h, i32 %k) ; set.c
```

## Why this is not undefined behaviour

Poison isn't UB just by existing. In LLVM it is a deferred-error value. Passing it around is fine. It becomes UB only when something depends on it. Here it goes into a parameter the callee never reads, so nothing observes it.

LLVM allows any bit pattern to stand in for poison, so a concrete value is a valid lowering. Here we use the same `0xdeadbeef`pattern that `serialiseUndefValue` already uses, and the recognisable pattern makes it easy to spot if something ever does read it.


Pavel-Durov:
Dealing with poison again in ykmruby 😶

ltratt:
I am not sure that making poison a concrete value _in_ AOT IR is the right thing. Ultimately, j2 should deal with poison, but we can't do that if it's been disappeared. I suspect that `poison` should flow into j2 as `Const::Poison` and it can then choose a concrete value when it has to. However, this might be a bit fiddly to do -- I'm not sure.

Pavel-Durov:
Updated 7e4441f5fcb1f3efb7e4536487f51e7ff92b9dcd

ltratt:
That's a merge commit?

Pavel-Durov:
> That's a merge commit?

which one?

ltratt:
> Updated https://github.com/ykjit/yk/commit/7e4441f5fcb1f3efb7e4536487f51e7ff92b9dcd

That one.

Pavel-Durov:
> 7e4441f

I copied the wrong commit message again, sorry
Its this one https://github.com/ykjit/yk/pull/2313/changes/8b131b15084455df6900adc2c5739899f02515bd

ltratt:
There now seems to be a lot of unnecessary churn in the PR for the change that's needed? AFAICS only the `p_operand` changes are relevant?

Pavel-Durov:
> There now seems to be a lot of unnecessary churn in the PR for the change that's needed? AFAICS only the `p_operand` changes are relevant?

Reduced https://github.com/ykjit/yk/pull/2313/changes/dd1ab623c3003a634a33cef83a847dd57527fd98
and d5e2cad812abdd910a1404baa3d52d1f6d5117c5

ykrt/src/compile/j2/aot_to_hir.rs
```diff
@@ -793,9 +793,24 @@ impl<'a, Reg: RegT + 'static> AotToHir<'a, Reg> {
...
-                match self.am.type_(c.tyidx()) {
+                // LLVM allows any bit pattern to be substituted for poison (using it is undefined
+                // behaviour), so we arbitrarily pick `0xdeadbeef` repeated.
+                let (tyidx, bytes) = match self.am.const_(*cidx) {
+                    Const::Poison(tyidx) => {
+                        let bytes = 0xdeadbeef_u32
+                            .to_ne_bytes()
+                            .into_iter()
+                            .cycle()
+                            .take(self.am.type_(*tyidx).bytew() as usize)
```
ltratt: We prefer `usize::try_from(...).unwrap()` because `as` is a footgun. You can do that at the same time as you squash (i.e. please squash).

Pavel-Durov: updated c1540643aa36b60febeee2cfda09409e0399a5d8

ltratt:
Please squash.

Pavel-Durov:
Squashed 3ac16b0deb63bc3c31152e63f5e046202b3a168c

